PR #30983, opened 2025-01-09
Change CTE order in EXPLAIN output

ggevay:
This PR modifies how EXPLAIN prints Let/LetRec: It used to be `Return` first, and CTEs in bottom-up order, now it's CTEs in top-down order, and then `Return`. The main argument for the old order was that data flowed upwards both inside CTEs and across CTEs. However, the bottom-up order of CTE bindings kept tripping up optimizer developers, so it's probably better to have the CTEs in binding order (top-down). As @mgree [noted](https://materializeinc.slack.com/archives/C063H5S7NKE/p1736447818231129?thread_ts=1736350194.794989&cid=C063H5S7NKE), other databases also print in top-down order.

For example:
```
EXPLAIN
WITH MUTUALLY RECURSIVE
  c0(n int) AS (
    (SELECT n FROM init)
    UNION ALL
    (SELECT * FROM c2)
  ),
  c1(n int) AS (
    SELECT n+n FROM c0 WHERE n<5
  ),
  c2(n int) AS (
    (SELECT * FROM c0)
    UNION ALL
    (SELECT * FROM c1)
    UNION ALL
    (SELECT * FROM c1)
  )
SELECT * FROM c2;
```
used to be
```
Explained Query:
  Return
    Get l2
  With Mutually Recursive
    cte l2 =
      Union
        Get l0
        Get l1
        Get l1
    cte l1 =
      Project (#1)
        Filter (#0 < 5)
          Map ((#0 + #0))
            Get l0
    cte l0 =
      Union
        Project (#0)
          ReadStorage materialize.public.init
        Get l2
```
and now it would be
```
Explained Query:
  With Mutually Recursive
    cte l0 =
      Union
        Project (#0)
          ReadStorage materialize.public.init
        Get l2
    cte l1 =
      Project (#1)
        Filter (#0 < 5)
          Map ((#0 + #0))
            Get l0
    cte l2 =
      Union
        Get l0
        Get l1
        Get l1
  Return
    Get l2
```

The first commit is straightforward: just a few lines of code changes, and a bunch of expected test output rewrites (most of it with auto-rewrite, except for `testdrive`).

The third commit updates the docs, where we show an example EXPLAIN with a CTE.

The second commit is unfortunately not so straightforward. The problem is that the `.spec` test framework's _input_ format is supposed to correspond to EXPLAIN's _output_ format, but now we are reversing EXPLAIN's output format. There are several options on how to handle this situation:
[...]

PR #31061, opened 2025-01-16
feature-benchmark: Old plan for old versions

def-:
Follow-up to https://github.com/MaterializeInc/materialize/pull/31045, follow-up to https://github.com/MaterializeInc/materialize/pull/30983

Awkward since feature benchmark runs old and new Mz versions, so we have to support both.

### Checklist


bobbyiliev (approved):

ggevay:
Oh, this is tricky, thanks for fixing!

Btw., I'm still thinking that we could just confine EXPLAINs to those tests that run in the normal CI. For example, this particular `EXPLAIN SELECT count(*) FROM accumulable;` doesn't seem too important, unless I'm missing something.

def-:
Yes, that would also work. I'm not sure what the motivation was of ensuring the specific plan in benchmarks. My best guess is that the benchmark is trying to bench a specific operator and thus it wants to ensure that the operator is still in the plan that is being executed.

ggevay:
> My best guess is that the benchmark is trying to bench a specific operator and thus it wants to ensure that the operator is still in the plan that is being executed.

Yes, that could certainly make sense in some situations. But for this particular plan, there is nothing specific in it e.g. for a particular way of executing the reduction. (The name of the test somewhat hints at aiming to test `ReductionType::Accumulable`, but the EXPLAIN is not testing that the `ReductionType` actually ends up being `Accumulable` for various reasons: 1. Only the physical plan would show that. 2. The reductions that are actually under test seem to be in the `CREATE VIEW accumulable`, whose plan is not included in this EXPLAIN, because it's not inlined, as it has an index.)

PR #31066, opened 2025-01-16
docs: tweak fallback versions.json path

kay-kim:


def- (approved):
Does this work when you have an empty base URL?

kay-kim:
Yes -- tested locally with `hugo server --disableFastRender --ignoreCache` since hugo defaults to localhost

PR #31071, opened 2025-01-16
catalog: Skip item optimization for renames

jkosh44:
Internally, the catalog models an update by a retraction followed by an
insert. Items are represented internally by their CREATE SQL string. So
when processing an item update, we first remove the existing item,
then parse, plan, and optimize the new SQL string, and then finally
insert the new item.

Parsing, planning, and optimization can be expensive, so we first check
if the CREATE SQL of the addition of an item matches the CREATE SQL of
the retraction of an item. If so, we can skip parsing, planning, and
optimization and just copy the old item. This can happen, for example,
when updating the privileges of an item, because the privileges are not
stored in the item AST.

Sometimes, when objects are renamed, the CREATE SQL string of an item
may change, but it ends up producing the same HIR and MIR. When this
happens we will spend unnecessary time re-optimizing the HIR, only to
produce an equivalent MIR when compared to the previous item. This
scenario is known to happen for `ALTER SCHEMA ... SWAP ...`.

This commit removes the unnecessary optimization work for equivalent
HIRs. After parsing and planning the CREATE SQL string of an item
addition, we check if the resulting HIR matches the HIR of the previous
item. If so, we skip optimization and re-use the previous item's MIR.
Importantly, the skipped optimization step is a local optimization, so
equal HIRs are guaranteed to produce equal MIRs.

### Motivation
This PR adds a known-desirable feature.

### Checklist

### DIFF
--- a/src/adapter/src/catalog.rs
+++ b/src/adapter/src/catalog.rs
@@ -2643,6 +2643,7 @@ mod tests {
                     &create_sql,
                     &BTreeMap::new(),
                     &mut LocalExpressionCache::Closed,
+                    None,
                 )
                 .expect("unable to parse view");
             let commit_ts = catalog.current_upper().await;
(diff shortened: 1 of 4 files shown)


ParkMyCar (approved):

ggevay (approved):
Thank you!

I checked with my repro for the original issue, and the optimizer is now indeed not called on either `alter cluster swap` or `alter schema swap`.

Let's wait for Nightly, and then merge it.

PR #10845, opened 2022-02-23
kafka: immediately cleanup metadata threads on drop

petrosagg:
### Motivation

The metadata thread was previously checking whether the source it was
spanwed for was dropped as the first step in its polling loop.

Since the poll loop by default was 5 minutes it could cause a lot of
idle threads to linger around for a maximum of 5 minutes.

Now the thread parks itself with a timeout instead of sleeping and we
make sure to unpark it when the parent source is dropped. This way the
thread notices that it should exit immediately.

Fixes #10843

### Checklist

- [x] This PR has adequate test coverage / QA involvement has been duly considered.
- [x] This PR adds a release note for any [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/user/content/release-notes.md#what-changes-require-a-release-note).

PR #31072, opened 2025-01-16
storage: round probe timestamps to the probe interval

teskje:
This PR ensures that probe timestamps, and therefore timestamps used for minting new bindings, are rounded down to the probe interval (typically 1s). This is to reduce the amount of distinct timestamps flowing through a dataflow joining multiple sources. Each distinct timestamp induces some amount of overhead, so forcing the timestamps of individual sources to the same values is more efficient.

### Motivation

Closes https://github.com/MaterializeInc/database-issues/issues/8885

### Checklist

### DIFF
--- a/src/storage/src/source/kafka.rs
+++ b/src/storage/src/source/kafka.rs
@@ -69,7 +69,7 @@ use crate::metrics::source::kafka::KafkaSourceMetrics;
 use crate::source::types::{
     Probe, ProgressStatisticsUpdate, SignaledFuture, SourceRender, StackedCollection,
 };
-use crate::source::{RawSourceCreationConfig, SourceMessage};
+use crate::source::{probe, RawSourceCreationConfig, SourceMessage};
 
 #[derive(Clone, Debug, Default, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
 struct HealthStatus {
@@ -1678,8 +1678,10 @@ fn spawn_metadata_thread<C: ConsumerContext>(
                 poll_interval =? poll_interval,
                 "kafka metadata thread: starting..."
             );
+
+            let mut ticker = probe::Ticker::new(poll_interval, config.now_fn);
             loop {
-                let probe_ts = (config.now_fn)().into();
+                let probe_ts = ticker.tick_blocking();
                 let result = fetch_partition_info(
                     &consumer,
                     &topic,
@@ -1737,8 +1739,6 @@ fn spawn_metadata_thread<C: ConsumerContext>(
                 if tx.send((probe_ts, update)).is_err() {
                     break;
                 }
-
-                thread::park_timeout(poll_interval);
             }
 
             info!(
--- a/src/storage/src/source/probe.rs
+++ b/src/storage/src/source/probe.rs
@@ -1,0 +1,96 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! Support for sending frontier probes to upstream systems.
+
+use std::time::Duration;
+
+use mz_ore::now::{EpochMillis, NowFn};
+use mz_repr::Timestamp;
+use tracing::trace;
+
+/// A ticker to drive source upstream probing.
+///
+/// This type works similar to [`tokio::time::Interval`] but returns timestamps from its
+/// [`Ticker::tick`] method that can be used as probe timestamps. These timestamps are rounded down
+/// to the nearest multiple of the tick interval, to reduce the amount of unique timestamps emitted
+/// by sources, thereby reducing churn in downstream dataflows.
+///
+/// The ticker also supports usage in non-async contexts, using [`Ticker::tick_blocking`].
+pub(super) struct Ticker {
+    interval: EpochMillis,
+    now: NowFn,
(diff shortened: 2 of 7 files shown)


src/storage/src/source/kafka.rs
```diff
@@ -1661,92 +1660,87 @@ fn render_metadata_fetcher<G: Scope<Timestamp = KafkaTimestamp>>(
     (metadata_stream, probe_stream, button.press_on_drop())
 }
 
-fn spawn_metadata_thread<C: ConsumerContext>(
+fn spawn_metadata_task<C: ConsumerContext>(
```
teskje, line 1663: The metadata thread is now a tokio task. I'm not sure why it was a thread before. Perhaps it's from a time where we didn't have async support in Timely operators?

antiguru (approved):
Looks reasonable!

src/storage/src/source/probe.rs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        trace!(target, now, "probe ticker skew: {}ms", now - target);
+        self.apply_tick(now)
+    }
+
+    /// Return the desired time of the next tick.
+    fn next_tick_target(&self) -> EpochMillis {
+        let target = match self.last_tick {
+            Some(ms) => ms + self.interval,
```
aljoscha, line 80: Might this over time get out of sync with the `NowFn`/wall-clock time? If we only ever keep adding an interval to a timestamp we got initially.

src/storage/src/source/kafka.rs
```diff
@@ -1737,8 +1739,6 @@ fn spawn_metadata_thread<C: ConsumerContext>(
                 if tx.send((probe_ts, update)).is_err() {
                     break;
                 }
-
-                thread::park_timeout(poll_interval);
```
aljoscha, line 1741: Is the change from `park_timeout` to using `sleep` a behavior change? Or can `sleep` also be woken up the same way? Might be worth considering but I haven't looked into the behavior of either function.

test/testdrive/source-timestamps.td
```diff
@@ -0,0 +1,93 @@
...
+
+$ postgres-execute connection=postgres://mz_system:materialize@${testdrive.materialize-internal-sql-addr}
+ALTER SYSTEM SET kafka_default_metadata_fetch_interval = '1s'
+ALTER SYSTEM SET pg_offset_known_interval = '1s'
+ALTER SYSTEM SET mysql_offset_known_interval = '1s'
+
+> ALTER CLUSTER test SET (REPLICATION FACTOR 0)
+> ALTER CLUSTER test SET (REPLICATION FACTOR 1)
+
+> SELECT name, write_frontier::text::uint8 % 1000 = 1
```
aljoscha, line 65: Are the write frontiers now, say `1`, `1001`, `2001`? My I-like-rounded-numbers brain would prefer `0`, `1000`, `2000`. I could be totally misunderstanding what's happening though.

aljoscha (approved):
I had some inline questions, but nothing should be blocking merging this.

src/storage/src/source/kafka.rs
teskje: It is! The difference is that a parked thread can be woken up by another thread, a sleeping thread cannot. I think in this case we don't need the wakeup-by-other-thread behavior and so `sleep` is the right way, but lmk if I missed something!

src/storage/src/source/probe.rs
teskje: Ah, but we're not doing that! We call the `NowFn` in each tick and set the returned timestamp (rounded down) as `last_tick` to which the interval is then added.

test/testdrive/source-timestamps.td
teskje: Indeed. I think what's happening here is that the source pipeline mints a binding for time, e.g. `1000`, writes out all the data it has for that binding and then steps forward the frontier to indicate that all future updates will be at _some_ later time. We could maybe use our knowledge of the tick interval instead of just stepping forward, but idk what the implications of that would be.

src/storage/src/source/kafka.rs
aljoscha: Park/unpark was introduced here: https://github.com/MaterializeInc/materialize/pull/10845, so it was meaningful. Not sure where things are currently, I think the thread now works somewhat differently. But still might be worth it to keep the immediate-cleanup-on-drop behavior.

teskje: Ah, good to know! Just changing the `sleep` to a `park_timeout` doesn't help much though. The sleep is now in the Ticker which only returns from a tick when the interval is reached, regardless of whether it's woken up before or not. The unpark-on-drop code also doesn't exist anymore.

We could change both of these things, but it doesn't seem worthwhile to pile on complexity just to shut down the thread slightly faster. With RLU the tick interval will be 1s, which is quite different from the 5m interval that existed when the `unpark_timeout` was added.

teskje:
TFTRs!

bosconi:
bugbot run

src/storage/src/source/probe.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// Apply a tick at the given time, returning the probe timestamp.
+    fn apply_tick(&mut self, time: EpochMillis) -> Timestamp {
+        let time = self.round_to_interval(time);
+        self.last_tick = Some(time);
+        time.into()
+    }
+
+    fn round_to_interval(&self, ms: EpochMillis) -> EpochMillis {
+        ms - (ms % self.interval)
+    }
```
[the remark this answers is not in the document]
teskje: https://github.com/MaterializeInc/materialize/pull/35565

PR #31077, opened 2025-01-16
Slow the rate of growth in `ConsolidatingVec`

frankmcsherry:
The growth of `ConsolidatingVec` was by doubling, which meant that we can end up with substantially more memory use than we really need. This PR attempts to make this configurable, growing by a smaller factor determined by a positive integer `slop`. If the result of consolidation is (1 - 1/slop)-full, we grow the vector by a factor of 1 + 1/(slop-1). These relationships were chosen to match the 2x behavior, but it's possible that they are imperfect.

I'm not certain how to test the impact of this, and whether we have "MV rehydration memory" tests, but that would be the right way to see if this is helpful. It comes at the potential expense of CPU.

### Checklist

### DIFF
--- a/src/compute-types/src/dyncfgs.rs
+++ b/src/compute-types/src/dyncfgs.rs
@@ -99,11 +99,12 @@ pub const LGALLOC_SLOW_CLEAR_BYTES: Config<usize> = Config::new(
     "Clear byte size per size class for every invocation",
 );
 
-/// The denominator `n` in the growth rate `1 + 1/(n-1)` for `ConsolidatingVec` growth.
-pub const CONSOLIDATING_VEC_GROWTH_DENOMINATOR: Config<usize> = Config::new(
-    "consolidating_vec_growth_denominator",
-    2,
-    "Denominator in growth rate for consolidating vector size",
+/// The term `n` in the growth rate `1 + 1/(n + 1)` for `ConsolidatingVec`.
+/// The smallest value `0` corresponds to the greatest allowed growth, of doubling.
+pub const CONSOLIDATING_VEC_GROWTH_DAMPENER: Config<usize> = Config::new(
+    "consolidating_vec_growth_dampener",
+    0,
+    "Dampener in growth rate for consolidating vector size",
 );
 
 /// The number of dataflows that may hydrate concurrently.
@@ -186,5 +187,5 @@ pub fn all_dyncfgs(configs: ConfigSet) -> ConfigSet {
         .add(&ENABLE_COMPUTE_REPLICA_EXPIRATION)
         .add(&COMPUTE_REPLICA_EXPIRATION_OFFSET)
         .add(&COMPUTE_APPLY_COLUMN_DEMANDS)
-        .add(&CONSOLIDATING_VEC_GROWTH_DENOMINATOR)
+        .add(&CONSOLIDATING_VEC_GROWTH_DAMPENER)
 }
--- a/src/compute-types/src/dyncfgs.rs
+++ b/src/compute-types/src/dyncfgs.rs
@@ -99,6 +99,13 @@ pub const LGALLOC_SLOW_CLEAR_BYTES: Config<usize> = Config::new(
     "Clear byte size per size class for every invocation",
 );
 
+/// The denominator `n` in the growth rate `1 + 1/(n-1)` for `ConsolidatingVec` growth.
+pub const CONSOLIDATING_VEC_GROWTH_DENOMINATOR: Config<usize> = Config::new(
+    "consolidating_vec_growth_denominator",
+    2,
+    "Denominator in growth rate for consolidating vector size",
+);
+
 /// The number of dataflows that may hydrate concurrently.
 pub const HYDRATION_CONCURRENCY: Config<usize> = Config::new(
     "compute_hydration_concurrency",
@@ -179,4 +186,5 @@ pub fn all_dyncfgs(configs: ConfigSet) -> ConfigSet {
         .add(&ENABLE_COMPUTE_REPLICA_EXPIRATION)
         .add(&COMPUTE_REPLICA_EXPIRATION_OFFSET)
         .add(&COMPUTE_APPLY_COLUMN_DEMANDS)
+        .add(&CONSOLIDATING_VEC_GROWTH_DENOMINATOR)
 }
--- a/src/compute/src/sink/correction.rs
+++ b/src/compute/src/sink/correction.rs
@@ -358,10 +358,11 @@ impl<D: Ord> ConsolidatingVec<D> {
             let length = self.data.len();
             let dampener = self.growth_dampener;
             if capacity < length + length / (dampener + 1) {
-                // Increase as a function of capacity rather than length.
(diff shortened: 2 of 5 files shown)


frankmcsherry:
From Slack, @teskje recommends a launch darkly flag to allow the number to be configurable, in case two was very important for some reasons (plausible!). I'll need to chase down what threading that through looks like.

src/compute/src/sink/correction.rs
```diff
@@ -304,13 +312,22 @@ pub(crate) struct ConsolidatingVec<D> {
     /// A lower bound for how small we'll shrink the Vec's capacity. NB: The cap
     /// might start smaller than this.
     min_capacity: usize,
+    /// Denominator in the growth rate, where 2 corresponds to doubling, and `n` to `1 + 1/(n-1)`.
+    ///
+    /// If consolidation didn't free enough space, at least a linear amount, increase the capacity
+    /// The `slop` term describes the rate of growth, where we scale up by factors of 1/slop.
```
teskje, line 318:
```suggestion
    /// The `slop` term describes the rate of growth, where we scale up by factors of 1/(slop-1).
```

... right?

src/compute-types/src/dyncfgs.rs
```diff
@@ -99,6 +99,13 @@ pub const LGALLOC_SLOW_CLEAR_BYTES: Config<usize> = Config::new(
     "Clear byte size per size class for every invocation",
 );
 
+/// The denominator `n` in the growth rate `1 + 1/(n-1)` for `ConsolidatingVec` growth.
+pub const CONSOLIDATING_VEC_GROWTH_DENOMINATOR: Config<usize> = Config::new(
+    "consolidating_vec_growth_denominator",
+    2,
```
teskje, line 105: I'd be okay with defaulting this to our target (`11`?) and then opting out were we observe issues. Seems safe enough, considering that this trades of less memory for more CPU and we usually don't have CPU issues.

src/compute/src/sink/correction.rs
```diff
@@ -47,17 +47,24 @@ pub(super) struct Correction<D> {
     metrics: SinkMetrics,
     /// Per-worker persist sink metrics.
     worker_metrics: SinkWorkerMetrics,
+    /// Configuration for `ConsolidatingVec` determining the rate of growth (doubling, or less).
+    growth_denominator: usize,
```
teskje, line 51: There isn't anything that enforces "doubling or less" here, right? People could set the config to `1` or `0` and end up having a bad time. How about checking that in `new` and falling back to `2` on invalid values? Alternatively, redefine the semantics to have `0` mean doubling.

teskje (approved):
Looks good! We could tighten down the API to prevent users from setting too-small denominator values.

src/compute/src/sink/correction.rs
frankmcsherry: Yes, but I think that term is *added* to the current capacity, so the factor is more like `slop/(slop - 1)`. I'll fix though!

src/compute-types/src/dyncfgs.rs
frankmcsherry: What do you think about starting this way, then trying out some new values, then a future item that cuts us over to something like 11 as a default? I'm flexible, but if you think that starting at 11 is the right way to make sure we actually test this and do it I'm happy to make the change!

src/compute/src/sink/correction.rs
frankmcsherry: Nope. Good point. This might be a great change also because `2` only coincidentally means "doubling", and perhaps we should break that association.

src/compute-types/src/dyncfgs.rs
teskje: No strong opinion from me either! If you want to go more slowly that's fine with me.

frankmcsherry:
PR updated, with the general theme of `denominator` turning in to `dampener`. One "surprise" was that `reserve` and `reserve_exact` are different things, and the former will proactively double the allocations (or .. something but seemingly doubling). This meant that before changing to `reserve_exact` we would
1. Allow a denser packing of the consolidating vector before increasing the capacity, but
2. Not grow more any more slowly.
I haven't entirely gotten my head around what this means, but I switched it over to `reserve_exact` just in case.

The main unknown for me at the moment is exactly what happens with odd settings of the dampening. For example, for a large value, we'll simply do nothing until the capacity exceeds the dampening, which means we'll potentially compact in place for each additional update, if they keep consolidating. Not the worst because that is what the dampening means, but also not the best we could do.

The main *known* problem is that this doesn't exactly track the current behavior any more, because `reserve_exact` doesn't have to double, which means whatever background doubling was going on will only be approximated with a dampener of zero. We should discuss this before merging!

src/compute-types/src/dyncfgs.rs
frankmcsherry: I think this is a bigger question now that it seems we aren't tracking the standard Rust behavior when we increase the capacity. We could try and sort that out to make it look like doubling (e.g. by ensuring that the capacity grows by `1 + 1/(damp + 1)`, by doing the math to hit this target based on the current length).

frankmcsherry:
The most recent commit `Target a specific capacity` attempts to get this right, by growing the capacity by 1+1/(n+1) and then subtracting the current length before calling `reserve_exact`, which should arrive at the intended capacity, more or less. Specifically, when `n` is zero, we intend that this doubles the capacity.

src/compute-types/src/dyncfgs.rs
frankmcsherry: Most recent commit tries to handle this.

src/compute/src/sink/correction.rs
```diff
@@ -326,21 +342,27 @@ impl<D: Ord> ConsolidatingVec<D> {
...
+            // We may need more capacity if our current capacity is within `1+1/(n+1)` of the length.
+            // This corresponds to `cap < len + len/(n+1)`, which is the logic we use.
+            let length = self.data.len();
+            let dampener = self.growth_dampener;
+            if capacity < length + length / (dampener + 1) {
+                // We would like to increase the capacity by a factor of `1+1/(n+1)`, which involves
+                // determining the target capacity, and then reserving an amount that achieves this
+                // while working around the existing length.
+                let new_cap = capacity + capacity / (dampener + 1);
+                self.data.reserve_exact(new_cap - length);
```
antiguru, line 365: We're starting at 128, so growing from the small size slowly will prevent us from ramping up quickly where it doesn't matter much -- I think everything below a few dozen KiB could just double. Is there a way we can get a fast start behavior that eventually switches over to slow growth?

frankmcsherry: We could, though by the same token I'm not sure that it matters there, for the same reason. I'm comfortable with whatever, but would love to not be in a state where we do not have to noodle on it too hard. The win of the early doubling would be reduced CPU utilization, and the risk would be .. probably mostly concern about whether we really nailed the logic for cutting over correctly, probably?

frankmcsherry:
Thanks for the review folks!

PR #31080, opened 2025-01-16
[persist] Structured file format

bkirwi:
### Motivation

Adds a new on-disk file format for Persist - the non-dual-write structured-data-only version.

We'd like this format to be supported in our on-prem releases, so folks aren't stuck with the migration midpoint.

### Tips for reviewer

I've intentionally left this off in CI, out of the parallel workload. For our existing cloud envs I think it's most important to verify that the (substantial) refactorings here don't cause any regression for existing flag settings. However, it is enabled in parallel workload so we should be able to spot any bugs. (And empirically those tests are good at finding bugs in this corner of Persist.)

### Checklist

### DIFF
--- a/src/persist-client/src/fetch.rs
+++ b/src/persist-client/src/fetch.rs
@@ -736,6 +736,7 @@ pub struct FetchedPart<K: Codec, V: Codec, T, D> {
     diffs: Int64Array,
     migration: PartMigration<K, V>,
     filter_pushdown_audit: Option<LazyPartStats>,
+    peek_stash: Option<((Result<K, String>, Result<V, String>), T, D)>,
     part_cursor: usize,
     key_storage: Option<K::Storage>,
     val_storage: Option<V::Storage>,
@@ -864,6 +865,7 @@ impl<K: Codec, V: Codec, T: Timestamp + Lattice + Codec64, D> FetchedPart<K, V, 
             metrics,
             ts_filter,
             part,
+            peek_stash: None,
             timestamps,
             diffs,
             migration,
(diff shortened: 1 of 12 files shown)


def-:
> I've intentionally left this off in CI, out of the parallel workload. For or existing cloud envs I think it's most important to verify that the (substantial) refactorings here don't cause any regression for existing flag settings. However, it is enabled in parallel workload so we should be able to spot any bugs. (And empirically those tests are good at finding bugs in this corner of Persist.)

One risk here is that what parallel workloads finds is usually harder to reproduce than other tests. I'll trigger a one-off run at least to see what would happen if we default all of CI to structured.

bkirwi:
Aight! Note that feature benchmarks and upgrade tests are very likely to fail; unsure about the rest of them.

def-:
That kind of failure is ok. I'm more hoping to catch some wrong result or panic.

def- (approved):
No wrong results, no panics, just some timeouts, and the expected upgrade errors. So all good from my side

bkirwi:
Found and fixed the performance regression - this should be ready for review.

([Nightly run with clean benchmarks.](https://buildkite.com/materialize/nightly/builds/10920) The only failure is in data ingest, which I think is the Azure perf regression?)

def-:
> ([Nightly run with clean benchmarks.](https://buildkite.com/materialize/nightly/builds/10920) The only failure is in data ingest, which I think is the recent Azure issue?)

Correct, please ignore.

src/persist-client/src/fetch.rs
```diff
@@ -781,96 +786,91 @@ impl<K: Codec, V: Codec, T: Timestamp + Lattice + Codec64, D> FetchedPart<K, V,
             None
         };
 
-        // TODO(parkmycar): We should probably refactor this since these columns are duplicated
-        // (via a smart pointer) in EncodedPart.
```
ParkMyCar, line 785: woohoo! love seeing this get deleted!

src/persist-client/src/fetch.rs
```diff
@@ -915,74 +915,85 @@ where
...
-            {
-                if (k, v) != (k_next, v_next) {
-                    break;
+        let mut consolidated = self.peek_stash.take();
+        loop {
+            // Fetch and decode the next tuple in the sequence. (Or break if there is none.)
+            let next = if self.part_cursor < self.timestamps.len() {
+                let next_idx = self.part_cursor;
+                self.part_cursor += 1;
+                let mut t = T::decode(self.timestamps.values()[next_idx].to_le_bytes());
```
ParkMyCar, line 924: This is probably fine, but I'm a bit surprised to see a `.to_le_bytes()` here? Feels like a bit of an abstraction leakage

src/persist-client/src/fetch.rs
```diff
@@ -915,74 +915,85 @@ where
...
+                if !self.ts_filter.filter_ts(&mut t) {
+                    continue;
                 }
-
-                if !self.ts_filter.filter_ts(&mut t_next) {
-                    break;
+                let d = D::decode(self.diffs.values()[next_idx].to_le_bytes());
```
ParkMyCar, line 928: same as above, surprised to see `.to_le_bytes()`

src/persist-client/src/fetch.rs
```diff
@@ -915,74 +915,85 @@ where
...
+                if d.is_zero() {
+                    continue;
                 }
-                if t != t_next {
+                let kv = if result_override.is_none() {
+                    self.decode_kv(next_idx, key, val)
+                } else {
+                    // This will be overridden later - just leave a placeholder here for now.
```
ParkMyCar, line 935: Mind mentioning that this will be overridden below? At first read I thought it meant at a different layer in the stack

ParkMyCar (approved):
WOOHOO!

src/persist-client/src/fetch.rs
bkirwi: Added a comment about this... it's an artifact of not always reading data from `ColumnarRecords`, where this code used to live. I've left a comment that we should re-encapsulate it once we're all in on structured data, if that feels like enough to cover it?

PR #31097, opened 2025-01-17
adapter/storage: Cast MySQL `bit` columns to `uint8`, add convience functions

ParkMyCar:
This PR changes the MySQL source to support ingesting the `bit` type as `uint8`. It also adds two new Postgres functions, `bit_count(bytea)` and `get_bit(bytea, int32)` to making working with byte strings easier.

### Motivation

Progress towards: https://github.com/MaterializeInc/database-issues/issues/8891

### Checklist


src/adapter/BUILD.bazel
```diff
@@ -24,7 +24,6 @@ rust_library(
     compile_data = [],
     crate_features = ["default"],
     data = [],
-    disable_pipelining = True,
```
jkosh44: Did you mean to include this in this PR? It seems a bit unrelated.

src/expr/src/scalar/func.rs
```diff
@@ -1361,6 +1361,26 @@ fn power_numeric<'a>(a: Datum<'a>, b: Datum<'a>) -> Result<Datum<'a>, EvalError>
...
+    let index = usize::try_from(index).map_err(|_| err.clone())?;
+
+    let byte_index = index / 8;
+    let bit_index = index % 8;
+
+    let i = bytes
+        .get(byte_index)
+        .map(|b| (*b >> bit_index) & 1)
+        .ok_or(err)?;
+    Ok(Datum::from(i32::from(i)))
```
jkosh44, line 1381: If I'm understanding correctly, the following might be useful: `assert!(i == 0 || i == 1)`

jkosh44 (approved):
LGTM, but I think we should add docs for the new functions.

src/adapter/BUILD.bazel
ParkMyCar: I didn't was rebased on the wrong branch by accident, removed!

src/expr/src/scalar/func.rs
ParkMyCar: Good call!

ParkMyCar:
> LGTM, but I think we should add docs for the new functions.

Good idea, I'll add them in a separate PR though so we can merge them when v0.131 goes out

def- (approved):
Thanks for adding tests!

PR #31106, opened 2025-01-18
bazel: tidy up `bin/bazel` invocations a bit

ParkMyCar:
Wrap the call to `subprocess` in a try-catch so when a Bazel build fails or is canceled we don't print the Python backtrace.

### Motivation

Makes `bin/bazel` a little nicer.

### Checklist

### DIFF
--- a/misc/python/materialize/cli/bazel.py
+++ b/misc/python/materialize/cli/bazel.py
@@ -116,7 +116,10 @@ def integrity_cmd(config: BuildConfig, args: list[str]):
 def bazel_cmd(config: BuildConfig, args: list[str]):
     """Forwards all arguments to Bazel, possibly with extra configuration."""
     remote_cache = remote_cache_arg(config)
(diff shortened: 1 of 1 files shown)


def- (changes requested):
What kind of side effects does it have that we continue running at this point? The command will also return 0 (success) now even when a failure happened. I'd suggest at least passing the error through and returning 1 from main, or alternatively catching the exception in main and returning the error there, without ever printing the python backtrace.

ParkMyCar:
> What kind of side effects does it have that we continue running at this point? The command will also return 0 (success) now even when a failure happened. I'd suggest at least passing the error through and returning 1 from main, or alternatively catching the exception in main and returning the error there, without ever printing the python backtrace.

There shouldn't be any side effect since this function is solely a wrapper around invoking `bazel` at the command line. But other folks in the future might use this function in a different way so IMO exiting with 1 is the right move.

jkosh44 (approved):

def- (approved):
Thanks

PR #31108, opened 2025-01-18
Add Mandelbrot set to WMR slt

ggevay:
The first commit adds a WMR test, drawing an ASCII art of the Mandelbrot set. The query is based on an [SQLite example](https://www.sqlite.org/lang_with.html#outlandish_recursive_query_examples), pointed out by @umanwizard.

Had to tweak the query a bit:
- `WITH MUTUALLY RECURSIVE` instead of `WITH RECURSIVE`
- explicit types for each CTE
- `least` instead of `min` (`min` is the aggregation function in Materialize)
- `string_agg` instead of `group_concat`
- `chr(10)` instead of `x'0a'`
- added `ORDER BY` to the `string_agg`, because otherwise Materialize chooses a non-useful order inside each group.

The second commit fixes a bug in sqllogictest: When the expected output starts with a space, the `trim_start()` used to eat this. The `trim_start` was there, because the `split_at(&QUERY_OUTPUT_REGEX)` ends at the end of `----`, leaving a newline there to be returned by the `split_at` that computes the `output_str`. Now we take care to eat just a newline instead of `trim_start` eating all whitespace.

(It would be tempting to just make `QUERY_OUTPUT_REGEX` eat the newline after `----`. However, this would cause an issue for tests like `joins.slt:753`, where the expected output doesn't have any rows, and there is only one empty line. We expect two newlines to end the expected output, but in this test one of these newlines is the one directly after `----`, so if `QUERY_OUTPUT_REGEX` eats this newline, then we get into trouble. One could just tweak this and other similar tests, but I wanted to make the bugfix mostly backwards compatible with existing tests. The one change that I still needed to make in an existing test was arguably in a malformed test.)

For some reason, the change in sqllogictest auto-summoned the Adapter team. But I guess it would be better if @def- reviewed the sqllogictest change. @ParkMyCar, feel free to remove yourself and the Adapter team from the reviewers.

### Motivation

  * The Mandelbrot set is cool.

### Checklist

### DIFF
--- a/src/sqllogictest/src/parser.rs
+++ b/src/sqllogictest/src/parser.rs
@@ -252,13 +252,27 @@ impl<'a> Parser<'a> {
         static HASH_REGEX: LazyLock<Regex> =
             LazyLock::new(|| Regex::new(r"(\S+) values hashing to (\S+)").unwrap());
         let sql = self.split_at(&QUERY_OUTPUT_REGEX)?;
-        let mut output_str = self
-            .split_at(if multiline {
-                &EOF_REGEX
-            } else {
-                &DOUBLE_LINE_REGEX
-            })?
-            .trim_start();
+        let mut output_str = self.split_at(if multiline {
+            &EOF_REGEX
+        } else {
+            &DOUBLE_LINE_REGEX
+        })?;
+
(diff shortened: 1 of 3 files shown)


def- (approved):
lgtm, assuming https://buildkite.com/materialize/nightly/builds/10917 is green

PR #31125, opened 2025-01-21
storage: make probe intervals runtime configurable

teskje:
This PR makes it possible to change the tick interval of the upstream frontier probes without a source restart.

To simplify things I've added a first commit that moves the Kafka interval configuration into dyncfg and starts ignoring the user-defined per-source interval (which is to be removed in https://github.com/MaterializeInc/database-issues/issues/8887).

### Motivation

Closes https://github.com/MaterializeInc/database-issues/issues/8886

### Checklist


src/storage-types/src/parameters.proto
```diff
@@ -18,7 +20,6 @@ import "service/src/params.proto";
 import "tracing/src/params.proto";
 
 message ProtoStorageParameters {
-  reserved 1, 2, 3, 25, 12, 16, 26, 31, 39;
```
antiguru, line 21: We could re-use the small identifiers, they have a more efficient proto encoding.

antiguru (approved):
Looks great, thank you!

src/storage-types/src/parameters.proto
teskje: Good point, might as well clean up the ids here too.

teskje: I've done that and also reordered the proto fields to match the order in Rust.

test/testdrive/source-timestamps.td
```diff
@@ -59,9 +59,6 @@ ALTER SYSTEM SET kafka_default_metadata_fetch_interval = '1s'
 ALTER SYSTEM SET pg_offset_known_interval = '1s'
 ALTER SYSTEM SET mysql_offset_known_interval = '1s'
 
-> ALTER CLUSTER test SET (REPLICATION FACTOR 0)
```
aljoscha: mostly just curious: why are these removed?

aljoscha (approved):
looks good! :raised_hands:

test/testdrive/source-timestamps.td
teskje: They existed because applying the changed intervals needed a source restart before this PR, but it doesn't need one now!

aljoscha: :man_facepalming:

teskje:
TFTRs!

PR #31127, opened 2025-01-21
[persist] refactor Blob impl for Azure for higher performance

ParkMyCar:
This refactors the impl of `Blob` for Azure in a way that should be faster. The `BlobClient` we use from the `azure_storage_blob` crate returns a `Stream` that when `await`-ed sends a ranged GET request for a chunk of a blob. This PR refactors our impl so we await each ranged request in a `tokio::task` which increases the concurrency at which we fetch chunks of a `Part`.

It also refactors how we handle the case when the `content-length` header is missing, and adds metrics so we can track how often this occurs.

### Motivation

Maybe progress against https://github.com/MaterializeInc/database-issues/issues/8892

### Checklist

### DIFF
--- a/src/persist/src/azure.rs
+++ b/src/persist/src/azure.rs
@@ -20,6 +20,7 @@ use azure_storage::{prelude::*, CloudLocation, EMULATOR_ACCOUNT};
 use azure_storage_blobs::blob::operations::GetBlobResponse;
 use azure_storage_blobs::prelude::*;
 use bytes::Bytes;
+use futures_util::stream::FuturesOrdered;
 use futures_util::StreamExt;
 use tracing::{info, warn};
 use url::Url;
@@ -206,10 +207,7 @@ impl Blob for AzureBlob {
                         .new_region(usize::cast_from(content_length));
                     PreSizedBuffer::Sized(region)
                 }
-                0 => {
-                    metrics.get_invalid_resp.inc();
-                    PreSizedBuffer::Unknown(SegmentedBytes::new())
-                }
+                0 => PreSizedBuffer::Unknown(SegmentedBytes::new()),
             };
 
             let mut body = response.data;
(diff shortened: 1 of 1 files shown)


src/persist/src/azure.rs
```diff
@@ -216,35 +200,97 @@ impl Blob for AzureBlob {
...
+            Ok(MaybeLgBytes::LgBytes(lgbytes))
+        }
+
+        let mut handles = Vec::new();
+        // TODO: the default chunk size is 1MB. We have not tried tuning it,
+        // but making this configurable / running some benchmarks could be
+        // valuable.
+        let mut stream = blob.get().into_stream();
+
+        while let Some(value) = stream.next().await {
```
bkirwi, line 249: Could we `map`/`buffer_unordered` here instead of spinning up individual tasks? A bit closer to the S3 impl (which does not fork off individual tasks) and makes it easier to cap the concurrency per fetch...

src/persist/src/azure.rs
```diff
@@ -216,35 +200,97 @@ impl Blob for AzureBlob {
...
-                        .metrics
+                    let region = metrics
                         .lgbytes
                         .persist_azure
                         .new_region(usize::cast_from(content_length));
                     PreSizedBuffer::Sized(region)
                 }
-                0 => PreSizedBuffer::Unknown(Vec::new()),
+                0 => {
+                    metrics.get_invalid_resp.inc();
```
bkirwi, line 210: The S3 metrics say that a "content-length of 0 isn't necessarily invalid", which makes sense to me. Could we inc this only if the size turns out to not match the header?

(Generally I'm not convinced of the need to have this defensive coding here, so it'd be handy if this metric fired only in the cases that it was actually load-bearing!)

src/persist/src/azure.rs
ParkMyCar: good call! refactored to use `FuturesOrdered` like the S3 blob impl

src/persist/src/azure.rs
ParkMyCar: Ahhh yeah you're totally right, updated!

bkirwi (approved):
Changes look good!

ParkMyCar:
TFTR!

PR #31147, opened 2025-01-22
storage: use exchange-based sequencing to broadcast internal commands

teskje:
The Timely `Sequencer` doesn't behave well when different workers instantiate their parts at points in time that are sufficiently spread apart. It delays commands by that creation time difference and is prone to spinning the CPU at 100%.

To work around these issues, this PR replaces the `Sequencer` used in storage for broadcasting internal commands with a dataflow that sequences by sending all commands through a single worker.

First commit introduces a simple version of the sequencer dataflow that assumes deterministic ordering of data sent through Timely channels. The second commit removes that assumption by explicitly keeping track of the command order and waiting for outstanding commands.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/8893

### Checklist

### DIFF
--- a/src/storage/src/internal_control.rs
+++ b/src/storage/src/internal_control.rs
@@ -6,8 +6,10 @@
 //! Types for cluster-internal control messages that can be broadcast to all
 //! workers from individual operators/workers.
 
+use std::cell::RefCell;
 use std::collections::BTreeMap;
-use std::time::Instant;
+use std::rc::Rc;
+use std::sync::mpsc;
 
 use mz_repr::{GlobalId, Row};
 use mz_rocksdb::config::SharedWriteBufferManager;
@@ -18,8 +20,11 @@ use mz_storage_types::sinks::{MetadataFilled, StorageSinkDesc};
 use mz_storage_types::sources::IngestionDescription;
 use serde::{Deserialize, Serialize};
 use timely::communication::Allocate;
+use timely::dataflow::channels::pact::Pipeline;
+use timely::dataflow::operators::generic::{source, OutputHandle};
+use timely::dataflow::operators::{Broadcast, Exchange, Operator};
 use timely::progress::Antichain;
-use timely::synchronization::Sequencer;
+use timely::scheduling::{Activator, Scheduler};
 use timely::worker::Worker as TimelyWorker;
 
 use crate::statistics::{SinkStatisticsRecord, SourceStatisticsRecord};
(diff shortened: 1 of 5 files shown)


src/storage/src/internal_control.rs
```diff
@@ -120,31 +125,202 @@ pub enum InternalStorageCommand {
...
     timely_worker: &'w mut TimelyWorker<A>,
-) -> Sequencer<InternalStorageCommand> {
-    // TODO(aljoscha): Use something based on `mz_ore::NowFn`?
-    Sequencer::new(timely_worker, Instant::now())
+) -> (InternalCommandSender, InternalCommandReceiver) {
+    let (input_tx, input_rx) = mpsc::channel();
+    let (output_tx, output_rx) = mpsc::channel();
+    let activator = Rc::new(RefCell::new(None));
+
+    timely_worker.dataflow_named::<(), _, _>("command_sequencer", {
```
antiguru, line 174: I think this is fine, but it essentially uses the `index` as a capability. I think there's an alternative design that uses Timely's capabilities, but I'm leaving it to you which one you'd like to implement here. I think the order from the source to the sequencer doesn't matter, as long as it's not reordered -- you could also think of this as a batch of data and a capability `(worker, index)`. After the sequencer, it's simpler as there is only a single operator instance that retains a capability, so it could just use a counter as its capability. The data then would be a vectors of updates per time, i.e., `Stream<_, Vec<Command>>`.

antiguru (approved):
Looks fine, left a comment about an alternative way to implementa the command distribution.

src/storage/src/internal_control.rs
teskje: I tried using timestamps first, instead of putting the index into the data and it turned out more complicated. Maybe I missed something though.

* I want to use a Timestamp type of `(worker, index)`, so `(usize, u64)`. Tuples already implement `Timestamp`, but they don't implement `Refines<()>`, so I can't use them directly. Instead I have to implement a new timestamp type, which requires a bunch of boilerplate.
* If I do the above, I think I'd still have to do the put-commands-into-a-map-and-wait-for-the-next-index thing in both the sequencer and the sink because there is no guarantee that an operator sees updates in it inputs ordered by time (right?). So in the end the code would look pretty much the same, just with a bunch of more timestamp impl and capability management code.

> I think the order from the source to the sequencer doesn't matter, as long as it's not reordered

The "as long as it's not reordered" is important though! Correct me if I'm wrong, but I think this could happen:

* Source sends command `C1` at `(0, 0)`.
* Source sends command `C2` at `(0, 1)`.
* Sequencer receives command `C2`.
* Sequencer receives command `C1`.

Or is there any guarantee that updates with different times are received in time order?

frankmcsherry: > I want to use a Timestamp type of `(worker, index)`, ...

Ah, I'm not sure you do! At least, `(worker, index)` is a totally ordered type, where `worker` trumps `index` no matter what, and roughly everyone should wait for a worker to finish out all of its `index` before moving on to the next worker. I think you might want `Product<worker, index>`, which I think should implement `Refines<()>`. Does that sound right? (I'm trying to back out the intent from my partial understanding).

teskje: True, though the thing I was considering would not involve waiting for frontiers or even looking at frontiers at all, but just using the timestamp to transport the command index (instead of putting it into the data as this PR does). If you don't look at frontiers, the partial order of times doesn't matter. It's entirely possible that working with frontiers would lead to a simpler implementation, but I don't know what that could look like!

> I think you might want `Product<worker, index>`

I'm not so sure! The only thing we need is that the commands coming from one worker retain their order in the output, between workers there is no defined order. So we'd need a `(worker, index)` where instances are only comparable if `worker` is equal. But that becomes awkward because `Timestamp` assumes the existence of a minimum that you can downgrade to all other timestamps. So I guess an `(Option<worker>, index)` where `(None, 0)` is that minimum?

teskje:
I'm going ahead and merging this since it improves the status quo. Happy to iterate if we can come up with a better/simpler way!

PR #31149, opened 2025-01-22
[storage] Minor controller cleanups and refactoring

bkirwi:
Some minor drive-by as I've been reading through the storage controller.

### Motivation

The first commit is inspired by a recent incident: we saw a bunch of Persist memory on the stack of an environment, some of which was allocated by this warmup code. Far from a smoking gun... but in general it feels wise to not attempt unbounded concurrent work, especially for an optimization like this one.

The others are hopefully self-explanatory, but let me know if anything seems odd!

### Checklist

### DIFF
--- a/src/adapter/src/coord/ddl.rs
+++ b/src/adapter/src/coord/ddl.rs
@@ -1289,11 +1289,6 @@ impl Coordinator {
         // Validate `sink.from` is in fact a storage collection
         self.controller.storage.check_exists(sink.from)?;
 
-        let status_id = self
-            .catalog()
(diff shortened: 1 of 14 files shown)


ParkMyCar (approved):
All makes sense to me!

My only hesitation is the removal of `status_id` from the `SinkDesc` because I'm not sure if we intend to use it in the future. Worse case though it can always be re-added!

bkirwi:
> My only hesitation is the removal of status_id from the SinkDesc because I'm not sure if we intend to use it in the future.

Thankfully this one I'm pretty confident about, since it dates back to when I was following this stuff more closely. 😅 Sinks originally wrote their own status updates to the status collection, but those writes got moved to the controller... I believe this just never got cleaned up.

PR #31152, opened 2025-01-22
[storage] Add a progress shard for sinks and skip the snapshot when possible

bkirwi:
### Motivation

Design sketch is here: https://github.com/MaterializeInc/database-issues/issues/8603#issuecomment-2611104841

Fixes https://github.com/MaterializeInc/database-issues/issues/8603

### Tips for reviewer

Adding a collection for sinks makes them much more like other types of things that the storage controller manages, which allows (and forces) us to handle them as a type of collection instead of a separate thing. Reworking this is much of the size of the diff - though the net result is slightly less / more uniform controller code, which is nice.

I've left a number of `// TODO(sinks)` todos for possible followups that I didn't want to bite off here.

### Checklist

### DIFF
--- a/src/storage-controller/src/lib.rs
+++ b/src/storage-controller/src/lib.rs
@@ -1440,6 +1440,7 @@ where
             .expect("missing dependency")
             .into_element();
         let from_storage_metadata = self.storage_collections.collection_metadata(from_id)?;
+        let to_storage_metadata = self.storage_collections.collection_metadata(id)?;
 
         // Check whether the sink's write frontier is beyond the read hold we got
         let cur_export = self
@@ -1480,6 +1481,7 @@ where
                 partition_strategy: new_description.sink.partition_strategy,
                 from_storage_metadata,
                 with_snapshot: new_description.sink.with_snapshot,
+                to_storage_metadata,
             },
         };
 
@@ -1529,6 +1531,7 @@ where
             let from_storage_metadata = self
                 .storage_collections
                 .collection_metadata(new_export_description.sink.from)?;
+            let to_storage_metadata = self.storage_collections.collection_metadata(id)?;
 
             let cmd = RunSinkCommand {
                 id,
@@ -1552,6 +1555,7 @@ where
                     // TODO(petrosagg): change the controller to explicitly track dataflow executions
                     as_of: as_of.to_owned(),
                     from_storage_metadata,
+                    to_storage_metadata,
                 },
             };
 
@@ -3155,6 +3159,7 @@ where
         let from_storage_metadata = self
             .storage_collections
             .collection_metadata(description.sink.from)?;
+        let to_storage_metadata = self.storage_collections.collection_metadata(id)?;
 
         let cmd = RunSinkCommand {
             id,
@@ -3168,6 +3173,7 @@ where
                 partition_strategy: description.sink.partition_strategy.clone(),
                 from_storage_metadata,
                 with_snapshot: description.sink.with_snapshot,
+                to_storage_metadata,
             },
         };
 
--- a/src/storage-controller/src/lib.rs
+++ b/src/storage-controller/src/lib.rs
@@ -1905,10 +1905,7 @@ where
             Some(StorageResponse::DroppedId(id)) => {
                 tracing::debug!("DroppedId for collection {id}");
 
-                if let Some(_collection) = self.collections.remove(&id) {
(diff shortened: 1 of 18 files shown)


bkirwi:
Still doing some testing, but I think this feels more or less right. @aljoscha or @petrosagg, able to take a look?

test/cluster/mzcompose.py
```diff
@@ -5140,12 +5140,12 @@ def workflow_test_constant_sink(c: Composition) -> None:
                 > SELECT write_frontier
                   FROM mz_internal.mz_frontiers
                   JOIN mz_sinks ON id = object_id
-                  WHERE name = 'snk'
+                  WHERE name = 'snk' AND write_frontier IS NOT NULL
 
                 > SELECT status
                   FROM mz_internal.mz_sink_statuses
                   WHERE name = 'snk'
-                dropped
```
bkirwi, line 5148: Previously, the read frontier was immediately set to {} for sinks, so having the write frontier progress to {} would mark the sink dropped. After this PR, the read frontier can evolve independently, so the sink is not immediately removed from introspection.

Looking at https://github.com/MaterializeInc/database-issues/issues/8842 - this seems like an improvement, though there may be more to do. (In the future, not on this PR.)

src/storage-controller/src/lib.rs
```diff
@@ -1768,27 +1746,42 @@ where
...
+        storage_metadata: &StorageMetadata,
+        mut identifiers: Vec<GlobalId>,
+    ) {
+        // Ignore exports that have already been removed.
+        identifiers.retain(|id| self.export(*id).is_ok());
 
-            // We don't explicitly remove read capabilities! Downgrading the
-            // frontier of the sink to `[]` (the empty Antichain), will
-            // propagate to the storage dependencies.
+        // TODO: ideally we'd advance the write frontier ourselves here, but this function's
```
aljoscha, line 1765: By this you mean the write frontier of the (new) "progress shard"?

src/storage-controller/src/lib.rs
```diff
@@ -1904,17 +1897,6 @@ where
                 if let Some(_collection) = self.collections.remove(&id) {
                     // Nothing to do, we already dropped read holds in
                     // `drop_sources_unvalidated`.
-                } else if let Some(export) = self.exports.get_mut(&id) {
-                    // TODO: Current main never drops export state, so we
```
aljoscha, line 1908: 🙌

src/storage-controller/src/lib.rs
```diff
@@ -3608,3 +3532,17 @@ fn replica_status_history_desc(params: &StorageParameters) -> StatusHistoryDesc<
         extract_time: Box::new(move |datums| datums[time_idx].unwrap_timestamptz()),
     }
 }
+
+/// Replace one antichain with another, tracking the overall changes in the returned `ChangeBatch`.
+fn swap_updates<T: Timestamp>(
```
aljoscha, line 3537: 🙌

aljoscha (approved):
Oh wow, it's interesting how small the actual change in the export logic is, compared to the refactorings!

Overall, this looks good. I have to say. though, that I probably trust @def-'s tests (which we already have) for sinks, with restarts and everything, more than my capability to suss out subtle concurrency issues from reading the code.

def- (approved):
The warnings in nightly are fixed if you rebase on main: https://buildkite.com/materialize/nightly/builds/11099 (which I'd suggest anyway)

def-:
new nightly just in case: https://buildkite.com/materialize/nightly/builds/11103

src/storage-controller/src/lib.rs
bkirwi: Yeah, that's right! It's safe to do - the sink code is robust to it - and it would simplify the finalization logic if we could assume that a dropped sink had an empty write / implied frontier.

bkirwi:
The Kafka migration tests need a fix - we're running the changed tests against an old version, so I'll need to tweak them to be more robust I think. Will get that up in a sec.

The copy to S3 test failures look unrelated?

def-:
Yes, copy to s3 is unrelated.

bkirwi:
Thanks all!

> Oh wow, it's interesting how small the actual change in the export logic is, compared to the refactorings!

There are a couple followups that I'd like to do that will probably expand this logic (moving the frontier management out of the hot path, writing actual data to the shard) but I'll take that in a followup... merging this in now.

PR #31153, opened 2025-01-22
docs: self-managed access control

kay-kim:
https://preview.materialize.com/materialize/31153/self-managed/v25.1/manage/access-control/

Converted to draft...will sort out the cross ref errors.


JLDLaughlin (approved):
👩🏻‍🍳

PR #31155, opened 2025-01-23
Consolidate produces batch of data

antiguru:
Instead of flattening the outputs of consolidate, provide the output as a
whole batch, encoded as an `Stream<_, Vec<C>>`. This allows downstream operators
to act on a whole batch at once, without relying on undocumented properties
of how Timely channels behave. For operators that need to flatten the
output need slightly different logic.

### Checklist

### DIFF
--- a/src/storage-operators/src/s3_oneshot_sink.rs
+++ b/src/storage-operators/src/s3_oneshot_sink.rs
@@ -16,7 +16,7 @@ use std::rc::Rc;
 
 use anyhow::anyhow;
 use aws_types::sdk_config::SdkConfig;
-use differential_dataflow::{Collection, Hashable};
+use differential_dataflow::Hashable;
 use futures::StreamExt;
 use mz_ore::cast::CastFrom;
 use mz_ore::error::ErrorExt;
@@ -31,6 +31,7 @@ use mz_storage_types::sinks::{S3SinkFormat, S3UploadInfo};
 use mz_timely_util::builder_async::{
     Event as AsyncEvent, OperatorBuilder as AsyncOperatorBuilder, PressOnDropButton,
 };
+use timely::container::columnation::TimelyStack;
 use timely::dataflow::channels::pact::{Exchange, Pipeline};
 use timely::dataflow::operators::Broadcast;
 use timely::dataflow::{Scope, Stream};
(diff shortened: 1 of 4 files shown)


src/timely-util/src/operator.rs
```diff
@@ -777,7 +794,7 @@ pub fn consolidate_pact<Ba, P, G>(
     stream: &StreamCore<G, Ba::Input>,
     pact: P,
     name: &str,
-) -> StreamCore<G, Ba::Output>
+) -> StreamCore<G, Rc<Vec<Ba::Output>>>
```
teskje, line 797: The purpose of the `Rc` isn't clear to me. Is it a safe-guard to avoid accidental expensive cloning of `Vec`s when the output stream is later forked?

antiguru: It is so that a chain of updates is inseparable, no matter what channel it traverses.

teskje: The concern being that Timely might otherwise try to pull the `Vec` apart while passing it through channels? Would be great to have a comment explaining this!

src/storage-operators/src/s3_oneshot_sink.rs
```diff
@@ -281,27 +288,30 @@ where
 /// Returns a `completion_stream` which contains 1 event per worker of
 /// the result of the upload operation, either an error or the number of rows
 /// uploaded by the worker.
+///
+/// The `input_collection` is a stream of chains, partitioned by the row's hash modulo the number
```
frankmcsherry, line 292: Same as above: "is a" vs "must be a".

src/timely-util/src/operator.rs
```diff
@@ -708,7 +707,18 @@ where
                 h.finish()
             });
             consolidate_pact::<Ba, _, _>(&self.map(|k| (k, ())).inner, exchange, name)
-                .map(|((k, ()), time, diff)| (k.clone(), time.clone(), diff.clone()))
+                .unary(Pipeline, "unpack consolidated", |_, _| {
```
frankmcsherry, line 710: Nit, but if this is coming in as a `Rc<Vec<...>>` we could check for unique ownership and avoid cloning everything, I think?

antiguru: It's columnar data, so we can only get references out of it!

src/timely-util/src/operator.rs
antiguru: I changed the PR such that `consolidate_pact` produces a `Stream<G, Vec<Ba::Output>>`, which means that the _elements_ we're transferring are `Vec<_>`. This gives us the certainty that Timely will not look into the chains.

src/storage-operators/src/s3_oneshot_sink.rs
```diff
@@ -54,9 +55,12 @@ mod pgcopy;
 ///   - completion: removes the sentinel file and calls the `worker_callback`
 ///
 /// Returns a token that should be held to keep the sink alive.
+///
+/// The `input_collection` is a stream of chains, partitioned by the row's hash modulo the number
+/// of batches.
```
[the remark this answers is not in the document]
antiguru: Fixed! Thanks

src/timely-util/src/operator.rs
frankmcsherry: I see! Owned containers, but the containers don't own keys, times, etc.

ParkMyCar (approved):
I don't totally understand the changes to `consolidate_pact`, but hopefully those are covered by Jan or Frank! But sending around batches of Rows instead of relying on the ordering of timely channels looks good to me!

PR #31166, opened 2025-01-23
adapter: Fix startup parameter default values

jkosh44:
There are some server configuration parameters that we need during
startup, before the catalog has been fully opened. To get the current
value we manually look at the current value in LaunchDarkly (which is
confusingly returned as `None` if it matches the default value in the
code), then we manually look in the catalog, then we manually look in
the server parameter default override CLI arg map.

Previously, for some of the parameters, if we weren't able to find the
value in any of those places, then we hard-code in the value false.
This is not correct, and instead we should fall back to their default
value, which is sometimes true. This commit fixes the issue by falling
back to their default values.

### Motivation
This PR fixes a previously unreported bug.

### Checklist

### DIFF
--- a/src/environmentd/src/lib.rs
+++ b/src/environmentd/src/lib.rs
@@ -401,7 +401,10 @@ impl Listeners {
                 strconv::parse_bool(x).map_err(|x| x.to_string())
             })?;
             let catalog = openable_adapter_storage.get_enable_0dt_deployment().await?;
-            let computed = ld.or(catalog).or(default).unwrap_or(false);
+            let computed = ld
+                .or(catalog)
+                .or(default)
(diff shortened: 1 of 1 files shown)


benesch (approved):
Oooooof.

jkosh44:
This is potentially now a confusing log message, but it does seem to work
```
determined value for enable_0dt_deployment system parameter","computed":"true","ld":"None","catalog":"None","default":"None"
```

benesch:
> This is potentially now a confusing log message, but it does seem to work
> 
> ```
> determined value for enable_0dt_deployment system parameter","computed":"true","ld":"None","catalog":"None","default":"None"
> ```

Arguably we should split the log line to have "CLI default" and "code default" but don't go back through CI just for that.

PR #31144, opened 2025-01-22
[copy_from] AWS source

ParkMyCar:
_Stacked on top of_: https://github.com/MaterializeInc/materialize/pull/30956

This PR implements a new AwsS3 `OneshotSource` that allows copying in files from S3, e.g.
```
COPY INTO my_table FROM 's3://my-test-bucket' (FORMAT CSV, FILES = ['important.csv']);
```

Along with `FILES = [<files>]` we also support a `PATTERN = <glob>` option which allows copying multiple files all at once.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/8860
Fixes https://github.com/MaterializeInc/database-issues/issues/8855

### Tips for reviewer

Review only the final commit, the one titled "start, implementation of an S3 oneshot source"

### Checklist

- [x] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [x] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [x] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [x] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [x] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #31174, opened 2025-01-24
[copy_from] Refactor ranged requests, fix HTTP source

ParkMyCar:
_Stacked on top of_ https://github.com/MaterializeInc/materialize/pull/31144

This PR refactors ranged get requests so the logic for forming the header value is shared for both the HTTP and S3 source. It also fixes a bug where sometimes `HEAD` requests (which we use to get metadata for a file) are not supported, so we fallback to a `GET` request but quickly drop the body.

### Motivation

Fix a bug in ranged HTTP requests I found in the demo today

### Tips for reviewer

review only the final commit, the one titled "start, fix ranged requests"

### Checklist

### DIFF
--- a/src/storage-operators/src/oneshot_source.rs
+++ b/src/storage-operators/src/oneshot_source.rs
@@ -105,6 +105,8 @@ pub mod csv;
 pub mod aws_source;
 pub mod http_source;
 
+mod util;
+
 /// Render a dataflow to do a "oneshot" ingestion.
 ///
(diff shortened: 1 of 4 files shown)


jkosh44 (approved):
LGTM, but we should probably add tests with the Range header if we don't already have some.

ParkMyCar:
> LGTM, but we should probably add tests with the Range header if we don't already have some.

Will do!

PR #31186, opened 2025-01-24
Switch dataflow fragments to columnar

antiguru:
Switch the specialized-arrange formation and the linear join preparation phase to use columnar data on dataflow edges. Requires unreleased changes from timely and differential.

This includes changes from #31158

### Checklist


src/compute/src/render/context.rs
```diff
@@ -1064,44 +1069,48 @@ where
...
-            .map_fallible::<CapacityContainerBuilder<_>, CapacityContainerBuilder<_>, _, _, _>(
-                "FormArrangementKey",
-                specialized_arrangement_key(key.clone(), thinning.clone()),
+        let (oks, errs) =
+            oks.inner
+                .unary_fallible::<ColumnBuilder<((Row, Row), S::Timestamp, i64)>, _, _, _>(
+                    Pipeline,
+                    "FormArrangementKey",
+                    move |_, _| {
+                        Box::new(move |input, ok, err| {
```
frankmcsherry, line 1081: This went from being a `map_fallible` to something more complicated. I think it would benefit from a comment explaining that (if correct) this is implementing a `map_fallible`, but it appears complicated in order to accommodate containers of rows.

frankmcsherry: The comment at the top of the method says weird stuff about specialization and bit lengths. This no longer seems to be accurate, and perhaps that is a fine place to leave the explanation about what the method does. Ideally it would describe the arguments `key` and `thinning` specifically, as I believe these are generally complicated (thinning, for sure). Not necessarily how they work, but if it is exactly the thing that has a clear meaning where created (the thinning and permutation stuff) perhaps a reference to it.

src/compute/src/render/context.rs
```diff
@@ -1064,44 +1069,48 @@ where
...
+                            let mut val_buf = Row::default();
+                            let mut datums = DatumVec::new();
+                            let temp_storage = RowArena::new();
+                            while let Some((time, data)) = input.next() {
+                                let mut ok_session = ok.session_with_builder(&time);
+                                let mut err_session = err.session(&time);
+                                for (row, time, diff) in data.iter() {
+                                    temp_storage.clear();
+                                    let datums = datums.borrow_with(row);
+                                    let val_datum_iter = thinning.iter().map(|c| datums[*c]);
```
frankmcsherry, line 1092: Bit of a nit, but can `val_datum_iter` be moved into the `Ok(())` branch? Mostly thinking about code clarity, and that one doesn't need to understand this term heading in to the `match` statement.

frankmcsherry (approved):
These all make sense. I have to puzzle out which things are columnar and which are not (these do some conversion, but I'm missing the larger picture so far, but trust that it will come together).

src/compute/src/render/context.rs
antiguru: Fixed.

src/compute/src/render/context.rs
antiguru: Updated the doc comment and added an inline comment.

PR #31202, opened 2025-01-27
docs: kapa.ai exclude self-managed sources in cloud docs

kay-kim:
https://preview.materialize.com/materialize/31202/
(the Documentation Cloud is set up on Kapa side as well so that it excludes docs/self-managed/)

### DIFF
--- a/doc/user/layouts/partials/head.html
+++ b/doc/user/layouts/partials/head.html
@@ -94,7 +94,7 @@ toCSS | fingerprint }}
   data-modal-open-on-command-k="true"
   data-modal-command-k-search-mode-default="true"
   data-search-result-link-target="_self"
-  data-search-include-source-names='["Documentation"]'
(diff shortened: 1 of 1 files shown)


SangJunBak:
At least from the preview link, it seems like searching in self managed still targets cloud:

Screenshot 2025-01-27 at 11 58 18 AM

kay-kim:
Oh .. the self-managed in this preview (of `main` branch) isn't the most recent self-managed (those are in the `self-managed-docs` branch).  Those files now are bit of a bridge while we added redirects from when we published the self-managed docs from main.  There will be some cleanup in the future, but left them there for now in case we need to make further changes.

kay-kim:
Although, now that I look at the config parameter, I think this might only restrict the Search and not the AI part 😞😞😞 Aw, man... drat.

JLDLaughlin (approved):

PR #31206, opened 2025-01-27
catalog: Key expression cache by binary version

jkosh44:
The expression cache caches optimized expressions in a persist shard. It uses the environment's deploy generation as part of the key for the cache. The assumption was that two environments with the same deploy generation would always have the same binary version. This assumption would allow two environments with the same deploy generation to safely deserialize each other's expressions, without worrying about backwards and forwards compatibility.

This assumption was not correct. Two environments with different binary version can use the same deploy generation as long as one environment never fully completed a deployment. This is especially bad because the expression cache is written to and read from in read-only mode, before a deployment is complete.

This commit updates the key of the expression cache to explicitly use the binary version of environmentd so that two environments with the same version can safely deserialize each other's expressions.

Fixes #MaterializeInc/database-issues/issues/8917

### Motivation
This PR fixes a recognized bug.

### Checklist

### DIFF
--- a/src/adapter/src/catalog/migrate.rs
+++ b/src/adapter/src/catalog/migrate.rs
@@ -9,17 +9,21 @@
 
 use std::collections::BTreeMap;
 
+use maplit::btreeset;
 use mz_catalog::builtin::BuiltinTable;
 use mz_catalog::durable::Transaction;
 use mz_catalog::memory::objects::StateUpdate;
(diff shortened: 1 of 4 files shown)


jkosh44:
WARNING: This commit is going to cause 15-20 seconds of startup time regression in the next release because it will wipe out the expression cache.

ParkMyCar (approved):
LGTM

PR #31207, opened 2025-01-27
materialized: strip debuginfo

benesch:
The materialized binary generates massive debuginfo: 8GB with full debuginfo and 3.5GB with limited debuginfo. (This is not the compressed size of the debuginfo on disk, but the amount of memory required to load the debuginfo into memory so that backtraces can be symbolized.)

We've historically shipped the materialized binary with full debuginfo, so that we get rich backtraces for any crashes users run into when running the emulator. Unfortunately, the full debuginfo is so large that it is itself the *cause* of OOMs [0], which is unacceptable. (When a console query encounters certain routine errors, like a connection failing to validate, the adapter attempts to log a backtrace, which requires loading the debuginfo.) Even the limited debuginfo size (3.5GB) is unacceptable for a Docker image that's meant to be run on developer laptops.

So, this commit adjusts the materialized image to strip all debuginfo from the binary.

If a user reports a crash with an unsymbolized backtrace, it's still *possible* (just painful) to manually symbolize the backtrace as long as they give us the exact version of Materialize they were running. We'll just need to manually run `addr2line` on each address reported in the backtrace. (We do irrevocably lose access to frames for inlined functions, but that seems tolerable.)

One silver lining here is that the Docker image will get much smaller. It's currently about 1GB. I expect this change to shave off a huge chunk of that.

[0]: https://materializeinc.slack.com/archives/C07FX1W1Y03/p1737414021061139

### Checklist

### DIFF
--- a/src/materialized/ci/mzbuild.yml
+++ b/src/materialized/ci/mzbuild.yml
@@ -14,4 +14,3 @@ pre-image:
     bin: [ materialized ]
     bazel-bin:
       materialized: "@//src/materialized:materialized"
-    strip: false


ParkMyCar (approved):
Wild how large the debug info is, but this LGTM!

def- (approved):
Agreed

PR #31220, opened 2025-01-28
[build] Get the `npm` crate working with Bazel

ParkMyCar:
This PR fixes our flamegraph rendering with our builtin memory and CPU profiling. Previously the vendored JS and CSS artifacts were not getting properly included into our build because the Cargo Build script that is responsible for downloading them gets run in a different sandbox than the action that builds the library.

Now the `npm::ensure(...)` call optionally takes an `OUT_DIR` param that it will copy the artifacts into and thus they will get included as part of the Bazel build. I manually built the `materialized` Docker image with Bazel and confirmed the flamegraphs worked!

### Motivation

Fix our builtin memory and CPU profiling

### Checklist

### DIFF
--- a/src/http-util/src/lib.rs
+++ b/src/http-util/src/lib.rs
@@ -50,8 +50,10 @@ macro_rules! make_handle_static {
         ) -> impl ::axum::response::IntoResponse {
             #[cfg(not(feature = "dev-web"))]
             const DIR_1: ::include_dir::Dir = $dir_1;
-            #[cfg(not(feature = "dev-web"))]
-            $(const DIR_2: ::include_dir::Dir = $dir_2;)?
+            $(
+                #[cfg(not(feature = "dev-web"))]
+                const DIR_2: ::include_dir::Dir = $dir_2;
+            )?
 
 
             #[cfg(not(feature = "dev-web"))]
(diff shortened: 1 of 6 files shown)


src/http-util/src/lib.rs
```diff
@@ -38,17 +38,25 @@ where
...
+    ) => {
         #[allow(clippy::unused_async)]
         pub async fn handle_static(
             path: ::axum::extract::Path<String>,
         ) -> impl ::axum::response::IntoResponse {
             #[cfg(not(feature = "dev-web"))]
-            const STATIC_DIR: ::include_dir::Dir = $static_dir;
+            const DIR_1: ::include_dir::Dir = $dir_1;
+            #[cfg(not(feature = "dev-web"))]
+            $(const DIR_2: ::include_dir::Dir = $dir_2;)?
```
antiguru, line 54: Should the cfg attribute go into the $(..) expansion?

antiguru (approved):
Thank you for figuring this out! Looks great, left an inline comment.

benesch:
Thank you, Parker!

src/http-util/src/lib.rs
ParkMyCar: Good call, updated!

PR #31222, opened 2025-01-29
doc: add ports aliugment note in Kafka privatelink guide

bobbyiliev:
### Motivation

Adding a small note to point out that misalignment between ports and broker addresses will result in connectivity issues.

### DIFF
--- a/doc/user/layouts/shortcodes/network-security/privatelink-kafka.md
+++ b/doc/user/layouts/shortcodes/network-security/privatelink-kafka.md
@@ -151,4 +151,7 @@ CREATE CONNECTION kafka_connection TO KAFKA (
 );
 ```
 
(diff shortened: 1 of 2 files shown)


doc/user/layouts/shortcodes/network-security/privatelink-kafka.md
```diff
@@ -151,4 +151,7 @@ CREATE CONNECTION kafka_connection TO KAFKA (
 );
 ```
 
-The `(PORT <port_number>)` value must match the port that you used when creating the **TCP listener** in the Network Load Balancer. Be sure to specify the correct availability zone for each broker.
+There are a few important things to note:
```
morsapaes, line 154: There's a few more places where this needs to be adapted. I can push the changes in tomorrow, since those are still fresh in my head from the cross-region docs PR!

jubrad (approved):
LGTM, but @morsapaes suggests this should be updated in a few more places?

morsapaes (approved):

PR #31226, opened 2025-01-29
storage: in remap, when since is empty, suspend instead of panic'ing

aljoscha:
As the comment describes, this is a race condition that is expected to happen and it's better to suspend rather than bring down the whole cluster, which causes pain for customers/the oncall.

@def- This fixes the panic of incident-360
Fixes https://github.com/MaterializeInc/incidents-and-escalations/issues/173
Fixes https://github.com/MaterializeInc/database-issues/issues/8936

### Checklist

### DIFF
--- a/src/storage/src/source/reclock/compat.rs
+++ b/src/storage/src/source/reclock/compat.rs
@@ -20,6 +20,7 @@ use differential_dataflow::lattice::Lattice;
 use fail::fail_point;
 use futures::stream::LocalBoxStream;
 use futures::StreamExt;
+use mz_ore::soft_panic_or_log;
 use mz_ore::vec::VecExt;
 use mz_persist_client::cache::PersistClientCache;
 use mz_persist_client::error::UpperMismatch;
@@ -119,6 +120,31 @@ where
         // Allow manually simulating the scenario where the since of the remap
         // shard has advanced too far.
         fail_point!("invalid_remap_as_of");
+
+        if since.is_empty() {
+            // This can happen when, say, a source is being dropped but we on
+            // the cluster are busy and notice that only later. In those cases
+            // it can happen that we still try to render an ingestion that is
+            // not valid anymore and where the shards it uses are not valid to
+            // use anymore.
+            //
+            // This is a rare race condition and something that is expected to
+            // happen every now and then. It's not a bug in the current way of
+            // how things work.
+            tracing::info!(
+                source_id = %id,
+                %worker_id,
+                "since of remap shard is the empty antichain, suspending...");
+
+            // We wait 5 hours to give the commands a chance to arrive at this
+            // replica and for it to drop our dataflow.
+            tokio::time::sleep(Duration::from_secs(5 * 60 * 60)).await;
+
+            // If we're still here after 5 hours, something has gone wrong and
+            // we complain.
+            soft_panic_or_log!("since of remap shard is the empty antichain, source_id = {id}, worker_id = {worker_id}");
+        }
+
         if !PartialOrder::less_equal(since, &as_of) {
             anyhow::bail!(
                 "invalid as_of: as_of({as_of:?}) < since({since:?}), \
--- a/src/storage/src/storage_state.rs
+++ b/src/storage/src/storage_state.rs
@@ -629,10 +629,10 @@ impl<'w, A: Allocate> Worker<'w, A> {
             }
             InternalStorageCommand::CreateIngestionDataflow {
                 id: ingestion_id,
-                ingestion_description,
+                mut ingestion_description,
                 as_of,
-                resume_uppers,
-                source_resume_uppers,
+                mut resume_uppers,
+                mut source_resume_uppers,
             } => {
                 info!(
                     ?as_of,
@@ -642,6 +642,10 @@ impl<'w, A: Allocate> Worker<'w, A> {
                     self.timely_worker.peers(),
                 );
 
+                // We initialize statistics before we prune finished exports. We
+                // still want to export statistics for these, plus the rendering
+                // machinery will get confused if there are not at least
+                // statistics for the "main" source.
                 for (export_id, export) in ingestion_description.source_exports.iter() {
                     let resume_upper = resume_uppers[export_id].clone();
                     self.storage_state.aggregated_statistics.initialize_source(
@@ -660,6 +664,18 @@ impl<'w, A: Allocate> Worker<'w, A> {
                         },
                     );
                 }
+
+                let finished_exports: BTreeSet<GlobalId> = resume_uppers
+                    .iter()
+                    .filter(|(_, frontier)| frontier.is_empty())
+                    .map(|(id, _)| *id)
+                    .collect();
+
+                resume_uppers.retain(|id, _| !finished_exports.contains(id));
+                source_resume_uppers.retain(|id, _| !finished_exports.contains(id));
+                ingestion_description
+                    .source_exports
+                    .retain(|id, _| !finished_exports.contains(id));
 
                 for id in ingestion_description.collection_ids() {
                     // If there is already a shared upper, we re-use it, to make


def- (commented):
What would be a good way to test this? Should I try just creating sources, using them, and then deleting them?

aljoscha:
The case we observed in the incident is the source being dropped right after creating it, I think that's the most promising path. It also helps when the cluster is _really_ busy, so it doesn't try and render the source in time but is delayed to after the source is dropped already

def-:
I have a simple testdrive test with which I'm trying to reproduce this, but I ran into another panic instead:
```
thread 'timely:work-0' panicked at src/storage/src/render/sources.rs:223:18:
resuming an already finished ingestion
   5: core::panicking::panic_fmt
   6: core::option::expect_failed
   7: mz_storage::render::sources::render_source::<timely::dataflow::scopes::child::Child<timely::worker::Worker<timely_communication::allocator::generic::Generic>, ()>, mz_storage_types::sources::kafka::KafkaSourceConnection>
   8: mz_storage::render::build_ingestion_dataflow::<timely_communication::allocator::generic::Generic>::{closure#0}::{closure#0}
   9: <timely::worker::Worker<timely_communication::allocator::generic::Generic>>::dataflow_core::<(), (), mz_storage::render::build_ingestion_dataflow<timely_communication::allocator::generic::Generic>::{closure#0}, alloc::boxed::Box<()>>
  10: mz_storage::render::build_ingestion_dataflow::<timely_communication::allocator::generic::Generic>
  11: <mz_storage::storage_state::Worker<timely_communication::allocator::generic::Generic>>::run_client
  12: <mz_storage::server::Config as mz_cluster::types::AsRunnableWorker<mz_storage_client::client::StorageCommand, mz_storage_client::client::StorageResponse>>::build_and_run::<timely_communication::allocator::generic::Generic>
```
I'll experiment around a bit with making it reproduce more reliably and open a separate PR.

def-:
Good news, my test in https://github.com/MaterializeInc/materialize/pull/31243 can also reproduces this panic, seen in this nightly run: https://buildkite.com/materialize/nightly/builds/11014#_
```
testdrive-materialized-1          | thread 'timely:work-0' panicked at src/storage/src/render/sources.rs:305:37: upsert_rehydration: u2689: sc109340c-2914-49d2-b5b7-3d0531da7bf8 cannot serve requested as_of Antichain { elements: [0] }: Since(Antichain { elements: [] })
```

src/storage/src/source/reclock/compat.rs
```diff
@@ -117,6 +117,30 @@ where
...
+        if since.is_empty() {
+            // This can happen when, say, a source is being dropped but we on
+            // the cluster are busy and notice that only later. In those cases
+            // it can happen that we still try to render an ingestion that is
+            // not valid anymore and where the shards it uses are not valid to
+            // use anymore.
+            //
+            // This is a rare race condition and something that is expected to
+            // happen every now and then. It's not a bug in the current way of
+            // how things work.
```
teskje, line 130: Storage has the `DroppedIds` protocol response that I assumed to exist to let the storage controller wait until all replicas have acknowledged the dropping of an object. Is that not true? Or does this race only come up in the context of 0dt upgrades?

aljoscha: DroppedIds only relate to state of the storage controller, read holds are dropped right away when the collection is dropped without waiting. Right now we drop things eagerly so that a struggling, slow, broken replica does not hold up cleaning up those things, finalizing shards ,etc.

src/storage/src/source/reclock/compat.rs
```diff
@@ -117,6 +117,30 @@ where
...
+                source_id = %id,
+                %worker_id,
+                "since of remap shard is the empty antichain, suspending...");
+
+            // Suspending will make it so we don't make progress and downstream
+            // operators don't think we're making progress. And this dataflow
+            // will eventually be removed once the news about removal make it
+            // from the controller to the cluster.
+            std::future::pending::<()>().await;
+            unreachable!("pending future never returns");
```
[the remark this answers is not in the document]
aljoscha: I think that's a good compromise before we do the more elaborate work of setting up a "channel" that allows reporting an Err back to the controller. I'll type that up!

src/storage/src/source/reclock/compat.rs
```diff
@@ -117,6 +118,31 @@ where
...
+                source_id = %id,
+                %worker_id,
+                "since of remap shard is the empty antichain, suspending...");
+
+            // We wait 5 hours to give the commands a chance to arrive at this
+            // replica and for it to drop our dataflow.
+            tokio::time::sleep(Duration::from_secs(5 * 60 * 60)).await;
+
+            // If we're still here after 5 hours, something has gone wrong and
+            // we complain.
```
petrosagg, line 142: Sounds like a good compromise!

petrosagg (approved):

aljoscha:
@def- I pushed another change that fixes the issue you reported above. And then we probably want to get your tests in in a speedy manner?

@petrosagg could you please have another look?

src/storage/src/storage_state.rs
```diff
@@ -523,15 +523,25 @@ impl<'w, A: Allocate> Worker<'w, A> {
...
-                    },
-                );
+                if resume_uppers.iter().any(|(_, f)| f.is_empty()) {
+                    tracing::info!(
+                        %id,
+                        ?as_of,
+                        ?resume_uppers,
+                        ?source_resume_uppers,
+                        "not resuming ingestion because at least one resume upper is empty"
+                    );
```
petrosagg, line 533: We should resume if any resume upper is non-empty because that means there is a subsource that has progress to be made. I think you could skip sending the command if all resume uppers are empty but I worry that skipping sending the command might also skip going through the standard state transitions on all the workers (which will send DroppedId responses to the controller)

def-:
My test in https://github.com/MaterializeInc/materialize/pull/31243 found another panic: https://github.com/MaterializeInc/database-issues/issues/8936 cc @petrosagg . So it's currently blocked from merging until that one is fixed too.

aljoscha:
> My test in #31243 found another panic: [MaterializeInc/database-issues#8936](https://github.com/MaterializeInc/database-issues/issues/8936) cc @petrosagg . So it's currently blocked from merging until that one is fixed too.

@def- that's the one I mentioned above, and should be fixed by my follow-up commits

def-:
Perfect, I have test runs with my new test running: https://github.com/MaterializeInc/materialize/pull/31243 & https://buildkite.com/materialize/nightly/builds/11262

def-:
I got a bunch of timeouts on my test run (https://buildkite.com/materialize/test/builds/99510#01953cb2-86b9-44b1-a5a1-cb2d81a2042e), seems to just hang in some `SELECT` queries, but maybe that's something you already fixed? I saw you just pushed another commit here.

aljoscha:
> I got a bunch of timeouts on my test run (https://buildkite.com/materialize/test/builds/99510#01953cb2-86b9-44b1-a5a1-cb2d81a2042e), seems to just hang in some `SELECT` queries, but maybe that's something you already fixed? I saw you just pushed another commit here.

They might be, the latest change is slightly different from what I pushed before, and it _could_ explain the hangs. Would you mind rebasing and testing again? Sorry about that!

def-:
Sure, running: https://buildkite.com/materialize/test/builds/99519 & https://buildkite.com/materialize/nightly/builds/11266

aljoscha:
@def- I don't know why that thing happened, but nightly now seems to "pass", if you want to rebase and try again with your changes

def-:
Done: https://buildkite.com/materialize/test/builds/99595 & https://buildkite.com/materialize/nightly/builds/11312

aljoscha:
@def- your nightly looks very unhappy. One of the failures is the old-upsert one that I also have (but you fixed that already). For the other things, I don't know ... 🙈

def-:
Those are definitely interesting failures, but maybe not caused by your PR. I guess merging my race test is not a good idea yet. Apparently when clusterd goes OoM at the right time we later panic with `thread 'timely:work-0' panicked at src/storage/src/storage_state.rs:1389:29: AllowCompaction command for non-existent u3130`
I'm also concerned by the `testdrive-materialized-1     | 2025-02-26T17:28:59.012803Z  thread 'timely:work-3' panicked at external/crates_io__columnar-0.3.0/src/lib.rs:207:75: index out of bounds: the len is 1 but the index is 1` without an OoM before.
The Azure blob store failure can be ignored.

Since your nightly is showing none of that I don't have enough data to block merging this PR, and it definitely fixes some races we had before. I'm not sure if you want to look into the two new ones or I should open separate issues.

def- (approved):

aljoscha:
tyty! I've run multiple nightlies now with my changes and they all came back good, so merging. Could you open issues for these other two, please. The first one, I'd look at, but the second one looks more like compute/Moritz territory.

def-:
Done: https://github.com/MaterializeInc/database-issues/issues/8995 & https://github.com/MaterializeInc/database-issues/issues/8996

aljoscha:
I diagnosed at least two problems that are making https://github.com/MaterializeInc/materialize/pull/31243 fail right now. I'll cut separate bug issues and link to your issue.

PR #30003, opened 2024-10-15
enable source replication for kafka sources

petrosagg:
### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #31227, opened 2025-01-29
storage: allow kafka/loadgen sources on multi-replica clusters, gated by dyncfg

aljoscha:
Version of https://github.com/MaterializeInc/materialize/pull/30003 with a dyncfg for enabling/disabling multi-replica sources

### Checklist


ParkMyCar (approved):
Woohoo!

def- (commented):
Test changes lgtm, I triggered a nightly run: https://buildkite.com/materialize/nightly/builds/11070

def-:
I'm a bit confused by the data ingest error: https://buildkite.com/materialize/nightly/builds/11070#0194dc61-63a3-4e71-b78a-7dad5aa1bac1
```
psycopg.errors.InternalError_: cannot create source in cluster with more than one replica
```
I thought that shouldn't happen anymore since we enable multi-cluster replication now?

aljoscha:
@def- This only enables multi-replica sources for Kafka sources, the other sources are still an open TODO after this. And yeah, sorry because the issue title is confusing.

Fwiw, here's the overall EPIC: https://github.com/MaterializeInc/database-issues/issues/5051

def- (approved):
Another nightly run: https://buildkite.com/materialize/nightly/builds/11102 Edit: I'm fixing up data ingest

I also want to write some more tests, but can do it after this is merged too.

def-:
Something interesting seems to have happened in the data ingest test: https://buildkite.com/materialize/nightly/builds/11104#_
```
data-ingest-materialized-1     | 2025-02-10T18:41:59.892749Z  thread 'main' panicked at src/adapter/src/catalog/apply.rs:1024:33: PlanError(Unstructured("cannot create source in cluster with more than one replica")): invalid persisted SQL: CREATE SOURCE "materialize"."public"."kafka_table0" IN CLUSTER [u2] FROM KAFKA CONNECTION [u1 AS "materialize"."public"."kafka_conn"] (TOPIC = 'data-ingest-0') FORMAT AVRO USING CONFLUENT SCHEMA REGISTRY CONNECTION [u2 AS "materialize"."public"."csr_conn"] SEED KEY SCHEMA '{"type":"record","name":"key","fields":[{"name":"key0","type":"long"},{"name":"key1","type":"int"}]}' VALUE SCHEMA '{"type":"record","name":"value","fields":[{"name":"value0","type":"int"},{"name":"value1","type":"double"},{"name":"value2","type":"double"},{"name":"value3","type":"string"},{"name":"value4","type":"double"},{"name":"value5","type":"int"},{"name":"value6","type":"double"},{"name":"value7","type":"int"},{"name":"value8","type":"double"}]}' ENVELOPE UPSERT EXPOSE PROGRESS AS [u17 AS "materialize"."public"."kafka_table0_progress"]
```

src/storage-controller/src/lib.rs
```diff
@@ -1915,11 +1915,6 @@ where
                         .read_hold
                         .try_downgrade(Antichain::new())
                         .expect("must be possible");
-                } else {
-                    soft_panic_or_log!(
-                        "DroppedId for ID {id} but we have neither ingestion nor export \
-                         under that ID"
-                    );
```
teskje, line 1922: I think that soft panic should stay? Since we're doing the `active_copies` tracking there isn't a reason we should ever get here, right?

src/storage/src/source/reclock.rs
```diff
@@ -130,7 +130,8 @@ where
 
         while *self.upper == [IntoTime::minimum()]
             || (PartialOrder::less_equal(&self.source_upper.frontier(), &new_from_upper)
-                && PartialOrder::less_than(&self.upper, &new_into_upper))
+                && PartialOrder::less_than(&self.upper, &new_into_upper)
+                && self.upper.less_equal(&binding_ts))
```
teskje, line 134: Just flagging that it's not clear to me why this was changed, though it might be clear to someone with more context on the reclock operator.

src/storage-controller/src/lib.rs
```diff
@@ -474,14 +475,28 @@ where
             location,
             grpc_client: self.config.parameters.grpc_client.clone(),
         };
+        for id in instance.active_ingestions() {
+            self.collections
+                .get_mut(id)
+                .expect("instance contains unknown ingestion")
```
teskje, line 481: Can we make this a soft panic instead? I'm wary of panics that could bring down envd, if it's not immediately obvious that we can't hit them. And I don't think that's obvious here.

src/storage-controller/src/lib.rs
```diff
@@ -948,6 +963,7 @@ where
                 data_source,
                 collection_metadata: metadata,
                 extra_state,
+                active_copies: 1,
```
teskje, line 966: How do we know that there is one active copy when a collection is created? Don't we need to look at the number of replicas for that?

src/storage-controller/src/lib.rs
```diff
@@ -474,14 +475,28 @@ where
...
             .get_mut(&instance_id)
-            .unwrap_or_else(|| panic!("instance {instance_id} does not exist"))
-            .drop_replica(replica_id);
+            .unwrap_or_else(|| panic!("instance {instance_id} does not exist"));
+
+        for id in instance.active_ingestions() {
+            self.collections
+                .get_mut(id)
+                .expect("instance contains unknown ingestion")
+                .active_copies -= 1;
```
teskje, line 497: We might have to remove collection state if this value drops to zero. For example, consider the case where a source is dropped but before we receive the `DroppedId` response from the replica we drop the replica. With the current code I think we would just leak the `CollectionState` for the dropped collection.

This is tricky because we can't always drop a collection when this counter goes to zero. All replicas in a cluster can be dropped without the sources on the same cluster being dropped as well.

teskje (commented):
Looks mostly good, but I'm suspicious that the `active_copies` tracking might not always be correct. At least I don't understand why we can unconditionally initialize the number of active copies to 1 when a collection is created, and I think we can leak `CollectionState` when a replica is dropped at the wrong moment.

aljoscha:
> Looks mostly good, but I'm suspicious that the `active_copies` tracking might not always be correct. At least I don't understand why we can unconditionally initialize the number of active copies to 1 when a collection is created, and I think we can leak `CollectionState` when a replica is dropped at the wrong moment.

@petrosagg Could you please look at the questions about `active_copies` and the changes to reclock (basically all the questions above 😅)

aljoscha:
@teskje and @petrosagg For some of the trickier questions around collection state, `active_copies`, concurrency, assertions, and leaking state, I think you already know my general stance 😅

aljoscha:
@def- the nightly is failing with:
```
data-ingest-materialized-1     | 2025-02-10T18:41:59.892749Z  thread 'main' panicked at src/adapter/src/catalog/apply.rs:1024:33: PlanError(Unstructured("cannot create source in cluster with more than one replica")): invalid persisted SQL: CREATE SOURCE "materialize"."public"."kafka_table0" IN CLUSTER [u2] FROM KAFKA CONNECTION [u1 AS "materialize"."public"."kafka_conn"] (TOPIC = 'data-ingest-0') FORMAT AVRO USING CONFLUENT SCHEMA REGISTRY CONNECTION [u2 AS "materialize"."public"."csr_conn"] SEED KEY SCHEMA '{"type":"record","name":"key","fields":[{"name":"key0","type":"long"},{"name":"key1","type":"int"}]}' VALUE SCHEMA '{"type":"record","name":"value","fields":[{"name":"value0","type":"int"},{"name":"value1","type":"double"},{"name":"value2","type":"double"},{"name":"value3","type":"string"},{"name":"value4","type":"double"},{"name":"value5","type":"int"},{"name":"value6","type":"double"},{"name":"value7","type":"int"},{"name":"value8","type":"double"}]}' ENVELOPE UPSERT EXPOSE PROGRESS AS [u17 AS "materialize"."public"."kafka_table0_progress"]
```

Which happens when `enable_multi_replica_sources=on` is disabled but has been enabled before and there are multi-replica clusters with sources. We currently can't go back, because we would have to delete either replicas or sources to get back to a valid state with the flag disabled.

src/storage/src/source/reclock.rs
petrosagg: This is a bug fix. We want the mint operation to be a no-op if either `new_from_upper` or `binding_ts` are not beyond the current `source_upper` and `upper` respectively. The code on main only checked the former which meant that if the requested `binding_ts` was in the past we would attempt to write to the remap shard updates that are not beyond the upper.

src/storage-controller/src/lib.rs
teskje: I see. Would be great to have that context in a comment too!

aljoscha:
Rebased and fixed, @def- we still need to get to the bottom of ☝️ , are the tests flipping the dyncfg on and off?

def-:
No, they are not. The dyncfg should always be on in the Data Ingest test.

def-:
What might happen is that during bootstrapping we parse the SQL without respecting the dyncfg setting?

aljoscha:
@def- what about that remaining cloudtest failure here: https://buildkite.com/materialize/nightly/builds/11180?

def-:
I have pushed a fix: https://github.com/MaterializeInc/materialize/pull/31227/commits/2e0132c60617dbc45e2d579c1b7ec808b502d779 @jubrad does this make sense?

aljoscha:
> I have pushed a fix: [2e0132c](https://github.com/MaterializeInc/materialize/commit/2e0132c60617dbc45e2d579c1b7ec808b502d779) @jubrad does this make sense?

doh! tyty! 🙇‍♂️

aljoscha:
@def- I pushed https://github.com/MaterializeInc/materialize/pull/31227/commits/a2fe527be7044b144d53cf91f056d8080e631c4d because this was yielding

```
error: expected error containing "creating cluster replica would violate max_replicas_per_cluster limit", got "unknown cluster replica size 2000"
```

aljoscha:
@def- I think you pushed a commit that changes the expected error message, on top of mine which changes size to `replication factor` 😅

def-:
Oops, didn't see your comment and just pushed something similar, sorry!

aljoscha:
> Oops, didn't see your comment and just pushed something similar, sorry!

No worries at all! ☺️

PR #31229, opened 2025-01-29
Combined logging dataflows

antiguru:
Combine all logging dataflows into a single dataflow.

This allows easier sharing between the dataflows as events can be send through streams, instead of injecting them into the log events for later processing. The change looks substantial, but shouldn't alter behavior other than that injected commands pass through actual streams instead of injecting them into the logging dataflows.

This has some benefits: The injected commands are consistent, i.e., they arrive at the expected time, and this change enables us to share information across logging dataflows more easily, for example to do analysis such as #31246.

### Checklist


src/compute/src/compute_state.rs
```diff
@@ -1632,13 +1641,16 @@ pub struct CollectionState {
     logging: Option<CollectionLogging>,
     /// Metrics tracked for this collection.
     metrics: CollectionMetrics,
+    /// The dataflow ID of the dataflow that hosts this collection.
+    pub dataflow_id: usize,
```
teskje, line 1645: This field also looks unused.

src/compute/src/logging/initialize.rs
```diff
@@ -96,54 +94,72 @@ struct LoggingContext<'a, A: Allocate> {
     shared_state: Rc<RefCell<SharedLoggingState>>,
 }
 
+pub(crate) struct LoggingDataflows {
+    pub logs: BTreeMap<LogVariant, (TraceBundle, usize)>,
+}
```
teskje, line 99: Since there is only a single dataflow index now, we don't need to mention it for each log.

```suggestion
pub(crate) struct LoggingDataflows {
    pub traces: BTreeMap<LogVariant, TraceBundle>,
    pub dataflow_index: usize,
}
```

Also removes the unnamed `usize`, which is nice.

src/compute/src/logging/initialize.rs
```diff
@@ -96,54 +94,72 @@ struct LoggingContext<'a, A: Allocate> {
...
-            .into_iter()
-            .map(|(log, collection)| {
-                let bundle =
-                    TraceBundle::new(collection.trace, errs.clone()).with_drop(collection.token);
-                (log, (bundle, collection.dataflow_index))
-            })
-            .collect()
+            // TODO(vmarcos): If we introduce introspection sources that would match
+            // type specialization for keys, we'd need to ensure that type specialized
+            // variants reach the map below (issue database-issues#6763).
```
teskje, line 152: Given that we've moved away from arrangement specialization, maybe this is a good time to remove this comment?

src/compute/src/logging/compute.rs
```diff
@@ -289,24 +285,28 @@ impl LirMetadata {
...
     shared_state: Rc<RefCell<SharedLoggingState>>,
-) -> BTreeMap<LogVariant, LogCollection> {
+) -> Return {
     let logging_interval_ms = std::cmp::max(1, config.interval.as_millis());
-    let worker_id = worker.index();
-    let worker2 = worker.clone();
-    let dataflow_index = worker.next_dataflow_index();
 
-    worker.dataflow_named("Dataflow: compute logging", move |scope| {
+    scope.scoped("Dataflow: compute logging", move |scope| {
```
teskje, line 309: I think we should remove the `Dataflow: ` prefix for the regions of the logging dataflow, to avoid confusion.

src/compute/src/logging/compute.rs
```diff
@@ -289,24 +285,28 @@ impl LirMetadata {
...
 ///
 /// Params
-/// * `worker`: The Timely worker hosting the log analysis dataflow.
+/// * `scope`: The Timely scope hosting the log analysis dataflow.
+/// * `scheduler`: The timely scheduler to obtainer activators.
 /// * `config`: Logging configuration.
 /// * `event_queue`: The source to read compute log events from.
-pub(super) fn construct<A: Allocate + 'static>(
-    worker: &mut timely::worker::Worker<A>,
+pub(super) fn construct<A: Scheduler + 'static, S: Scope<Timestamp = Timestamp>>(
```
teskje, line 299: Nit: `A` seems like a strange name for a `Scheduler`. But `S` is already taken, so I don't have a better suggestion either 😅

teskje (commented):
Looks good, my comments are mostly about removing/simplifying stuff. There is one about not naming the logging regions `Dataflow: *` that I think is important though.

src/compute/src/compute_state.rs
```diff
@@ -175,6 +175,9 @@ pub struct ComputeState {
     /// replica can drop diffs associated with timestamps beyond the replica expiration.
     /// The replica will panic if such dataflows are not dropped before the replica has expired.
     pub replica_expiration: Antichain<Timestamp>,
+
+    /// The interval for rounding logging timestamps.
+    logging_interval: Duration,
```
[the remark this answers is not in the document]
antiguru: Removed.

src/compute/src/logging/differential.rs
```diff
@@ -143,7 +152,7 @@ pub(super) fn construct<A: Allocate>(
         let arrangement_records = stream_to_collection(&records, ArrangementRecords);
         let sharing = stream_to_collection(&sharing, Sharing);
         let batcher_records = stream_to_collection(&batcher_records, BatcherRecords);
-        let batcher_size = stream_to_collection(&batcher_size, BatcherSize);
+        let batcher_size_rows = stream_to_collection(&batcher_size, BatcherSize);
```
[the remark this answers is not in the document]
antiguru: Removed

src/compute/src/logging/compute.rs
antiguru: Solved! Scope is G, scheduler is S. At least we G for scope in many places, so it's not too bad.

teskje: Works for me! Could you also change the name of the `DemuxState` type variable?

teskje (approved):
LGTM once CI agrees. Might be a good idea to run the Nightlies as well.

PR #31230, opened 2025-01-29
Support lgalloc for columnar

antiguru:
Adds support for allocating memory for columnar's aligned type from
lgalloc. Adds a flag `enable_lgalloc_columnar` to control the behavior.

### Checklist


src/timely-util/src/containers.rs
```diff
@@ -217,6 +231,28 @@ mod builder {
...
+    /// Allocate a region of memory with a capacity of at least `len` that is aligned to 8 bytes
+    /// and zeroed.
+    #[inline]
+    pub(crate) fn alloc_aligned_zeroed(len: usize) -> Region<u64> {
+        if enable_lgalloc_columnar() {
+            Region::new_auto_zeroed(len)
+        } else {
+            Region::new_heap_zeroed(len)
+        }
+    }
```
ParkMyCar, line 254: Mind adding to this comment that it's aligned to 8 bytes because it's a Region of `u64`s?

src/timely-util/src/containers.rs
```diff
@@ -229,14 +265,32 @@ mod builder {
...
+                    let mut alloc = alloc_aligned_zeroed(round);
+                    let mut writer = std::io::Cursor::new(bytemuck::cast_slice_mut(&mut alloc[..]));
+                    for (align, bytes) in current.borrow().as_bytes() {
+                        assert!(align <= 8);
+                        let length = u64::cast_from(bytes.len());
+                        let length_slice = bytemuck::cast_slice(std::slice::from_ref(&length));
+                        writer.write_all(length_slice).unwrap();
+                        writer.write_all(bytes).unwrap();
+                        let padding = usize::cast_from((8 - (length % 8)) % 8);
+                        writer.write_all(&[0; 8][..padding]).unwrap();
```
ParkMyCar, line 287: I'm not totally up-to-date on the encoded layout of a `Column`, but my read is this layouts data like:
```
<u64 encoded in native endianness><buffer whose length is the u64><optional padding>
```
If that's what it's supposed to do then it LGTM!

ParkMyCar (approved):

src/timely-util/src/containers.rs
teskje, line 287: Is there a `columnar` method we could invoke so we don't have to worry about this code diverging from the `columnar` implementation?

src/timely-util/src/containers/array.rs
```diff
@@ -120,6 +128,22 @@ impl<T> Deref for Array<T> {
...
 
+impl<T> DerefMut for Array<T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        // TODO: Use `slice_assume_init_ref` once stable.
+        // Context: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.slice_assume_init_ref
+        // The following safety argument is adapted from the source.
+        // SAFETY: casting `elements` to a `*const [T]` is safe since the caller guarantees that
+        // `slice` is initialized, and `MaybeUninit` is guaranteed to have the same layout as `T`.
+        // The pointer obtained is valid since it refers to memory owned by `elements` which is a
+        // reference and thus guaranteed to be valid for reads.
```
teskje, line 139: This safety argument talks about `*const` and reads, but we're casting to a `*mut` and deference that for writes. I think a similar argument still holds, based on that `elements` is mutably borrowed here, so we can produce a `&mut` into it.

src/timely-util/src/containers/array.rs
```diff
@@ -120,6 +128,22 @@ impl<T> Deref for Array<T> {
     }
 }
 
+impl<T> DerefMut for Array<T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        // TODO: Use `slice_assume_init_ref` once stable.
```
frankmcsherry, line 133: Probably `slice_assume_init_mut` rather than `_ref`, I think?

frankmcsherry: If nothing else, we'll want to have the `unsafe` justification for `_mut` instead, I think. Maybe a nit, but "have the right explanation when we use unsafe" seems like a great place to pick nits.

frankmcsherry (approved):
Seems good, though there are a few moments I'm not entirely clear on. But structurally the PR seems ok!

src/timely-util/src/containers.rs
antiguru: Removed code and replaced by a call to columnar's API, which now supports `write`.

src/ore/src/region.rs
```diff
@@ -342,6 +344,48 @@ impl<T> Region<T> {
...
+    /// capacity of the returned region can be larger than requested to accommodate page sizes.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the memory allocation fails.
+    #[inline(always)]
+    pub fn new_mmap_zeroed(capacity: usize) -> Result<Region<u64>, lgalloc::AllocError> {
+        lgalloc::allocate::<u64>(capacity).map(|(ptr, capacity, handle)| {
+            // SAFETY: `allocate` returns a valid memory block
+            unsafe { ptr.as_ptr().write_bytes(0, capacity) }
```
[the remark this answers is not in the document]
antiguru: I'm not too worried about the cost of setting the memory to zero, especially since we'll write to it rather soon, anyways. We could consider not touching the memory, which would leave whatever was there previously in-place, but I'd prefer to do so only if we notice that this is a potential bottleneck.

src/timely-util/src/containers.rs
```diff
@@ -194,16 +200,24 @@ mod container {
...
     use columnar::{AsBytes, Clear, Columnar, Len, Push};
+    use mz_ore::cast::CastFrom;
+    use mz_ore::region::Region;
     use timely::container::PushInto;
     use timely::container::{ContainerBuilder, LengthPreservingContainerBuilder};
 
     use super::Column;
 
+    thread_local! {
+        static ENABLE_LGALLOC_COLUMNAR: std::cell::RefCell<bool> = const { std::cell::RefCell::new(false) };
```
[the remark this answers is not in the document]
antiguru: I'll switch it to `Cell`. The `const` is important because otherwise Rust will generate constructors/destructors for the thread-local. It's significantly less code if wrapped in `const`. See https://godbolt.org/z/WYYd6M85z.

src/timely-util/src/containers/array.rs
antiguru: Fortunately, that code is gone :)

src/ore/src/region.rs
```diff
@@ -342,6 +344,48 @@ impl<T> Region<T> {
     }
 }
 
+impl Region<u64> {
```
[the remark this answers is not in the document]
antiguru: Changing to `bytemuck::AnyBitPattern` to be less opinionated (and special-cased to columnar).

src/timely-util/src/containers/array.rs
antiguru: Fortunately, that code is gone :)

src/timely-util/src/containers/array.rs
antiguru: Code has been removed.

src/compute-types/src/dyncfgs.rs
```diff
@@ -50,10 +50,17 @@ pub const LINEAR_JOIN_YIELDING: Config<&str> = Config::new(
...
 pub const ENABLE_COLUMNATION_LGALLOC: Config<bool> = Config::new(
     "enable_columnation_lgalloc",
-    false,
+    true,
     "Enable allocating regions from lgalloc.",
 );
 
+/// Enable lgalloc for columnation.
+pub const ENABLE_LGALLOC_COLUMNAR: Config<bool> = Config::new(
+    "enable_lgalloc_columnar",
```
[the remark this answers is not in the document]
antiguru: Done

teskje (approved):

PR #31233, opened 2025-01-29
adapter: Enrich expr cache version in dev

jkosh44:
The expression cache uses the build version as part of the key to ensure that a materialize instance only tries to deserialize expressions that were serialized using the same version of the code. However, a single dev build version can be used for many different builds, which can lead to errors in deserialization when running locally. This commit fixes the issue by enriching dev build versions with build metadata containing the build ID.

Note: we could enrich all build versions with build ID metadata without any correctness issues. However, the ID is fairly long and must be stored for each expression. Since we already have the guarantee that a prod build version is only ever used with a single build, it would be wasted space for prod build versions.

Fixes #MaterializeInc/database-issues/issues/8930

### Motivation
This PR fixes a recognized bug.

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1410,6 +1410,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "buildid"
+version = "1.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0188344028070a23e23eebec37f4c5699db7b5a5fd088d97e6fae7727ac0dd77"
+dependencies = [
+ "cfg-if",
+ "libc",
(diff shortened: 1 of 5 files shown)


jkosh44:
Can someone help me with this lint error? It looks like `buildid` pulls in `log` which is banned, but I'm not entirely sure why that would be or what my options are to prevent that.

https://buildkite.com/materialize/test/builds/97900#0194b302-309f-48d2-a708-a6d7fbed708a

def-:
let me check

antiguru (approved):
Looks great, thanks!

PR #31239, opened 2025-01-29
mysql: add support for invisible columns

martykulma:
To support invisible columns in MySQL, explicitly request columns in snapshot select statement. And while I'm here, I corrected the glob root path in mzcompose.py to get the tests to run.

### Motivation

Adds support for invisible columns in mysql
- https://github.com/MaterializeInc/database-issues/issues/7635
- https://github.com/MaterializeInc/database-issues/issues/7782

### Checklist

### DIFF
--- a/ci/nightly/pipeline.template.yml
+++ b/ci/nightly/pipeline.template.yml
@@ -927,7 +927,7 @@ steps:
       - id: checks-0dt-restart-entire-mz-forced-migrations-azurite
         label: "Checks 0dt restart of the entire Mz with forced migrations with :azure: blob store"
         depends_on: build-aarch64
-        timeout_in_minutes: 120
+        timeout_in_minutes: 180
         parallelism: 2
         agents:
           queue: hetzner-aarch64-16cpu-32gb
--- a/ci/test/pipeline.template.yml
+++ b/ci/test/pipeline.template.yml
@@ -410,7 +410,7 @@ steps:
 
       - id: mysql-cdc-resumption
         label: "MySQL CDC resumption tests"
-        parallelism: 4
+        parallelism: 6
         depends_on: build-aarch64
         timeout_in_minutes: 30
         inputs: [test/mysql-cdc-resumption]


src/storage/src/source/mysql/snapshot.rs
```diff
@@ -404,7 +405,18 @@ pub(crate) fn render<G: Scope<Timestamp = GtidPartition>>(
...
+                    // to be retrieved.
+                    let columns = outputs
+                        .iter()
+                        .flat_map(|info| {
+                            info.desc
+                                .columns
+                                .iter()
+                                .map(|col_info| format!("`{}`", col_info.name))
+                        })
+                        .join(", ");
```
petrosagg, line 418: Hm, I think this might be doing a different thing than what we need. Here you iterate over all outputs and concatenate the name of the columns from each one of them. So if I understand correctly, if the same table is ingested twice in Materialize we will end up with a query like `SELECT col1, col2, col1, col2` and transfer twice the amount of data. Am I reading this right?

test/mysql-cdc/mzcompose.py
```diff
@@ -99,7 +99,9 @@ def workflow_cdc(c: Composition, parser: WorkflowArgumentParser) -> None:
 
     matching_files = []
     for filter in args.filter:
-        matching_files.extend(glob.glob(filter, root_dir="test/mysql-cdc"))
+        matching_files.extend(
+            glob.glob(filter, root_dir=MZ_ROOT / "test" / "mysql-cdc")
```
petrosagg, line 103: Thanks for this fix! I've been bitten by it!

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -517,6 +517,22 @@ where
...
+#[must_use]
+fn build_snapshot_query(outputs: &[SourceOutputInfo]) -> String {
+    let info = outputs.first().expect("MySQL table info");
+    let columns = info
+        .desc
+        .columns
+        .iter()
+        .map(|col| format!("`{}`", col.name))
+        .join(", ");
+    format!("SELECT {} FROM {}", columns, info.table_name)
```
antiguru, line 533: (sorry for the drive-by review) Do we have to worry about SQL injection here?

petrosagg: I did wonder about that during review and concluded that we don't have to be worried as to perform an injection you have to somehow create weirdly named columns, which means you already admin level have access to the database. My understanding is that we can't use prepared statements here since with those you can put placeholders for values but here we want to put placeholders for column names. If it's possible and easy to do we should maybe do it

src/mysql-util/src/lib.rs
```diff
@@ -114,10 +114,31 @@ pub enum MySqlError {
...
+mod tests {
+
+    use super::quote_identifier;
+    #[mz_ore::test]
+    fn test_identifier_quoting() {
+        let expected = vec!["`a`", "`naughty``sql`", "```;naughty;sql;```"];
+        let input = ["a", "naughty`sql", "`;naughty;sql;`"].iter().map(|raw_str| quote_identifier(&raw_str)).collect::<Vec<_>>();
+        assert_eq!(expected, input);
+    }
+}
```
petrosagg: Looks like your editor doesn't add trailing new lines, this might generate spurious diffs. Not blocking for this PR but something to configure

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -524,11 +526,20 @@ where
 #[must_use]
 fn build_snapshot_query(outputs: &[SourceOutputInfo]) -> String {
     let info = outputs.first().expect("MySQL table info");
+    if outputs.len() > 1
+        && !outputs
+            .iter()
+            .skip(1)
+            .map(|other_info| &other_info.desc.columns)
+            .all(|other_columns| *other_columns == info.desc.columns)
```
petrosagg, line 534: small nit: in idiomatic Rust, at least in the MZ codebase, it usually better to avoid long chains of iterator transformations with either a short chain if possible or just a imperative for loop. This functional style becomes harder to read after a certain point. 

So I would write this either as:

 ```rust
assert!(outputs.iter().all(|o| info.desc.columns == o.desc.columns), "Mismatch...");
 ```
 
 or:
 
 ```rust
for output in outputs[1..] {
    assert!(info.desc.columns == output.desc.columns), "Mismatch...");
}
```

File: src/storage/src/source/mysql/snapshot.rs (hunk 3 above)
Comment (petrosagg) [reply src/storage/src/source/mysql/snapshot.rs:544]: nice!

**Review by petrosagg — APPROVED** [review]
Looks great!

**Review by def- — APPROVED** [review]
Thanks for adding a test. I have added a [platform checks](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/platform-checks.md) check, nightly run with it: https://buildkite.com/materialize/nightly/builds/11054

**Comment by def-** [comment]
I find it slightly suspicious that the MySQL CDC resumption test timed out twice on this PR: https://buildkite.com/materialize/test/builds/98149#0194d5c5-ceb8-4010-9f7f-a41aa8ce2acf
It doesn't seem to get stuck, but just seems to run slower. I will bump the timeout for now, but is this expected?

**Comment by petrosagg** [comment]
hm, https://github.com/MaterializeInc/database-issues/issues/8884 seems a bit in the weeds since it involves a failover scenario so I wouldn't dig into it at the moment. 

I compared the set of dataflow restarts that happened in this run vs one from main that succeeds and this is the difference:

```diff
-INFO mz_storage::healthcheck: Broadcasting suspend-and-restart command because of Some((MySql, Stalled { error: "Input/output error: Input/output error: Connection refused (os error 111): Input/output error: Connection refused (os error 111): Connection refused (os error 111)", hint: None, should_halt: true })) after 5s delay
+INFO mz_storage::healthcheck: Broadcasting suspend-and-restart command because of Some((MySql, Stalled { error: "Input/output error: Input/output error: connection closed: Input/output error: connection closed: connection closed", hint: None, should_halt: true })) after 5s delay
+INFO mz_storage::healthcheck: Broadcasting suspend-and-restart command because of Some((MySql, Stalled { error: "Input/output error: Input/output error: connection closed: Input/output error: connection closed: connection closed", hint: None, should_halt: true })) after 5s delay
+INFO mz_storage::healthcheck: Broadcasting suspend-and-restart command because of Some((MySql, Stalled { error: "Input/output error: Input/output error: connection closed: Input/output error: connection closed: connection closed", hint: None, should_halt: true })) after 5s delay
```

(For Marty: these logs are in each buildkite run under the "Artifacts" tab and named "services.log")

So it looks like in this run the dataflow found the MySQL service down more times. The run on main takes 24 minutes which is already close to the 30 minute timeout. At the same time these three restarts should take less than 6 minutes to complete so it's not that they are a complete explanation.

Given the change is pretty simple I'm inclined to just merge this with the increased timeout without further investigation.

PR #31246, opened 2025-01-30
Introduce option to limit the heap capacity for a dataflow

antiguru:
Observe a dataflow's arrangement sizes and notify controller once it's over a limit. Let the controller handle the next step.

Related design: #31277

Missing from this PR:
* A feature flag to disable the behavior.
* Documentation for users.

### Checklist


src/adapter/src/coord.rs
```diff
@@ -467,6 +467,7 @@ pub struct PeekStageLinearizeTimestamp {
     validity: PlanValidity,
     plan: mz_sql::plan::SelectPlan,
     max_query_result_size: Option<u64>,
+    max_query_heap_size: Option<u64>,
```
teskje, line 470: Seems like a good idea to bundle these into a `QueryResourceLimits` or similar, to making passing them around less boilerplate-y.

src/adapter/src/coord/sequencer/inner.rs
```diff
@@ -3004,6 +3005,7 @@ impl Coordinator {
             },
             TargetCluster::Active,
             None,
+            None,
```
teskje, line 3008: Why are we not enforcing limits here? Afaict this is the code path for `INSERT/UPDATE ... SELECT` and nothing prevents the SELECT from being expensive.

src/compute-client/src/controller/instance.rs
```diff
@@ -2015,6 +2016,39 @@ where
...
+                    peek_uuid = Some(*uuid);
+                    break;
+                }
+            }
+            if let Some(uuid) = peek_uuid {
+                self.cancel_peek(
+                    uuid,
+                    PeekResponse::Error("Dataflow limit exceeded".to_string()),
+                );
+            }
```
teskje, line 2051: Some thoughts:
* It's awkward that we can't immediately cancel the offending dataflow here because controller clients don't expect dataflows to get dropped silently. If we had a way to mark dataflows as "failed" we could use that here to drop them aggressively. Since we don't have that yet, I think the current way to report an error and then rely on the client to cancel the dataflow in a timely fashion makes sense.
* Controller clients can specify a heap size limit for dataflows that are neither the target of peeks nor subscribes. We don't support this currently, so I think we should log an error here if we don't find a subscribe/peek.
* It seems important to report, _which_ limit was exceeded, because users will definitely ask that once they get an error like this. The current "dataflow limit exceeded" response doesn't really provide that level of detail, so maybe we should just rename it to "dataflow heap limit exceeded"? Or add the limit type to the status, if we think that we might have more limits like this in the future.

src/compute-client/src/controller/instance.rs
```diff
@@ -2015,6 +2016,39 @@ where
...
+                SubscribeBatch {
+                    lower: subscribe.frontier.clone(),
+                    upper: subscribe.frontier.clone(),
+                    updates: Err("Dataflow limit exceeded".to_string()),
+                },
+            ))
+        } else {
+            // Look for a matching peek
+            let mut peek_uuid = None;
+            for (uuid, peek) in self.peeks.iter() {
```
teskje, line 2040: Nit: `.find(...).map(...)` might end up a bit more readable... or not :D

src/compute-client/src/protocol/response.proto
```diff
@@ -107,6 +107,7 @@ message ProtoSubscribeBatch {
 message ProtoStatusResponse {
   oneof kind {
     ProtoOperatorHydrationStatus operator_hydration = 1;
+    ProtoDataflowLimitStatus dataflow_limit_status = 2;
```
teskje, line 110: I think this field should be called `dataflow_limit_exceeded`, since that's what the `StatusResponse` variant is called too?

src/compute-client/src/protocol/response.rs
```diff
@@ -650,6 +656,29 @@ impl RustType<ProtoOperatorHydrationStatus> for OperatorHydrationStatus {
     }
 }
 
+/// A dataflow exceeded some limit.
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, Arbitrary)]
+pub struct DataflowLimitStatus {
+    /// The ID of the compute collection exported by the dataflow.
+    pub collection_id: GlobalId,
+}
```
teskje, line 664: Mentioned above, but we should also report, _which_ limit!

teskje (approved):
Releasing some comments already, but I only made it through adapter and the compute controller. Looks good so far though!

src/adapter/src/coord.rs
frankmcsherry: I think there's maybe a larger refactoring around here (the first ~900 lines of `coord.rs` are types without documentation, other than some field-level doccomments). Imo, maybe this PR isn't the place to do that, but rather a more holistic sweep?

frankmcsherry: Having said that, the next several pages of diffs do seem to be more boilerplate, pairing up the new parameter with the prior one. I may be coming around!

teskje: Should have called this a nit! Imo it's good to follow the boy scout rule when it's not too much of a hassle, because we might never get to that holistic refactor. But nothing I'd block the PR on.

src/adapter/src/coord/sequencer/inner.rs
ParkMyCar: +1

What's tricky here is the already existing variable `max_query_result_size != max_result_size`. The former limits how large of a result we'll return to the user, while the later limits how large of a result we'll allow `clusterd` to send to `environmentd`.

Some more context is [here](https://github.com/MaterializeInc/database-issues/issues/8747) but because the results of this peek are not generally user facing (except in the case where the user specifies a `RETURNING` clause) we don't want to set a `max_query_result_size`.

test/testdrive/query_heap_size.td
```diff
@@ -0,0 +1,46 @@
...
+> FETCH 1 c;
+<TIMESTAMP> 1 10000
+
+# No output should be produced. Instead an error .. notice?
+! FETCH 1 c;
+contains:Dataflow limit exceeded
+
+> ROLLBACK;
+
+> SET max_query_heap_size = '128MB';
```
ParkMyCar, line 35: Can you also test the behavior of `RESET max_query_heap_size` and `SET max_query_heap_size = 0`? Both should reset the size to `None`

src/adapter/src/coord.rs
ParkMyCar: +1 to bundling into a `QueryResourceLimits` but also not blocking

ParkMyCar (approved):
Adapter changes LGTM

antiguru:
Superseded by #38547, which picks this up against current `main`. The replica-side code there is a rewrite rather than a rebase, so the review threads here do not carry over; #38547 records how each of them was handled.

🤖 Posted by [Claude Code](https://claude.com/claude-code)

PR #31248, opened 2025-01-30
Advance and consolidate remap updates

frankmcsherry:
The `reclock` operator accepts and enqueues all remap updates, and requires memory proportional to the number of updates supplied. This has the potential to be much larger than required if the updates when advanced using `as_of` would consolidate to a much smaller collection.

This PR advances all inbound times using `as_of`, and consolidates them before enqueueing them in the heap-based holding pen. This does not guarantee that they will fully consolidate, as it depends on non-determinism of data arrival, and a more thorough solution might look more like DD's `MergeBatcher` that uses a continual merge sort to maintain consolidated priority order. Ideally this temporary fix relieves some pressure, though.

### Checklist

### DIFF
--- a/src/timely-util/src/reclock.rs
+++ b/src/timely-util/src/reclock.rs
@@ -249,6 +249,10 @@ where
         let mut binding_buffer = Vec::new();
         let mut interesting_times = Vec::new();
 
+        // Accumulation buffer for `remap_input` updates.
+        use timely::progress::ChangeBatch;
+        let mut remap_accum_buffer: ChangeBatch<(IntoTime, FromTime)> = ChangeBatch::new();
+
         // The operator drains `remap_input` and organizes new bindings that are not beyond
         // `remap_input`'s frontier into the time ordered `remap_trace`.
         //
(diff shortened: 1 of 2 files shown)


sjwiesman:
Not here to review the code but looking at an live customer environment, I see a tremendous amount of memory a vec and binary heap and it looks like we never shrink those collections even after the data is conslidated. Does it make sense to add some sort of shink to fit if the size is < 50% of the capacity?

antiguru:
Doesn't the `reclock_committed_upper` operator have a similar problem? At least, it stands out in the profile, too, and it is upstream to the `reclock` operator this PR adjusts.

frankmcsherry:
It seems like it might, but I'm not clear on what this function does. It also drains bindings directly into a vector, and if it is provided with many bindings will simply fall over. It seems plausible it could also be drained into a `ChangeBatch` to consolidate in place, but I'm not exactly sure what it needs to do. It .. fills, sorts, and then drains, but does not appear to perform any consolidation. I think we need to tap in the code owner to figure this out!

src/timely-util/src/reclock.rs
```diff
@@ -268,11 +268,19 @@ where
...
-                    pending_remap.push(Reverse((into, from, diff)));
+                for (from, mut into, diff) in data.drain(..) {
+                    into.advance_by(as_of.borrow());
+                    consolidated.update((into, from), diff);
                 }
             }
+            // Drain consolidated bindings into the `pending_remap` heap.
+            for ((into, from), diff) in consolidated.drain() {
+                pending_remap.push(Reverse((into, from, diff)));
+            }
```
petrosagg, line 283: I was expecting a more aggressive optimization here but maybe this one is enough to fix our memory problems. I was expecting that we have a separate holding pen just for the timestamp of the `as_of` frontier that gets all updates that advance to exactly that timestamp and gets consolidated entirely in each operator invocation. This would avoid the sensitivity of this implementation to how finely batched the snapshot comes in as.

frankmcsherry: Good point. I didn't do that as I think it requires more thought to understand the cross-invocation data use. The PR will consolidate everything that is newly available each time it is scheduled, but to fully consolidate a snapshot, even when the snapshot updates are slow-rolled by the source, seems to want a replacement for the binary heap that has, potentially, a pair of heap and vector, with the vector maintaining the consolidated snapshot.

frankmcsherry: I think it should be the case that we could tweak the PR logic to accumulate into a `ChangeBatch` until such a point that the input frontier has passed `as_of`, at which point we can disgorge the change batch accumulation into the binary heap. I would say that we could guard the
```rust
            // Drain consolidated bindings into the `pending_remap` heap.
            for ((into, from), diff) in consolidated.drain() {
                pending_remap.push(Reverse((into, from, diff)));
            }
```
logic by looking at `remap_upper` and relating it to `as_of`: for as long as the former is less or equal to the latter, we should continue to accumulate, and once that is no longer the case we can drain the consolidated bindings whenever they occur.

If that checks out, it is a pretty easy change to make.

frankmcsherry: Pushed a new commit that believes it implements this.

frankmcsherry: This still has opportunities for improvement for partially ordered time, in that we could continue accumulating any time in `as_of`, rather than waiting for at least one time and then stopping accumulation for all of them. I *think* that we aren't worried about partially ordered `IntoTime` here, but .. wanted to point this out.

frankmcsherry:
Oops. Bug with the most recent commit: the `consolidated` change batch needs to be defined outside the closure now that it is not always drained each invocation. Will fix tomorrow!

petrosagg (approved):

petrosagg (approved):
Last commit also LGTM, thank you!

PR #31249, opened 2025-01-30
Listen on 0.0.0.0:6876 through cli

antiguru:
At the moment, we're exposing the default DB port 6876 only to localhost,
but the internal one to all addresses. This makes it difficult to test the
console locally because it needs access to 6876, and running it on the
host's network can cause other port collisions.

For this reason, open up connections to 6876 to all hosts. This is not a
security regression as we're already opening the internal endpoint to all
addresses.

### Checklist

### DIFF
--- a/misc/python/materialize/cli/run.py
+++ b/misc/python/materialize/cli/run.py
@@ -277,6 +277,7 @@ def main() -> int:
                 # Setting the listen addresses below to 0.0.0.0 is required
                 # to allow Prometheus running in Docker (misc/prometheus)
                 # access these services to scrape metrics.
+                "--http-listen-addr=0.0.0.0:6876",
                 "--internal-http-listen-addr=0.0.0.0:6878",
                 "--orchestrator=process",
                 f"--orchestrator-process-secrets-directory={MZDATA}/secrets",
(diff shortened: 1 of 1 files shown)


benesch (commented):
Seems fine to me, but deferring to @def- who defaulted these to bind to 127.0.0.1 in the first place. Perhaps we should just wholesale roll that change back? If 6876 and 6878 are exposed, there's no security benefit to leaving 6875 and 6877 closed.

def- (approved):
I cared mostly about changing the defaults for the docker emulator so that people don't run our tutorials and accidentally expose Materialize to the world.

misc/python/materialize/cli/run.py
```diff
@@ -277,6 +277,7 @@ def main() -> int:
                 # Setting the listen addresses below to 0.0.0.0 is required
                 # to allow Prometheus running in Docker (misc/prometheus)
                 # access these services to scrape metrics.
```
teskje, line 279: Maybe adjust the comment too? Otherwise someone might in the future realize that Prometheus only needs 6878 and revert the change.

antiguru: Updated!

PR #31259, opened 2025-01-31
Remove redundant types in compute

antiguru:
Removes the MzArrangement, MzArrangementImport, SpecializedTraceHandle, and FromDatumIter types. They were mostly boilerplate that didn't add much value, or were enums with a single variant.

### Checklist

### DIFF
--- a/src/compute/src/render/context.rs
+++ b/src/compute/src/render/context.rs
@@ -36,7 +36,7 @@ use timely::dataflow::channels::pact::{ExchangeCore, Pipeline};
 use timely::dataflow::operators::generic::OutputHandleCore;
 use timely::dataflow::operators::Capability;
 use timely::dataflow::scopes::Child;
-use timely::dataflow::Scope;
+use timely::dataflow::{Scope, Stream};
 use timely::progress::timestamp::Refines;
 use timely::progress::{Antichain, Timestamp};
 use timely::Container;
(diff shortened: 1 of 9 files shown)


src/compute/src/render/context.rs
```diff
@@ -550,14 +332,30 @@ where
...
-            ),
+            ArrangementFlavor::Local(oks, errs) => {
+                let mut datums = DatumVec::new();
+                (
+                    oks.as_collection(move |k, v| {
+                        let mut datums_borrow = datums.borrow();
+                        datums_borrow.extend(k.to_datum_iter());
+                        datums_borrow.extend(v.to_datum_iter());
+                        SharedRow::pack(&**datums_borrow)
+                    }),
```
frankmcsherry, line 343: I can't really tell why this got more complicated. I can believe it all ties out, but paused here to acknowledge an understanding gap.

frankmcsherry (approved):
Looks good to me. I left a comment at a moment where things got more complicated rather than less complicated, and if it turns out there is a thing to say there about why it is now complicated, or .. what needs to be true for it to become less complicated, that would rule.

antiguru:
I pushed another commit that cleans up how we use `as_collection` and `flat_map`. I realized that previously `as_collection` wasn't fueled, it went directly to differential's `as_collection`. With the additional commit, it uses `flat_map` instead, so calls to `as_collection` are fueled, too.

frankmcsherry:
Ah, looks reasonable though I've lost the thread of some of the changes (looking at commits and losing context). If the PR is also changing the fueling of operators, I think it makes sense to split that out as a behavioral change.

antiguru:
I restored the old behavior that `as_collection` is not fueled.

PR #31261, opened 2025-01-31
storage: flatten `StorageCommand` and `StorageResponse`

petrosagg:
### Tips for reviewer

Each commit flattens one variant

### Checklist

### DIFF
--- a/src/storage-client/src/client.rs
+++ b/src/storage-client/src/client.rs
@@ -127,8 +127,8 @@ pub enum StorageCommand<T = mz_repr::Timestamp> {
     AllowWrites,
     /// Update storage instance configuration.
     UpdateConfiguration(StorageParameters),
-    /// Run the enumerated sources, each associated with its identifier.
-    RunIngestions(Vec<RunIngestionCommand>),
+    /// Run the specified ingestion dataflow.
+    RunIngestion(RunIngestionCommand),
     /// Enable compaction in storage-managed collections.
     ///
     /// A collection id and a frontier after which accumulations must be correct.
@@ -167,7 +167,7 @@ impl<T> StorageCommand<T> {
             // TODO(cf2): multi-replica oneshot ingestions. At the moment returning
             // true here means we can't run `COPY FROM` on multi-replica clusters, this
             // should be easy enough to support though.
-            RunIngestions(_) | RunSinks(_) | RunOneshotIngestion(_) => true,
+            RunIngestion(_) | RunSinks(_) | RunOneshotIngestion(_) => true,
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/storage-controller/src/lib.rs
```diff
@@ -1892,122 +1887,129 @@ where
...
-                        .expect("must be possible");
-                } else {
-                    soft_panic_or_log!(
-                        "DroppedId for ID {id} but we have neither ingestion nor export \
-                         under that ID"
-                    );
+
+        // Take the allocation so that we can call mut receiver functions in the loop. We put it
+        // back at the end of the loop.
+        let mut stashed_responses = std::mem::take(&mut self.stashed_responses);
```
teskje, line 1896: Maybe we shouldn't? If we drop the allocation instead, we don't have to worry about it ever getting large. Keeping it lets us safe some allocation overhead, but is that really relevant in this code?

src/storage-controller/src/lib.rs
```diff
@@ -1886,16 +1886,17 @@ where
...
         // back at the end of the loop.
         let mut stashed_responses = std::mem::take(&mut self.stashed_responses);
         for resp in stashed_responses.drain(..) {
             match resp {
-                StorageResponse::FrontierUppers(updates) => {
-                    self.update_write_frontiers(&updates);
-                    updated_frontiers = Some(Response::FrontierUpdates(updates));
+                StorageResponse::FrontierUpper(id, upper) => {
+                    let update = (id, upper);
+                    self.update_write_frontiers(std::slice::from_ref(&update));
```
teskje, line 1901: Looks like `Controller::update_write_frontiers` is now only called with a single update everywhere, so you could rename it to `update_write_frontier` and remove its loop.

src/storage-client/src/client.rs
```diff
@@ -599,29 +581,31 @@ impl From<StatusUpdate> for AppendOnlyUpdate {
...
+    FrontierUpper(GlobalId, Antichain<T>),
     /// Punctuation indicates that no more responses will be transmitted for the specified id
     DroppedId(GlobalId),
     /// Batches that have been staged in Persist and maybe will be linked into a shard.
     StagedBatches(BTreeMap<uuid::Uuid, Vec<Result<ProtoBatch, String>>>),
-
     /// A list of statistics updates, currently only for sources.
     StatisticsUpdates(Vec<SourceStatisticsUpdate>, Vec<SinkStatisticsUpdate>),
-    /// A list of status updates for sources and sinks. Periodically sent from
+    /// A status updates for a source or a sink. Periodically sent from
```
teskje, line 592:
```suggestion
    /// A status update for a source or a sink. Periodically sent from
```

src/storage-client/src/client.rs
```diff
@@ -146,7 +145,7 @@ impl<T> StorageCommand<T> {
             | InitializationComplete
             | AllowWrites
             | UpdateConfiguration(_)
-            | AllowCompaction(_) => false,
+            | AllowCompaction(_, _) => false,
```
teskje, line 148: Did you know you can use `..` on tuple variants too? Not important at all, but I found this neat when I learned about it :)

teskje (approved):
LGTM!

src/storage-controller/src/lib.rs
petrosagg: >  but is that really relevant in this code?

It's not, you're right. Switching to the simpler version

src/storage-client/src/client.rs
petrosagg: I did not! Very useful tip, thanks

PR #31275, opened 2025-02-03
Fueled as_specific_collection

antiguru:
Adds fueling to `as_specific_collection` by calling into `flat_map` instead
of `as_collection`. Add a flag to restore the previous behavior.

### Checklist

### DIFF
--- a/src/compute/src/render/context.rs
+++ b/src/compute/src/render/context.rs
@@ -11,7 +11,7 @@
 //! dataflow.
 
 use std::collections::BTreeMap;
-use std::rc::Weak;
+use std::rc::{Rc, Weak};
 use std::sync::mpsc;
 
 use columnar::Columnar;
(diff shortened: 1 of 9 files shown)


src/compute/src/render/context.rs
```diff
@@ -97,6 +98,24 @@ where
     /// The expiration time for dataflows in this context. The output's frontier should never advance
     /// past this frontier, except the empty frontier.
     pub dataflow_expiration: Antichain<T>,
+    /// The flags active in this context.
+    pub flags: ContextFlags,
```
teskje, line 102: Have you considered just passing a `ConfigSet` directly? That would reduce some boilerplate when adding a new config. You could simply access it everywhere that has access to a `Context`, without needing to change the fields and `From` implement of `ContextFlags`. Also makes it possible to dynamically change some configuration at runtime (without recreating the dataflow), should we ever want that.

teskje (approved):
LGTM. Thanks for adding the feature flag!

src/compute/src/render/context.rs
antiguru: Great point! Changed it such that we just have a single `Rc<ConfigSet>` (updating doesn't require mutability), which seems better than introducing a custom struct.

PR #31280, opened 2025-02-03
mysql: add connection timeout

martykulma:
Wraps `mysql_async::conn::Conn::new()` in a timeout, which is 60s by default, and can be overridden via `mysql_source_connect_timeout` configuration.

## default
```
% time psql postgres://materialize@localhost:6875/materialize -c "create source mysrc from mysql connection mycon"
NOTICE:  connected to Materialize v0.133.0-dev.0
  Org ID: e891b6ab-fd70-4db8-bcc7-3b81c86c9775
  Region: local/az1
  User: materialize
  Cluster: quickstart
  Database: materialize
  Schema: public
  Session UUID: b93976f5-bb9c-47e4-8141-337de9aeaddc

Issue a SQL query to get started. Need help?
  View documentation: https://materialize.com/s/docs
  Join our Slack community: https://materialize.com/s/chat
    
ERROR:  failed to connect to MySQL database: connection attempt timed out after 60s
psql postgres://materialize@localhost:6875/materialize -c   0.01s user 0.01s system 0% cpu 1:00.04 total
```

## override
```
materialize=> alter system set mysql_source_connect_timeout = '7s';
NOTICE:  variable "mysql_source_connect_timeout" was updated for the system, this will have no effect on the current session
ALTER SYSTEM
```

```
% time psql postgres://materialize@localhost:6875/materialize -c "create source mysrc from mysql connection mycon"
NOTICE:  connected to Materialize v0.133.0-dev.0
  Org ID: e891b6ab-fd70-4db8-bcc7-3b81c86c9775
  Region: local/az1
  User: materialize
  Cluster: quickstart
  Database: materialize
  Schema: public
  Session UUID: 6733c6ac-c1ea-4c01-bfd8-ffbcfb90e143

Issue a SQL query to get started. Need help?
  View documentation: https://materialize.com/s/docs
  Join our Slack community: https://materialize.com/s/chat
    
ERROR:  failed to connect to MySQL database: connection attempt timed out after 7s
psql postgres://materialize@localhost:6875/materialize -c   0.01s user 0.01s system 0% cpu 7.038 total
```

### Motivation

This PR adds mysql connection timeout that (somewhat) matches up with the postgresql client
- https://github.com/MaterializeInc/database-issues/issues/8295

### Checklist

### DIFF
--- a/src/mysql-util/src/tunnel.rs
+++ b/src/mysql-util/src/tunnel.rs
@@ -14,8 +14,9 @@ use std::time::Duration;
 
 use mysql_async::{Conn, Opts, OptsBuilder};
 
-use mz_ore::future::{InTask, OreFutureExt, TimeoutError};
+use mz_ore::future::{InTask, TimeoutError};
 use mz_ore::option::OptionExt;
+use mz_ore::task::{spawn, JoinHandleExt};
 use mz_repr::CatalogItemId;
 use mz_ssh_util::tunnel::{SshTimeoutConfig, SshTunnelConfig};
 use mz_ssh_util::tunnel_manager::{ManagedSshTunnelHandle, SshTunnelManager};
@@ -352,8 +353,13 @@ impl Config {
         &self,
         opts_builder: OptsBuilder,
     ) -> Result<mysql_async::Conn, MySqlError> {
-        let connection_future =
-            Conn::new(opts_builder).run_in_task_if(self.in_task, || "mysql_connect".to_string());
+        let connection_future = if let InTask::Yes = self.in_task {
+            spawn(|| "mysql_connect".to_string(), Conn::new(opts_builder))
+                .abort_on_drop()
+                .wait_and_assert_finished()
+        } else {
+            Conn::new(opts_builder)
+        };
 
         if let Some(connect_timeout) = self.mysql_timeout_config.connect_timeout {
             mz_ore::future::timeout(connect_timeout, connection_future)


def- (approved):
Thanks for adding a test! I added another one with a toxiproxy latency to verify that you can still connect with a manually set mysql_source_connect_timeout.

src/mysql-util/src/tunnel.rs
```diff
@@ -318,13 +341,30 @@ impl Config {
...
+
+    async fn connect_with_timeout(
+        &self,
+        opts_builder: OptsBuilder,
+    ) -> Result<mysql_async::Conn, MySqlError> {
+        let connection_future =
+            Conn::new(opts_builder).run_in_task_if(self.in_task, || "mysql_connect".to_string());
+
+        if let Some(connect_timeout) = self.mysql_timeout_config.connect_timeout {
+            mz_ore::future::timeout(connect_timeout, connection_future)
```
petrosagg, line 359: There is potentially an unintended interaction here. When `self.in_task` is false then the code works as intended and the connection will be aborted after the timeout is reached. When `self.in_task` is true then the Future you get back is actually just the `JoinHandle` of the background tokio task. However, dropping that `JoinHandle` doesn't cancel the task. This can be done by calling `JoinHandle::abort`, but that type has been erased at this point so we can't call it.

So the bad thing that can happen is that we set a short timeout that leads to dataflow restarts and we end up accumulating background tasks that will linger until the underlying Future completes or errors.

I think what you want to do here is match on `self.in_task` and use the lower level methods so that you get access to the `JoinHandle`. Then you can call [`abort_on_drop`](https://dev.materialize.com/api/rust-private/mz_ore/task/struct.JoinHandle.html#method.abort_on_drop) before passing it to the timeout function to properly clean up if the timeout fires.

petrosagg (approved):
Looks great, thank you!

PR #31281, opened 2025-02-03
Join order improvement: tweak key length vs. already arranged

ggevay:
This PR changes the join ordering heuristics as discussed here:
https://materializeinc.slack.com/archives/C08ACQNGSQK/p1738616351815239?thread_ts=1738605053.141629&cid=C08ACQNGSQK

The changes are behind a feature flag. I'll create an LD flag and put it in the off state before merging this PR to disable it in production for now, and then we can decide when to roll it out. However, in the code, I've set the flag to be on by default, so that slts show the enabled state, because we expect the flag to be soon turned on in production. (I also ran a full slt rewrite locally with the flag disabled in the code by default, which showed that there wouldn't be any plan changes if the flag is off.)

### Motivation

See [slack discussion](https://materializeinc.slack.com/archives/C08ACQNGSQK/p1738616351815239?thread_ts=1738605053.141629&cid=C08ACQNGSQK). Some joins will need less new arrangements, due to trying harder to use existing arrangements, thus saving memory and CPU.

### Tips for reviewer

The first commit just adds the feature flag (which is a lot of boilerplate, so best reviewed separately), and the second commit adds the actual functionality.

### Checklist

### DIFF
--- a/src/expr-test-util/tests/testdata/rel
+++ b/src/expr-test-util/tests/testdata/rel
@@ -104,23 +104,24 @@ cat
 ----
 ok
 
-build
-(join
-    [(get y) (get y)]
-    [[#0 #3]]
-    (delta_query [[[0 [#0] (false 1 true 1 (false false false 0 false) 0)] [1 [#0] (false 1 true 1 (false false false 0 false) 0)]]
-                  [[1 [#0] (false 1 true 1 (false false false 0 false) 0)] [0 [#0] (false 1 true 1 (false false false 0 false) 0)]]]))
-----
-----
-Join on=(#0 = #3) type=delta
-  implementation
-    %0:u1 » %0:u1[#0]KA|1| » %1:u1[#0]KA|1|
-    %1:u1 » %1:u1[#0]KA|1| » %0:u1[#0]KA|1|
-  Get u1
-  Get u1
-
-----
-----
+# Commented out when JoinInputCharacteristics got versioned.
+#build
+#(join
+#    [(get y) (get y)]
+#    [[#0 #3]]
+#    (delta_query [[[0 [#0] (false 1 true 1 (false false false 0 false) 0)] [1 [#0] (false 1 true 1 (false false false 0 false) 0)]]
+#                  [[1 [#0] (false 1 true 1 (false false false 0 false) 0)] [0 [#0] (false 1 true 1 (false false false 0 false) 0)]]]))
+#----
+#----
+#Join on=(#0 = #3) type=delta
+#  implementation
+#    %0:u1 » %0:u1[#0]KA|1| » %1:u1[#0]KA|1|
+#    %1:u1 » %1:u1[#0]KA|1| » %0:u1[#0]KA|1|
+#  Get u1
+#  Get u1
+#
+#----
+#----
 
 build
 (top_k (get y) [1] [0] 5 1)
--- a/src/expr/src/relation.rs
+++ b/src/expr/src/relation.rs
@@ -3428,13 +3428,154 @@ impl JoinImplementation {
 ///
 /// A candidate is described by a collection and a key, and may have various liabilities.
 /// Primarily, the candidate may risk substantial inflation of records, which is something
-/// that concerns us greatly. Additionally the candidate may be unarranged, and we would
+/// that concerns us greatly. Additionally, the candidate may be unarranged, and we would
 /// prefer candidates that do not require additional memory. Finally, we prefer lower id
-/// collections in the interest of consistent tie-breaking.
+/// collections in the interest of consistent tie-breaking. For more characteristics, see
+/// comments on individual fields.
+///
+/// This has more than one version. `new` instantiates the appropriate version based on a
+/// feature flag.
 #[derive(
     Eq, PartialEq, Ord, PartialOrd, Debug, Clone, Serialize, Deserialize, Hash, MzReflect, Arbitrary,
 )]
-pub struct JoinInputCharacteristics {
+pub enum JoinInputCharacteristics {
+    /// Old version, with `enable_join_prioritize_arranged` turned off.
+    V1(JoinInputCharacteristicsV1),
+    /// Newer version, with `enable_join_prioritize_arranged` turned on.
+    V2(JoinInputCharacteristicsV2),
+}
+
+impl JoinInputCharacteristics {
+    /// Creates a new instance with the given characteristics.
+    pub fn new(
+        unique_key: bool,
+        key_length: usize,
+        arranged: bool,
+        cardinality: Option<usize>,
+        filters: FilterCharacteristics,
+        input: usize,
+        enable_join_prioritize_arranged: bool,
+    ) -> Self {
+        if enable_join_prioritize_arranged {
+            Self::V2(JoinInputCharacteristicsV2::new(
+                unique_key,
+                key_length,
+                arranged,
+                cardinality,
+                filters,
+                input,
+            ))
+        } else {
+            Self::V1(JoinInputCharacteristicsV1::new(
+                unique_key,
+                key_length,
+                arranged,
+                cardinality,
+                filters,
+                input,
+            ))
+        }
+    }
+
+    /// Turns the instance into a String to be printed in EXPLAIN.
+    pub fn explain(&self) -> String {
+        match self {
+            Self::V1(jic) => jic.explain(),
+            Self::V2(jic) => jic.explain(),
+        }
+    }
+
+    /// Whether the join input described by `self` is arranged.
+    pub fn arranged(&self) -> bool {
+        match self {
+            Self::V1(jic) => jic.arranged,
+            Self::V2(jic) => jic.arranged,
+        }
+    }
+
+    /// Returns the `FilterCharacteristics` for the join input described by `self`.
+    pub fn filters(&mut self) -> &mut FilterCharacteristics {
+        match self {
+            Self::V1(jic) => &mut jic.filters,
+            Self::V2(jic) => &mut jic.filters,
+        }
+    }
+}
+
+/// Newer version of `JoinInputCharacteristics`, with `enable_join_prioritize_arranged` turned on.
+#[derive(
+    Eq, PartialEq, Ord, PartialOrd, Debug, Clone, Serialize, Deserialize, Hash, MzReflect, Arbitrary,
+)]
+pub struct JoinInputCharacteristicsV2 {
+    /// An excellent indication that record count will not increase.
+    pub unique_key: bool,
+    /// Cross joins are bad.
+    /// (`key_length > 0` also implies that it is not a cross join. However, we need to note cross
+    /// joins in a separate field, because not being a cross join is more important than `arranged`,
+    /// but otherwise `key_length` is less important than `arranged`.)
+    pub not_cross: bool,
+    /// Indicates that there will be no additional in-memory footprint.
+    pub arranged: bool,
+    /// A weaker signal that record count will not increase.
+    pub key_length: usize,
+    /// Estimated cardinality (lower is better)
+    pub cardinality: Option<std::cmp::Reverse<usize>>,
+    /// Characteristics of the filter that is applied at this input.
+    pub filters: FilterCharacteristics,
+    /// We want to prefer input earlier in the input list, for stability of ordering.
+    pub input: std::cmp::Reverse<usize>,
+}
+
+impl JoinInputCharacteristicsV2 {
+    /// Creates a new instance with the given characteristics.
+    pub fn new(
+        unique_key: bool,
+        key_length: usize,
+        arranged: bool,
+        cardinality: Option<usize>,
+        filters: FilterCharacteristics,
+        input: usize,
+    ) -> Self {
+        Self {
+            unique_key,
+            not_cross: key_length > 0,
+            arranged,
+            key_length,
+            cardinality: cardinality.map(std::cmp::Reverse),
+            filters,
+            input: std::cmp::Reverse(input),
+        }
+    }
+
+    /// Turns the instance into a String to be printed in EXPLAIN.
+    pub fn explain(&self) -> String {
+        let mut e = "".to_owned();
+        if self.unique_key {
+            e.push_str("U");
+        }
+        // Don't need to print `not_cross`, because that is visible in the printed key.
+        // if !self.not_cross {
+        //     e.push_str("C");
+        // }
+        for _ in 0..self.key_length {
+            e.push_str("K");
+        }
+        if self.arranged {
+            e.push_str("A");
+        }
+        if let Some(std::cmp::Reverse(cardinality)) = self.cardinality {
+            e.push_str(&format!("|{cardinality}|"));
+        }
+        e.push_str(&self.filters.explain());
+        e
+    }
+}
+
+/// Old version of `JoinInputCharacteristics`, with `enable_join_prioritize_arranged` turned off.
+#[derive(
+    Eq, PartialEq, Ord, PartialOrd, Debug, Clone, Serialize, Deserialize, Hash, MzReflect, Arbitrary,
+)]
+pub struct JoinInputCharacteristicsV1 {
     /// An excellent indication that record count will not increase.
     pub unique_key: bool,
     /// A weaker signal that record count will not increase.
@@ -3449,7 +3590,7 @@ pub struct JoinInputCharacteristics {
     pub input: std::cmp::Reverse<usize>,
 }
 
-impl JoinInputCharacteristics {
+impl JoinInputCharacteristicsV1 {
     /// Creates a new instance with the given characteristics.
     pub fn new(
         unique_key: bool,
--- a/src/transform/src/join_implementation.rs
+++ b/src/transform/src/join_implementation.rs
@@ -381,6 +381,7 @@ impl JoinImplementation {
                     &unique_keys,
                     &cardinalities,
                     &filters,
+                    features,
                 ) {
                     tracing::debug!(plan = ?delta_query_plan, "replacing differential join with delta join");
                     *relation = delta_query_plan;
@@ -399,6 +400,7 @@ impl JoinImplementation {
                 &unique_keys,
                 &cardinalities,
                 &filters,
+                features,
             )
             .expect("Failed to produce a differential join plan");
 
@@ -463,6 +465,7 @@ impl JoinImplementation {
                 &unique_keys,
                 &cardinalities,
                 &filters,
+                features,
             ) {
                 // If delta plan's inputs need no new arrangements, pick the delta plan.
                 Ok((delta_query_plan, 0)) => {
@@ -577,6 +580,7 @@ mod delta_queries {
     use mz_expr::{
         FilterCharacteristics, JoinImplementation, JoinInputMapper, MirRelationExpr, MirScalarExpr,
     };
+    use mz_repr::optimize::OptimizerFeatures;
 
     use crate::TransformError;
 
@@ -591,6 +595,7 @@ mod delta_queries {
         unique_keys: &[Vec<Vec<usize>>],
         cardinalities: &[Option<usize>],
         filters: &[FilterCharacteristics],
+        optimizer_features: &OptimizerFeatures,
     ) -> Result<(MirRelationExpr, usize), TransformError> {
         let mut new_join = join.clone();
 
@@ -608,23 +613,23 @@ mod delta_queries {
                 cardinalities,
                 filters,
                 input_mapper,
+                optimizer_features.enable_join_prioritize_arranged,
             )?;
 
             // Count new arrangements.
-            let new_arrangements: usize =
-                orders
-                    .iter()
-                    .flat_map(|o| {
-                        o.iter().skip(1).filter_map(|(c, key, input)| {
-                            if c.arranged {
-                                None
-                            } else {
-                                Some((input, key))
-                            }
-                        })
+            let new_arrangements: usize = orders
+                .iter()
+                .flat_map(|o| {
+                    o.iter().skip(1).filter_map(|(c, key, input)| {
+                        if c.arranged() {
+                            None
+                        } else {
+                            Some((input, key))
+                        }
                     })
-                    .collect::<BTreeSet<_>>()
-                    .len();
+                })
+                .collect::<BTreeSet<_>>()
+                .len();
 
             // Convert the order information into specific (input, key, characteristics) information.
             let mut orders = orders
@@ -666,8 +671,9 @@ mod differential {
 
     use mz_expr::{JoinImplementation, JoinInputMapper, MirRelationExpr, MirScalarExpr};
     use mz_ore::soft_assert_eq_or_log;
+    use mz_repr::optimize::OptimizerFeatures;
 
-    use crate::join_implementation::{FilterCharacteristics, JoinInputCharacteristics};
+    use crate::join_implementation::FilterCharacteristics;
     use crate::TransformError;
 
     /// Creates a linear differential plan, and any predicates that need to be lifted.
@@ -679,6 +685,7 @@ mod differential {
         unique_keys: &[Vec<Vec<usize>>],
         cardinalities: &[Option<usize>],
         filters: &[FilterCharacteristics],
+        optimizer_features: &OptimizerFeatures,
     ) -> Result<(MirRelationExpr, usize), TransformError> {
         let mut new_join = join.clone();
 
@@ -704,6 +711,7 @@ mod differential {
                 cardinalities,
                 filters,
                 input_mapper,
+                optimizer_features.enable_join_prioritize_arranged,
             )?;
 
             // Count new arrangements.
@@ -714,7 +722,7 @@ mod differential {
                 .map(|o| {
                     o.iter()
                         .filter_map(|(c, key, input)| {
-                            if c.arranged {
+                            if c.arranged() {
                                 None
                             } else {
                                 Some((*input, key.clone()))
@@ -732,9 +740,9 @@ mod differential {
             // any input before. For examples, see chbench.slt Query 02 and 11.
             orders.iter_mut().for_each(|order| {
                 let mut sum = FilterCharacteristics::none();
-                for (JoinInputCharacteristics { filters, .. }, _, _) in order {
-                    *filters |= sum;
-                    sum = filters.clone();
+                for (jic, _, _) in order {
+                    *jic.filters() |= sum;
+                    sum = jic.filters().clone();
                 }
             });
 
@@ -992,6 +1000,7 @@ fn optimize_orders(
     cardinalities: &[Option<usize>],     // cardinalities of input relations
     filters: &[FilterCharacteristics],   // filter characteristics per input
     input_mapper: &JoinInputMapper,      // join helper
+    enable_join_prioritize_arranged: bool,
 ) -> Result<Vec<Vec<(JoinInputCharacteristics, Vec<MirScalarExpr>, usize)>>, TransformError> {
     let mut orderer = Orderer::new(
         equivalences,
@@ -1000,6 +1009,7 @@ fn optimize_orders(
         cardinalities,
         filters,
         input_mapper,
+        enable_join_prioritize_arranged,
     );
     (0..available.len())
         .map(move |i| orderer.optimize_order_for(i))
@@ -1024,6 +1034,8 @@ struct Orderer<'a> {
     arrangement_active: Vec<Vec<usize>>,
     priority_queue:
         std::collections::BinaryHeap<(JoinInputCharacteristics, Vec<MirScalarExpr>, usize)>,
+
+    enable_join_prioritize_arranged: bool,
 }
 
 impl<'a> Orderer<'a> {
@@ -1034,6 +1046,7 @@ impl<'a> Orderer<'a> {
         cardinalities: &'a [Option<usize>],
         filters: &'a [FilterCharacteristics],
         input_mapper: &'a JoinInputMapper,
+        enable_join_prioritize_arranged: bool,
     ) -> Self {
         let inputs = arrangements.len();
         // A map from inputs to the equivalence classes in which they are referenced.
@@ -1078,6 +1091,7 @@ impl<'a> Orderer<'a> {
             equivalences_active,
             arrangement_active,
             priority_queue,
+            enable_join_prioritize_arranged,
         }
     }
 
@@ -1114,6 +1128,7 @@ impl<'a> Orderer<'a> {
                         cardinality,
                         self.filters[input].clone(),
                         input,
+                        self.enable_join_prioritize_arranged,
                     ),
                     vec![],
                     input,
@@ -1127,6 +1142,7 @@ impl<'a> Orderer<'a> {
                         cardinality,
                         self.filters[input].clone(),
                         input,
+                        self.enable_join_prioritize_arranged,
                     ),
                     vec![],
                     input,
@@ -1162,6 +1178,7 @@ impl<'a> Orderer<'a> {
                 self.cardinalities[start],
                 self.filters[start].clone(),
                 start,
+                self.enable_join_prioritize_arranged,
             ),
             vec![],
             start,
@@ -1198,6 +1215,7 @@ impl<'a> Orderer<'a> {
                         cardinality,
                         self.filters[start].clone(),
                         start,
+                        self.enable_join_prioritize_arranged,
                     ),
                     candidate_start_key,
                     start,
@@ -1288,6 +1306,7 @@ impl<'a> Orderer<'a> {
                                                     self.cardinalities[rel],
                                                     self.filters[rel].clone(),
                                                     rel,
+                                                    self.enable_join_prioritize_arranged,
                                                 ),
                                                 key.clone(),
                                                 rel,
@@ -1310,6 +1329,7 @@ impl<'a> Orderer<'a> {
                                         self.cardinalities[rel],
                                         self.filters[rel].clone(),
                                         rel,
+                                        self.enable_join_prioritize_arranged,
                                     ),
                                     self.bound[rel].clone(),
                                     rel,
--- a/test/sqllogictest/chbench.slt
+++ b/test/sqllogictest/chbench.slt
@@ -314,32 +314,30 @@ Explained Query:
             Filter like["EUROP%"](padchar(#1)) // { arity: 3 }
               ReadStorage materialize.public.region // { arity: 3 }
     Return // { arity: 8 }
-      Project (#2, #3, #12, #0, #1, #4, #6, #7) // { arity: 8 }
-        Join on=(#0 = #8 = #15 AND #2 = #10 AND #5 = #11 AND #9 = #16 AND #13 = #14) type=delta // { arity: 17 }
-          implementation
-            %0:item » %5[#0]UKA » %2:stock[#0, #1]KK » %1:supplier[#0]UK » %3:nation[#0]UK » %4:l0[#0]UKlf
-            %1:supplier » %3:nation[#0]UK » %4:l0[#0]UKlf » %2:stock[#2]KA » %5[#0, #1]UKK » %0:item[#0]UKlf
-            %2:stock » %5[#0, #1]UKK » %0:item[#0]UKlf » %1:supplier[#0]UK » %3:nation[#0]UK » %4:l0[#0]UKlf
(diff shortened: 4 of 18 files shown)


test/sqllogictest/advent-of-code/2023/aoc_1217.slt
```diff
@@ -266,7 +266,7 @@ Explained Query:
     cte l4 =
       Reduce aggregates=[min(#0)]
         Project (#2)
-          Join on=(#0 = #3 AND #1 = #4) type=differential
+          Join on=(#0 = #3 AND #1 = #4) type=delta
```
ggevay, line 269: Probably good change: intermediate results don't need to be arranged.

test/sqllogictest/advent-of-code/2023/aoc_1217.slt
```diff
@@ -312,7 +312,7 @@ Explained Query:
       cte l7 =
         Reduce aggregates=[min(#0)]
           Project (#2)
-            Join on=(#0 = #3 AND #1 = #4) type=differential
+            Join on=(#0 = #3 AND #1 = #4) type=delta
```
ggevay, line 315: Probably good change: intermediate results don't need to be arranged.

test/sqllogictest/chbench.slt
```diff
@@ -314,32 +314,30 @@ Explained Query:
...
+            ArrangeBy keys=[[#0], [#3]] // { arity: 6 }
+              Project (#0..=#4, #6) // { arity: 6 }
+                ReadIndex on=supplier fk_supplier_nationkey=[*** full scan ***] // { arity: 7 }
+            ArrangeBy keys=[[#0], [#17]] // { arity: 18 }
+              ReadIndex on=stock fk_stock_item=[delta join lookup] fk_stock_supplier=[delta join lookup] // { arity: 18 }
+            ArrangeBy keys=[[#0], [#2]] // { arity: 3 }
+              Project (#0..=#2) // { arity: 3 }
+                ReadIndex on=nation fk_nation_regionkey=[*** full scan ***] // { arity: 4 }
+            Get l0 // { arity: 1 }
+            ArrangeBy keys=[[#0]] // { arity: 2 }
```
ggevay, line 340: Good change: 3 less arrangements!

test/sqllogictest/chbench.slt
```diff
@@ -508,28 +507,26 @@ ORDER BY revenue DESC
...
           ArrangeBy keys=[[#0..=#2], [#2, #1, #3]] // { arity: 4 }
             Project (#0..=#3) // { arity: 4 }
               Filter (#3) IS NOT NULL AND (date_to_timestamp(#4) >= 2007-01-02 00:00:00) // { arity: 8 }
                 ReadIndex on=order fk_order_customer=[*** full scan ***] // { arity: 8 }
-          ArrangeBy keys=[[#0..=#3], [#2, #1, #0]] // { arity: 5 }
-            Project (#0..=#2, #4, #8) // { arity: 5 }
-              Filter (#4) IS NOT NULL // { arity: 10 }
-                ReadIndex on=orderline fk_orderline_order=[*** full scan ***] // { arity: 10 }
+          ArrangeBy keys=[[#2, #1, #0]] // { arity: 10 }
+            ReadIndex on=orderline fk_orderline_order=[delta join lookup] // { arity: 10 }
```
ggevay, line 529: 2 less arrangements. (1 being reused instead of building 2 new.)

test/sqllogictest/chbench.slt
```diff
@@ -508,28 +507,26 @@ ORDER BY revenue DESC
...
             %2:orderline » %1:order[#0..=#2]UKKKif » %0:customer[#0..=#2]UKKK » %3:stock[#0, #1]UKK » %4:supplier[#0, #1]UKK » %5:nation[#0]UK » %6:region[#0]UKef
-            %3:stock » %4:supplier[#0]UK » %5:nation[#0]UK » %6:region[#0]UKef » %0:customer[#2, #3]KK » %1:order[#2, #1, #3]KKKAif » %2:orderline[#0..=#3]KKKK
+            %3:stock » %4:supplier[#0]UK » %5:nation[#0]UK » %6:region[#0]UKef » %0:customer[#3]KA » %1:order[#2, #1, #3]KKKAif » %2:orderline[#2, #1, #0]KKKA
             %4:supplier » %5:nation[#0]UK » %6:region[#0]UKef » %0:customer[#3]KA » %1:order[#2, #1, #3]KKKAif » %2:orderline[#2, #1, #0]KKKA » %3:stock[#0..=#2]UKKK
             %5:nation » %6:region[#0]UKef » %0:customer[#3]KA » %1:order[#2, #1, #3]KKKAif » %2:orderline[#2, #1, #0]KKKA » %3:stock[#0, #1]UKK » %4:supplier[#0, #1]UKK
             %6:region » %5:nation[#2]KA » %0:customer[#3]KA » %1:order[#2, #1, #3]KKKAif » %2:orderline[#2, #1, #0]KKKA » %3:stock[#0, #1]UKK » %4:supplier[#0, #1]UKK
-          ArrangeBy keys=[[#0..=#2], [#2, #3], [#3]] // { arity: 4 }
+          ArrangeBy keys=[[#0..=#2], [#3]] // { arity: 4 }
```
ggevay, line 520: 1 less arrangement. (Arrangements being built in the new plan are a subset of arrangements in the old plan.)

test/sqllogictest/chbench.slt
```diff
@@ -1145,20 +1142,19 @@ Explained Query:
...
               Get l0 // { arity: 2 }
-          ArrangeBy keys=[[#0]] // { arity: 1 }
-            Filter (#0) IS NOT NULL // { arity: 1 }
-              Reduce aggregates=[max(#0)] // { arity: 1 }
-                Project (#1) // { arity: 1 }
-                  Get l0 // { arity: 2 }
+          ArrangeBy keys=[[]] // { arity: 1 }
+            Reduce aggregates=[max(#0)] // { arity: 1 }
+              Project (#1) // { arity: 1 }
+                Get l0 // { arity: 2 }
```
ggevay, line 1157: Ugh, **regression**! This will skew the join to exactly 1 worker.

We could tweak the heuristic to not consider such a `Reduce` arranged that is a global aggregate. Reusing that arrangement can't be a positive thing, but it can be a negative thing, as in this example.

test/sqllogictest/chbench.slt
```diff
@@ -1527,22 +1523,20 @@ Explained Query:
...
+                %4:nation » %0:supplier[#2]KA » %3:stock[#2]KA » %1:orderline[#4]KA » %2:order[#0..=#2]UKKK
               ArrangeBy keys=[[#0], [#2]] // { arity: 3 }
                 Project (#0, #1, #3) // { arity: 3 }
                   ReadIndex on=supplier fk_supplier_nationkey=[*** full scan ***] // { arity: 7 }
-              ArrangeBy keys=[[#2, #1, #0], [#2, #3]] // { arity: 5 }
-                Project (#0..=#2, #4, #6) // { arity: 5 }
-                  Filter (#4) IS NOT NULL // { arity: 10 }
-                    ReadIndex on=orderline fk_orderline_order=[*** full scan ***] // { arity: 10 }
+              ArrangeBy keys=[[#2, #1, #0], [#4]] // { arity: 10 }
+                ReadIndex on=orderline fk_orderline_order=[delta join lookup] fk_orderline_item=[delta join lookup] // { arity: 10 }
```
ggevay, line 1539: Good: 2 less arrangements. (0 being created instead of 2 being created.)

test/sqllogictest/explain/aggregates.slt
```diff
@@ -305,10 +305,10 @@ Explained Query:
   Return
     Project (#0, #1)
       Join on=(#0 = #2 AND #1 = #3) type=differential
-        ArrangeBy keys=[[#0, #1]]
+        ArrangeBy keys=[[#0]]
           Reduce group_by=[#0] aggregates=[array_agg[order_by=[#0 asc nulls_last]](row(array[#1], #1))]
             Get l0
-        ArrangeBy keys=[[#0, #1]]
+        ArrangeBy keys=[[#0]]
```
ggevay, line 311: Good change: 2 less arrangements getting created.

test/sqllogictest/explain/optimized_plan_as_text.slt
```diff
@@ -776,20 +776,20 @@ WHERE a = c and d = e and b = f
...
-        ArrangeBy keys=[[#0{c}]]
-          ReadIndex on=u u_c_idx=[differential join]
-        ArrangeBy keys=[[#0{e}, #1{f}]]
-          Filter (#0{e}) IS NOT NULL AND (#1{f}) IS NOT NULL
-            ReadIndex on=v v_e_idx=[*** full scan ***]
+          ReadIndex on=t t_a_idx=[delta join 1st input (full scan)]
+        ArrangeBy keys=[[#0{c}], [#1{d}]]
+          ReadIndex on=u u_c_idx=[delta join lookup] u_d_idx=[delta join lookup]
+        ArrangeBy keys=[[#0{e}]]
+          ReadIndex on=v v_e_idx=[delta join lookup]
```
ggevay, line 786: Probably a good change: 1 less arrangement on inputs because of a reuse + 1 less intermediate arrangement due to switching from differential to delta join.

test/sqllogictest/explain/optimized_plan_as_text.slt
```diff
@@ -804,22 +804,24 @@ WHERE a = c and d = e and b = f
...
```
ggevay, line 818: Good: 2 less arrangements (same as above)

test/sqllogictest/explain/physical_plan_aggregates.slt
```diff
@@ -581,39 +581,30 @@ Explained Query:
...
+          raw=true
+      Reduce::Basic
+        aggr=(0, array_agg[order_by=[#0 desc nulls_first]](row(array[#1], #1)))
+        val_plan
+          project=(#2)
+          map=(row(array[#1], #1))
+        key_plan
+          project=(#0)
+        Get::PassArrangements l0
+          raw=true
```
ggevay, line 607: Good: 2 less arrangements (I think)

test/sqllogictest/explain/physical_plan_as_text.slt
```diff
@@ -1508,70 +1508,71 @@ Explained Query:
...
-        key=#0
-        raw=false
-        arrangements[0]={ key=[#0], permutation=id, thinning=(#1) }
-        types=[integer?, integer?]
+      arrangements[1]={ key=[#1], permutation={#0: #1, #1: #0}, thinning=(#0) }
+      types=[integer?, integer?]
+    Get::PassArrangements materialize.public.v
+      raw=false
+      arrangements[0]={ key=[#0], permutation=id, thinning=(#1) }
+      types=[integer?, integer?]
```
ggevay, line 1569: Good: 2 less arrangements on inputs due to reuse.

test/sqllogictest/ldbc_bi.slt
```diff
@@ -2537,23 +2537,20 @@ Explained Query:
...
-                    Project (#1{messageid}, #9{creatorpersonid}, #10{containerforumid}, #12{parentmessageid}) // { arity: 4 }
-                      ReadIndex on=message message_messageid=[*** full scan ***] // { arity: 13 }
-                  ArrangeBy keys=[[#0{creatorpersonid}, #2{parentmessageid}], [#1{containerforumid}]] // { arity: 3 }
-                    Project (#9{creatorpersonid}, #10{containerforumid}, #12{parentmessageid}) // { arity: 3 }
-                      Filter (#12{parentmessageid}) IS NOT NULL // { arity: 13 }
-                        ReadIndex on=message message_messageid=[*** full scan ***] // { arity: 13 }
+                  ArrangeBy keys=[[#9{creatorpersonid}], [#10{containerforumid}]] // { arity: 13 }
+                    ReadIndex on=message message_containerforumid=[delta join lookup] message_creatorpersonid=[delta join lookup] // { arity: 13 }
+                  ArrangeBy keys=[[#10{containerforumid}], [#12{parentmessageid}]] // { arity: 13 }
+                    ReadIndex on=message message_containerforumid=[delta join lookup] message_parentmessageid=[delta join lookup] // { arity: 13 }
```
ggevay, line 2553: Good: Whoa, 5 less arrangements!

test/sqllogictest/ldbc_bi.slt
```diff
@@ -2733,23 +2732,20 @@ Explained Query:
...
-                  Project (#1{messageid}, #9{creatorpersonid}, #10{containerforumid}, #12{parentmessageid}) // { arity: 4 }
-                    ReadIndex on=message message_messageid=[*** full scan ***] // { arity: 13 }
-                ArrangeBy keys=[[#0{creatorpersonid}, #2{parentmessageid}], [#1{containerforumid}]] // { arity: 3 }
-                  Project (#9{creatorpersonid}, #10{containerforumid}, #12{parentmessageid}) // { arity: 3 }
-                    Filter (#12{parentmessageid}) IS NOT NULL // { arity: 13 }
-                      ReadIndex on=message message_messageid=[*** full scan ***] // { arity: 13 }
+                ArrangeBy keys=[[#9{creatorpersonid}], [#10{containerforumid}]] // { arity: 13 }
+                  ReadIndex on=message message_containerforumid=[delta join lookup] message_creatorpersonid=[delta join lookup] // { arity: 13 }
+                ArrangeBy keys=[[#10{containerforumid}], [#12{parentmessageid}]] // { arity: 13 }
+                  ReadIndex on=message message_containerforumid=[delta join lookup] message_parentmessageid=[delta join lookup] // { arity: 13 }
```
ggevay, line 2748: (Good: same as above)

test/sqllogictest/ldbc_bi_eager.slt
```diff
@@ -2544,23 +2544,20 @@ Explained Query:
...
-                    Project (#1{messageid}, #9{creatorpersonid}, #10{containerforumid}, #12{parentmessageid}) // { arity: 4 }
-                      ReadIndex on=message message_messageid=[*** full scan ***] // { arity: 13 }
-                  ArrangeBy keys=[[#0{creatorpersonid}, #2{parentmessageid}], [#1{containerforumid}]] // { arity: 3 }
-                    Project (#9{creatorpersonid}, #10{containerforumid}, #12{parentmessageid}) // { arity: 3 }
-                      Filter (#12{parentmessageid}) IS NOT NULL // { arity: 13 }
-                        ReadIndex on=message message_messageid=[*** full scan ***] // { arity: 13 }
+                  ArrangeBy keys=[[#9{creatorpersonid}], [#10{containerforumid}]] // { arity: 13 }
+                    ReadIndex on=message message_containerforumid=[delta join lookup] message_creatorpersonid=[delta join lookup] // { arity: 13 }
+                  ArrangeBy keys=[[#10{containerforumid}], [#12{parentmessageid}]] // { arity: 13 }
+                    ReadIndex on=message message_containerforumid=[delta join lookup] message_parentmessageid=[delta join lookup] // { arity: 13 }
```
ggevay, line 2560: (Good: same as above)

test/sqllogictest/ldbc_bi_eager.slt
```diff
@@ -2740,23 +2739,20 @@ Explained Query:
...
-                  Project (#1{messageid}, #9{creatorpersonid}, #10{containerforumid}, #12{parentmessageid}) // { arity: 4 }
-                    ReadIndex on=message message_messageid=[*** full scan ***] // { arity: 13 }
-                ArrangeBy keys=[[#0{creatorpersonid}, #2{parentmessageid}], [#1{containerforumid}]] // { arity: 3 }
-                  Project (#9{creatorpersonid}, #10{containerforumid}, #12{parentmessageid}) // { arity: 3 }
-                    Filter (#12{parentmessageid}) IS NOT NULL // { arity: 13 }
-                      ReadIndex on=message message_messageid=[*** full scan ***] // { arity: 13 }
+                ArrangeBy keys=[[#9{creatorpersonid}], [#10{containerforumid}]] // { arity: 13 }
+                  ReadIndex on=message message_containerforumid=[delta join lookup] message_creatorpersonid=[delta join lookup] // { arity: 13 }
+                ArrangeBy keys=[[#10{containerforumid}], [#12{parentmessageid}]] // { arity: 13 }
+                  ReadIndex on=message message_containerforumid=[delta join lookup] message_parentmessageid=[delta join lookup] // { arity: 13 }
```
ggevay, line 2755: (Good: same as above)

test/sqllogictest/tpch_create_index.slt
```diff
@@ -501,27 +501,24 @@ materialize.public.q05_primary_idx:
...
-            Project (#0{l_orderkey}, #2{l_suppkey}, #5{l_extendedprice}, #6{l_discount}) // { arity: 4 }
-              ReadIndex on=lineitem pk_lineitem_orderkey_linenumber=[*** full scan ***] // { arity: 16 }
-          ArrangeBy keys=[[#0{s_suppkey}, #1{s_nationkey}]] // { arity: 2 }
-            Project (#0{s_suppkey}, #3{s_nationkey}) // { arity: 2 }
-              Filter (#0{s_suppkey}) IS NOT NULL // { arity: 7 }
-                ReadIndex on=supplier pk_supplier_suppkey=[*** full scan ***] // { arity: 7 }
+          ArrangeBy keys=[[#0{l_orderkey}], [#2{l_suppkey}]] // { arity: 16 }
+            ReadIndex on=lineitem fk_lineitem_orderkey=[delta join lookup] fk_lineitem_suppkey=[delta join lookup] // { arity: 16 }
+          ArrangeBy keys=[[#3{s_nationkey}]] // { arity: 7 }
+            ReadIndex on=supplier fk_supplier_nationkey=[delta join lookup] // { arity: 7 }
```
ggevay, line 521: Good: 3 less arrangements. (This is the query that originally prompted this investigation.)

test/sqllogictest/tpch_create_index.slt
```diff
@@ -1303,14 +1301,14 @@ materialize.public.q15:
...
-          %1:l0 » %2[#0]UK » %0:supplier[#0]KA
+          %0:supplier » %1:l0[#0]UKA » %2[×]UA
+          %1:l0 » %2[×]UA » %0:supplier[#0]KA
           %2 » %1:l0[#1]K » %0:supplier[#0]KA
         ArrangeBy keys=[[#0{s_suppkey}]] // { arity: 7 }
           ReadIndex on=supplier pk_supplier_suppkey=[delta join 1st input (full scan)] // { arity: 7 }
         ArrangeBy keys=[[#0{l_suppkey}], [#1{sum}]] // { arity: 2 }
           Get l0 // { arity: 2 }
-        ArrangeBy keys=[[#0{max_sum}]] // { arity: 1 }
+        ArrangeBy keys=[[]] // { arity: 1 }
```
ggevay, line 1311: **Regression**! Same issue [as the other regression above](https://github.com/MaterializeInc/materialize/pull/31281/files#r1941037320), and the same fix would work.

test/sqllogictest/tpch_create_materialized_view.slt
```diff
@@ -453,27 +453,24 @@ ORDER BY
...
-            Project (#0{l_orderkey}, #2{l_suppkey}, #5{l_extendedprice}, #6{l_discount}) // { arity: 4 }
-              ReadIndex on=lineitem pk_lineitem_orderkey_linenumber=[*** full scan ***] // { arity: 16 }
-          ArrangeBy keys=[[#0{s_suppkey}, #1{s_nationkey}]] // { arity: 2 }
-            Project (#0{s_suppkey}, #3{s_nationkey}) // { arity: 2 }
-              Filter (#0{s_suppkey}) IS NOT NULL // { arity: 7 }
-                ReadIndex on=supplier pk_supplier_suppkey=[*** full scan ***] // { arity: 7 }
+          ArrangeBy keys=[[#0{l_orderkey}], [#2{l_suppkey}]] // { arity: 16 }
+            ReadIndex on=lineitem fk_lineitem_orderkey=[delta join lookup] fk_lineitem_suppkey=[delta join lookup] // { arity: 16 }
+          ArrangeBy keys=[[#3{s_nationkey}]] // { arity: 7 }
+            ReadIndex on=supplier fk_supplier_nationkey=[delta join lookup] // { arity: 7 }
```
ggevay, line 473: (Good: same query as above)

test/sqllogictest/tpch_create_materialized_view.slt
```diff
@@ -1165,14 +1163,14 @@ materialize.public.q15:
...
-          %1:l0 » %2[#0]UK » %0:supplier[#0]KA
+          %0:supplier » %1:l0[#0]UKA » %2[×]UA
+          %1:l0 » %2[×]UA » %0:supplier[#0]KA
           %2 » %1:l0[#1]K » %0:supplier[#0]KA
         ArrangeBy keys=[[#0{s_suppkey}]] // { arity: 7 }
           ReadIndex on=supplier pk_supplier_suppkey=[delta join 1st input (full scan)] // { arity: 7 }
         ArrangeBy keys=[[#0{l_suppkey}], [#1{sum}]] // { arity: 2 }
           Get l0 // { arity: 2 }
-        ArrangeBy keys=[[#0{max_sum}]] // { arity: 1 }
+        ArrangeBy keys=[[]] // { arity: 1 }
```
ggevay, line 1173: (Same query with the same regression as above)

test/sqllogictest/tpch_select.slt
```diff
@@ -452,27 +452,24 @@ ORDER BY
...
-              Project (#0{l_orderkey}, #2{l_suppkey}, #5{l_extendedprice}, #6{l_discount}) // { arity: 4 }
-                ReadIndex on=lineitem pk_lineitem_orderkey_linenumber=[*** full scan ***] // { arity: 16 }
-            ArrangeBy keys=[[#0{s_suppkey}, #1{s_nationkey}]] // { arity: 2 }
-              Project (#0{s_suppkey}, #3{s_nationkey}) // { arity: 2 }
-                Filter (#0{s_suppkey}) IS NOT NULL // { arity: 7 }
-                  ReadIndex on=supplier pk_supplier_suppkey=[*** full scan ***] // { arity: 7 }
+            ArrangeBy keys=[[#0{l_orderkey}], [#2{l_suppkey}]] // { arity: 16 }
+              ReadIndex on=lineitem fk_lineitem_orderkey=[delta join lookup] fk_lineitem_suppkey=[delta join lookup] // { arity: 16 }
+            ArrangeBy keys=[[#3{s_nationkey}]] // { arity: 7 }
+              ReadIndex on=supplier fk_supplier_nationkey=[delta join lookup] // { arity: 7 }
```
ggevay, line 472: (same query as above)

test/sqllogictest/tpch_select.slt
```diff
@@ -1162,14 +1160,14 @@ Explained Query:
...
-            %1:l0 » %2[#0]UK » %0:supplier[#0]KA
+            %0:supplier » %1:l0[#0]UKA » %2[×]UA
+            %1:l0 » %2[×]UA » %0:supplier[#0]KA
             %2 » %1:l0[#1]K » %0:supplier[#0]KA
           ArrangeBy keys=[[#0{s_suppkey}]] // { arity: 7 }
             ReadIndex on=supplier pk_supplier_suppkey=[delta join 1st input (full scan)] // { arity: 7 }
           ArrangeBy keys=[[#0{l_suppkey}], [#1{sum}]] // { arity: 2 }
             Get l0 // { arity: 2 }
-          ArrangeBy keys=[[#0{max_sum}]] // { arity: 1 }
+          ArrangeBy keys=[[]] // { arity: 1 }
```
ggevay, line 1170: (same query as above)

test/sqllogictest/transform/aggregation_nullability.slt
```diff
@@ -1056,16 +1056,15 @@ Explained Query:
...
         ArrangeBy keys=[[#0]] // { arity: 1 }
           Project (#1) // { arity: 1 }
             ReadStorage materialize.public.t1 // { arity: 2 }
-        ArrangeBy keys=[[], [#0]] // { arity: 1 }
-          Filter (#0) IS NOT NULL // { arity: 1 }
-            Reduce aggregates=[max(#0)] // { arity: 1 }
-              Get l1 // { arity: 1 }
+        ArrangeBy keys=[[]] // { arity: 1 }
+          Reduce aggregates=[max(#0)] // { arity: 1 }
+            Get l1 // { arity: 1 }
```
ggevay, line 1067: **Regression:** Again, very similar to the above 2 regressions, with the same fix possible.

test/sqllogictest/transform/join_fusion.slt
```diff
@@ -120,21 +120,20 @@ Explained Query:
...
-        ArrangeBy keys=[[#0, date_to_timestamp(#3)]] // { arity: 4 }
-          Filter (#0) IS NOT NULL // { arity: 4 }
-            ReadIndex on=orders pk_orders_orderkey=[*** full scan ***] // { arity: 4 }
+          %0:lineitem » %1:orders[#0]KA » %2:customer[×]
+          %1:orders » %0:lineitem[#0]KA » %2:customer[×]
+          %2:customer » %0:lineitem[×] » %1:orders[#0]KA
+        ArrangeBy keys=[[], [#0]] // { arity: 8 }
+          ReadIndex on=lineitem fk_lineitem_orderkey=[*** full scan ***, delta join lookup] // { arity: 8 }
+        ArrangeBy keys=[[#0]] // { arity: 4 }
+          ReadIndex on=orders pk_orders_orderkey=[delta join lookup] // { arity: 4 }
```
ggevay, line 129: Probably a good change because of 2 less arrangements. But it's hard to tell, because all these cross joins make this plan very sensitive to the join order, depending on data characteristics.

test/sqllogictest/chbench.slt
mgree: This is very interesting---beyond wanting to be able to reach in to `Reduce`'s arrangements to reorder things in join planning, we might need to treat a `Reduce`'s arrangements differently for certain aggregates? (By global, do you mean with an empty key, or something else?)

frankmcsherry: That's my read (on global = unit key). This is an example of where the trade-off "makes sense" to prefer longer keys, of at least a certain complexity. We could .. potentially think specifically about empty keys, which are visibly skewed. I don't want to randomize the PR, but the "key length" as a goal was mostly to attempt some minimal distribution, not necessarily maximal entropy. So, `[]` has problems, but `[expr1]` vs `[expr2, expr3]` is much harder to have a helpful opinion on (because 1. who knows which has more entropy, and 2. once past a certain level, who cares).

test/sqllogictest/tpch_create_index.slt
frankmcsherry: Unhelpful thought, but longer term the pattern of joining against a max can/should be `TopK` supporting ties. I think. At least, that's what you'd want someone to type, you can as of SQL2008, but we don't support it yet.

frankmcsherry:
The changes look qualitatively much better. Although some moments where we have aggregates with `[]` keys, it's hard to feel terrible there in that the queries say as much themselves. It's a very good hint for a potential broadcast join implementation (at least in the case of a unit key that is also a primary key, as many of the cases seemed to be).

test/sqllogictest/chbench.slt
ggevay: Ok, so there are several ways to tweak our heuristics to avoid this type of regression.

Maybe the best way would be to add a new field to `JoinInputCharacteristics` that is just a bool `non_empty_key`, and put this field ahead of `arranged`. So, this way, `non_empty_key` would be more important than `arranged`, but otherwise `arranged` would be more important than `key_length`:

Current `JoinInputCharacteristics`:
```
pub struct JoinInputCharacteristics {
    /// An excellent indication that record count will not increase.
    pub unique_key: bool,
    /// A weaker signal that record count will not increase.
    pub key_length: usize,
    /// Indicates that there will be no additional in-memory footprint.
    pub arranged: bool,
    /// Estimated cardinality (lower is better)
    pub cardinality: Option<std::cmp::Reverse<usize>>,
    /// Characteristics of the filter that is applied at this input.
    pub filters: FilterCharacteristics,
    /// We want to prefer input earlier in the input list, for stability of ordering.
    pub input: std::cmp::Reverse<usize>,
}
```

New `JoinInputCharacteristics` after the PR:
```
pub struct JoinInputCharacteristics {
    /// An excellent indication that record count will not increase.
    pub unique_key: bool,
    
    /// THE NEW FIELD. More important than `arranged`!
    pub non_empty_key: bool,

    /// Indicates that there will be no additional in-memory footprint.
    pub arranged: bool,

    /// A weaker signal that record count will not increase. THIS WAS MOVED DOWN, because it's less important than `arranged`.
    pub key_length: usize,

    /// Estimated cardinality (lower is better)
    pub cardinality: Option<std::cmp::Reverse<usize>>,
    /// Characteristics of the filter that is applied at this input.
    pub filters: FilterCharacteristics,
    /// We want to prefer input earlier in the input list, for stability of ordering.
    pub input: std::cmp::Reverse<usize>,
}
```

mgree: Yes, this makes sense and seems like a gentler tweak. For feature flagging, seems a bit annoying to have to support multiple types here, but it seems equally finicky to sneak the feature flag into the `Ord` instance.

ggevay: Yeah, the feature flag will get a bit ugly, but we'll just have to bite the bullet. I have the following plan: Copy-paste `JoinInputCharacteristics` to have a `JoinInputCharacteristicsOld` and `JoinInputCharacteristicsNew`. Also, have a `JoinInputCharacteristics` that is just an enum that is the old or the new. And then we can pass the feature flag's value to the constructor of `JoinInputCharacteristics`, which will create either the old or the new.

test/sqllogictest/tpch_create_index.slt
ggevay: Yes. (Btw., in the long run, another reason why we'll need a TopK that supports ties is because we'll eventually want to automatically optimize `rank() OVER (...) = 1` to TopK: https://github.com/MaterializeInc/database-issues/issues/5060.)

What do you exactly mean by "you can as of SQL2008"? `max_by`?

test/sqllogictest/chbench.slt
mgree: Nice! That or a trait and `Box<dyn JICImpl>`. Probably simplest to be direct.

antiguru: Here's an alternative: Implement `PartialOrd/Ord` manually, and use a thread-local bool to decide which one sorts first. This only works if we don't keep the struct around for long in a btree or any other structure that expected consistent sort order.

Otherwise, the enum of old and new variant seem like the obvious choice :/

ggevay: Hmm, the join planning does put these into complicated structures, but these shouldn't escape the thread, so this would be a viable alternative.

Still, I think the enum is kind of fine. It's a bit of copy-paste, but not the worst kind of copy-paste, in that we are not expecting to ever modify the `...Old` struct, so we won't be in the situation where we'd need to "maintain" both versions, e.g. having to mirror some later modifications in both versions. Btw. it's somewhat similar to what we have in the audit log with the `...V2`, `...V3`, etc. structs.

I'll think about `Box<dyn JICImpl>`. But I have to admit I developed kind of an allergy to `Box<dyn ...>` by working on data processing stuff in Java/Scala for 6 years, where pretty much every single thing is a `Box<dyn ...>` :D

src/transform/src/join_implementation.rs
```diff
@@ -611,20 +611,19 @@ mod delta_queries {
...
+                            None
+                        } else {
+                            Some((input, key))
+                        }
                     })
-                    .collect::<BTreeSet<_>>()
-                    .len();
+                })
+                .collect::<BTreeSet<_>>()
+                .len();
```
ggevay, line 626: (Just changed `c.arranged` to `c.arranged()`, but `cargo fmt` reformatted the whole block.)

test/sqllogictest/advent-of-code/2023/aoc_1217.slt
ggevay: Well, this somehow reverted to a differential join after I [tweaked the heuristics](https://github.com/MaterializeInc/materialize/pull/31281#discussion_r1943159959).

test/sqllogictest/advent-of-code/2023/aoc_1217.slt
ggevay: Same as above, reverted to a differential join after I [tweaked the heuristics](https://github.com/MaterializeInc/materialize/pull/31281#discussion_r1943159959).

test/sqllogictest/chbench.slt
ggevay: Ok, regression is gone after tweaking the heuristics!

test/sqllogictest/tpch_create_index.slt
ggevay: Ok, regression is gone after tweaking the heuristics!

test/sqllogictest/transform/aggregation_nullability.slt
ggevay: Ok, regression is gone after tweaking the heuristics!

src/expr-test-util/tests/testdata/rel
```diff
@@ -104,23 +104,24 @@ cat
...
-Join on=(#0 = #3) type=delta
-  implementation
-    %0:u1 » %0:u1[#0]KA|1| » %1:u1[#0]KA|1|
-    %1:u1 » %1:u1[#0]KA|1| » %0:u1[#0]KA|1|
-  Get u1
-  Get u1
-
-----
-----
+# Commented out when JoinInputCharacteristics got versioned.
```
mgree, line 107: We should form a plan about how we're going to roll out and then uncomment this once we're on a single version.

src/expr/src/relation.rs
```diff
@@ -3428,13 +3428,154 @@ impl JoinImplementation {
...
-/// collections in the interest of consistent tie-breaking.
+/// collections in the interest of consistent tie-breaking. For more characteristics, see
+/// comments on individual fields.
+///
+/// This has more than one version. `new` instantiates the appropriate version based on a
+/// feature flag.
+#[derive(
+    Eq, PartialEq, Ord, PartialOrd, Debug, Clone, Serialize, Deserialize, Hash, MzReflect, Arbitrary,
+)]
+pub enum JoinInputCharacteristics {
```
mgree, line 3441: Nice solution to an annoying problem. 😁

src/sql/src/plan/statement/ddl.rs
```diff
@@ -4415,7 +4415,8 @@ generate_extracted_config!(
     (EnableEagerDeltaJoins, Option<bool>, Default(None)),
     (EnableNewOuterJoinLowering, Option<bool>, Default(None)),
     (EnableVariadicLeftJoinLowering, Option<bool>, Default(None)),
-    (EnableLetrecFixpointAnalysis, Option<bool>, Default(None))
+    (EnableLetrecFixpointAnalysis, Option<bool>, Default(None)),
+    (EnableJoinPrioritizeArranged, Option<bool>, Default(None))
```
mgree, line 4419:
```suggestion
    (EnableJoinPrioritizeArranged, Option<bool>, Default(None)),
```
if the macro will allow it?

src/transform/src/join_implementation.rs
mgree: 🙃

mgree (approved):
Optimizer changes LGTM!

src/sql/src/plan/statement/ddl.rs
ggevay: It won't, unfortunately... (Btw., some of our similar macros also don't allow having a comma there, while some other of our macros don't allow _not_ having a comma there :sweat_smile: )

src/expr-test-util/tests/testdata/rel
ggevay: Tbh, nowadays I'm getting quite lazy with these `mzreflect`-based tests, because we mostly decided that we'll just remove this whole test framework, and probably work with the `spec` test framework instead. See https://github.com/MaterializeInc/database-issues/issues/8774

ggevay:
Thanks for the review!

I've created the LD flag, and set it to Off in prod, and On in staging.

Merging!

PR #31282, opened 2025-02-04
feat: Add self-managed debug tool

SangJunBak:
Introduces a new CLI tool for debugging self-managed Materialize deployments in Kubernetes. The tool can:
- Collect pod logs
- Collect info similar to `kubectl get all -o wide`
- Up next will be getting events, describe output, rest of TODOs, and tests

I hope to merge all changes in a stack but decided to create a PR for early feedback.  

### Motivation

### Tips for reviewer

- Currently want to leave it all in one file and separate things out later
- Curious if the amount of code for something like this is normal!

### Checklist


src/self-managed-debug/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
```
SangJunBak, line 1: I manually created this file by copying `.toml`s from other projects. Wondering if there's an easier way to do this though via CLI commands.

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,724 @@
...
+            "NetworkPolicies:\n{}\n",
+            Table::new(netpol_info).with(Style::psql())
+        )?;
+    }
+
+    println!("Finished exporting k8s get all for {}", namespace);
+    Ok(file_name)
+}
+
+fn format_duration(duration: Duration) -> String {
```
SangJunBak, line 640: Wondering if there's a better function I can reuse here

src/self-managed-debug/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+name = "mz-self-managed-debug"
+description = "Debug tool for self-managed Materialize."
+version = "0.130.1"
```
antiguru, line 4: You probably want the current version here and elsewhere (right now it's `0.133.0-dev.0`, but you'll want to update it prior to merging.)

Also, to update it automatically, adjust line 41 and 61 in `bin/bump-version` so that the version gets updated automatically. The script takes care of calling `bin/bazel gen` to sync the bazel files.

src/self-managed-debug/Cargo.toml
antiguru: That's what I do. You start with some similar crate and use it as a starting point.

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,722 @@
...
+
+use mz_ore::cli::{self, CliConfig};
+
+use k8s_openapi::api::core::v1::{ContainerStatus, Pod, Service};
+
+use kube::api::{Api, ListParams, LogParams, ObjectMeta};
+use mz_ore::error::ErrorExt;
+
+use std::fs::File;
+use std::io::Write;
```
antiguru, line 37: Nit: Structure uses as:
```
use std::...;

use other-crate::...;

use crate::...;
```
I.e., three sections: std, other crates, `crate`, separated by a newline.

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,722 @@
...
+        };
+    }
+
+    // TODO: Compress files to ZIP
+    Ok(())
+}
+
+/**
+ * Creates a k8s client given a context. If no context is provided, the default context is used.
+ */
```
antiguru, line 103:
```suggestion
/// Creates a k8s client given a context. If no context is provided, the default context is used.
```

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,722 @@
...
+
+    let client = Client::try_from(kubeconfig)?;
+
+    Ok(client)
+}
+
+/**
+ * Write k8s pod logs to a file per pod as mz-pod-logs.<namespace>.<pod-name>.log.
+ * Returns a list of file names on success.
+ */
```
antiguru, line 120:
```suggestion
/// Write k8s pod logs to a file per pod as `mz-pod-logs.<namespace>.<pod-name>.log`.
/// Returns a list of file names on success.
```

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,722 @@
...
+
+            println!("Finished exporting logs for {}", pod_name);
+            file_names.push(file_name);
+            Ok::<(), anyhow::Error>(())
+        }
+        .await
+        {
+            let pod_name = pod.metadata.name.clone().unwrap_or_default();
+            eprintln!("Failed to process pod {}: {}", pod_name, e);
+        }
```
antiguru, line 174: This might not do what you want: the `?` operator returns from the enclosing function, not the block.

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,722 @@
...
+                    age: format_age(&netpol.metadata),
+                }
+            })
+        })
+        .collect();
+    if !netpol_info.is_empty() {
+        writeln!(
+            file,
+            "NetworkPolicies:\n{}\n",
+            Table::new(netpol_info).with(Style::psql())
```
antiguru, line 630: Did you consider writing json (or something else structured) instead of formatting as a table? If it's json, we could always format it later, but the table doesn't allow us easily read the contents in a program.

src/self-managed-debug/src/main.rs
antiguru: You should be able to get a `TimeDelta`, which has a `Display` implementation.

antiguru (commented):
Left some comments inline. Thanks for taking this on!

A high-level question I have is why we don't format the data as JSON (or another serialization framework). This would reduce complexity here as a lot of the types already have serde implementations, and we'd not accidentally lose information on the way.

We could then have a second binary that takes the serialized format and pretty-prints it for human consumption. (Or, send both files because it allows users to more easily verify what information we include.)

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,722 @@
...
+        );
+        process::exit(1);
+    }
+}
+
+async fn run(args: Args) -> Result<(), anyhow::Error> {
+    let client = create_k8s_client(args.kubernetes_context.clone()).await?;
+    // TODO: Make namespaces mandatory
+    // TODO: Print a warning if namespace doesn't exist
+    for namespace in args.kubernetes_namespaces {
```
alex-hunt-materialize, line 73: We probably want to store things in an organized directory structure. Maybe something like:
`materialize-debug-{iso_datetime}/{namespace}/{resource_type}/{resource_name}.yaml`

or, in the case of logs:
`materialize-debug-{iso_datetime}/{namespace}/logs/{pod_name}.{current_or_previous}.log`

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,722 @@
...
+    let mut file_names = Vec::new();
+
+    let pods: Api<Pod> = Api::<Pod>::namespaced(client.clone(), namespace);
+
+    let pod_list = pods.list(&ListParams::default()).await?;
+
+    for pod in &pod_list.items {
+        if let Err(e) = async {
+            let pod_name = pod.metadata.name.clone().unwrap_or_default();
+            let file_name = format!("mz-pod-logs.{}.{}.log", namespace, pod_name);
```
alex-hunt-materialize, line 134: See above about directory structure.

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,722 @@
...
+        type_: String,
+        #[tabled(rename = "CLUSTER-IP")]
+        cluster_ip: String,
+        #[tabled(rename = "EXTERNAL-IP")]
+        external_ip: String,
+        #[tabled(rename = "PORT(S)")]
+        ports: String,
+        #[tabled(rename = "AGE")]
+        age: String,
+    }
```
alex-hunt-materialize, line 217: We probably don't need most of this info, but we may want to know if the service has endpoints (ie: is it ready).

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,722 @@
...
+            })
+        })
+        .collect();
+
+    if !pod_info.is_empty() {
+        writeln!(
+            file,
+            "Pods:\n{}\n",
+            Table::new(pod_info).with(Style::psql())
+        )?;
```
alex-hunt-materialize, line 406: These structs might be useful for printing things out for a human running the tool, but I really just want the yaml serialized object in its entirety.

alex-hunt-materialize (commented):
+1 to everything Moritz said.

src/self-managed-debug/src/main.rs
SangJunBak: That makes sense, even if we're zipping at the end!

SangJunBak:
After talking to the cloud team, decided to remove `dump_k8s_get_all` and can add it back later if we need to

src/self-managed-debug/src/main.rs
SangJunBak: Ah but if we're referring to the usage of `?` in lines 132-168, wouldn't it return from the block given it's an async block?

src/self-managed-debug/Cargo.toml
SangJunBak: Acc ik Marta mentioned we'd most likely want to decouple the tool from LTS versions/materialize   https://materializeinc.slack.com/archives/C07PN7KSB0T/p1737567670003019?thread_ts=1737566344.547089&cid=C07PN7KSB0T, so we might want to make this just version 0.1.0. I think it's very doable that we can have the tool work for any version of materialize.

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,348 @@
...
+        .stderr(std::process::Stdio::null()) // Silence stderr
+        .output()
+        .await?;
+
+    if !output.status.success() {
+        return Err(anyhow::anyhow!(
+            "{}",
+            String::from_utf8_lossy(&output.stderr)
+        ));
+    }
```
SangJunBak, line 319: I noticed when we usually get an auth error here, per resource type, we get an error message like:
```
Failed to dump kubectl describe for deployments: 
The SSO session associated with this profile has expired or is otherwise invalid. To refresh this SSO session run aws sso login with the corresponding profile.
E0210 01:49:49.975938   34995 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"https://26380169F507ACFBF59BE93A7134F6B1.gr7.us-east-1.eks.amazonaws.com/api?timeout=32s\": getting credentials: exec: executable aws failed with exit code 255"

The SSO session associated with this profile has expired or is otherwise invalid. To refresh this SSO session run aws sso login with the corresponding profile.
E0210 01:49:50.491841   34995 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"https://26380169F507ACFBF59BE93A7134F6B1.gr7.us-east-1.eks.amazonaws.com/api?timeout=32s\": getting credentials: exec: executable aws failed with exit code 255"

The SSO session associated with this profile has expired or is otherwise invalid. To refresh this SSO session run aws sso login with the corresponding profile.
E0210 01:49:50.934489   34995 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"https://26380169F507ACFBF59BE93A7134F6B1.gr7.us-east-1.eks.amazonaws.com/api?timeout=32s\": getting credentials: exec: executable aws failed with exit code 255"

The SSO session associated with this profile has expired or is otherwise invalid. To refresh this SSO session run aws sso login with the corresponding profile.
E0210 01:49:51.425720   34995 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"https://26380169F507ACFBF59BE93A7134F6B1.gr7.us-east-1.eks.amazonaws.com/api?timeout=32s\": getting credentials: exec: executable aws failed with exit code 255"

The SSO session associated with this profile has expired or is otherwise invalid. To refresh this SSO session run aws sso login with the corresponding profile.
E0210 01:49:51.899735   34995 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"https://26380169F507ACFBF59BE93A7134F6B1.gr7.us-east-1.eks.amazonaws.com/api?timeout=32s\": getting credentials: exec: executable aws failed with exit code 255"
Unable to connect to the server: getting credentials: exec: executable aws failed with exit code 255
```
wondering if it's worth pointing out, if we should truncate this, or if we should output this to an error log file?

alex-hunt-materialize: That message is useful for debugging permissions in AWS. We should definitely show that to the user.

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,348 @@
...
+
+            file.write_all(logs.as_bytes())?;
+
+            println!("Exported {}", &current_logs_file_name);
+            Ok::<(), anyhow::Error>(())
+        }
+        .await
+        {
+            eprintln!("Failed to export {}: {}", &current_logs_file_name, e);
+        }
```
alex-hunt-materialize, line 261: There is a lot of duplication here. Maybe pull this all out into a separate function that is called for both the previous and current logs?

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,348 @@
...
+            let mut file = File::create(&previous_logs_file_name)?;
+
+            file.write_all(previous_logs.as_bytes())?;
+
+            println!("Exported {}", &previous_logs_file_name);
+            Ok::<(), anyhow::Error>(())
+        }
+        .await
+        {
+            if let Some(kube::Error::Api(e)) = e.downcast_ref::<kube::Error>() {
```
alex-hunt-materialize, line 225: This would probably be nicer with a three branch match statement:
```rust
match result.await {
    Ok(_) => {},
    Err(kube::Error::Api(e)) if e.code == 400 => {
        eprintln!("No previous logs available for pod {}", pod_name);
    },
    Err(e) => {
        eprintln!("Failed to export {}: {}", &previous_logs_file_name, e);
    },
}
```

I'm also surprised it's a 400, rather than a 404. It sucks that they aren't more specific here.

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,348 @@
...
+        ));
+    }
+
+    if output.stdout.is_empty() {
+        println!("No {} found in namespace {}", resource_type, namespace);
+        return Ok(());
+    }
+
+    let file_path = format_resource_path(context.start_time, namespace, resource_type);
+    let file_name = format!("{}/describe.txt", file_path);
```
alex-hunt-materialize, line 327: It's not super important, but for cross-platform stuff, you should probably use PathBuf and Path instead of formatting Strings.

alex-hunt-materialize (commented):
We're still missing the most important thing I want: the object yaml.

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,348 @@
...
+
+    if event_list.items.is_empty() {
+        println!("No events found for namespace {}", namespace);
+        return Ok(());
+    }
+
+    let file_path = format_resource_path(context.start_time, namespace, &K8sResourceType::Event);
+    let file_name = format!("{}/events.yaml", file_path);
+    create_dir_all(&file_path)?;
+    let mut file = File::options().append(true).create(true).open(&file_name)?;
```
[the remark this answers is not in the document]
SangJunBak: Ah at this stage, we assumed there isn't already a file. But I added `append(true)` such that `        serde_yaml::to_writer(&mut file, &event)?;` appends per event, which seems like what we want! Lemme know if there's a better way to achieve what I want tho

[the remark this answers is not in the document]
alex-hunt-materialize: I'm pretty sure the append is just for how the file gets opened. You shouldn't need it for the calls to `to_writer` to append.

src/self-managed-debug/src/main.rs
SangJunBak: Ah so the reason I wrote it this way is the compiler's forcing me to downcast the error to a kube::Error first. I tried running
```
Ok(_) => {},
Err(e)
    if let Some(kube::Error::Api(e)) = e.downcast_ref::<kube::Error>()
        && e.code == 400 =>
{
    eprintln!("No previous logs available for pod {}", pod_name);
}
```
but the compiler gives me the error 
```
error[E0658]: `let` expressions in this position are unstable
   --> src/self-managed-debug/src/main.rs:227:20
    |
227 |                 if let Some(kube::Error::Api(e)) = e.downcast_ref::<kube::Error>()
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667> for more information
```
It's something to do with the `&& e.code ==` being on the same line as the `let Some`. 

Given we don't need the result however, I refactored it to:
```
 if let Err(e) = export_pod_logs(&pods, &pod_name, &file_path, true).await {
            let print_error = || {
                eprintln!(
                    "Failed to export previous logs for pod {}: {}",
                    &pod_name, e
                );
            };

            if let Some(kube::Error::Api(e)) = e.downcast_ref::<kube::Error>() {
                if e.code == 400 {
                    eprintln!("No previous logs available for pod {}", pod_name);
                } else {
                    print_error();
                }
            } else {
                print_error();
            }
        }
```

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,348 @@
...
+            for namespace in context.args.k8s_namespaces.clone() {
+                let _ = match dump_k8s_pod_logs(&context, client.clone(), &namespace).await {
+                    Ok(_) => Some(()),
+                    Err(e) => {
+                        eprintln!(
+                            "Failed to write k8s pod logs for namespace {}: {}",
+                            namespace, e
+                        );
+                        if let Some(_) = e.downcast_ref::<kube::Error>() {
+                            eprintln!("Ensure that {} exists.", namespace);
```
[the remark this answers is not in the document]
SangJunBak: Decided to get rid of this given the error we print above is informative enough!

SangJunBak:
Btw @alex-hunt-materialize the full list of resource types I could get are:
```
➜  materialize git:(jun/#8908/debug-tool) ✗ kubectl api-resources                                                                                           
NAME                              SHORTNAMES   APIVERSION                             NAMESPACED   KIND
bindings                                       v1                                     true         Binding
componentstatuses                 cs           v1                                     false        ComponentStatus
configmaps                        cm           v1                                     true         ConfigMap
endpoints                         ep           v1                                     true         Endpoints
events                            ev           v1                                     true         Event
limitranges                       limits       v1                                     true         LimitRange
namespaces                        ns           v1                                     false        Namespace
nodes                             no           v1                                     false        Node
persistentvolumeclaims            pvc          v1                                     true         PersistentVolumeClaim
persistentvolumes                 pv           v1                                     false        PersistentVolume
pods                              po           v1                                     true         Pod
podtemplates                                   v1                                     true         PodTemplate
replicationcontrollers            rc           v1                                     true         ReplicationController
resourcequotas                    quota        v1                                     true         ResourceQuota
secrets                                        v1                                     true         Secret
serviceaccounts                   sa           v1                                     true         ServiceAccount
services                          svc          v1                                     true         Service
mutatingwebhookconfigurations                  admissionregistration.k8s.io/v1        false        MutatingWebhookConfiguration
validatingwebhookconfigurations                admissionregistration.k8s.io/v1        false        ValidatingWebhookConfiguration
customresourcedefinitions         crd,crds     apiextensions.k8s.io/v1                false        CustomResourceDefinition
apiservices                                    apiregistration.k8s.io/v1              false        APIService
controllerrevisions                            apps/v1                                true         ControllerRevision
daemonsets                        ds           apps/v1                                true         DaemonSet
deployments                       deploy       apps/v1                                true         Deployment
replicasets                       rs           apps/v1                                true         ReplicaSet
statefulsets                      sts          apps/v1                                true         StatefulSet
tokenreviews                                   authentication.k8s.io/v1               false        TokenReview
localsubjectaccessreviews                      authorization.k8s.io/v1                true         LocalSubjectAccessReview
selfsubjectaccessreviews                       authorization.k8s.io/v1                false        SelfSubjectAccessReview
selfsubjectrulesreviews                        authorization.k8s.io/v1                false        SelfSubjectRulesReview
subjectaccessreviews                           authorization.k8s.io/v1                false        SubjectAccessReview
horizontalpodautoscalers          hpa          autoscaling/v2                         true         HorizontalPodAutoscaler
cronjobs                          cj           batch/v1                               true         CronJob
jobs                                           batch/v1                               true         Job
certificatesigningrequests        csr          certificates.k8s.io/v1                 false        CertificateSigningRequest
leases                                         coordination.k8s.io/v1                 true         Lease
endpointslices                                 discovery.k8s.io/v1                    true         EndpointSlice
events                            ev           events.k8s.io/v1                       true         Event
flowschemas                                    flowcontrol.apiserver.k8s.io/v1beta3   false        FlowSchema
prioritylevelconfigurations                    flowcontrol.apiserver.k8s.io/v1beta3   false        PriorityLevelConfiguration
materializes                      mzs          materialize.cloud/v1alpha1             true         Materialize
vpcendpoints                      vpce         materialize.cloud/v1                   true         VpcEndpoint
ingressclasses                                 networking.k8s.io/v1                   false        IngressClass
ingresses                         ing          networking.k8s.io/v1                   true         Ingress
networkpolicies                   netpol       networking.k8s.io/v1                   true         NetworkPolicy
runtimeclasses                                 node.k8s.io/v1                         false        RuntimeClass
poddisruptionbudgets              pdb          policy/v1                              true         PodDisruptionBudget
clusterrolebindings                            rbac.authorization.k8s.io/v1           false        ClusterRoleBinding
clusterroles                                   rbac.authorization.k8s.io/v1           false        ClusterRole
rolebindings                                   rbac.authorization.k8s.io/v1           true         RoleBinding
roles                                          rbac.authorization.k8s.io/v1           true         Role
priorityclasses                   pc           scheduling.k8s.io/v1                   false        PriorityClass
csidrivers                                     storage.k8s.io/v1                      false        CSIDriver
csinodes                                       storage.k8s.io/v1                      false        CSINode
csistoragecapacities                           storage.k8s.io/v1                      true         CSIStorageCapacity
storageclasses                    sc           storage.k8s.io/v1                      false        StorageClass
volumeattachments                              storage.k8s.io/v1                      false        VolumeAttachment
```

I based my list off the resources `kubectl get all` gives in the `materialize` and `materialize-environment` namespaces + nodes, but lemme know if any of these would be useful!

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,451 @@
...
+        Ok(client) => {
+            for namespace in context.args.k8s_namespaces.clone() {
+                if let Err(e) = dump_k8s_pod_logs(&context, client.clone(), &namespace).await {
+                    eprintln!(
+                        "Failed to write k8s pod logs for namespace {}: {}",
+                        namespace, e
+                    );
+                }
+
+                let events_api = Api::<Event>::namespaced(client.clone(), &namespace);
```
SangJunBak, line 150: I initially wanted to enumerate each API resource type and loop over them instead of going one by one. Maybe even map them to K8sResource type. But I figured it was more code than and more complex than worth. Lemme know if it's easier than I think though!

src/self-managed-debug/src/main.rs
alex-hunt-materialize: This syntax only works in the match, not in `if let`.
We use it in our kubernetes shared lib in the cloud repo https://github.com/MaterializeInc/cloud/blob/a33823b83dfe8b5b3bcb41d6b8db25dca85648bb/src/k8s/src/k8sutil.rs#L55 .

Don't spend a ton of time on it, I just figured you weren't aware of the things rust can do in match destructuring.

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,451 @@
...
+                let pods_api = Api::<Pod>::namespaced(client.clone(), &namespace);
+                if let Err(e) =
+                    dump_k8s_resources(&context, pods_api, &K8sResourceType::Pod, Some(&namespace))
+                        .await
+                {
+                    eprintln!(
+                        "Failed to write k8s pods for namespace {}: {}",
+                        namespace, e
+                    );
+                }
```
alex-hunt-materialize, line 173: There isn't an easy way to just loop through the types, but you can reduce duplication quite a bit here by handling the error inside the function call, and possibly creating the API inside the function as well.

You may also want to make wrapper functions for namespaced and non-namespaced resources.

Alternatively, you could create a generic type that instantiates the api internally. Maybe something like:

```rust
struct ResourceDumper<'n, K> {
    context: &'n Context,
    api: Api<K>,
    namespace: Option<&'n str>,
}

impl<'n, K> ResourceDumper<'n, K>
where
    K: kube::Resource<DynamicType = ()>,
    K: ListableResource,
    K: Clone,
    K: std::fmt::Debug,
    K: serde::Serialize,
    K: serde::de::DeserializeOwned,
{
    fn cluster(context: &Context, client: Client) -> Self {
        Self {
            context,
            api: Api::<K>::all(client),
            namespace: None,
        }
    }

    fn namespaced(context: &Context, client: Client, namespace: &str) -> Self
    where
        K: kube::Resource<Scope = NamespaceResourceScope>,
    {
        Self {
            context,
            api: Api::<K>::namespaced(client, namespace),
            namespace: Some(namespace),
        }
    }

    async fn _dump(&self) -> Result<(), Error> {
        let object_list = self.api.list(&ListParams::default()).await?;

        let resource_type = K::URL_PATH_SEGMENT;
        if object_list.items.is_empty() {
            let mut err_msg = format!("No {} found", resource_type);
            if let Some(namespace) = self.namespace {
                err_msg = format!("{} for namespace {}", err_msg, namespace);
            }
            println!("{}", err_msg);
            return Ok(());
        }

        let file_path =
            format_resource_path(self.context.start_time, resource_type, self.namespace);
        let file_name = file_path.join(format!("{}.yaml", resource_type));
        create_dir_all(&file_path)?;
        let mut file = File::create(&file_name)?;

        for item in &object_list.items {
            serde_yaml::to_writer(&mut file, &item)?;
        }

        println!("Exported {}", file_name.display());

        Ok(())
    }

    async fn dump(&self) {
        if let Err(e) = self._dump().await {
            eprintln!("Failed to write k8s {}: {}", K::URL_PATH_SEGMENT, e);
        }
    }
}

pub async fn run(context: &Context, client: &Client, namespace: String) {
    ResourceDumper::<Pod>::namespaced(context, client.clone(), &namespace)
        .dump()
        .await;
    ResourceDumper::<Service>::namespaced(context, client.clone(), &namespace)
        .dump()
        .await;
    ResourceDumper::<Deployment>::namespaced(context, client.clone(), &namespace)
        .dump()
        .await;
    ResourceDumper::<StatefulSet>::namespaced(context, client.clone(), &namespace)
        .dump()
        .await;
    ResourceDumper::<ReplicaSet>::namespaced(context, client.clone(), &namespace)
        .dump()
        .await;
    ResourceDumper::<NetworkPolicy>::namespaced(context, client.clone(), &namespace)
        .dump()
        .await;
    ResourceDumper::<Event>::namespaced(context, client.clone(), &namespace)
        .dump()
        .await;
    ResourceDumper::<Node>::cluster(context, client.clone())
        .dump()
        .await;
}
```

src/self-managed-debug/src/main.rs
```diff
@@ -0,0 +1,451 @@
...
+    }
+
+    let file_path = format_resource_path(context.start_time, &resource_type.to_string(), namespace);
+    let file_name = file_path.join(format!("{}.yaml", resource_type));
+    create_dir_all(&file_path)?;
+    let mut file = File::create(&file_name)?;
+
+    for item in &object_list.items {
+        serde_yaml::to_writer(&mut file, &item)?;
+    }
```
alex-hunt-materialize, line 381: If we're breaking this up per item, we probably want separate files for each item. Alternatively, just serialize the list output directly, rather than looping over items.
Sorry I missed this last time.

```suggestion
    create_dir_all(&file_path)?;

    for item in &object_list.items {
        let file_name = file_path.join(format!("{}.yaml", item.name_unchecked()));
        let mut file = File::create(&file_name)?;
        serde_yaml::to_writer(&mut file, &item)?;
    }
```

src/self-managed-debug/src/main.rs
SangJunBak: That's so clean! I'll try that :D

SangJunBak: Decided to do the same for the kubectl describe function and eliminate our custom enum. Changes here! https://github.com/MaterializeInc/materialize/compare/8d0b2d4be313d2faaf62553d0ea6db47010e5768..e31486cfff017b6736ff7bc5e242acbd38c4c8bd

src/self-managed-debug/Cargo.toml
antiguru: If we want to include anything from `mz_introspection`, we'll need to version it the same way as the rest of Materialize. It's inherently unstable and not compatible across versions.

src/self-managed-debug/src/k8s_resource_dumper.rs
```diff
@@ -0,0 +1,303 @@
...
+
+        if let Err(e) = export_pod_logs(&pods, &pod_name, &file_path, false).await {
+            eprintln!("Failed to export current logs for pod {}: {}", &pod_name, e);
+        }
+    }
+    Ok(())
+}
+
+/// Write k8s pod logs to a yaml file per pod.
+async fn dump_k8s_pod_logs(context: &Context, client: Client, namespace: &String) -> () {
```
antiguru, line 179:
```suggestion
async fn dump_k8s_pod_logs(context: &Context, client: Client, namespace: &String) {
```

src/self-managed-debug/src/k8s_resource_dumper.rs
```diff
@@ -0,0 +1,303 @@
...
+    })
+}
+
+fn format_resource_path(
+    date_time: DateTime<Utc>,
+    resource_type: &str,
+    namespace: Option<&String>,
+) -> PathBuf {
+    let mut path = PathBuf::from(format!(
+        "mz-debug/{}/{}",
```
antiguru, line 295: What OS do we want to support? This will only work on Linux/Mac/etc., but might fail on Windows where the path separator is \. If we'd like to be portable, we should use https://doc.rust-lang.org/std/path/constant.MAIN_SEPARATOR.html instead.

src/self-managed-debug/src/k8s_resource_dumper.rs
```diff
@@ -0,0 +1,303 @@
...
+            self.context.start_time,
+            resource_type,
+            self.namespace.as_ref(),
+        );
+        create_dir_all(&file_path)?;
+
+        for item in &object_list.items {
+            let file_name = file_path.join(format!(
+                "{}.yaml",
+                &item.meta().name.clone().unwrap_or_default()
```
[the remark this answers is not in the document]
alex-hunt-materialize: I _think_ kubernetes objects obtained with `get` or `list` are guaranteed to have a name, so I think this is maybe ok?

It might be good to have a randomized default or something maybe like `unknown_{hash}.yaml` instead, in case that isn't true.

src/self-managed-debug/src/k8s_resource_dumper.rs
```diff
@@ -0,0 +1,303 @@
...
+        .await;
+    K8sResourceDumper::<ReplicaSet>::namespaced(context, client.clone(), namespace.clone())
+        .dump()
+        .await;
+    K8sResourceDumper::<NetworkPolicy>::namespaced(context, client.clone(), namespace.clone())
+        .dump()
+        .await;
+    K8sResourceDumper::<Event>::namespaced(context, client.clone(), namespace.clone())
+        .dump()
+        .await;
```
alex-hunt-materialize, line 207: Looking at your list of api-resources, there are a few things that might make sense to include:

High priority:
- materializes
- roles
- rolebindings
- configmaps
- secrets (probably only the names, maybe with an arg saying whether to include the contents)

Lower priority:
- daemonsets (may want to get this from all namespaces, since they can affect the node even if they are in a separate namespace)
- storageclasses
- persistentvolumes
- persistentvolumeclaims
- serviceaccounts
- mutatingwebhookconfigurations
- validatingwebhookconfigurations
- customresourcedefinitions
- certificates (I mean the ones from cert-manager, you probably need the full name, and they may not exist)

alex-hunt-materialize (commented):
+1 to all Moritz's comments.

src/self-managed-debug/src/k8s_resource_dumper.rs
SangJunBak: > materializes

Ah what are materializes?  Everything else I can try to get!

src/self-managed-debug/src/k8s_resource_dumper.rs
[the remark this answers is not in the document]
SangJunBak: I think this is correct ^^. I wasn't sure what to hash so I went with the easiest approach and used the index of the loop

src/self-managed-debug/src/k8s_resource_dumper.rs
SangJunBak: Was able to get everything but:

> certificates

> daemonsets (may want to get this from all namespaces, since they can affect the node even if they are in a separate namespace)

Not too sure where the certificates in cert-manager live but if they're in :secrets, then seems like getting the secret names cover it! For daemonsets, given their a namespaced resource, I need to actually define a namespaced resource. I'm also trying to reduce the number of `:all` API calls given I don't want the query to fail in case the user doesn't have RBAC permissions for some of the namespaces.

src/self-managed-debug/src/k8s_resource_dumper.rs
```diff
@@ -211,6 +228,34 @@ pub async fn dump_namespaced_resources(context: &Context, client: &Client, names
     K8sResourceDumper::<Event>::namespaced(context, client.clone(), namespace.clone())
         .dump()
         .await;
+    K8sResourceDumper::<Materialize>::namespaced(context, client.clone(), namespace.clone())
```
SangJunBak, line 231: It actually blew me away that we could just reuse this struct here and it just works.

src/self-managed-debug/Cargo.toml
SangJunBak: sgtm!

src/cloud-resources/src/crd/materialize.rs
```diff
@@ -151,6 +151,15 @@ pub mod v1alpha1 {
...
     }
 
+    impl k8s_openapi::Resource for Materialize {
+        const API_VERSION: &'static str = "v1";
+        const GROUP: &'static str = "materialize.cloud";
+        const KIND: &'static str = "Materialize";
+        const VERSION: &'static str = "v1alpha1";
+        const URL_PATH_SEGMENT: &'static str = "materializes";
+        type Scope = k8s_openapi::NamespaceResourceScope;
+    }
```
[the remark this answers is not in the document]
SangJunBak: You'd think but I don't think so! At least to my understanding, inputting a `k8s_openapi` to `kube::Api` I think morphs it into the kube resource https://github.com/kube-rs/kube/blob/8876639ae5624b8ac0632d34b7b1f0fa06ca2477/kube-core/src/resource.rs#L170-L205. And I couldn't find any references to URL_PATH_SEGMENT being set in the kube macro https://github.com/kube-rs/kube/blob/8876639ae5624b8ac0632d34b7b1f0fa06ca2477/kube-derive/src/resource.rs#L42. Lemme know if that checks out with you!

src/self-managed-debug/src/k8s_resource_dumper.rs
```diff
@@ -0,0 +1,376 @@
...
+
+            if let Some(kube::Error::Api(e)) = e.downcast_ref::<kube::Error>() {
+                if e.code == 400 {
+                    eprintln!("No previous logs available for pod {}", pod_name);
+                } else {
+                    print_error();
+                }
+            } else {
+                print_error();
+            }
```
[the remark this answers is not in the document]
SangJunBak: Screenshot 2025-02-19 at 2 24 10 PM

src/cloud-resources/src/crd/materialize.rs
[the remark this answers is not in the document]
doy-materialize: what breaks if you don't do this? i'm confused why this has never come up before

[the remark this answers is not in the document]
SangJunBak: So for our Dumper struct, we need to implement where K also implements `k8s_openapi::Resource` https://github.com/MaterializeInc/materialize/pull/31282/files#diff-32953559f78a9d106cef813ef38b5c319946d920a19ddb3ca89f645572685168R54-R60. We need this in order to use K::URL_PATH_SEGMENT.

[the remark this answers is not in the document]
doy-materialize: can you just use `K::kind()` instead? it seems like that would work fine everywhere you are currently using `K::URL_PATH_SEGMENT` (unless i'm missing something), and that should work directly without needing to do any of this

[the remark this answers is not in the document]
SangJunBak: Actually that's a good point. Or event `K::plural()` (since that's pretty much URL_PATH_SEGMENT). Lemme do that change real quick!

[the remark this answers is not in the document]
SangJunBak: Change done [`cc342bf` (#31282)](https://github.com/MaterializeInc/materialize/pull/31282/commits/cc342bf71f729e8a22fa02059ddf818b60998e46) . Thanks!

alex-hunt-materialize (approved):
I'd prefer separate describe files per object, rather than per type, but I won't block the PR over it.

src/self-managed-debug/src/k8s_resource_dumper.rs
```diff
@@ -0,0 +1,376 @@
...
+        let mut err_msg = format!("Describe: No {} found", resource_type);
+        if let Some(namespace) = namespace {
+            err_msg = format!("{} for namespace {}", err_msg, namespace);
+        }
+        eprintln!("{}", err_msg);
+        return Ok(());
+    }
+
+    let file_path = format_resource_path(context.start_time, resource_type, namespace);
+    let file_name = file_path.join("describe.txt");
```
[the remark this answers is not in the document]
SangJunBak: ohh, yeah that's a good point. Honestly shouldn't be hard. I've created as a TODO in https://github.com/MaterializeInc/database-issues/issues/8908#issue-2805156377 and can create a followup PR!

bosconi:
bugbot run

PR #30576, opened 2024-11-20
storage: track latest reported status and report on reconnect

aljoscha:
Before, we would purely pump upgrades from running dataflows over to the
controller. Which would make it so that when a controller re-connects it
doesn't get status messages for the latest state.

Now, we track the latest status update, per object, and also what has
been reported, per object. And on re-connect we clear out what has been
reported, which makes it so that we report the latest status when a
reconnect happens.

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #31293, opened 2025-02-05
0dt: in preflight checks, notice DDL changes and restart read-only envd

aljoscha:
Before, a hydrating read-only environment would read a catalog snapshot on bootstrap and then not subscribe to further changes. This means that we would not be hydrating collections/replicas that are created in the old version that is still running, and those would then have to be hydrated after cutting over to the new version.

With this change, we periodically check if new collections/replicas where created and we also check right before announcing as ready to promote. When we _do_ notice there was relevant DDL, we halt. This will make it so we're restarted in read-only mode again, and can now read an up-to-date catalog snapshot.

It's important to note that any running clusters are not restarted, so any work that has already gone into hydration will not be lost.

I think there's no one left besides me who has seen this code, but @ParkMyCar might be best to review. @alex-hunt-materialize there's a question in there for you about what cloud can or wants to do. I'll tag you on the specific code line.

@def- We probably want more testing. The new behavior is described above, I imagine a test would be to create tables or things while hydrating, and make sure that they show up in the read-only environment.

### Checklist

### DIFF
--- a/src/environmentd/src/deployment/preflight.rs
+++ b/src/environmentd/src/deployment/preflight.rs
@@ -189,14 +189,16 @@ pub async fn preflight_0dt(
             check_ddl_changes_interval
                 .set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
 
+            let mut should_skip_catchup = false;
             loop {
                 tokio::select! {
-                    () = &mut caught_up_receiver => {
-                        info!("deployment caught up");
-                        break;
-                    }
                     () = &mut skip_catchup => {
                         info!("skipping waiting for deployment to catch up due to administrator request");
+                        should_skip_catchup = true;
+                        break;
+                    }
+                    () = &mut caught_up_receiver => {
+                        info!("deployment caught up");
                         break;
                     }
                     () = &mut caught_up_max_wait_fut => {
@@ -224,17 +226,19 @@ pub async fn preflight_0dt(
 
             // Check for DDL changes one last time before announcing as ready to
             // promote.
-            check_ddl_changes(
-                boot_ts,
-                persist_client.clone(),
-                environment_id.clone(),
-                deploy_generation,
-                Arc::clone(&catalog_metrics),
-                bootstrap_args.clone(),
-                initial_next_user_item_id,
-                initial_next_replica_id,
-            )
-            .await;
+            if !should_skip_catchup {
+                check_ddl_changes(
+                    boot_ts,
+                    persist_client.clone(),
+                    environment_id.clone(),
+                    deploy_generation,
+                    Arc::clone(&catalog_metrics),
+                    bootstrap_args.clone(),
+                    initial_next_user_item_id,
+                    initial_next_replica_id,
+                )
+                .await;
+            }
 
             // Announce that we're ready to promote.
             let promoted = deployment_state.set_ready_to_promote();
--- a/src/environmentd/src/deployment/preflight.rs
+++ b/src/environmentd/src/deployment/preflight.rs
@@ -192,6 +192,8 @@ pub async fn preflight_0dt(
             let mut should_skip_catchup = false;
             loop {
                 tokio::select! {
+                    biased;
+
                     () = &mut skip_catchup => {
                         info!("skipping waiting for deployment to catch up due to administrator request");
                         should_skip_catchup = true;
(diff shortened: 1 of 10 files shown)


src/environmentd/src/deployment/preflight.rs
```diff
@@ -190,6 +241,19 @@ pub async fn preflight_0dt(
             // Take over the catalog.
             info!("promoted; attempting takeover");
 
+            // NOTE: There _is_ a window where DDL can happen in the old
```
aljoscha, line 244: @alex-hunt-materialize If cloud is set up for this, or wants to make that change, we could try and shorten the window for missed DDL a bit more.

def-:
> @def- We probably want more testing. The new behavior is described above, I imagine a test would be to create tables or things while hydrating, and make sure that they show up in the read-only environment.

On it, will push the new test into this PR.

src/environmentd/src/deployment/preflight.rs
alex-hunt-materialize: I don't think we should worry too much about DDL made in the last few seconds before promotion. At most, it will only be a few seconds lost of hydration.

def-:
The existing parallel-workload test is already failing: https://buildkite.com/materialize/nightly/builds/11058#_
```
materialize.ui.UIError: Timed out waiting for materialized2 to reach Mz deployment status ReadyToPromote, still in status Initializing
```
I'm seeing a very similar error in my new 0dt ddl scenario. Just pushed it, reproduces with `bin/mzcompose --find 0dt down && bin/mzcompose --find 0dt run ddl`, after running a `CREATE TABLE t1 (c INT)` it never manages to promote. For the tests do I need to set Materialize to auto-restart? It's currently `restart="on-failure"`

aljoscha:
> I'm seeing a very similar error in my new 0dt ddl scenario. Just pushed it, reproduces with `bin/mzcompose --find 0dt down && bin/mzcompose --find 0dt run ddl`, after running a `CREATE TABLE t1 (c INT)` it never manages to promote. For the tests do I need to set Materialize to auto-restart? It's currently `restart="on-failure"`

Hmm, maybe, what happens mechanically is that the new `envd` will halt and then expects to be restarted. So tentatively I'd say yes it needs `restart="on-failure"`.

aljoscha:
@def- No forget what I said. I'm pretty sure it's because of a known bug that is fixed in https://github.com/MaterializeInc/materialize/pull/30576. I even meant to talk to you about that other fix, and how the 0dt tests should already be a good enough test for it, once they have restarts of the read-only `envd` (which is what this here PR introduces). It's clearly getting late... 🙈

src/environmentd/src/deployment/preflight.rs
```diff
@@ -154,34 +155,84 @@ pub async fn preflight_0dt(
...
-                () = caught_up_receiver => {
-                    info!("deployment caught up");
-                }
-                () = skip_catchup => {
-                    info!("skipping waiting for deployment to catch up due to administrator request");
-                }
-                () = caught_up_max_wait_fut => {
-                    if panic_after_timeout {
-                        panic!("not caught up within {:?}", caught_up_max_wait);
+            let mut check_ddl_changes_interval = tokio::time::interval(Duration::from_secs(5 * 60));
```
ParkMyCar, line 185: Is it possible to make this configurable via LaunchDarkly? Alternatively could we make this an optional CLI arg that can be provided to `environmentd`. Having a knob here to tune this seems useful

src/environmentd/src/deployment/preflight.rs
```diff
@@ -154,34 +155,84 @@ pub async fn preflight_0dt(
...
+            check_ddl_changes_interval
+                .set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+
+            loop {
+                tokio::select! {
```
ParkMyCar, line 190: maybe make this a biased `select!` loop? The chances of two things resolving at once is low, but it would make it easier to reason about if we have a defined order

ParkMyCar (approved):
Did you mean to include the storage status reporting changes here? Everything LGTM but just wanted to double check

aljoscha:
@def- what do you make of these remaining nightly failures?

aljoscha:
@ParkMyCar I pushed individual commits that implement your suggestions, ptal 🙏

def-:
> @def- what do you make of these remaining nightly failures?

Sorry, just saw your message. So far the new run is looking good, but I haven't seen the SQLancer error before (but probably not related to this PR): https://buildkite.com/materialize/nightly/builds/11085

def- (approved):

PR #31295, opened 2025-02-05
ci: improve aws terraform module test s3 bucket deletion

bobbyiliev:
### Motivation

As per the suggestion [here](https://github.com/MaterializeInc/terraform-aws-materialize/issues/17#issuecomment-2632764448), disabling S3 bucket versioning and encryption to improve the S3 bucket deletion time in tests.


def- (approved):
Thanks!

PR #31300, opened 2025-02-05
[adapter] Stream BuiltinTable retractions at bootstrap

ParkMyCar:
This PR refactors Coordinator startup, specifically `bootstrap_tables()` to stream our retractions into a Persist `BatchBuilder` instead of buffering them in-memory. It also makes two `txn-wal` related fixes that were required to get this passing. A more detailed explanation of them is in a comment below.

### Motivation

Fix/reduce a memory spike we've seen at startup.

### Checklist

### DIFF
--- a/src/adapter/src/catalog/builtin_table_updates.rs
+++ b/src/adapter/src/catalog/builtin_table_updates.rs
@@ -46,6 +46,7 @@ use mz_expr::MirScalarExpr;
 use mz_orchestrator::{CpuLimit, DiskLimit, MemoryLimit, ServiceProcessMetrics};
 use mz_ore::cast::CastFrom;
 use mz_ore::collections::CollectionExt;
+use mz_persist_client::batch::ProtoBatch;
 use mz_repr::adt::array::ArrayDimension;
 use mz_repr::adt::interval::Interval;
 use mz_repr::adt::jsonb::Jsonb;
@@ -67,6 +68,7 @@ use mz_sql::plan::{ClusterSchedule, ConnectionDetails, SshKey};
 use mz_sql::session::user::SYSTEM_USER;
 use mz_sql::session::vars::SessionVars;
 use mz_sql_parser::ast::display::AstDisplay;
+use mz_storage_client::client::TableData;
 use mz_storage_types::connections::aws::{AwsAuth, AwsConnection};
 use mz_storage_types::connections::inline::ReferencedConnection;
 use mz_storage_types::connections::string_or_secret::StringOrSecret;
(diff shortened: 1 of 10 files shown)


ParkMyCar:
@aljoscha I believe I figured out what was wrong! The root cause seemed to be two batches getting appended to a single collection at the same timestamp. One of the batches getting appended are the retractions we use to reset the table, the other I'm not 100% positive what it is, but it results in the [`datas` field in `TxnCacheState`](https://github.com/MaterializeInc/materialize/blob/main/src/txn-wal/src/txn_cache.rs#L101-L105) having writes with duplicate timestamps. According to the invariants documented on that field it seems like this is within spec. The problem comes when compacting the [`times` field within `datas`](https://github.com/MaterializeInc/materialize/blob/main/src/txn-wal/src/txn_cache.rs#L618-L628) we don't consider this duplicate case and thus the duplicates remain. Later when validating that everything is compacted, we're allowed to have one "write" that is less than or equal to the `min_unapplied_ts`, because it's the latest. But this check fails because we have two timestamps, thus we're not compacted, but they're the same thing.

The first fix was to deduplicate the `times` field as part of compaction.

The second issue is also from having two batches at the same time, specifically one that we're re-writing the timestamp for (the retractions) and another that we're not (the second unknown batch). That causes [this check](https://github.com/MaterializeInc/materialize/blob/main/src/persist-client/src/write.rs#L483-L489) within Persist to fail ([check impl](https://github.com/MaterializeInc/materialize/blob/main/src/persist-client/src/batch.rs#L1467-L1474)) because when `txn-wal` [generates a Batch with in-memory Rows we use a lower of T::minimum()](https://github.com/MaterializeInc/materialize/blob/main/src/txn-wal/src/txn_write.rs#L199). Then when calling `validate_truncate_batch` [here](https://github.com/MaterializeInc/materialize/blob/main/src/persist-client/src/batch.rs#L1467-L1468) the `part_lower_bound` ends up being `T::minimum()` (aka `0`) and `truncate.lower()` is our commit timestamp.

I see two ways to fix this second issue:

1. Relax the validation in Persist so if instead of "any_batch_rewrite" we do the validation on a Batch-by-Batch basis.
2. Change `txn-wal` to rewrite the timestamp for all batches it generates. Since we already rewrite the timestamp on staged batches I figured this was the easier to reason about change and thus opted for this!

Alternatively to both of these fixes we could try and prevent two batches from getting appended at the same timestamp, but that seems like something we should support in general, so I opted for these fixes instead of trying to prevent that problem.

aljoscha (approved):
Before merging, could you please organize the two txn/persist fixes into two neat commits that have as a description your description of the bugs from the comment here, along with the rationale for how we fix it? That way we can keep track of how and why these where changes. You might already have planned that but wanted to make sure! 😅

ParkMyCar:
@aljoscha TFTR!

Totally agree with the proposed patch, I went ahead and made that, it's in the commit titled "txn-wal fix 1". I also updated the commit messages for "txn-wal fix 1" and "txn-wal fix 2" with a description of the actual fix and root cause.

PR #31307, opened 2025-02-06
source/postgres: move schema validation onto a separate task

teskje:
This PR moves the postgres source's periodic schema validation onto a separate task and gives it a separate upstream client. The intention is to avoid that (potentially slow) fetching of schema information stalls the progress of the replication reader, leading to responsiveness regressions for consumers of the source.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/8945

### Tips for reviewers

I recommend reviewing with whitespace disabled!

### Checklist

### DIFF
--- a/src/storage/src/source/postgres/replication.rs
+++ b/src/storage/src/source/postgres/replication.rs
@@ -91,7 +91,7 @@ use mz_postgres_util::PostgresError;
 use mz_postgres_util::{simple_query_opt, Client};
 use mz_repr::{Datum, DatumVec, Diff, Row};
 use mz_sql_parser::ast::{display::AstDisplay, Ident};
-use mz_storage_types::dyncfgs::PG_OFFSET_KNOWN_INTERVAL;
+use mz_storage_types::dyncfgs::{PG_OFFSET_KNOWN_INTERVAL, PG_SCHEMA_VALIDATION_INTERVAL};
 use mz_storage_types::errors::DataflowError;
 use mz_storage_types::sources::SourceTimestamp;
 use mz_storage_types::sources::{MzOffset, PostgresSourceConnection};
@@ -111,7 +111,7 @@ use timely::dataflow::operators::Capability;
 use timely::dataflow::operators::Concat;
 use timely::dataflow::{Scope, Stream};
 use timely::progress::Antichain;
-use tokio::sync::watch;
+use tokio::sync::{mpsc, watch};
 use tokio_postgres::error::SqlState;
 use tokio_postgres::types::PgLsn;
 use tracing::{error, trace};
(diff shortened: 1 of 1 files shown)


src/storage/src/source/postgres/replication.rs
```diff
@@ -500,68 +508,48 @@ pub(crate) fn render<G: Scope<Timestamp = MzOffset>>(
...
-                                        Ok(()) => {
-                                            trace!(%id, "timely-{worker_id} schema of output \
-                                                index {output_index} for oid {oid} valid");
-                                        }
-                                        Err(err) => {
-                                            trace!(%id, "timely-{worker_id} schema of output \
-                                                index {output_index} for oid {oid} invalid");
+
+                        // Take the opportunity to report any schema validation errors.
+                        while let Ok(error) = schema_errors.try_recv() {
```
petrosagg, line 513: Using the tokio `Watch` here is just as good since we only ever want to process the most recent schema

src/storage/src/source/postgres/replication.rs
```diff
@@ -1104,3 +1092,66 @@ async fn ensure_replication_timeline_id(
...
+                        trace!(
+                            %source_id,
+                            "schema of output index {output_index} for oid {oid} valid",
+                        );
+                    }
+                }
+            }
+
+            let validation_interval = PG_SCHEMA_VALIDATION_INTERVAL.get(&config_set);
+            tokio::time::sleep(validation_interval).await;
```
petrosagg, line 1152: nit: this will make the validation happen every `validation_interval + schema_fetch_duration`. We can define a tokio Interval outside the loop so that we can tick every `validation_interval` reliably

petrosagg (approved):
Looks great! Only some minor non-blocking comments

src/storage/src/source/postgres/replication.rs
teskje: I tried this out: A `Watch` is a bit annoying to use in the schema validator task because we can have multiple errors, so we need to collect them all in a vec and then send them at the end. In the replication reader we also can't `take` the errors because we only get immutable references to them, so we have to clone them and then remember that we've already seen them. Both doable but more noisy than using a channel.

We could use an `Arc<Mutex<...>` instead of a `Watch` to solve the immutable references issue, but we'd still have to do the collection into a vec in the schema validator task.

src/storage/src/source/postgres/replication.rs
teskje: A tokio interval makes it difficult to apply updated `PG_SCHEMA_VALIDATION_INTERVAL` values at runtime, so I instead kept the `sleep` but am now accounting for how long the validation took.

teskje:
TFTR!

PR #31309, opened 2025-02-06
[adapter] don't block on builtin table write in `Session` creation

ParkMyCar:
This PR does two things:

1. Changes explicitly triggered group commits (e.g. a write to `mz_sessions`) to have a minimum batching interval. Today when we trigger a group commit we only allow 1 to be in-flight at a time, this allows pending writes to naturally batch. But if the write is small there's a good chance Persist will inline it and it will take low 10s of milliseconds. Added in this PR is a dyncfg `group_commit_batch_duration` which ensures we will wait X duration before running the next group commit. The default value is 100ms.
2. When starting a `Session`/connection we have to update the `mz_sessions` table. Waiting for this write to complete has been moved to when a Session receives a query, and if that query reads from the `mz_sessions` table. This means starting a `Session` is a completely in-memory operation, and waiting for the builtin table write to complete happens concurrently with the network roundtrip of responding to the client, and them sending their request.

Theoretically it could _increase_ latency for opening a new connection, if that connection runs `SELECT ... FROM mz_sessions` because we now we might wait up to 100ms for the write to complete. But this seems like an acceptable tradeoff because it speeds up all other queries and is tune-able in LaunchDarkly if it becomes an issue.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/8951

### Checklist

### DIFF
--- a/src/adapter/src/coord/appends.rs
+++ b/src/adapter/src/coord/appends.rs
@@ -44,16 +44,21 @@ use crate::session::{GroupCommitWriteLocks, Session, WriteLocks};
 use crate::util::{CompletedClientTransmitter, ResultExt};
 use crate::{AdapterError, ExecuteContext};
 
-/// An operation that was deferred waiting for [`WriteLocks`].
+/// Tables that we emit updates for when starting a new session.
+pub(crate) static REQUIRED_BUILTIN_TABLES: &[&LazyLock<BuiltinTable>] = &[&MZ_SESSIONS];
+
+/// An operation that was deferred waiting on a resource to be available.
+///
+/// For example when inserting into a table we defer on acquiring [`WriteLocks`].
 #[derive(Debug)]
-pub enum DeferredWriteOp {
+pub enum DeferredOp {
     /// A plan, e.g. ReadThenWrite, that needs locks before sequencing.
     Plan(DeferredPlan),
     /// Inserts into a collection.
     Write(DeferredWrite),
 }
 
-impl DeferredWriteOp {
+impl DeferredOp {
     /// Certain operations, e.g. "blind writes"/`INSERT` statements, can be optimistically retried
     /// because we can share a write lock between multiple operations. In this case we wait to
     /// acquire the locks until [`group_commit`], where writes are groupped by collection and
@@ -67,19 +72,19 @@ impl DeferredWriteOp {
     /// [`group_commit`]: crate::coord::Coordinator::group_commit
     pub(crate) fn can_be_optimistically_retried(&self) -> bool {
         match self {
-            DeferredWriteOp::Plan(_) => false,
-            DeferredWriteOp::Write(_) => true,
+            DeferredOp::Plan(_) => false,
+            DeferredOp::Write(_) => true,
         }
     }
 
     /// Returns an Iterator of all the required locks for current operation.
     pub fn required_locks(&self) -> impl Iterator<Item = CatalogItemId> + '_ {
         match self {
-            DeferredWriteOp::Plan(plan) => {
+            DeferredOp::Plan(plan) => {
                 let iter = plan.requires_locks.iter().copied();
                 itertools::Either::Left(iter)
             }
-            DeferredWriteOp::Write(write) => {
+            DeferredOp::Write(write) => {
                 let iter = write.writes.keys().copied();
                 itertools::Either::Right(iter)
             }
@@ -89,16 +94,16 @@ impl DeferredWriteOp {
     /// Returns the [`ConnectionId`] associated with this deferred op.
     pub fn conn_id(&self) -> &ConnectionId {
         match self {
-            DeferredWriteOp::Plan(plan) => plan.ctx.session().conn_id(),
-            DeferredWriteOp::Write(write) => write.pending_txn.ctx.session().conn_id(),
+            DeferredOp::Plan(plan) => plan.ctx.session().conn_id(),
+            DeferredOp::Write(write) => write.pending_txn.ctx.session().conn_id(),
         }
     }
 
     /// Consumes the [`DeferredWriteOp`], returning the inner [`ExecuteContext`].
     pub fn into_ctx(self) -> ExecuteContext {
         match self {
-            DeferredWriteOp::Plan(plan) => plan.ctx,
-            DeferredWriteOp::Write(write) => write.pending_txn.ctx,
+            DeferredOp::Plan(plan) => plan.ctx,
+            DeferredOp::Write(write) => write.pending_txn.ctx,
         }
     }
 }
@@ -221,7 +226,7 @@ impl Coordinator {
         };
 
         match op {
-            DeferredWriteOp::Plan(mut deferred) => {
+            DeferredOp::Plan(mut deferred) => {
                 if let Err(e) = deferred.validity.check(self.catalog()) {
                     deferred.ctx.retire(Err(e))
                 } else {
@@ -249,7 +254,7 @@ impl Coordinator {
                         .await;
                 }
             }
-            DeferredWriteOp::Write(DeferredWrite {
+            DeferredOp::Write(DeferredWrite {
                 span,
                 writes,
                 pending_txn,
@@ -431,7 +436,7 @@ impl Coordinator {
 
         // Queue all of our deferred ops.
         for (acquire_future, write) in deferred_writes {
-            self.defer_op(acquire_future, DeferredWriteOp::Write(write));
+            self.defer_op(acquire_future, DeferredOp::Write(write));
         }
 
         // The value returned here still might be ahead of `now()` if `now()` has gone backwards at
@@ -636,7 +641,7 @@ impl Coordinator {
         BuiltinTableAppend { coord: self }
     }
 
-    pub(crate) fn defer_op<F>(&mut self, acquire_future: F, op: DeferredWriteOp)
+    pub(crate) fn defer_op<F>(&mut self, acquire_future: F, op: DeferredOp)
     where
         F: Future<Output = Option<(CatalogItemId, tokio::sync::OwnedMutexGuard<()>)>>
             + Send
@@ -983,9 +988,6 @@ pub(crate) fn waiting_on_startup_appends(
     session: &mut Session,
     plan: &Plan,
 ) -> Option<(BTreeSet<CatalogItemId>, BoxFuture<'static, ()>)> {
-    /// Tables that we emit updates for when starting a new session.
-    static REQUIRED_BUILTIN_TABLES: &[&LazyLock<BuiltinTable>] = &[&MZ_SESSIONS];
-
     let depends_on = match plan {
         Plan::Select(plan) => plan.source.depends_on(),
         Plan::ReadThenWrite(plan) => plan.selection.depends_on(),
--- a/src/adapter/src/coord/appends.rs
+++ b/src/adapter/src/coord/appends.rs
@@ -11,12 +11,15 @@
 
 use std::collections::{BTreeMap, BTreeSet};
 use std::future::Future;
+use std::pin::Pin;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use derivative::Derivative;
-use futures::future::{BoxFuture, FutureExt};
+use futures::future::FutureExt;
 use mz_adapter_types::connection::ConnectionId;
+use mz_adapter_types::dyncfgs::GROUP_COMMIT_BATCH_DURATION;
+use mz_dyncfg::{ConfigSet, ConfigValHandle};
 use mz_ore::metrics::MetricsFutureExt;
 use mz_ore::task;
 use mz_ore::tracing::OpenTelemetryContext;
@@ -27,12 +30,14 @@ use mz_sql::plan::Plan;
 use mz_sql::session::metadata::SessionMetadata;
 use mz_storage_client::client::TableData;
 use mz_timestamp_oracle::WriteTimestamp;
+use prometheus::Histogram;
 use smallvec::SmallVec;
 use tokio::sync::{oneshot, Notify, OwnedMutexGuard, OwnedSemaphorePermit, Semaphore};
 use tracing::{debug_span, info, warn, Instrument, Span};
 
 use crate::catalog::BuiltinTableUpdate;
 use crate::coord::{Coordinator, Message, PendingTxn, PlanValidity};
+use crate::metrics::GroupCommitMetrics;
 use crate::session::{GroupCommitWriteLocks, WriteLocks};
 use crate::util::{CompletedClientTransmitter, ResultExt};
 use crate::{AdapterError, ExecuteContext};
@@ -703,7 +708,7 @@ pub struct BuiltinTableAppend<'a> {
 /// Note: builtin table writes need to talk to persist, which can take 100s of milliseconds. This
 /// type allows you to execute a builtin table write, e.g. via [`BuiltinTableAppend::execute`], and
 /// wait for it to complete, while other long running tasks are concurrently executing.
-pub type BuiltinTableAppendNotify = BoxFuture<'static, ()>;
+pub type BuiltinTableAppendNotify = Pin<Box<dyn Future<Output = ()> + Send + Sync + 'static>>;
 
 impl<'a> BuiltinTableAppend<'a> {
     /// Submit a write to a system table to be executed during the next group commit. This method
@@ -749,7 +754,7 @@ impl<'a> BuiltinTableAppend<'a> {
                 .expect("in read-only mode")
                 .append(&mut updates);
 
-            return futures::future::ready(()).boxed();
+            return Box::pin(futures::future::ready(()));
         }
 
         let (tx, rx) = oneshot::channel();
@@ -782,7 +787,7 @@ impl<'a> BuiltinTableAppend<'a> {
                 .expect("in read-only mode")
                 .append(&mut updates);
 
-            return (futures::future::ready(()).boxed(), None);
+            return (Box::pin(futures::future::ready(())), None);
         }
 
         let (tx, rx) = oneshot::channel();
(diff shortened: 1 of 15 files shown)


def- (commented):
Nightly triggered: https://buildkite.com/materialize/nightly/builds/11069

I assume no new tests required since this shouldn't have noticeable differences from the outside.

ParkMyCar:
Thanks for triggering Nightlies @def-!

ParkMyCar:
Re-triggering nightlies

def-:
The MySQL platform-checks errors in Nightly should be fixed by rebasing, but the performance regressions are probably expected?
Should we disable the dyncfg in `ADDITIONAL_BENCHMARKING_SYSTEM_PARAMETERS` so that benchmarks are not affected?

ParkMyCar:
Yeah the performance regressions are expected, and almost perfectly line up with the current thresholds. e.g. 4 connections per-second on a single thread makes sense because of the 250ms batching duration.

That being said, it made me realize the current threshold is maybe a little too slow, so I reduced it down to 100ms, and also set a test override to 25ms so we can benchmark what the full throughput would be

ParkMyCar:
> I'm a bit worried about this increasing the latency for opening a connection and then doing a peek.

The more I thought about it the more I was concerned too! Thanks for your experimentation, I just pushed up another commit which should do exactly what we want, only block the query if it's trying to read from the `mz_sessions` table itself, otherwise the query is allowed to proceed.

I also reverted the change to tweak to `group_commit_batch_duration` for tests, it will now use the default of 100ms, let's see what the benchmarks have to say!

ParkMyCar:
A few nightly test failures but they don't look related? Some nice improvements in connection scaling!
```
mprovements:
* Scalability improvement in workload 'EstablishConnectionWorkload' at concurrency 1 with MaterializeContainer (None specified as HEAD): 121.42 tps vs. 25.35 tps (96.07 tps; 378.97%)
* Scalability improvement in workload 'EstablishConnectionWorkload' at concurrency 2 with MaterializeContainer (None specified as HEAD): 251.47 tps vs. 57.68 tps (193.8 tps; 336.0%)
* Scalability improvement in workload 'EstablishConnectionWorkload' at concurrency 6 with MaterializeContainer (None specified as HEAD): 372.64 tps vs. 134.13 tps (238.51 tps; 177.82%)
* Scalability improvement in workload 'EstablishConnectionWorkload' at concurrency 16 with MaterializeContainer (None specified as HEAD): 412.48 tps vs. 228.05 tps (184.43 tps; 80.87%)
* Scalability improvement in workload 'EstablishConnectionWorkload' at concurrency 39 with MaterializeContainer (None specified as HEAD): 418.7 tps vs. 260.13 tps (158.58 tps; 60.96%)
* Scalability improvement in workload 'EstablishConnectionWorkload' at concurrency 98 with MaterializeContainer (None specified as HEAD): 464.28 tps vs. 295.17 tps (169.11 tps; 57.29%)
* Scalability improvement in workload 'EstablishConnectionWorkload' at concurrency 244 with MaterializeContainer (None specified as HEAD): 452.65 tps vs. 321.64 tps (131.01 tps; 40.73%)
* Scalability improvement in workload 'EstablishConnectionWorkload' at concurrency 610 with MaterializeContainer (None specified as HEAD): 433.84 tps vs. 294.21 tps (139.63 tps; 47.46%)
* Scalability improvement in workload 'EstablishConnectionWorkload' at concurrency 1526 with MaterializeContainer (None specified as HEAD): 349.57 tps vs. 215.8 tps (133.78 tps; 61.99%)
```

src/adapter/src/coord/appends.rs
```diff
@@ -958,3 +967,140 @@ impl Drop for GroupCommitPermit {
...
+/// So as an optimization we do not wait for these writes to complete. But if a [`Session`] tries
+/// to query any of these builtin objects, we need to block that query on the writes completing to
+/// maintain linearizability.
+pub(crate) fn waiting_on_startup_appends(
+    catalog: &Catalog,
+    session: &mut Session,
+    plan: &Plan,
+) -> Option<(BTreeSet<CatalogItemId>, BoxFuture<'static, ()>)> {
+    /// Tables that we emit updates for when starting a new session.
+    static REQUIRED_BUILTIN_TABLES: &[&LazyLock<BuiltinTable>] = &[&MZ_SESSIONS];
```
ggevay, line 987: Maybe put a comment about this also where the table updates happen (`handle_startup`), so that if somebody adds more table updating there, then they'll know to also update this constant.

ggevay:
Thank you! The code looks good to me, but I'm not an expert, so maybe somebody from the Adapter team should also review it.

One minor thing is that you could update the comment of `DeferredWriteOp`, because it specifically says it's for waiting for `WriteLocks`, but now we will be waiting for a different thing if I understand correctly.

> A few nightly test failures but they don't look related?

Well, the benchmark failures are not impossible to be related somehow, I guess. But I'm not sure.

ParkMyCar:
> Thank you! The code looks good to me, but I'm not an expert, so maybe somebody from the Adapter team should also review it.

We're a bit short on review bandwidth at the moment! We can have @aljoscha take a cursory pass but it would be great if you don't mind giving a review too @ggevay :)

> One minor thing is that you could update the comment of `DeferredWriteOp`, because it specifically says it's for waiting for `WriteLocks`, but now we will be waiting for a different thing if I understand correctly.

Good callout, I updated the comment and renamed the enum from `DeferredWriteOp` to `DeferredOp`.

> Well, the benchmark failures are not impossible to be related somehow, I guess. But I'm not sure.

Just rebased on `main` and will re-run them again. Will def debug ones that aren't also failing on main.

ggevay:
Ok, I'll look into it more deeply tomorrow, delving into the `Future` stuff!

ggevay:
The Nightly benchmark failures (e.g. `ManySmallInserts`) are because of `group_commit_batch_duration` unfortunately slowing down inserts. In a sequence of inserts, each insert after the first one waits for ~100ms:
(compiled in debug mode)
```
materialize=> CREATE TABLE t(x int, y int);
CREATE TABLE
materialize=> \timing
Timing is on.
materialize=> INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT 0 1
Time: 40.276 ms
INSERT 0 1
Time: 108.467 ms
INSERT 0 1
Time: 87.613 ms
INSERT 0 1
Time: 101.167 ms
INSERT 0 1
Time: 99.876 ms
INSERT 0 1
Time: 102.039 ms
INSERT 0 1
Time: 103.497 ms
INSERT 0 1
Time: 98.783 ms
INSERT 0 1
Time: 107.178 ms
INSERT 0 1
Time: 94.033 ms
INSERT 0 1
Time: 102.344 ms
```
If I do `alter system set group_commit_batch_duration = '10ms';`, then:
```
materialize=> CREATE TABLE t(x int, y int);
CREATE TABLE
materialize=> \timing
Timing is on.
materialize=> INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT 0 1
Time: 63.708 ms
INSERT 0 1
Time: 30.211 ms
INSERT 0 1
Time: 29.694 ms
INSERT 0 1
Time: 30.768 ms
INSERT 0 1
Time: 31.536 ms
INSERT 0 1
Time: 29.493 ms
INSERT 0 1
Time: 33.950 ms
INSERT 0 1
Time: 34.574 ms
INSERT 0 1
Time: 39.077 ms
INSERT 0 1
Time: 27.500 ms
INSERT 0 1
Time: 27.021 ms
```
If I do `alter system set group_commit_batch_duration = '1000ms';`, then:
```
materialize=> CREATE TABLE t(x int, y int);
CREATE TABLE
materialize=> \timing
Timing is on.
materialize=> INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT INTO t VALUES (1,2);
INSERT 0 1
Time: 47.987 ms
INSERT 0 1
Time: 1005.307 ms (00:01.005)
INSERT 0 1
Time: 1012.090 ms (00:01.012)
INSERT 0 1
Time: 989.670 ms
INSERT 0 1
Time: 1024.006 ms (00:01.024)
INSERT 0 1
Time: 983.476 ms
INSERT 0 1
Time: 1006.929 ms (00:01.007)
INSERT 0 1
Time: 1000.307 ms (00:01.000)
INSERT 0 1
Time: 1004.053 ms (00:01.004)
INSERT 0 1
Time: 996.474 ms
INSERT 0 1
Time: 1007.284 ms (00:01.007)
```

I'm thinking about what to do about this.

I think the many small inserts use case being as fast as possible is somewhat important, because it's probably often used. It's maybe not so much used in production, but I'm thinking it's definitely used during initial exploration, where users will notice if this is slow. (Or wait, webhook sources also do something like inserts, right? How often do they want to do inserts?)

However, if we don't batch inserts, then they can similarly put pressure on `mz_catalog_server` in the same way as the `mz_session` writes in the incident. I've tried this in my staging env (not a custom build, just what we currently have on staging), and `mz_catalog_server`'s CPU usage indeed jumps up to 100% while I'm doing inserts in a hot loop. The Console is still somewhat responsive, but noticeably slower than if I'm not doing anything.

ggevay:
As a reference, these are the timings for 100 inserts pasted into the Console's shell:
```
INSERT 0 1
Returned in 95.0ms
INSERT 0 1
Returned in 7.6ms
INSERT 0 1
Returned in 11.4ms
INSERT 0 1
Returned in 23.6ms
INSERT 0 1
Returned in 32.0ms
INSERT 0 1
Returned in 31.7ms
INSERT 0 1
Returned in 33.5ms
INSERT 0 1
Returned in 31.5ms
INSERT 0 1
Returned in 37.0ms
INSERT 0 1
Returned in 29.7ms
INSERT 0 1
Returned in 34.2ms
INSERT 0 1
Returned in 32.8ms
INSERT 0 1
Returned in 33.4ms
INSERT 0 1
Returned in 34.8ms
INSERT 0 1
Returned in 31.3ms
INSERT 0 1
Returned in 34.2ms
INSERT 0 1
Returned in 31.0ms
INSERT 0 1
Returned in 33.2ms
INSERT 0 1
Returned in 33.2ms
INSERT 0 1
Returned in 32.8ms
INSERT 0 1
Returned in 32.8ms
INSERT 0 1
Returned in 30.9ms
INSERT 0 1
Returned in 31.7ms
INSERT 0 1
Returned in 32.4ms
INSERT 0 1
Returned in 29.8ms
INSERT 0 1
Returned in 35.2ms
INSERT 0 1
Returned in 38.2ms
INSERT 0 1
Returned in 33.9ms
INSERT 0 1
Returned in 33.0ms
INSERT 0 1
Returned in 27.9ms
INSERT 0 1
Returned in 30.0ms
INSERT 0 1
Returned in 32.1ms
INSERT 0 1
Returned in 37.0ms
INSERT 0 1
Returned in 35.0ms
INSERT 0 1
Returned in 26.9ms
INSERT 0 1
Returned in 33.8ms
INSERT 0 1
Returned in 31.4ms
INSERT 0 1
Returned in 33.8ms
INSERT 0 1
Returned in 31.7ms
INSERT 0 1
Returned in 34.3ms
INSERT 0 1
Returned in 31.7ms
INSERT 0 1
Returned in 39.6ms
INSERT 0 1
Returned in 36.3ms
INSERT 0 1
Returned in 31.6ms
INSERT 0 1
Returned in 30.7ms
INSERT 0 1
Returned in 33.9ms
INSERT 0 1
Returned in 34.0ms
INSERT 0 1
Returned in 35.0ms
INSERT 0 1
Returned in 31.0ms
INSERT 0 1
Returned in 33.6ms
INSERT 0 1
Returned in 31.4ms
INSERT 0 1
Returned in 34.4ms
INSERT 0 1
Returned in 32.0ms
INSERT 0 1
Returned in 31.5ms
INSERT 0 1
Returned in 31.0ms
INSERT 0 1
Returned in 37.1ms
INSERT 0 1
Returned in 33.5ms
INSERT 0 1
Returned in 32.7ms
INSERT 0 1
Returned in 31.6ms
INSERT 0 1
Returned in 30.7ms
INSERT 0 1
Returned in 31.7ms
INSERT 0 1
Returned in 33.0ms
INSERT 0 1
Returned in 36.2ms
INSERT 0 1
Returned in 27.2ms
INSERT 0 1
Returned in 37.6ms
INSERT 0 1
Returned in 34.3ms
INSERT 0 1
Returned in 34.4ms
INSERT 0 1
Returned in 29.7ms
INSERT 0 1
Returned in 28.9ms
INSERT 0 1
Returned in 31.0ms
INSERT 0 1
Returned in 28.7ms
INSERT 0 1
Returned in 34.1ms
INSERT 0 1
Returned in 35.6ms
INSERT 0 1
Returned in 31.7ms
INSERT 0 1
Returned in 30.3ms
INSERT 0 1
Returned in 31.4ms
INSERT 0 1
Returned in 35.4ms
INSERT 0 1
Returned in 34.8ms
INSERT 0 1
Returned in 35.1ms
INSERT 0 1
Returned in 35.0ms
INSERT 0 1
Returned in 33.4ms
INSERT 0 1
Returned in 31.8ms
INSERT 0 1
Returned in 32.9ms
INSERT 0 1
Returned in 33.5ms
INSERT 0 1
Returned in 32.6ms
INSERT 0 1
Returned in 32.6ms
INSERT 0 1
Returned in 31.4ms
INSERT 0 1
Returned in 36.9ms
INSERT 0 1
Returned in 30.7ms
INSERT 0 1
Returned in 32.4ms
INSERT 0 1
Returned in 32.4ms
INSERT 0 1
Returned in 36.7ms
INSERT 0 1
Returned in 35.8ms
INSERT 0 1
Returned in 32.0ms
INSERT 0 1
Returned in 30.9ms
INSERT 0 1
Returned in 35.3ms
INSERT 0 1
Returned in 36.6ms
INSERT 0 1
Returned in 29.8ms
INSERT 0 1
Returned in 32.5ms
INSERT 0 1
Returned in 33.0ms
```

ParkMyCar:
Thanks @ggevay for checking the benchmarks, totally agree keeping many small inserts fast is important!

Just pushed a commit which makes two changes:
1. Changes the `mz_sessions` write to use "background". Now that we are only blocking queries to `mz_sessions` and not all queries, I think using "background" priority is okay! This totally removes the possibility of many sessions causing lots of group commits.
2. [1] kind of makes the addition of `group_commit_batch_duration` orthogonal, so I reduced the group commit latency to 25ms by default!

ggevay:
Ok, thank you, this sounds good! I'll continue reviewing the code tomorrow.

src/adapter/src/coord/appends.rs
```diff
@@ -892,4 +942,171 @@ impl GroupCommitWaiter {
...
+///
+/// So as an optimization we do not wait for these writes to complete. But if a [`Session`] tries
+/// to query any of these builtin objects, we need to block that query on the writes completing to
+/// maintain linearizability.
+pub(crate) fn waiting_on_startup_appends(
+    catalog: &Catalog,
+    session: &mut Session,
+    plan: &Plan,
+) -> Option<(BTreeSet<CatalogItemId>, BoxFuture<'static, ()>)> {
+    let depends_on = match plan {
```
ggevay, line 995: I think we also need to call `transitive_uses` on these ids, in order to traverse referenced view / mat view definitions.

src/adapter/src/coord/appends.rs
```diff
@@ -892,4 +942,171 @@ impl GroupCommitWaiter {
...
+        | Plan::AlterSystemReset(_)
+        | Plan::AlterSystemResetAll(_)
+        | Plan::AlterRole(_)
+        | Plan::AlterOwner(_)
+        | Plan::AlterTableAddColumn(_)
+        | Plan::Declare(_)
+        | Plan::Fetch(_)
+        | Plan::Close(_)
+        | Plan::Prepare(_)
+        | Plan::Execute(_)
```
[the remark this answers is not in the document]
ParkMyCar: We don't need to handle `Plan::Execute` here, the flow for prepared statements is:
1. [sequence_execute(...)](https://github.com/MaterializeInc/materialize/blob/main/src/adapter/src/coord/sequencer/inner.rs#L4409-L4426) creates a new "Portal" which essentially registers the query
2. We sent a [`Command::Execute` to the coordinator](https://github.com/MaterializeInc/materialize/blob/main/src/adapter/src/coord/sequencer.rs#L557-L567) which will then plan and sequence the prepared statement, hitting this code path. So we should be covered!

We also don't need to cover `Plan::Insert` because users can't write to builtin tables, and for the case of `INSERT INTO ... SELECT ... FROM ...`, `Plan::ReadThenWrite` is already covered at the top of this match statement :)

src/adapter/src/coord/appends.rs
ParkMyCar: I think we're okay for now since we don't support creating Views or MatViews on top of objects in the `mz_internal` schema, but I added a TODO about this and a soft assert to make sure all the tables we depend on are in `mz_internal`.

Happy to make the change anyways if you'd prefer, although initially it feels a bit bad looking up all of the transitive IDs when we know it won't matter?

ggevay: Ok, this is fine for now! Thanks for adding the assert!

Btw. there is a third case in which this could go wrong besides the cases mentioned in the added comment: We might at some point create a _builtin_ view on top of `mz_sessions` (which _is_ allowed), and a user might query that builtin view. But I've checked that we currently don't have any builtin that would depend on `mz_sessions`.

src/adapter/src/coord/appends.rs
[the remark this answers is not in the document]
ParkMyCar: No worries at all, thank you for being thorough!

src/adapter/src/coord/appends.rs
ParkMyCar: Ahh good callout! I'll keep that in mind 👍

PR #31314, opened 2025-02-06
[adapter] use an unconsolidated snapshot in table bootstrapping

ParkMyCar:
_Pared down version of_ https://github.com/MaterializeInc/materialize/pull/31300

When the Coordinator starts we bootstrap tables by first retracting their entire contents so we have a clean slate, and then re-emitting all of the updates. We've seen in some recent releases large memory spikes on Coordinator startup related to creating a consolidated snapshot, and I believe it's this table bootstrapping that is causing the issue.

We don't need a consolidated snapshot to retract the contents, thus this API uses Persist's `snapshot_and_stream` API which should use a smaller bound of memory, specifically it keeps one `Part` in memory at a time, and parts are capped at approximately 128MiB.

### Motivation

Fix memory spike in `envd` startup

### Checklist

### DIFF
--- a/src/adapter/src/coord.rs
+++ b/src/adapter/src/coord.rs
@@ -2462,24 +2462,34 @@ impl Coordinator {
             let current_contents_fut = self
                 .controller
                 .storage_collections
-                .snapshot(system_table.table.global_id_writes(), read_ts);
+                .snapshot_and_stream(system_table.table.global_id_writes(), read_ts);
             // Fetch a snapshot of the current tables concurrently.
             let task = spawn(|| format!("snapshot-{table_id}"), async move {
-                let current_contents = current_contents_fut
+                let contents_stream = current_contents_fut
                     .await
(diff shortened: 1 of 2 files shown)


bkirwi:
At a first blush I'd expect this to make memory use worse, not better!

This may only fetch one part file at once... but it keeps the entire unconsolidated decoded snapshot in memory, whereas the previous version kept only the consolidated data. For most shards this seems like a bad tradeoff: if the unconsolidated data is big enough to cause a lot of memory use in the consolidator, it seems likely to cause about as much stored as a vec of rows. And the vec of rows can't benefit from lgalloc...

Have we seen this improve memory use in practice?

ParkMyCar:
@bkirwi ahh good thinking! In a small scale test I ran locally it did appear as though memory usage decreased, but your reasoning totally makes sense. I was able to figure out the `txn-wal` issue in https://github.com/MaterializeInc/materialize/pull/31300 so I'm going to close this PR and switch back to that one

PR #31317, opened 2025-02-06
[persist] Scale the size of the in-mem blob cache based on number of threads

ParkMyCar:
This PR introduces a dyncfg that allows us to tune the size of the in-mem blob cache based on the number of threads in the current process. Specifically it adds two new flags:
* `persist_blob_cache_scale_with_threads: bool` - enables this new sizing. default is `false`
* `persist_blob_cache_scale_factor_bytes: usize` - scaling factor, the number of bytes we'd size for the cache, pre-thread in the process. default is `32MiB`

The default scaling factor is less than the current static size of the cache (128MiB) so to prevent regressing the size of the cache on small clusters we use the static size as a minimum.

### Motivation

No concrete ask but seems like a useful knob to have

### Checklist

### DIFF
--- a/misc/python/materialize/mzcompose/__init__.py
+++ b/misc/python/materialize/mzcompose/__init__.py
@@ -148,6 +148,7 @@ def get_default_system_parameters(
         "persist_use_critical_since_snapshot": "false" if zero_downtime else "true",
         "persist_use_critical_since_source": "false" if zero_downtime else "true",
         "persist_part_decode_format": "row_with_validate",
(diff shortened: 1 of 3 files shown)


bkirwi (approved):
I'm still not clear on whether / when we'd want these, pending testing... but shouldn't hurt to have them in the pocket!

ParkMyCar:
> I'm still not clear on whether / when we'd want these, pending testing... but shouldn't hurt to have them in the pocket!

Same, but seems useful for experimentation! TFTR!

PR #31395, opened 2025-02-07
Implement MzArrange for column containers

antiguru:
Provide an implementation for MzArrange for collections and streams backed
by Column. This is strictly better than calling `mz_exchange_core`
wherever we need to arrange columnar data.

### Checklist

### DIFF
--- a/src/compute/src/extensions/arrange.rs
+++ b/src/compute/src/extensions/arrange.rs
@@ -7,16 +7,19 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::hash::Hash;
 use std::rc::Rc;
 
+use columnar::Columnar;
 use differential_dataflow::difference::Semigroup;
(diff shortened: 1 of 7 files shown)


frankmcsherry:
Still to read, but: this is starting to get close to scary moments where we need to make sure that hashes of things like `Row` and `Ref<'a>` actually match. I'd be inclined to see if we can externalize this a bit, *not* implement `Hash`, and instead require a function that comes from a common interface (e.g. as bytes, or as a datum sequence, or something).

antiguru:
Yes, the `Hash` implementation is a concern. We document Row's `Hash` implementation to defer to `RowRef` for exactly this reason. This won't work with `DatumSeq` in the long run, because it can have a fundamentally different way to represent data.

https://github.com/MaterializeInc/materialize/blob/385305b8d137335141c8dbdb6563480396ea6604/src/repr/src/row.rs#L296-L301

PR #31401, opened 2025-02-07
compute/correction-v2: introduce a Stage to speed up small inserts

teskje:
This PR extends the `CorrectionV2` data structure with a `Stage` that accumulates inserted updates before they get inserted into the sorted chains. This significantly reduces the amount of chain merges that need to be done in workloads that trickle in large amounts of updates in small batches, greatly speeding up these workloads.

For some reason, the memory tests introduced in https://github.com/MaterializeInc/materialize/pull/31267 are such a workload. Brief testing reveals that this PR speeds up the MV hydration in the `-800mb` test, when pointed against `CorrectionV2`, by roughly an order of magnitude, from ~300s to ~30s.

Note that this change doesn't have much impact when running locally on macOS. For some reason the batch sizes entering the MV sink are much smaller when running in `mzcompose`, not sure why that is.

### Motivation

Part of https://github.com/MaterializeInc/database-issues/issues/8464

### Checklist

### DIFF
--- a/misc/python/materialize/mzcompose/__init__.py
+++ b/misc/python/materialize/mzcompose/__init__.py
@@ -96,6 +96,7 @@ def get_default_system_parameters(
         "enable_alter_swap": "true",
         "enable_columnation_lgalloc": "true",
         "enable_compute_chunked_stack": "true",
+        "enable_compute_correction_v2": "true",
         "enable_connection_validation_syntax": "true",
         "enable_continual_task_builtins": (
             "true" if version > MzVersion.parse_mz("v0.127.0-dev") else "false"
--- a/src/compute/src/sink/correction_v2.rs
+++ b/src/compute/src/sink/correction_v2.rs
@@ -63,6 +63,13 @@
 //! A batch of updates is appended as a new chain. Then chains are merged at the end of the chain
 //! list until the chain invariant is restored.
 //!
+//! Inserting an update into the correction buffer can be expensive: It involves allocating a new
+//! chunk, copying the update in, and then likely merging with an existing chain to restore the
+//! chain invariant. If updates trickle in in small batches, this can cause a considerable
+//! overhead. The amortize this overhead, new updates aren't immediately inserted into the sorted
+//! chains but instead stored in a [`Stage`] buffer. Once enough updates have been staged to fill a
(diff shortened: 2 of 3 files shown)


antiguru (commented):
The implementation looks very similar to a `ConsolidatingContainerBuilder`, is there a reason why it couldn't reuse the implementation? It has a slightly different API, but maybe it doesn't need to. In Timely, a good rule-of-thumb is that during an operator invocation, it's great to leave state across input batches, but once the operator ceases to work, it's better to finish pending work. The builder API gives you this in the form of the `extract` and `finish` API, but it's harder to track this with the APIs implemented by `Stage`.

Following the builder pattern, the `Stage` wouldn't need to hook into logic to account for size and be aware of frontiers.

teskje:
> The implementation looks very similar to a `ConsolidatingContainerBuilder`, is there a reason why it couldn't reuse the implementation?

Looks like the `ConsolidatingContainerBuilder` produces containers up to a preferred size, which makes it awkward to use in the context of the correction buffer. For example, in the current implementation if the `Stage` gets enough data to fill two chunks at once, it can consolidate all this data together and then build a chain out of two chunks directly. Whereas with the `ConsolidatingContainerBuilder` I think I would get two consolidated `Vec`s and to build a chain from them I'd have to merge them into one `Vec` and then consolidate them again. So we'd spend more time consolidating than we do now.

> it's great to leave state across input batches, but once the operator ceases to work, it's better to finish pending work

Why would it be preferable to finish pending work when the operator becomes idle?

Keep in mind that the correction buffer is used in an async operator which, afaick, doesn't have a good way to know when it will become idle. It just does `loop { select! { ... } }` on a bunch of input channels. We could probably twiddle that control flow to do something special when we find out that non of the channels is ready, but that would make the code more complex, so we should have a good reason for doing so.

teskje:
I tested using the `ConsolidatingContainerBuilder` instead in https://github.com/MaterializeInc/materialize/pull/31456. It reduces the amount of code, which is nice. But for some reason it also has way worse performance, compared to the hand-rolled `Stage` implementation in this PR.

antiguru (approved):
LGTM, let's take the discussion around the consolidating container builder offline!

teskje:
TFTR!

PR #31406, opened 2025-02-07
Temporal buckets

antiguru:
Because, who doesn't like buckets?

WIP

### Checklist


src/timely-util/src/temporal.rs
```diff
@@ -0,0 +1,328 @@
...
+    /// element contains all data with a timestamp strictly less than `timestamp`, and the second
+    /// all other data.
+    fn split(self, timestamp: &Self::Timestamp, fuel: &mut isize) -> (Self, Self);
+}
+
+/// A sorted list of buckets, representing data bucketed by timestamp.
+///
+/// Bucket chains support three main APIs: finding buckets for a given timestamp, peeling
+/// off buckets up to a frontier, and restoring the chain property. All operations aim to be
+/// amortized logarithmic in the number of outstanding timestamps.
```
teskje, line 44: Another data structure with logarithmic complexities in these operations (insert, find, remove) is the b-tree. What advantages does the bucket chain provide over a b-tree?

src/timely-util/src/temporal.rs
```diff
@@ -0,0 +1,328 @@
...
+        assert!(collect_and_sort(peeled).into_iter().eq(63..65));
+        let peeled = chain.peel(AntichainRef::new(&[]));
+        let mut fuel = 1000;
+        chain.restore(&mut fuel);
+        assert!(fuel > 0);
+        assert!(collect_and_sort(peeled).into_iter().eq(65..=255));
+    }
+
+    /// Test a chain with 10M disjoint elements. The same with a vector would take too long for a
+    /// test.
```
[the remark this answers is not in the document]
antiguru: (Removing the comment) I took a vector with the same elements, and repeatedly removed the front elements up to a specific point. The cost of moving elements around becomes substantial at some point.

src/repr/Cargo.toml
```diff
@@ -56,6 +56,7 @@ mz-persist-types = { path = "../persist-types", default-features = false }
 mz-pgtz = { path = "../pgtz", default-features = false }
 mz-proto = { path = "../proto", features = ["chrono"], default-features = false }
 mz-sql-parser = { path = "../sql-parser", default-features = false }
+mz-timely-util = { path = "../timely-util", default-features = false }
```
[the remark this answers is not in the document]
antiguru: The implementation of `BucketTimestamp` for `Timestamp` can only live in `mz-timely-util` if it depends on `mz-repr`, which seems like an even bigger surprise---the crate specifically doesn't depend on anything in Mz due to its license. Because of Rust's orphan rules, the only place for the implementation to live is next to the type itself. Does that make sense?

src/timely-util/src/temporal.rs
antiguru: Yup, I agree. The implementation here just came together very naturally and now needs little dependencies.

src/repr/Cargo.toml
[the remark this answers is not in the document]
teskje: Huh, I didn't realize we had a special license for `mz-timely-util`! What's the reason for that?

I guess it's fine. I view `mz-repr` as a crate that provides fundamental data structures and should be a leaf dependency, but it already depends on `mz-sql-parser`, so 🤷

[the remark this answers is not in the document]
antiguru: `mz-timely-util` is code that we might want to upstream, and it's much easier if it doesn't need to go through a license change.

misc/python/materialize/parallel_workload/action.py
```diff
@@ -1130,6 +1130,7 @@ def __init__(
             "enable_compute_replica_expiration",
             "compute_replica_expiration_offset",
             "enable_compute_render_fueled_as_specific_collection",
+            "enable_compute_temporal_bucketing",
```
def-, line 1140: What happens if we change the setting during a test run, could that be interesting to test?

antiguru: It only applies when rendering a dataflow. Would it be interesting to toggle it in this case?

def-: I think so

src/timely-util/src/temporal.rs
antiguru: I switched the implementation to a `BTreeMap` fwiw.

src/compute/src/extensions/temporal_bucket.rs
```diff
@@ -0,0 +1,215 @@
...
+use timely::container::{ContainerBuilder, PushInto};
+use timely::dataflow::channels::pact::Pipeline;
+use timely::dataflow::operators::Operator;
+use timely::dataflow::{Scope, StreamCore};
+use timely::progress::{Antichain, Timestamp};
+use timely::{Data, PartialOrder};
+
+use crate::typedefs::MzData;
+
+/// Sort outstanding updates into a [`BucketChain`], and reveal data not in advance of the frontier.
```
teskje: I was confused by the "reveal data not in advance of the frontier" (here and below), because I first thought "frontier" here means "output frontier". But it means "input frontier", right? Good to make that explicit in the comments.

src/ore/Cargo.toml
```diff
@@ -94,7 +94,7 @@ hyper-util = { version = "0.1.14", optional = true }
 opentelemetry = { version = "0.24.0", features = ["trace"], optional = true }
 opentelemetry-otlp = { version = "0.17.0", optional = true }
 opentelemetry_sdk = { version = "0.24.1", features = [
-    "rt-tokio",
+  "rt-tokio",
 ], optional = true }
```
teskje: Nit: I'd prefer to remove the line breaks instead

test/testdrive/replica-expiration.td
```diff
@@ -24,9 +24,12 @@ ALTER SYSTEM SET compute_replica_expiration_offset = '30d'
   WHERE mz_now() <= event_ts + INTERVAL '20 days';
 > CREATE DEFAULT INDEX ON events_view;
 > INSERT INTO events SELECT x::text, now() FROM generate_series(1, 1000) AS x;
+# TODO: The following query should return 2000, but it returns 1000 because the
+# the arrangement sizes does not account for the temporal bucket. It is part of
+# a different operator.
```
teskje: I thought this would work, since the temporal bucket operator uses the MergeBatcher under the hood and that produces record counts as `BatcherEvent`s. Are we losing those somewhere along the way?

src/timely-util/src/temporal.rs
```diff
@@ -0,0 +1,328 @@
...
+use timely::progress::Timestamp;
+use timely::progress::frontier::AntichainRef;
+
+/// Timestamp extension for timestamps that can advance by `2^exponent`.
+///
+/// Most likely, this is only relevant for totally ordered timestamps.
+pub trait BucketTimestamp: Timestamp {
+    /// The number of bits in the timestamp.
+    const DOMAIN: usize = size_of::<Self>() * 8;
+    /// Advance this timestamp by `2^exponent`.
```
teskje: Can you also mention when this should return `None`?

src/timely-util/src/temporal.rs
```diff
@@ -0,0 +1,328 @@
...
+    fn advance_by_power_of_two(&self, exponent: u32) -> Option<Self>;
+}
+
+/// A type that can be split into two parts based on a timestamp.
+pub trait Storage: Sized {
+    /// The timestamp type associated with this storage.
+    type Timestamp: BucketTimestamp;
+    /// Split self in two, based on the timestamp. The result a pair of self, where the first
```
teskje:
```suggestion
    /// Split self in two, based on the timestamp. The result is a pair of self, where the first
```

src/timely-util/src/temporal.rs
```diff
@@ -0,0 +1,328 @@
...
+
+/// A sorted list of buckets, representing data bucketed by timestamp.
+///
+/// Bucket chains support three main APIs: finding buckets for a given timestamp, peeling
+/// off buckets up to a frontier, and restoring the chain property. All operations aim to be
+/// amortized logarithmic in the number of outstanding timestamps.
+///
+/// We achieve this by storing buckets of increasing size for timestamps that are further out
+/// in the future. At the same time, in a well-formed chain, adjacent buckets span a time range at
+/// most factor 4 different.
```
teskje: How was 4 selected as the factor?

src/timely-util/src/temporal.rs
```diff
@@ -0,0 +1,328 @@
...
+}
+
+/// A type that can be split into two parts based on a timestamp.
+pub trait Storage: Sized {
+    /// The timestamp type associated with this storage.
+    type Timestamp: BucketTimestamp;
+    /// Split self in two, based on the timestamp. The result a pair of self, where the first
+    /// element contains all data with a timestamp strictly less than `timestamp`, and the second
+    /// all other data.
+    fn split(self, timestamp: &Self::Timestamp, fuel: &mut isize) -> (Self, Self);
```
teskje: `isize` seems like a strange type for fuel. I assume you want a signed integer so when we overshoot our available fuel, the call can give us less fuel next time. But there is not relationship to memory sizes, so `i64` seems better suited.

src/timely-util/src/temporal.rs
```diff
@@ -0,0 +1,328 @@
...
+    pub fn len(&self) -> usize {
+        self.content.len()
+    }
+
+    /// Split the bucket specified by `(bits, offset, storage)` and insert the new buckets.
+    /// Updates `fuel`.
+    ///
+    /// Panics if the bucket cannot be split, i.e, it covers 0 bits.
+    #[inline(always)]
+    fn splice(&mut self, fuel: &mut isize, bits: u32, offset: S::Timestamp, storage: S) {
```
teskje: Why is this called "splice"? Doesn't that mean more or less the opposite of "split"? 😅

src/timely-util/src/temporal.rs
```diff
@@ -0,0 +1,328 @@
...
+        assert!(chain.is_empty());
+    }
+
+    #[mz_ore::test]
+    fn test_bucket_chain_u8() {
+        let mut chain = BucketChain::new(TestStorage::<u8> {
+            inner: (0..=255).collect(),
+        });
+        let mut fuel = -1;
+        while fuel < 0 {
```
teskje: Doesn't this need to be `while fuel <= 0`?

src/timely-util/src/temporal.rs
```diff
@@ -0,0 +1,328 @@
...
+/// Most likely, this is only relevant for totally ordered timestamps.
+pub trait BucketTimestamp: Timestamp {
+    /// The number of bits in the timestamp.
+    const DOMAIN: usize = size_of::<Self>() * 8;
+    /// Advance this timestamp by `2^exponent`.
+    fn advance_by_power_of_two(&self, exponent: u32) -> Option<Self>;
+}
+
+/// A type that can be split into two parts based on a timestamp.
+pub trait Storage: Sized {
```
teskje: Nit: Consider calling this `Bucket`. I think that would make the usage of this trait more obvious, given that it's a part of a `BucketChain` and contains `BucketTimestamp`s.

teskje (approved):
LGTM, I only have minor comments.

src/timely-util/src/temporal.rs
antiguru: `i64` it is!

src/timely-util/src/temporal.rs
antiguru: Changed to `split_and_insert`.

src/timely-util/src/temporal.rs
antiguru: Added some explanation, but we might want to do the math to figure out what the best factor actually is. I don't recall the exact reasoning :/

test/testdrive/replica-expiration.td
antiguru: This is https://github.com/MaterializeInc/database-issues/issues/9455

antiguru:
Thanks for the reviews!

PR #31407, opened 2025-02-07
kafka + avro: Add a test for subject/schema mismatch

martykulma:
I didn't see any tests that cover subject/schema mismatches.  One of the failure scenarios that came out of looking at [8933](https://github.com/MaterializeInc/database-issues/issues/8933) is this type of mismatch. Though I'm not sure the best way to incorporate this test (or how likely it is, so it might not be worth the effort).

In order for this test to run reliably, it needs 2 subjects with compatible schemas.  We publish a message to a topic, but reference a different schema id in the header.  The existing tests keep reusing the schema registry, so the schema id is nondeterministic as the schema ids keep increasing with each schema registered.

options seem to be:
- reset the schema registry (i.e. drop and recreate kafka + registry) so that we are highly confident of the schema ids
- add an action to retrieve the id from schema registry (in the same vein as `set-from-sql`)

for now, it lives in a separate subdirectory so default doesn't try to run it (as it might fail with a different type of error, depending on what schema id 2 ends up being)

```
 % bin/mzcompose --find testdrive-old-kafka-src-syntax down && bin/mzcompose --find testdrive-old-kafka-src-syntax run default isolated/avro-resolution-schema-subject-mismatch.td
```

### Motivation

coverage for a subject/schema mismatch

### Checklist

### DIFF
--- a/test/testdrive-old-kafka-src-syntax/mzcompose.py
+++ b/test/testdrive-old-kafka-src-syntax/mzcompose.py
@@ -175,7 +175,9 @@ def workflow_default(c: Composition, parser: WorkflowArgumentParser) -> None:
     matching_files = []
     for files in args.files:
         matching_files.extend(
-            glob.glob(files, root_dir=MZ_ROOT / "test" / "testdrive-old-kafka-src-syntax")
+            glob.glob(
+                files, root_dir=MZ_ROOT / "test" / "testdrive-old-kafka-src-syntax"
+            )
         )
 
     # test files that need to have the schema registry reset
@@ -237,7 +239,6 @@ def workflow_default(c: Composition, parser: WorkflowArgumentParser) -> None:
             )
 
         junit_report = ci_util.junit_report_filename(c.name)
-
 
         try:
             junit_report = ci_util.junit_report_filename(c.name)


def-:
Testdrive already has logic to clean up Kafka state (when run without `--no-reset`) in
https://github.com/MaterializeInc/materialize/blob/109406ce5137a01d2c98b124f5bd45b8c3f4dea0/src/testdrive/src/lib.rs#L109-L124
So my suggestion would be to extend that with the schema registry so that it is reset properly too. Looking more closely I found
https://github.com/MaterializeInc/materialize/blob/109406ce5137a01d2c98b124f5bd45b8c3f4dea0/src/testdrive/src/lib.rs#L109-L124
So this should already be reset? I guess the index doesn't start from 1 even after cleaning? I looked around a bit and couldn't find a way to solve this.

A second idea would be to have the id available somewhere in Materialize via SQL, but I couldn't find any existing view.

Third idea is to just leave it in an isolated folder and create a separate workflow that fully resets the state after each test, but I'd prefer to avoid that.

def-:
I was afraid that was the case. So that means to go with option 3 with the `isolated` directory and running those in actual isolation I guess? I can write that up tomorrow.

PR #31408, opened 2025-02-07
[storage] Atomic alter table in the storage controller

bkirwi:
This avoids a possible race in the storage controller, where a downgrade of our read capability can race the selection of new capabilities for a new version of the table. (More detail in a comment downthread.)

We instead switch to:
- using the frontiers of the old version directly;
- doing the frontier selection and the state update atomically while holding the lock;

...which together should avoid this particular race. 

### Motivation

https://github.com/MaterializeInc/database-issues/issues/8952

### Tips for reviewer

This bug is quite tricky to reproduce, so I'm not 100% confident that this is the issue we're seeing in CI. Happy to take suggestions for improving the tests! But also please check my logic below: that it sounds like an accurate description of the current behaviour, and that the particular interleaving of events it imagines would explain the symptoms we see.

### Checklist

### DIFF
--- a/src/storage-client/src/storage_collections.rs
+++ b/src/storage-client/src/storage_collections.rs
@@ -2232,7 +2232,7 @@ where
         new_desc: RelationDesc,
         expected_version: RelationVersion,
     ) -> Result<(), StorageError<Self::Timestamp>> {
-        let (data_shard, existing_write_frontier, existing_read_capabilities) = {
+        let data_shard = {
             let self_collections = self.collections.lock().expect("lock poisoned");
             let existing = self_collections
                 .get(&existing_collection)
@@ -2243,11 +2243,7 @@ where
                 return Err(StorageError::IdentifierInvalid(existing_collection));
             }
 
-            (
-                existing.collection_metadata.data_shard,
-                existing.write_frontier.clone(),
-                existing.read_capabilities.clone(),
-            )
+            existing.collection_metadata.data_shard
         };
 
         let persist_client = self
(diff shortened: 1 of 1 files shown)


bkirwi:
# A hypothesis

At the storage collections level, alter table is roughly:

1. Grab a snapshot of the frontiers of the previous version, as known to the storage controller.
2. Evolve the schema of the shard.
3. Obtain new handles for the shard. Notably, these frontiers are not advanced: they’re based on whatever’s currently durable in Persist.
4. Generate new metadata for the collection. The initial since hold of this collection is:
    1. the current critical since of the shard. (Read from Persist.)
    2. the overall frontier of the read capabilities obtained in step 1.
5. Atomically modify the controller state: add the new version and point the old version to it. Among other things, this means that only changes to the new version’s frontiers get sent to the background worker and synced back to Persist.

Observations:

- The frontier we get in 3 will generally be *behind* what we got in 1: since we only write sinces back to Persist periodically. (And in any case it would be impossible to update the durable state *instantly* once we were ready to downgrade the frontier, for distributed-systems reasons.)
- However, it’s also possible for the frontier we get in 3 to be *ahead*, since this whole setup is not atomic, and we might have downgraded the critical handle in the meantime.

So: suppose we `ALTER TABLE` right as we downgrade the since of our table from t=10 to t=20, enqueueing but not applying a downgrade. In that case, we might see t=20 in step 1 but t=10 in step 3. Then, as we go to write the new metadata (with an overall since of t=10) into our storage collections state. At this point, the background worker might finally apply the downgrade, making t=10 permanently unreadable even though the storage controller still thinks it can issue handles at that frontier.

aljoscha (approved):
Yeah, I think this can happen! At first I was skeptical because I thought re-acquiring the since handle would protect us, because I mistakenly thought that someone somewhere would have to fail because of a mismatched since when trying to downgrade. _But_, we only compare against the epoch: when downgrading a critical since handle we don't check that the actual since _frontier_ is what we expect. So in the alter table situation it can (does) happen that there are two valid critical since handles in the process that can downgrade the since (the one we freshly acquired, and the one in the background worker), and they both map to the one logical since handle (the one since handle ID).

Very nice find! 🙌

bkirwi:
Great! So far the only nightly errors are either also present on main, or seem like flakes - since they don't alter any tables, so this code wouldn't run - but I'll try and confirm that before merging.

PR #31409, opened 2025-02-07
helm-chart: Replace metrics-server kubectl install with Helm

bobbyiliev:
### Motivation

In the local test instructions, updating the steps for the metrics-server to simplify installation and configuration by using Helm instead of manual kubectl apply and deployment edits.

### DIFF
--- a/misc/helm-charts/testing/README.md
+++ b/misc/helm-charts/testing/README.md
@@ -54,32 +54,39 @@ After adding the labels, verify that they were successfully applied by running t
 kubectl get nodes --show-labels
 ```
 
-## Metrics service
+## (Optional) Metrics service
 
-The metrics service is required for the `environmentd` pod to function correctly.
+Materialize does not require the Kubernetes metrics service to function. However, if `observability.enabled: true` and `observability.podMetrics.enabled: true` are set in the Helm values file, `environmentd` would expect a running metrics service. Without it, and with `observability` disabled, Materialize will still operate, but metrics will not be available in the web console.
 
+For more information, see the [Metrics Server documentation](https://github.com/kubernetes-sigs/metrics-server/blob/master/README.md). To install the metrics server, follow these steps:
+
+First, add the metrics-server Helm repository:
+
+```bash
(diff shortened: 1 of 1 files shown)


misc/helm-charts/testing/README.md
```diff
@@ -56,32 +56,37 @@ kubectl get nodes --show-labels
 
 ## Metrics service
 
-The metrics service is required for the `environmentd` pod to function correctly.
+The metrics service is required for the `environmentd` pod to function correctly. For more information, see the [Metrics Server documentation](https://github.com/kubernetes-sigs/metrics-server/blob/master/README.md).
```
kay-kim, line 59: Q: If this is required for the environmentd to function correctly, should we update the Chart's values.yaml file to have the observability.podMetrics.enabled be true?  Or mention that we should set it to true?

kay-kim (approved):
LGTM - thanks again!

misc/helm-charts/testing/README.md
bobbyiliev: Ah yes, good point! This was actually true in the past, now if the `observability` is disabled in the Helm values file, envd will still work. It will only fail if the `observability` is enabled but the metrics service is missing. I've added a bit of clarification for this. Let me know what you think!

misc/helm-charts/testing/README.md
```diff
@@ -56,32 +56,37 @@ kubectl get nodes --show-labels
...
+  --namespace kube-system \
+  --set args="{--kubelet-insecure-tls,--kubelet-preferred-address-types=InternalIP\,Hostname\,ExternalIP}"
 ```
 
+As this is a local testing environment, we are disabling TLS for the metrics server to avoid issues with the `environmentd` pod. This is not recommended for production environments.
+
 Get the metrics server pod status:
 
-```sh
+```bash
```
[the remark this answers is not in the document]
bobbyiliev: Neat! Great idea!

kay-kim (approved):

misc/helm-charts/testing/README.md
kay-kim: Looks great! Thanks for the clarification.

PR #31451, opened 2025-02-10
mz-ore: conditional compilation for chrono

martykulma:
Adds conditional compilation for `chrono` in `mz-ore` crate.

### Motivation

cargo buid/test failures
```
error[E0432]: unresolved import `chrono`
  --> src/ore/src/panic.rs:31:5
   |
31 | use chrono::Utc;
   |     ^^^^^^ use of undeclared crate or module `chrono`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `mz-ore` (lib) due to 1 previous error
```

### Checklist

### DIFF
--- a/src/ore/src/panic.rs
+++ b/src/ore/src/panic.rs
@@ -28,6 +28,7 @@ use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use std::{env, thread};
 
+#[cfg(feature = "chrono")]
 use chrono::Utc;
 use itertools::Itertools;
 #[cfg(feature = "async")]
@@ -86,7 +87,12 @@ pub fn install_enhanced_handler() {
             return;
         }
 
-        let timestamp = Utc::now().format("%Y-%m-%dT%H:%M:%S%.6fZ").to_string();
+        // can't use if cfg!() here because that will require chrono::Utc import
+        #[cfg(feature = "chrono")]
+        let timestamp = Utc::now().format("%Y-%m-%dT%H:%M:%S%.6fZ ").to_string();
+        #[cfg(not(feature = "chrono"))]
+        let timestamp = String::new();
+
         let thread = thread::current();
         let thread_name = thread.name().unwrap_or("<unnamed>");
 
@@ -142,7 +148,7 @@ pub fn install_enhanced_handler() {
         //
         // See https://github.com/rust-lang/rust/issues/64413 for details.
         let buf = format!(
-            "{timestamp}  thread '{thread_name}' panicked at {location}:\n{msg}\n{backtrace}"
+            "{timestamp}thread '{thread_name}' panicked at {location}:\n{msg}\n{backtrace}"
         );
 
         // Ideal path: spawn a thread that attempts to lock the Rust-managed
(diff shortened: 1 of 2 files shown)


src/ore/src/panic.rs
```diff
@@ -86,7 +87,12 @@ pub fn install_enhanced_handler() {
             return;
         }
 
-        let timestamp = Utc::now().format("%Y-%m-%dT%H:%M:%S%.6fZ").to_string();
+        // can't use if cfg!() here because that will require chrono::Utc import
+        #[cfg(feature = "chrono")]
+        let timestamp = Utc::now().format("%Y-%m-%dT%H:%M:%S%.6fZ ").to_string();
```
doy-materialize, line 92: not sure if this is relevant but it does change the output from two spaces to one space in between the timestamp and the message

doy-materialize (approved):
looks good aside from that one nit

src/ore/src/panic.rs
```diff
@@ -86,7 +87,12 @@ pub fn install_enhanced_handler() {
             return;
         }
 
-        let timestamp = Utc::now().format("%Y-%m-%dT%H:%M:%S%.6fZ").to_string();
+        // can't use if cfg!() here because that will require chrono::Utc import
+        #[cfg(feature = "chrono")]
+        let timestamp = Utc::now().format("%Y-%m-%dT%H:%M:%S%.6fZ  ").to_string();
+        #[cfg(not(feature = "chrono"))]
+        let timestamp = String::new();
```
def-, line 94: Can we make this non-optional? I think we should always have timestamps in the panics.

def- (approved):
Oops, sorry about that. Shouldn't this have failed in CI?

src/ore/src/panic.rs
doy-materialize: my understanding is that mz-ore is not supposed to have non-optional dependencies?

teskje: I agree with Dennis that this should be non-optional. Timestamps in panics are really useful and in the current state it's very easy to forget enabling `chrono` when depending on mz-ore to install the panic hook. I think what we could do instead is gate `install_enhanced_handler` behind a `#[cfg(feature = "chrono")]`. Then everyone who wants to use the handler needs to enable `chrono`, but nobody else does.

teskje: > Wouldn't people still have to remember to enable it?

You wouldn't have to remember because you'd get a compile error if you tried to use `install_enhanced_handler` but didn't have `chrono` enabled!

> If your goal is to always have it, then we should default to something. What about `std::time::SystemTime` in the non-chrono case?

That's better than nothing, but there might be regexes parsing logs in CI/cloud and those might get confused if the timestamp format for panics is different than the timestamp format for other logs.

Would we gain anything from allowing to use `install_enhanced_handler` without `chrono`? At least in the database I don't think we are too worried about including it because ~everything already depends on it anyway, directly or indirectly. Not sure about crates in the cloud repo.

doy-materialize: i don't think renaming it is necessary - feel free to do whatever makes the most sense here and i'm happy to handle updating the cloud repo once the changes land

teskje: Yeah, I think this should be as easy as (a) slapping the `#[cfg]` on `install_enhanced_handler` and (b) adding the `chrono` feature to the `mz-ore` dependency declarations that need it in this repo (which you can find by following compile errors). And then later, when cloud bumps the vendored version of the code do (b) also for the cloud crates.

PR #31454, opened 2025-02-10
[persist] Remove the "gone to sleep" panics

bkirwi:
- Include more debugging info in the handle-expiry message.
- Remove a panic when attempting to downgrade the since of a handle that is expired.

### Motivation

https://github.com/MaterializeInc/incidents-and-escalations/issues/170

### Tips for reviewer

Silencing the error here might be controversial? My thinking is roughly:
- In theory: semantically, we generally treat an expired handle as one that has had its frontier advanced all the way, so this is in line with some existing behaviour.
- In practice: this should never lead to incorrectness that wasn't already possible, because extending the lease never guaranteed that it would be present for any particular future operation... just more likely. It does increase the risk that we'll hit some other panic down the line, but that should only happen in cases where this would have panicked anyways... so this should be a net stability benefit.

### Checklist

### DIFF
--- a/src/persist-client/src/internal/state.rs
+++ b/src/persist-client/src/internal/state.rs
@@ -2305,19 +2305,26 @@ where
     pub fn expire_at(&mut self, walltime_ms: EpochMillis) -> ExpiryMetrics {
         let mut metrics = ExpiryMetrics::default();
         let shard_id = self.shard_id();
-        self.collections.leased_readers.retain(|k, v| {
-            let retain = v.last_heartbeat_timestamp_ms + v.lease_duration_ms >= walltime_ms;
+        self.collections.leased_readers.retain(|id, state| {
+            let retain = state.last_heartbeat_timestamp_ms + state.lease_duration_ms >= walltime_ms;
             if !retain {
-                info!("Force expiring reader ({k}) of shard ({shard_id}) due to inactivity");
+                info!(
+                    "Force expiring reader {id} ({}) of shard {shard_id} due to inactivity",
+                    state.debug.purpose
+                );
                 metrics.readers_expired += 1;
             }
             retain
         });
         // critical_readers don't need forced expiration. (In fact, that's the point!)
-        self.collections.writers.retain(|k, v| {
-            let retain = (v.last_heartbeat_timestamp_ms + v.lease_duration_ms) >= walltime_ms;
+        self.collections.writers.retain(|id, state| {
+            let retain =
+                (state.last_heartbeat_timestamp_ms + state.lease_duration_ms) >= walltime_ms;
             if !retain {
-                info!("Force expiring writer ({k}) of shard ({shard_id}) due to inactivity");
+                info!(
+                    "Force expiring writer {id} ({}) of shard {shard_id} due to inactivity",
+                    state.debug.purpose
+                );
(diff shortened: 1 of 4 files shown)


bkirwi:
Nightly run: https://buildkite.com/materialize/nightly/builds/11112

src/persist-client/src/internal/state.rs
```diff
@@ -1660,7 +1660,14 @@ where
...
         }
 
-        let reader_state = self.leased_reader(reader_id);
+        // The only way to have a missing reader in state is if it's been expired... and in that
+        // case, we behave the same as though that reader had been downgraded to the empty antichain.
+        let Some(reader_state) = self.leased_reader(reader_id) else {
+            tracing::warn!(
+                "Leased reader {reader_id} was expired due to inactivity. Did the machine go to sleep?",
+            );
+            return Break(NoOpStateTransition(Since(Antichain::new())));
```
ParkMyCar, line 1669: While this function does return the new Since, it looks like our caller doesn't check how the returned Since compares to the `new_since` it provided ([code](https://github.com/MaterializeInc/materialize/blob/main/src/persist-client/src/read.rs#L608-L634))?

Thinking out loud a bit here... I imagine theoretically this is okay, but this is what might result in panics elsewhere when someone can no longer read with their handle because the `since` has been advanced to the empty Antichain? Like you call out in the PR description I think this will lead to a net win in stability though, so it works for me!

src/persist-client/src/internal/state.rs
```diff
@@ -2310,19 +2305,26 @@ where
     pub fn expire_at(&mut self, walltime_ms: EpochMillis) -> ExpiryMetrics {
         let mut metrics = ExpiryMetrics::default();
         let shard_id = self.shard_id();
-        self.collections.leased_readers.retain(|k, v| {
-            let retain = v.last_heartbeat_timestamp_ms + v.lease_duration_ms >= walltime_ms;
+        self.collections.leased_readers.retain(|id, state| {
```
ParkMyCar, line 2308: 🙏

ParkMyCar (approved):

src/persist-client/src/internal/state.rs
bkirwi: Yeah, that all sounds right!

A slightly more invasive version would be to have an explicit "expired" state, and return a targeted error when folks try and snapshot on an expired handle. I think it mostly comes to the same thing, but may be more explicit... preference?

ParkMyCar: An explicit expired state sounds nice! Definitely don't think it's a blocker though, happy to make that change at a later date

bkirwi:
I did a bit more digging:
- The `since` field is set ~only by `downgrade_since`, and read ~only by the `since()` accessor.
  - So: to observe a different since frontier from this change, you'd need to call `downgrade_since` and then `since` on the same expired handle.
- `downgrade_since` is only called by the `Listen` machinery (which does not access the since of the handle it takes ownership of) and the read-only-mode controller (which does access the since, but is already designed to fail open if it loses its holds).

So I think as-is this should be low-impact, and designing an explicit "expired" state is a little trickier. I'll get this in for now to make sure it gets lots of coverage by the next release...

PR #31457, opened 2025-02-11
storage: pick largest possible `as_of` frontiers for ingestions

petrosagg:
Fixes https://github.com/MaterializeInc/database-issues/issues/8898
### Checklist

### DIFF
--- a/src/storage/src/storage_state/async_storage_worker.rs
+++ b/src/storage/src/storage_state/async_storage_worker.rs
@@ -267,15 +267,47 @@ impl<T: Timestamp + TimestampManipulation + Lattice + Codec64 + Display + Sync>
                         // amount of memory required to hydrate.
                         //
                         // For each output `o` and for each input `i` of the ingestion the
-                        // controller guarantees that i.since < o.upper. Therefore the largest
-                        // as-of for a particular output `o` is `{ t - 1 | t in o.upper }`.
+                        // controller guarantees that i.since < o.upper except when o.upper is
+                        // [T::minimum()]. Therefore the largest as-of for a particular output `o`
+                        // is `{ (t - 1).advance_by(i.since) | t in o.upper }`.
                         //
                         // To calculate the global as_of frontier we take the minimum of all those
                         // per-output as-of frontiers.
+                        let client = persist_clients
+                            .open(
+                                ingestion_description
+                                    .ingestion_metadata
+                                    .persist_location
+                                    .clone(),
+                            )
+                            .await
+                            .expect("error creating persist client");
+                        let read_handle = client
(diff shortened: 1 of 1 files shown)


src/storage/src/storage_state/async_storage_worker.rs
```diff
@@ -199,73 +201,12 @@ impl<T: Timestamp + Lattice + Codec64 + Display + Sync> AsyncStorageWorker<T> {
...
-                        // that upon restart, or upon observing the persist state like this
-                        // function, one can see non-sensical results like the since of A be in
-                        // advance of B even when B depends on A! This can happen because the
-                        // downgrade of B gets reordered and lost. Here is our best attempt at
-                        // playing detective of what the controller meant to do by blindly
-                        // assuming that the since of the remap shard is a suitable since
-                        // frontier without consulting the since frontier of the outputs. One
-                        // day we will enforce order to chaos and this comment will be deleted.
-                        let remap_shard = ingestion_description
+                        let seen_remap_shard = ingestion_description
```
teskje, line 206: Nit: The `seen_` prefix here confuses me as I can't guess what it means! It kind of indicates that this isn't necessarily the actual remap shard?

src/storage/src/storage_state/async_storage_worker.rs
```diff
@@ -317,6 +258,61 @@ impl<T: Timestamp + Lattice + Codec64 + Display + Sync> AsyncStorageWorker<T> {
...
+                        let remap_since = read_handle.since().clone();
+                        mz_ore::task::spawn(move || "deferred_expire", async move {
+                            tokio::time::sleep(std::time::Duration::from_secs(300)).await;
+                            read_handle.expire().await;
+                        });
+                        let mut as_of = Antichain::new();
+                        for upper in resume_uppers.values() {
+                            for t in upper.elements() {
+                                let mut t_prime = t.step_back().unwrap_or(T::minimum());
+                                if !remap_since.is_empty() {
```
teskje, line 309: This special case could use a comment! I think an empty since here means that the ingestion has already been dropped so we should return `as_of = []`? Would be good to spell that out a bit.

teskje (approved):
This looks reasonable to me but I'm lacking a ton of context. I think this would benefit from an @aljoscha review!

src/storage/src/storage_state/async_storage_worker.rs
petrosagg: I will find a better name, that variable existed before the PR, I just moved it around

src/storage/src/storage_state/async_storage_worker.rs
petrosagg: will do

PR #31463, opened 2025-02-11
Remove obsolete sink partition strategy

martykulma:
Refactors Kafka sink to only use v1 strategy for publishing.  This starts to remove the concept of a partitioning strategy version, which was use for migration from old (v0) to new (v1).

This is phase 1:
- retain ability to parse CREATE SINK ... WITH (PARTITION STRATEGY ...) statements
- default to v1 behavior regardless of PARTITION STRATEGY
- adds a migration to remove partition strategy on upgrade

Phase 2 will remove the migration code and the remaining bits.

### Motivation

Migration for existing users is completed, removing legacy code.

### Checklist

### DIFF
--- a/src/adapter/src/catalog/migrate.rs
+++ b/src/adapter/src/catalog/migrate.rs
@@ -18,6 +18,7 @@ use mz_ore::now::NowFn;
 use mz_persist_types::ShardId;
 use mz_repr::{CatalogItemId, Timestamp};
 use mz_sql::ast::display::AstDisplay;
+use mz_sql::ast::CreateSinkOptionName;
 use mz_sql::names::FullItemName;
 use mz_sql_parser::ast::{IdentError, Raw, Statement};
 use mz_storage_client::controller::StorageTxn;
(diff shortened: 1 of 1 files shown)


def- (approved):
Test changes lgtm. I like removing old code. I have triggered a nightly run (using https://trigger-ci.dev.materialize.com/): https://buildkite.com/materialize/nightly/builds/11125

def- (changes requested):
Nightly shows some fun panics after upgrading: https://buildkite.com/materialize/nightly/builds/11125
```
platform-checks-materialized-1  | 2025-02-12T18:26:47.574399Z  thread 'main' panicked at src/sql-parser/src/parser.rs:3198:18: internal error: entered unreachable code
```
Something like this might reproduce it, but haven't run it locally yet: `bin/mzcompose --find platform-checks down && bin/mzcompose --find platform-checks run default --scenario=UpgradeEntireMz --check=SinkUpsert`
I guess we persist the SQL with the old partition strategy, and then try to parse it in the new version during startup. We should probably still support parsing it so that we can migrate.

def-:
The reproducer I posted above works locally.

def- (commented):
New nightly triggered: https://buildkite.com/materialize/nightly/builds/11141

def- (approved):
The new nightly failures seem unrelated, we also have a similar set on main, that I'm looking into.

ParkMyCar (approved):
Sorry for the delay here @martykulma! Was catching up after some PTO.

I'm not super familiar with the history here, but I read through the [design doc](https://www.notion.so/materialize/Stable-partitioning-9e4b4f8113df45739bbf1f5d9a34a56e?pvs=4) and linked GitHub issue and everything you've done here makes sense to me! Let's chat in Slack about the rollout :)

PR #31466, opened 2025-02-12
cluster: ensure workers get activated on controller reconnect

teskje:
This PR fixes a race condition that could cause workers to not get activated when their connection to the controller dropped. The `LocalClient` connected to each worker makes sure that an activation is sent when it drops, but there is a possibility that the controller connection fails before a `LocalClient` has been created, in which case any worker activators created so far would be dropped without activating the worker.

The fix implemented here is to immediately wrap worker activators in a `LocalActivator` type with a `Drop` impl that ensures activation on drop. Some care needs to be taken, by ordering declarations appropriately, to ensure that activators are always dropped _after_ the corresponding command channels have been dropped, to ensure workers observe the command channel disconnection when they are activated.

### Checklist


src/service/src/local.rs
```diff
@@ -36,23 +37,52 @@ impl Activatable for Thread {
...
 
+/// An activator for a thread.
+///
+/// This wraps any `Activatable` and has a `Drop` impl to ensure the thread is always activated
+/// when the activator is dropped. This is important to ensure workers have a chance to observe
+/// that their command channel has closed and prepare for reconnection.
+#[derive(Debug)]
+pub struct LocalActivator {
+    inner: Box<dyn Activatable>,
+}
```
teskje, line 48: I avoided making this type generic over `Activatable` to reduce some of the noise in the using code.

antiguru (approved):
I think this looks good! I'm not sure about the name `LocalActivator`, but I see that it borrows from `LocalClient`, so maybe it's not so bad? Anyway, that shouldn't block progress on this PR!

teskje:
TFTR!

PR #31479, opened 2025-02-13
storage: allow singleton sources on multi-replica clusters

aljoscha:
This is stacked on https://github.com/MaterializeInc/materialize/pull/31227, so only the last few commits are relevant for singleton sources.

NOTE: There is still murkiness around `active_copies` in the controller itself, which have to be fixed before we can merge!

Opening this PR early to so I can run nightly. @def- With this, all sources should now work on multi-replica clusters, I imagine there is tests that you would like to change?

### DIFF
--- a/src/adapter/src/coord/caught_up.rs
+++ b/src/adapter/src/coord/caught_up.rs
@@ -245,8 +245,7 @@ impl Coordinator {
             .controller
             .storage
             .active_ingestions(cluster.id)
-            .iter()
-            .copied()
+            .into_iter()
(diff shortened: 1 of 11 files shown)


def-:
> @def- With this, all sources should now work on multi-replica clusters, I imagine there is tests that you would like to change?

Yes, we should probably switch more clusters in tests over to having multiple replicas just to get more coverage.

aljoscha:
@def This now passes tests and nightly, so next step would be to tweak tests so more clusters (with pg/mysql) sources have multiple replicas.

PR #31490, opened 2025-02-13
[sinks] Set with_snapshot=false even during history reduction

bkirwi:
### Checklist

### DIFF
--- a/src/dyncfg/src/lib.rs
+++ b/src/dyncfg/src/lib.rs
@@ -299,6 +299,19 @@ impl<T: ConfigType> ConfigValHandle<T> {
     pub fn get(&self) -> T {
         T::from_val(self.val.load())
     }
+
+    /// Return a new handle that returns the constant value provided,
+    /// generally for testing.
+    pub fn disconnected<X>(value: X) -> Self
+    where
+        X: ConfigDefault<ConfigType = T>,
+    {
+        let config_val: ConfigVal = value.into_config_type().into();
+        Self {
+            val: config_val.into(),
+            _type: Default::default(),
+        }
+    }
 }
 
 /// A type-erased configuration value for when set of different types are stored
--- a/src/storage-controller/src/history.rs
+++ b/src/storage-controller/src/history.rs
@@ -11,10 +11,12 @@
 
 use std::collections::BTreeMap;
 
+use mz_dyncfg::ConfigValHandle;
 use mz_ore::cast::{CastFrom, CastInto};
 use mz_storage_client::client::StorageCommand;
 use mz_storage_client::metrics::HistoryMetrics;
 use mz_storage_types::parameters::StorageParameters;
+use timely::PartialOrder;
 
 /// A history of storage commands.
 #[derive(Debug)]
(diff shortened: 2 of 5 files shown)


bkirwi:
[Nightlies](https://buildkite.com/materialize/nightly/builds?branch=bkirwi%3Asink-progress-3) - the first run was pretty red, but it seems to be unrelated breakage in main... rerunning the failed tests after a rebase seems greener.

bkirwi:
Bump for @aljoscha or @petrosagg if either of you have a minute to think it over!

src/dyncfg/src/lib.rs
```diff
@@ -299,6 +299,19 @@ impl<T: ConfigType> ConfigValHandle<T> {
     pub fn get(&self) -> T {
         T::from_val(self.val.load())
     }
+
+    /// Return a new handle that returns the constant value provided,
+    /// generally for testing.
+    pub fn disconnected<X>(value: X) -> Self
```
petrosagg, line 305: We could additionally mark this with `#[cfg(test)]` if it should never be used in normal code

src/storage-controller/src/history.rs
```diff
@@ -29,17 +31,20 @@ pub(crate) struct CommandHistory<T> {
     commands: Vec<StorageCommand<T>>,
     /// Tracked metrics.
     metrics: HistoryMetrics,
+    /// Config: whether to use the snapshot-frontier optimization for sinks.
+    enable_snapshot_frontier: ConfigValHandle<bool>,
 }
 
-impl<T: std::fmt::Debug> CommandHistory<T> {
+impl<T: std::fmt::Debug + timely::progress::Timestamp> CommandHistory<T> {
```
petrosagg, line 38: nit: `timely::progress::Timestamp` implies `std::fmt::Debug`

src/storage-controller/src/history.rs
```diff
@@ -155,7 +160,16 @@ impl<T: std::fmt::Debug> CommandHistory<T> {
                 if frontier.is_empty() {
                     continue;
                 }
-                sink.description.as_of = frontier;
+                // The as-of is at least the implied capability of the sink collection. If the as-of
+                // advances for an existing export, that can only be because the implied capability
+                // has advanced, which means that the write frontier has advanced, which means the
+                // snapshot has definitely been written out.
+                if PartialOrder::less_than(&sink.description.as_of, &frontier) {
```
petrosagg, line 167: This is fine when `T` is a `TotalOrder`, and maybe we should add that to the `T` bound to make it clear and not complicate the code.

If `T` is partially ordered then `less_than` is not sufficient to conclude that no snapshot is needed because that test returns true if **any** of the timestamps of `description.as_of` `frontier` is greater than the  but we want to check that **all** the timestamps of `frontier` are greater than `description.as_of`. I believe the test for the general `T: Timestamp` case is `description.as_of.all(|t| !frontier.less_equal(t))`, which we also use in reclocking [here](https://github.com/MaterializeInc/materialize/blob/main/src/timely-util/src/reclock.rs#L336).

src/dyncfg/src/lib.rs
bkirwi: Tried this, but no luck... since it ends up being used in _another_ crate's test.

To bad! Thankfully shouldn't be too much of a footgun, given how this stuff is used in practice.

src/storage-controller/src/history.rs
bkirwi: Cool! It looks like adding the bound is not too invasive, so I'll go that route for now.

bkirwi:
Followups are in - thanks for the review!

I've rebased and kicked off another set of nightlies - a subset that gives us some sinks coverage but is green on main. I'll merge assuming that looks good.

PR #31496, opened 2025-02-14
parallel-workload: Deal with changes in parallelism handling

def-:
It seems that dropping objects now fails more often

Failure seen in https://buildkite.com/materialize/nightly/builds/11127#0194fcbb-11fa-4b32-884d-9f3de1b2bfbf

I'm thinking this is fallout from https://github.com/MaterializeInc/materialize/pull/31309

### Checklist

### DIFF
--- a/misc/python/materialize/parallel_workload/parallel_workload.py
+++ b/misc/python/materialize/parallel_workload/parallel_workload.py
@@ -95,29 +95,29 @@ def run(
     with system_conn.cursor() as system_cur:
         system_exe = Executor(rng, system_cur, None, database)
         system_exe.execute(
-            f"ALTER SYSTEM SET max_schemas_per_database = {MAX_SCHEMAS * 20 + num_threads}"
+            f"ALTER SYSTEM SET max_schemas_per_database = {MAX_SCHEMAS * 40 + num_threads}"
         )
         # The presence of ALTER TABLE RENAME can cause the total number of tables to exceed MAX_TABLES
         system_exe.execute(
-            f"ALTER SYSTEM SET max_tables = {MAX_TABLES * 20 + num_threads}"
+            f"ALTER SYSTEM SET max_tables = {MAX_TABLES * 40 + num_threads}"
         )
         system_exe.execute(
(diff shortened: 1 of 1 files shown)


def-:
Green nightly run: https://buildkite.com/materialize/nightly/builds/11148

ggevay:
> It seems that dropping objects now fails more often

Could you point me to the error msg where a drop failed? I'm only seeing the following error in the linked run of Parallel Workload, which seems to be for an object creation:
```
[worker_0] Query failed: CREATE ROLE role1130; creating role would violate max_roles limit (desired: 303, limit: 302, current: 302)
HINT:  Drop an existing role or contact support to request a limit increase.
Exception in thread worker_0:
  File "/var/lib/buildkite-agent/builds/hetzner-aarch64-16cpu-32gb-7e2606e1/materialize/nightly/misc/python/materialize/parallel_workload/executor.py", line 152
, in execute
    self.cur.execute(query.encode())
  File "/usr/local/lib/python3.12/dist-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
psycopg.errors.InsufficientResources: creating role would violate max_roles limit (desired: 303, limit: 302, current: 302)
HINT:  Drop an existing role or contact support to request a limit increase.
```

> I'm thinking this is fallout from https://github.com/MaterializeInc/materialize/pull/31309

Well, nothing is impossible, but I would be surprised, because [the resource limit checks look at the catalog directly](https://github.com/MaterializeInc/materialize/blob/e8c42c65afb7acd55eb7e530a92c89a9165f2e33/src/adapter/src/coord/ddl.rs#L1802), rather than at builtin tables, so the way table updates happen shouldn't influence resource limit checks.

Btw., I'm wondering if we shouldn't simply set all these limits to something like 1000000000, i.e., infinity, for this test. As far as I know, the test does random DDL, so I would guess that nothing is ensuring that the numbers of various objects won't go over standard limits. And if we would like to check that resource limit checks themselves are working, then we could have directed tests for that, creating controlled numbers of certain types of objects.

def-:
I don't think we have good logging for it, so just my intuition. There could for example be a timeout during the drop, then the client thinks the drop failed, but environmentd can still process it.

PR #31506, opened 2025-02-14
compute: add mz_compute_collection_count metric

teskje:
This commit adds a replica metric, `mz_compute_collection_count`, that tracks how many compute collections are installed on each replica, split up by type and hydration status. The metric enables tracking hydration progress during 0dt deployments as the percentage of hydrated collections can be derived from it.

[Slack thread.](https://materializeinc.slack.com/archives/C08ACQNGSQK/p1738774881494979)

### Motivation

Part of https://github.com/MaterializeInc/database-issues/issues/8948

### Checklist

### DIFF
--- a/src/compute/src/metrics.rs
+++ b/src/compute/src/metrics.rs
@@ -13,7 +13,7 @@ use mz_compute_client::metrics::{CommandMetrics, HistoryMetrics};
 use mz_ore::cast::CastFrom;
 use mz_ore::metric;
 use mz_ore::metrics::{raw, MetricsRegistry, UIntGauge};
-use mz_repr::SharedRow;
+use mz_repr::{GlobalId, SharedRow};
 use prometheus::core::{AtomicF64, GenericCounter};
 use prometheus::proto::LabelPair;
 use prometheus::{Histogram, HistogramVec};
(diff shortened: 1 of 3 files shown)


src/compute/src/metrics.rs
```diff
@@ -161,6 +164,11 @@ impl ComputeMetrics {
                 help: "The remaining seconds until replica expiration. Can go negative, can lag behind.",
                 var_labels: ["worker_id"],
             )),
+            collection_count: registry.register(metric!(
+                name: "mz_compute_collection_count",
+                help: "The number and hydration status of maintained compute collections.",
+                var_labels: ["worker_id", "type", "hydrated"],
```
teskje, line 170: @materializekatrina I decided to add a "type" field with values such as (`user`, `system`, `log`, `transient`) because I imagine we want to filter collections for 0dt hydration tracking. We'll at least want to filter out `transient` collections (SELECTs, subscribes) as they can cause noise, and we may want to only look at `user` collections because the number of `system`/`log` collections can change between Mz versions.

antiguru (approved):
This looks fine. A downside of reporting this through metrics only is that it's not possible to see the information through SQL. Do we have enough information in the introspection views to surface the collection count? If not, do you mind recording this as a feature request?

teskje:
> Do we have enough information in the introspection views to surface the collection count?

I think so! All the information exposed by this metric is available through `mz_compute_hydration_times_per_worker`, and the unified introspection collection `mz_compute_hydration_times` based on it. We already have a promsql-exporter metric based on the latter, but the trouble with that is that it doesn't include the read-only environment during 0dt upgrades (because the read-only env cannot write to persist). We could have a metric based on `mz_compute_hydration_times_per_worker` directly, but that metric would be unreliable during hydration, which is a deal-breaker when we want to use it to track hydration progress.

teskje:
TFTR!

PR #31511, opened 2025-02-14
Always enable lgalloc, unless disabled

antiguru:
Modernize the infrastructure around lgalloc. In the past, it needed to be explicitly enabled, but now it needs to be disabled and is on by default. This reflects our current setup better where we expect to have access to lgalloc if we have disk. The setting still serves as a switch to disable it in case it is needed.

Introduces a new setting `enable_lgalloc` that gates access to lgalloc as a whole. The old `enable_columnation_lgalloc` controls the behavior of `Region::new_auto` instead, which is closer to the name.

### Checklist

### DIFF
--- a/src/compute-types/src/dyncfgs.rs
+++ b/src/compute-types/src/dyncfgs.rs
@@ -47,7 +47,7 @@ pub const LINEAR_JOIN_YIELDING: Config<&str> = Config::new(
      work, respectively, rather than falling back to some default.",
 );
 
-/// Enable lgalloc for columnation.
+/// Enable lgalloc.
 pub const ENABLE_LGALLOC: Config<bool> = Config::new("enable_lgalloc", true, "Enable lgalloc.");
 
 /// Enable lgalloc for columnation.
--- a/src/compute-types/src/dyncfgs.rs
+++ b/src/compute-types/src/dyncfgs.rs
@@ -48,9 +48,12 @@ pub const LINEAR_JOIN_YIELDING: Config<&str> = Config::new(
 );
 
 /// Enable lgalloc for columnation.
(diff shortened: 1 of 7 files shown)


src/compute-types/src/dyncfgs.rs
```diff
@@ -47,10 +47,13 @@ pub const LINEAR_JOIN_YIELDING: Config<&str> = Config::new(
      work, respectively, rather than falling back to some default.",
 );
 
+/// Enable lgalloc for columnation.
```
teskje, line 50: Copy-paste-o?

src/orchestrator-process/src/lib.rs
```diff
@@ -350,6 +356,7 @@ impl NamespacedOrchestrator for NamespacedProcessOrchestrator {
         self.send_command(WorkerCommand::EnsureService {
             id: id.to_string(),
             config,
+            scheduling_config,
```
teskje, line 359: Did you consider applying `always_use_disk` to the `ServiceConfig` here, instead of sending the whole `ServiceSchedulingConfig`? This looks like it would end up a bit less noisy, but more importantly it would be consistent with what we do in the k8s orchestrator.

src/compute-types/src/dyncfgs.rs
antiguru: Fixed.

src/orchestrator-process/src/lib.rs
antiguru: I cleaned up the code. Instead of sending the raw config and scheduling config, we send a custom struct. This only encodes the information we need downstream.

teskje (approved):
LGTM

PR #31515, opened 2025-02-16
storage: when dropping objects, eagerly clean up state and update builtins

aljoscha:
NOTE: This is now a whole bunch of other cleanups that were enabled/or made feasible by this initial change to eagerly clean out state and update builtin collections 

Before this change, we had a more sequentialized/blocking protocol for
dropping sources (and other collections):

1. envd receives DDL that drops source
2. envd releases read holds and sends AllowCompaction to clusterd
3. clusterd drops source and sends back a DroppedId message
4. envd cleans up it's collection state and updates builtin collections:
 - updating mz_storage_shards (the global id -> shard id mapping)
 - updating source/sink status to "dropped"
 - removing statistics tracking

Especially the transition between 3. and 4. can block a while if a
cluster is unresponsive.

Plus, in some situations we don't get DroppedId messages, so we would
not do these updates/cleanups. The common of these situations are:
- dropping a replica right after dropping sources/sinks
- dropping a whole cluster

Now, we eagerly apply updates to our builtin collections and also
eagerly clean up state. The protocol becomes:

1. envd receives DDL that drops source
2. in parallel:
 2.a: envd releases read holds and sends AllowCompaction to clusterd
 2.b. envd cleans up it's collection state and updates builtin collections:
  - updating mz_storage_shards (the global id -> shard id mapping)
  - updating source/sink status to "dropped"
  - removing statistics tracking
3. clusterd drops source and sends back a DroppedId message

The benefits of this change:
- less code
- control flow of dropping things is more localized and easier to
  understand
- we fix problems where we inadvertently didn't drop state and update
  builtin collections (this could be done independently of this change, but with the changed code structure it's more obvious what happens, and should happen, when dropping an object)

And, for the more-protocol-asserts aficionado: this fixes bugs in "active copies" tracking and re-enables stricter assertions about what types of messages we expect when.

Here's Lamport Diagrams that show the change in the distributed protocol:
dropping-sink-non-blocking

### DIFF
--- a/src/storage-controller/src/instance.rs
+++ b/src/storage-controller/src/instance.rs
@@ -395,7 +395,7 @@ where
                 // `tokio::sync::mpsc::UnboundedReceiver::recv` is documented as cancel safe.
                 command = self.command_rx.recv() => {
                     let Some(mut command) = command else {
-                        // Controller is no longer interested in this replica. Shut down.
+                        tracing::debug!(%self.replica_id, "controller is no longer interested in this replica, shutting down message loop");
                         break;
                     };
 
@@ -410,7 +410,7 @@ where
                     };
 
                     if self.response_tx.send((Some(self.replica_id), response)).is_err() {
-                        // Controller is no longer interested in this replica. Shut down.
+                        tracing::debug!(%self.replica_id, "controller (receiver) is no longer interested in this replica, shutting down message loop");
                         break;
                     }
                 }
--- a/src/storage-controller/src/instance.rs
+++ b/src/storage-controller/src/instance.rs
@@ -110,11 +110,6 @@ where
         instance
     }
 
-    /// Returns the number of replicas of this storage instance.
-    pub fn replica_count(&self) -> usize {
-        self.replicas.len()
-    }
-
     /// Returns the IDs of all replicas connected to this storage instance.
     pub fn replica_ids(&self) -> impl Iterator<Item = ReplicaId> + '_ {
         self.replicas.keys().copied()
--- a/src/storage-controller/src/instance.rs
+++ b/src/storage-controller/src/instance.rs
@@ -69,7 +69,11 @@ pub(crate) struct Instance<T> {
     /// A function that returns the current time.
     now: NowFn,
     /// A sender for responses from replicas.
-    response_tx: mpsc::UnboundedSender<StorageResponse<T>>,
+    ///
+    /// Responses are tagged with the [`ReplicaId`] of the replica that sent the
+    /// response. Responses that don't originate from a replica (e.g. a "paused"
+    /// status update, when no replicas are connected) are tagged with `None`.
+    response_tx: mpsc::UnboundedSender<(Option<ReplicaId>, StorageResponse<T>)>,
 }
 
 impl<T> Instance<T>
@@ -82,7 +86,7 @@ where
         envd_epoch: NonZeroI64,
         metrics: InstanceMetrics,
         now: NowFn,
-        instance_response_tx: mpsc::UnboundedSender<StorageResponse<T>>,
+        instance_response_tx: mpsc::UnboundedSender<(Option<ReplicaId>, StorageResponse<T>)>,
         enable_snapshot_frontier: ConfigValHandle<bool>,
     ) -> Self {
         let history = CommandHistory::new(metrics.for_history(), enable_snapshot_frontier);
@@ -194,7 +198,10 @@ where
 
         if !status_updates.is_empty() {
             let response = StorageResponse::StatusUpdates(status_updates);
-            let _ = self.response_tx.send(response);
+            // NOTE: If we lift this "inject paused status" logic to the
+            // controller, we could instead return ReplicaId instead of an
+            // Option<ReplicaId>.
+            let _ = self.response_tx.send((None, response));
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/storage-controller/src/lib.rs
```diff
@@ -1951,16 +2045,9 @@ where
...
-
-                if let Some(_collection) = self.collections.remove(&id) {
-                    // Nothing to do, we already dropped read holds in
-                    // `drop_sources_unvalidated`.
-                } else {
-                    soft_panic_or_log!(
-                        "DroppedId for ID {id} but we have neither ingestion nor export \
-                         under that ID"
-                    );
+                let prev = self.dropped_ids.remove(&id);
```
bkirwi, line 2048: How does this interact with multi-replica storage clusters? Do we only expect to get one dropped id message globally?

aljoscha: Ah yeah, dammit, `DroppedId` is a bit confusing, because for some things I thing the replica client logic only forwards a thing when all of the replicas agree, but I think you're right and for `DroppedId` they pass through as is, so we get multiple of them. I would just not remove things from `dropped_ids` and accept the small cost of keeping these around until a restart. ((I don't like using `dropped_ids` to begin with, but it's a compromise. ))

teskje: > I would just not remove things from dropped_ids and accept the small cost of keeping these around until a restart.

I'm not a fan of intentionally leaking memory. I think the storage controller has done that previously with the collection (and/or export) state and it turned out fine, but someone running `CREATE`/`DROP` in a loop could still oom envd. We definitely can't do it in compute where transient dataflows are created and dropped all the time.

src/storage-controller/src/lib.rs
```diff
@@ -1003,9 +1021,10 @@ where
                 data_source,
                 collection_metadata: metadata,
                 extra_state,
-                // Default to 1, which is good for tables and webhooks. For
-                // ingestions, this will be updated in run_ingestion.
-                active_copies: 1,
+                // Default to empty as tables and webhooks don't run on
+                // replicas.  For ingestions, this will be updated in
+                // run_ingestion.
+                active_replicas: BTreeSet::new(),
```
petrosagg, line 1027: This is the only piece that confused me since the empty set represents zero, but previously the active copies was one. Is the explanation that we no longer touch this based on `DroppedId` responses?

petrosagg (approved):
The changes look good to me! The only thing that wasn't entirely clear is the comment on the default active replicas.

Also, given both Ben and Jan have looked into this it would be nice to get one more sign off.

src/storage-controller/src/lib.rs
aljoscha: This now tracks how many replicas are running the collection (i.e. are interested in it's state still existing, in terms you once used with me), and tables and webhooks run on zero replicas

aljoscha:
@bkirwi and/or @teskje Could you please take another look? I think your comments are resolved and there's a bunch of new changes. Petros has approved, but mentioned to me that it would be good if you could also take a look again 🙏

src/storage-controller/src/lib.rs
```diff
@@ -1911,9 +1911,42 @@ where
         );
         self.set_hold_policies(drop_policy);
 
+        // Record the drop status for all sink drops.
+        //
+        // We also delete the items' statistics objects.
+        //
+        // The locks are held for a short time, only while we do some hash map removals.
```
teskje, line 1945: Nit: They are all btree maps, no? :)

src/storage-controller/src/instance.rs
```diff
@@ -67,7 +67,7 @@ pub(crate) struct Instance<T> {
     /// A function that returns the current time.
     now: NowFn,
     /// A sender for responses from replicas.
```
teskje, line 69: Could you extend this comment to say what the meaning of the `Option<ReplicaId>` is, especially if it's `None`?

src/storage-controller/src/instance.rs
```diff
@@ -389,6 +389,7 @@ where
                 command = self.command_rx.recv() => {
                     let Some(mut command) = command else {
                         // Controller is no longer interested in this replica. Shut down.
```
teskje, line 386: Nit: I think the comment is now redundant.

teskje (approved):
Lgtm, for whatever that's worth!

src/storage-controller/src/instance.rs
aljoscha: clearly! 😅

src/storage-controller/src/lib.rs
aljoscha: Very true, I just copied this comment over ... 🙈

src/storage-controller/src/lib.rs
```diff
@@ -1786,7 +1796,8 @@ where
                         // Ingestion exports also have ReadHolds that we need to
                         // downgrade, and much of their drop machinery is the
                         // same as for the "main" ingestion.
-                        ingestions_to_drop.insert(id);
+                        ingestions_to_drop.insert(*id);
+                        source_statistics_to_drop.push(*id);
                     }
                     DataSource::Other | DataSource::Introspection(_) | DataSource::Progress => (),
                     DataSource::Sink { .. } => {}
```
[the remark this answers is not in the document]
aljoscha: Yep, I wanted to not roll too many behavior changes/bugfixes into this one

aljoscha:
Thanks for the review! I addressed your comments and will merge now! 🙇‍♂️

PR #31518, opened 2025-02-17
compute/correction_v2: fix stage size tracking

teskje:
This commit fixes a bug in size metrics tracking for the `CorrectionV2` stage. In `consolidate_before`, we would call `stage.flush()` and if this returned no updates we would assume that the stage had been empty and immediately return without updating the metrics. However, `stage.flush()` performs a `consolidate`, which can cancel out updates, and we would miss recording these deletions in the metrics.

This fix chosen here is to perform the emptiness checking and early return before calling `stage.flush()`. I find this more obvious than to keeping the current flow and adding a second `update_metrics` to `consolidate_before`, but both would work.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/8971

### Checklist

### DIFF
--- a/src/compute/src/sink/correction_v2.rs
+++ b/src/compute/src/sink/correction_v2.rs
@@ -271,12 +271,17 @@ impl<D: Data> CorrectionV2<D> {
     /// Once this method returns, all remaining updates before `upper` are contained in a single
     /// chain. Note that this chain might also contain updates beyond `upper` though!
     fn consolidate_before(&mut self, upper: &Antichain<Timestamp>) {
-        let mut chains = std::mem::take(&mut self.chains);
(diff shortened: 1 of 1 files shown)


antiguru (approved):
Looks good, thank you! Do you mind adding a comment in the code saying what the PR description says? Otherwise I can imagine our future selves might 'optimize' this at some point, only to discover that metrics don't quite add up anymore.

teskje:
I added the comment. Then I figured if we're explaining how to correctly early-return based on the flushed stage contents, we can also just do it. I kept the early return in the beginning though: This way we avoid the work of updating the metrics in the common case where the correction buffer is empty.

teskje:
TFTR!

PR #31452, opened 2025-02-10
conditionally create mz_system and mz_probe cluster replicas

SangJunBak:
See commit messages for details

### Motivation
  * This PR adds a known-desirable feature.
https://github.com/MaterializeInc/database-issues/issues/8954

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #31527, opened 2025-02-18
terraform test: New version with local helm chart support

def-:
### Checklist

### DIFF
--- a/test/terraform/aws-persistent/main.tf
+++ b/test/terraform/aws-persistent/main.tf
@@ -16,22 +16,41 @@ resource "random_password" "db_password" {
   special = false
 }
 
+variable "operator_version" {
+  type    = string
+  default = "v25.2.0-beta.1"
+}
+
+variable "orchestratord_version" {
+  type    = string
+  default = "v0.130.3"
+}
+
 module "materialize_infrastructure" {
-  source = "git::https://github.com/MaterializeInc/terraform-aws-materialize.git?ref=v0.2.0"
+  source = "git::https://github.com/MaterializeInc/terraform-aws-materialize.git?ref=v0.2.5"
 
   # Basic settings
   namespace    = "aws-persistent"
   environment  = "dev"
   install_materialize_operator = true
+  use_local_chart = true
+  helm_chart = "materialize-operator-v25.2.0-beta.1.tgz"
+  operator_version = var.operator_version
+  orchestratord_version = var.orchestratord_version
 
-  helm_values = {
-      defaultReplicationFactor = {
-          system = 1
-          probe = 1
-          support = 1
-          analytics = 1
-      }
-  }
+  # TODO: Doesn't seem to work yet
+  # helm_values = {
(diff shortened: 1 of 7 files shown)


def-:
@bobbyiliev I'm wondering why this doesn't work: https://buildkite.com/materialize/test/builds/99223#01951a82-de29-48b1-979f-897baf998e06
```
error: getting semver of materialize: db error: ERROR: CLUSTER "mz_system" has no replicas available to service request
```
I'm explicitly setting the defaultReplicationFactor in helm_values, I assume that should work together with use_local_helm_chart.
I'm passing in the operator version here: https://github.com/MaterializeInc/materialize/pull/31527/files#diff-83d8f0169fbca62050b7506c763a1ce53697a5979da1d3759d83cb5a48e3f3eaR163 but I'm actually wondering if that will use the operator from the current source code how I could specify that.

This is somehow related to https://github.com/MaterializeInc/materialize/pull/31452, but I don't understand how we can both use a new operator version that sets the defaults to 0, but doesn't accept me to override them. cc @SangJunBak maybe you have an idea too

SangJunBak:
@def- from first glance, it seems like `defaultReplicationFactor` isn't scoped properly! If you look in `values.yaml`, it's scoped as `clusters.defaultReplicationFactor`. I am curious... If we specify `clusters.defaultReplicationFactor`, would we need to specify `clusters.sizes` and other variables in the `clusters` scope?

def-:
I tried that but no difference still:
```
error: getting semver of materialize: db error: ERROR: CLUSTER "mz_system" has no replicas available to service request
```

test/terraform/aws-persistent/main.tf
```diff
@@ -16,20 +16,81 @@ resource "random_password" "db_password" {
...
+  operator_version = var.operator_version
+  orchestratord_version = var.orchestratord_version
 
   helm_values = {
-      defaultReplicationFactor = {
-          system = 1
-          probe = 1
-          support = 1
-          analytics = 1
+      clusters = {
```
bobbyiliev, line 42: The `clusters` section is under the `operator` key in the values file:

https://github.com/MaterializeInc/materialize/blob/main/misc/helm-charts/operator/values.yaml#L11

def-: So is it true that if I set anything in `operator` I have to override everything?

bobbyiliev: No, this should not be the case. It will override only the values that you are modifying:

> https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#overriding-values-from-a-parent-chart

Are you seeing a different behavior?

def-: I didn't get it into a working state, I'm not sure what I'm doing wrong with the helm_values or if they don't work. I just manually alter the cluster now and that works for me. So can be merged in this state.

bobbyiliev (approved):

PR #31529, opened 2025-02-18
cluster: clean up worker activator handling

teskje:
This PR performs some cleanup around the handling of worker activators in the cluster code, unlocked by https://github.com/MaterializeInc/materialize/pull/31523.

Now that the compute command channel dataflow is persistent, rather than recreated on every client connection, there is no need to send back connection-specific activators from the workers to the `ClusterClient`. Instead the `ClusterClient` can simply unpark the threads running the client workers, which it can get at through the `TimelyContainer`.

Note that the thread unparking isn't actually necessary for the compute cluster because workers are activated by the channel adapter thread sending new commands into the internal channel already. But the storage cluster needs it, so we just do it for both.

Also, the `Activable` trait and the `LocalActivator` wrapper type are not required anymore and are therefore deleted here.

### Motivation

### Checklist


antiguru (approved):
Thank you, much clearer!

teskje:
TFTR!

PR #31535, opened 2025-02-18
storage: when envd reconnects, report current status at latest timestamp

aljoscha:
This makes it so that status updates from the latest deployment, for example when coming out of read-only mode, sort last in mz_source/sink_status_history, so get surfaced as the latest status in mz_source/sink_statuses.

This fixes (or makes less likely) a bug where a failing/stalled source/sink in the old deployment would crowd out the running status of objects in the new deployment.

This fix work, but doesn't feel ideal. A problem is that the raw status collections are append-only collections, so `envd` will simply append status updates to the end. One could argue that the smarts should be in `envd`/the controller. But ultimately, it would also have to append a status update that "trumps" the earlier updates, so would do the same thing that we're achieving with this change.

Fixes MaterializeInc/database-issues#8863

### Checklist

### DIFF
--- a/src/storage/src/storage_state.rs
+++ b/src/storage/src/storage_state.rs
@@ -846,7 +846,22 @@ impl<'w, A: Allocate> Worker<'w, A> {
         // If we haven't done the initial status report, report all current
         // statuses
         if !self.storage_state.initial_status_reported {
-            to_report.extend(self.storage_state.latest_status_updates.values().cloned());
+            // We pull initially reported status updates to "now", so that they
+            // sort as the latest update in internal status collections. This
+            // makes it so that a newly bootstrapped envd can append status
+            // updates to internal status collections that report an accurate
+            // view as of the time when they came up.
+            let now_ts = mz_ore::now::to_datetime((self.storage_state.now)());
+            let status_updates = self
+                .storage_state
+                .latest_status_updates
+                .values()
+                .cloned()
+                .map(|mut update| {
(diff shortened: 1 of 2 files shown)


def- (approved):
Thanks for adding a test!

teskje (approved):
Seems fine! I agree it's not ideal with the reliance on clock synchronization between different clusterd processes.

Have you considered selecting the `occurred_at` time in the controller, before writing down a new status? That would make sure that times are always increasing and cannot race. It would change the semantics of `occurred_at` from "when the cluster learned about it" to "when the controller learned about it" but does that matter?

aljoscha:
> Have you considered selecting the `occurred_at` time in the controller, before writing down a new status? That would make sure that times are always increasing and cannot race. It would change the semantics of `occurred_at` from "when the cluster learned about it" to "when the controller learned about it" but does that matter?

That's a good point, but I didn't want to futz with the semantics too much, plus if we just took the "now" timestamp on envd that also wouldn't protect against clock skew. I think the proper solution would be to use the timestamp oracle, but the controller (who writes down these status updates) doesn't have access to that right now. So it's a whole long tail of things that would have to be changed. 🙈

teskje:
True... Another proper solution would be to remove the `occurred_at` column and use the update timestamp instead, i.e. have a retained-history relation instead of log. The catch of course being that retained-history relations are not ergonomic to use in Mz at the moment.

PR #31541, opened 2025-02-18
[persist] Fix up the Persist malestrom test

bkirwi:
- We assert that since < upper-1 on the next trip through the loop, but if we don't obtain the two frontiers atomically, so it's possible for the since to be larger than the write frontier if it takes a while to fetch. Now we fetch them in the other order.
- `compare_and_downgrade_since` is not _actually_ idempotent! If we're trying to compare and downgrade changing the token from X to Y, it's possible to succeed, changing the token to Y, but fail with some indeterminate error... and then a retry will fail since the token is already Y. This doesn't actually affect our real-life usage, but it does affect the particular expectation in `advance_since`. Now we check whether the since has advanced unexpectedly no matter what the latest version of the opaque was.

### Motivation

Fixes: https://github.com/MaterializeInc/database-issues/issues/8623

### Checklist


def- (approved):
LGTM

PR #31551, opened 2025-02-19
doc/design: zero-downtime upgrades: milestone 1

aljoscha:
I'd like to make this available for historic context, see the first comment in the doc:

```
NOTE: At the time this was merged, we were past zero-downtime upgrades,
milestone 2 already. But we merged this so that historic context is available
in our repository of design docs. Some of the descriptions in here are half
baked, but overall the implementation followed closely what we laid out here.
```

@ParkMyCar you are closes to the adapter/0dt work, but I'm largely looking for a stamp to merge it as is. If you think this is alright.


ParkMyCar (approved):
LGTM!

aljoscha:
tyty! 🙇‍♂️

PR #31556, opened 2025-02-19
storage: remove non-feedback upsert implementation

aljoscha:
The feedback upsert implementation is now rolled out to all
environments, is required for doing 0dt upgrades, and is a prerequisite
for multi-replica UPSERT sources.

Removing the code makes sure we can't accidentally switch
implementations and get into incorrect situations.


aljoscha:
superseded by https://github.com/MaterializeInc/materialize/pull/31593

PR #31567, opened 2025-02-20
Prefix the environment role, rolebinding, and serviceaccount

alex-hunt-materialize:
Prefix the environment role, rolebinding, and serviceaccount with the `mz{resource_id}-` like all other objects created by orchestratord.

### Motivation

These objects were missed when adding prefixes earlier. Without the prefixes, there can be conflicts when running multiple materialize installations in the same namespace.

### Checklist

  No changes needed currently, but may need to be included in my upcoming migration to reconciling Materialize CRs.

### DIFF
--- a/src/cloud-resources/src/crd/materialize.rs
+++ b/src/cloud-resources/src/crd/materialize.rs
@@ -161,15 +161,15 @@ pub mod v1alpha1 {
         }
 
         pub fn service_account_name(&self) -> String {
-            self.name_unchecked()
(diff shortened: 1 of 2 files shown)


doy-materialize (commented):
is this actually necessary? they are using the materialize resource name as their name, so it's not actually possible for them to conflict with anything.

if we want to do this for consistency, we should do something like `self.name_prefixed("environmentd")` instead, since using both the prefix and the resource name together feels confusing, and will likely run into resource name length limits.

PR #31568, opened 2025-02-20
[storage] Always register sinks last

bkirwi:
Shouldn't hurt and may help!

### Checklist


src/storage-client/src/storage_collections.rs
```diff
@@ -1921,20 +1922,20 @@ where
...
-        let (tables_to_register, collections_to_register): (Vec<_>, Vec<_>) = to_register
-            .into_iter()
-            .partition(|(_id, desc, ..)| matches!(desc.data_source, DataSource::Table { .. }));
-        let to_register = tables_to_register
-            .into_iter()
-            .rev()
-            .chain(collections_to_register.into_iter());
+        #[derive(Ord, PartialOrd, Eq, PartialEq)]
+        enum DependencyOrder {
+            /// Tables should always be registered first, and large ids before small ones.
```
aljoscha, line 1927: OOC, do you remember while we sort tables this way?

bkirwi: `alter table` creates new versions of the table, and has the old versions depend on the new versions. So it's important to create new versions first, I guess! (But I don't recall having seen this before, so that's just guessing from context.)

def-:
I added a test for https://github.com/MaterializeInc/database-issues/issues/8977, triggered nightly: https://buildkite.com/materialize/nightly/builds/11219

misc/python/materialize/checks/all_checks/sink.py
```diff
@@ -976,6 +978,61 @@ def validate(self) -> Testdrive:
...
+            dedent(
+                """
+                > CREATE SOURCE sink_alter_mv_source_src
+                  FROM KAFKA CONNECTION kafka_conn (TOPIC 'sink-alter-mv')
+                > CREATE TABLE sink_alter_mv_source FROM SOURCE sink_alter_mv_source_src (REFERENCE "sink-alter-mv")
+                  FORMAT AVRO USING CONFLUENT SCHEMA REGISTRY CONNECTION csr_conn
+                  ENVELOPE NONE
+
+                > SELECT before IS NULL, (after).a FROM sink_alter_mv_source
+                true 3
```
bkirwi, line 1028: @def- - I don't think this is true in general... if the sink publishes its snapshot I don't think we'll ever retract the old value and replace it with the new one. ([ALTER SINK has awkward semantics in this way](https://materialize.com/docs/sql/alter-sink/#cutover-scenarios).)

def-: Oh, you are right. My bad, will fix it.

bkirwi: Ah, just pushed a change for this -- but not one I'm attached to if you have other ideas!

def-: My idea is a sleep, and checking that we actually get values from all 3 MVs. I'll push it in a minute

src/storage-client/src/storage_collections.rs
petrosagg: This is the reason, yes

misc/python/materialize/checks/all_checks/sink.py
bkirwi: Cool, yeah, I suspect borrowing the approach from `AlterSink` should work fine here too!

src/storage-client/src/storage_collections.rs
```diff
@@ -1921,20 +1922,20 @@ where
...
+            Table(Reverse<GlobalId>),
+            /// For most collections the id order is the correct one.
+            Collection(GlobalId),
+            /// Sinks should always be registered last.
+            Sink(GlobalId),
```
ParkMyCar, line 1932: FWIW we have this heuristic of Sinks being sorted last in the Adapter as well! Can find the exact spot if necessary, but point being there is precedent for this

def- (approved):
Confirmed with the test that this is fixed: https://buildkite.com/materialize/nightly/builds/11222

aljoscha (approved):
I think this is a good change, regardless of the discussion of relying on inherent dependency order in table IDs. We have to fix that one independently.

PR #31580, opened 2025-02-21
[copy_from] Support specifying a column map in `COPY ... FROM <url>`

ParkMyCar:
This PR adds support for specifying a column mapping and filling in default values for `COPY ... FROM <url>` (aka `OneshotSource`).

Let's say we ran `CREATE TABLE t1 (c1 text, c2 text, c3 text DEFAULT 'apple')`, so our table has 3 columns, but the file we're copying into Materialize only has 2 columns. After this PR you could run `COPY INTO t1 (c2, c1) FROM <url>` and it would map the first column of the data to `t1.c2`, the second column to `t1.c1`, and it would fill in the default value of `'apple'` for all values in `t1.c3`.

We do this by planning a `MapFilterProject`, converting it into a `SafeMfpPlan`, and then in `clusterd` when decoding data evaluating the `SafeMfpPlan` for every row.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/8858

### Tips for reviewer

There are no tests in this PR, they are included as part of https://github.com/MaterializeInc/materialize/pull/31581

### Checklist

### DIFF
--- a/src/sql/src/plan.rs
+++ b/src/sql/src/plan.rs
@@ -37,7 +37,9 @@ use ipnet::IpNet;
 use maplit::btreeset;
 use mz_adapter_types::compaction::CompactionWindow;
 use mz_controller_types::{ClusterId, ReplicaId};
-use mz_expr::{CollectionPlan, ColumnOrder, MirRelationExpr, MirScalarExpr, RowSetFinishing};
+use mz_expr::{
+    CollectionPlan, ColumnOrder, MapFilterProject, MirRelationExpr, MirScalarExpr, RowSetFinishing,
+};
 use mz_ore::now::{self, NOW_ZERO};
 use mz_pgcopy::CopyFormatParams;
 use mz_repr::adt::mz_acl_item::{AclMode, MzAclItem};
@@ -47,8 +49,8 @@ use mz_repr::optimize::OptimizerFeatureOverrides;
 use mz_repr::refresh_schedule::RefreshSchedule;
 use mz_repr::role_id::RoleId;
 use mz_repr::{
-    CatalogItemId, ColumnName, ColumnType, Diff, GlobalId, RelationDesc, Row, ScalarType,
-    Timestamp, VersionedRelationDesc,
+    CatalogItemId, ColumnIndex, ColumnName, ColumnType, Diff, GlobalId, RelationDesc, Row,
+    ScalarType, Timestamp, VersionedRelationDesc,
 };
 use mz_sql_parser::ast::{
     AlterSourceAddSubsourceOption, ClusterAlterOptionValue, ConnectionOptionName, QualifiedReplica,
@@ -930,10 +932,21 @@ pub struct ShowColumnsPlan {
 
 #[derive(Debug)]
 pub struct CopyFromPlan {
+    /// Table we're copying into.
     pub id: CatalogItemId,
+    /// Source we're copying data from.
     pub source: CopyFromSource,
-    pub columns: Vec<usize>,
+    /// How input columns map to those on the destination table.
+    ///
+    /// TODO(cf2): Remove this field in favor of the mfp.
+    pub columns: Vec<ColumnIndex>,
+    /// [`RelationDesc`] describing the input data.
+    pub source_desc: RelationDesc,
+    /// Changes the shape of the input data to match the destination table.
+    pub mfp: MapFilterProject,
+    /// Format specific params for copying the input data.
     pub params: CopyFormatParams<'static>,
+    /// Filter for the source files we're copying from, e.g. an S3 prefix.
     pub filter: Option<CopyFromFilter>,
 }
 
--- a/src/sql/src/plan/query.rs
+++ b/src/sql/src/plan/query.rs
@@ -510,7 +510,11 @@ pub fn plan_copy_item(
     // should be simplified. The reason they are currently separate code paths is so we can roll
     // out `COPY ... FROM <url>` without touching the current `COPY ... FROM ... STDIN` behavior.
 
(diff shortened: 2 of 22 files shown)


src/sql/src/plan/statement/dml.rs
```diff
@@ -1210,11 +1210,18 @@ fn plan_copy_from(
         bail_unsupported!("COPY FROM ... WITH (FILES ...) only supported from a URL")
     }
 
-    let (id, _, columns) = query::plan_copy_from(scx, table_name, columns)?;
+    let (id, source_desc, columns, maybe_mfp) = query::plan_copy_from(scx, table_name, columns)?;
+
+    let Some(mfp) = maybe_mfp else {
+        bail_unsupported!("COPY FROM ... expects an MFP to be produced");
```
petrosagg, line 1216: This will produce an error string like so: "COPY FROM ... expects an MFP to be produced not yet supported", which is probably confusing. Is this something users can trigger with syntax? If not what we do in other places in the planner is something like `sql_bail!("[internal error]: COPY FROM ... expects an MFP to be produced")`

src/sql/src/plan.rs
```diff
@@ -932,10 +934,21 @@ pub struct ShowColumnsPlan {
...
 pub struct CopyFromPlan {
+    /// Table we're copying into.
     pub id: CatalogItemId,
+    /// Source we're copying data from.
     pub source: CopyFromSource,
-    pub columns: Vec<usize>,
+    /// How input columns map to those on the destination table.
+    ///
+    /// TODO(cf2): Remove this field in favor of the mfp.
+    pub columns: Vec<ColumnIndex>,
```
petrosagg, line 944: Sounds like an easy win to fold this into the current PR. Is it difficult to add the projection into the MFP?

src/sql/src/plan/query.rs
```diff
@@ -476,27 +478,93 @@ pub fn plan_insert_query(
...
-    let mut desc = item
-        .desc(&scx.catalog.resolve_full_name(item.name()))?
-        .into_owned();
+    // TODO(cf2): The logic here to create the `source_desc` and the MFP are a bit duplicated and
+    // should be simplified. The reason they are currently separate code paths is so we can roll
+    // out `COPY ... FROM <url>` without touching the current `COPY ... FROM ... STDIN` behavior.
 
-    let mut ordering = Vec::with_capacity(columns.len());
+    // If we're copying data into a table, then we generate an MFP.
+    let mfp = if let Some(table_defaults) = item.writable_table_details() {
```
petrosagg, line 514: The comment says that we do something in case the item is a table but the code checks something more specific. I'm not super familiar with this but my initial doubt is whether there are tables for which writable_table_details returns None. Can this happen? Should we check if the item is a table and then assert that we get table_defaults inside?

src/storage-operators/src/oneshot_source.rs
```diff
@@ -998,6 +1027,10 @@ pub enum StorageErrorXKind {
     MissingSize,
     #[error("object is missing the required '{0}' field")]
     MissingField(Arc<str>),
+    #[error("the supplied mfp filtered out a row, it is only supposed to map and project")]
```
petrosagg, line 1030: should we also prefix that with [internal error]? A user won't understand what an mfp is if this ever makes it to the console

petrosagg (commented):
Looks great overall! I was looking for tests to understand how it's used and I was about to request that they are added before I noticed the other stacked PR.

Should we merge the two PRs here? I saw the stacked commit also had Rust changes so I think it makes sense to review this as a unit.

src/sql/src/plan/statement/dml.rs
ParkMyCar: Users should not be able to trigger this syntax, it's solely a programming error. Changed the returned error to be what you suggested!

src/sql/src/plan.rs
ParkMyCar: It would be a bit tricky to fold this into the current PR because the current `COPY FROM ... STDIN` still uses these columns. We have https://github.com/MaterializeInc/database-issues/issues/8859 tracking the work to fold all of that into this new setup

src/sql/src/plan/query.rs
ParkMyCar: It can happen! For example objects created via `CREATE TABLE ... FROM SOURCE` will return `None` here. I updated the comment to explain some more of that context

src/storage-operators/src/oneshot_source.rs
ParkMyCar: Done!

ParkMyCar:
> Looks great overall! I was looking for tests to understand how it's used and I was about to request that they are added before I noticed the other stacked PR.
> 
> Should we merge the two PRs here? I saw the stacked commit also had Rust changes so I think it makes sense to review this as a unit.

Good idea! I merged https://github.com/MaterializeInc/materialize/pull/31581 into here

bkirwi (approved):
Clever to use the MFP machinery for this!

src/storage-operators/src/oneshot_source.rs
```diff
@@ -998,6 +1030,10 @@ pub enum StorageErrorXKind {
     MissingSize,
     #[error("object is missing the required '{0}' field")]
     MissingField(Arc<str>),
+    #[error("[internal error] the supplied mfp filtered out a row, it is only supposed to map and project")]
+    MfpFiltered,
```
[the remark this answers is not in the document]
ParkMyCar: Fair! Updated to use a `soft_assert!` but otherwise not error

ParkMyCar:
TFTR!

PR #31596, opened 2025-02-24
Fix `EXPLAIN` for constant views

ggevay:
The first commit fixes `EXPLAIN LOCALLY OPTIMIZED PLAN` (https://github.com/MaterializeInc/database-issues/issues/8985).

The second commit renames `explain/view.slt` to `explain/locally_optimized_plan.slt`, because now the file is explaining not just views. Also, it moves out the one `EXPLAIN RAW PLAN` statement from it to `explain/raw_plan_as_text.slt`.

The third commit is somewhat unrelated: just adds a test for `EXPLAIN` without an explicit stage or format, so that we test the default. (Currently defaults to `EXPLAIN OPTIMIZED PLAN AS TEXT FOR`.) This is to prevent issues like the one fixed in https://github.com/MaterializeInc/materialize/pull/31482 (cc @def-, @mgree)

### Motivation

  * This PR fixes a recognized bug: https://github.com/MaterializeInc/database-issues/issues/8985

### Checklist

### DIFF
--- a/src/adapter/src/optimize/view.rs
+++ b/src/adapter/src/optimize/view.rs
@@ -78,6 +78,8 @@ impl Optimize<HirRelationExpr> for Optimizer {
         // MIR ⇒ MIR optimization (local)
         let expr = if expr.as_const().is_some() {
             // No need to optimize further, because we already have a constant.
+            // But trace this at "local", so that `EXPLAIN LOCALLY OPTIMIZED PLAN` can pick it up.
+            trace_plan!(at: "local", &expr);
             OptimizedMirRelationExpr(expr)
         } else {
             // Call the real optimization.
--- a/src/catalog/src/memory/objects.rs
+++ b/src/catalog/src/memory/objects.rs
@@ -1309,7 +1309,7 @@ pub struct View {
     pub global_id: GlobalId,
     /// Unoptimized high-level expression from parsing the `create_sql`.
     pub raw_expr: Arc<HirRelationExpr>,
-    /// Optimized mid-level expression from optimizing the `raw_expr`.
+    /// Optimized mid-level expression from (locally) optimizing the `raw_expr`.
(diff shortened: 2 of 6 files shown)


src/adapter/src/optimize/view.rs
```diff
@@ -78,6 +78,8 @@ impl Optimize<HirRelationExpr> for Optimizer {
         // MIR ⇒ MIR optimization (local)
         let expr = if expr.as_const().is_some() {
             // No need to optimize further, because we already have a constant.
+            // But trace this at "local", so that `EXPLAIN LOCALLY OPTIMIZED PLAN` can pick it up.
+            trace_plan!(at: "local", &expr);
```
ParkMyCar, line 82: I don't know much about the `trace_plan!` macro, but if it's okay to run at this location then LGTM!

ParkMyCar (approved):

src/adapter/src/optimize/view.rs
ggevay: Yes, it's intended to be run at any moment where we need to emit an extra plan tracing event. Usually, plan tracing events are emitted by `mz_ore::instrument` annotations on functions ([example](https://github.com/MaterializeInc/materialize/blob/7893991ee11dd9601bce5592e0c73bcb1f42f365/src/transform/src/predicate_pushdown.rs#L124)), but sometimes we need to emit some extra ones, in which case we use `trace_plan!`.

(Plan tracing events are used by EXPLAIN, which usually just picks out one tracing event, which corresponds to the stage requested by the EXPLAIN statement, e.g., EXPLAIN LOCALLY OPTIMIZED PLAN picks out "local". Plan tracing events are also used by https://optimizer-trace.dev.materialize.com/ , which shows all the plan tracing events in a tree structure.)

PR #31602, opened 2025-02-24
[sql_server] Add `sqlserver-util` crate that provides a wrapper around the `tiberius` library

ParkMyCar:
This PR introduces a new crate `sql-server-util` which adds a wrapper around a `tiberius::Client` to do automatic SQL transaction management. Specifically, if a `mz_sql_server_util::Client` is dropped, e.g. something returned an error, any open transaction will be rolled back.

It also changes testdrive to use this new wrapper client, and runs the stubbed SQL Server CDC tests in CI.

### Motivation

### Checklist


src/sql-server-util/src/lib.rs
```diff
@@ -0,0 +1,519 @@
...
+///
+/// When creating a [`Client`] we return a [`Connection`] which implements [`std::future::Future`]
+/// and must be polled for queries to make progress. Internally a [`Client`] holds the sending side
+/// of a channel and the [`Connection`] receives query requests to run. This enables us to
+/// introduce a [`Transaction`] type that when dropped will cause the `TRANSACTION` in the
+/// connected SQL Server instance to get rolled back.
+#[derive(Debug)]
+pub struct Client {
+    tx: UnboundedSender<Request>,
+}
```
ParkMyCar, line 38: @SangJunBak if you're interested in reviewing this the meat of the change is in this file. Basically `mz_sql_server_util::Client` holds the producer side of an mpsc channel, and when it wants to issue a query sends a request down the channel to a single worker which holds the consumer side of the channel and a `tiberius::Client`.

What this allows us to do is when using a `mz_sql_sever_util::Client` create a `Transaction` object where the lifetime of the object in Rust matches the lifetime of the transaction in SQL Server. Specifically if the `mz_sql_server_util::Transaction` gets dropped, e.g. some random code hit an error while we have the transaction open, then we'll execute a `ROLLBACK TRANSACTION` on the SQL Server side to cleanup our open transaction.

This is useful because the snapshotting process for the SQL Server source will use a transaction and I want to make sure if some intermediate thing, e.g. decoding a row, fails that we successfully cleanup our state.

src/sql-server-util/src/lib.rs
```diff
@@ -0,0 +1,522 @@
...
+    /// to be sent.
+    ///
+    /// [`Future`]: std::future::Future
+    pub async fn query<'a, 'q>(
+        &'a mut self,
+        query: impl Into<Cow<'q, str>>,
+        params: &'q [&'q dyn tiberius::ToSql],
+    ) -> Result<SmallVec<[tiberius::Row; 1]>, anyhow::Error>
+    where
+        'a: 'q,
```
SangJunBak, line 114: We probably don't need `'q` given we're converting params into owned data anyways

src/sql-server-util/src/lib.rs
```diff
@@ -0,0 +1,522 @@
...
+            .collect();
+        let kind = RequestKind::Query {
+            query: query.into().to_string(),
+            params,
+        };
+        self.tx
+            .send(Request { tx, kind })
+            .context("sending request")?;
+
+        let response = rx.await.context("channel")?.context("execute")?;
```
SangJunBak, line 130: Should be "query"

SangJunBak (approved):
Small nits but otw lgtm!

src/sql-server-util/src/lib.rs
ParkMyCar: Fixed!

src/sql-server-util/src/lib.rs
ParkMyCar: Good catch, fixed!

PR #31605, opened 2025-02-24
[storage] Fix and re-enable alter sink tests

bkirwi:
From our docs:

> When using the Avro format with a schema registry, the generated Avro schema for the new relation must be compatible with the previously published schema. If that’s not the case, the ALTER SINK command will succeed, but the subsequent execution of the sink will result in errors and will not be able to make progress.

And the avro docs:

> [...] if both are records: the ordering of fields may be different: fields are matched by name.

So to check compatibility, we should check by name instead of by order.

### Motivation

Fixes: https://github.com/MaterializeInc/database-issues/issues/8445

### Tips for reviewer

I've opted out JSON from any checks, since I don't think any rules are enforced for that type of sink, and that should help get our test coverage up.

### Checklist

### DIFF
--- a/misc/python/materialize/parallel_workload/action.py
+++ b/misc/python/materialize/parallel_workload/action.py
@@ -823,6 +823,10 @@ class AlterKafkaSinkFromAction(Action):
                     if len(o.columns) == 1
                     and o.columns[0].data_type == old_object.columns[0].data_type
                 ]
+            elif sink.format in ["FORMAT JSON"]:
+                # We should be able to format all data types as JSON, and they have no
+                # particular backwards-compatiblility requirements.
+                objs = [o for o in exe.db.db_objects_without_views()]
             else:
                 # Avro schema migration checking can be quite strict, and we need to be not only
                 # compatible with the latest object's schema but all previous schemas.
--- a/misc/python/materialize/parallel_workload/action.py
+++ b/misc/python/materialize/parallel_workload/action.py
@@ -824,18 +824,19 @@ class AlterKafkaSinkFromAction(Action):
                     and o.columns[0].data_type == old_object.columns[0].data_type
                 ]
             else:
-                # multi column formats require at least as many columns as before
-                # columns also have to be of the same type, see database-issues#8385
-                objs = [
-                    o
-                    for o in exe.db.db_objects_without_views()
-                    # Webhook sources can include all headers, then we don't know the exact columns
-                    if not isinstance(old_object, WebhookSource)
-                    and not isinstance(o, WebhookSource)
-                    and len(o.columns) >= len(old_object.columns)
-                    and [c.data_type for c in o.columns[: len(old_object.columns)]]
-                    == [c.data_type for c in old_object.columns]
-                ]
+                # Avro schema migration checking can be quite strict, and we need to be not only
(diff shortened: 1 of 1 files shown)


bkirwi:
I've kicked off the parallel workload tests in Nightlies, but even if they pass it's possible I haven't got to the original problem: I think the original failures were flaky, and I wasn't able to reliably reproduce on a scratch machine. OTOH I feel confident enough that I'd like to re-enable the tests, and if we see future failures I can follow up on those directly.

misc/python/materialize/parallel_workload/action.py
```diff
@@ -800,7 +800,7 @@ def run(self, exe: Executor) -> bool:
 
 class AlterKafkaSinkFromAction(Action):
     def run(self, exe: Executor) -> bool:
-        if exe.db.scenario in (Scenario.Kill, Scenario.ZeroDowntimeDeploy):
+        if exe.db.scenario in (Scenario.ZeroDowntimeDeploy,):
```
def-, line 803: Why was this changed?

misc/python/materialize/parallel_workload/action.py
```diff
@@ -823,19 +823,24 @@ def run(self, exe: Executor) -> bool:
...
-                    if not isinstance(old_object, WebhookSource)
-                    and not isinstance(o, WebhookSource)
-                    and len(o.columns) >= len(old_object.columns)
-                    and [c.data_type for c in o.columns[: len(old_object.columns)]]
-                    == [c.data_type for c in old_object.columns]
-                ]
+                # Avro schema migration checking can be quite strict, and we need to be not only
+                # compatible with the latest object's schema but all previous schemas.
+                # Only allow a conservative case for now: where all types and names match.
+                objs = []
```
def-, line 834: That's a bit unfortunate, I guess we should make the FORMAT JSON more likely

def- (approved):
Test changes are mostly ok, except the one comment. Your nightly run was green, but here's another one: https://buildkite.com/materialize/nightly/builds/11258

misc/python/materialize/parallel_workload/action.py
bkirwi: Yeah... at least "schemas match" is a fairly common case IIUC even if it isn't the most interesting one.

misc/python/materialize/parallel_workload/action.py
bkirwi: Not an interesting reason: I was trying to re-enable this for as many scenarios as possible, and it seemed plausible to me that this or recent changes fixed whatever caused us to exclude it earlier.

But you definitely have more context than I do, and if you prefer I'm happy to leave this line alone!

def-: I think it will be slightly flaky with this change.

bkirwi: Aight, I'll change it back!

Is it easy to explain why `alter sink` is not expected to work in the kill scenario? I'm curious!

def-: https://github.com/MaterializeInc/database-issues/issues/8421
> I believe what happens is that we have interrupted queries during a kill and thus are not aware of the exact state of a table, thus we can't replace it.

PR #31606, opened 2025-02-25
disable db network policies by default in helm chart

jubrad:
### Motivation
Network policies don't work with self-managed deploys. This is in part due to the fact that all port forwarding will appear to come from local host, and our current NGINX setup for console isn't doing the tcp proxy headers that we rely on in our cloud product.

Need to check if this needs a corresponding cloud change

### Checklist


misc/helm-charts/operator/values.yaml
```diff
@@ -27,6 +27,9 @@ operator:
     # -- Whether to enable environmentd rbac checks
     # TODO: this is not yet supported in the helm chart
     authentication: false
+    # -- Whether to enable netowrk policies
+    # -- Network policies don't yet work in self-managed
+    disable_database_network_policies: true
```
alex-hunt-materialize, line 32: Nitpick, comment spelling.

Nitpick, negative bools are often confusing, especially if we are inverting that for the actual parameter passed to environmentd.

alex-hunt-materialize (approved):

misc/helm-charts/operator/templates/deployment.yaml
```diff
@@ -79,6 +79,9 @@ spec:
         {{- if not .Values.operator.features.authentication }}
         - "--disable-authentication"
         {{- end }}
+        {{- if not .Values.operator.features.databaseNetworkPolicies}}
+        - "--disable-database-network-policies"
```
alex-hunt-materialize, line 83: We're now flopping back and forth between positive and negative args. Can we get consistent?

jubrad: Rather than having this be a negative all the way through, which was also commented on as being less than ideal, I'm trying to follow the `--diable-authentication` flag which does this flopping.

PR #31622, opened 2025-02-25
self-managed-debug: Add port forwarding functionality

SangJunBak:
- Add `system_catalog_dumper.rs` module for internal SQL port forwarding
- Update `Cargo.toml` to include `tokio-postgres`
- Modify `main.rs` to include new command-line arguments and initial port forwarding call
- Implement `spawn_internal_sql_port_forwarding_process` function for Kubernetes port forwarding

TODO: Spawn tasks to scrape the catalog using tokio-postgres

### Motivation

Part of https://github.com/MaterializeInc/database-issues/issues/8908

### Checklist

### DIFF
--- a/src/self-managed-debug/src/main.rs
+++ b/src/self-managed-debug/src/main.rs
@@ -37,6 +37,7 @@ use mz_ore::cli::{self, CliConfig};
 use mz_ore::error::ErrorExt;
 
 mod k8s_resource_dumper;
+mod system_catalog_dumper;
 
 pub const BUILD_INFO: BuildInfo = build_info!();
 pub static VERSION: LazyLock<String> = LazyLock::new(|| BUILD_INFO.human_version(None));
@@ -51,6 +52,17 @@ pub struct Args {
     k8s_namespaces: Vec<String>,
     #[clap(long = "k8s-dump-secret-values", action = clap::ArgAction::SetTrue)]
     k8s_dump_secret_values: bool,
+    /// If true, the tool will not attempt to port-forward the SQL port.
+    #[clap(long = "skip-port-forward", action = clap::ArgAction::SetTrue)]
+    skip_port_forward: bool,
+    /// The SQL port we want to port-forward to.
+    /// By default, we will attempt to find this port by looking for an environmentd service with a port named "internal sql"
+    #[clap(long = "sql-target-port")]
+    sql_target_port: Option<i32>,
+    /// The port on the local machine that will be forwarded to sql-target-port.
+    /// By default, this will be the same as the target port.
+    #[clap(long = "sql-local-port")]
+    sql_local_port: Option<i32>,
 }
 
 #[derive(Clone)]
@@ -176,18 +188,37 @@ async fn run(context: Context) -> Result<(), anyhow::Error> {
         ),
     ]);
 
-    match create_k8s_client(context.args.k8s_context.clone()).await {
-        Ok(client) => {
-            for namespace in context.args.k8s_namespaces.clone() {
-                k8s_resource_dumper::dump_namespaced_resources(&context, &client, namespace).await;
-            }
-            k8s_resource_dumper::dump_cluster_resources(&context, &client).await;
-        }
+    let client = match create_k8s_client(context.args.k8s_context.clone()).await {
+        Ok(client) => client,
         Err(e) => {
             eprintln!("Failed to create k8s client: {}", e);
+            return Err(e);
         }
     };
 
+    for namespace in context.args.k8s_namespaces.clone() {
+        k8s_resource_dumper::dump_namespaced_resources(&context, &client, namespace).await;
+    }
+    k8s_resource_dumper::dump_cluster_resources(&context, &client).await;
+
+    let _port_forward_handle;
+    if !context.args.skip_port_forward {
+        // Find the namespace, service name, local port, and target port.
+        let port_forwarding_info =
+            system_catalog_dumper::get_sql_port_forwarding_info(&client, &context.args).await?;
+
+        _port_forward_handle = system_catalog_dumper::spawn_sql_port_forwarding_process(
+            &port_forwarding_info,
+            context.args.k8s_context.clone(),
+        );
+        // There may be a delay between when the port forwarding process starts and when it's ready to use.
+        // We wait a few seconds to ensure that port forwarding is ready.
+        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
+    }
+
+    // TODO: Use the forwarded port to dump the system catalog.
+
+    // Wait for all the describe processes to finish.
     join_all(describe_handles).await;
 
     // TODO: Compress files to ZIP
--- a/src/self-managed-debug/src/main.rs
+++ b/src/self-managed-debug/src/main.rs
@@ -48,23 +48,30 @@ pub static VERSION: LazyLock<String> = LazyLock::new(|| BUILD_INFO.human_version
 #[clap(name = "self-managed-debug", next_line_help = true, version = VERSION.as_str())]
 pub struct Args {
     // === Kubernetes options. ===
-    #[clap(long = "k8s-context", env = "KUBERNETES_CONTEXT")]
+    #[clap(long, env = "KUBERNETES_CONTEXT")]
     k8s_context: Option<String>,
-    #[clap(long = "k8s-namespace", required = true, action = clap::ArgAction::Append)]
+    #[clap(long= "k8s-namespace", required = true, action = clap::ArgAction::Append)]
     k8s_namespaces: Vec<String>,
-    #[clap(long = "k8s-dump-secret-values", action = clap::ArgAction::SetTrue)]
+    #[clap(long , action = clap::ArgAction::SetTrue)]
     k8s_dump_secret_values: bool,
+    // === Port forwarding options. ===
     /// If true, the tool will not attempt to port-forward the SQL port.
-    #[clap(long = "skip-port-forward", action = clap::ArgAction::SetTrue)]
+    #[clap(long , action = clap::ArgAction::SetTrue)]
     skip_port_forward: bool,
-    /// The SQL port we want to port-forward to.
-    /// By default, we will attempt to find this port by looking for an environmentd service with a port named "internal sql"
-    #[clap(long = "sql-target-port")]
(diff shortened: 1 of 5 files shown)


src/self-managed-debug/src/main.rs
```diff
@@ -176,18 +188,36 @@ async fn run(context: Context) -> Result<(), anyhow::Error> {
...
+        let port_forwarding_info =
+            system_catalog_dumper::get_sql_port_forwarding_info(&client, &context.args).await?;
+
+        let _port_forward_handle = system_catalog_dumper::spawn_sql_port_forwarding_process(
+            &port_forwarding_info,
+            context.args.k8s_context.clone(),
+        );
+        // There may be a delay between when the port forwarding process starts and when it's ready to use.
+        // We wait a few seconds to ensure that port forwarding is ready.
+        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
```
SangJunBak, line 215: The `wait` here is kind of sus, but it's what we do in mzcompose too. We could grep for the messages, but that couples us to the current kubernetes version.

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -0,0 +1,167 @@
...
+                            return Ok(SqlPortForwardingInfo {
+                                namespace: namespace.clone(),
+                                service_name,
+                                target_port: port,
+                                local_port: args.sql_local_port.unwrap_or(port),
+                            });
+                        }
+                    }
+                }
+            }
```
ParkMyCar, line 77: It seems like here we're doing some filtering, finding, and mapping! A great use for iterators! Personally I find the following more read-able and I think it's more idiomatic Rust.
```suggestion
        let services = services.list(&Default::default()).await?;

        // Check if any of the services in this namespace target this port.
        let maybe_port_info = services
            .iter()
            // Filter down to just services that have specifications.
            .filter_map(|service| {
                let spec = service.spec.as_ref()?;
                let service_name = service.metadata.name.as_ref()?;
                Some((spec, service_name))
            })
            // For those services, we just need the associated ports.
            .flat_map(|(spec, service_name)| {
                spec.ports
                    .iter()
                    .flatten()
                    .map(move |port| (port, service_name))
            })
            // Find the first service who is listening on our target port.
            .find_map(|(port_info, service_name)| {
                let service_port = port_info.port;
                // Default to forwarding the same local port to the target port.
                let local_port = args.sql_local_port.unwrap_or(service_port);

                if let Some(target_port) = &args.sql_target_port {
                    if *target_port == service_port {
                        return Some(SqlPortForwardingInfo {
                            namespace: namespace.to_owned(),
                            service_name: service_name.to_owned(),
                            target_port: service_port,
                            local_port,
                        });
                    }
                }

                if let Some(port_name) = &port_info.name {
                    if port_name.eq_ignore_ascii_case("internal")
                        && port_name.eq_ignore_ascii_case("sql")
                    {
                        return Some(SqlPortForwardingInfo {
                            namespace: namespace.clone(),
                            service_name: service_name.to_owned(),
                            target_port: service_port,
                            local_port,
                        });
                    }
                }

                None
            });

        if let Some(port_info) = maybe_port_info {
            return Ok(port_info);
        } else {
            tracing::debug!("no services in '{namespace}' matched target port")
        }
```

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -0,0 +1,167 @@
...
+    pub service_name: String,
+    pub target_port: i32,
+    pub local_port: i32,
+}
+
+pub async fn get_sql_port_forwarding_info(
+    client: &Client,
+    args: &Args,
+) -> Result<SqlPortForwardingInfo, anyhow::Error> {
+    for namespace in &args.k8s_namespaces {
```
ParkMyCar, line 39: Could there be two different services across two different namespaces that are targeting the same port? In that case it looks like we'd forward to... which ever namespace is alphabetically higher? Not opposed to merging as-is but it seems like we'd want to do this as a follow up?

Btw a technique I've grown to using is leaving notes like `TODO(debug_toolX): ...` where `X` signifies the milestone that we need to fix this by. e.g. Fix all `1`s before giving to any user, `2`s before GAing, and `3`s at some point in time

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -0,0 +1,167 @@
...
+        if let Err(err) = retry::Retry::default()
+            .max_duration(Duration::from_secs(60))
+            .retry_async(|retry_state| {
+                let k8s_context = k8s_context.clone();
+                let namespace = port_forwarding_info.namespace.clone();
+                let service_name = port_forwarding_info.service_name.clone();
+                let local_port = port_forwarding_info.local_port;
+                let target_port = port_forwarding_info.target_port;
+
+                println!(
```
ParkMyCar, line 105: Instead of `println!(...)` should we be using `tracing::info!`? Maybe a nice follow up?

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -0,0 +1,167 @@
...
+                    "Spawning port forwarding process for {} from ports {} -> {}",
+                    service_name, local_port, target_port
+                );
+                async move {
+                    let port_arg_str = format!("{}:{}", &local_port, &target_port);
+                    let service_name_arg_str = format!("services/{}", &service_name);
```
ParkMyCar, line 111: Feels like this string formatting might be nice to pull into separate functions, mainly so we can clearly document said format?

Here's an [example](https://github.com/MaterializeInc/materialize/blob/main/src/storage-operators/src/oneshot_source/util.rs#L12-L32) of where I've done something similar. No need to make a trait but I've found the associated documentation useful for myself because in a few weeks I forget the specifics 😅

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -0,0 +1,167 @@
...
+                    // the type checker.
+                    RetryResult::Ok(())
+                }
+            })
+            .await
+        {
+            eprintln!("{}", err);
+        }
+    })
+    .abort_on_drop()
```
ParkMyCar, line 166: Presumably we're calling `.abort_on_drop()` because we want to kill the `kubectl` process we spawned when the debug tool exits?

A pattern I've seen before which might be nice here is something like:
```
/// A handle to the `kubectl` process we spawned to do the port fowarding.
#[must_use]
struct PortForwardHandle {
  inner: Option<AbortOnDropHandle<()>>,
}

impl PortForwardHandle {
    // Shutdown/kill the `kubectl` process that is port forwarding for us.
    pub fn shutdown(self) {
        let handle = self.inner.take();
        drop(handle);
    }
}

impl Drop for PortForwardHandle {
    fn drop(&mut self) {
        if let Some(handle) = self.inner.take() {
            mz_ore::soft_panic_or_log!("forgot to shutdown the kubectl process");
            drop(handle);
        }
    }
}
```

This `PortForwardHandle` lightly forces you to manage the lifetime of the process yourself, and makes it a bit more obvious in code as to how all the pieces fit together.

ParkMyCar (approved):
If this does what you need it to do then I'm happy to merge! Would love to talk through some of the comments though, ideally add them as `TODO(debug_tool)`s, happy to fix them in follow up PRs

src/self-managed-debug/src/system_catalog_dumper.rs
SangJunBak: ohh makes sense to me

src/self-managed-debug/src/system_catalog_dumper.rs
SangJunBak: > Could there be two different services across two different namespaces that are targeting the same port?

Ah that's a good callout! I think it might make sense to change the interface of the `local_port` and `target_port` argument variables. Here's what I'm thinking:
1. Accept an `IpAddr` for the local port since they can listen on an IP that's not `localhost`
2. Instead of just `target_port`, they can override with a service name **and** a target port. Can even do validation to see if the service name is correct in the CLI. That way it's 1:1 with `kubectl port-forward`'s interface. 

How does that sound? P.S. we use services instead of pods since we want a stable IP.

>Btw a technique I've grown to using is leaving notes like TODO(debug_toolX): ... where X signifies the milestone that we need to fix this by. e.g. Fix all 1s before giving to any user, 2s before GAing, and 3s at some point in time

I like this a lot. I think I'll do it!

src/self-managed-debug/src/system_catalog_dumper.rs
SangJunBak: Very functional, I think it reads better too! Especially since most TypeScript is written like this too

src/self-managed-debug/src/system_catalog_dumper.rs
SangJunBak: It kind of feels overkill creating a function per format function, especially since we're not reusing it anywhere else. I wonder if a comment is better?

src/self-managed-debug/src/system_catalog_dumper.rs
SangJunBak: Actually I realized we don't even need the `abort_on_drop` since what really matters is `.kill_on_drop(true)` on line 129. When the program ends, the task gets aborted anyways. I had a few questions about your suggestion tho:

1. Were you suggesting we call `shutdown` at the end of the program? Then if it doesn't get shutdown, we at least have the `Drop` implementation?
2. If we were to implement this pattern for `tokio::process::Command::new`, it seems like we'd need to pass the reference to the `Command` up to main to call the equivalent `shutdown` method. Is it possible to do so without a channel / mutex?

src/self-managed-debug/src/main.rs
SangJunBak: Thought about how we could implement this, and I think there needs to be two functions.

The first one like `init_port_forward` where it's an async function that waits until 
```
Forwarding from 127.0.0.1:58203 -> 8080
Forwarding from [::1]:58203 -> 8080
```
gets printed, then returns the tokio process. Then the second one like `retry_loop(tokio_process)` that repeatedly checks if the process has errored/exited – via `retry_async`– then restarts the process. Key thing being we don't await on `retry_loop`.

I don't think we can do this all in one function since if we awaited it, then that implies that the retry loop finished. Is there maybe a better implementation tho?

PR #31633, opened 2025-02-26
self_hosted: Update Persist's list of `LTS_VERSIONS`

ParkMyCar:
While debugging a legacy upgrade test I noticed that Persist's list of `LTS_VERSIONS` does not match the [list we maintain in Python](https://github.com/MaterializeInc/materialize/blob/main/misc/python/materialize/version_list.py#L30-L34). This PR updates Persist's list to match Python.

This difference could be intentional! In which case please ignore me :) But figured I would toss up this PR in case it's a bug.

### Motivation

Fix missing LTS version in Persist

### Checklist

### DIFF
--- a/src/persist-client/src/cfg.rs
+++ b/src/persist-client/src/cfg.rs
@@ -37,8 +37,10 @@ use crate::project::OPTIMIZE_IGNORED_DATA_DECODE;
 use crate::read::READER_LEASE_DURATION;
 
 const LTS_VERSIONS: &[Version] = &[
-    // 25.1
-    Version::new(0, 130, 0),
+    // 25.1.0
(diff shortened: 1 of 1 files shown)


doy-materialize:
yeah, this isn't correct - we guarantee that patch releases can't break compatibility, and so this is just intended to be a mapping between minor versions (and the logic using this list expects it to always contain versions with the patch version being 0).

ParkMyCar:
> yeah, this isn't correct - we guarantee that patch releases can't break compatibility, and so this is just intended to be a mapping between minor versions (and the logic using this list expects it to always contain versions with the patch version being 0).

Makes sense, thanks @doy-materialize!

PR #30360, opened 2024-11-06
compute: Improve the Union-Negate consolidation heuristic

ggevay:
DRAFT

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #31643, opened 2025-02-26
[design doc] `EXPLAIN` in Postgres syntax

mgree:
An outline of how to implement the last stage of https://github.com/MaterializeInc/database-issues/issues/8889.

[Rendered document](https://github.com/mgree/materialize/blob/explain-postgres-syntax/doc/developer/design/20250226_postgres_style_explain.md).

### Motivation

### Checklist

### DIFF
--- a/doc/developer/design/20250226_postgres_style_explain.md
+++ b/doc/developer/design/20250226_postgres_style_explain.md
@@ -76,32 +76,33 @@ We will need three pieces of work, which should all land together:
 
 ### Concrete Mapping
 
-| LIR node    | `mz_lir_mapping` node                    | New, Postgres-style syntax              |
-| :---------- | :--------------------------------------- | :-------------------------------------- |
-| `Constant`  | `Constant`                               | `Constant`                              |
-| `Get`       | `Get::PassArrangements l0`               | `Index Scan on l0 using ...`            |
-| `Get`       | `Get::Arrangement l0 (val=...)`          | `Index Lookup on l0 using ...`          |
-| `Get`       | `Get::Collection l0`                     | `Read l0`                               |
-| `Mfp`       | `MapFilterProject`                       | `Map/Filter/Project`                    |
-| `FlatMap`   | `FlatMap`                                | `Flat Map`                              |
-| `Join`      | `Join::Differential`                     | `Differential Join`                     |
-| `Join`      | `Join::Delta`                            | `Delta Join`                            |
-| `Reduce`    | `Reduce::Distinct`                       | `Distinct GroupAggregate`               |
-| `Reduce`    | `Reduce::Accumulable`                    | `Accumulable GroupAggregate`            |
-| `Reduce`    | `Reduce::Hierarchical (monotonic)`       | `Monotonic Hierarchical GroupAggregate` |
-| `Reduce`    | `Reduce::Hierarchical (buckets: ...)`    | `Bucketed Hierarchical GroupAggregate`  |
-| `Reduce`    | `Reduce::Basic`                          | `Non-incremental GroupAggregate`        |
-| `Reduce`    | `Reduce::Collation`                      | `Collated GroupAggregate` (details?)    |
-| `TopK`      | `TopK::MonotonicTop1`                    | `Monotonic Top1`                        |
-| `TopK`      | `TopK::MonotonicTopK`                    | `Monotonic TopK`                        |
-| `TopK`      | `TopK::Basic`                            | `Non-monotonic TopK`                    |
-| `Negate`    | `Negate`                                 | `Negate Diffs`                          |
-| `Threshold` | `Threshold`                              | `Threshold Diffs`                       |
-| `Union`     | `Union`                                  | `Union`                                 |
-| `Union`     | `Union (consolidates output)`            | `Consolidating Union`                   |
-| `ArrangeBy` | `Arrange`                                | `Arrange`                               |
-| `Let`       | `e0 With l1 = e1 ...`                    | `e1 With l1 = e1 ...`                   |
-| `LetRec`    | `e0 With Mutually Recursive l1 = e1 ...` | `e0 With Mutually Recursve l1 = e1 ...` |
+| LIR node    | `mz_lir_mapping` node                    | New, Postgres-style syntax                           |
+| :---------- | :--------------------------------------- | :--------------------------------------------------- |
+| `Constant`  | `Constant`                               | `Constant`                                           |
+| `Get`       | `Get::PassArrangements l0`               | `Index Scan on l0 using ...` or `Stream Scan on l0`  |
+| `Get`       | `Get::Arrangement l0 (val=...)`          | `Index Lookup on l0 using ...`                       |
+| `Get`       | `Get::Arrangement l0`                    | `Index Scan on l0 using ...` (showing mfp)           |
+| `Get`       | `Get::Collection l0`                     | `Read l0`                                            |
+| `Mfp`       | `MapFilterProject`                       | `Map/Filter/Project`                                 |
+| `FlatMap`   | `FlatMap`                                | `Table Function`                                     |
+| `Join`      | `Join::Differential`                     | `Differential Join`                                  |
+| `Join`      | `Join::Delta`                            | `Delta Join`                                         |
+| `Reduce`    | `Reduce::Distinct`                       | `Distinct GroupAggregate`                            |
+| `Reduce`    | `Reduce::Accumulable`                    | `Accumulable GroupAggregate`                         |
+| `Reduce`    | `Reduce::Hierarchical (monotonic)`       | `Monotonic Hierarchical GroupAggregate`              |
+| `Reduce`    | `Reduce::Hierarchical (buckets: ...)`    | `Bucketed Hierarchical GroupAggregate`               |
+| `Reduce`    | `Reduce::Basic`                          | `Non-incremental GroupAggregate`                     |
+| `Reduce`    | `Reduce::Collation`                      | `Collated GroupAggregate` (details?)                 |
+| `TopK`      | `TopK::MonotonicTop1`                    | `Monotonic Top1`                                     |
+| `TopK`      | `TopK::MonotonicTopK`                    | `Monotonic TopK`                                     |
+| `TopK`      | `TopK::Basic`                            | `Non-monotonic TopK`                                 |
+| `Negate`    | `Negate`                                 | `Negate Diffs`                                       |
+| `Threshold` | `Threshold`                              | `Threshold Diffs`                                    |
+| `Union`     | `Union`                                  | `Union`                                              |
+| `Union`     | `Union (consolidates output)`            | `Consolidating Union`                                |
+| `ArrangeBy` | `Arrange`                                | `Arrange`                                            |
+| `Let`       | `e0 With l1 = e1 ...`                    | `e1 With l1 = e1 ...`                                |
+| `LetRec`    | `e0 With Mutually Recursive l1 = e1 ...` | `e0 With Mutually Recursve l1 = e1 ...`              |
 
 In the new Postgres-style syntax, extra information will appear on the
 next line: for joins, it will be the join pipelines; for
@@ -111,6 +112,23 @@ filters.
 For `Delta Join` in particular, we will want to push information
 further down in the listing; see [TPC-H query 3](#tpc-h-query-3) below
 for an example.
+
+### Formerly Open Questions
+
+**Should we show `Project`? Should we show _all_ expressions for `Map`
+and `Filter`?** Yes: we will show all Mfp expressions by default.
+
+**How much of this data should `mz_lir_mapping` show?** I propose
+showing the first line plus anything involving scalar expressions
+(e.g., `Map/Filter/Project`s, `Join` equivalences, etc.).
+
+**What about names?** Separate efforts
+([#31802](https://github.com/MaterializeInc/materialize/pull/31802)
+will help us [get more column
+names](https://github.com/MaterializeInc/database-issues/issues/8960)). Showing
+_only_ column names (without numbers) can induce some confusion when
+we have self-joins, as in outer-join lowering. We will want to add
+context (e.g., the table alias, `f1.f_col = f2.f_col`).
 
 ## Minimal Viable Prototype
 
@@ -185,17 +203,17 @@ New Materialize `EXPLAIN`:
   Finish
     Order by: l_returnflag, l_linestatus
     ->  Project (columns=10)
-        Columns: l_returnflag..=sum, #9..=#11, count
-        -> Map (columns=12)
-           (bigint_to_numeric(case when (count = 0) then null else count end), (sum_l_quantity / #8), (sum_l_extendedprice / #8), (sum_l_discount / #8))
-           -> Accumulable GroupAggregate (columns=8)
-              Group Key: l_returnflag, l_linestatus
-              Aggregates: sum(l_quantity), sum(l_extendedprice), sum((l_extendedprice * (1 - l_discount))), sum(((l_extendedprice * (1 - l_discount)) * (1 + l_tax))), count(*), sum(l_discount)
-              -> Project (columns=6)
-                 Columns: l_quantity..=l_linestatus
-                 -> Filter (columns=16)
-                    Predicates: date_to_timestamp(l_shipdate) <= 1998-10-02 00:00:00
-                      -> Index Scan using pk_lineitem_orderkey_linenumber on lineitem (columns=16)
+          Columns: l_returnflag..=sum, #9..=#11, count
+          -> Map (columns=12)
+             (bigint_to_numeric(case when (count = 0) then null else count end), (sum_l_quantity / #8), (sum_l_extendedprice / #8), (sum_l_discount / #8))
+             -> Accumulable GroupAggregate (columns=8)
+                  Group Key: l_returnflag, l_linestatus
+                  Aggregates: sum(l_quantity), sum(l_extendedprice), sum((l_extendedprice * (1 - l_discount))), sum(((l_extendedprice * (1 - l_discount)) * (1 + l_tax))), count(*), sum(l_discount)
+                  -> Project (columns=6)
+                       Columns: l_quantity..=l_linestatus
+                       -> Filter (columns=16)
+                            Predicates: date_to_timestamp(l_shipdate) <= 1998-10-02 00:00:00
+                            -> Index Scan using pk_lineitem_orderkey_linenumber on lineitem (columns=16)
 
 Used Indexes:
   - materialize.public.pk_lineitem_orderkey_linenumber (*** full scan ***)
@@ -285,34 +303,34 @@ New Materialize `EXPLAIN`:
 
 ```
 Finish
-  Order by: sum desc nulls_first, o_orderdate
+  Order by: sum desc, o_orderdate
   -> Project (columns=4)
-     Columns: o_orderkey, sum, o_orderdate, o_shippriority
-     -> Reduce (columns=4)
-        Group key: o_orderkey..=#2o_shippriority
-        Aggregates: sum((l_extendedprice * (1 - l_discount)))
-        -> Project (columns=5)
-           Columns: o_orderkey, o_orderdate, o_shippriority, l_extendedprice, l_discount
-           -> Filter (columns=33)
-              Predicates: (c_mktsegment = "BUILDING") AND (o_orderdate < 1995-03-15) AND (l_shipdate > 1995-03-15)
-              -> Delta Join (columns=33)
-                 Conditions: c_custkey = o_custkey AND o_orderkey = l_orderkey
-                 Pipelines:
-                   %0:customer » %1:orders[#1]KAif » %2:lineitem[#0]KAif
-                   %1:orders » %0:customer[#0]KAef » %2:lineitem[#0]KAif
-                   %2:lineitem » %1:orders[#0]KAif » %0:customer[#0]KAef
-                 -> Arranged (columns=8)
-                    Keys: [c_custkey]
-                    -> Index Scan using pk_customer_custkey on customer (columns=8)
-                       Delta join first input (full scan): pk_customer_custkey
-                 -> Arrange (columns=9)
-                    Keys: [o_orderkey], [o_custkey]
-                    -> Index Scan using pk_orders_orderkey, fk_orders_custkey on orders (columns=9)
-                       Delta join lookup: pk_orders_orderkey, fk_orders_custkey
-                 -> Arrange (columns=16)
-                    Keys: [l_orderkey]
-                    -> Index Scan using fk_lineitem_orderkey on lineitem (columns=16)
-                       Delta join lookup: fk_lineitem_orderkey
+       Columns: o_orderkey, sum, o_orderdate, o_shippriority
+       -> Reduce (columns=4)
+            Group key: o_orderkey..=#2o_shippriority
+            Aggregates: sum((l_extendedprice * (1 - l_discount)))
+            -> Project (columns=5)
+                 Columns: o_orderkey, o_orderdate, o_shippriority, l_extendedprice, l_discount
+                 -> Filter (columns=33)
+                      Predicates: (c_mktsegment = "BUILDING") AND (o_orderdate < 1995-03-15) AND (l_shipdate > 1995-03-15)
+                      -> Delta Join (columns=33)
+                           Conditions: c_custkey = o_custkey AND o_orderkey = l_orderkey
+                           Pipelines:
+                             %0:customer » %1:orders[#1]KAif » %2:lineitem[#0]KAif
+                             %1:orders » %0:customer[#0]KAef » %2:lineitem[#0]KAif
+                             %2:lineitem » %1:orders[#0]KAif » %0:customer[#0]KAef
+                           -> Arranged (columns=8)
+                                Keys: [c_custkey]
+                                -> Index Scan using pk_customer_custkey on customer (columns=8)
+                                   Delta join first input (full scan): pk_customer_custkey
+                           -> Arrange (columns=9)
+                                Keys: [o_orderkey], [o_custkey]
+                                -> Index Scan using pk_orders_orderkey, fk_orders_custkey on orders (columns=9)
+                                   Delta join lookup: pk_orders_orderkey, fk_orders_custkey
+                           -> Arrange (columns=16)
+                                Keys: [l_orderkey]
+                                -> Index Scan using fk_lineitem_orderkey on lineitem (columns=16)
+                                   Delta join lookup: fk_lineitem_orderkey
 
 Used Indexes:
   - materialize.public.pk_customer_custkey (delta join 1st input (full scan))
@@ -327,12 +345,8 @@ Target cluster: quickstart
 
 Should we more radically reduce the AST?
 
-Should we abandon static `EXPLAIN` and encourage `mz_lir_mapping` use?
+~~Should we abandon static `EXPLAIN` and encourage `mz_lir_mapping`
+use?~~ No: being able to `EXPLAIN` ahead of time is valuable, and also
+`mz_lir_mapping` forces us to munge strings in SQL.
 
 ## Open questions
-
-Should we show `Project`?
-
-Should we show _all_ expressions for `Map` and `Filter`?
-
-How much of this data should `mz_lir_mapping` show?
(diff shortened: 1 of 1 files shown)


doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+
+`EXPLAIN` is meant to help users understand how Materialize actually
+runs their queries. In the name of streamlining the education process,
+we should make our output as much like Postgres's as is practicable.
+
+Changing `EXPLAIN` is tricky, though: we rely heavily on `EXPLAIN`'s
+completionist output to test our optimizer and debug queries. We must
+be careful to keep these tests while enabling the new behavior.
+
+[https://github.com/MaterializeInc/materialize/pull/31185
```
ggevay, line 15: Missing `]`. Also, maybe you wanted to make this into a sentence?

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+### Concrete Mapping
+
+| LIR node    | `mz_lir_mapping` node                    | New, Postgres-style syntax              |
+| :---------- | :--------------------------------------- | :-------------------------------------- |
+| `Constant`  | `Constant`                               | `Constant`                              |
+| `Get`       | `Get::PassArrangements l0`               | `Index Scan on l0 using ...`            |
+| `Get`       | `Get::Arrangement l0 (val=...)`          | `Index Lookup on l0 using ...`          |
+| `Get`       | `Get::Collection l0`                     | `Read l0`                               |
+| `Mfp`       | `MapFilterProject`                       | `Map/Filter/Project`                    |
+| `FlatMap`   | `FlatMap`                                | `Flat Map`                              |
```
ggevay, line 86: Does Postgres have `Flat Map`? If not, we could consider `Table Function`.

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+| `Join`      | `Join::Differential`                     | `Differential Join`                     |
+| `Join`      | `Join::Delta`                            | `Delta Join`                            |
+| `Reduce`    | `Reduce::Distinct`                       | `Distinct GroupAggregate`               |
```
ggevay, line 89: Why not just `Distinct`?

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+| `Reduce`    | `Reduce::Hierarchical (buckets: ...)`    | `Bucketed Hierarchical GroupAggregate`  |
+| `Reduce`    | `Reduce::Basic`                          | `Non-incremental GroupAggregate`        |
+| `Reduce`    | `Reduce::Collation`                      | `Collated GroupAggregate` (details?)    |
+| `TopK`      | `TopK::MonotonicTop1`                    | `Monotonic Top1`                        |
+| `TopK`      | `TopK::MonotonicTopK`                    | `Monotonic TopK`                        |
+| `TopK`      | `TopK::Basic`                            | `Non-monotonic TopK`                    |
+| `Negate`    | `Negate`                                 | `Negate Diffs`                          |
+| `Threshold` | `Threshold`                              | `Threshold Diffs`                       |
+| `Union`     | `Union`                                  | `Union`                                 |
+| `Union`     | `Union (consolidates output)`            | `Consolidating Union`                   |
```
ggevay, line 101: I'm wondering whether users need to know about whether a `Union` consolidates. Seems like a somewhat esoteric implementation-detail, so maybe not. Also, I can't really think of a situation where user would need to make some decision based on knowing whether a Union consolidates or not.

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+   using this output everywhere `EXPLAIN` is used.
+
+### Concrete Mapping
+
+| LIR node    | `mz_lir_mapping` node                    | New, Postgres-style syntax              |
+| :---------- | :--------------------------------------- | :-------------------------------------- |
+| `Constant`  | `Constant`                               | `Constant`                              |
+| `Get`       | `Get::PassArrangements l0`               | `Index Scan on l0 using ...`            |
+| `Get`       | `Get::Arrangement l0 (val=...)`          | `Index Lookup on l0 using ...`          |
+| `Get`       | `Get::Collection l0`                     | `Read l0`                               |
```
ggevay, line 84: I think `GetPlan`'s correspondence to reality is a bit more complicated. I don't have the time today to delve into this, but here are my raw notes:

GetPlan is unncessarily hard to understand:
- GetPlan::PassArrangements:
  - no MFP
  - ! BUT: there might or might not be an input arrangement that we are passing (so it's not always an `Index Scan`)
- GetPlan::Arrangement:
  - there is an input arrangement
  - no output arrangement
  - there is an MFP
  - we might be seeking a key (but that code path is deprecated, so it occurs very rarely; still occurs in aggregates.slt and is_null_propagation.slt) (so it's usually not an `Index Lookup`)
- GetPlan::Collection:
  - no input arrangement (and no output arrangement)
  - currently, there is always an MFP (but this will change when we change PassArrangements)

possible refactoring:
- GetPlan::PassArrangements -- add an if to only produce it if there is an arrangement
- split off a new lookup variant from GetPlan::Arrangement
- rename GetPlan variants
- Get::keys -- make it clear that these are _output_ arrangements

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+| `Negate`    | `Negate`                                 | `Negate Diffs`                          |
+| `Threshold` | `Threshold`                              | `Threshold Diffs`                       |
+| `Union`     | `Union`                                  | `Union`                                 |
+| `Union`     | `Union (consolidates output)`            | `Consolidating Union`                   |
+| `ArrangeBy` | `Arrange`                                | `Arrange`                               |
+| `Let`       | `e0 With l1 = e1 ...`                    | `e1 With l1 = e1 ...`                   |
+| `LetRec`    | `e0 With Mutually Recursive l1 = e1 ...` | `e0 With Mutually Recursve l1 = e1 ...` |
+
+In the new Postgres-style syntax, extra information will appear on the
+next line: for joins, it will be the join pipelines; for
```
ggevay, line 107: LIR has MFPs inside pipelines. We'll need to decide what to do with these. (MIR doesn't have these, so the current MIR EXPLAIN is often lying by saying that the MFP is executed after the join, when in fact the lowering pushes the MFP inside the pipelines.)

Maybe we can just leave off all the joins paths by default, unless `WITH (JOIN IMPLEMENTATIONS)` is specified. But the Filters (and also maps and projects if we generally decide to show them) will have to appear somewhere.

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+| `Reduce`    | `Reduce::Basic`                          | `Non-incremental GroupAggregate`        |
+| `Reduce`    | `Reduce::Collation`                      | `Collated GroupAggregate` (details?)    |
+| `TopK`      | `TopK::MonotonicTop1`                    | `Monotonic Top1`                        |
+| `TopK`      | `TopK::MonotonicTopK`                    | `Monotonic TopK`                        |
+| `TopK`      | `TopK::Basic`                            | `Non-monotonic TopK`                    |
+| `Negate`    | `Negate`                                 | `Negate Diffs`                          |
+| `Threshold` | `Threshold`                              | `Threshold Diffs`                       |
+| `Union`     | `Union`                                  | `Union`                                 |
+| `Union`     | `Union (consolidates output)`            | `Consolidating Union`                   |
+| `ArrangeBy` | `Arrange`                                | `Arrange`                               |
```
ggevay, line 102: If there are no keys, but `raw` is true, then maybe think up another name for this case.

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+                   %2:lineitem » %1:orders[#0]KAif » %0:customer[#0]KAef
+                 -> Arranged (columns=8)
+                    Keys: [c_custkey]
+                    -> Index Scan using pk_customer_custkey on customer (columns=8)
+                       Delta join first input (full scan): pk_customer_custkey
+                 -> Arrange (columns=9)
+                    Keys: [o_orderkey], [o_custkey]
+                    -> Index Scan using pk_orders_orderkey, fk_orders_custkey on orders (columns=9)
+                       Delta join lookup: pk_orders_orderkey, fk_orders_custkey
+                 -> Arrange (columns=16)
```
ggevay, line 312: These `Arrange`s won't actually be here in this case, because they are not present in LIR, due to reusing existing indexes. (Which is a very good thing -- it was confusing as hell in the MIR EXPLAIN that ArrangeBy might not might not do a thing)

But we might want to still somehow show the keys that the join picked for each input.

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+
+Should we more radically reduce the AST?
+
+Should we abandon static `EXPLAIN` and encourage `mz_lir_mapping` use?
+
+## Open questions
+
+Should we show `Project`?
+
+Should we show _all_ expressions for `Map` and `Filter`?
```
ggevay, line 336: (https://github.com/MaterializeInc/database-issues/issues/8889#issuecomment-2612639970 discusses this in the penultimate paragraph. I'm not sure, but leaning towards yes.)

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+        -> Project (columns=5)
+           Columns: o_orderkey, o_orderdate, o_shippriority, l_extendedprice, l_discount
+           -> Filter (columns=33)
+              Predicates: (c_mktsegment = "BUILDING") AND (o_orderdate < 1995-03-15) AND (l_shipdate > 1995-03-15)
+              -> Delta Join (columns=33)
+                 Conditions: c_custkey = o_custkey AND o_orderkey = l_orderkey
+                 Pipelines:
+                   %0:customer » %1:orders[#1]KAif » %2:lineitem[#0]KAif
+                   %1:orders » %0:customer[#0]KAef » %2:lineitem[#0]KAif
+                   %2:lineitem » %1:orders[#0]KAif » %0:customer[#0]KAef
```
ggevay, line 303: We can probably leave these off by default, unless `WITH (JOIN IMPLEMENTATIONS)` is specified.

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+
+## Minimal Viable Prototype
+
+These examples are adapted from existing MIR explain plans, so they
+are not completely faithful to the language above (e.g., `Map` and
+`Filter` are separate, when they will be combined in
+`Map/Filter/Project`).
+
+Arity is included in the Postgres style (cf. "width="), though we will
+hopefully not need it when we have good column names.
```
ggevay, line 123: Yes, it could be off by default when we have good column names.

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+  - materialize.public.fk_lineitem_orderkey (delta join lookup)
+
+Target cluster: quickstart
+```
+
+## Alternatives
+
+Should we more radically reduce the AST?
+
+Should we abandon static `EXPLAIN` and encourage `mz_lir_mapping` use?
```
ggevay, line 330: I don't think that's viable, because there are a lot of minor things to take care of when implementing plan printing, and it would be hard to do all those in SQL.

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+  - materialize.public.fk_lineitem_orderkey (delta join lookup)
+
+Target cluster: quickstart
+```
+
+New Materialize `EXPLAIN`:
+
+```
+Finish
+  Order by: sum desc nulls_first, o_orderdate
```
ggevay, line 288: We could leave off `nulls_first`/`nulls_last` if they are at their default setting. (Which I think is `nulls_first` for `desc`, and `nulls_last` for `asc`.

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+  - materialize.public.fk_orders_custkey (delta join lookup)
+  - materialize.public.fk_lineitem_orderkey (delta join lookup)
+
+Target cluster: quickstart
+```
+
+New Materialize `EXPLAIN`:
+
+```
+Finish
```
ggevay, line 287: I'm not sure what Postgres prints for LIMIT/OFFSET, but we might try to match that too if it makes sense.

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,338 @@
...
+     Columns: o_orderkey, sum, o_orderdate, o_shippriority
+     -> Reduce (columns=4)
+        Group key: o_orderkey..=#2o_shippriority
+        Aggregates: sum((l_extendedprice * (1 - l_discount)))
+        -> Project (columns=5)
+           Columns: o_orderkey, o_orderdate, o_shippriority, l_extendedprice, l_discount
+           -> Filter (columns=33)
+              Predicates: (c_mktsegment = "BUILDING") AND (o_orderdate < 1995-03-15) AND (l_shipdate > 1995-03-15)
+              -> Delta Join (columns=33)
+                 Conditions: c_custkey = o_custkey AND o_orderkey = l_orderkey
```
ggevay, line 299: As mentioned elsewhere, we'll have a problem with self-joins: we'll be seeing conditions like `x = x`, where one `x` is from one join input, and the other `x` is from another join input. Even including the table name won't resolve it if it's a self-join. And we have a lot of self-joins due to outer join lowering / subquery lowering creating them.

A way out could be to write something like `%1.x = %2.x`, where `%1`, `%2`, ... are the join inputs.

How does Postgres resolve this btw.?

ggevay (commented):
I wrote some comments.

It looks good; hopefully the Postgres syntax will be more familiar to users.

And I'm happy that we are basing it on LIR!

doc/developer/design/20250226_postgres_style_explain.md
mgree: Happy to skip it! My understanding was that we sometimes cared about this in accounting for memory footprint, but maybe that's an us/`VERBOSE TEXT` thing?

doc/developer/design/20250226_postgres_style_explain.md
mgree: Just following Postgres, which always calls reduces a `GroupAggregate`:

```
EXPLAIN SELECT DISTINCT l_discount FROM lineitem;
                            QUERY PLAN                            
------------------------------------------------------------------
 HashAggregate  (cost=12.50..14.50 rows=200 width=18)
   Group Key: l_discount
   ->  Seq Scan on lineitem  (cost=0.00..12.00 rows=200 width=18)
(3 rows)
```

doc/developer/design/20250226_postgres_style_explain.md
mgree: Let's talk about this (maybe with @frankmcsherry?) and find the right way to factor this out so that things are actually clear.

doc/developer/design/20250226_postgres_style_explain.md
mgree: Not exactly:

```
michaelgreenberg=# EXPLAIN SELECT DISTINCT l_discount, generate_series(1, 5) from lineitem;
                               QUERY PLAN                               
------------------------------------------------------------------------
 HashAggregate  (cost=23.50..36.00 rows=1000 width=22)
   Group Key: l_discount, generate_series(1, 5)
   ->  ProjectSet  (cost=0.00..18.50 rows=1000 width=22)
         ->  Seq Scan on lineitem  (cost=0.00..12.00 rows=200 width=18)
(4 rows)
```

I'm fine with `Table Function` or `ProjectSet`.

doc/developer/design/20250226_postgres_style_explain.md
mgree: Yes, good idea! Feels related to the `Get` refactoring.

doc/developer/design/20250226_postgres_style_explain.md
mgree: I think we're going to want to see join order though. Postgres always shows join order (implicitly, via AST structure) along with the join conditions at each step of the join (`Hash Cond`, `Index Cond`).

doc/developer/design/20250226_postgres_style_explain.md
mgree: They seem to show `LIMIT` but not `OFFSET`:

```
michaelgreenberg=# EXPLAIN SELECT DISTINCT l_discount FROM lineitem LIMIT 10 OFFSET 15;
                               QUERY PLAN                               
------------------------------------------------------------------------
 Limit  (cost=12.65..12.75 rows=10 width=18)
   ->  HashAggregate  (cost=12.50..14.50 rows=200 width=18)
         Group Key: l_discount
         ->  Seq Scan on lineitem  (cost=0.00..12.00 rows=200 width=18)
(4 rows)
```

doc/developer/design/20250226_postgres_style_explain.md
mgree: They use the relevant aliases:

```
michaelgreenberg=# explain select l1.l_orderkey, l2.l_orderkey from lineitem l1, lineitem l2 where l1.l_partkey = l2.l_partkey and l1.l_orderkey <> l2.l_orderkey and l1.l_shipdate < l2.l_shipdate;
                                      QUERY PLAN                                       
---------------------------------------------------------------------------------------
 Hash Join  (cost=14.50..35.00 rows=66 width=8)
   Hash Cond: (l1.l_partkey = l2.l_partkey)
   Join Filter: ((l1.l_orderkey <> l2.l_orderkey) AND (l1.l_shipdate < l2.l_shipdate))
   ->  Seq Scan on lineitem l1  (cost=0.00..12.00 rows=200 width=12)
   ->  Hash  (cost=12.00..12.00 rows=200 width=12)
         ->  Seq Scan on lineitem l2  (cost=0.00..12.00 rows=200 width=12)
(6 rows)
```

I don't _think_ it's possible to have a self-join without such aliases, but I'm not 100% on that.

```
michaelgreenberg=# explain select * from lineitem join lineitem on (l_partkey);
ERROR:  table name "lineitem" specified more than once
```

doc/developer/design/20250226_postgres_style_explain.md
mgree: Fair enough!

doc/developer/design/20250226_postgres_style_explain.md
ggevay: Yeah, it's more for us, enough in `VERBOSE TEXT`. (But even I haven't looked at this in a long time. Currently, the heuristic for whether a Union consolidates is very simple: if there is at least 1 negated input, then we consolidate. If we were to make the heuristic more complex then we'd need to more often look at it. There was an abandoned attempt on this here: https://github.com/MaterializeInc/materialize/pull/30360 )

doc/developer/design/20250226_postgres_style_explain.md
ggevay: (Discussed in a meeting since then: Our lowering for outer joins / subqueries also creates self-joins. The solution will probably be to make the lowering think up synthetic aliases.)

doc/developer/design/20250226_postgres_style_explain.md
frankmcsherry: I vote `Table Function`!

doc/developer/design/20250226_postgres_style_explain.md
mgree: Ah, good point. Tried to update this, not 100% on it.

doc/developer/design/20250226_postgres_style_explain.md
mgree: I propose we write `Stream/Arrange` when `raw=true`, on the principle that we should indicate when we _could_ be streaming rather than using an arrangement.

mgree:
Okay, I tried to incorporate these changes---please take a look!

doc/developer/design/20250226_postgres_style_explain.md
ggevay: - If there are no `keys`, but there is `raw`, then it could be something like `Stream` or `Unarrange`.
- If it's just `keys` but no `raw`, then `Arrange` is fine.

But I'm not sure what to do with the weird cases:
- Both `keys` and `raw`: how to indicate that there are two things being produced? Or maybe it's enough to just say `Arrange`, because the `raw` part won't be important performance-wise in this case?
- Neither `keys` nor `raw`: hopefully this won't occur, because it doesn't make a whole lot of sense :D

doc/developer/design/20250226_postgres_style_explain.md
mgree: The changes in https://github.com/MaterializeInc/materialize/pull/31878 should make it so we have table aliases in joins: https://github.com/MaterializeInc/materialize/blob/8e29cdc91e355feb652c462a3c9bdc78232a2498/src/sql/src/plan/query.rs#L6625-L6633

doc/developer/design/20250226_postgres_style_explain.md
ggevay: I think it's fine to show it as `Distinct`. I'd say this is a clear improvement over Postgres not showing it as `Distinct`. There is no need to follow Postgres just for familiarity here, because we don't really need to rely on familiarity, as it should be pretty clear for users what `Distinct` is.

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,358 @@
...
+                      -> Delta Join (columns=33)
+                           Conditions: c_custkey = o_custkey AND o_orderkey = l_orderkey
+                           Pipelines:
+                             %0:customer » %1:orders[#1]KAif » %2:lineitem[#0]KAif
+                             %1:orders » %0:customer[#0]KAef » %2:lineitem[#0]KAif
+                             %2:lineitem » %1:orders[#0]KAif » %0:customer[#0]KAef
+                           -> Index Scan using pk_customer_custkey on customer (columns=8)
+                                Delta join first input (full scan): pk_customer_custkey
+                           -> Index Scan using pk_orders_orderkey, fk_orders_custkey on orders (columns=9)
+                                Delta join lookup: pk_orders_orderkey (%1), fk_orders_custkey (%0, %2)
```
ggevay, line 337: Do the `%` numbers mean which join path uses the index? (If yes, then there seems to be a mixup: `%0` and `%2` user different indexes.

doc/developer/design/20250226_postgres_style_explain.md
```diff
@@ -0,0 +1,358 @@
...
+| `Reduce`    | `Reduce::Hierarchical (buckets: ...)`    | `Bucketed Hierarchical GroupAggregate`               |
+| `Reduce`    | `Reduce::Basic`                          | `Non-incremental GroupAggregate`                     |
+| `Reduce`    | `Reduce::Collation`                      | `Collated GroupAggregate` (details?)                 |
+| `TopK`      | `TopK::MonotonicTop1`                    | `Monotonic Top1`                                     |
+| `TopK`      | `TopK::MonotonicTopK`                    | `Monotonic TopK`                                     |
+| `TopK`      | `TopK::Basic`                            | `Non-monotonic TopK`                                 |
+| `Negate`    | `Negate`                                 | `Negate Diffs`                                       |
+| `Threshold` | `Threshold`                              | `Threshold Diffs`                                    |
+| `Union`     | `Union`                                  | `Union`                                              |
+| `Union`     | `Union (consolidates output)`            | `Consolidating Union`                                |
```
ggevay, line 111: We might want to leave off the `Consolidating` case.

doc/developer/design/20250226_postgres_style_explain.md
ggevay: I'm still not sure how to show the MFPs that are pushed inside join paths. The current MIR EXPLAIN doesn't show these, which sometimes creates confusion, while the current EXPLAIN PHYSICAL PLAN shows these in an extremely verbose way.

> Postgres always shows join order (implicitly, via AST structure

This would get quite verbose for Delta joins, because we'd have a big tree for each join path. In other words, the verboseness of our printing would be quadratic with the number of join inputs.

ggevay: I think we could just show joins similarly to how current MIR EXPLAIN shows them, and think about this later.

ggevay (approved):
Great, thanks!

doc/developer/design/20250226_postgres_style_explain.md
mgree: It's meant to indicate the join path, yeah... I might have messed it up manually coming up with this!

antiguru (approved):
(I don't have a strong opinion as I lack some background in the area, so take my words with a grain of salt.)

I think this proposal is fine, as in it moves us from one potentially confusing representation to something that users might be more familiar with. I like the way it maps LIR nodes to more practical terms (what does `Basic` even mean?), it should make it easier to follow for users who aren't familiar with the internal workings of Materialize.

The flipside is that I'm not sure by _how much_ we'll improve user experience, and for _how long_ the proposal will make sense in the future. While LIR is closest to what we're executing, it doesn't need to stay like this forever, or we might change LIR to something that is more amenable to dataflow rendering, i.e., different MirScalarExpr evaluation. Basing our explains off something that is even more tuned for a computer to be interpreted than a human might have the effect that explaining will need to recover information that is otherwise not available.

The other part I'm missing is whether a tree-based representation is the best we can do. We've had different approaches in the past, including the current tree-based one, and they all had trade-offs. A stack-machine-like syntax is much closer to what we're rendering (we're _not_  rendering trees), but confusing for people who assume the database follows some volcano-style execution pattern. The tree-based variant is in some way a mis-representation of what's happening, but caters to people who've used explain in the past.

What I mean to say is that this proposal seems like a step in one direction, and fine in the small, but I'm missing the bigger picture -- what _is_ the best way to represent what essentially boils down to a program?

ggevay:
Random thought about MFP fusion. Maybe we could print fused MFP as a separate operator, similarly to our current MIR EXPLAIN, but add a note that it's fused. For example, in the following example it's critically important that the MFP is fused into the FlatMap, because otherwise `data` would be copied for every attendee:
```
explain
SELECT
      (e.data -> 'fullDocument' -> '_id' ->> '$oid') AS id,
      jsonb_array_elements(e.data -> 'fullDocument' -> 'attendees') AS attendee
    FROM materialize.prod_sources.src_mongodb_events e;


Explained Query:
  Project (#2, #1) // { arity: 2 }
    Map ((((#0{data} -> "fullDocument") -> "_id") ->> "$oid")) // { arity: 3 }
      FlatMap jsonb_array_elements(((#0{data} -> "fullDocument") -> "attendees")) // { arity: 2 }
        Project (#1{data}) // { arity: 1 }
          ReadStorage materialize.prod_sources.src_mongodb_events // { arity: 2 }
```

```
explain physical plan for
SELECT
      (e.data -> 'fullDocument' -> '_id' ->> '$oid') AS id,
      jsonb_array_elements(e.data -> 'fullDocument' -> 'attendees') AS attendee
    FROM materialize.prod_sources.src_mongodb_events e;


Explained Query:
  FlatMap jsonb_array_elements(((#0 -> "fullDocument") -> "attendees"))
    mfp_after
      project=(#2, #1)
      map=((((#0 -> "fullDocument") -> "_id") ->> "$oid"))
    Get::Collection materialize.prod_sources.src_mongodb_events
      raw=true

Source materialize.prod_sources.src_mongodb_events
  project=(#1)
```

And it could be something like:
```
  Project (#2, #1) // (fused down into FlatMap)
    Map ((((#0{data} -> "fullDocument") -> "_id") ->> "$oid")) // (fused down into FlatMap)
      FlatMap jsonb_array_elements(((#0{data} -> "fullDocument") -> "attendees"))
        Project (#1{data}) // { arity: 1 }
          ReadStorage materialize.prod_sources.src_mongodb_events // { arity: 2 }
```

And maybe we could do the same also for MFPs that are pushed into joins: put the MFP on top of the join, but add a note that it's fused/pushed into the join.

mgree:
If the new explain (after https://github.com/MaterializeInc/materialize/pull/31878 lands) will be using LIR, I think we'll be presenting things closer to this way anyway! When I have a prototype for that new EXPLAIN, let's revisit this and make sure we're happy with what we're producing.

PR #3149, opened 2020-05-26
sql-parser: render CAST expressions using Postgres syntax

benesch:
The PostgreSQL-style casts

    'foo'::int4

are a lot less verbose than standard SQL casts:

    CAST('foo' AS int4)

When formatting SQL statements, use the PostgreSQL style. This will make
`SHOW CREATE VIEW` more readable, and allow a forthcoming PR (#3146) to
avoid introducing a new AST node for typed string literals.

---
This change is [Reviewable](https://reviewable.io/reviews/materializeinc/materialize/3149)

PR #31696, opened 2025-02-27
sql: don't re-write `CAST`s to 'list's with the `::` operator

ParkMyCar:
This PR changes our SQL rendering to use the standard SQL `CAST` expression when a cast involves the `list` type, instead of rewriting the expression to use the Postgres `::` operator.

```
-- For example consider the following expression:
CAST(<field> AS <type> list)[index]
-- Today get's re-written as:
(<field>)::<type> list[index]

-- these two expressions are not the same!
```

The rewrite to the Postgres operator was originally introduced in https://github.com/MaterializeInc/materialize/pull/3149

I went back and forth on a number of different solutions to fix this panic, and what's in this PR is sort of the least bad. My major gripe with this implementation is adding the `AstDataType` trait, it feels like it breaks the abstraction of `AstInfo::DataType`, but maybe this isn't too bad?

Other changes I tried:
1. Parsing `int list[]` as indexing into a list as opposed to the type `Array<List<int>>`. We already don't support arrays of lists, but this behavior felt like a very big change.
2. Supporting wrapping parens around the type, e.g. `(int list)`. Also not a bad solution but the parser impl wasn't too straight forward, and given the root cause was re-writing the `CAST` statements, this felt like a bit of a hack?
3. Wrapping the entire `::` expression in parens, e.g. `((<field>)::<type> list)[index]`. But this also changes the semantics a bit because now the expression provided to the subscript operator is a nested expression. Concretely going from text -> AST -> text -> AST failed to roundtrip AST.

### Motivation

A customer hit this panic in their environment.

### Checklist

### DIFF
--- a/src/sql-parser/src/ast/defs/expr.rs
+++ b/src/sql-parser/src/ast/defs/expr.rs
@@ -24,7 +24,7 @@ use mz_ore::soft_assert_eq_or_log;
 use mz_sql_lexer::keywords::*;
 
 use crate::ast::display::{self, AstDisplay, AstFormatter};
-use crate::ast::{AstInfo, Ident, OrderByExpr, Query, UnresolvedItemName, Value};
+use crate::ast::{AstDataType, AstInfo, Ident, OrderByExpr, Query, UnresolvedItemName, Value};
 
 /// An SQL expression of any type.
 ///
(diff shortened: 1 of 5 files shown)


src/sql-parser/src/ast/defs/expr.rs
```diff
@@ -321,36 +321,60 @@ impl<T: AstInfo> AstDisplay for Expr<T> {
...
+                // can cause the behavior to change.
+                //
+                // For example consider the following expression:
+                //      CAST(<field> AS <type> list)[index]
+                // would get re-written as:
+                //      (<field>)::<type> list[index]
+                //
+                // The original statement is indexing into the casted list, but
+                // the re-written statement is casting to an Array<List<type>>.
+                let can_rewrite_to_pg = !data_type.is_list();
```
ggevay, line 335: I think we unfortunately also need an `is_array`, because the following still triggers the problem:
```
CREATE MATERIALIZED VIEW json_mv2 AS (
    SELECT * FROM json_list WHERE random_id = CAST(payload->>'my_field' AS uuid[])[random_index]
);
```

ggevay: Btw. do we want to have the invariant that "if it parses then it roundtrips"? Because even if the cast is not to list or array, but to any other type, the statement still parses, but does not roundtrip. For example, if I write the following parser test fails the roundtrip:
```
parse-statement
SELECT CAST(my_json->>'field' AS uuid)[my_index] FROM fake_table;
----
```
This doesn't plan successfully of course, so I can't see a way to trigger an actual issue in the product with it, so whether this is a problem depends on whether we want to have the above invariant. Maybe at some point we'll want to have fuzzing for the parser, in which case the above invariant would be nice to have.

ggevay: Also, I'm not sure how easy it is to check for the array type. It's easy for `RawDataType`, because there is the `RawDataType::Array` enum variant, but might be harder for `ResolvedDataType`, because arrays are `ResolvedDataType::Named` at that point. We'd have to check the id inside the `Named`, but I don't see any conveniently available list of array type ids, or do hacky string matching on `full_name`.

ggevay: An idea for a completely different alternative approach: When printing `Expr::Subscript` in `AstDisplay for Expr`, check whether `expr`'s printing can end with a type name, and put parens around `expr` in that case. The only `Expr` variant whose printing ends with a type name is `Expr::Cast`, as far as I can see.

ggevay: For 3., the objection was

> But this also changes the semantics a bit because now the expression provided to the subscript operator is a nested expression. Concretely going from text -> AST -> text -> AST failed to roundtrip AST.

However, the `needs_wrap` thing already makes us have such a problem even on `main`:
```
parse-statement
SELECT CAST(a + b AS integer) FROM fake_table;
----
```
fails with
```
reparse comparison failed:
Select(SelectStatement { query: Query { ctes: Simple([]), body: Select(Select { distinct: None, projection: [Expr { expr: Cast { expr: Op { op: Op { namespace: None, op: "+" }, expr1: Identifier([Ident("a")]), expr2: Some(Identifier([Ident("b")])) }, data_type: Other { name: Name(UnresolvedItemName([Ident("int4")])), typ_mod: [] } }, alias: None }], from: [TableWithJoins { relation: Table { name: Name(UnresolvedItemName([Ident("fake_table")])), alias: None }, joins: [] }], selection: None, group_by: [], having: None, qualify: None, options: [] }), order_by: [], limit: None, offset: None }, as_of: None })
!=
Select(SelectStatement { query: Query { ctes: Simple([]), body: Select(Select { distinct: None, projection: [Expr { expr: Cast { expr: Nested(Op { op: Op { namespace: None, op: "+" }, expr1: Identifier([Ident("a")]), expr2: Some(Identifier([Ident("b")])) }), data_type: Other { name: Name(UnresolvedItemName([Ident("int4")])), typ_mod: [] } }, alias: None }], from: [TableWithJoins { relation: Table { name: Name(UnresolvedItemName([Ident("fake_table")])), alias: None }, joins: [] }], selection: None, group_by: [], having: None, qualify: None, options: [] }), order_by: [], limit: None, offset: None }, as_of: None })
SELECT (a + b)::int4 FROM fake_table
```
on main.

I'm not sure what to do.

ggevay: I have an idea to resolve the `reparse comparison failed` problem for both 3. and for the `needs_wrap` thing: Instead of tweaking the AST printing, we should tweak the parsing:
- For 3., tweak `parse_subscript`: if `expr` matches `Expr::Cast`, then wrap `expr` in `Expr::Nested`.
- For the `needs_wrap` thing, tweak `parse_cast_expr` and `parse_pg_cast` to do the same `needs_wrap` check, and add `Expr::Nested` when needed.

ggevay: Alternative PR: https://github.com/MaterializeInc/materialize/pull/31705

ggevay:
Closing this one, as we have merged an alternative PR: https://github.com/MaterializeInc/materialize/pull/31705

PR #31707, opened 2025-02-28
Optimize some local crates

antiguru:
This doesn't seem to increase build times much, but comes with a dramatic
speed up of sqllogictest:

main:
```
bin/sqllogictest --build-only
	3677.11s user 382.87s system 1402% cpu 4:49.48 total
bin/sqllogictest test/sqllogictest/introspection
	1088.87s user 97.06s system 421% cpu 4:41.33 total
```

this:
```
bin/sqllogictest --build-only
	3906.03s user 390.20s system 1444% cpu 4:57.37 total
bin/sqllogictest test/sqllogictest/introspection
	206.07s user 5.38s system 260% cpu 1:21.13 total
```

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -262,6 +262,10 @@ gimli = { opt-level = 3 }
 miniz_oxide = { opt-level = 3 }
 object = { opt-level = 3 }
 rustc-demangle = { opt-level = 3 }
+timely = { opt-level = 3 }
+differential-dataflow = { opt-level = 3 }
+mz-compute = { opt-level = 3 }
+mz-transform = { opt-level = 3 }
 
 [profile.release]
 # Compile time seems similar to "lto = false", runtime ~10% faster


ggevay:
Good idea! Doing this for the `timely` and `differential` crates sounds definitely good! For `compute` and `transform`, I'm less sure. In your test, build time increased from 3677.11s to 3906.03s, which is a low percentage increase. However, I guess this was a full build of all crates. I'd be curious to see how are the build times affected after a small code change just in `compute` or `transform` (or, say, `expr`). If the build time of a small code change would increase on the order of few a hundred seconds, then it might not be worth it, as the common scenario of running just one slt is on the order of low tens of seconds usually.

antiguru:
This was a full rebuild after a cargo clean. I had this enabled for a while and I didn't notice a difference, but I might be ignorant.

def- (approved):

def-:
There is also `bin/sqllogictest --optimized`, but now debug is pretty similar to it ;)
```
$ rm -rf target*
$ time bin/sqllogictest --build-only --optimized
bin/sqllogictest --optimized --build-only  4098.25s user 302.73s system 1226% cpu 5:58.86 total
$ time bin/sqllogictest --optimized test/sqllogictest/introspection
bin/sqllogictest --optimized --build-only  499.01s user 42.95s system 670% cpu 1:20.86 total
```

ggevay:
I've tried now what happens when modifying code in the transform crate, and it's working really well! Compilation times are affected by just a few seconds, but slts became seriously faster!

PR #29139, opened 2024-08-21
cluster: use async channel to return activators

teskje:
Previously, the channel through which cluster workers would return dataflow activators to the tokio tasks managing client connections was a sync channel. When the connection tasks called `recv` on them, that call would synchronously block, stalling the current thread and even the entire tokio runtime if it was single-threaded.

The fix is to use an async channel instead.

### Checklist

- [x] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [x] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [x] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [x] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [x] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #31629, opened 2025-02-26
k8s-orchestrator: create a StatefulSet per process

teskje:
This PR changes the k8s orchestrator to create a separate statefulset for each process requested, instead of using a single statefulset with multiple replicas.

This is to allow us to supply different configurations to different cluster processes. In particular, we want to be able to provide each process with its index through a command line argument.

### Motivation

  * This PR refactors existing code.

Having the process ID as a command line argument to clusterd processes lets us use it for the Timely cluster initialization. See [thread](https://materializeinc.slack.com/archives/C08AENAP691/p1740413822244849) and [thread](https://materializeinc.slack.com/archives/CL68GT3AT/p1740572254878829).

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #31710, opened 2025-02-28
cluster: initialize Timely runtimes at cluster startup

teskje:
This PR moves the initialization of the Timely cluster from controller connection time to replica process startup time. When a replica process starts up, it now first initializes its Timely cluster and connects to its peer processes before starting to accept connections from the controller. This simplifies controller-replica communication in two ways:

* On controller (re)connects, we don't have to run the (potentially blocking) Timely cluster setup. This simplifies the replica-side connection code, hopefully reducing the risk of deadlock bugs we had there in the past ([1](https://github.com/MaterializeInc/database-issues/issues/8467), [2](https://github.com/MaterializeInc/materialize/pull/29139), [3](https://github.com/MaterializeInc/database-issues/issues/8957), [4](https://github.com/MaterializeInc/database-issues/issues/8964)).
* The `CreateTimely` command is removed, simplifying the cluster protocol and every client interacting with it.

The necessary Timely cluster config is now passed through the orchestrator as a command line argument to new clusterd processes.

### Motivation

### Tips for reviewer

I recommend reviewing the commits separately.

The diff is unfortunately quite large. I've tried splitting off as much as possible into preparatory PRs, but a large bulk of the diff comes from adjusting the tests, which can't be pulled out. If people prefer, we can pull out the first three commits that implement the orchestrator changes into a separate PR.

### Checklist

### DIFF
--- a/ci/mkpipeline.py
+++ b/ci/mkpipeline.py
@@ -28,6 +28,7 @@ import sys
 import traceback
 from collections import OrderedDict
 from collections.abc import Iterable, Iterator
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import Any
 
@@ -375,19 +376,27 @@ def prioritize_pipeline(pipeline: Any, priority: int) -> None:
 def switch_jobs_to_aws(pipeline: Any, priority: int) -> None:
     """Switch jobs to AWS if Hetzner is currently overloaded"""
 
-    branch = os.getenv("BUILDKITE_BRANCH")
+    # branch = os.getenv("BUILDKITE_BRANCH")
 
     # If Hetzner is entirely broken, you have to take these actions to switch everything back to AWS:
     # - CI_FORCE_SWITCH_TO_AWS env variable to 1
     # - Reconfigure the agent from hetzner-aarch64-4cpu-8gb to linux-aarch64-small in https://buildkite.com/materialize/test/settings/steps and other pipelines
     # - Reconfigure the agent from hetzner-aarch64-4cpu-8gb to linux-aarch64-small in ci/mkpipeline.sh
-    if not ui.env_is_truthy("CI_FORCE_SWITCH_TO_AWS", "0"):
+
+    stuck: set[str] = set()
+    # TODO(def-): Remove me when Hetzner fixes its aarch64 availability
+    stuck.add("aarch64")
+
+    if ui.env_is_truthy("CI_FORCE_SWITCH_TO_AWS", "0"):
+        stuck = set(["x86-64", "x86-64-dedi", "aarch64"])
+    else:
+        # TODO(def-): Reenable me when Hetzner fixes its aarch64 availability
         # If priority has manually been set to be low, or on main branch, we can
         # wait for agents to become available
-        if branch == "main" or priority < 0:
-            return
+        # if branch == "main" or priority < 0:
+        #     return
 
-        # Consider Hetzner to be overloaded when at least 600 jobs exist with priority >= 0
+        # Consider Hetzner to be overloaded/broken when an important job is stuck waiting for an agent for > 60 minutes, per arch (x86-64/x86-64-dedi/aarch64)
         try:
             builds = generic_api.get_multiple(
                 "builds",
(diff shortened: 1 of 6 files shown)


test/cluster/mzcompose.py
```diff
@@ -113,120 +113,86 @@ def workflow_test_smoke(c: Composition, parser: WorkflowArgumentParser) -> None:
...
-                WORKERS 2
-            )
-        );
 
-        GRANT ALL ON CLUSTER cluster1 TO materialize;
-        """,
-        port=6877,
-        user="mz_system",
-    )
+    with c.override(
```
teskje, line 117: Most of the diffs in this file come from adding overrides, which force indentation changes. I recommend reviewing with whitespace changes disabled!

src/orchestrator-process/src/lib.rs
```diff
@@ -628,13 +613,47 @@ impl OrchestratorWorker {
             None
         };
 
+        // The service might already exist. If it has the same config as requested (currently we
+        // can check only the scale), we have nothing to do. Otherwise we need to drop and recreate
+        // it.
```
teskje, line 618: This is a functional change: Before we would simply leave existing processes unchanged and just add/drop ones according to the requested `scale`. I'm not sure if that approach made sense before, but it definitely doesn't make sense in a world where each process needs to start up with a compatible Timely config.

src/controller/src/clusters.rs
```diff
@@ -625,12 +628,39 @@ where
...
         let persist_pubsub_url = self.persist_pubsub_url.clone();
         let secrets_args = self.secrets_args.to_flags();
+
+        // TODO(teskje): use the same values as for compute?
+        let storage_proto_timely_config = TimelyConfig {
+            // This value is not currently used by storage, so we just choose
+            // some identifiable value.
+            arrangement_exert_proportionality: 1337,
+            ..Default::default()
+        };
```
teskje, line 638: It would be really easy to use the same config as compute here. Still deferring to a follow-up since it's a behavior change.

src/controller/src/lib.rs
```diff
@@ -200,11 +201,25 @@ pub struct Controller<T: ComputeControllerTimestamp = mz_repr::Timestamp> {
     ///
     /// See [`self.install_watch_set`] for a description of watch sets.
     immediate_watch_sets: Vec<WatchSetId>,
+
+    /// Dynamic system configuration.
+    dyncfg: ConfigSet,
+    /// `arrangement_exert_proportionality` value passed to new replicas.
+    arrangement_exert_proportionality: u32,
```
teskje, line 208: Possible follow-up: Make `arrangement_exert_proportionality` a dyncfg.

src/clusterd/ci/entrypoint.sh
```diff
@@ -21,11 +21,15 @@ export CLUSTERD_INTERNAL_HTTP_LISTEN_ADDR=${CLUSTERD_INTERNAL_HTTP_LISTEN_ADDR:-
...
-# in Docker), this is likely not desirable.
 if [[ "${KUBERNETES_SERVICE_HOST:-}" ]]; then
+    # Pass the host's FQDN as the host to be used for GRPC request validation
+    # only when running in Kubernetes. In other contexts (like when running
+    # locally, or in Docker), this is likely not desirable.
     export CLUSTERD_GRPC_HOST=${CLUSTERD_GRPC_HOST:-$(hostname --fqdn)}
+
+    # When running in Kubernetes, pass the StatefulSet replica's ordinal index
+    # as the process index.
+    export CLUSTERD_PROCESS=${CLUSTERD_PROCESS:-HOSTNAME##*-}
```
teskje, line 32: The plan is to have a StatefulSet per process and supply the process index in the `TimelyConfig`s passed to each. Changing to multiple StatefulSets is a bigger change though ([draft](https://github.com/MaterializeInc/materialize/pull/31629)), so deferring it to a follow-up and doing the simple thing here.

test/limits/mzcompose.py
```diff
@@ -1946,7 +1973,7 @@ def workflow_main(c: Composition, parser: WorkflowArgumentParser) -> None:
                             password=app_password(ADMIN_USER),
                         ) as cur:
                             start_time = time.time()
-                            cur.execute(scenario.EXPLAIN)
+                            cur.execute(scenario.EXPLAIN) # type: ignore
```
teskje, line 1976: Type checking complains here because `Cursor.execute` only takes `LiteralString`, or some psycopg type that prevents SQL injection. But we do dynamically build the value of `EXPLAIN`, so I'm not sure there is anything better to do here than ignore the type check error.

src/clusterd/ci/entrypoint.sh
```diff
@@ -21,11 +21,15 @@ export CLUSTERD_INTERNAL_HTTP_LISTEN_ADDR=${CLUSTERD_INTERNAL_HTTP_LISTEN_ADDR:-
...
-# in Docker), this is likely not desirable.
 if [[ "${KUBERNETES_SERVICE_HOST:-}" ]]; then
+    # Pass the host's FQDN as the host to be used for GRPC request validation
+    # only when running in Kubernetes. In other contexts (like when running
+    # locally, or in Docker), this is likely not desirable.
     export CLUSTERD_GRPC_HOST=${CLUSTERD_GRPC_HOST:-$(hostname --fqdn)}
+
+    # When running in Kubernetes, pass the StatefulSet replica's ordinal index
+    # as the process index.
+    export CLUSTERD_PROCESS=${CLUSTERD_PROCESS:-${HOSTNAME##*-}}
```
teskje, line 32: The plan is to have a StatefulSet per process and supply the process index in the `TimelyConfig`s passed to each. Changing to multiple StatefulSets is a bigger change though ([draft](https://github.com/MaterializeInc/materialize/pull/31629)), so deferring it to a follow-up and doing the simple thing here.

test/limits/mzcompose.py
```diff
@@ -1946,7 +1973,7 @@ def workflow_main(c: Composition, parser: WorkflowArgumentParser) -> None:
                             password=app_password(ADMIN_USER),
                         ) as cur:
                             start_time = time.time()
-                            cur.execute(scenario.EXPLAIN)
+                            cur.execute(scenario.EXPLAIN)  # type: ignore
```
teskje, line 1976: Type checking complains here because `Cursor.execute` only takes `LiteralString`, or some psycopg type that prevents SQL injection. But we do dynamically build the value of `EXPLAIN`, so I'm not sure there is anything better to do here than ignore the type check error.

src/cluster/src/communication.rs
```diff
@@ -139,237 +182,305 @@ where
...
+    my_index: usize,
+    addresses: &[String],
+) -> Result<Vec<Option<Stream>>, CreateSocketsError> {
+    let my_address = &addresses[my_index];
+
+    // Binding to a TCP address of the form `hostname:port` unnecessarily involves a DNS query. We
+    // should get the port from here, but otherwise just bind to `0.0.0.0`.
+    let port_re = Regex::new(r":(\d{1,5})$").unwrap();
+    let listen_address = match port_re.captures(my_address) {
+        Some(cap) => format!("0.0.0.0:{}", &cap[1]),
```
antiguru, line 288: Just pointing out that this may hardcode IPv4.

teskje: That's true, but not sure if it's a problem. Also it's not new in this PR, so I'm not planning to change it here. Can do as a follow up, if we think it's necessary/worthwhile.

def-:
If you want to run tests/nightly I'd suggest rebasing since Hetzner is still out of aarch64 machines.

teskje:
Ah, I'm actually using this as a repro for https://github.com/MaterializeInc/database-issues/issues/9415 since for some reason it occurs very reliably with this branch. I'm afraid rebasing will break the repro 😅

def-:
Switching to x86-64 or AWS could also make it harder to reproduce? I have picked the relevant commits for that into https://github.com/MaterializeInc/materialize/pull/32854 and triggered new runs:
- https://buildkite.com/materialize/test/builds/105320
- https://buildkite.com/materialize/nightly/builds/12461

teskje:
Looks like it still reproduces with the CI changes! I've picked them into this branch.

teskje:
Closing in favor of https://github.com/MaterializeInc/materialize/pull/32869.

PR #31705, opened 2025-02-28
sql: Wrap subscripted expression in parens when it ends with a type name

ggevay:
This is an alternative approach for solving the problem mentioned in https://github.com/MaterializeInc/materialize/pull/31696. The advantages of this approach are:
- Roundtrips through the parser even if the cast is not to a list or array type. [Details.](https://github.com/MaterializeInc/materialize/pull/31696/files#r1975213111)
- Does not break the abstraction of `AstInfo::DataType`. ([Breaking into `AstInfo::DataType` might be non-trivial for arrays.](https://github.com/MaterializeInc/materialize/pull/31696/files#r1975238697))
- Future-proof against the same problem occurring again if we add subscripting for more types at some point.
- ([The same approach can be used to solve the `needs_wrap` roundtripping problem](https://github.com/MaterializeInc/materialize/pull/31696/files#r1975321220) in a follow-up PR.)

Nightly: https://buildkite.com/materialize/nightly/builds/11348

### Motivation

[A customer encountered this.](https://materializeinc.slack.com/archives/C08ACQNGSQK/p1740675512403819?thread_ts=1740674779.093879&cid=C08ACQNGSQK)

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #31717, opened 2025-03-01
[parser] Refactor the handling of complex expressions in casts

ggevay:
This is a follow-up to https://github.com/MaterializeInc/materialize/pull/31705. When working on that PR, we noticed that if we fix precedence issues by adding parenthesis during the _printing_ of the AST, then the problem is that the AST won't roundtrip in parser tests. Therefore that PR fixed a precedence issue during AST parsing instead of printing. However, even before that PR, the AST printing already had an earlier precedence fix during printing. This was not causing a problem so far only because we hadn't had a parser test for it. So this PR moves that precedence fix also to the parsing, and adds a parser test for it. This is in the second commit.

The first commit is a related refactoring in the parser tests: When there was a parser roundtrip problem in tests, it would often manifest as an error msg complaining about the redacted query, even if the non-redacted query would also have the same roundtrip problem, which was sometimes misleading me into thinking that there is some issue specifically with the redaction. So, this refactoring moves the redacted reparse check to happen after the normal reparse checks. (See also commit msg.)

Edit: Note that the roundtrip problem that this PR is solving is only a problem in the sense that it's not an identical AST that we get back after reparsing, but the meaning of the AST is still the same. So, this is a test-only issue, as far as I can see.

Edit: I added a 3rd and 4th commit, each of which adds a case where the extra wrapping parens are not needed. If we have the extra wrapping, that can cause roundtrip failures, because the parser removes superfluous parens in some cases.

(Edit: In the long run, the cleanest solution would be to have a representation for both casting styles, and then both would trivially roundtrip. Then, if we still like the Postgres-style better, then an AST-transformation (like the ones in `transform_ast.rs`) could change from the standard to the Postgres style, and add Expr::Nested when needed.)

Nightly: https://buildkite.com/materialize/nightly/builds/11365

### Motivation

### Checklist

### DIFF
--- a/src/environmentd/tests/testdata/http/post
+++ b/src/environmentd/tests/testdata/http/post
@@ -324,7 +324,7 @@ http
 {"queries":[{"query":"select $1+$2::int as col","params":["1"]}]}
 ----
 200 OK
-{"results":[{"error":{"message":"request supplied 1 parameters, but SELECT $1 + ($2)::int4 AS col requires 2","code":"XX000"},"notices":[]}]}
+{"results":[{"error":{"message":"request supplied 1 parameters, but SELECT $1 + $2::int4 AS col requires 2","code":"XX000"},"notices":[]}]}
 
 # NaN
 http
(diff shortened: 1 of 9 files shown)


ParkMyCar (approved):
Sorry it took me so long to review this one Gabor! Thanks for making this change!

ggevay:
No problem! Thanks for reviewing it!

PR #31763, opened 2025-03-04
upsert: inititalize rocksdb in core loop thread

martykulma:
After looking into the logs and talking over with Ben, it looks like we can make 2 changes to improve the behavior. The goal is to fall back to suspend-and-restart where possible and not wait around for threads.

1. Don't try to clean up rocks DB as part of instantiation.  The existing approach blocks the creation of the operator, and results in a panic if there was an issue cleaning up or starting up RocksDB.
Initialization is moved into the dedicated thread that services requests for RocksDB.  If the initialization fails, the thread exits, closing the command channel. The next command to RocksDB will result in a suspend-and-restart.

2. If the main loop encounters a fatal error during write, return the error via normal channel, and then exit the loop to begin cleanup.  If RocksDB is in an error state, there's no benefit in waiting for something to terminate it.

### Motivation

fixes https://github.com/MaterializeInc/database-issues/issues/8981

### Checklist

### DIFF
--- a/src/rocksdb/src/lib.rs
+++ b/src/rocksdb/src/lib.rs
@@ -881,7 +881,7 @@ fn rocksdb_core_loop<K, V, M, O, IM, F>(
                         let _ = response_sender.send(Err(e));
                         if let Some(db_err) = db_err {
                             if !matches!(db_err.kind(), ErrorKind::TryAgain) {
-                                tracing::error!(
+                                tracing::warn!(
                                     "exiting on fatal rocksdb error at {}: {}",
                                     instance_path.display(),
                                     db_err.display_with_causes(),


def- (approved):

src/rocksdb/src/lib.rs
```diff
@@ -655,13 +641,13 @@ fn rocksdb_core_loop<K, V, M, O, IM, F>(
...
-        Ok(db) => {
-            drop(creation_error_tx);
-            db
-        }
+        Ok(db) => db,
         Err(e) => {
-            // Communicate the error back to `new`.
-            let _ = creation_error_tx.send(e);
+            tracing::error!(
+                "failed to create rocksdb at {}: {}",
```
petrosagg, line 647: This will make it to Sentry. Is this a rare/interesting enough error that we should be notified if it happens?

src/rocksdb/src/lib.rs
```diff
@@ -890,12 +879,23 @@ fn rocksdb_core_loop<K, V, M, O, IM, F>(
                     }
                     Err(e) => response_sender.send(Err(e)),
                 };
+
+                // TODO(maz): should we also terminate when retries are exhausted?
+                if let Some(fatal_error) = fatal_error {
+                    tracing::error!(
+                        "existing on fatal rocksdb error at {}: {}",
```
petrosagg, line 886: Did you mean "exiting" here?

src/rocksdb/src/lib.rs
```diff
@@ -869,7 +855,10 @@ fn rocksdb_core_loop<K, V, M, O, IM, F>(
                             }
                             Err(e) => match e.kind() {
                                 ErrorKind::TryAgain => RetryResult::RetryableErr(Error::RocksDB(e)),
-                                _ => RetryResult::FatalErr(Error::RocksDB(e)),
+                                _ => {
+                                    fatal_error.replace(Error::RocksDB(e.clone()));
```
petrosagg, line 859: nit: `Option::replace` is mostly useful when you want to do something with the old value, otherwise this code is the same as the normal assignment `fatal_error = Some(..)`.

src/rocksdb/src/lib.rs
```diff
@@ -890,12 +879,23 @@ fn rocksdb_core_loop<K, V, M, O, IM, F>(
                     }
                     Err(e) => response_sender.send(Err(e)),
                 };
+
+                // TODO(maz): should we also terminate when retries are exhausted?
+                if let Some(fatal_error) = fatal_error {
```
petrosagg, line 884: I couldn't figure out why we track this error separately instead of inspecting the `retry_result` above. Do we lose some information through that variable?

src/rocksdb/src/lib.rs
def-: I think so, yes. We have a bunch of errors in Senry which are mostly user errors.

petrosagg (approved):
LGTM!

PR #31766, opened 2025-03-04
[wip] license key implementation

doy-materialize:
### Motivation

https://github.com/MaterializeInc/cloud/issues/10759

this is the first draft of an implementation of license keys (more details can be found at https://www.notion.so/materialize/License-keys-1a713f48d37b80298db2d241b90fcfd5). the basic outline is that environmentd will read the max credit consumption rate from the signed license key as well as a flag to determine whether both the credit consumption rate limit and the credits each cluster replica uses can be overridden by the user. if no license key is provided, the user will be limited to 24G of clusters. in cloud, we'll be able to provide a special license key which disables these checks.

### Tips for reviewer

opening this as a draft to see how it works in ci and if there is anything that needs to be fixed in the test suite. also, the public key used for license validation is currently coming from a random kms key in the scratch account, and i'll need to do the real implementation of license key issuing on the cloud side first before we can merge this, in order to get the real pubkey to use.

### Checklist


src/sql/src/session/vars.rs
```diff
@@ -1361,6 +1385,9 @@ impl SystemVars {
 
     /// Returns whether or not this parameter can be modified by a superuser.
     pub fn user_modifiable(&self, name: &str) -> bool {
+        if self.check_read_only(name).is_err() {
+            return false;
+        }
```
jubrad, line 1390: MAX_CREDIT_CONSUMPTION_RATE shouldn't be a session_var so it shouldn't need a readonly check here, although it's not hurting anything to be here.

src/adapter/src/config.rs
```diff
@@ -43,6 +43,8 @@ pub struct SystemParameterSyncConfig {
     /// to use when populating the [SynchronizedParameters]
     /// instance in [SystemParameterFrontend::pull].
     ld_key_map: BTreeMap<String, String>,
+    /// max_credit_consumption_rate override from the license key.
+    pub max_credit_consumption_rate: Option<f64>,
```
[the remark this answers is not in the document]
doy-materialize: we probably will need to do this eventually, but that would be a more complicated change that i'd like to leave for the future, to limit the scope here (there's already a lot going on in this pr)

src/environmentd/src/environmentd/main.rs
```diff
@@ -1026,8 +1048,19 @@ fn run(mut args: Args) -> Result<(), anyhow::Error> {
         },
     };
 
-    let cluster_replica_sizes: ClusterReplicaSizeMap =
+    let mut cluster_replica_sizes: ClusterReplicaSizeMap =
         serde_json::from_str(&args.cluster_replica_sizes).context("parsing replica size map")?;
+    if !license_key_config.allow_credit_consumption_override {
+        for (name, replica) in cluster_replica_sizes.0.iter_mut() {
+            let memory_limit = replica.memory_limit.get_or_insert_with(|| {
+                warn!("No memory limit found in cluster definition for {name}, defaulting to 4GiB");
```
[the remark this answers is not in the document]
doy-materialize: yeah, i wasn't quite sure what to do here - if a user is using a license key with limited credits/memory usage, we can't allow using clusters that don't have memory limits set, but a lot of our tests currently want to do this. it might be better to just fail here instead of setting a default, and update the tests instead, i'll look into it

src/environmentd/src/license_keys.rs
```diff
@@ -0,0 +1,54 @@
...
+        }
+    }
+
+    pub fn max_credit_consumption_rate(&self) -> Option<f64> {
+        if self.allow_credit_consumption_override {
+            None
+        } else {
+            Some(self.max_credit_consumption_rate)
+        }
+    }
```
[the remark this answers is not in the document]
doy-materialize: the first reason is to use in cloud - if we want to use the same binaries in cloud (which i think we do), we'll need to be able to provide a license key which allows unrestricted access so that we can define our own credit consumption for cluster sizes and allow the limit to be changed by launchdarkly. additionally, nate mentioned we'll need the ability to issue license keys with unlimited resource capacity (presumably for larger customers)

src/license-keys/src/signing.rs
```diff
@@ -0,0 +1,109 @@
...
+        .key_id(key_id)
+        .send()
+        .await?
+        .public_key
+    {
+        Ok(pubkey.into_inner())
+    } else {
+        Err(anyhow!("failed to get pubkey"))
+    }
+}
```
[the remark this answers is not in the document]
doy-materialize: this get_pubkey function here was just for testing - the actual implementation will have pulumi create the kms key and emit the pubkey as an output, which we can then hardcode into the binary here (in `const PUBLIC_KEYS` in lib.rs)

src/sql/src/session/vars.rs
doy-materialize: ah, fair enough, but yeah, i think it still makes sense to do this to make it easier to understand in the future

src/sql/src/session/vars.rs
```diff
@@ -1051,6 +1051,8 @@ pub enum NetworkPolicyError {
 pub struct SystemVars {
     /// Allows "unsafe" parameters to be set.
     allow_unsafe: bool,
+    /// max_credit_consumption_rate set by the license key.
+    max_credit_consumption_rate: Option<f64>,
```
ParkMyCar, line 1055: Instead of adding the license key's version of `max_credit_consumption_rate` to `SystemVars`, can we instead check the value from the `CatalogConfig` in the existing [`validate_resource_limit_numeric(...)`](https://github.com/ParkMyCar/materialize/blob/main/src/adapter/src/coord/ddl.rs#L1879-L1913) function?

I think this should prevent the need for drilling the license key value down so far and prevent the need for introducing a read_only concept of a system variable. wdyt?

src/catalog/src/config.rs
```diff
@@ -92,6 +92,8 @@ pub struct StateConfig {
     pub enable_0dt_deployment: bool,
     /// Helm chart version
     pub helm_chart_version: Option<String>,
+    /// max_credit_consumption_rate override from the license key
+    pub max_credit_consumption_rate: Option<f64>,
```
ParkMyCar, line 96: Instead of just storing the `max_credit_consumption_rate` here, it seems like it might be nice to store an entire `ValidatedLicenseKey`? Up to you!

src/sql/src/session/vars/definitions.rs
```diff
@@ -493,7 +493,7 @@ pub static MAX_REPLICAS_PER_CLUSTER: VarDefinition = VarDefinition::new(
 
 pub static MAX_CREDIT_CONSUMPTION_RATE: VarDefinition = VarDefinition::new_lazy(
     "max_credit_consumption_rate",
-    lazy_value!(Numeric; || 1024.into()),
+    lazy_value!(Numeric; || 24.into()),
```
ParkMyCar, line 496: I'm a little concerned this will break existing environments. Looking at LaunchDarkly everything seems to be configured correctly, but this does open the risk for if LaunchDarkly fails to sync we'll now "fail closed" and prevent users from creating new objects, whereas before we'd "fail open"

src/sql/src/session/vars.rs
doy-materialize: ah, yeah, this is much simpler. thanks!

src/catalog/src/config.rs
doy-materialize: yeah, may as well

src/sql/src/session/vars/definitions.rs
doy-materialize: yeah, this probably doesn't actually need to change since we check the license key first for the value anyway

src/environmentd/src/environmentd/main.rs
```diff
@@ -681,6 +686,28 @@ fn run(mut args: Args) -> Result<(), anyhow::Error> {
...
+            match license_key.expiration_behavior {
+                ExpirationBehavior::Warn | ExpirationBehavior::DisableClusterCreation => {
+                    warn!("{message}");
+                }
+                ExpirationBehavior::Disable => panic!("{message}"),
+            }
+        }
+        license_key
+    } else if matches!(args.orchestrator, OrchestratorKind::Kubernetes) {
+        panic!("--license-key is required when running in Kubernetes");
```
ParkMyCar, line 706: Any chance we could further match here on the cloud provider to skip this check entirely for the Cloud offering? (Note: I'm not sure what cloud provider self hosted folks get reported as)

Also any reason for a panic instead of returning an error?

src/environmentd/src/environmentd/main.rs
```diff
@@ -1026,8 +1053,18 @@ fn run(mut args: Args) -> Result<(), anyhow::Error> {
         },
     };
 
-    let cluster_replica_sizes: ClusterReplicaSizeMap =
+    let mut cluster_replica_sizes: ClusterReplicaSizeMap =
         serde_json::from_str(&args.cluster_replica_sizes).context("parsing replica size map")?;
+    if !license_key.allow_credit_consumption_override {
+        for (name, replica) in cluster_replica_sizes.0.iter_mut() {
+            let Some(memory_limit) = replica.memory_limit else {
+                panic!("No memory limit found in cluster definition for {name}");
```
ParkMyCar, line 1061: Maybe return an error here instead of panicking?

src/environmentd/src/environmentd/main.rs
```diff
@@ -1026,8 +1053,18 @@ fn run(mut args: Args) -> Result<(), anyhow::Error> {
...
+            };
+            replica.credits_per_hour = Numeric::from(
+                (memory_limit.0 * replica.scale * u64::try_from(replica.workers).unwrap()).0,
+            ) / Numeric::from(1024 * 1024 * 1024);
```
ParkMyCar, line 1065: Mind documenting the use of `1024 * 1024 * 1024`? It seems to assume that memory_limit values are in GiB? Also FWIW we generally try to use the `checked_mul(...)` or `saturating_mul(...)` methods instead of just `*`

ParkMyCar (approved):
Overall looks good to me! Thanks for backing out the `SystemVar` changes!

src/environmentd/src/environmentd/main.rs
doy-materialize: i don't think we want to be able to bypass needing a license key like that? my understanding was we wanted all community edition usage to be restricted. self-hosted users will also be using the `aws` cloud provider (the same thing our cloud product uses) if they are running in aws

doy-materialize:
fixed up the panics and hardcoded constants

misc/python/materialize/cloudtest/k8s/environmentd.py
```diff
@@ -265,6 +293,9 @@ def args(self) -> list[str]:
                 "--announce-egress-ip=88.77.66.55",
             ]
 
+        if self._meets_minimum_version("0.137.0-dev"):
```
def-, line 296: 0.140.0-dev and you should rebase

doy-materialize:
okay, this appears to be working now - i still need to do some testing on the cloud side to make sure that things don't break when a release with these changes goes out, but it should be ready for one final review

doy-materialize:
fairly sure the remaining sqlsmith errors are https://github.com/MaterializeInc/materialize/pull/32081, so i'm going to ignore those for now

misc/python/materialize/mzcompose/services/materialized.py
```diff
@@ -289,6 +290,16 @@ def __init__(
             config["platform"] = platform
 
         volumes = []
+
+        if image_version is None or image_version >= "v0.140.0-dev":
```
def-, line 294: Does string comparison work for a Version object?

alex-hunt-materialize: Probably not correctly.

doy-materialize: it does https://python-semver.readthedocs.io/en/latest/usage/compare-versions.html

PR #31774, opened 2025-03-05
adapter: add some debug logging for requests to get their latency

ParkMyCar:
Trying to debug some high latencies with our auth provider Frontegg. This PR just adds a bit of debug logging (off by default) to report the latencies of requests with the "frontegg-trace-id" header.

### Motivation

Debug auth latencies

### Checklist

### DIFF
--- a/src/frontegg-auth/src/client/tokens.rs
+++ b/src/frontegg-auth/src/client/tokens.rs
@@ -7,6 +7,8 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::time::Instant;
+
 use mz_ore::instrument;
 use mz_ore::metrics::MetricsFutureExt;
 use uuid::Uuid;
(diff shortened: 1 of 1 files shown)


src/frontegg-auth/src/client/tokens.rs
```diff
@@ -29,14 +30,18 @@ impl Client {
...
             .json(&request)
             .send()
-            .wall_time()
-            .observe(histogram)
             .await?;
+        let duration = start.elapsed();
+
+        // Authentication is on the blocking path for connection startup so we
+        // want to make sure it stays fast.
+        histogram.observe(duration.as_secs_f64());
```
SangJunBak, line 44: Ik the previous implementation returns a `WallTimeFuture` which I'm assuming that once awaited, will pass the update to a channel and propagate to Prometheus. But I noticed histogram.observe just records it into its object. How do we know for sure that the duration is still being observed in prometheus?

ParkMyCar: The impl of `Future` for `WallTimeFuture` does essentially the same thing I'm manually doing here. On [first poll record some starting instant](https://github.com/MaterializeInc/materialize/blob/main/src/ore/src/metrics.rs#L607-L609), then once the future finishes (returns `Poll::Ready`) it [calculates the duration since we started](https://github.com/MaterializeInc/materialize/blob/main/src/ore/src/metrics.rs#L615). It's a bit tricky, but [calling `.observe(...)`](https://github.com/MaterializeInc/materialize/blob/main/src/ore/src/metrics.rs#L554) just sets the histogram we want to [record into when the future completes](https://github.com/MaterializeInc/materialize/blob/main/src/ore/src/metrics.rs#L623).

The code now is doing the same exact thing that it used to, but now we have the duration so we can pass it in a `tracing::debug!` log

SangJunBak: Ah that makes sense! I should've looked at the implementation of `poll` more closely 😅 I was just confused that one action was async and the other wasn't, but yeah although it doesn't seem like the exact code path, both seem to be doing the same thing

SangJunBak (approved):

PR #31776, opened 2025-03-06
docs: self-managed aws upgrade

kay-kim:


doc/user/content/installation/install-on-aws/upgrade.md
```diff
@@ -0,0 +1,87 @@
...
+1. Open a Terminal window.
+
+1. Configure AWS CLI with your AWS credentials. For details, see the [AWS
+   documentation](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html).
+
+1. Go to the Terraform directory for your Materialize deployment. For example,
+   if you deployed from the `examples/simple` directory:
+
+   ```bash
+   cd terraform-aws-materialize/examples/simple
```
kay-kim, line 62: If they checked out a particular tag, do we want them to checkout a more recent tag compatible with their current version?

PR #31791, opened 2025-03-07
Run `ProjectionPushdown` after the last `RelationCSE`

ggevay:
_**Edit: I've added the `lts-backport-v25.1` label, but if this turns out to be too annoying to cherry-pick, then it's ok to skip this, as it's just a performance improvement.**_

This adds a call to `ProjectionPushdown` after the last call to `RelationCSE`. As explained in a code comment:
```
// `RelationCSE` can create new points of interest for `ProjectionPushdown`: If an MFP
// is cut in half by `RelationCSE`, then we'd like to push projections behind the new
// Get as much as possible. This is because a fork in the plan involves copying the
// data. (But we need `ProjectionPushdown` to skip joins, because it can't deal with
// filled in JoinImplementations.)
```

Slack discussion here:
https://materializeinc.slack.com/archives/C07SUTL7JMQ/p1741358397379089?thread_ts=1741341927.561309&cid=C07SUTL7JMQ

It's guarded by a feature flag as a precaution, but I've enabled it by default, as it's hard to imagine this hurting plans.

The first commit is the actual functionality, while the second commit just adds the feature flag.

### Motivation

* This PR increases performance.

### Checklist

### DIFF
--- a/src/repr/src/optimize.rs
+++ b/src/repr/src/optimize.rs
@@ -122,6 +122,8 @@ optimizer_feature_flags!({
     enable_join_prioritize_arranged: bool,
     // Extract expressions in MFP::extract_common, making them available to sources.
     extract_common_mfp_expressions: bool,
+    // See the feature flag of the same name.
+    enable_projection_pushdown_after_relation_cse: bool,
 });
 
 /// A trait used to implement layered config construction.
--- a/src/sql-lexer/src/keywords.txt
+++ b/src/sql-lexer/src/keywords.txt
@@ -29,6 +29,7 @@ Add
 Added
 Address
 Addresses
+After
 Aggregate
 Aggregation
 Aligned
@@ -112,6 +113,7 @@ Createnetworkpolicy
 Createrole
 Creation
 Cross
+Cse
 Csv
 Current
 Cursor
@@ -342,6 +344,7 @@ Prioritize
 Privatelink
 Privileges
 Progress
+Projection
 Protobuf
 Protocol
 Public
@@ -368,6 +371,7 @@ Refresh
 Regex
 Region
 Registry
+Relation
 Rename
 Reoptimize
 Repeatable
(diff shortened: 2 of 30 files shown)


test/sqllogictest/introspection/attribution_sources.slt
```diff
@@ -32,19 +32,157 @@ SELECT mz_unsafe.mz_sleep(8)
...
+40  u55
+41  u56
+46  t11884
+47  t11887
+48  t11893
+49  u66
+50  u68
+61  t12091
+94  u116
+94  u117
```
ggevay, line 61: (Ok, this slt file totally needs a `reset-server`, independently from this PR.)

mgree (approved):
Looks good to me: code is simple enough and SLTs look good.

The join skipping is an unwelcome bit of complexity (it's _almost_ the same logic), but here we are. Perhaps an issue for another time, but how hard would it be to actually look at `JoinImplementation`s here?

ggevay:
Thanks for the review!

> The join skipping is an unwelcome bit of complexity (it's _almost_ the same logic), but here we are. Perhaps an issue for another time, but how hard would it be to actually look at `JoinImplementation`s here?

It wouldn't be impossible:
- We'd have to attend to all the scalar expressions in the `JoinImplementation` enum. 
- We'd have to attend to the ArrangeBys below the joins.
- (I think `JoinInputCharacteristics` are unchanged. E.g., key fields are demanded, so key length is not changing.)
- (I think we don't need to touch the `Row`s that are inside `JoinImplementation::IndexedFilter`, because those fields are involved in join conditions, so are demanded.)

One reason for which I decided to not go for `JoinImplementation` is that in the long run we want to move the `JoinImplementation` transform to just before the MIR lowering, and then manipulating the `JoinImplementation` enum in other transforms won't be needed anymore.

ggevay:
It's good that I've run Nightly, because there is an EXPLAIN output change :sweat_smile: Discussing [here](https://materializeinc.slack.com/archives/C01LKF361MZ/p1741872074245179).

def-:
> Edit: I've added the lts-backport-v25.1 label, but if this turns out to be too annoying to cherry-pick, then it's ok to skip this, as it's just a performance improvement.

There is a bunch of conflicts indeed. I'm not sure if we should backport these kinds of changes in general, since there is also some risk for them.

PR #31808, opened 2025-03-09
Assert that there are no more projection pushdown opportunities at the end of the MIR pipeline, according to `Demand`

ggevay:
This PR adds a check for an optimizer invariant: There should be no more projection pushdown opportunities at the end of the MIR pipeline, according to `Demand`.

Compared to what we discussed on Friday with @frankmcsherry and @mgree, the only change is that we skip checking the invariant when there is an `ArrangeBy` on top of a global Get, because a projection might have been lifted away by `JoinImplementation` to re-use an index.

(I've run this locally on some slts, but not all. We'll see what CI says.)

The first commits are from https://github.com/MaterializeInc/materialize/pull/31791

### Motivation

   * This PR adds a soft assert.

### Checklist

### DIFF
--- a/src/transform/src/demand.rs
+++ b/src/transform/src/demand.rs
@@ -10,14 +10,15 @@
 //! Transformation based on pushing demand information about columns toward sources.
 
 use itertools::Itertools;
-use mz_ore::assert_none;
 use std::collections::{BTreeMap, BTreeSet};
 
+use mz_expr::visit::Visit;
 use mz_expr::{
-    AggregateExpr, AggregateFunc, Id, JoinInputMapper, MirRelationExpr, MirScalarExpr,
-    RECURSION_LIMIT,
+    AggregateExpr, AggregateFunc, Id, JoinInputMapper, MapFilterProject, MirRelationExpr,
+    MirScalarExpr, RECURSION_LIMIT,
 };
 use mz_ore::stack::{CheckedRecursion, RecursionGuard};
+use mz_ore::{assert_none, soft_assert_or_log};
 use mz_repr::{Datum, Row};
 
 use crate::TransformCtx;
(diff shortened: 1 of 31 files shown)


src/transform/src/demand.rs
```diff
@@ -365,4 +366,65 @@ impl Demand {
...
+        // We use `visit_pre_post`, because we'd like to control how we descend to children: we want
+        // to jump over an MFP that we find at the current node, in order to avoid observing partial
+        // MFPs, that are without their Projects.
+        relation.visit_pre_post(&mut |expr| {
+            let (mfp, expr) = MapFilterProject::extract_from_expression(expr);
+            match expr {
+                MirRelationExpr::Get { id: Id::Global(id), .. } => {
+                    let demand = demand_on_ids.get(&Id::Global(*id)).expect("`Demand` should have an opinion on all ids");
+                    let actual_proj_above_get = mfp.projection.iter().filter(|c| **c < mfp.input_arity).collect_vec();
+                    soft_assert_or_log!(
```
mgree, line 407: Should `relation.pretty()` and `expr.pretty()` and `mfp` be redacted?

mgree (commented):
Fundamentally okay by me, but see my comment. Is the plan to merge just this and drop #31791.

This feature flag boilerplate is a little much---very redundant changes across 6 files. If feature flags are the way forward for now, we'll likely want to find a nicer way to manage the flagging.

src/transform/src/demand.rs
ggevay: Good point!
- `pretty` already redacts in prod: `pretty` calls `ExplainConfig::default()`, which has
`redacted: !mz_ore::assert::soft_assertions_enabled()`.
- `Display for MapFilterProject` unfortunately doesn't redact, and would be hard to make it do so properly. But since `Display for MapFilterProject` is only used in debugging code, I've now tweaked it to just omit the whole MFP if `!soft_assertions_enabled()`.

ggevay:
Thanks for the review!

> Is the plan to merge just this and drop https://github.com/MaterializeInc/materialize/pull/31791.

No, I will merge that one first, then rebase on main, and then merge this one. (I'm currently fighting it out with Nightly over there.)

> This feature flag boilerplate is a little much---very redundant changes across 6 files. If feature flags are the way forward for now, we'll likely want to find a nicer way to manage the flagging.

Yeah, we could look into this.

mgree:
> > This feature flag boilerplate is a little much---very redundant changes across 6 files. If feature flags are the way forward for now, we'll likely want to find a nicer way to manage the flagging.
> 
> Yeah, we could look into this.

Made an issue: https://github.com/MaterializeInc/database-issues/issues/9099. Good luck with the RQG 🫣

PR #31843, opened 2025-03-10
optimizer: Call canonicalize_equivalence_classes in ReductionPushdown

ggevay:
Draft fix for https://github.com/MaterializeInc/database-issues/issues/9013

### Checklist


frankmcsherry (approved):
The fix probably fixes the issue, but it feels like the larger issue is that we have `MRE::Join` nodes with non-conforming equivalences. Ideally `Join::equivalences` would be equivalence classes, and if not the bug is with the operator that introduces them. The field is document as being a sequence of equivalence classes, and the standard math definition of such is the classes of an equivalence relation closed under transitivity.

It makes sense to fix the symptom asap, but we'll need to fix the cause also.

mgree (approved):
Talked this over out-of-band. Looks good to me, but I'll speak briefly with Frank about this transformation to make sure I understand it correctly.

ggevay:
Thanks for the reviews! Merging.

I've created an issue about maybe making `Join::equivalences` always be in canonical form. https://github.com/MaterializeInc/database-issues/issues/9069

PR #31845, opened 2025-03-10
doc: add note about Aurora PostgreSQL 16.1 logical replication bug

bobbyiliev:
### Motivation

Adding a small note about a known issue with Aurora PostgreSQL 16.1 that affects logical replication:

> https://repost.aws/questions/QU4RXUrLNQS_2oSwV34pmwww/error-could-not-map-filenumber-after-aurora-upgrade-to-16-1

I've wrapped this in a warning tag, but I'm open to suggestions if we should use a different formatting approach or tag type for this information.


kay-kim (approved):
lgtm

PR #31851, opened 2025-03-10
s3 metrics: more granular errors

martykulma:
Expands S3 metrics to break out errors by operation and error code.

I opted not to make these too granular, as that can create a lot of noise in telemetry that makes it harder to diagnose issues. If the granularity isn't sufficient, we can always improve it.

There is a minor overlap between the existing S3 timeout metrics and the metrics this PR adds.  The existing timeout metrics provide more granularity with regards to where the timeout occurred (connection timeout, TTFB, etc.), while the metrics in the PR provide context of which operation timed out (GetObject, CreateMultipartUpload, etc.).  It isn't clear to me which one is more useful.

There wasn't a nice way to test that that was worth checking in. I wrote a few unit tests and tested by overriding the AWS env vars to point to local services.  I tried to plumb in an `impl Intercept`, but that set of changes kept growing as it requires a lot of plumbing to get things into mz_aws_util where we create the S3 configuration and finally the client.

### Motivation

fix https://github.com/MaterializeInc/incidents-and-escalations/issues/195

### Checklist

### DIFF
--- a/src/persist/src/s3.rs
+++ b/src/persist/src/s3.rs
@@ -950,20 +950,22 @@ impl S3Blob {
                 None => "UnknownServiceError",
             },
             SdkError::DispatchFailure(e) => {
-                let error_kind = if let Some(connector_error) = e.as_connector_error() {
-                    connector_error.as_other()
-                } else {
-                    e.as_other()
-                };
-                match error_kind {
-                    Some(error_kind) => match error_kind {
+                if let Some(other_error) = e.as_other() {
+                    match other_error {
                         aws_config::retry::ErrorKind::TransientError => "TransientError",
                         aws_config::retry::ErrorKind::ThrottlingError => "ThrottlingError",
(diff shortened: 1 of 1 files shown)


bkirwi:
Neat! Mind deploying this to staging, so we can see how the metrics look in a live environment?

src/persist/src/s3.rs
```diff
@@ -908,6 +939,65 @@ impl S3Blob {
         );
         Ok(())
     }
+
+    fn update_error_metrics<E, R>(&self, op: &str, err: &SdkError<E, R>)
+    where
+        E: ProvideErrorMetadata,
+    {
+        match err {
```
bkirwi, line 947: Nit: I'd consider returning the `(op, code)` tuple from this match and then doing the `self.metrics.error_counts.with_label_values(&[op, label]).inc()` call at the end. Less duplication, and makes it clearer that the metric will always be incremented once.

PR #31859, opened 2025-03-11
Add system catalog dumper for self-managed debug tool

SangJunBak:
- See commit messages for details

### Motivation

### Checklist


SangJunBak:
I just realized it's probably a bad idea to expose tracing as stdout/stderr... 

@ParkMyCar Should we revert back to `println/eprintln`? At the end, I also want to take everything we printed to stdout, stderr, and maybe some additional debug info, then include that as a file in the debug zip.

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -206,3 +217,577 @@ pub fn spawn_sql_port_forwarding_process(
...
+        name: "mz_kafka_sinks",
+        category: RelationCategory::Basic,
+    },
+];
+
+const TIMEOUT: Duration = Duration::from_secs(60);
+
+#[derive(Debug, Clone)]
+pub struct ClusterReplica {
+    // TODO (debug_tool1): Use this when scraping introspection relations
```
ParkMyCar, line 622: love the use of these TODOs!

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -206,3 +217,577 @@ pub fn spawn_sql_port_forwarding_process(
...
+    context: &'n Context,
+    pg_client: Arc<PgClient>,
+    // TODO (debug_tool1): Use this when scraping introspection relations
+    _cluster_replicas: Vec<ClusterReplica>,
+    _pg_conn_handle: JoinHandle<Result<(), tokio_postgres::Error>>,
+}
+
+impl<'n> SystemCatalogDumper<'n> {
+    pub async fn new(context: &'n Context, connection_string: &str) -> Result<Self, anyhow::Error> {
+        let (pg_client, pg_conn) = retry::Retry::default()
```
ParkMyCar, line 637: FWIW we have our own `PostgresClient` [here](https://github.com/MaterializeInc/materialize/blob/main/src/postgres-client/src/lib.rs#L94-L96) that handles a lot of nice things like connection pooling.

Not necessary to use now but might be worth taking a look!

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -206,3 +217,577 @@ pub fn spawn_sql_port_forwarding_process(
...
+
+        let handle = task::spawn(|| "postgres-connection", pg_conn);
+
+        // Set search path to system catalog tables
+        pg_client
+            .execute(
+                "SET search_path = mz_internal, mz_catalog, mz_introspection",
+                &[],
+            )
+            .await?;
```
ParkMyCar, line 657: would be nice to add some `.context("set search path")` here, and to the other DB calls! Otherwise when we get an error back we don't know which query caused the error to fail

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -206,3 +217,577 @@ pub fn spawn_sql_port_forwarding_process(
...
+
+        info!("Connected to PostgreSQL server at {}...", connection_string);
+
+        let handle = task::spawn(|| "postgres-connection", pg_conn);
+
+        // Set search path to system catalog tables
+        pg_client
+            .execute(
+                "SET search_path = mz_internal, mz_catalog, mz_introspection",
```
ParkMyCar, line 654: Slight preference to moving these queries to a static string, e.g.
```
static SET_SEARCH_PATH_QUERY: &str = "SET search_path = mz_internal, mz_catalog, mz_introspection";
```
but up to you!

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -206,3 +217,577 @@ pub fn spawn_sql_port_forwarding_process(
...
+            .await?;
+
+        let cluster_replicas = match pg_client
+            .query("SELECT c.name as cluster_name, cr.name as replica_name FROM mz_clusters AS c JOIN mz_cluster_replicas AS cr ON c.id = cr.cluster_id;", &[])
+            .await
+            {
+                Ok(rows) => rows.into_iter()
+                            .map(|row| ClusterReplica {
+                                _cluster_name: row.get::<_, String>("cluster_name"),
+                                _replica_name: row.get::<_, String>("replica_name"),
```
ParkMyCar, line 666: `get(...)` panics if the column doesn't exist, probably better to use `try_get(...)` here! e.g. something like this should get you close:
```
rows.into_iter()
    .map(|row| Ok(ClusterReplica {
        _cluster_name: row.try_get::<_, String>("cluster_name")?,
        _replica_name: row.try_get::<_, String>("replica_name")?,
    }))
    .collect::<Result<Vec<_>, _>>(),
```

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -206,3 +217,577 @@ pub fn spawn_sql_port_forwarding_process(
...
+
+                    async move {
+                        match relation_category {
+                            RelationCategory::Introspection => {
+                                // TODO (debug_tool1): Create different outputs depending on the relation category
+                                return Ok::<(), anyhow::Error>(());
+                            }
+                            _ => {}
+                        }
+                        // TODO: Change None based on relation category
```
ParkMyCar, line 714: nit, use `TODO (debug_tool)` here

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -206,3 +217,577 @@ pub fn spawn_sql_port_forwarding_process(
...
+}
+
+#[derive(Debug, Clone)]
+pub struct Relation {
+    pub name: &'static str,
+    pub category: RelationCategory,
+}
+
+const RELATIONS: &[Relation] = &[
+    // Basic object information
```
ParkMyCar, line 239: nit: can you move this to a doc comment on the `const RELATIONS`?

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -206,3 +217,577 @@ pub fn spawn_sql_port_forwarding_process(
...
+    Basic,
+}
+
+#[derive(Debug, Clone)]
+pub struct Relation {
+    pub name: &'static str,
+    pub category: RelationCategory,
+}
+
+const RELATIONS: &[Relation] = &[
```
ParkMyCar, line 238: `static` probably works better here instead of `const`.

When something is `const` the compiler will inline the entire list wherever it's used, so this list will get copied around a bunch. When it's `static` we'll have a single copy in static memory that we pass around references to.

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -206,3 +217,577 @@ pub fn spawn_sql_port_forwarding_process(
...
+    },
+    Relation {
+        name: "mz_object_dependencies",
+        category: RelationCategory::Basic,
+    },
+    Relation {
+        name: "mz_object_transitive_dependencies",
+        category: RelationCategory::Basic,
+    },
+    // Compute
```
ParkMyCar, line 312: ohhhh I see now, these comments categorize the types of relations we're querying. Mind calling _that_ out in a doc comment so when folks add new relations they know they're sorted?

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -206,3 +217,577 @@ pub fn spawn_sql_port_forwarding_process(
...
+fn format_file_path(date_time: DateTime<Utc>, replica_name: Option<&str>) -> PathBuf {
+    let path = format_base_path(date_time).join("system-catalog");
+    if let Some(replica_name) = replica_name {
+        path.join(replica_name)
+    } else {
+        path
+    }
+}
+
+pub fn create_postgres_connection_string(
```
ParkMyCar, line 775: Can you document the defaults that get used if params are not provided?

src/self-managed-debug/src/utils.rs
```diff
@@ -0,0 +1,24 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::path::PathBuf;
+
+use chrono::{DateTime, Utc};
+
+pub fn format_base_path(date_time: DateTime<Utc>) -> PathBuf {
```
ParkMyCar, line 20: What does base_path mean in this context? I'm guessing it's where we write the data dump to? Would be great to document

src/self-managed-debug/src/k8s_resource_dumper.rs
```diff
@@ -366,13 +367,8 @@ fn format_resource_path(
...
     namespace: Option<&String>,
 ) -> PathBuf {
-    let mut path = PathBuf::from(format!(
-        "mz-debug{}{}{}{}",
-        MAIN_SEPARATOR,
-        date_time.format("%Y-%m-%dT%H:%MZ"),
-        MAIN_SEPARATOR,
-        resource_type,
-    ));
+    let mut path = format_base_path(date_time).join(resource_type);
```
ParkMyCar, line 370: Not necessarily in this PR, but just realizing now, what happens if this path already exists?

SangJunBak: Ah so at the very end, once the full path is composed. i.e. in `system_catalog_dumper` lines 766 to 768:
```
            let file_path = format_file_path(start_time, None);
            let file_path_name = file_path.join(relation_name).with_extension("csv");
            tokio::fs::create_dir_all(&file_path).await?;
```

We do a `create_dir_all` which recursively creates the `dir` if missing:

Screenshot 2025-03-20 at 12 41 04 AM

SangJunBak:
@ParkMyCar Thanks for the review! Addressed all your feedback. but in https://github.com/MaterializeInc/materialize/pull/31951 from commit  [Annotate postgresql query errors](https://github.com/MaterializeInc/materialize/pull/31951/commits/317bcdac40b5d7bd66318e139a3b9da0ee50246c) and onwards if that's cool! For the tracing feedback you gave, I actually wanna print out to the user what I'm currently tracing! But it's good to know that I can just change the tracing subscriber logic. I'll most likely clean it up when I'm implementing the zipping at the end!

At the end, I'll merge both PRs into a single PR then squash & commit!

PR #31861, opened 2025-03-11
Push `MSE::If` into expressions

frankmcsherry:
When we have an expression `IF <cond> THEN e1 ELSE e2`, where `e1` and `e2` are the same operator except for one argument (e.g. the same unary operator, or the same binary operator and one input matches), we can push the condition into the non-matching branch. We can actually do this more generally, always pushing the condition down when the operators match, but it's less certainly a win in those cases.

One case that the conservative approach will miss is `IF <cond> THEN a = b ELSE c = d`. This could be planned as
```
(IF <cond> THEN a ELSE c) = (IF <cond> THEN b ELSE d)
```
which presents as an equivalence, for join purposes, in a way that the `IF`-wrapped expression does not.

This is equivalent to hoisting some logic out of the `If`, though it is logic evaluated in both branches, which I believe means it is safe to hoist (any erroring behavior would occur in either case).

### Checklist

### DIFF
--- a/src/expr/src/scalar.rs
+++ b/src/expr/src/scalar.rs
@@ -1358,6 +1358,55 @@ impl MirScalarExpr {
                                 }
                                 _ => {}
                             }
+                        } else {
+                            // Equivalent expression structure would allow us to push the `If` into the expression.
+                            // For example, `IF <cond> THEN x = y ELSE x = z` becomes `x = IF <cond> THEN y ELSE z`.
+                            match (&mut **then, &mut **els) {
+                                (
+                                    MirScalarExpr::CallUnary { func: f1, expr: e1 },
+                                    MirScalarExpr::CallUnary { func: f2, expr: e2 },
+                                ) if f1 == f2 => {
+                                    *e = cond
+                                        .take()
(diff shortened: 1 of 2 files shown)


antiguru (approved):
Looks good; thank you!

frankmcsherry:
@ggevay I don't mind adding tests, but the changes in slts seem like they might count already. In the interest of being a grown up, it should probably be feature flagged as well, but I still need to figure that out for the other PR.

ggevay:
Ok, the slt changes seem enough. (I think they weren't there when I wrote the comment about tests.)

Yes, if you could add a feature flag, that would be good!

frankmcsherry:
> Ok, the slt changes seem enough. (I think they weren't there yet when I wrote the comment about tests.)

100%; I wasn't sure if there would be diffs, and if not then tests make sense. I'll add the flags, having gotten around it for the other PR!

frankmcsherry:
Update: not going to feature flag, unless we have a theory about how to thread the flag through `MSE::reduce`, which is used pervasively. If that doesn't work for folks, we can discuss how to move forward, but I don't think its worth the squeeze in this case.

PR #31862, opened 2025-03-11
[persist] skip generating `Codec` data in compaction if it's not necessary

ParkMyCar:
Something small I noticed when checking what metrics we have to measure our structured rollout.

My read of this code is the only time we call `BlobTraceUpdates::concat` is when we're compacting runs and our preferred `SortOrder` is `Structured`. In `concat(...)` we currently always ensure `Codec` data exists, which AFAICT means we always decode and then re-encode structured data into `Codec` data. Right after calling `concat(...)` we call `flush_part(...)` which if our `batch_columnar_format` is structured will throw away the `Codec` data we just generated.

This PR adds a bit of new logic to gate when we generate the `Codec` data in `concat(...)`. If a CYA dyncfg is enabled and our batch columnar format is structured, then we'll skip generating `Codec` data. Given we're currently rolling out our columnar format, this will soon be the common case.

@bkirwi it's totally possible I missed something and we still need Codec data here, but it seems like with the right flags on this code path operates entirely on structured data!

### Motivation

(Hopefully) Make compaction a bit faster by skipping unnecessary work

### Checklist


bkirwi (approved):
Nice catch!

bkirwi:
Oh, and we should get nightlies going if you haven't already.

ParkMyCar:
> Oh, and we should get nightlies going if you haven't already.

Good call! Just kicked them off https://buildkite.com/materialize/nightly/builds/11435

PR #31868, opened 2025-03-11
Fix a memory leak in arrangement size logging

antiguru:
This fixes a memory leak in arrangement size logging, which was
inadvertently introduced when combining the logging dataflows into a single
dataflow and adding streams to share information between different
fragments. The problem is that the data needs to be processed in order,
i.e., the handler needs to see the announcement of a new arrangement
followed by updates, and at the end an indication that it was dropped.
Passing data only through the log infrastructure ensures a total order, as
it maintains FIFO semantics. Adding additional channels to the mix removed
the property that data arrives in order, rendering the log analysis
incorrect.

The fix in the change is only a partial fix. It reverts to the previous
behavior that all data needs to pass through the logging infrastructure to
ensure its FIFO semantics, but it doesn't address the fact that the replay
operator doesn't maintain ordering explicitly -- Timely channels aren't
guaranteed to preserve order, even if the implementation currently does.
A proper fix is to introduce a sequencer, but that needs work to not
regress in performance.

### Checklist


bosconi:
Fixes https://github.com/MaterializeInc/database-issues/issues/9054

antiguru:
Thanks for the review!

PR #31869, opened 2025-03-11
[sinks] Config checks for progress topics

bkirwi:
Attempt to fetch the config of already-existing topics, so we can confirm that the configs match what we'd expect.

### Motivation

https://github.com/MaterializeInc/database-issues/issues/9032

### Tips for reviewer

I'd prefer to have `ensure_topic` actually attempt to update the configuration of the topic when it can, because I'd expect that to make the overall system easier to reason about. However, we don't necessarily have the right permissions to do that... and this should give us a better sense of how urgent that work might be.

### Checklist

### DIFF
--- a/src/kafka-util/src/admin.rs
+++ b/src/kafka-util/src/admin.rs
@@ -13,11 +13,44 @@ use std::iter;
 use std::time::Duration;
 
 use anyhow::{anyhow, bail};
+use itertools::Itertools;
 use mz_ore::collections::CollectionExt;
 use mz_ore::retry::Retry;
-use rdkafka::admin::{AdminClient, AdminOptions, NewTopic};
+use rdkafka::admin::{
+    AdminClient, AdminOptions, ConfigEntry, ConfigResource, NewTopic, OwnedResourceSpecifier,
+    ResourceSpecifier,
+};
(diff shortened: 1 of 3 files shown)


src/kafka-util/src/admin.rs
```diff
@@ -103,51 +158,24 @@ where
...
+                        new_topic.num_partitions
+                    );
                 }
             }
             Ok(false)
         })
         .await
 }
 
-/// An error while creating a Kafka topic.
```
aljoscha, line 125: you inlined this one because each variant is only used in one place, and as it was before it was a  whole bunch of boilerplate?

aljoscha (approved):
🙌

src/kafka-util/src/admin.rs
bkirwi: Brief rationale in https://github.com/MaterializeInc/materialize/pull/31869/commits/0297e073c3aebf51e56cc35c075dc5fe2ea11178

For me the main motive was... all the callers just convert to anyhow anyways. I had more logic (and more possible errors) to add, so it seemed like useless work to reshape the structure to add my own variants etc. when all that structure was immediately thrown away.

PR #31870, opened 2025-03-11
docs: housecleaning ... use relative docs links

kay-kim:
As we do in the self-managed docs, we should in the cloud docs use relative links instead of absolute `https://materialize.com/docs/blah`.


frankmcsherry (approved):
Looks great.

PR #31872, opened 2025-03-12
Set reduced_count in storage command history

antiguru:
We never set the `reduced_count` to the last reduced length, causing us to
reduce the storage command history every time we'd push a command to it.
The fix is simple: set the reduced count at the end of the reduce function.

It's hard to measure the impact of this because we limit the number of storage objects to a low two-digit (base 10) number. In local tests, this problem caused environmentd's coordinator thread to spend more than half its time in reducing the storage history.

Related: MaterializeInc/incidents-and-escalations#211

### Checklist


frankmcsherry (approved):
Looks good to me. Are we missing a clippy pass that might have told us that we never write to this field?

PR #31873, opened 2025-03-12
Maintenance interval to rate-limit expensive tasks

antiguru:
Similarly to what compute has been doing for a while, introduce a
maintenance interval in the storage server to rate limit specific tasks.
At the moment, this only covers reporting frontiers to the controller. The
downside is that we might delay reporting frontier updates by the
maintenance interval.

At the moment, the maintenance interval is set to 10ms, the same as used by
compute. The largest expected frontier reporting regression is thus 10ms.

Related: https://github.com/MaterializeInc/incidents-and-escalations/issues/211

### Checklist

### DIFF
--- a/src/storage-types/src/dyncfgs.rs
+++ b/src/storage-types/src/dyncfgs.rs
@@ -249,6 +249,13 @@ pub const STORAGE_USE_CONTINUAL_FEEDBACK_UPSERT: Config<bool> = Config::new(
     "Whether to use the new continual feedback upsert operator.",
 );
 
+/// The interval at which the compute server performs maintenance tasks.
+pub const STORAGE_SERVER_MAINTENANCE_INTERVAL: Config<Duration> = Config::new(
+    "storage_server_maintenance_interval",
+    Duration::from_millis(10),
+    "The interval at which the storage server performs maintenance tasks. Zero enables maintenance on every iteration.",
+);
+
 /// Adds the full set of all storage `Config`s.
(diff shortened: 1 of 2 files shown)


src/storage-types/src/dyncfgs.rs
```diff
@@ -249,6 +249,13 @@ pub const STORAGE_USE_CONTINUAL_FEEDBACK_UPSERT: Config<bool> = Config::new(
     "Whether to use the new continual feedback upsert operator.",
 );
 
+/// The interval at which the compute server performs maintenance tasks.
```
aljoscha, line 252:
```suggestion
/// The interval at which the storage server performs maintenance tasks.
```

src/storage/src/storage_state.rs
```diff
@@ -417,11 +424,32 @@ impl<'w, A: Allocate> Worker<'w, A> {
...
 
+        // The last time we did periodic maintenance.
+        let mut last_maintenance = std::time::Instant::now();
+
         let mut disconnected = false;
         while !disconnected {
             let config = &self.storage_state.storage_configuration;
             let stats_interval = config.parameters.statistics_collection_interval;
 
+            // Get the maintenance interval, default to zero if we don't have a compute state.
```
aljoscha, line 435:
```suggestion
            // Get the maintenance interval, default to zero if we don't have a storage state.
```

aljoscha (approved):
Looks good! I don't know if the code is sensitive to where in this method we report frontiers, but I think that's fine.

PR #31879, opened 2025-03-12
[sinks] Iteratively search through the progress topic

bkirwi:
### Motivation

https://github.com/MaterializeInc/database-issues/issues/9032#issuecomment-2711978494

### Tips for reviewer

This behaviour is difficult to meaningfully test. I've tried to isolate the semantic changes as much as possible... consider reviewing commit-by-commit to be convinced of this.

### Checklist

### DIFF
--- a/src/storage-types/src/dyncfgs.rs
+++ b/src/storage-types/src/dyncfgs.rs
@@ -256,6 +256,13 @@ pub const STORAGE_SERVER_MAINTENANCE_INTERVAL: Config<Duration> = Config::new(
     "The interval at which the storage server performs maintenance tasks. Zero enables maintenance on every iteration.",
 );
 
+/// If set, iteratively search the progress topic for a progress record with increasing lookback.
+pub const SINK_PROGRESS_SEARCH: Config<bool> = Config::new(
+    "storage_sink_progress_search",
+    true,
+    "If set, iteratively search the progress topic for a progress record with increasing lookback.",
+);
+
 /// Adds the full set of all storage `Config`s.
 pub fn all_dyncfgs(configs: ConfigSet) -> ConfigSet {
     configs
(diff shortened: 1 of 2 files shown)


src/storage/src/sink/kafka.rs
```diff
@@ -1019,17 +1019,33 @@ async fn determine_sink_progress(
...
-            Arc::clone(&metrics)
-        )
+        // This topic might be long, but the desired offset will usually be right near the end.
+        // Instead of always scanning through the entire topic, we scan through exponentially-growing
+        // suffixes of it. (Because writes are ordered, the largest progress record in any suffix,
+        // if present, is the global max.) If we find it in one of our suffixes, we've saved at least
+        // an order of magnitude of work; if we don't, we've added at most a constant factor.
+        let mut start_indices = vec![lo];
+        let mut lookback = hi.saturating_sub(lo) / 10;
+        while lookback >= 1000 {
```
aljoscha: You know off the top of your head how fast we can zip through 1000 records?

src/storage-types/src/dyncfgs.rs
```diff
@@ -256,6 +256,13 @@ pub const STORAGE_SERVER_MAINTENANCE_INTERVAL: Config<Duration> = Config::new(
     "The interval at which the storage server performs maintenance tasks. Zero enables maintenance on every iteration.",
 );
 
+/// The interval at which the storage server performs maintenance tasks.
```
aljoscha, line 259: "Whether sinks iteratively search through the progress topic, with exponentially larger lookback."

aljoscha (approved):
This looks nice! A failing nightly test is concerning but this one felt unrelated.

src/storage-types/src/dyncfgs.rs
bkirwi: 🤦 - thanks.

src/storage/src/sink/kafka.rs
bkirwi: Quite fast normally - I think we saw at least tens of thousands of records per second for the slow catchups we were looking at recently. Could go higher if this seems low?

bkirwi: I have now increased this to 20_000. That means that on average, for a large topic, we'll start by scanning the latest 20k to 200k records, which for the typical sink looks like it will take a couple of seconds and should include the most recent progress record even for environments with many sinks. Happy to follow up if that seems like the wrong tradeoff!

bkirwi:
Merging to make the release cut - if you have any followup feedback on the threshold I'm happy to patch or do a followup PR.

PR #31880, opened 2025-03-13
storage/pg: produce probes during snapshotting

teskje:
This PR fixes a defect in the postgres source where it would only start producing upstream probes once it was finished with snapshotting and started replication. The result of which being that with RLU enabled, the entire snaphot would have to be buffered in memory because no timestamp bindings would be produced during snapshotting, and thus no updates could transit the reclock operator.

This fix adds a separate "statistics" operator to the postgres source, like the mysql source already has. As in the mysql source, this operator is responsible for issueing upstream probes, so probing is decoupled from the replication process and can start immediately when the source is rendered.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9061

### Checklist

### DIFF
--- a/src/storage/src/source/postgres.rs
+++ b/src/storage/src/source/postgres.rs
@@ -81,6 +81,7 @@
 
 use std::collections::BTreeMap;
 use std::convert::Infallible;
+use std::pin::pin;
 use std::rc::Rc;
 use std::time::Duration;
 
@@ -106,8 +107,11 @@ use timely::dataflow::operators::core::Partition;
 use timely::dataflow::operators::{Concat, Map, ToStream};
 use timely::dataflow::{Scope, Stream};
 use timely::progress::Antichain;
+use tokio::sync::mpsc;
 use tokio_postgres::error::SqlState;
 use tokio_postgres::types::PgLsn;
+use tokio_stream::wrappers::UnboundedReceiverStream;
+use tokio_stream::StreamExt;
 
 use crate::healthcheck::{HealthStatusMessage, HealthStatusUpdate, StatusNamespace};
 use crate::source::types::{Probe, ProgressStatisticsUpdate, SourceRender, StackedCollection};
@@ -115,6 +119,7 @@ use crate::source::{RawSourceCreationConfig, SourceMessage};
 
 mod replication;
 mod snapshot;
+mod statistics;
 
 impl SourceRender for PostgresSourceConnection {
     type Time = MzOffset;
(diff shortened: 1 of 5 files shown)


src/storage/src/source/postgres.rs
```diff
@@ -172,6 +177,24 @@ impl SourceRender for PostgresSourceConnection {
...
+            let mut resume_uppers = pin!(resume_uppers);
+            while let Some(upper) = resume_uppers.next().await {
+                let r1 = resume_uppers_tx_1.send(upper.clone());
+                let r2 = resume_uppers_tx_2.send(upper);
+
+                if r1.is_err() && r2.is_err() {
+                    break;  // both receivers disconnected
+                }
+            }
+        });
```
teskje, line 196: This part is a major smell, but I don't know how to avoid it. The `statistics` operator needs to look at the committed uppers to write them into its statistics updates. The `replication` operator needs to look at them to send keepalives back to the PG instance.

It seems like the stream we get here originally comes from a `watch`, so we could consider passing in the `watch` instead of an opaque `impl Stream`. But that'd probably require touching a whole bunch of unrelated code.

teskje:
The approach chosen here doesn't seem to work. It makes some of the `CREATE TABLE ... FROM SOURCE` tests get stuck in a "slot overcompacted" loop. The postgres source keeps restarting, but just finds the slow overcompacted the same way when it comes back up.

As far as I can tell, this happens:
* A `CREATE SOURCE` installs a bare postgres source without exports.
* The first probe at `probe_ts = T1` reports an upstream lsn `L1`, and the remap operator mints a binding accordingly.
* The source snapshot is committed at time `T1`.
* The replication slot is created and starts at an lsn `L2` > `L1`.
* A `CREATE TABLE ... FROM SOURCE` statement forces the source dataflow to be recreated.
* The async worker looks at the frontier of the committed data and finds `T1` as the resume upper and `L1` as the resume lsn (I assume through mapping through the remap collection).
* Now we are trying to resume the source at lsn `L1` but the replication slot only starts later at `L2` and so source creation fails.

It's not clear to me which part violates the spec here, or why the same approach seems to work for the mysql source.

petrosagg:
> It's not clear to me which part violates the spec here,

Producing a binding for lsn `T1` puts an obligation to the source to produce the state of tables at `T1`. But if the replication slot is created at some lsn `T2 > T1` this obligation is impossible to fulfill. We can only produce accurate states of tables for LSNs that are beyond the slot LSN.

So we must take into account the slot's initial LSN before producing probes.

> or why the same approach seems to work for the mysql source.

MySQL doesn't have the concept of replication slots and will happily give you the WAL updates if they are available on disk

teskje:
Closing in favor of https://github.com/MaterializeInc/materialize/pull/31912

PR #31883, opened 2025-03-13
[ore] Avoid unnecessary copies of shared lgalloc data

bkirwi:
Recent versions of `bytes` have a [`Bytes::from_owner`](https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.from_owner) method specially designed for lgalloc-like data.

Switching some of our custom types over to `Bytes` allows us to avoid copies in many cases... including in `SegmentedBytes::copy_to_bytes`, which is used heavily in protobuf decoding etc., and can now just return a reference to a single segment where possible.

### Motivation

https://github.com/MaterializeInc/database-issues/issues/9060#event-16723742556

Whether or not we're worried about Persist's overall memory usage, being able to avoid reallocating lgalloc data for as long as possible ought to be a good thing across the board.

### Checklist


antiguru (approved):
Not too familiar with the code, but seems like a good idea!

src/ore/src/bytes.rs
```diff
@@ -265,16 +207,23 @@ impl parquet::file::reader::ChunkReader for SegmentedBytes {
...
 
     fn get_bytes(&self, start: u64, length: usize) -> parquet::errors::Result<Bytes> {
-        let mut reader = self.clone().reader();
-        reader.seek(std::io::SeekFrom::Start(start))?;
-
-        // TODO(parkmycar): Make this possible to achieve zero-copy.
-        //
-        // It's currently non-trivial to do this because LgBytes is not easily reference counted.
-        let mut buf = vec![0; length];
-        reader.read_exact(&mut buf)?;
```
ParkMyCar, line 275: Thank you 🙏

ParkMyCar (approved):
Fantastic! Thank you!

PR #31884, opened 2025-03-14
compute,storage: replica connection metrics

teskje:
This PR adds a set of replica collection metrics, tracking reconnects of both compute and storage replicas. The hope is to improve observability into issues caused by unstable connections between the controller and its replicas.

### Motivation

Closes https://github.com/MaterializeInc/database-issues/issues/4895.

The issue is about compute only, but the observability is useful for storage replicas as well.

### Checklist

### DIFF
--- a/src/compute-client/src/metrics.rs
+++ b/src/compute-client/src/metrics.rs
@@ -20,17 +20,18 @@ use mz_ore::cast::CastFrom;
 use mz_ore::metric;
 use mz_ore::metrics::raw::UIntGaugeVec;
 use mz_ore::metrics::{
-    DeleteOnDropCounter, DeleteOnDropGauge, DeleteOnDropHistogram, HistogramVec, IntCounterVec,
-    MetricVecExt, MetricsRegistry,
+    CounterVec, DeleteOnDropCounter, DeleteOnDropGauge, DeleteOnDropHistogram, HistogramVec,
+    IntCounterVec, MetricVecExt, MetricsRegistry,
 };
 use mz_ore::stats::histogram_seconds_buckets;
 use mz_repr::GlobalId;
 use mz_service::codec::StatsCollector;
-use prometheus::core::AtomicU64;
+use prometheus::core::{AtomicF64, AtomicU64};
 
 use crate::protocol::command::{ComputeCommand, ProtoComputeCommand};
 use crate::protocol::response::{PeekResponse, ProtoComputeResponse};
 
+pub(crate) type Counter = DeleteOnDropCounter<AtomicF64, Vec<String>>;
 pub(crate) type IntCounter = DeleteOnDropCounter<AtomicU64, Vec<String>>;
 pub(crate) type UIntGauge = DeleteOnDropGauge<AtomicU64, Vec<String>>;
 type Histogram = DeleteOnDropHistogram<Vec<String>>;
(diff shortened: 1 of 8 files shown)


src/compute-client/src/metrics.rs
```diff
@@ -163,6 +169,21 @@ impl ComputeControllerMetrics {
...
+            replica_connects_total: metrics_registry.register(metric!(
+                name: "mz_compute_controller_replica_connects_total",
+                help: "The total number of replica (re-)connections made by the compute controller.",
+                var_labels: ["instance_id", "replica_id"],
+            )),
+            replica_connect_wait_time_seconds_total: metrics_registry.register(metric!(
+                name: "mz_compute_controller_replica_connect_wait_time_seconds_total",
+                help: "The total time the compute controller spent waiting for replica (re-)connection.",
+                var_labels: ["instance_id", "replica_id"],
+            )),
```
antiguru, line 186: Should this be a histogram?

src/storage-client/src/metrics.rs
```diff
@@ -151,6 +191,21 @@ pub struct ReplicaMetrics {
...
+        self.inner.replica_connects_total.inc();
+        self.observe_connect_time(wait_time);
+    }
+
+    /// Observe time spent waiting for a replica connection.
+    pub fn observe_connect_time(&self, wait_time: Duration) {
+        self.inner
+            .replica_connect_wait_time_seconds_total
+            .inc_by(wait_time.as_secs_f64());
+    }
```
antiguru, line 206: Nit: From the name alone, I wouldn't have `observe_connect` expected to accept a `wait_time` parameter. Consider calling `observe_connect_time` explicitly instead of threading it through `observe_connect`.

src/storage-controller/src/instance.rs
```diff
@@ -230,6 +233,22 @@ where
...
     }
+
+    /// Refresh the controller state metrics for this instance.
+    ///
+    /// We could also do state metric updates directly in response to state changes, but that would
+    /// mean littering the code with metric update calls. Encapsulating state metric maintenance in
+    /// a single method is less noisy.
+    ///
+    /// This method is invoked by `Controller::maintain`, which we expect to be called once per
+    /// second during normal operation.
```
antiguru, line 244: Cound the state be derived from the replica status, too? I think it might be worth leaving a comment to that note, if true. I agree with the implementation that it's easier to maintain the shared bools here than to wire up a SQL query, but we should still explain the design decision if there was any.

antiguru (approved):
Thank you! It's a but unfortunate that we need to duplicate the implementation for compute and storage, but there doesn't seem to be much we can do in the scope of this PR. I left some inline comments!

src/storage-controller/src/instance.rs
teskje: I don't think so! At least as it's implemented now `mz_cluster_replica_statuses` only has the k8s/pod-level status. That tells us nothing about the application-level status of the replica (e.g. whether the Timely cluster is initialized or whether it's connected to the controller).

src/compute-client/src/metrics.rs
teskje: My guess is that between the two metrics "number of connects" and "total wait time" we should have enough information to not require a histogram. At least it will let us differentiate between a replica that takes very long to connect and a replica in a connect-disconnect cycle. Can always add a histogram later if we notice that we are missing insight after all.

src/storage-controller/src/instance.rs
antiguru: Maybe we should include the actual connection status somewhere accessible through SQL? Might be worth capturing this in an issue.

teskje: I added a note to the existing issue about making the status more useful: https://github.com/MaterializeInc/database-issues/issues/4228

src/storage-client/src/metrics.rs
teskje: Done. I also restructured the two `ReplicaTask::connect` to make them more similar.

teskje:
TFTR!

PR #31885, opened 2025-03-14
Improve subsource planning

antiguru:
Improve subsource planning by making various parts of the adapter more scalable.

This PR includes changes to:
* The *durable transaction* implementation to improve updates and validity checks. Updates are more efficient by cloning less and passing references where possible. Validity checks are more efficient by converting some that assert on name equality to a $n\cdot\log n$, where it previously was $n^2$.
* *Planning subsources* is more efficient by allocating all IDs in one operation rather than one-by-one.
* Avoiding *computing trace data* in the storage client when the level isn't high enough.

Takes planning 10k subsources from I-terminated-it-after-one-hour to around one minute.

### Checklist

### DIFF
--- a/src/adapter/src/coord/sequencer/inner.rs
+++ b/src/adapter/src/coord/sequencer/inner.rs
@@ -409,16 +409,18 @@ impl Coordinator {
     /// Because of this, we have usually "missed" the opportunity to plan them
     /// through the normal statement execution life cycle (the exception being
     /// during bootstrapping).
+    ///
+    /// The caller needs to provide a `CatalogItemId` and `GlobalId` for the sub-source.
     pub(crate) async fn plan_subsource(
         &mut self,
         session: &Session,
         params: &mz_sql::plan::Params,
         subsource_stmt: CreateSubsourceStatement<mz_sql::names::Aug>,
+        item_id: CatalogItemId,
+        global_id: GlobalId,
     ) -> Result<CreateSourcePlanBundle, AdapterError> {
         let catalog = self.catalog().for_session(session);
         let resolved_ids = mz_sql::names::visit_dependencies(&catalog, &subsource_stmt);
-        let id_ts = self.get_catalog_write_ts().await;
-        let (item_id, global_id) = self.catalog_mut().allocate_user_id(id_ts).await?;
 
         let plan = self.plan_statement(
             session,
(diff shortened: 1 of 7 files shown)


src/adapter/src/coord/sequencer/inner.rs
```diff
@@ -467,9 +469,15 @@ impl Coordinator {
...
+        let id_ts = self.get_catalog_write_ts().await;
+        let mut ids = self
+            .catalog_mut()
+            .allocate_user_ids(u64::cast_from(subsource_stmts.len()), id_ts)
+            .await?;
         for subsource_stmt in subsource_stmts {
+            let (item_id, global_id) = ids.pop().unwrap();
             let s = self
-                .plan_subsource(session, &params, subsource_stmt)
+                .plan_subsource(session, &params, subsource_stmt, item_id, global_id)
```
antiguru, line 480: This changes the order of the IDs .. not sure if it's relevant!

src/catalog/src/durable/transaction.rs
```diff
@@ -3206,12 +3321,20 @@ mod tests {
         )
         .unwrap();
 
-        table
+        // Ideally, we compare for errors here, but it's hard/impossible to implement PartialEq
+        // for DurableCatalogError.
+        assert!(table
```
ParkMyCar, line 3326: Might be better to use `assert_ok!` here

ParkMyCar (approved):
Nice! LGTM!

src/adapter/src/coord/sequencer/inner.rs
```diff
@@ -467,10 +469,14 @@ impl Coordinator {
...
+            .allocate_user_ids(u64::cast_from(subsource_stmts.len()), id_ts)
+            .await?;
         for subsource_stmt in subsource_stmts {
-            let s = self
-                .plan_subsource(session, &params, subsource_stmt)
-                .await?;
+            let (item_id, global_id) = ids.pop().expect("allocated sufficient ids");
+            let s = self.plan_subsource(session, &params, subsource_stmt, item_id, global_id)?;
             subsource_plans.push(s);
         }
```
ParkMyCar, line 481: to retain the ID ordering we had before you could zip `ids` and `subsource` statements together, would eliminate the need for the `.pop`

src/adapter/src/coord/sequencer/inner.rs
ParkMyCar: We've been bitten by ID ordering before, would be great to retain the existing order. See my other comment related to using `.zip(...)`

ParkMyCar:
If you don't mind kicking off a round of Nightlies, it would be much appreciated!

antiguru:
Nightly in https://buildkite.com/materialize/nightly/builds/11470, but I couldn't trigger it through our tool for whatever reason.

src/catalog/src/durable/transaction.rs
antiguru: TIL!

src/adapter/src/coord/sequencer/inner.rs
antiguru: Yup, zipping is the obvious choice here!

PR #31890, opened 2025-03-15
storage: allow singleton sources on multi-replica clusters

aljoscha:
Implements https://github.com/MaterializeInc/database-issues/issues/9079

Per the design in https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/20250127_multi_replica_scheduling_singleton_sources.md

There's now an issue around tracking "active replicas" in the storage controller, for the purposes of tracking `DroppedId` messages. I have a good idea how to fix it but wanted to get this out for review now.

### DIFF
--- a/src/storage-controller/src/instance.rs
+++ b/src/storage-controller/src/instance.rs
@@ -90,9 +90,6 @@ pub(crate) struct Instance<T> {
 
 #[derive(Debug)]
 struct ActiveIngestion {
-    /// Whether the ingestion prefers running on a single replica.
-    prefers_single_replica: bool,
-
     /// The set of replicas that this ingestion is currently running on.
     active_replicas: BTreeSet<ReplicaId>,
 }
@@ -358,12 +355,6 @@ where
     /// in state and potentially update scheduling decisions.
     fn absorb_ingestions(&mut self, ingestions: Vec<RunIngestionCommand>) {
         for ingestion in ingestions {
-            let prefers_single_replica = ingestion
-                .description
-                .desc
-                .connection
-                .prefers_single_replica();
-
             let existing_ingestion_state = self.active_ingestions.get_mut(&ingestion.id);
 
             // Always update our mapping from export to their ingestion.
@@ -376,10 +367,6 @@ where
                 // change anything about our scheduling decisions, no need to
                 // update active_ingestions.
 
-                assert!(
-                    ingestion_state.prefers_single_replica == prefers_single_replica,
-                    "single-replica preference changed"
-                );
                 tracing::debug!(
                     ingestion_id = %ingestion.id,
                     active_replicas = %ingestion_state.active_replicas.iter().map(|id| id.to_string()).join(", "),
@@ -388,7 +375,6 @@ where
             } else {
                 // We create a new ingestion state for this ingestion.
                 let ingestion_state = ActiveIngestion {
-                    prefers_single_replica,
                     active_replicas: BTreeSet::new(),
                 };
                 self.active_ingestions.insert(ingestion.id, ingestion_state);
@@ -418,16 +404,39 @@ where
     /// If `send_commands` is true, will send commands for newly-scheduled
     /// single-replica ingestions.
     fn update_ingestion_scheduling(&mut self, send_commands: bool) {
-        // We have to collect first and then send later, because we're borrowing
-        // self mutably just below.
+        // We first collect scheduling preferences and then schedule below.
+        // Applying the decision needs a mutable borrow but we also need to
+        // borrow for determining `prefers_single_replica`, so we split this
+        // into two loops.
+        let mut scheduling_preferences: Vec<(GlobalId, bool)> = Vec::new();
+
+        for ingestion_id in self.active_ingestions.keys() {
+            let ingestion_description = self
+                .get_ingestion_description(ingestion_id)
+                .expect("missing ingestion description");
+
+            let prefers_single_replica = ingestion_description
+                .desc
+                .connection
+                .prefers_single_replica();
+
+            scheduling_preferences.push((*ingestion_id, prefers_single_replica));
+        }
+
+        // Collect all commands per replica and send them in one go.
         let mut commands_by_replica: BTreeMap<ReplicaId, Vec<GlobalId>> = BTreeMap::new();
 
-        for (ingestion_id, ingestion_state) in self.active_ingestions.iter_mut() {
-            if ingestion_state.prefers_single_replica {
+        for (ingestion_id, prefers_single_replica) in scheduling_preferences {
+            let ingestion_state = self
+                .active_ingestions
+                .get_mut(&ingestion_id)
+                .expect("missing ingestion state");
+
+            if prefers_single_replica {
                 // For single-replica ingestion, schedule only if it's not already running.
                 if ingestion_state.active_replicas.is_empty() {
-                    let first_replica = self.replicas.keys().min().copied();
-                    if let Some(first_replica_id) = first_replica {
+                    let target_replica = self.replicas.keys().min().copied();
+                    if let Some(first_replica_id) = target_replica {
                         tracing::info!(
                             ingestion_id = %ingestion_id,
                             replica_id = %first_replica_id,
@@ -437,7 +446,7 @@ where
                         commands_by_replica
                             .entry(first_replica_id)
                             .or_default()
-                            .push(ingestion_id.clone());
+                            .push(ingestion_id);
                     }
                 } else {
                     tracing::info!(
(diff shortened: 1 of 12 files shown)


aljoscha:
@bkirwi & @petrosagg If we're happy with this approach, we can also use it for sinks on multi-replica clusters.

src/storage-controller/src/instance.rs
```diff
@@ -82,6 +88,15 @@ pub(crate) struct Instance<T> {
     response_tx: mpsc::UnboundedSender<(Option<ReplicaId>, StorageResponse<T>)>,
 }
 
+#[derive(Debug)]
+struct ActiveIngestion {
+    /// Whether the ingestion prefers running on a single replica.
+    prefers_single_replica: bool,
```
petrosagg, line 94: Is this a property of an active ingestion? I would expect this information to be found in the collection state for the given `GlobalId`. Having it here implies that it can potentially disagree with the ingestion, right?

src/storage-controller/src/instance.rs
```diff
@@ -132,18 +148,66 @@ where
...
+        self.replicas.insert(id, replica);
+
+        self.update_ingestion_scheduling(false);
+
+        self.replay_commands(id);
+    }
+
+    /// Replays commands to the specified replica.
+    pub fn replay_commands(&mut self, replica_id: ReplicaId) {
+        let commands = self.history.iter().cloned().collect::<Vec<_>>();
```
petrosagg, line 160: Haven't tried it but it feels like the `collect` and the `into_iter` below can be deleted.

src/storage-controller/src/instance.rs
```diff
@@ -231,38 +295,263 @@ where
...
+    ///
+    /// This does _not_ send commands to replicas, we only record the ingestion
+    /// in state and potentially update scheduling decisions.
+    fn absorb_ingestions(&mut self, ingestions: Vec<RunIngestionCommand>) {
+        for ingestion in ingestions {
+            let prefers_single_replica = ingestion
+                .description
+                .desc
+                .connection
+                .prefers_single_replica();
```
petrosagg, line 358: I see, so this method already exists there. Is there a reason we can't use this instead of de-normalizing it in the `ActiveIngestion` struct?

src/storage-controller/src/instance.rs
```diff
@@ -231,38 +295,263 @@ where
...
+    /// exists.
+    ///
+    /// This function searches through the command history to find the most
+    /// recent RunIngestionCommand for the specified ingestion ID and returns
+    /// its description.  Returns None if no ingestion with the given ID is
+    /// found.
+    pub fn get_ingestion_description(
+        &self,
+        id: &GlobalId,
+    ) -> Option<mz_storage_types::sources::IngestionDescription<CollectionMetadata>> {
```
petrosagg, line 489: nit: maybe import `IngestionDescription` at the top to avoid the long path

src/storage-controller/src/instance.rs
```diff
@@ -231,38 +295,263 @@ where
...
+            }
+        }
+    }
+
+    /// Returns the replicas that are actively running the given object (ingestion or export).
+    fn active_replicas(&mut self, id: &GlobalId) -> Box<dyn Iterator<Item = &mut Replica<T>> + '_> {
+        if let Some(ingestion_id) = self.ingestion_exports.get(id) {
+            // Right now, only ingestions can have per-replica scheduling
+            // decisions.
+            if let Some(ingestion) = self.active_ingestions.get(ingestion_id) {
```
petrosagg, line 523: nit: feels cleaner to make this a match instead of an if let with an early return to avoid falling through

src/storage-controller/src/instance.rs
```diff
@@ -231,38 +295,263 @@ where
...
+    /// Update scheduling decisions for ingestions, that is what replicas should
+    /// be running a given ingestion, if needed.
+    ///
+    /// A single-replica ingestion is scheduled on the last replica, if it is
+    /// not already running. The rationale being that the latest replica is the
+    /// one that was created to take over load from other replicas. Crucially,
+    /// we never change the scheduling decision for single-replica ingestions
+    /// unless we have to, that is unless the replica that they are running on
+    /// goes away. We do this, so that we don't send a mix of "run"/"allow
+    /// compaction"/"run" messages to replicas, which wouldn't deal well with
```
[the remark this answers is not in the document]
aljoscha: In that case replica A would receive a `RunIngestion` and then nothing more. From above:

> Crucially, we never change the scheduling decision for single-replica ingestions unless we have to, that is unless the replica that they are running on goes away. We do this, so that we don't send a mix of "run"/"allow compaction"/"run" messages to replicas

I can highlight it more that this is the more important property we uphold

src/storage-controller/src/instance.rs
aljoscha: It's a property of the `IngestionDescription`, and yeah, this is a cache of that, so that I don't have to re-derive it from the history. If you prefer that, I can also change it so we get it from the history instead. I don't think this is performance-sensitive code.

src/storage-controller/src/instance.rs
aljoscha: By "already exist", you mean it's added in this PR? 😅 

But yes, see my answer up here: https://github.com/MaterializeInc/materialize/pull/31890/files/9413c5366af22285e88e7c7ad279fcfcf1a20d65#r1999229940

bkirwi:
In general, the more similar the replicas of a particular cluster are, the happier I am. For the Kafka source, where we only filter at the end, things feel very similar and I am very happy. For these other sources, where we end up sending different command streams to different replicas, I am more nervous!

I think it probably doesn't make sense to revisit the design here, especially given the timing of everything, but I'd be interested in exploring a different approach for the sink. (eg. having the controller pick a leader, but only doing the filtering of the Kafka writes "at the last minute" in the sink.)

src/storage-controller/src/instance.rs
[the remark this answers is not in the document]
petrosagg: Does that mean that if I create a replica and then drop it my pg sources will never get scheduled again?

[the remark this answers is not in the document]
aljoscha: No, in that case it would be scheduled on another replica.

Here's the docstring, but slightly massaged to highlight the important property. I'll change the code to that.

```
    /// An important property of this scheduling algorithm is that we never
    /// change the scheduling decision for single-replica ingestions unless we
    /// have to, that is unless the replica that they are running on goes away.
    /// We do this, so that we don't send a mix of "run"/"allow
    /// compaction"/"run" messages to replicas, which wouldn't deal well with
    /// this. When we _do_ have to make a scheduling decision we schedule a
    /// single-replica ingestion on the last replica.
```

Does this help?

I want to change the last part (along with the implementation) to this, though:

```
    /// this. When we _do_ have to make a scheduling decision we schedule a
    /// single-replica ingestion on the first replica, according to the sort
    /// order of `ReplicaId`. We do this latter so that the scheduling decision
    /// is stable across restarts of `environmentd`/the controller.
```

aljoscha:
> In general, the more similar the replicas of a particular cluster are, the happier I am. For the Kafka source, where we only filter at the end, things feel very similar and I am very happy. For these other sources, where we end up sending different command streams to different replicas, I am more nervous!
> 
> I think it probably doesn't make sense to revisit the design here, especially given the timing of everything, but I'd be interested in exploring a different approach for the sink. (eg. having the controller pick a leader, but only doing the filtering of the Kafka writes "at the last minute" in the sink.)

Agreed on all those accounts! This design was chosen because it was determined to be quite hard for the sources to share their replication slot. I'd be very happy if we can come up with a design that allows us to run the same dataflows on all replicas. Also, though, this thing here is not at all a trap door decision, we can easily change things in future releases.

src/storage-controller/src/instance.rs
```diff
@@ -132,18 +148,66 @@ where
...
+                    cmds.retain(|cmd| self.is_active_replica(&cmd.id, &replica_id));
+                    StorageCommand::RunIngestions(cmds)
+                }
+                StorageCommand::AllowCompaction(mut cmds) => {
+                    cmds.retain(|cmd| self.is_active_replica(&cmd.0, &replica_id));
+                    StorageCommand::AllowCompaction(cmds)
+                }
+                command => command,
+            })
+            .collect::<Vec<_>>();
```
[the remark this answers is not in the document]
aljoscha: this one is getting around a borrowing problem, the iterator wants to access self and the call for getting the replica does as well

aljoscha:
@petrosagg I addressed your easy comments, and the question around scheduling is now resolved. Could you take another look? If you like, I'll also remove the cached `prefers_single_replica` field on `ActiveIngestion` and recompute that from the history when needed.

bkirwi:
(Just TBC - once Petros is satisfied I am also satisfied!)

petrosagg (approved):
Looks great! Thanks for explaining the scheduling thing in the call

PR #31912, opened 2025-03-17
storage/pg: emit an early probe during snapshotting

teskje:
This PR changes the Postgres source to emit an initial upstream probe immediately after the creation of the replication slot. The purpose of this is to produce the first timestamp binding (under RLU) as early as possible, so it can be used to reclock the snapshot updates rather than buffering them in the reclock operator.

Prior to this change, the first probe would only be emitted after the snapshot had been fully processed, which meant that that all snapshot updates needed to be buffered in the reclock operator, making the Postgres source require memory proportional to the size of the snapshot.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9061

### Checklist

### DIFF
--- a/src/storage/src/source/postgres/replication.rs
+++ b/src/storage/src/source/postgres/replication.rs
@@ -326,6 +326,18 @@ pub(crate) fn render<G: Scope<Timestamp = MzOffset>>(
             upper_cap_set.downgrade([&resume_lsn]);
             trace!(%id, "timely-{worker_id} replication reader started lsn={resume_lsn}");
 
+            // Emitting an initial probe before we start waiting for rewinds ensures that we will
+            // have a timestamp binding in the remap collection while the snapshot is processed.
+            // This is important because otherwise the snapshot updates would need to be buffered
+            // in the reclock operator, instead of being spilled to S3 in the persist sink.
+            //
+            // Note that we need to fetch the probe LSN _after_ having created the replication
+            // slot, to make sure the fetched LSN will be included in the replication stream.
+            let probe_ts = (config.now_fn)().into();
+            let max_lsn = super::fetch_max_lsn(&*metadata_client).await?;
+            let probe = Probe { probe_ts, upstream_frontier: Antichain::from_elem(max_lsn) };
+            probe_output.give(&probe_cap[0], probe);
+
             let mut rewinds = BTreeMap::new();
             while let Some(event) = rewind_input.next().await {
                 if let AsyncEvent::Data(_, data) = event {
--- a/test/bounded-memory/mzcompose.py
+++ b/test/bounded-memory/mzcompose.py
@@ -236,25 +236,25 @@ SCENARIOS = [
                     INSERT INTO t1 (f3) SELECT '{i}' || REPEAT('a', {PAD_LEN}) FROM generate_series(1, {REPEAT});
                     """
                 )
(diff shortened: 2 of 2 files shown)


test/bounded-memory/mzcompose.py
```diff
@@ -236,25 +236,25 @@ class KafkaScenario(Scenario):
...
             # We do not do DELETE post-restart, as it will cause OOM for clusterd
-            > SELECT * FROM v1; /* expect {ITERATIONS * REPEAT} */
-            {ITERATIONS * REPEAT}
+            > SELECT * FROM v1; /* expect {ITERATIONS * 10 * REPEAT} */
+            {ITERATIONS * 10 * REPEAT}
             """
         ),
         materialized_memory="4.5Gb",
-        clusterd_memory="3.5Gb",
+        clusterd_memory="1Gb",
```
teskje, line 257: Had to both bump the #iterations and reduce the memory limit to make the original issue reproducible. @def- lmk if you'd rather have this as a separate test. But we should have _some_ regression test.

src/storage/src/source/postgres/replication.rs
```diff
@@ -326,6 +326,18 @@ pub(crate) fn render<G: Scope<Timestamp = MzOffset>>(
...
+            // have a timestamp binding in the remap collection while the snapshot is processed.
+            // This is important because otherwise the snapshot updates would need to be buffered
+            // in the reclock operator, instead of being spilled to S3 in the persist sink.
+            let probe_ts = (config.now_fn)().into();
+            let max_lsn = super::fetch_max_lsn(&*metadata_client).await?;
+            // We can only produce the upstream table state accurately for LSN's >= the
+            // `resume_lsn`, so our first probe frontier must not be before `resume_lsn`.
+            let max_lsn = std::cmp::max(max_lsn, resume_lsn);
+            let probe = Probe { probe_ts, upstream_frontier: Antichain::from_elem(max_lsn) };
+            probe_output.give(&probe_cap[0], probe);
```
teskje, line 339: This seems to work, but I have some uncertainty about whether `max(max_lsn, resume_lsn)` is the correct `upstream_frontier` to use here. We have to use something >= `slot_metadata.confirmed_flush_lsn`, and we _may_ have to use something >= `resume_lsn`, not sure. Also fetching the `max_lsn` seems like the right thing to do, to respect the probe semantics, but I'm not sure how important it is.

petrosagg: I think using the max lsn unmodified is fine as long as it is fetched after the slot has been created (which it is, here). After creating a slot we can always recreate the state of the max LSN of the database. I don't think it can ever happen that `resume_lsn` is higher than a fresh `max_lsn`

petrosagg (approved):
Looks great! I think the code can be simplified by not caring about the `resume_lsn`. The only requirement is that we fetch the max lsn after having created the slot

src/storage/src/source/postgres/replication.rs
teskje: Put in a soft assert now instead.

test/bounded-memory/mzcompose.py
def-: Fine to have it in here I think.

def- (approved):

PR #31913, opened 2025-03-17
docs,etc: in ALTER CLUSTER docs, rename graceful -> zero downtime

aljoscha:
Resolves https://github.com/MaterializeInc/database-issues/issues/9080

We also go ahead and change mentions in code, including tests.

For the docs changes: @kay-kim, and @sjwiesman and @josharenberg so you know for field-eng. Would be good if one of you could stamp this. 🙏

### DIFF
--- a/doc/user/content/releases/previews.md
+++ b/doc/user/content/releases/previews.md
@@ -35,12 +35,14 @@ contact your Materialize support.
 {{</ note >}}
 
 
-### Graceful cluster resizing
+### Zero-downtime cluster resizing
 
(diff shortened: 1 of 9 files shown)


kay-kim (approved):

ParkMyCar (approved):

aljoscha:
Also, fyi @jubrad as the original implementor. In case you have any comments. 🕊️

aljoscha:
Also, I created the renamed flag in LD, and copied the rules from the current flag, so that we can smoothly transition.

PR #31918, opened 2025-03-17
parallel-workload: Reenable two checks in backup&restore

def-:
Since the associated issue has been closed

Seen in https://buildkite.com/materialize/nightly/builds/11474#0195a14d-2d4d-4344-a5e2-3ba10e2b647b

### Checklist

### DIFF
--- a/misc/python/materialize/parallel_workload/action.py
+++ b/misc/python/materialize/parallel_workload/action.py
@@ -1913,7 +1913,7 @@ class CreateMySqlSourceAction(Action):
         return result
 
     def run(self, exe: Executor) -> bool:
-        # TODO: Reenable when database-issues#6881 is fixed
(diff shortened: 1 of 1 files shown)


bkirwi:
Now that that issue is closed we don't expect PG sources to _panic_ after a restore, but we also don't really expect them to work normally - if a source has compacted its replication slot, restoring to a time before that will result in a stuck source. (This is AFAICT an irresolvable tension, not a bug we can fix... restoring a backup can restore MZ to a specific state, but not an external system that's managed separately.)

I don't know if the kind of assertions we want to run in parallel workload would succeed for a source that's been restored from backup. I think we probably want to leave sources and sinks disabled for tests that do backup/restore.

bkirwi (approved):
Thank you!

PR #31930, opened 2025-03-17
tests: Add newline handling to slt when `multiline` is not specified

ggevay:
The first commit changes the CODEOWNERS of `/src/sqllogictest` from `@MaterializeInc/adapter` to `@MaterializeInc/testing`. (But let me know if there was a good reason to have the Adapter team be the owner.) Edit: Also added myself, as discussed in DMs.

The second commit fixes a minor bug in space handling in slt in `colnames`.

The third commit is the main thing, which adds newline handling to slt result values, even when `multiline` is not specified. We use the replacement character trick to avoid newlines inside result values conflicting with newlines that are for field or row separation: when rewriting, we replace newlines with ⏎; when parsing, we replace ⏎ with newlines.

The motivation is that I'd like to have this feature for my next PR, which will make `SHOW CREATE ...` pretty-print its output, adding newlines. We have a gazillion tests that call `SHOW CREATE`, and it would be painful to change all of them to use `multiline`, especially that they have multi-column outputs, which, I think, is not really supported with `multiline`.

(I've added `lts-backport-v25.1`, because the `SHOW CREATE` PR will need to be backported, because `mzexplore` will rely on that, and `mzexplore` is part of the self-managed debug tool.)

### Motivation

### Checklist

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -121,7 +121,7 @@
 /src/sql/src/plan/lowering.rs       @MaterializeInc/cluster
 /src/sql-lexer                      @MaterializeInc/adapter
 /src/sql-parser                     @MaterializeInc/adapter
-/src/sqllogictest                   @MaterializeInc/adapter
+/src/sqllogictest                   @MaterializeInc/testing @ggevay
 /src/ssh-util                       @MaterializeInc/cluster
 /src/storage                        @MaterializeInc/cluster
 /src/storage-client                 @MaterializeInc/cluster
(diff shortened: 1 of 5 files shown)


def- (approved):
> The first commit changes the CODEOWNERS of /src/sqllogictest from @MaterializeInc/adapter to @MaterializeInc/testing. (But let me know if there was a good reason to have the Adapter team be the owner.)

Generally we haven't written many SLT tests in the testing team, and more focused on higher level frameworks. It still makes sense to review test changes though.

PR #31933, opened 2025-03-18
Pretty-print in `SHOW CREATE`

ggevay:
This makes `SHOW CREATE` and `SHOW REDACTED CREATE` pretty-print the result (e.g., add line breaks). (https://github.com/MaterializeInc/database-issues/issues/9078, and [slack discussion](https://materializeinc.slack.com/archives/C063H5S7NKE/p1742066518991299?thread_ts=1742058798.775019&cid=C063H5S7NKE))

The first commit just does a renaming, and then the next commit is the main thing. I recommend reviewing commit-by-commit, and starting the review of the main commit from `humanize_sql_for_show_create`. The diff looks somewhat big, but most of the code diff is just threading `FormatMode` through the `sql-pretty` crate, which was needed to enable pretty-printing for both the normal `SHOW CREATE` and for `SHOW REDACTED CREATE`. Also, there was a staggering amount of manual test rewrites needed, because Testdrive doesn't have auto-rewriting. (I also did some "spring cleaning": deleted some old tests, which were mirrors of other tests but with an old Kafka syntax, as discussed [here](https://materializeinc.slack.com/archives/C01LKF361MZ/p1741873647311339?thread_ts=1741872074.245179&cid=C01LKF361MZ).)

Note that in addition to pretty-printing, this also changes the output format of `SHOW CREATE` from `FormatMode::Stable` to `FormatMode::Simple`. The main effect of this is less quoting of identifiers: stable mode quotes _all_ identifiers, thus cluttering up the screen quite a bit, while the simple mode quotes only when it's needed. I have made some efforts recently to get the "when it's needed" logic bug-free, so hopefully the simple mode is enough.

### Motivation

  * This PR adds a known-desirable feature: https://github.com/MaterializeInc/database-issues/issues/9078, and [slack discussion](https://materializeinc.slack.com/archives/C063H5S7NKE/p1742066518991299?thread_ts=1742058798.775019&cid=C063H5S7NKE)

### Checklist


misc/python/materialize/checks/all_checks/create_table.py
```diff
@@ -60,9 +60,6 @@ def validate(self) -> Testdrive:
                 > SELECT * FROM create_table_view2;
                 2
 
-                > SHOW CREATE TABLE create_table1;
-                materialize.public.create_table1 "CREATE TABLE \\"materialize\\".\\"public\\".\\"create_table1\\" (\\"f1\\" \\"pg_catalog\\".\\"int4\\", \\"f2\\" \\"pg_catalog\\".\\"int4\\" NOT NULL DEFAULT 1234)"
```
teskje, line 64: There are a bunch of deleted SHOW CREATE statements here and in other test files. Are those not useful anymore?

src/sql-pretty/src/util.rs
```diff
@@ -33,7 +33,7 @@ where
 
 pub(crate) fn title_comma_separate<'a, F, T, S>(title: S, f: F, v: &'a [T]) -> RcDoc<'a, ()>
 where
-    F: Fn(&'a T) -> RcDoc<'a>,
+    F: FnMut(&'a T) -> RcDoc<'a>,
```
ParkMyCar, line 36: I'm curious what motivated these changes from `Fn` to `FnMut`?

src/sql-pretty/src/lib.rs
```diff
@@ -22,39 +23,69 @@ use crate::doc::{
...
+pub struct PrettyConfig {
+    pub width: usize,
+    pub format_mode: FormatMode,
+}
+
 /// Pretty prints a statement at a width.
-pub fn to_pretty<T: AstInfo>(stmt: &Statement<T>, width: usize) -> String {
-    format!("{};", to_doc(stmt).pretty(width))
+pub fn to_pretty<T: AstInfo>(stmt: &Statement<T>, config: PrettyConfig) -> String {
+    format!("{};", to_doc(stmt, config).pretty(config.width))
```
ParkMyCar, line 51: Threading a `PrettyConfig` into the `doc*` functions doesn't seem right?

test/legacy-upgrade/check-from-v0.111.0-kafka-sink.td
```diff
@@ -7,17 +7,14 @@
 # the Business Source License, use of this software will be governed
 # by the Apache License, Version 2.0.
 
-$ set-from-sql var=expected-compression-implicit-create-sql
-SELECT sql FROM expected_compression_implicit_create_sql
-
 > SHOW CREATE SINK compression_implicit;
-"materialize.public.compression_implicit" ${expected-compression-implicit-create-sql}
+materialize.public.compression_implicit "CREATE SINK materialize.public.compression_implicit IN CLUSTER quickstart FROM materialize.public.kafka_sink_from INTO KAFKA CONNECTION materialize.public.kafka_conn (TOPIC = 'kafka-sink') FORMAT JSON ENVELOPE DEBEZIUM;"
```
teskje, line 11: Dumb question, but why do some of the SHOW CREATE outputs have newlines and some don't?

teskje, line 11: Also the pretty-printing adds a semicolon where there previously was none. Is this a compatibility concern? I think it would be if people would somehow run SHOW CREATE in their scripts, and then expect output without a trailing semicolon. Not sure why they would do either though.

src/sql/src/plan/statement/show.rs
```diff
@@ -1182,9 +1183,15 @@ fn humanize_sql_for_show_create(
...
+        &resolved,
+        PrettyConfig {
+            width: mz_sql_pretty::DEFAULT_WIDTH,
+            format_mode: if redacted {
+                FormatMode::SimpleRedacted
+            } else {
+                FormatMode::Simple
+            },
+        },
+    ))
```
ParkMyCar, line 1196: It seems like a large part of this change is because we need to plumb the `FormatMode` into the `sql-pretty` crate?

IMO pretty printing shouldn't need to know about whether or not something is redacted. What do you think about something like:

```
let raw_str = if redacted {
  resolved.to_ast_string_redacted()
} else {
  resolved.to_ast_string_stable()
};

Ok(mz_sql_pretty::pretty_str(&redacted))
```

Unfortunate that we need to format the string twice, but given this is only for `SHOW CREATE` I don't feel too bad about it? In a future world it feels like there is an API we could introduce for the `mz-sql-pretty` crate that would allow us to wrap a statement in some context which would automatically format values as redacted if necessary.

teskje (approved):
LGTM! Not sure about the removed SHOW CREATE statements in some of the test files, but I trust they all have good reasons.

The `config` parameter added to ~all functions in `mz-sql-pretty` makes me think that we should turn them into methods on a struct that holds the config. But I can understand that you don't want to deal with such a large refactor now.

src/sql/src/plan/statement/show.rs
teskje: > IMO pretty printing shouldn't need to know about whether or not something is redacted.

This didn't occur to me during the review, but I agree!

test/legacy-upgrade/check-from-v0.111.0-kafka-sink.td
ggevay: `sql-pretty` doesn't support every statement kind. E.g., `to_doc` [has a case](https://github.com/MaterializeInc/materialize/blob/bf9abe7fd897f8c45324d951e0caf16ca395f462/src/sql-pretty/src/lib.rs#L38) for `CreateSource`, but not for `CreateSink`.

ggevay: Well, I'd say we should just risk this.

It's probably slightly better to have a semicolon than to not have, because if you want to paste these statements into some SQL shell, then it's slightly easier if they already have a semicolon.

Btw. `mzexplore` might actually run into this problem. For example, its cluster cloning functionality might use `SHOW CREATE` in this way. I'm working on `mzexplore`, so I'll fix these things in follow-up PRs.

src/sql/src/plan/statement/show.rs
ggevay: There is the issue that we might not be able to parse a redacted statement back. This is because maybe the parser expects a number somewhere, but then it gets something like `<redacted>`. https://github.com/MaterializeInc/database-issues/issues/8796 aims to solve this problem, but there might be a long tail of cases to solve there, so in the meantime we'd have to have some error handling after the `pretty_str` call, and just print the non-pretty result. This is doable, but maybe it tips the balance in favor of the PR's current approach. What do you think?

ggevay: One more consideration:

There are various aspects of AST printing that we want to control (see https://github.com/MaterializeInc/database-issues/issues/9082). If we want to keep `mz-sql-pretty` oblivious to all of them, and just use Parker's trick for adding redaction on top of any of the other AST printing options, then we run the risk that `mz-sql-pretty` undoes some of the other formatting options when it calls `AstDisplay` as its "base case" with its hardwired `FormatMode`.

For example, imagine that we'd like to pretty-print in `FormatMode::Stable`. The above trick can't be adopted for this: if we first print with `FormatMode::Stable`, then parse back, then run `mz-sql-pretty`, then the problem is that  `mz-sql-pretty` has `FormatMode::Simple` hardwired into its own calls of `AstDisplay` (which it does when it can't or doesn't want to deal with further chunking up an AST fragment), so it undoes the earlier `FormatMode::Stable` and just prints in `FormatMode::Simple`.

This wouldn't be a concern for this particular PR, but I think in the future it would be great to make all formatting options orthogonal (https://github.com/MaterializeInc/database-issues/issues/9082), for which it seems to me that we'd have to wire `FormatMode` through `mz-sql-pretty`, to avoid `mz-sql-pretty` undoing some formatting option.

src/sql-parser/src/ast/display.rs
```diff
@@ -154,23 +154,21 @@ pub trait AstDisplay {
     where
         W: fmt::Write;
 
-    fn to_ast_string(&self) -> String {
-        let mut buf = String::new();
-        let mut f = AstFormatter::new(&mut buf, FormatMode::Simple);
-        self.fmt(&mut f);
-        buf
+    fn to_ast_string_simple(&self) -> String {
+        self.to_ast_string(FormatMode::Simple)
```
antiguru, line 158: I think to reduce noise, it'd make sense not to rename this function.

src/sql-pretty/src/doc.rs
```diff
@@ -17,26 +17,39 @@ use crate::util::{
...
+fn doc_display_pass_mapper<T: AstDisplay>(
+    config: PrettyConfig,
+) -> impl for<'b> FnMut(&'b T) -> RcDoc<'b, ()> {
+    move |v| doc_display_pass(v, config)
+}
+
+pub(crate) fn doc_create_source<T: AstInfo>(
+    v: &CreateSourceStatement<T>,
+    config: PrettyConfig,
+) -> RcDoc {
```
antiguru, line 52: It seems the PR needs to introduce a bunch of complexity to work around the existing code structure. It strikes me as potentially the wrong approach: Why don't we convert the freestanding functions to functions on a type instead? That way we wouldn't need to pass the config everywhere.

src/sql/src/func.rs
```diff
@@ -3838,7 +3838,7 @@ pub static MZ_CATALOG_BUILTINS: LazyLock<BTreeMap<&'static str, Func>> = LazyLoc
         "pretty_sql" => Scalar {
             params!(String, Int32) => BinaryFunc::PrettySql => String, oid::FUNC_PRETTY_SQL;
             params!(String) => Operation::unary(|_ecx, s| {
-                let width = HirScalarExpr::literal(Datum::Int32(100), ScalarType::Int32);
+                let width = HirScalarExpr::literal(Datum::Int32(mz_sql_pretty::DEFAULT_WIDTH.try_into().expect("must fit")), ScalarType::Int32);
```
antiguru, line 3841: You could make the constant a 32-bit number, which should be plenty enough. Then you can do cheap up conversion and don't have to unwrap here.

antiguru (approved):
Left some comments inline.

I'll approve, but I think there could be more work done to improve the PR.

src/sql-pretty/src/doc.rs
ggevay: Yeah, I'll probably do this. I don't think it will really reduce the complexity, because the `self` parameter will just take the place of the current `config` parameter, but should help readability a bit.

(But first, I'd like to resolve [the question of whether we even need to plumb `FormatMode` through the `sql-pretty` crate](https://github.com/MaterializeInc/materialize/pull/31933#discussion_r2017283330). If not, then most of the changes to `sql-pretty` can simply be reverted.)

src/sql-parser/src/ast/display.rs
ggevay: I renamed it after finding myself jumping to the definition repeatedly to see what `FormatMode` it uses. Now it's clear from the name.

Note that the renaming is separated into its own commit (as mentioned in the PR description). This way it doesn't really add noise when reviewing: One can look at the commits individually, and all the diff that is in the renaming commit doesn't need reviewing, just the name itself.

src/sql/src/func.rs
ggevay: Good idea, will do!

ggevay: Actually, it's not so simple, because then the other use of `DEFAULT_WIDTH`, which gives it to `PrettyConfig::width`, would need to convert from i32 to usize. I could make `PrettyConfig::width` also an i32, but all these widths being signed types would look weird.

I think the original problem is that the `pretty_sql` scalar function takes a signed width. We could also change that, but changing the parameter types of an existing scalar function is probably more hassle than this is worth.

So, I'd like to just stay with these widths being unsigned types (as they should be), and just work around the problem that `pretty_sql` takes an `Int32` by doing a conversion there.

src/sql-pretty/src/lib.rs
ggevay: See discussion [here](https://github.com/MaterializeInc/materialize/pull/31933#discussion_r2017283330).

ggevay:
The "Refactor sql-pretty: pass around PrettyConfig as &self." commit did the refactoring that @teskje and @antiguru suggested.

The only remaining open question is whether we even need to pass around the config. I think we do, see [here](https://github.com/MaterializeInc/materialize/pull/31933#discussion_r2017450031).

src/sql/src/plan/statement/show.rs
ParkMyCar: Thanks for thinking through this Gabor!

ParkMyCar (approved):

def-:
@ggevay This is very messy to backport, is it really required?

ggevay:
(We can live without it, as discussed on Slack.)

PR #31938, opened 2025-03-18
persist-client: avoid short-lived `FuturesUnordered`

petrosagg:
This PR is fixing a subtle performance problem that arises from short-lived instances of the `FuturesUnordered` type. The high CPU load was observed by @antiguru when instatiating a very large number of sources.

When a future registered with a `FuturesUnordered` instance is woken it enters a ready queue. Then, when the `FuturesUnordered::poll_next` method is called each future from the ready queue is popped and polled. Ideally one of the futures resolves and `poll_next` returns with `Poll::Ready(_)`. If no future is ready the entire `FuturesUnordered` returns `Poll::Pending`.

One thing that can happen, especially if many futures are registered, is that while polling a future in the ready queue another previously polled future gets woken, re-entering the ready queue. When a large number of futures is managed by `FuturesUnordered` it is possible that futures enter the ready queue faster than they can be polled, starving the executor.

To defend against this degenerate behavior `FuturesUnordered` purpusefully calls its waker and immediately returns `Poll::Pending` after having polled all the futures it manages. This allows the executor to schedule other tasks, including the task containing the `FuturesUnordered` instance.

This is all well and good in the steady state, but it falls apart the *first* time a `FuturesUnordered` is polled. On construction, all registered futures are forcefully put into the ready queue so that they get polled at least once and then the waker mechanism can take over. However, the code that tracks how many polls have happened does not differentiate between an initial poll and subsequent polls, so just the *first* time around the defending mechanism will kick in and cause a spurious wake.

You can see a demo of this happening here
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1c832e329b746b49d877a8b9ec7ff5a5

The code in `machine.rs` would construct a temporary `FuturesUnordered` that would always fall into the bad interaction described above, since it's always a brand new instance. The large amount of spurious wake ups can amplify the executor overhead since it is now doing a lot more work swapping tasks. When the executor is timely, this can really hurt performance.

To fix the problem we can simply drive the two futures with a select.

Separately, I think the behavior of `FuturesUnordered` can be improved by taking into acount that a certain number of futures have forcefully entered the ready queue.

### Checklist

### DIFF
--- a/src/persist-client/src/internal/machine.rs
+++ b/src/persist-client/src/internal/machine.rs
@@ -17,7 +17,7 @@ use std::time::{Duration, Instant, SystemTime};
 
 use differential_dataflow::difference::Semigroup;
 use differential_dataflow::lattice::Lattice;
-use futures::future::BoxFuture;
+use futures::future::{self, BoxFuture};
 use futures::stream::{FuturesUnordered, StreamExt};
 use futures::FutureExt;
 use mz_dyncfg::{Config, ConfigSet};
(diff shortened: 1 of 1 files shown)


antiguru (approved):
Looks great; thank you!

bkirwi:
Ah, nice find! Clear that a select is a better solution here in any case.

A while ago we made some effort to try not to call Persist API methods directly on timely threads, since when Timely was saturated it would cause very odd-looking Persist metrics as futures failed to be promptly polled. It sounds like that effort was incomplete? Worth looking into when we have some time.

PR #31949, opened 2025-03-19
storage: avoid rendering fat timely operator

petrosagg:
We used to have a usecase for rendering a timely operator with many inputs and many outputs connected in a complicated way but after #31260 this usecase no longer exists.

This PR removes the fat operator in favor of many smaller operators with simple connectivity.

### Checklist

### DIFF
--- a/src/storage/src/source/source_reader_pipeline.rs
+++ b/src/storage/src/source/source_reader_pipeline.rs
@@ -362,7 +362,7 @@ where
             let mut health_cap = Some(caps.remove(0));
 
             move |frontiers| {
-                let mut statuses = vec![];
+                let mut last_status = None;
                 let mut health_output = health_output.activate();
 
                 if frontiers[0].is_empty() {
@@ -392,8 +392,9 @@ where
                             namespace: C::STATUS_NAMESPACE.clone(),
                             update: status,
                         };
-                        if statuses.last() != Some(&status) {
-                            statuses.push(status);
+                        if last_status.as_ref() != Some(&status) {
+                            last_status = Some(status.clone());
+                            health_output.session(&health_cap).give(status);
                         }
 
                         match message {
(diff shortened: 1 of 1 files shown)


src/storage/src/source/source_reader_pipeline.rs
```diff
@@ -447,7 +437,7 @@ where
     (
         export_collections,
         progress,
-        health.concat(&derived_health),
+        health.concatenate(health_streams),
```
antiguru, line 440: Correct me if I'm wrong, but I think this changes the progress tracking for the health output: previously, it was disconnected, but now the progress information from all sources flows into it. Looking through the code, we might not actually care, in the past the health cap was either 0 or dropped?

src/storage/src/source/source_reader_pipeline.rs
```diff
@@ -419,17 +410,16 @@ where
...
-                        if statuses.is_empty() {
-                            continue;
-                        }
-                        health_output.session(&health_cap).give_container(statuses);
-                        statuses.clear()
+                    if !statuses.is_empty() {
+                        health_output
+                            .session(&health_cap)
+                            .give_container(&mut statuses);
+                        statuses.clear();
```
antiguru, line 417: You could just write the status directly I think, without going through the `statuses` vector. That way you'd actually construct even-sized containers provided by the Timely infrastructure.

antiguru (approved):
Looks great, I left some comments inline. Thank you!

src/storage/src/source/source_reader_pipeline.rs
petrosagg: The health stream is a bit of a weirdo and its progress events aren't used in anything. It communicates "writes that need to happen" which eventually the controllers write out to the system table at whatever timestamp it happens to be at. So I opted for the simpler API that doesn't need me to construct connection vectors

src/storage/src/source/source_reader_pipeline.rs
petrosagg: Good point, fixed in the top commit of the PR

PR #31952, opened 2025-03-19
Fix the test_retain_history test

DAlperin:
This was probably always racy based on timestamp selection and compaction. This removes the racy parts and will be followed up with a PR to actually plan queries correctly with an index over a history retained view.

Closes https://github.com/MaterializeInc/database-issues/issues/7450

### Checklist

### DIFF
--- a/src/environmentd/tests/sql.rs
+++ b/src/environmentd/tests/sql.rs
@@ -3271,7 +3271,6 @@ async fn test_explain_as_of() {
 
 // Test that RETAIN HISTORY results in the since and upper being separated by the specified amount.
 #[mz_ore::test(tokio::test(flavor = "multi_thread", worker_threads = 1))]
-#[ignore] // TODO: Reenable when database-issues#7450 is fixed
 async fn test_retain_history() {
     let server = test_util::TestHarness::default().start().await;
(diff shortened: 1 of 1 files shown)


src/environmentd/tests/sql.rs
```diff
@@ -3366,34 +3365,29 @@ async fn test_retain_history() {
...
+        // We should be able to query the views now and in the past.
         let ts = get_explain_timestamp(name, &client).await;
-        assert_contains!(
-            client
-                .query(&format!("SELECT * FROM {name} AS OF {ts}-2000"), &[])
-                .await
-                .unwrap_err()
-                .to_string(),
-            "not valid for all inputs"
-        );
```
ParkMyCar, line 3384: While this behavior is not optimal, it still seems worthwhile to assert? This way were exercising known behavior?

It's fine to leave as-is, but I'm curious if this assert was failing?

ParkMyCar (approved):

src/environmentd/tests/sql.rs
DAlperin: This assert was designed to fail due to landing on the index. But we could add a similar assertion to verify that times totally out of bounds of the collection fail.

PR #31967, opened 2025-03-20
Set console's pod security context fs_group to 101

alex-hunt-materialize:
Set console's pod security context fs_group to 101, as we use an nginx-unprivileged container that runs as uid/gid 101, and our TLS certificate mount was previously unreadable by that user.

### Motivation

TLS doesn't work with the console ever since we stopped running as root.

### Checklist

### DIFF
--- a/src/orchestratord/src/controller/materialize/console.rs
+++ b/src/orchestratord/src/controller/materialize/console.rs
@@ -11,9 +11,9 @@ use k8s_openapi::{
     api::{
         apps::v1::{Deployment, DeploymentSpec},
         core::v1::{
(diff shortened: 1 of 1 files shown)


alex-hunt-materialize:
@def- This doesn't currently have test coverage, but it blocks the terraform certificate support PR. I think it will be easier to wait and add TLS to our AWS terraform tests after that gets merged.

def- (approved):
Fine by me for now, we should add tests later.

> Alternatively, do we have a place we're already testing the helm chart directly that I can add TLS to?

Currently not.

PR #31975, opened 2025-03-20
make halt! use an exit code of 166

doy-materialize:
### Motivation

panic! already uses an exit code of 2, and it would be useful for us to be able to differentiate between these cases (because halt! is used for situations which aren't actually real errors, they should just trigger retries)

### Checklist

### DIFF
--- a/src/ore/src/process.rs
+++ b/src/ore/src/process.rs
@@ -18,7 +18,7 @@
 /// Halts the process.
 ///
 /// `halt` forwards the provided arguments to the [`tracing::warn`] macro, then
-/// terminates the process with exit code 2.
+/// terminates the process with exit code 166.
 ///
 /// Halting a process is a middle ground between a graceful shutdown and a
 /// panic. Use halts for errors that are severe enough to require shutting down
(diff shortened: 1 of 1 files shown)


src/ore/src/process.rs
```diff
@@ -42,7 +42,7 @@
 macro_rules! halt {
     ($($arg:expr),* $(,)?) => {{
         $crate::__private::tracing::warn!("halting process: {}", format!($($arg),*));
-        $crate::process::exit_thread_safe(2);
+        $crate::process::exit_thread_safe(3);
```
alex-hunt-materialize, line 45:
```suggestion
        $crate::process::exit_thread_safe(166);
```

Can we use something in the 166-199 range?
Discussion https://materializeinc.slack.com/archives/CTESPM7FU/p1743687737423139?thread_ts=1743684376.799089&cid=CTESPM7FU

doy-materialize:
to clarify, it's not panic that uses an exit code of 2 (panic normally exits with 101, but we override panic to abort which causes an exit of 134), it's clap that returns an exit code of 2 on argument parsing failures.

alex-hunt-materialize (approved):

PR #31976, opened 2025-03-20
Scrape retained catalog data as CSVs

SangJunBak:

### Motivation

### Checklist

We were running into this error where if we tried to retry, because the transaction was hoisted, we couldn't rollback without moving ownership of it. By cancelling using the client, we avoid having to hoist the transaction.

### DIFF
--- a/src/self-managed-debug/src/system_catalog_dumper.rs
+++ b/src/self-managed-debug/src/system_catalog_dumper.rs
@@ -871,12 +871,6 @@ impl<'n> SystemCatalogDumper<'n> {
 
         let relation_name = relation.name.to_string();
 
-        let mut pg_client_lock = pg_client.lock().await;
-        // TODO (debug_tool3): Use a transaction for the entire dump instead of per query.
-        let transaction = pg_client_lock.transaction().await?;
-
-        let cancel_token = transaction.cancel_token();
-
         if let Err(err) = retry::Retry::default()
             .max_duration(PG_QUERY_TIMEOUT)
             .initial_backoff(Duration::from_secs(2))
@@ -884,10 +878,13 @@ impl<'n> SystemCatalogDumper<'n> {
                 let start_time = start_time.clone();
                 let relation_name = relation.name;
                 let cluster_replica = cluster_replica.clone();
-                let transaction = &transaction;
 
                 async move {
-                    match query_relation(transaction, start_time, &relation, cluster_replica).await
+                    // TODO (debug_tool3): Use a transaction for the entire dump instead of per query.
+                    let mut pg_client_lock = pg_client.lock().await;
+                    let transaction = pg_client_lock.transaction().await?;
+
+                    match query_relation(&transaction, start_time, &relation, cluster_replica).await
                     {
                         Ok(()) => Ok(()),
                         Err(err) => {
@@ -903,6 +900,10 @@ impl<'n> SystemCatalogDumper<'n> {
             })
             .await
         {
+            let pg_client_lock = pg_client.lock().await;
+
+            let cancel_token = pg_client_lock.cancel_token();
+
             if let Err(_) = async {
                 let tls = self.pg_tls.clone();
 
--- a/src/self-managed-debug/src/system_catalog_dumper.rs
+++ b/src/self-managed-debug/src/system_catalog_dumper.rs
@@ -889,7 +889,7 @@ impl<'n> SystemCatalogDumper<'n> {
                         Ok(()) => Ok(()),
                         Err(err) => {
                             error!(
-                                "{}: {}. Retrying...",
+                                "{}: {:#}. Retrying...",
                                 format_catalog_dump_error_message(relation_name, cluster_replica),
                                 err
                             );
@@ -991,7 +991,7 @@ impl<'n> SystemCatalogDumper<'n> {
                 };
 
                 error!(
-                    "{}: {}.\nConsider increasing the size of the cluster {}",
+                    "{}: {:#}.\nConsider increasing the size of the cluster {}",
                     format_catalog_dump_error_message(relation.name, replica),
                     err,
                     docs_link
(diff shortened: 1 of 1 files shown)


src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -681,41 +693,110 @@ pub async fn create_postgres_connection(
...
+    relation_name: &str,
+) -> Result<(), anyhow::Error> {
+    let copy_stream = transaction
+        .copy_out(copy_query)
+        .await
+        .context(format!("Failed to COPY TO for {}", relation_name))?
+        .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e));
+    let copy_stream = std::pin::pin!(copy_stream);
+    let mut reader = StreamReader::new(copy_stream);
+    tokio::io::copy(&mut reader, file).await?;
```
ParkMyCar, line 709: Call `file.sync_all().await` after copying the data

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -681,41 +693,110 @@ pub async fn create_postgres_connection(
...
+            );
+
+            let copy_fut = write_copy_stream(transaction, &copy_query, &mut file, relation.name);
+
+            // We use a timeout to cut the SUBSCRIBE query short since it's expected to run indefinitely.
+            // Alternatively, we could use a `DECLARE...FETCH ALL` for the same effect, but then we'd have
+            // to format the result as CSV ourselves, leading to more code. Another alternative is to
+            // specify an UPTO, but it gets finicky to get the UPTO frontier right since we can't rely on
+            // wallclock time.
+            let res = timeout(SUBSCRIBE_SCRAPE_TIMEOUT, copy_fut).await;
```
ParkMyCar, line 739: Why are we switching from a `SELECT` to a `SUBSCRIBE`? Especially if we're using a timeout to cancel a `SUBSCRIBE` I don't think there are any guarantees about what we'll actually receive?

src/self-managed-debug/src/system_catalog_dumper.rs
```diff
@@ -681,41 +693,110 @@ pub async fn create_postgres_connection(
...
+    Ok::<(), anyhow::Error>(())
+}
+
+pub async fn query_column_names(
+    pg_client: &PgClient,
+    relation: &Relation,
+) -> Result<Vec<String>, anyhow::Error> {
+    let relation_name = relation.name;
+    // We query the column names to write the header row of the CSV file.
+    // TODO (SangJunBak): Use `WITH (HEADER TRUE)` once database-issues#2846 is implemented.
```
ParkMyCar, line 770: I think this comment is stale?

ParkMyCar (commented):
Only comment is I'm not really sure why we want to use a `SUBSCRIBE` over a `SELECT`?

src/self-managed-debug/src/system_catalog_dumper.rs
SangJunBak: To get the retained history, including the snapshot! I don't think there's any other way of getting retained data, but I might be wrong?

> I don't think there are any guarantees about what we'll actually receive?

So if it's empty, I think that's fine! But I think you're right that there's no guarantee that we'll get the full result set using the method. My reasoning though is it isn't that much worse than using `DECLARE ... FETCH ALL` given we'd need to supply a timeout to `FETCH ALL` anyways. I think `FETCH ALL` does have stronger guarantees that, if the fetch doesn't timeout, will terminate once ALL the rows at the time of `FETCH` is received. But I didn't think that guarantee was worth the tradeoff of parsing the result stream into a CSV ourselves. But curious on your take!

ParkMyCar: I think it's definitely worth it to use `FETCH ALL`, because otherwise reasoning about what data we get is pretty difficult. Also tuning the `SUBSCRIBE_SCRAPE_TIMEOUT` would be hard because we'd need to account for network lag, data set sizes, etc.

If you'd rather merge as-is and follow up, that works for me, but I would really like to see this change! Happy to help write the logic to transform the result stream into CSVs :)

ParkMyCar (approved):

src/self-managed-debug/src/system_catalog_dumper.rs
SangJunBak: > Also tuning the SUBSCRIBE_SCRAPE_TIMEOUT would be hard because we'd need to account for network lag, data set sizes, etc.

But wouldn't we still need the timeout though when using `FETCH ALL`?

> because otherwise reasoning about what data we get is pretty difficult

I agree with you here! I'll push up a change to this PR~

ParkMyCar: > But wouldn't we still need the timeout though when using `FETCH ALL`?

What I'm thinking of is `FETCH ALL FROM <cursor> WITH (timeout = '0s');` I haven't tested it but I believe this should return a snapshot and all of the retained history that is immediately available. In other words it returns everything from `now() - retained history` up until `now()`, which is the goal?

We'll still need to wrap the query in a timeout in case the connection or database is slow, but I don't think we'll ever receive a partial set of data. For example, if we timeout the `SUBSCRIBE` we could receive the initial snapshot and then some indeterminate amount of diffs.

I might be wrong here though! Please feel free to correct me :)

SangJunBak: > it returns everything from now() - retained history up until now()

Ah that's a really good point. My logic is because it's streaming in sequentially (ordered by `mz_timestamp`), even if results are indeterminate, it'll still at least be correct / in order. But having the guarantee that we'll never get the snapshot cut off is definitely important.

SangJunBak: This commit should fix it! [`b8d117a` (#31976)](https://github.com/MaterializeInc/materialize/pull/31976/commits/b8d117aa50f9b5279c408f5dd18206b6ad8d6069)

PR #31979, opened 2025-03-21
optimizer: Make `EQProp` call `MSE::reduce` only after a change in `reduce_expr`

ggevay:
[We have a CPU profile](https://pprof.me/631f3c6fd71e6b13baf0e7b2c1505b7e/) where [a user](https://github.com/MaterializeInc/accounts/issues/3) running some very complex ad-hoc queries made envd quite busy with optimization. ([Slack discussion](https://materializeinc.slack.com/archives/C03R1QZ101Z/p1742565209009279?thread_ts=1742550520.344209&cid=C03R1QZ101Z).) The profile is showing that most of the time is spent in `MirScalarExpr::reduce` calls from `EquivalencePropagation`.

This PR modifies `EquivalencePropagation` to call `reduce` only if there was a recent change in the expression. Specifically, all of EQProp's calls to `reduce` are just after a `reduce_expr`, which conveniently returns a bool that indicates whether it made any changes.

There are no EXPLAIN changes across all of our slts. I've also done a run on all slts where I added a temporary assert that the `reduce`s that we are skipping wouldn't do anything. This failed only one time across all of our slts: in `joins.slt:1275`. But looking at the EXPLAINs of this query on `main` vs. the PR, the final plan does not change even for this query.

Added a feature flag, enabled by default.

The first commit is the actual functionality, the second just adds the feature flag and wires it up.

### Motivation

  * This PR speeds up the optimizer.

### Checklist

### DIFF
--- a/src/repr/src/optimize.rs
+++ b/src/repr/src/optimize.rs
@@ -124,6 +124,8 @@ optimizer_feature_flags!({
     enable_projection_pushdown_after_relation_cse: bool,
     // See the feature flag of the same name.
     enable_let_prefix_extraction: bool,
+    // See the feature flag of the same name.
+    enable_less_reduce_in_eqprop: bool,
 });
 
 /// A trait used to implement layered config construction.
--- a/src/sql/src/plan/statement/ddl.rs
+++ b/src/sql/src/plan/statement/ddl.rs
@@ -4654,7 +4654,9 @@ pub fn unplan_create_cluster(
                 enable_join_prioritize_arranged,
                 enable_projection_pushdown_after_relation_cse,
                 enable_let_prefix_extraction: _,
+                enable_less_reduce_in_eqprop: _,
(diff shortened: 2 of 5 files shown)


frankmcsherry (approved):
I can't really tell what this does to the overall behavior, but I think `MSE::reduce` doesn't have specific behavior in any case (e.g. not necessarily idempotent; may or may not perform various simplifications) so it should be equally correct w/o calling it, and the behavior of the whole thing can't rely on it being called or not. E.g. no clue whether we produce the same plans but have to run EQProp more often to tickle reductions we miss for some reason.

ggevay:
My thinking is that we do call `reduce` in the same fixpoint loops with EQProp, as well as shortly before those fixpoint loops. So, when EQProp itself is not making any changes, then these other calls should be enough.

PR #31982, opened 2025-03-21
cluster: make create_sockets work without an external epoch

teskje:
The current `create_sockets` logic relies on an external epoch that is strictly increasing between replica connections. [As it turns out](https://materializeinc.slack.com/archives/C08ACQNGSQK/p1739960920798069) the controller is not able to guarantee a strictly increasing epoch in read-only mode, making `create_sockets` unsound. Additionally, we would like to move Timely cluster initialization into the clusterd startup sequence, where no external epoch value is available at all.

This PR adds a new `create_sockets` implementation that works without an external epoch. The new version is similar to the old one in that it prevents processes from different generations from connecting to each other. It achieves that using a pseudo epoch based on  the system time and a more sequential connection protocol that uses process 0 as the generation leader.

The old version is still available by setting the feature flag `enable_create_sockets_v2 = false`. The expectation is that we can remove it once the new version has baked in production for a bit.

To derisk the change, this PR also adds a [turmoil](https://github.com/tokio-rs/turmoil)-based chaos test for `create_sockets`, which can be used for fuzzing. To support this, `mz_ore::netio::socket` is extended to also support `turmoil`'s socket wrapper types.

### Motivation

`create_sockets` relies on a strictly increasing epoch, but in read-only mode none can be provided.

### Tips for reviewer

Commits are meaningful and I recommend reviewing them separately.

This was initially part of the PR that moves Timely cluster initialization into clusterd startup (https://github.com/MaterializeInc/materialize/pull/31710). I pulled it out to reduce the scope of that PR and because it makes sense to discuss both changes separately.

### Checklist


teskje:
Looks like this is ready to review. [Nightlies](https://buildkite.com/materialize/nightly/builds/11803) are green, except for one timeout that also happens on main.

src/cluster/src/communication_v2.rs
```diff
@@ -185,9 +185,12 @@ pub(crate) async fn create_sockets(
 
     // Binding to a TCP address of the form `hostname:port` unnecessarily involves a DNS query. We
     // should get the port from here, but otherwise just bind to `0.0.0.0`.
-    let port_re = Regex::new(r":(\d{1,5})$").unwrap();
+    let port_re = Regex::new(r"(?<proto>\w+:)?(?<host>.*):(?<port>\d{1,5})$").unwrap();
```
antiguru, line 188: Though: should we use the `url` crate to parse the address?

antiguru (approved):
Looks great; thank you!

teskje:
TFTR!

PR #32000, opened 2025-03-24
doc/user: improve percentile queries

teskje:
This commit makes a number of improvements to the percentile queries in our docs, specifically the ones using HDR histograms.

* Fixes handling of 0 input values, which previously would cause the `hdr_histogram` view to error because `log(2, 0)` is undefined.
* Fixes handling of negative values, which previously would end up in positive (and duplicate) buckets because the sign was applied twice.
* Makes the `hdr_histogram` view contain bucket values instead of `(sign, mantissa, exponent)` triples.
* Makes use of CTEs to give names to things and improve readability.
* Removes an `ORDER BY` clause from the `hdr_distribution` definition where it had no effect.

### Motivation

The HDR histogram queries in our docs didn't handle zero and negative input values correctly.

### Tips for reviewer

Note that this change removes the splitting out of the `sign` part and instead keeps the mantissa signed. This simplifies the query a bit (no need to compute and later re-apply the sign, no need to `abs` the mantissa) but there might have been a good reason to split out the sign that I'm not aware of.

### Checklist

### DIFF
--- a/doc/user/content/transform-data/patterns/percentiles.md
+++ b/doc/user/content/transform-data/patterns/percentiles.md
@@ -69,39 +69,54 @@ HDR histograms are related to how [floating point numbers are represented](https
 
 The following snippets reduce the precision of numbers to limit the amount of required buckets. Numbers are decomposed into their floating point representation
 
-	n = sign * mantissa * 2**exponent
+	n = mantissa * 2**exponent
 
 and the precision of the mantissa is then reduced to compute the respective bucket.
 
-The basic ideas of using histograms to compute percentiles remain the same; but determining the bucket becomes more involved because it’s now composed of the triple (sign, mantissa, exponent).
+The basic ideas of using histograms to compute percentiles remain the same; but determining the bucket becomes more involved because it’s now composed of the tuple (mantissa, exponent).
 
 ```mzsql
 -- precision for the representation of the mantissa in bits
 \set precision 4
 
 CREATE VIEW hdr_histogram AS
+WITH
+  input_parts AS (
+    SELECT
+      CASE WHEN value = 0 THEN NULL
+           ELSE trunc(log(2, abs(value)))::int
+      END AS exponent,
+      CASE WHEN value = 0 THEN NULL
+           ELSE value / pow(2.0, trunc(log(2, abs(value)))::int)
+      END AS mantissa
+    FROM input
+  ),
+  buckets AS (
+    SELECT
+      trunc(mantissa * pow(2.0, :precision)) / pow(2.0, :precision)
+        * pow(2.0, exponent)
+        AS bucket
+    FROM input_parts
+  )
 SELECT
-  CASE WHEN value<0 THEN -1 ELSE 1 END AS sign,
-  trunc(log(2.0, abs(value)))::int AS exponent,
-  trunc(pow(2.0, :precision) * (value / pow(2.0, trunc(log(2.0, abs(value)))::int) - 1.0))::int AS mantissa,
+  COALESCE(bucket, 0) AS bucket,
   count(*) AS frequency
-FROM input
-GROUP BY sign, exponent, mantissa;
+FROM buckets
+GROUP BY bucket;
 ```
 
-The `hdr_distribution` view below reconstructs the `bucket` (with reduced precision), and determines the cumulative count and cumulative distribution.
+The `hdr_distribution` view below determines the cumulative count and cumulative distribution.
 
 ```mzsql
 CREATE VIEW hdr_distribution AS
 SELECT
-  h.sign*(1.0+h.mantissa/pow(2.0, :precision))*pow(2.0,h.exponent) AS bucket,
+  h.bucket,
   h.frequency,
   sum(g.frequency) AS cumulative_frequency,
   sum(g.frequency) / (SELECT sum(frequency) FROM hdr_histogram) AS cumulative_distribution
 FROM hdr_histogram g, hdr_histogram h
-WHERE (g.sign,g.exponent,g.mantissa) <= (h.sign,h.exponent,h.mantissa)
-GROUP BY h.sign, h.exponent, h.mantissa, h.frequency
-ORDER BY cumulative_distribution;
+WHERE g.bucket <= h.bucket
+GROUP BY h.bucket, h.frequency;
 ```
 
 This view can then be used to query _approximate_ percentiles. More precisely, the query returns the lower bound for the percentile (the next larger bucket represents the upper bound).
--- a/doc/user/content/transform-data/patterns/percentiles.md
+++ b/doc/user/content/transform-data/patterns/percentiles.md
@@ -18,41 +18,43 @@ values.
 Instead, more memory efficient alternatives are to use **histograms** and **High
 Dynamic Range (HDR) histograms**:
 
-- [Histograms](#using-histograms-to-compute-percentiles) have a lower memory
-  footprint that is linear to the number of _unique_ values and can compute
-  precise percentiles. However, for domains with _high_ cardinality, calculating
-  precise percentiles may be computationally expensive in processing time.
+- [Histograms](#using-histograms-to-compute-exact-percentiles) have a lower
+  memory footprint that is linear to the number of _unique_ values and can
+  compute precise percentiles. However, for domains with _high_ cardinality,
+  calculating precise percentiles may be computationally expensive.
 
 - [HDR histograms](#using-hdr-histograms-to-compute-approximate-percentiles)
-  further reduce the memory footprint but computes approximate percentiles. HDR
-  histograms are recommended if there is a long tail of large values that you
-  want to track, such as with latency measurements.
+  further reduce the memory footprint but computes approximate percentiles.
+  Depending on the precision needed for the percentiles, HDR histograms may be
+  preferred for domains with a high cardinality and dynamic range of values.
 
-## Using histograms to compute percentiles
+## Using histograms to compute exact percentiles
 
-Histograms track the frequency for each unique value/bucket instead of
-tracking all values, and can be used to compute exact percentiles. To compute
-exact percentiles, However, for
-domains with high cardinality, because of the cross join to calcuate
-distributions, the memory usage can become significant. As such, using
-histograms to compute percentiles are better suited for domains with _low_
-cardinality.
+Histograms summarize data sets by grouping values into ranges and counting how
+many elements fall into each range. From this summary, you can get the
+percentile information by identifying the range where the cumulative count
+crosses the desired percentile threshold. By grouping each distinct value into
+its own range, you can get exact percentiles; however, this can be
+computationally expensive if there are large number of distinct values.
+Alternatively, you can get an approximate percentiles by using [HDR
+histograms](#using-hdr-histograms-to-compute-approximate-percentiles).
 
-To use histograms to compute percentiles:
 
-- First, create a histogram view that tracks the unique values and their
-  associated frequency.
+To use histograms to compute exact percentiles:
+
+- First, create a histogram view that groups each distinct value into its own
+  bucket and counts the number of each distinct value.
 
 - Then, using a cross join on the histogram view, create a distribution view
-  that calculates the distribution of the values. The distribution is calculated
-  by dividing the cumulative frequency (the frequency of all values up to and
-  including that bucket) by the total frequency of all bucket values.
+  that calculates the cumulative density for a bucket by dividing the cumulative
+  counts (sum of the counts for all buckets up to and including that bucket) by
+  the total count.
 
   {{< note >}}
 
-  For domain with high cardinality, because of the cross join (which grows
-  quadratic in the number of unique values), calculating precise percentiles may
-  be computationally expensive(i.e., time-consuming).
+  The use of the cross join produces a number of outputs that is quadratic in
+  the input. And, while the results will only be linear in size, it may take a
+  disproportionate amount of time to produce and maintain.
 
   {{</ note >}}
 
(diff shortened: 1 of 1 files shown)


kay-kim (commented):
Thanks Jan!  I'll add a patch to this to refactor the existing content a bit.  It didn't seem fair to make you sort out existing crazy layout 😄

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -69,39 +69,54 @@ HDR histograms are related to how [floating point numbers are represented](https
...
+	n = mantissa * 2**exponent
 
 and the precision of the mantissa is then reduced to compute the respective bucket.
 
-The basic ideas of using histograms to compute percentiles remain the same; but determining the bucket becomes more involved because it’s now composed of the triple (sign, mantissa, exponent).
+The basic ideas of using histograms to compute percentiles remain the same; but determining the bucket becomes more involved because it’s now composed of the tuple (mantissa, exponent).
 
 ```mzsql
 -- precision for the representation of the mantissa in bits
 \set precision 4
```
kay-kim, line 80: Will fix in patch.

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -10,134 +10,306 @@ menu:
...
+   SELECT
+     h.bucket,
+     h.frequency,
+     sum(g.frequency) AS cumulative_frequency,
+     sum(g.frequency) / (SELECT sum(frequency) FROM hdr_histogram) AS  cumulative_distribution
+   FROM hdr_histogram g, hdr_histogram h
+   WHERE g.bucket <= h.bucket
+   GROUP BY h.bucket, h.frequency;
+   ```
+
```
kay-kim, line 274: Double checking: The cross join here is still problematic but not to the same degree since we're reducing the bucket numbers?  That is, we alleviate that impact since we reduced the number of buckets.

teskje: Yes, exactly! The number of buckets should end up quite small (depending on the precision of the histogram), so the cross join doesn't hurt.

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -10,134 +10,306 @@ menu:
...
+   ```mzsql
+   CREATE VIEW histogram AS
+   SELECT
+     value AS bucket,
+     count(*) AS frequency
+   FROM input
+   GROUP BY value;
+   ```
+
+1. Create a view `distribution` to calculate the cumulative frequency/count and
```
frankmcsherry, line 86: We have sometimes "count/frequency" and sometimes "frequency/count". Specifically here were we are discussing accumulating a thing, we risk looking like a division sign. I think probably we could also lock in either count or frequency, earlier in the doc, and stick with that.

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -10,134 +10,306 @@ menu:
...
+   frequency (the count of all values up to and including that bucket) by the
+   total counts for all bucket values.
+
+   ```mzsql
+   CREATE VIEW distribution AS
+   SELECT
+     h.bucket,
+     h.frequency,
+     sum(g.frequency) AS cumulative_frequency,
+     sum(g.frequency) / (SELECT sum(frequency) FROM histogram) AS  cumulative_distribution
```
frankmcsherry, line 98: I think perhaps this should be `cumulative_density` rather than `cumulative_distribution`?

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -10,134 +10,306 @@ menu:
...
+   WHERE g.bucket <= h.bucket
+   GROUP BY h.bucket, h.frequency
+   ORDER BY cumulative_distribution;
+   ```
+
+   {{< note >}}
+
+   This view uses a self-join that scales quadratically with the number of
+   unique bucket values. If the data has high cardinality, memory usage may grow
+   significantly.
```
frankmcsherry, line 109: Fwiw, there's no reason memory needs to be large. If we have a cross join, it is the *output* that may be large, but the output above heads in to a reduction that can consolidate updates in place. It's possible that this is just linear memory, as it is (we should check the plan, though). The compute will be quadratic, and updates likely linear.

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -10,134 +10,306 @@ menu:
...
+   SELECT bucket AS percentile90
+   FROM distribution
+   WHERE cumulative_distribution >= 0.9
+   ORDER BY cumulative_distribution
+   LIMIT 1;
+   ```
+
+1. To keep the `distribution` view up to date as new data comes into `input`,
+   you can create an index on the view. To optimize for percentile queries,
+   index on the `cumulative_distribution` column to support [point lookup on
```
frankmcsherry, line 127: I feel like this index may not help much, and may just confuse the reader. The above query would work well, but only if indexes were ordered rather than hash indexes. I think creating such an index would not improve any queries other than ones where you know the numeric `cumulative_distribution` value you are looking for, which seems rare.

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -10,134 +10,306 @@ menu:
...
 
+### Example
 
-## Using HDR histograms to compute approximate percentiles
+{{< tip >}}
 
-[HDR histograms](https://github.com/HdrHistogram/HdrHistogram) can be used to approximate percentiles in a space efficient manner that scales well even for large domains with many distinct values. HDR histograms reduce the precision of values that are tracked and use buckets with variable width. Buckets that are closer to 0 are smaller whereas buckets far away from 0 are wider. This works particularly well for data that exhibits a long tail of large values, e.g., latency measurements.
+The following example assumes you have not previously created and populated the
+`input` table from the [Using histograms to compute percentiles
+example](#example). If you have created and populated the table, skip to step 3.
```
frankmcsherry, line 168: Maybe fine, but I can see the "step 3" reference rotting if we change the steps ever. Alternately, maybe: "skip the corresponding steps".

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -10,134 +10,306 @@ menu:
...
 
-```mzsql
-SELECT bucket AS percentile
-FROM distribution
-WHERE cumulative_distribution >= 0.9
-ORDER BY cumulative_distribution
-LIMIT 1;
-```
+In the example below, to reduce the number of buckets, the values are first
+decomposed into their floating point representation (i.e., `mantissa *
```
frankmcsherry, line 147: I came back to here when I realized that the mantissa is signed, and we don't have a separate sign bit. I think IEEE floating point has a sign bit, which was the source of my confusion. Perhaps there's a way to flag that around here?

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -10,134 +10,306 @@ menu:
...
+CREATE VIEW hdr_histogram AS
+WITH
+  input_parts AS (
+    SELECT
+      CASE WHEN value = 0 THEN NULL
+          ELSE trunc(log(2, abs(value)))::int
+      END AS exponent,
+      CASE WHEN value = 0 THEN NULL
+          ELSE value / pow(2.0, trunc(log(2, abs(value)))::int)
+      END AS mantissa
```
frankmcsherry, line 204: I understand the spirit of what is going on here, but perhaps a good moment to put a comment in describing the properties of this field? Something like `-- A signed value in [-1, -0.5) or (0.5, 1] that multiplies 2^exponent to reproduce value`

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -10,134 +10,306 @@ menu:
...
+    FROM input
+  ),
+  buckets AS (
+    SELECT
+      trunc(mantissa * pow(2.0, 4)) / pow(2.0, 4) -- precision is reduced to 4
```
frankmcsherry, line 209: This is where I needed to go back and understand `mantissa`. I think this is saying something like
```
-- map (0.5, 1] to (8, 16], and round toward zero
```
which .. typing it out seems a bit weird because `16` only shows up if we had exactly a mantissa of `1.0` if I understand correctly. Not terrible, just "wasting" a bucket. Can ponder if we should fix, but perhaps it's not the most important thing.

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -10,134 +10,306 @@ menu:
...
+   LIMIT 1;
+   ```
+
+1. To keep the `hdr_distribution` view up to date as new data comes into
+   `input`, you can create an index on the view. To optimize for percentile
+   queries, index on the `cumulative_distribution` column to support [point
+   lookup on the index](/concepts/indexes/#point-lookups):
+
+   ```mzsql
+   CREATE INDEX hdr_distribution_idx ON hdr_distribution (cumulative_distribution);
```
frankmcsherry, line 297: Same comment as above: I suspect this index will not help other than by maintaining the result (the `cumulative_distribution` column is probably hard to guess).

frankmcsherry (approved):
Generally looks good! I have a few nits here and there, and also a "technical" nit around how we pick bucket boundaries. If there's an easy fix, great (or if I'm wrong, also great!). If no easy fix, no worries.

doc/user/content/transform-data/patterns/percentiles.md
teskje: I also wasn't sure about this part. Removing the separate sign column makes the query simpler, so I think it's a good change. But the reference to floating point representations might be confusing. We could consider just not mentioning floats?

Fwiw, the wikipedia page for [floating point arithmetic](https://en.wikipedia.org/wiki/Floating-point_arithmetic) mostly doesn't mention a separate sign bit but uses a signed mantissa like we do here. Though your confusion makes me suspect that people are generally more familiar with the IEEE format than with the theoretical concept.

doc/user/content/transform-data/patterns/percentiles.md
teskje: This part is made awkward by SQL limitations. Ideally this would be `value / pow(2.0, exponent)`, which would be quite obvious imo (especially because we point out the `mantissa * 2^exponent` representation above. But we can't name `exponent` here :/

doc/user/content/transform-data/patterns/percentiles.md
teskje: Hm, the way I think about this (I didn't write the logic, just cleaned it up) is that multiplying with `2^x` is a left shift by `x` and dividing is a right shift. So here we shift left by 4 bits, truncate to an integer, which clears all the bits beyond our selected precision of 4, and then shift back.

I'm not sure where we'd end up wasting a bucket in this. But I also don't understand where the `(0.5, 1]` interval comes from, as noted above I _think_ it should be `[1, 2)`.

doc/user/content/transform-data/patterns/percentiles.md
kay-kim: So ... is it more: 

> However, for domains with _high_ cardinality, calculating precise percentiles may be computationally expensive in processing time.

doc/user/content/transform-data/patterns/percentiles.md
kay-kim: removed in patch 75f32afbf3839b0dc7b5f84321847c6c88161bff

doc/user/content/transform-data/patterns/percentiles.md
kay-kim: Updated in patch 75f32afbf3839b0dc7b5f84321847c6c88161bff

doc/user/content/transform-data/patterns/percentiles.md
kay-kim: Updated in patch 75f32afbf3839b0dc7b5f84321847c6c88161bff

doc/user/content/transform-data/patterns/percentiles.md
kay-kim: Removed in patch 75f32afbf3839b0dc7b5f84321847c6c88161bff

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -144,9 +136,8 @@ their relative error becomes less relevant (e.g. 10000 ms and 10001 ms are
 almost the same and can reside in the same bucket).
 
 In the example below, to reduce the number of buckets, the values are first
-decomposed into their floating point representation (i.e., `mantissa *
-2^exponent`), and then with the precision of the mantissa lowered,
-reconstructed for the respective bucket value.
+decomposed into `signed_mantissa * 2^exponent`, and then with the precision of
```
frankmcsherry, line 139: `signed_mantissa` eventually becomes `mantissa --signed`. Perhaps we want to bring those together somehow. It seems fine to put here "`mantissa * 2^exponent` where the mantissa is signed"? I think it's perhaps a terminological bug in that the thing we are discussing is the "[significand](https://en.wikipedia.org/wiki/Significand)", and the mantissa is only the part after the decimal point in the normalized significand.
Screenshot 2025-03-27 at 12 38 02 PM

frankmcsherry (approved):
I have a bunch of nits. Mainly around the use of "frequency" instead of "count", and the use of "mantissa" where we mean "significand". None of these are existential, but there's the potential to be clearer with different terms.

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -20,20 +20,19 @@ Dynamic Range (HDR) histograms**:
 
 - [Histograms](#using-histograms-to-compute-percentiles) have a lower memory
   footprint that is linear to the number of _unique_ values and can compute
-  precise percentiles. However, calculating distributions using histograms
-  involves a self-join, which can cause memory usage to grow quadratically.  As
-  such, percentile calculations using histograms are better suited for domains
-  with _low_ cardinality.
+  precise percentiles. However, for domains with _high_ cardinality, calculating
+  precise percentiles may be computationally expensive in processing time.
```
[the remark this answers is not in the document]
kay-kim: 👍  updated in patch https://github.com/MaterializeInc/materialize/pull/32000/commits/6bb2d03d917cc7f8037fbc2ecf87730364a747c3

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -20,20 +20,19 @@ Dynamic Range (HDR) histograms**:
...
 
 - [HDR histograms](#using-hdr-histograms-to-compute-approximate-percentiles)
   further reduce the memory footprint but computes approximate percentiles. HDR
   histograms are recommended if there is a long tail of large values that you
-  want to track, which is often the case for latency measurements.
+  want to track, such as with latency measurements.
```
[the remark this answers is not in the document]
kay-kim: 👍  updated in patch https://github.com/MaterializeInc/materialize/pull/32000/commits/6bb2d03d917cc7f8037fbc2ecf87730364a747c3

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -20,20 +20,19 @@ Dynamic Range (HDR) histograms**:
...
 
 ## Using histograms to compute percentiles
 
-Histograms track the count/frequency for each unique value/bucket instead of
-tracking all values, and can be used to compute exact percentiles. However, for
+Histograms track the frequency for each unique value/bucket instead of
+tracking all values, and can be used to compute exact percentiles. To compute
```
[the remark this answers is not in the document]
kay-kim: 👍  updated in patch https://github.com/MaterializeInc/materialize/pull/32000/commits/6bb2d03d917cc7f8037fbc2ecf87730364a747c3

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -83,11 +83,11 @@ To use histograms to compute percentiles:
    GROUP BY value;
    ```
 
-1. Create a view `distribution` to calculate the cumulative frequency/count and
-   the cumulative distribution. The view definition uses a cross join on the
+1. Create a view `distribution` to calculate the cumulative frequency and the
+   cumulative distribution. The view definition uses a cross join on the
```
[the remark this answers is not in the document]
kay-kim: 👍  updated in patch https://github.com/MaterializeInc/materialize/pull/32000/commits/6bb2d03d917cc7f8037fbc2ecf87730364a747c3

doc/user/content/transform-data/patterns/percentiles.md
```diff
@@ -104,9 +104,9 @@ To use histograms to compute percentiles:
 
    {{< note >}}
 
-   This view uses a self-join that scales quadratically with the number of
-   unique bucket values. If the data has high cardinality, memory usage may grow
-   significantly.
+   Because of the cross join (which grows quadratic in the number of unique
```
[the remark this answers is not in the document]
kay-kim: 👍  updated in patch https://github.com/MaterializeInc/materialize/pull/32000/commits/6bb2d03d917cc7f8037fbc2ecf87730364a747c3

PR #32002, opened 2025-03-24
Revert "storage,0dt: include all ingestion collections in caught-up check"

ParkMyCar:
This reverts commit 323190b1136ce3494c709d340a3baa03031f6926.

### Motivation

Subsources lazily report their status, so for certain loadgen subsources that produce data once and never again, we never pass the hydration checks and thus never switch over.

### Checklist

### DIFF
--- a/src/adapter/src/coord/caught_up.rs
+++ b/src/adapter/src/coord/caught_up.rs
@@ -244,7 +244,8 @@ impl Coordinator {
         let storage_frontiers = self
             .controller
             .storage
-            .active_ingestion_collections(cluster.id)
+            .active_ingestions(cluster.id)
+            .copied()
             .filter(|id| !id.is_transient() && !exclude_collections.contains(id))
             .map(|id| {
                 let (_read_frontier, write_frontier) =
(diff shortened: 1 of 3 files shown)


bkirwi (approved):
This looks like a clean revert - and I've checked that the reverted logic included the overall ingestion id in the returned set, so this should block on strictly fewer ids after the revert.

I guess the risk here is that we cut over too soon, when not all subsources have caught up? cc @petrosagg, who reviewed the original PR, in case he has insight on this.

ParkMyCar:
> I guess the risk here is that we cut over too soon, when not all subsources have caught up? cc @petrosagg, who reviewed the original PR, in case he has insight on this.

+1 I was concerned about this risk, but it seems like it's the behavior in v0.137 so... I'm not too concerned?

teskje (approved):

aljoscha:
This got re-implemented in https://github.com/MaterializeInc/materialize/pull/33589, and then fixed in https://github.com/MaterializeInc/materialize/pull/33672

PR #32005, opened 2025-03-25
WIP Password auth design doc

jubrad:
### Checklist

### DIFF
--- a/doc/developer/design/20250321_password_authentication.md
+++ b/doc/developer/design/20250321_password_authentication.md
@@ -74,6 +74,19 @@ Users should be able to manage their passwords and set passwords for roles they 
 
 -   **Password Dating**: The creation date of passwords will be recorded along with the hashed and salted password. When the password changes a creation date should be reset. The creation date of passwords should be queryable.
 
+-   **Storage Location Password Hash**: The password will be stored in the catalog in its own object, `RoleAuth`. This object must record the following:
+    - role_id
+    - creation_date
+    - algorithm / algorithm meta: IE for sha256: algo: sha256, salt: "some-string", iteration_count: some-int
+    Example struct:
+    ```Rust
+    RoleAuth {
+        role_id: ident
+        password_hash: String
+        creation_date: Date
+    }
+    ```
+
 -   **OPTIONAL: Password Versioning**: Updates to the password hashing mechanisms may be required. As long as we are receiving passwords in plain text we should be able to take a validated password and replace the existing a new securely hashed value. This may require prefixing the password with data about the hash algorithm or parameters.
 
 ### 4. Configurable admin system login:
@@ -83,11 +96,38 @@ Passwords for `mz_system` and `mz_support` roles will be settable via environmen
 
 ### 5. HTTP Authentication:
 
-Environmentd's HTTP endpoint will need to have added support for session-based authentication. This includes session creation, storage, and deleting/setting cookies. To handle sessions we should use [tower session-store](https://docs.rs/tower-sessions/latest/tower_sessions/). This should explicitly require secure cookies `.with_secure(true)` when running with TLS.
+Environmentd's HTTP endpoint will have added support for session-based authentication. This includes session creation, storage, and deleting/setting cookies. To handle sessions we should use [tower session-store](https://docs.rs/tower-sessions/latest/tower_sessions/). This should explicitly require secure cookies `.with_secure(true)` when running with TLS.
+Required new endpoints:
+  - POST: api/login 
+    - Accepts username, password Authentication Headers
+    - imposes strict rate limiting : returning 429
+    - returns 401 on auth failure
+    - returns 201 on success and sets a session token in the cookies
+  - POST: api/logout
+    - returns 204 if session is found
+    - returns 404 if session is not found
 
-### Configurable Authentication Mechanism
+Additionally, we'll need to add a tower middleware layer that validates cookie authentication for all endpoints returning 403 when the user is not authenticated.
+
+
+### Related Superuser Details:
+Enabling this feature we will require us to provide an alternative method to declare superusers. This is currently done through JWT metadata. In order for this to work with passwords we will need to provide syntax:
+```sql
+ALTER ROLE ... WITH SUPERUSER;
+ALTER ROLE ... WITH NO SUPERUSER;
+```
+
+A role's superuser state should be stored as a rolsuper along with the user in the catalog. It may make sense to make this an Option<bool>. If the value of rolsuper is not `Some(v)` then we defer to metadata if we're using frontegg auth. 
+
+
+### Note on Configurable Authentication Mechanism
 
 The authentication mechanism for an environment must be configurable. A new flag will be created to select the authentication `mz_authentication_type` with the options of `frontegg`, `password`, and `disabled`. This will subsume the enable_authentication flag.
+
+
+### Note on Console builds:
+
+We currently re-use the same Console build for the emulator and self-managed. For practical purposes we can build both with password auth enabled with no option to disable. 
 
 ## Minimal Viable Prototype
 
@@ -98,6 +138,7 @@ A minimal viable prototype (MVP) for this solution will include:
 3. **Secure Password Storage**: Implement hashed salted password storage using compatible with `scram-sha-256`.
 4. **Bootstrapping System Users**: Provide a mechanism for bootstrapping the password and enabling login for system users.
 5. **HTTP Session Management**: Environmentd's HTTP layer needs to handle session management via cookies and internal session storage.
+6. **Enhanced Superuser Management**: Support assigning / unassigning superuser for roles.
 
 ## Alternatives
 
@@ -111,7 +152,7 @@ A minimal viable prototype (MVP) for this solution will include:
 
 3. **Store Passwords in K8s Secrets**: It would be viable to store passwords in Kubernetes secrets.
 
--   **Reasons Not Chosen**: This may create an untenable number of Kubernetes secrets.
+-   **Reasons Not Chosen**: This may create an untenable number of Kubernetes secrets, and may cause issues with passing security reviews.
 
 4. **App Passwords**: Rather than allowing users to configure a password per user we could generate "app-passwords" that have a many-to-one relationship with users, are auto-generated, and stored in k8s secrets.
 
--- a/doc/developer/design/20250321_password_authentication.md
+++ b/doc/developer/design/20250321_password_authentication.md
@@ -87,11 +87,14 @@ Users should be able to manage their passwords and set passwords for roles they 
     }
     ```
 
--   **OPTIONAL: Password Versioning**: Updates to the password hashing mechanisms may be required. As long as we are receiving passwords in plain text we should be able to take a validated password and replace the existing a new securely hashed value. This may require prefixing the password with data about the hash algorithm or parameters.
+-   **OPTIONAL: Password Versioning**: Updates to the password hashing mechanisms may be required. 
+  - For passwords received in plain text we should be able re-encrypt them with stronger algorithms.
+  - For client side challenges we may be able to increase the number of hash iterations.
+  We will already be storing the necessary password hashing metadata to perform these actions.
 
 ### 4. Configurable admin system login:
 
-Passwords for `mz_system` and `mz_support` roles will be settable via environment variables `MZ_MZ_<USER>_EXTERNAL_LOGIN_PASSWORD`. Orchestratord should provide a set of parameters to set these variables via a Kubernetes secret. Additionally, We should enable login of system users through the external ports when they have external login passwords set. Login through the external port must not be possible unless this flag is set, this logic should not rely on whether the internal user has a password.
+Passwords for `mz_system` and `mz_support` roles will be settable via environment variables `MZ_MZ_<USER>_EXTERNAL_LOGIN_PASSWORD`. Orchestratord should provide a set of parameters to set these variables via a Kubernetes secret. Additionally, We should enable login of system users through the external ports when they have external login passwords set. Login through the external port must not be possible unless this flag is set, this logic should not rely on whether the internal user has a password. Our Helm chart and Orchestratord will be adjusted to support these parameters.
 
 
 ### 5. HTTP Authentication:
@@ -127,7 +130,7 @@ The authentication mechanism for an environment must be configurable. A new flag
 
 ### Note on Console builds:
 
-We currently re-use the same Console build for the emulator and self-managed. For practical purposes we can build both with password auth enabled with no option to disable. 
+Console does currently do not support runtime or startup configuration. Configuration is handled only at build time. To resolve this we should add a `config.json` or `config.js` file which can be mounted directly into the Nginx container assets. This file should come from a materialize-console config map which must be setup by Orchestratord.  We will also need changes to the console to support reading in configuration from this map. The initial config value here should be `authentication_type: password`, in cloud we should use `authentication_type: frontegg` or `authentication_type: jwt`. The console build process can still be used to set default values for this config file. 
 
 ## Minimal Viable Prototype
 
--- a/doc/developer/design/20250321_password_authentication.md
+++ b/doc/developer/design/20250321_password_authentication.md
@@ -1,7 +1,7 @@
 # Adding Password Authentication to Materialize
 
-- Associated: 
-  - https://github.com/MaterializeInc/cloud/issues/10754
+-   Associated:
+    -   https://github.com/MaterializeInc/cloud/issues/10754
 
 ## The Problem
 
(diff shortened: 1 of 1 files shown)


doc/developer/design/20250321_password_authentication.md
```diff
@@ -0,0 +1,121 @@
...
+
+- **Password Dating**: The creation date of passwords will be recorded along with the hashed and salted password. When the password changes a creation date should be reset. The creation date of passwords should be queryable.
+
+- **OPTIONAL: Password Versioning**: Updates to the password hashing mechanisms may be required. As long as we are receiving passwords in plain text we should be able to take a validated password and replace the existing a new securely hashed value. This may require prefixing the password with data about the hash algorithm or parameters.
+
+### 4. Configurable System User Passwords:
+Passwords for `mz_system` and `mz_support` roles will be settable via environment variables `MZ_MZ_SYSTEM_ROLE_PASSWORD` and `MZ_MZ_SUPPORT_ROLE_PASSWORD`. Orchestratord should provide a set of parameters to set these variables via a kubernetes secret. Additionally, We should enable login of `mz_system` through the external ports when they have passwords set. This should require both an explicit feature flag, `enable_public_internal_user_login` to be set to true, and a password set in an environment variable.
```
alex-hunt-materialize, line 74: Setting these in an environment variable would work, but would require a restart to change the password. If we instead mount the secret as a file, we can use an inotify watch to trigger an immediate reload of the file.

I'm also not sure about the behavior of this feature flag. Wouldn't we want that to default to true? Do we even need a feature flag if we have a CLI argument for the file path of this system user secrets file?

doc/developer/design/20250321_password_authentication.md
```diff
@@ -0,0 +1,121 @@
...
+
+- **Access Control**: Hashed passwords should not be extracted from the catalog and should not be accessible through queryable tables. Passwords must not be logged or output in stack traces.
+
+- **Password Dating**: The creation date of passwords will be recorded along with the hashed and salted password. When the password changes a creation date should be reset. The creation date of passwords should be queryable.
+
+- **OPTIONAL: Password Versioning**: Updates to the password hashing mechanisms may be required. As long as we are receiving passwords in plain text we should be able to take a validated password and replace the existing a new securely hashed value. This may require prefixing the password with data about the hash algorithm or parameters.
```
alex-hunt-materialize, line 71: This should be required, not optional.

doc/developer/design/20250321_password_authentication.md
```diff
@@ -0,0 +1,121 @@
...
+
+ - **Reasons Not Chosen**: This is slightly more complicated and non-compliant with Postgres.
+ 
+ - **Reasons I still might prefer this**: It does have a number of benefits, including slightly better security via enforcement of random strong passwords, and a better rotation story. It's also possible we could have different app password protocols/versions that allow users to manage their passwords entirely through secrets directly or something like Hashicorp vault.
+
+## Open questions
+1. **Where exactly are passwords stored**: While it makes sense to store passwords in the catalog, I have not yet identified exactly where they should be stored. Should it go alongside `mz_roles`, or in a new `mz_auth_id` table. The latter makes sense as it could store password metadata, and might let us store multiple passwords down the road.
+
+2. **Audit Logging and Brute Force Detection**: Do we need to log all failed password attempts and the IPs of those attempts? Do we need to take action to lock an IP or User when an attempt threshold is met?
```
alex-hunt-materialize, line 117: Any action taken should be user configurable. We don't want to allow DOS attacks by a user just spamming incorrect passwords, for example.

doc/developer/design/20250321_password_authentication.md
```diff
@@ -0,0 +1,121 @@
...
+**Password Rotation or Expiration**: While not currently in scope, the design should allow for additions in the future, such as password rotation and expiration.
+
+**Abuse Detection**: Action should be taken action when brute force attempts are detected, either by locking the user or blocking the requesting client IP.
+
+### 2. Password Management Syntax:
+Users should be able to manage their passwords and set passwords for roles they create. Superusers should be able to alter any non-internal user's password.
+- **Syntax Examples**:
+    - `CREATE ROLE hunter WITH PASSWORD 'hunter2';` 
+    - `ALTER ROLE hunter SET (PASSWORD 'hunter2');` 
+     - `ALTER USER user_name PASSWORD NULL`
```
jubrad, line 56: Should be `ALTER ROLE`

SangJunBak:
Regarding closing off the internal port for Cloud in the cloud sync: 
- We should consider Teleport too since i think it uses some internal port?

jubrad:
Option A) - preferred
We should consider adding confimap syncing as a self-managed replacement for LD for system vars. If we do this we can probably drop the need for mz_system (in self-managed) and just close off the internal port for self-managed, perhaps not even start it up.

Option B) - backup
- We need to evaluate other mz_system capabilities and ensure that it is not needed for any other purposes. 
If mz_system is needed then we may need to optionally allow external port login for mz_system with a password.

jubrad:
On system users... We might be able to pass in a bootstrap user that is also considered internal/system. This user should be considered superuser and internal user and it's password should come from a k8s secret. The k8s secret passed in can be used as it's password secrets reference in the catalog.

doc/developer/design/20250321_password_authentication.md
```diff
@@ -0,0 +1,128 @@
...
+
+-   **Password Dating**: The creation date of passwords will be recorded along with the hashed and salted password. When the password changes a creation date should be reset. The creation date of passwords should be queryable.
+
+-   **OPTIONAL: Password Versioning**: Updates to the password hashing mechanisms may be required. As long as we are receiving passwords in plain text we should be able to take a validated password and replace the existing a new securely hashed value. This may require prefixing the password with data about the hash algorithm or parameters.
+
+### 4. Configurable admin system login:
+
+Passwords for `mz_system` and `mz_support` roles will be settable via environment variables `MZ_MZ_<USER>_EXTERNAL_LOGIN_PASSWORD`. Orchestratord should provide a set of parameters to set these variables via a Kubernetes secret. Additionally, We should enable login of system users through the external ports when they have external login passwords set. Login through the external port must not be possible unless this flag is set, this logic should not rely on whether the internal user has a password.
```
SangJunBak, line 81: Nit: Should be `MZ_<USER>_EXTERNAL_LOGIN_PASSWORD` given `mz_system` would give `MZ_MZ_MZ_SYSTEM_EXTERNAL_LOGIN_PASSWORD`

SangJunBak:
Still not clear to me the flow of setting up auth. In my head, it looks like:
1. As they initially set up via the helm chart, orchestratord will have the env variable to set the password of the `mz_system` user. This is probably configurable via the helm chart?
2. Once the user deploys the materialize CR, orchestratord will pass that env variable in order to start the environment with an mz_system type user
3. The user can now login via that user and add more roles

Does that sound correct? Then we just make it a limitation that a user can't use the Console unless they have . In the Console, we'll most likely need the following information 
- Is (password) auth enabled?
- Does there exist an authenticated user?
But I'm also confused how the Console might receive this information if we can't query the database to begin with. 


Should also consider the migration path from self managed customers without auth to customers with auth.

doc/developer/design/20250321_password_authentication.md
jubrad: If this is not optional I think we need to define the scope, for instance we cannot re-hash password to a better algorithm unless we're provided with the raw password value, so we may be able to provide something near term, but that would fall apart if we started using SCARM.  Also assuming we stay on scram-sha-256 (seems reasonable) we may be able to change the hash iterations, without re-reading the raw password but that's the extent. 

Since the password encryption type is going to add compatibility constraints the user specified pgwire authentication type I think we're going to need to:
1. keep track of the password algorithm
2. surface the algorithm used for a password to users
3. let user reset their password specifying which algorithm they want. 

This is largely the take that both postgres and cockroach use.

For V1 I think we just keep track of the password metadata: hash algo, iterations, creation date, secret location

doc/developer/design/20250321_password_authentication.md
alex-hunt-materialize: Maybe the user should be last? `MZ_EXTERNAL_LOGIN_PASSWORD_<USER>`

doc/developer/design/20250321_password_authentication.md
DAlperin: Pg stores them as `SCRAM-SHA-256$<iteration count>:<salt>$<StoredKey>:<ServerKey>` I propose we start with scram-sha256 and store as such

jubrad: ok I spoke with @alex-hunt-materialize  on this..

We need to keep track of iterations any way... lets just use that for now. If we need to bump PW strength we can do it seamlessly as long as it's just through iterations. If we need a new algo because sha256 gets cracked we will cross that road then.

doc/developer/design/20250321_password_authentication.md
```diff
@@ -0,0 +1,128 @@
...
+
+    - **Reasons Not Chosen**: This solution is more involved, requiring significant additional work to secure, develop, and maintain. Additionally, it does not align with the Postgres.
+
+2. **Password Authentication in Ingress Layer**: Using a front-end proxy like PGbouncer or NGINX to handle password authentication before traffic hits Materialize.
+
+    - **Reasons Not Chosen**: This approach requires significant customization and doesn't solve the problem within the Materialize database itself. It also still necessitates internal authentication for role management.
+
+3. **Store Passwords in K8s Secrets**: It would be viable to store passwords in Kubernetes secrets.
+
+-   **Reasons Not Chosen**: This may create an untenable number of Kubernetes secrets.
```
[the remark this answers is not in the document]
DAlperin: Yeah I think this is from the original pass. Let's remove

alex-hunt-materialize:
@SangJunBak +1 to everything you said.

>But I'm also confused how the Console might receive this information if we can't query the database to begin with.

Maybe we can pass that to the console also? Maybe through an env var? That gets a bit tricky, since I think we update the balancers and console immediately upon getting an updated Materialize, while only updating environmentd if we trigger a rollout. Perhaps this logic needs to change, but that may be difficult too.

doc/developer/design/20250321_password_authentication.md
alex-hunt-materialize: Sorry for my poor initial wording on this. I just meant keeping track of the algorithm, iteration, etc... not implementing the actual rotation. The PG stored values Dov mentioned look good to me.

doc/developer/design/20250321_password_authentication.md
jubrad: Sounds good!

`EXTERNAL_LOGIN_PASSWORD_<USER>`

with a env variable of 

`MZ_EXTERNAL_LOGIN_PASSWORD_<USER>`

SangJunBak:
@jubrad 
> Unless they have what?
Ah my bad. Unless there's a registered role! Edited. 

>The standard auth for postgres has login and non login users see [pg_authid](https://www.postgresql.org/docs/current/catalog-pg-authid.html).

But I mean in order for the Console to query `pg_authid` via the HTTP API, wouldn't they need to be authenticated?

SangJunBak:
>Maybe we can pass that to the console also? Maybe through an env var? That gets a bit tricky, since I think we update the balancers and console immediately upon getting an updated Materialize

@alex-hunt-materialize and with the current design, an env variable won't do much good given the Console's already built by the time orchestratord installs the console pod. There's currently no way for orchestratord to edit build variables for the Console, which I think we should fix!

jubrad:
> > Unless they have what?
> > Ah my bad. Unless there's a registered role! Edited.
> 
> > The standard auth for postgres has login and non login users see [pg_authid](https://www.postgresql.org/docs/current/catalog-pg-authid.html).
> 
> But I mean in order for the Console to query `pg_authid` via the HTTP API, wouldn't they need to be authenticated?

Console shouldn't need to query pg_authid. It should just need to attempt to connect using username and password. If the role isn't a login role or there's a password mismatch that should all be handled server side. To determine if auth is enabled it could just attempt to run a query and, on failed login, route to the login box, or we can install a configurable endpoint on NGINX that console can hit to tell it what kind of auth it should attempt to use.

doc/developer/design/20250321_password_authentication.md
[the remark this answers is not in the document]
jubrad: After thinking about this some more and talking it over, I think we should just store passwords in the catalog. This is considered user metadata, it seems like it should be relatively simple to keep there may simplify security reviews. 

see:
https://materializeinc.slack.com/archives/C07PN7KSB0T/p1743173985961509

SangJunBak (approved):
LGTM on my end! Thanks for writing all this out!

doc/developer/design/20250321_password_authentication.md
```diff
@@ -127,7 +130,7 @@ The authentication mechanism for an environment must be configurable. A new flag
 
 ### Note on Console builds:
 
-We currently re-use the same Console build for the emulator and self-managed. For practical purposes we can build both with password auth enabled with no option to disable. 
+Console does currently do not support runtime or startup configuration. Configuration is handled only at build time. To resolve this we should add a `config.json` or `config.js` file which can be mounted directly into the Nginx container assets. This file should come from a materialize-console config map which must be setup by Orchestratord.  We will also need changes to the console to support reading in configuration from this map. The initial config value here should be `authentication_type: password`, in cloud we should use `authentication_type: frontegg` or `authentication_type: jwt`. The console build process can still be used to set default values for this config file.
```
SangJunBak, line 133: in the case where someone wants to bump the Console version or if we ever need to reinstall the `console` service, we can most likely just swap out the `config` file given the Console will fetch it on runtime.

SangJunBak: We'll most likely just block the UI until the config map is loaded in (i.e. via a loading screen). This is fine given loading it in should be really quick.

doc/developer/design/20250321_password_authentication.md
jasonhernandez: Agree with @alex-hunt-materialize - there's too much risk to availability from DOS or automatic retries with an incorrect password.
Our default behavior should log all login attempts, and we should aim to replicate PostgreSQL log messages. This will allow customers to send the data to their SIEM and/or use fail2ban or another tool to act on any failures.
We'd need to add a way for customers to lock users with fail2ban or a script, but I think we can leave that out for now unless it is very simple to implement.

here's some discussion on how to configure fail2ban w/ Postgres, for example
https://gist.github.com/rc9000/fd1be13b5c8820f63d982d0bf8154db1

doc/developer/design/20250321_password_authentication.md
```diff
@@ -0,0 +1,176 @@
...
+    -   `ALTER ROLE hunter SET (PASSWORD 'hunter2');`
+    -   `ALTER ROLE user_name PASSWORD NULL`
+
+-   **Password Policies**: We will not implement password policies but should enforce a configurable minimum password length.
+
+-   **Login Role Determination**: We should display whether a role has a password and can be used as a login role in `mz_roles`. See [pg_roles](https://www.postgresql.org/docs/9.1/view-pg-roles.html).
+
+### 3. Hashed Password Storage
+
+-   **Secure Hashing**: For future compatibility with SCRAM, passwords should be hashed using `scram-sha-256` encryption. The RFC for [SCRAM](https://datatracker.ietf.org/doc/html/rfc5802) dictates an iteration count of at least 4096. However, the RFC for SCRAM-SHA-256 dictates that "[the hash iteration-count should be such that a modern machine will take 0.1 seconds to perform the complete algorithm](https://www.rfc-editor.org/rfc/rfc7677)". This also dictates at least 4096 with a recommendation of 15000. We should aim for 15000 if it does not negatively impact connection times or load.
```
jasonhernandez, line 71: This RFC is 10 years old.
I ran a quick benchmark on two computers:
t4g.nano - 200,000 iterations in 0.100s
Macbook Pro M3 Max - 450,000 iterations in 0.102s

I recommend we plan on a minimum of 200,000 iterations for now. I would also like us to store the iteration count in a column in the database. If practical, it would be very nice to randomly generate an iteration count for each password, perhaps in the range of 200,000-210,000 iterations. This would significantly reduce the probability of birthday attacks and the usefulness of rainbow tables. We will want to track the iteration count for passwords regardless, because we may need to increase the number of iterations as computers get faster.

I will try and do some more research and might adjust my recommended iteration count.

doc/developer/design/20250321_password_authentication.md
```diff
@@ -0,0 +1,176 @@
...
+
+    - **Reasons Not Chosen**: This solution is more involved, requiring significant additional work to secure, develop, and maintain. Additionally, it does not align with the Postgres.
+
+2. **Password Authentication in Ingress Layer**: Using a front-end proxy like PGbouncer or NGINX to handle password authentication before traffic hits Materialize.
+
+    - **Reasons Not Chosen**: This approach requires significant customization and doesn't solve the problem within the Materialize database itself. It also still necessitates internal authentication for role management.
+
+3. **Store Passwords in K8s Secrets**: It would be viable to store passwords in Kubernetes secrets.
+
+-   **Reasons Not Chosen**: This may create an untenable number of Kubernetes secrets, and may cause issues with passing security reviews.
```
jasonhernandez, line 162: Using Kubernetes secrets also creates some additional complexity for backup / restore / migration. There's an advantage to keeping the same mental model as Postgres where possible.

doc/developer/design/20250321_password_authentication.md
```diff
@@ -0,0 +1,176 @@
...
+
+-   **Reasons I still might prefer this**: It does have a number of benefits, including slightly better security via enforcement of random strong passwords, and a better rotation story. It's also possible we could have different app password protocols/versions that allow users to manage their passwords entirely through secrets directly or something like Hashicorp vault.
+
+## Open questions
+
+1. **Where exactly are passwords stored**: While it makes sense to store passwords in the catalog, I have not yet identified exactly where they should be stored. Should it go alongside `mz_roles`, or in a new `mz_auth_id` table. The latter makes sense as it could store password metadata, and might let us store multiple passwords down the road.
+
+2. **Audit Logging and Brute Force Detection**: Do we need to log all failed password attempts and the IPs of those attempts? Do we need to take action to lock an IP or User when an attempt threshold is met?
+
+3. **Password Strength Validation**: Should Materialize include built-in password strength validation, or should it be handled outside the system (e.g., via Kubernetes or other security tools)?
```
jasonhernandez, line 176: I'm not sure if this is necessary yet. If we included the requirements, we would want to include an override setting for developer mode.
If we were to implement this, I would suggest copying the behavior of Postgres passwordcheck without cracklib + requiring an entropy score of 3+ from zxcvbn. 
https://crates.io/crates/zxcvbn

Docs on Postgres passwordcheck
https://www.postgresql.org/docs/current/passwordcheck.html
https://github.com/postgres/postgres/blob/master/contrib/passwordcheck/passwordcheck.c

jasonhernandez (changes requested):
My "hard requirements" 
1. Set the minimum iteration count >=200,000.
2. Do not allow lockouts by default.
3. Log successful and failed login attempts by default.

I've made some additional comments and recommendations for discussion / consideration. Thanks for writing this all up and the consideration in this design.

doc/developer/design/20250321_password_authentication.md
jasonhernandez: This (expired) draft RFC from 2021 recommends a minimum of 310,000 iterations.
https://www.ietf.org/archive/id/draft-ietf-kitten-password-storage-07.txt

We should also specify a length on salts. We should use at least 32 bytes of cryptographically random data for salts.

doc/developer/design/20250321_password_authentication.md
```diff
@@ -0,0 +1,128 @@
...
+1. The user sends a username and password to the Materialize server.
+2. The protocol layer will communicate with the coordinator layer to validate the password. (via the adapter Client)
+   a. The coordinator layer will check the password against the stored hash. If the password is correct, the user will be authenticated.
+   b. If the password is incorrect, the user will be denied access.
+
+3. **OPTIONAL: Book-Keeping**: Login and failed login requests should be tracked. Metrics should be created to monitor potential security events.
+
+**Password Rotation or Expiration**: While not currently in scope, the design should allow for additions in the future, such as password rotation and expiration.
+
+**Abuse Detection**: Action should be taken action when brute force attempts are detected, either by locking the user or blocking the requesting client IP.
```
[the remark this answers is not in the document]
jubrad: I think the niftiest way to do fail2ban would be to have a table that tracks login attempts, success, fail, ip, username... we can log for now and keep metrics or tables for v2

doc/developer/design/20250321_password_authentication.md
ptravers: we also will need rate limiting if we want to defend against credential stuffing.

doc/developer/design/20250321_password_authentication.md
```diff
@@ -0,0 +1,183 @@
...
+3. **Secure Password Storage**: Implement hashed salted password storage using compatible with `scram-sha-256`.
+4. **Bootstrapping System Users**: Provide a mechanism for bootstrapping the password and enabling login for system users.
+5. **HTTP Session Management**: Environmentd's HTTP layer needs to handle session management via cookies and internal session storage.
+6. **Enhanced Superuser Management**: Support assigning / unassigning superuser for roles.
+
+## Alternatives
+
+1. **Authd Service**: This would involve creating a separate authentication service to handle authentication. This approach is more complex and introduces security concerns, as it would require developing secure endpoints and integrating a user management system that could be prone to additional complexity and security risks.
+
+    - **Reasons Not Chosen**: This solution is more involved, requiring significant additional work to secure, develop, and maintain. Additionally, it does not align with the Postgres.
```
ptravers, line 157: if this is the sentence then in what ways does it not align with postgres?

```suggestion
    - **Reasons Not Chosen**: This solution is more involved, requiring significant additional work to secure, develop, and maintain. Additionally, it does not align with Postgres.
```

ptravers:
should this include a threat analysis?

jubrad:
> should this include a threat analysis?

This will be a self-managed only feature which will be scoped different for security audits. I'll leave it up to @jasonhernandez to sort out how we deal with this.

jasonhernandez (approved):
Thanks for setting a 400k iteration count as the default. This looks like a good v1 design to ship.

PR #32010, opened 2025-03-25
Record wallclock lag histogram

teskje:
This PR adds a new append-only storage-managed collection, `mz_wallclock_global_lag_histogram`, and makes the controllers start populating it.

The schema is:
* `period_start` timestamp
* `period_end` timestamp
* `object_id` text
* `lag_seconds` uint8
* `labels` jsonb
* `count` uint8

Histograms are per object and per time period, as specified by `object_id` and `period_*`, respectively. The time period is configurable through a config flag and defaults to one day.

The lag values are bucketed as powers-of-2 seconds. I decided to store seconds instead of milliseconds to (a) avoid the illusion of millisecond accuracy and (b) significantly reduce the number of buckets. I also omitted adding additional precision controls for now to reduce the complexity in this first version and because I suspect that powers of 2 might be sufficient. We can always iterate if it turns out we need more precision. The bucket values are upper bounds, mirroring what we also use for compute introspection histograms.

`labels` currently includes a single optional `workload_class` label, reflecting the workload class of the cluster maintaining the respective object, according to `mz_cluster_workload_classes`. I included this label specifically because it provides a simple way to filter out most of the noise when building a metric focused on production experience.

Measurements are collected every second but only written once per a configurable refresh interval (one minute by default). This is to reduce the load on persist.

There is also a config flag to disable wallclock lag histogram recording entirely. This exists only to derisk the feature rollout and the expectation is that we'll remove it soon.

### Motivation

See [proposal in Notion](https://www.notion.so/materialize/Ship-internal-only-freshness-dashboard-for-business-reviews-and-daily-check-ins-17513f48d37b802485b9de9a74d15817?pvs=4#1bc13f48d37b802e9163fae6660adad8).

### Tips for reviewer

Individual commits are meaningful.

### Checklist


src/controller-types/src/dyncfgs.rs
```diff
@@ -32,6 +32,18 @@ pub const WALLCLOCK_LAG_HISTORY_REFRESH_INTERVAL: Config<Duration> = Config::new
...
+    Duration::from_secs(60),
+    "The interval at which to refresh `WallclockLagHistogram` introspection.",
+);
+
+pub const WALLCLOCK_LAG_HISTOGRAM_PERIOD_LENGTH: Config<Duration> = Config::new(
+    "wallclock_lag_histogram_period_length",
+    Duration::from_secs(24 * 60 * 60),
+    "The period length of histograms in `WallclockLagHistogram` introspection.",
+);
+
```
antiguru: Maybe interval would be more specific than length?

src/compute-client/src/controller/instance.rs
```diff
@@ -608,26 +612,34 @@ impl<T: ComputeControllerTimestamp> Instance<T> {
...
 
         let mut histogram_updates = Vec::new();
+        let mut row_buf = Row::default();
         for (collection_id, collection) in &mut self.collections {
             if let Some(stash) = &mut collection.wallclock_lag_histogram_stash {
                 let lag = collection.shared.lock_write_frontier(|f| frontier_lag(f));
                 let bucket = lag.as_secs().next_power_of_two();
 
-                let key = (histogram_period, bucket);
+                let key = (histogram_period, bucket, histogram_labels.clone());
```
antiguru, line 627: This clone isn't nice, but I can't think of an easy way to avoid it.

antiguru (commented):
I think this looks good, let me know when it's ready. I left some comments inline.

I'm wondering if we need both `period_start` and `period_end`. Ideally, the end of the previous period corresponds to the start of the current period.

src/controller-types/src/dyncfgs.rs
teskje: SGTM!

teskje:
> I'm wondering if we need both `period_start` and `period_end`. Ideally, the end of the previous period corresponds to the start of the current period.

Where it can be useful is when we change the period length/interval. There will be some time where the measurements are glitchy because intervals overlap. Having both the start and end allows us to remove ambiguity in these cases. Though there are no specific plans to actually use the `period_end` value right now it seems like a good idea to include it to future-proof things. I think once we have the period length locked in we can consider removing the `period_end` column.

src/compute-client/src/controller/instance.rs
teskje: The whole handling of labels in `wallclock_lag_histogram_stash` is not great. Label values are fixed to strings right now, which will stop working once we add other types (like bools). Ideally we'd store a `Datum` but it's impossible to store an owned `String` as a `Datum`. I've been considering storing a `DatumMap` instead of the `BTreeMap`, but I don't know how to construct one from its contents. Should we store a `Row`?

teskje:
Should be ready to review now! Things I changed since @antiguru's last look:

* Made the `period_*` columns the first two columns, after @bkirwi's suggestion that this would make persist order them by time.
* Renamed the source to `mz_wallclock_global_lag_histogram_raw` and added a `mz_wallclock_global_lag_histogram` view that lifts the count into a column. We need this for compatibility with the catalog extractor, which can't run transformations.
* Renamed the `wallclock_lag_histogram_period_length` dyncfg to `wallclock_lag_histogram_period_interval`.
* Added an `enable_wallclock_lag_histogram_collection` dyncfg to make it possible to switch of histogram collection in a pinch.
* Added tests.

teskje:
TFTR!

PR #32015, opened 2025-03-26
Display error if user tries to create subsources for single-output load generators

martykulma:
Return appropriate error if user tries to create a single-output load generator with subsources (e.g. `FOR ALL TABLES`, `FOR TABLES ...`, `FOR SCHEMAS ...`)

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9028

### Checklist

### DIFF
--- a/test/testdrive/load-generator.td
+++ b/test/testdrive/load-generator.td
@@ -52,31 +52,31 @@ exact:COUNTER load generators do not support SCALE FACTOR values
 
 # Error if trying to create with subsources
 ! CREATE SOURCE g FROM LOAD GENERATOR COUNTER FOR ALL TABLES;
-contains:FOR ALL TABLES
+contains:FOR ALL TABLES is only valid for multi-output sources
 
 ! CREATE SOURCE g FROM LOAD GENERATOR CLOCK FOR ALL TABLES;
-contains:FOR ALL TABLES
+contains:FOR ALL TABLES is only valid for multi-output sources
 
 ! CREATE SOURCE g FROM LOAD GENERATOR DATUMS FOR ALL TABLES;
-contains:FOR ALL TABLES
+contains:FOR ALL TABLES is only valid for multi-output sources
 
 ! CREATE SOURCE g FROM LOAD GENERATOR COUNTER FOR TABLES ("foo");
-contains:FOR TABLES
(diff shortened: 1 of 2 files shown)


test/testdrive-old-kafka-src-syntax/load-generator-key-value.td
```diff
@@ -0,0 +1,298 @@
+# Copyright Materialize, Inc. and contributors. All rights reserved.
```
aljoscha, line 1: This one is longer because it first adds test coverage that wasn't there before, right?

aljoscha (approved):
This does what it says on the tin, and the code is good!

@petrosagg might want to look at this because he's currently in the business of moving all our sources to only allow the `CREATE TABLE FROM SOURCE ...` syntax. Which I'm thinking currently doesn't work with these load generator sources, maybe?

petrosagg:
Disallowing this syntax in the current version of the product is inline with the migration so no concerns from me

test/testdrive-old-kafka-src-syntax/load-generator-key-value.td
def-: It doesn't hurt much either way, but no need to add a separate file here if there is nothing interesting being exercised. All of these `old-kafka-src-syntax` tests was supposed to be long removed, but then the new `CREATE TABLE ... FROM SOURCE` syntax work was paused for a while.

test/testdrive/load-generator.td
```diff
@@ -50,6 +50,34 @@ exact:COUNTER load generators do not support SCALE FACTOR values
...
+contains:FOR TABLES
+
+! CREATE SOURCE g FROM LOAD GENERATOR COUNTER FOR SCHEMAS ("foo");
+contains:FOR SCHEMAS
+
+! CREATE SOURCE g FROM LOAD GENERATOR CLOCK FOR SCHEMAS ("foo");
+contains:FOR SCHEMAS
+
+! CREATE SOURCE g FROM LOAD GENERATOR DATUMS FOR SCHEMAS ("foo");
+contains:FOR SCHEMAS
```
def-, line 79: could be more verbose with what the error message is

def- (approved):

PR #32017, opened 2025-03-26
compute: handle future epochs in channel adapter

teskje:
Previously the compute command channel adapter assumed that it is impossible to receive responses with epoch values greater than the channel adapter's current epoch. This assumption is wrong because the Timely cluster-side of workers learns about new connection epochs by observing commands from worker 0, which can be ahead of other workers' epochs.

This PR removes the assumption and instead makes the channel adapter handle responses with future epochs by stashing them until sufficiently many client reconnects have been observed.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9124

### Tips for reviewer

I've tried to write a regression test using mzcompose and toxiproxy, spawning a multi-process cluster and resetting its controller connection in a loop. Didn't manage to reproduce the race that leads to the "epoch from the future" panic though.

### Checklist

### DIFF
--- a/src/compute/src/server.rs
+++ b/src/compute/src/server.rs
@@ -747,28 +747,50 @@ fn spawn_channel_adapter(
             // for previous clients.
             let mut epoch = 0;
 
+            // It's possible that we receive responses with epochs from the future: Worker 0 might
+            // have increased its epoch before us and broadcasted it to our Timely cluster. When we
+            // receive a response with a future epoch, we need to wait with forwarding it until we
+            // have increased our own epoch sufficiently (by observing new client connections). We
+            // need to stash the response in the meantime.
+            let mut stashed_response = None;
+
             while let Ok((command_rx, response_tx)) = client_rx.recv() {
                 epoch += 1;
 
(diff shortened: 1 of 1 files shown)


src/compute/src/server.rs
```diff
@@ -747,28 +747,50 @@ fn spawn_channel_adapter(
...
+                        // Response for a previous connection; discard it.
+                        continue;
+                    } else if resp_epoch > epoch {
+                        // Response for a future connection; stash it and reconnect.
+                        stashed_response = Some((resp, resp_epoch));
+                        break;
+                    } else {
+                        // Response for the current connection; forward it.
+                        if response_tx.send(resp).is_err() {
+                            break;
```
teskje, line 793: This would all be much nicer if we could have a `crossbeam_channel::Receiver` wrapper with a builtin stash that you could `push_front` to. Unfortunately `crossbeam_channel::select` doesn't support other types so I had to move the stash outside and ended up with this somewhat convoluted control flow.

antiguru (approved):
Looks good, thank you!

teskje:
TFTR!

PR #32030, opened 2025-03-27
adapter: properly retry when LD client initialization fails

teskje:
Previously the code initializing the LD client would correctly await `initialized_async` to see if the initialization succeeded. However, if it didn't succeed it would simply wait a bit and then call `initialized_async` again. Reading the LD server sdk code, there is no reason to assume that the call would return something different if repeated.

This PR changes the logic to call `start_with_default_executor` again when `initialized_async` reports failure, to attempt a new initialization. It also moves to the mz-ore `Retry` type, instead of implementing manual retry logic.

### Motivation

If initializing the LD client fails the first time, the code gets stuck forever in a retry loop.

### Tips for reviewer

I only stumbled over this when reading the code and I might be missing something. Please check my work!

### Checklist

### DIFF
--- a/src/dyncfg-launchdarkly/src/lib.rs
+++ b/src/dyncfg-launchdarkly/src/lib.rs
@@ -15,6 +15,7 @@ use launchdarkly_server_sdk as ld;
 use mz_build_info::BuildInfo;
 use mz_dyncfg::{ConfigSet, ConfigUpdates, ConfigVal};
 use mz_ore::cast::CastLossy;
+use mz_ore::retry::Retry;
 use mz_ore::task;
 use tokio::time;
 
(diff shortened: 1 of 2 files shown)


src/dyncfg-launchdarkly/src/lib.rs
```diff
@@ -48,18 +49,27 @@ where
...
+                        Err(())
+                    }
+                })
+                .await
+                .expect("retries forever")
         };
+
         if tokio::time::timeout(config_sync_timeout, init)
             .await
             .is_err()
```
teskje, line 75: The timeout error handling here also looks very suspicious to me. We just log an INFO event, but the LD sync might be broken. Shouldn't we at least return an error?

ParkMyCar: +1 it should be an error

ParkMyCar (approved):
Sorry for the delay on this one @teskje, thanks for making the change though!

PR #32045, opened 2025-03-28
Add string_to_array function

ptravers:
add string_to_array function

### Motivation
https://github.com/MaterializeInc/database-issues/issues/7101

### Checklist


ptravers:

src/expr/src/scalar/func.rs
```diff
@@ -2125,6 +2125,72 @@ fn parse_ident<'a>(
...
+    delimiter: &str,
+    null_string: Datum<'a>,
+    temp_storage: &'a RowArena,
+) -> Result<Datum<'a>, EvalError> {
+    let mut row = Row::default();
+    let mut packer = row.packer();
+    let empty_flags = "";
+    let regex = build_regex(delimiter, empty_flags)?;
+
+    let found = mz_regexp::regexp_split_to_array(string, &regex);
```
ParkMyCar, line 2170: The regex here works, but it might be faster and easier to just use [`std::str::split`](https://doc.rust-lang.org/std/primitive.str.html#method.split) any particular reason you used a regex?

src/expr/src/scalar/func.rs
```diff
@@ -2125,6 +2125,72 @@ fn parse_ident<'a>(
...
+    let array_dimensions = [ArrayDimension {
+        lower_bound: 1,
+        length: found.len(),
+    }];
+
+    if !null_string.is_null() {
```
ParkMyCar, line 2176: nit: mind switching these `if cases`, "not null_string is null" is a little confusing :)

src/sql/src/func.rs
```diff
@@ -3161,6 +3161,10 @@ pub static PG_CATALOG_BUILTINS: LazyLock<BTreeMap<&'static str, Func>> = LazyLoc
             }) => String, 3538;
             params!(Bytes, Bytes) => Operation::binary(|_ecx, _l, _r| bail_unsupported!("string_agg on BYTEA")) => Bytes, 3545;
         },
+        "string_to_array" => Scalar {
+            params!(String, Any) => VariadicFunc::StringToArray => ScalarType::Array(Box::new(ScalarType::String)), 3947;
+            params!(String, Any, String) => VariadicFunc::StringToArray => ScalarType::Array(Box::new(ScalarType::String)), 3948;
```
ParkMyCar, line 3166: We match the oids from Postgres, it looks like these should be 394 and 376 respectively ([postgres](https://github.com/postgres/postgres/blob/master/src/include/catalog/pg_proc.dat#L1642-L1647)).

Also, I think the second argument in both of these cases should be `String`? I think even with the type of `String` you should be able to pass `NULL` as an argument.

ParkMyCar (approved):
Overall looks good, but can you address the OIDs and argument types before merging? Want to make sure we adhere to Postgres compatibility.

WOOHOO on your first PR!

src/sql/src/func.rs
ptravers: done! ty!

src/expr/src/scalar/func.rs
ptravers: yep! will do!

ptravers: done

src/expr/src/scalar/func.rs
ggevay: Note that `build_regex` is expensive (see https://github.com/MaterializeInc/database-issues/issues/6199#issuecomment-2764228653 for a measurement), so `std::str::split` would be preferable (if its semantics match).

src/expr/src/scalar/func.rs
```diff
@@ -2125,6 +2125,72 @@ fn parse_ident<'a>(
     })?)
 }
 
+fn string_to_array<'a>(
+    string_datum: Datum<'a>,
+    delimiter: Datum<'a>,
+    null_string: Datum<'a>,
+    temp_storage: &'a RowArena,
+) -> Result<Datum<'a>, EvalError> {
+    let string = string_datum.unwrap_str();
```
ggevay, line 2134: Note that the input might be null. E.g., the following panics:
```
select string_to_array(null, ' ');
```

src/expr/src/scalar/func.rs
```diff
@@ -8102,6 +8175,7 @@ impl VariadicFunc {
                 ScalarType::Array(Box::new(ScalarType::String)).nullable(in_nullable)
             }
             RegexpReplace => ScalarType::String.nullable(in_nullable),
+            StringToArray => ScalarType::Array(Box::new(ScalarType::String)).nullable(true),
```
ggevay, line 8178: I think the output is null if and only if the first argument is null. If this is true, you could give a tighter nullability here by copying the nullability of the first argument. (This would be similar to some of the above lines that use `in_nullable`, but you can't use that directly, because that takes the nullability of all arguments, whereas you'd need only the first argument.)

src/expr/src/scalar/func.rs
ptravers: gotcha, will reimplement into std::str::split

ptravers: done

src/expr/src/scalar/func.rs
ptravers: ah neat, thanks! what difference does making the nullability more stringent like that make to the behaviour of the database?

src/expr/src/scalar/func.rs
```diff
@@ -8102,6 +8190,9 @@ impl VariadicFunc {
                 ScalarType::Array(Box::new(ScalarType::String)).nullable(in_nullable)
             }
             RegexpReplace => ScalarType::String.nullable(in_nullable),
+            StringToArray => {
+                ScalarType::Array(Box::new(ScalarType::String)).nullable(input_types[0].nullable)
```
ptravers, line 8194: do I need to protect against empty input_types here?

test/sqllogictest/string.slt
```diff
@@ -1757,3 +1757,80 @@ select reverse(word) from words;
...
+----
+{}
+
+# string_to_array - whitespace
+query T
+select string_to_array(' ', '') from string_to_array_words;
+----
+{" "}
+
+# string_to_array - whitespace
```
ggevay, line 1830: (comment copy-pasted from above)

src/expr/src/scalar/func.rs
ggevay: No need, the length of `input_types` should correspond to the length of the argument list of a valid function call. You could add this fact as a doc comment on top of `output_type`.

src/expr/src/scalar/func.rs
ggevay: - `... IS NULL`/`IS NOT NULL` [are rewritten to](https://github.com/MaterializeInc/materialize/blob/b05bdef24d17c53fa8b5094b629ac3249e1dbfdf/src/expr/src/scalar.rs#L763) `false`/`true`. (Or not introduced in other cases, e.g., [here](https://github.com/MaterializeInc/materialize/blob/7893991ee11dd9601bce5592e0c73bcb1f42f365/src/transform/src/predicate_pushdown.rs#L243).) Generally, we have a lot of `IS NOT NULL` checks, because `MirRelationExpr::Join` matches up nulls, which doesn't correspond to SQL join behavior for nulls, so we introduce a lot of null checks just below joins to match the SQL join behavior. So, eliminating these checks usually means slightly less CPU work due to simply not performing null checks. Also, in some cases it can have more dramatic consequences, for example:
  - Certain subquery simplifications (`try_simplify_quantified_comparisons`) are possible only for non-nullable columns.
  - `NOT IN` subqueries choke on nullable inputs: https://github.com/MaterializeInc/database-issues/issues/382#issuecomment-2368827498
  - https://github.com/MaterializeInc/database-issues/issues/8396
- `coalesce` call argument lists [are truncated](https://github.com/MaterializeInc/materialize/blob/b05bdef24d17c53fa8b5094b629ac3249e1dbfdf/src/expr/src/scalar.rs#L1176) after the first non-nullable argument.

src/expr/src/scalar/func.rs
ptravers: great, thanks!

src/expr/src/scalar/func.rs
ptravers: ah thanks so much! that makes sense

doc/user/data/sql_funcs.yml
```diff
@@ -473,6 +473,15 @@
...
 
+  - signature: 'string_to_array(s: str, delimiter: str, null_string: str) -> [str]'
+    description: |
+      Splits the string at occurrences of delimiter and forms the resulting fields into a text array.
+      If delimiter is NULL, each character in the string will become a separate element in the array.
+      If delimiter is an empty string, then the string is treated as a single field. If null_string is supplied
+      and is not NULL, fields matching that string are replaced by NULL.
+
+      string_to_array('xx~~yy~~zz', '~~', 'yy') → {xx,NULL,zz}
+
```
kay-kim, line 484: Maybe:

```suggestion
      Splits the string at occurrences of delimiter and returns a text array of
      the split segments. 
      
      If `delimiter` is NULL, each character in the string will become a
      separate element in the array.
      
      If `delimiter` is an empty string, then the string is treated as a single
      field. 
      
      If `null_string` is supplied and is not NULL, fields matching that string
      are replaced by NULL.

      For example: `string_to_array('xx~~yy~~zz', '~~', 'yy')` → `{xx,NULL,zz}`
```

doc/user/data/sql_funcs.yml
```diff
@@ -471,7 +471,22 @@
...
+      If `delimiter` is NULL, each character in the string will become a
+      separate element in the array.
+      
+      If `delimiter` is an empty string, then the string is treated as a single
+      field. 
+      
+      If `null_string` is supplied and is not NULL, fields matching that string
+      are replaced by NULL.
+
+      For example: `string_to_array('xx~~yy~~zz', '~~', 'yy')` → `{xx,NULL,zz}`
```
kay-kim, line 489: Heh ... sorry for the whitespaces 😄 
```suggestion
    description: |
      Reverse the characters in `s`.

      Splits the string at occurrences of delimiter and returns a text array of
      the split segments.

      If `delimiter` is NULL, each character in the string will become a
      separate element in the array.

      If `delimiter` is an empty string, then the string is treated as a single
      field.

      If `null_string` is supplied and is not NULL, fields matching that string
      are replaced by NULL.

      For example: `string_to_array('xx~~yy~~zz', '~~', 'yy')` → `{xx,NULL,zz}`
```

PR #32052, opened 2025-03-31
Improve interface of debug tool

SangJunBak:

Sorry the first commit is so large. I would've broken the first commit more granularly, but it was hard to given all changes were needed in order to make it functional. Most of it is moving files and refactoring functions to structs w/ methods. I've tried annotating where i've acc made changes. 

### Motivation

### Checklist


src/self-managed-debug/src/kubectl_port_forwarder.rs
```diff
@@ -0,0 +1,167 @@
...
+pub struct KubectlPortForwarder {
+    pub namespace: String,
+    pub service_name: String,
+    pub target_port: i32,
+    pub local_address: String,
+    pub local_port: i32,
+    pub context: Option<String>,
+}
+
+impl KubectlPortForwarder {
```
SangJunBak, line 38: This is pretty much copied from the port_forwarding methods from system_catalog_dumper. Only thing that's changed is `SqlPortforwardingInfo` -> `KubetlPortForwarder`, converting the functions to methods, and everything else I mentioned in the commit message (i.e. getting rid of certain overrides).

src/self-managed-debug/src/main.rs
```diff
@@ -8,83 +8,135 @@
...
+    /// An example URL is `postgres://root@localhost:6875/materialize?sslmode=disable`.
+    // TODO(debug_tool3): Allow users to specify the pgconfig via separate variables
+    #[clap(long, required = true, value_parser = validate_pg_connection_string)]
+    url: String,
+    /// If true, the tool will not dump the system catalog.
+    #[clap(long, default_value = "false")]
+    skip_system_catalog_dump: bool,
+}
+
+pub trait ContainerDumper {
```
SangJunBak, line 105: Implemented this as a unified interface for both dumpers

src/mz-debug/src/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+use std::fs::{create_dir_all, remove_dir_all, File};
+use std::io::{copy, BufWriter};
+use std::path::PathBuf;
+use std::str::FromStr;
+
+use chrono::{DateTime, Utc};
+use zip::write::SimpleFileOptions;
+use zip::ZipWriter;
+
+static DEBUG_FOLDER: &str = "mz-debug";
```
alex-hunt-materialize, line 25: You are naming the debug directory the same thing as your binary. If they are in the same directory, they will conflict.

src/mz-debug/src/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+
+pub fn format_base_path(date_time: DateTime<Utc>) -> PathBuf {
+    let path = PathBuf::from(DEBUG_FOLDER).join(date_time.format("%Y-%m-%dT%H:%MZ").to_string());
```
alex-hunt-materialize, line 28: Can we instead just put this directly in the current directory, named with an `mz-debug` prefix?

Why do we need to nest this in that DEBUG_FOLDER?

src/mz-debug/src/utils.rs
```diff
@@ -0,0 +1,73 @@
...
+        let path = entry.path();
+
+        if path.is_file() {
+            zip_writer.start_file(path.to_string_lossy(), SimpleFileOptions::default())?;
+            let mut file = File::open(path)?;
+            copy(&mut file, &mut zip_writer)?;
+        }
+    }
+
+    remove_dir_all(&DEBUG_FOLDER)?;
```
alex-hunt-materialize, line 69: Nitpick, it is unintuitive to me that zipping a directory would delete that directory.

src/mz-debug/src/utils.rs
SangJunBak: Good call!

src/mz-debug/src/utils.rs
SangJunBak: Ah are you referring to all files generated or just the stdout/stderr file? Mainly because I want a centralized location to zip everything such that the user can send it to me!

src/mz-debug/src/main.rs
```diff
@@ -0,0 +1,286 @@
...
+    /// An example URL is `postgres://root@localhost:6875/materialize?sslmode=disable`.
+    // TODO(debug_tool3): Allow users to specify the pgconfig via separate variables
+    #[clap(long, required = true, value_parser = validate_pg_connection_string)]
+    url: String,
+    /// The name of the zip file to create. Should end with `.zip`.
+    #[clap(long, default_value = "mz-debug.zip")]
+    zip_file_name: String,
+    /// If true, the tool will not dump the system catalog.
+    #[clap(long, default_value = "false")]
+    skip_system_catalog_dump: bool,
```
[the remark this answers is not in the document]
SangJunBak: Ah that makes sense!

src/mz-debug/src/main.rs
```diff
@@ -0,0 +1,286 @@
...
+#[derive(Parser, Debug, Clone)]
+#[clap(name = "mz-debug", next_line_help = true, version = VERSION.as_str())]
+pub struct Args {
+    #[clap(subcommand)]
+    debug_mode: DebugMode,
+    /// The URL of the SQL connection used to dump the system catalog.
+    /// An example URL is `postgres://root@localhost:6875/materialize?sslmode=disable`.
+    // TODO(debug_tool3): Allow users to specify the pgconfig via separate variables
+    #[clap(long, required = true, value_parser = validate_pg_connection_string)]
+    url: String,
```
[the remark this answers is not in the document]
SangJunBak: This is for Materialize! And yeah good point, I'll give it a more descriptive name and make it more clear in the documentation~

src/self-managed-debug/src/docker_dumper.rs
```diff
@@ -24,17 +24,17 @@ use mz_ore::retry::{self, RetryResult};
...
-pub struct DockerResourceDumper {
+pub struct DockerDumper {
     container_id: String,
     directory_path: PathBuf,
 }
 
-impl DockerResourceDumper {
-    pub fn new(context: &Context, container_id: String) -> Self {
+impl<'n> DockerDumper {
+    pub fn new(context: &'n Context, container_id: String) -> Self {
```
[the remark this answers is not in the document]
SangJunBak: oh good point! I think initially I kept it in because we stored the context as part of `DockerDumper` but given we don't need it anymore, it can definitely be elided

src/mz-debug/src/kubectl_port_forwarder.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+pub async fn create_kubectl_port_forwarder(
+    client: &Client,
+    args: &SelfManagedDebugMode,
+) -> Result<KubectlPortForwarder, anyhow::Error> {
+    for namespace in &args.k8s_namespaces {
+        let services: Api<Service> = Api::namespaced(client.clone(), namespace);
+        let services = services.list(&Default::default()).await?;
```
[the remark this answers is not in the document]
SangJunBak: Good point! I can filter on the name in the `filter_map` but I don't think `services.list` explicitly allows regex matching (i.e. `LIKE %environmentd%`)

src/mz-debug/src/utils.rs
alex-hunt-materialize: I meant, why have it as `mz-debug/%Y-%m-%dT%H:%MZ` as opposed to something like `mz-debug_%Y-%m-%dT%H:%MZ`? If everything is going inside that dated folder, why have a base directory that just contains the date directory?

src/mz-debug/src/kubectl_port_forwarder.rs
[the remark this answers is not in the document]
alex-hunt-materialize: You can filter by label at least.

Even better, you can probably construct the exact name of the service you want if you first get the Materialize CR, then get the resource ID from its status. The service for environmentd will then be called `mz<resource_id>-environmentd`.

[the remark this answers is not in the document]
SangJunBak: Ohh I didn't know we could query the resource ID, that's super helpful! However, I do feel like just fetching all the services and filtering client side gives us the best bang for our buck. Especially since we're already filtering on user-provided namespaces. But are there any other consequences of fetching all services in a namespace other than over-fetching? My reasoning is fetching the Materialize CR then getting the resource ID from the status leads to more points of failure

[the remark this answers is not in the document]
SangJunBak: Also filtering by label doesn't really help unless we get the resource ID given the labels of each service looks like:
```
Labels:          materialize.cloud/mz-resource-id=audbbieima                                                                                                              ││                    materialize.cloud/organization-name=12345678-1234-1234-1234-123456789012                                                                                 ││                    materialize.cloud/organization-namespace=materialize-environment
```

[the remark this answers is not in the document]
alex-hunt-materialize: I'm not worried about over fetching.

>My reasoning is fetching the Materialize CR then getting the resource ID from the status leads to more points of failure

This is fair. At the least, you can select for things that have the labels we use, even if you don't have the exact value. You can then do more specific filtering in the `filter_map`.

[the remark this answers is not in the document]
SangJunBak: ooh i see, by label key. That makes sense!

src/mz-debug/src/main.rs
```diff
@@ -89,17 +91,14 @@ pub enum DebugMode {
...
     // TODO(debug_tool3): Allow users to specify the pgconfig via separate variables
     #[clap(long, required = true, value_parser = validate_pg_connection_string)]
-    url: String,
-    /// The name of the zip file to create. Should end with `.zip`.
-    #[clap(long, default_value = "mz-debug.zip")]
-    zip_file_name: String,
-    /// If true, the tool will not dump the system catalog.
-    #[clap(long, default_value = "false")]
-    skip_system_catalog_dump: bool,
+    mz_connection_url: String,
```
alex-hunt-materialize, line 98: This is much clearer! I wonder if this should have a default value based on the port forward? That's probably hard to do in clap, but can be done after parsing if you make this an `Option<String>`.

alex-hunt-materialize (commented):
This is much nicer!

src/mz-debug/src/main.rs
SangJunBak: That makes sense to me! I initially made it required since I think the emulator should have it explicitly required, but there's no reason why we can't make it required for the emulator and optional for the self-managed option.

src/mz-debug/src/kubectl_port_forwarder.rs
```diff
@@ -130,6 +130,11 @@ pub async fn create_kubectl_port_forwarder(
             .filter_map(|service| {
                 let spec = service.spec.as_ref()?;
                 let service_name = service.metadata.name.as_ref()?;
+                if !service_name.to_lowercase().contains("environmentd")
+                    && !service_name.to_lowercase().contains("balancerd")
```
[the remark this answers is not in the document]
SangJunBak: Ah fair haha!

src/mz-debug/src/main.rs
```diff
@@ -43,11 +45,11 @@ static ENV_FILTER: &str = "mz_debug=info";
 #[derive(Parser, Debug, Clone)]
 pub struct SelfManagedDebugMode {
     // === Kubernetes options. ===
-    /// If true, the tool will not dump the Kubernetes cluster and will not port-forward the external SQL port.
-    #[clap(long, default_value = "false")]
-    skip_k8s_dump: bool,
+    /// If true, the tool will dump debug information in Kubernetes cluster such as logs, pod describes, etc.
+    #[clap(long, default_value = "true")]
+    k8s_dump_debug: bool,
```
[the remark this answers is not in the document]
SangJunBak: ohh because it's a boolean and by default, decides based on the existence of the flag! Good callout

src/mz-debug/src/kubectl_port_forwarder.rs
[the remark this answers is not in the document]
alex-hunt-materialize: I just had another thought here. We could have multiple Materialize CRs in the same namespace. Returning the first balancerd service we find will only be for one of them.

src/mz-debug/src/kubectl_port_forwarder.rs
```diff
@@ -0,0 +1,182 @@
...
+                            service_name: service_name.to_owned(),
+                            target_port: port_info.port,
+                            local_address: args.port_forward_local_address.clone(),
+                            local_port: args.port_forward_local_port,
+                        });
+                    }
+                }
+
+                None
+            });
```
alex-hunt-materialize, line 163: There may be multiple Materialize CRs in a single namespace. This will only find and forward one of them.

alex-hunt-materialize (approved):
This looks great, but will fail to forward and connect to multiple Materialize CRs in the same namespace. I think that might be easier to fix in a follow up PR.

Additionally, you are failing clippy lints for an unneeded reference to a reference.

src/mz-debug/src/kubectl_port_forwarder.rs
SangJunBak: Gah you're right. I'll handle in a follow-up PR, but the way I'm thinking we can handle this is having an optional flag to specify the Materialize CR they want to port forward from?

SangJunBak: That being said, I recall us discussing how it's uncommon for customers to have multiple Materialize CRs(maybe unless they're doing an upgrade). So this shouldn't be common, right?

PR #27300, opened 2024-05-27
sql: optimize regexp_replace for static regexes

sdht0:
Mirroring the `regexp_match` sql function, this PR pre-processes static regexes for `regexp_replace`.

Fixes #22172 

### Motivation

The `Regex` object is now constructed from their string representation only once when possible, instead of constructing the same object repeatedly across multiple rows.

### Tips for reviewer

- The implementation follows the style of `regexp_match`, except we need a binary function here.
- Because of the `g` flag, we need to carry around an extra integer. I created a new proto object `ProtoRegexReplace` to be able to represent this.
- I've extended the tests to ensure the fast and slow paths are taken appropriately. 
- [x] I've tried to guess the correct output for the functions in `impl BinaryFunc`. These should be verified properly though, as I don't yet understand the exact implications of these functions.
- [x] The code for computing `n` is repeated in two places and should be hoisted out to a helper function. What's the best file for that?

### Checklist

- [x] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] This PR includes the following [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note):
  -

PR #32053, opened 2025-03-31
Simplify regexp replace handling

antiguru:
Simplify regexp replace handling by using a literal error

Instead of encapsulating a result, replace the function call by a literal error. The hope is that this is semantically equivalent to the existing implementation.

Part of MaterializeInc/database-issues#9138

### Checklist

### DIFF
--- a/src/expr/src/scalar.rs
+++ b/src/expr/src/scalar.rs
@@ -1241,10 +1241,16 @@ impl MirScalarExpr {
                                         BinaryFunc::RegexpReplace { regex, limit },
                                     )
                                 }
-                                Err(err) => MirScalarExpr::literal(
-                                    Err(err),
-                                    e.typ(column_types).scalar_type,
-                                ),
+                                Err(err) => {
+                                    let mut exprs = mem::take(exprs);
+                                    let source = exprs.swap_remove(0);
+                                    let scalar_type = e.typ(column_types).scalar_type;
+                                    // We need to return `NULL` on `NULL` input, and error otherwise.
+                                    source.call_is_null().if_then_else(
+                                        MirScalarExpr::literal_null(scalar_type.clone()),
+                                        MirScalarExpr::literal(Err(err), scalar_type),
+                                    )
+                                }
                             };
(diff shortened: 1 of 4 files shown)


ggevay:
Unfortunately, there was a reason for this complication: https://github.com/MaterializeInc/materialize/pull/27300#issuecomment-2135451577
(which totally should have been recorded in a code comment on the field)

src/expr/src/scalar.rs
```diff
@@ -1231,15 +1231,29 @@ impl MirScalarExpr {
...
+                                    let source = exprs.swap_remove(0);
+                                    let scalar_type = e.typ(column_types).scalar_type;
+                                    // We need to return `NULL` on `NULL` input, and error otherwise.
+                                    source
+                                        .call_is_null()
+                                        .if_then_else(
+                                            MirScalarExpr::literal_null(scalar_type.clone()),
+                                            MirScalarExpr::literal(Err(err), scalar_type),
+                                        )
+                                }
```
antiguru, line 1255: @ggevay I changed the error path to return `null` on `null` input, and error otherwise. I think this should match the current behavior!

src/expr/src/scalar.rs
```diff
@@ -1231,15 +1231,27 @@ impl MirScalarExpr {
...
+                                    source.call_binary(
+                                        replacement,
+                                        BinaryFunc::RegexpReplace { regex, limit },
+                                    )
+                                }
+                                Err(err) => {
+                                    let mut exprs = mem::take(exprs);
+                                    let source = exprs.swap_remove(0);
+                                    let scalar_type = e.typ(column_types).scalar_type;
+                                    // We need to return `NULL` on `NULL` input, and error otherwise.
```
ggevay, line 1248: Maybe extend the comment a bit. Something like:
"We'd like the same behaviors between the regex argument being dynamic or a literal. Therefore, we need to return `NULL` on `NULL` input, and error otherwise, because this is what the original, dynamic function does."

ggevay (approved):
Good idea!

antiguru:
Thanks for the review!

PR #32055, opened 2025-03-31
Turn jsonb stringify functions into separate functions

antiguru:
Some of the jsonb functions can be used in contexts where the output is either jsonb or strings. Previously, we'd encode this as part of the function call by supplying a `bool` field. This is an abstraction leak and instead it would be better to have separate functions.

This change absorbs the `stringify` parameter into different functions. No behavior change expected.

### Checklist

### DIFF
--- a/src/buf.yaml
+++ b/src/buf.yaml
@@ -55,6 +55,8 @@ breaking:
     - compute-types/src/sinks.proto
     # reason: Ignore because plans are currently not persisted.
     - expr/src/relation.proto
+    # reason: Ignore because plans are currently not persisted.
+    - expr/src/scalar.proto
     # reason: we very carefully evolve these protobuf definitions
     - persist-client/src/internal/state.proto
     # reason: does currently not require backward-compatibility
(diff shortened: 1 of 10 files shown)


ggevay (approved):
LGTM; the CI failures have straightforward fixes:

The lint is just the protobuf compatibility stuff, also discussed on slack.

The `cargo test` fail is due to some `mzreflect`-based tests. There are several occurrences of the functions that are being modified in the PR. For example
```
(call_binary (jsonb_get_string true) #0 ("created_ms" string))
```
has to be modified to
```
(call_binary (jsonb_get_string_stringify) #0 ("created_ms" string))
```
There are also some mzreflect-based tests that are not complaining, because it seems `mzreflect` just ignores the extra argument, but still should be changed. For example, there is a
```
(call_binary (jsonb_get_int64 false) #1 #0))
```
in `src/expr/tests/testdata/reduce`.
Normally, I'd debug why `mzreflect` is ignoring the extra argument, but `mzreflect` is on its way out, so we can just quickly fix the individual tests. I recommend grepping for `jsonb_get_string true`/`jsonb_get_string false` with all the modified functions to find all relevant tests.

The slt fail in `test/sqllogictest/distinct_arrangements.slt` is because some arrangement names mention `JsonbGetString` where this is part of the arrangement key in some builtin objects. This just needs a `--rewrite-results`.

antiguru:
Thanks for the review!

PR #32058, opened 2025-03-31
doc: Adjust cluster size in Postgres source guide

bobbyiliev:
### Motivation
The Postgres source guide initially suggests a `50cc` size but later on talks about a `200cc` which is not mentioned in the doc.


kay-kim (approved):
thank you!

kay-kim (approved):

PR #32066, opened 2025-04-01
storage: allow sinks on multi-replica clusters

petrosagg:
This PR extends the approach of
https://github.com/MaterializeInc/materialize/pull/31890 to the case of sinks. The handling is identical to that of single replica sources.

### Checklist

### DIFF
--- a/src/storage-controller/src/instance.rs
+++ b/src/storage-controller/src/instance.rs
@@ -31,10 +31,11 @@ use mz_repr::GlobalId;
 use mz_service::client::{GenericClient, Partitioned};
 use mz_service::params::GrpcClientParameters;
 use mz_storage_client::client::{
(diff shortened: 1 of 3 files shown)


src/storage-controller/src/instance.rs
```diff
@@ -411,12 +473,17 @@ where
     ///
     /// If `send_commands` is true, will send commands for newly-scheduled
```
ParkMyCar, line 474: This doc comment only references ingestions, it would be nice to update it to all talk about exports. Specifically I'm thinking a description of how a Kafka ingestion is multi-replica, but a Kafka export is single-replica

ParkMyCar (approved):
Maybe get a second look at the storage-controller bits, but as far as I understand how the controller works, it all LGTM!

src/storage-controller/src/instance.rs
petrosagg: Good catch, wlll fix

PR #32070, opened 2025-04-01
[persist] Clean up the legacy zero-column projection pushdown logic

bkirwi:
The persist source has a special-cased "projection pushdown" optimization... in the case where all columns are projected away (and some other conditions are met) we know that all rows will collapse to a single KVT and we can replace an entire part with its summary diff.

This updates this path to match some recent changes:
- We've moved to a structured / columnar data representation instead of a row-oriented proto encoding. Switch this path over too.
- We've since added a second, more general form of projection pushdown that discards unneeded columns. Reuse that here instead of duplicating its work.
  - This includes removing a special-case zero-column decode optimization, since the standard decode path now only decodes the columns it needs.
 
### Motivation

Followup to https://github.com/MaterializeInc/database-issues/issues/7411

(This is one of the few remaining codepaths where we generate codec-only data.

### Checklist


ParkMyCar (approved):
Looks great!!

PR #32081, opened 2025-04-03
fix string_to_array handle non string delimiter

ptravers:
fixes non string handling in string_to_array. currently causing panics.

### Motivation

errors in buildkite: https://buildkite.com/materialize/nightly/builds/11690#annotation-0195f3b5-b1bd-440b-8c7e-a445219baf63-error

### Checklist

### DIFF
--- a/src/expr/src/scalar/func.rs
+++ b/src/expr/src/scalar/func.rs
@@ -8207,7 +8207,9 @@ impl VariadicFunc {
                 ScalarType::Array(Box::new(ScalarType::String)).nullable(in_nullable)
             }
             RegexpReplace => ScalarType::String.nullable(in_nullable),
-            StringToArray => ScalarType::Array(Box::new(ScalarType::String)).nullable(in_nullable),
+            StringToArray => {
+                ScalarType::Array(Box::new(ScalarType::String)).nullable(input_types[0].nullable)
+            }
         }
     }
 
--- a/src/expr/src/scalar/func.rs
+++ b/src/expr/src/scalar/func.rs
@@ -2151,7 +2151,12 @@ fn string_to_array<'a>(
         return string_to_array_impl(string, split_all_chars_delimiter, null_string, temp_storage);
     }
 
-    let delimiter = delimiter.unwrap_str();
+    let Datum::String(delimiter) = delimiter else {
+        return Err(EvalError::PrettyError(
+            "Invalid delimiter passed. Must be string or null.".into(),
+        ));
+    };
+
     if delimiter.is_empty() {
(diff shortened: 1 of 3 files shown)


src/expr/src/scalar/func.rs
```diff
@@ -2151,7 +2151,12 @@ fn string_to_array<'a>(
         return string_to_array_impl(string, split_all_chars_delimiter, null_string, temp_storage);
     }
 
-    let delimiter = delimiter.unwrap_str();
+    let Datum::String(delimiter) = delimiter else {
+        return Err(EvalError::PrettyError(
+            "Invalid delimiter passed. Must be string or null.".into(),
```
ptravers, line 2156: should this message replicate the postgres error message? it would be less clear

```
ERROR:  function string_to_array(unknown, integer) does not exist
LINE 1: select string_to_array('hsdfk', 0) from string_to_array_word...

HINT: No function matches the given name and argument types. You might need to add explicit type casts.
```

not sure if I can easily do LINE 1 part from this point in the parser

ggevay: Typechecking of parameters is usually done in the `params!` call [in planning](https://github.com/MaterializeInc/materialize/blob/4a986681ee2ad2fcc30c5cef0ae84d4b6c630ae8/src/sql/src/func.rs#L3166). `delimiter` is the second argument, right? Is there a reason why it is `Any` instead of `String`? I've tried changing it to `String`, and all tests passed in `string.slt`.

Btw. [the docs](https://github.com/MaterializeInc/materialize/blob/2aa249cff128c8f847db9defbf8fc7c3c5a22fd7/doc/user/data/sql_funcs.yml#L476) currently seem to be missing a part of the parameter list.

ParkMyCar: If you change the second parameters [here](https://github.com/MaterializeInc/materialize/blob/main/src/sql/src/func.rs#L3165-L3166) from `Any` to `String` then `delimiter` should always be `Datum::String`, I can't remember if we would accept `Null` there though. Do you mind trying this?

Otherwise ideally the error would match Postgres, or get as close as possible, but what you have here works for me :)

ParkMyCar (approved):
Approving to unblock! Would love to see if changing the function signature would work here though

src/expr/src/scalar/func.rs
ggevay: Yes, there are tests for Null delimiter in `string.slt`. `String` would accept Null.

ptravers: fixed!! thanks!

PR #32083, opened 2025-04-03
Procmacro to derive unary/binary function definitions

antiguru:
Similar to the `sqlfunc!` macro, this proc macro derives the helper traits for unary and binary functions.

Here's a binary function for which we generate the `EagerBinaryFunc` trait implementation:
```rust
#[sqlfunc(is_monotone = (true, true), output_type = i16, is_infix_op = true, sqlname="+")]
fn add_int16<'a>(a: Datum<'a>, b: Datum<'a>) -> Result<Datum<'a>, EvalError> {
    a.unwrap_int16()
        .checked_add(b.unwrap_int16())
        .ok_or(EvalError::NumericFieldOverflow)
        .map(Datum::from)
}
```

The parameters control the behavior of the code that we generate and allow the function to override the default behavior.

### Checklist


src/expr-derive-impl/src/sqlfunc.rs
```diff
@@ -0,0 +1,441 @@
...
+    } else {
+        quote! {
+            stringify!(#fn_name)
+        }
+    };
+
+    let (output_type, introduces_nulls_fn) = if let Some(output_type) = output_type {
+        let introduces_nulls_fn = quote! {
+            fn introduces_nulls(&self) -> bool {
+                <#output_type as mz_repr::DatumType<'_, ()>>::nullable()
```
ParkMyCar, line 242: in a macro when referencing an external crate I think it's best practice to prefix it with a `::`, e.g.
```
<#output_type as ::mz_repr::DatumType<'_, ()>>::nullable()
```
to ensure it doesn't conflict with a module of the same name in the crate the macro is being used. Not a huge deal, but might help someone in the future, especially for a crate like `serde` which _maybe_ could also be the name of a module

src/expr-derive-impl/src/sqlfunc.rs
```diff
@@ -0,0 +1,441 @@
...
+            }
+
+            fn output_type(&self, input_type: mz_repr::ColumnType) -> mz_repr::ColumnType {
+                use mz_repr::AsColumnType;
+                let output = #output_type::as_column_type();
+                let propagates_nulls = crate::func::EagerUnaryFunc::propagates_nulls(self);
+                let nullable = output.nullable;
+                // The output is nullable if it is nullable by itself or the input is nullable
+                // and this function propagates nulls
+                output.nullable(nullable || (propagates_nulls && input_type.nullable))
```
ParkMyCar, line 280: What about `introduces_nulls`?

ParkMyCar (approved):
Before merging could you take a pass on the comments? Especially would love to see the documentation improved but overall love this PR!

src/expr-derive-impl/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "mz-expr-derive-impl"
```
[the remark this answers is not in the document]
antiguru: Yes: We can't export non-proc-macro symbols from the `mz-expr-derive` crate, but we want to have access to the raw implementation to enable the tests. I'll leave a comment.

src/expr-derive-impl/src/sqlfunc.rs
antiguru: I _think_ it's handled via the output type, but I'm honestly not sure. I think I'll merge as-is, and revisit next.

ptravers: I don't know if you would ever want to expand this to variadic functions but it might be nice to take `introduces_nulls` as a parameter because the output type of `string_to_array` is not nullable but it can return nulls iff the first parameter is null

src/expr-derive-impl/src/sqlfunc.rs
```diff
@@ -0,0 +1,441 @@
...
+/// Determines the argument type of the nth argument of the function.
+///
+/// Adds a lifetime `'a` to the argument type if it is a reference type.
+///
+/// Panics if the function has fewer than `nth` arguments. Returns an error if
+/// the parameter is a `self` receiver.
+fn arg_type(arg: &syn::ItemFn, nth: usize) -> Result<syn::Type, syn::Error> {
+    match &arg.sig.inputs[nth] {
+        syn::FnArg::Typed(pat) => {
+            // Patch lifetimes to be 'a if reference
```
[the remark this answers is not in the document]
antiguru: It might, but if we change it, we need to patch things like `Datum<'a>`. I'll leave a comment and not fix it, because I first have to figure out what's the best way forward.

src/expr-derive-impl/src/sqlfunc.rs
antiguru: Good point, fixed.

src/expr-derive-impl/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+workspace-hack = { version = "0.0.0", path = "../workspace-hack", optional = true }
+
+[dev-dependencies]
+insta = { version = "1.42", features = ["serde"] }
+mz-ore = { path = "../ore", features = ["test"] }
+prettyplease = "0.2"
+
+[package.metadata.cargo-udeps.ignore]
+normal = ["workspace-hack"]
+
```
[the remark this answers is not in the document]
antiguru: Added this, and also insta and similar as recommended by insta.

src/expr-derive/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! Proc macros to derive SQL function traits.
+
+/// Derive function traits for SQL functions.
+#[proc_macro_attribute]
+pub fn sqlfunc(
```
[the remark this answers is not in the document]
antiguru: Added more documentation. It'd be nice to link to the traits we're generating implementations for, but that's not easy as the dependency points the other way.

src/expr-derive-impl/src/sqlfunc.rs
```diff
@@ -0,0 +1,441 @@
...
+        inverse,
+        is_infix_op,
+        output_type,
+        negate,
+        could_error,
+        propagates_nulls,
+    } = modifiers;
+
+    if is_infix_op.is_some() {
+        return Err(darling::Error::unknown_field("is_infix_op"));
```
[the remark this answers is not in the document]
antiguru: Added more information to the error.

src/expr-derive-impl/src/sqlfunc.rs
antiguru: Makes sense! I added a `introduces_nulls` parameters, which likely needs refinement in a later PR, but we should def. have it.

PR #32084, opened 2025-04-03
docs: replace tools-integration Overview page with as generic landing…

kay-kim:
… page placeholder + clean up references to old content

- we had a hidden Cube page ... marked that as draft so that it's not published.  We could either publish it or rm it.

https://preview.materialize.com/materialize/32084/integrations/


doc/user/content/integrations/cube.md
```diff
@@ -3,6 +3,7 @@ title: "Cube"
 description: "How to connect a Cube deployment to Materialize"
 aliases:
   /third-party/cube/
+draft: true
```
kay-kim, line 6: Not sure what to do about this page.  It wasn't in the nav ... so, just marked as draft.  We can either rm it or surface it.

doc/user/content/sql/create-source/_index.md
```diff
@@ -24,12 +24,13 @@ Materialize bundles **native connectors** that allow ingesting data from the fol
...
-- [Other message brokers](/integrations/#message-brokers)
 {{</ linkbox >}}
 {{< linkbox title="Databases (CDC)" >}}
-- [PostgreSQL](/sql/create-source/postgres)
-- [MySQL](/sql/create-source/mysql)
-- [Other databases](/integrations/#other-databases)
+- [PostgreSQL](/ingest-data/postgres/)
+- [MySQL](/ingest-data/mysql/)
+- [SQL Server](/ingest-data/cdc-sql-server/)
+- [MongoDB](https://github.com/MaterializeIncLabs/materialize-mongodb-debezium)
```
kay-kim, line 32: Copied this list from  ingest_data page  -- we listed MongoDB there ....
Not sure what the status of the MongoDB stuff is --  There was discussion of both incorporating into docs as well as putting the mdb work on hold 😄

doc/user/content/integrations/client-libraries/_index.md
```diff
@@ -0,0 +1,10 @@
+---
+title: "Client Libraries"
+description: ""
```
SangJunBak: nit: Should maybe add a description to be consistent? Something like "Appendix for client libraries". Not sure where the descriptions are used tho, maybe o11y?

SangJunBak:
It feels bad that we're losing the prescribed drivers from the following section:
Screenshot 2025-04-03 at 3 57 53 PM

and the "Submit a feature request" functionality. It'd be nice to keep it in the bottom of the index pages in `/integrations/client-libraries/` and `integrations/sql-clients/`, but were we thinking of getting rid of it because not enough people actually submit request?

doc/user/content/integrations/client-libraries/_index.md
kay-kim: Oh ... that was done in the 2nd patch.  But, I'll port over that content you mention above to be here in the next patch.

doc/user/content/integrations/client-libraries/_index.md
```diff
@@ -0,0 +1,33 @@
+---
+title: "SQL drivers"
```
kay-kim, line 2: Should I just rename the folder to be "SQL-drivers"? or rename the title of this page "Client libraries"?

What's currently being published uses client-libraries in the path but named it SQL drivers ... which is why I created an actual folder called client-libraries ... but it then seems strange to name it SQL drivers.

doc/user/content/integrations/client-libraries/_index.md
kay-kim: Added a patch with the table and file a request.

SangJunBak (approved):
Thanks for doing those changes! LGTM

PR #32088, opened 2025-04-03
[sql_server] Update `CdcStream` to emit `CdcEvent::Progress`

ParkMyCar:
This PR changes the implementation of `mz_sql_server_util::cdc::CdcStream` to emit not only changes to upstream tables, but also "progress" events. Specifically the `CdcEvent` type is now an enum, `CdcEvent::Data` and `CdcEvent::Progress`. This matches TimelyDataflow semantics, and makes it easier to integrate into our Source pipeline.

### Motivation

Progress towards: https://github.com/MaterializeInc/database-issues/issues/8762

### Checklist

### DIFF
--- a/src/sql-server-util/src/cdc.rs
+++ b/src/sql-server-util/src/cdc.rs
@@ -67,18 +67,6 @@ impl<'a> CdcStream<'a> {
         self
     }
 
-    /// Adds "one" to the provided [`Lsn`].
-    ///
-    /// Note: This method calls the upstream SQL Server instance to increment
-    /// the [`Lsn`] for us, but it is an entirely stateless operation, in other
-    /// words it does not modify the server state. While we understand the
-    /// structure of an [`Lsn`], the values are opaque to us so we need to call
-    /// the server to figure out what the logically next [`Lsn`] is.
-    pub async fn increment_lsn(&mut self, lsn: Lsn) -> Result<Lsn, SqlServerError> {
-        let new_lsn = crate::inspect::increment_lsn(self.client, lsn).await?;
-        Ok(new_lsn)
-    }
-
     /// Takes a snapshot of the upstream table that the specified `capture_instance` is
     /// replicating changes from.
     ///
--- a/src/sql-server-util/src/cdc.rs
+++ b/src/sql-server-util/src/cdc.rs
@@ -67,8 +67,14 @@ impl<'a> CdcStream<'a> {
         self
     }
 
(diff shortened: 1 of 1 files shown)


src/sql-server-util/src/cdc.rs
```diff
@@ -67,8 +67,14 @@ impl<'a> CdcStream<'a> {
         self
     }
 
-    /// Takes a snapshot of the upstream tables that the specified `capture_instance`s
-    /// are replicating changes from.
+    /// Bumps the provided [`Lsn`] by one.
+    pub async fn bump_lsn(&mut self, lsn: Lsn) -> Result<Lsn, SqlServerError> {
```
petrosagg, line 71: Was this accidentally included in this PR? I don't see it being used.

Separately, it would be nice to explain a bit more what it does in the doc comment. As it stands it's as if it simply computes `lsn + 1` but it seems to be modifying server state. Is this function reporting our progress of ingestion? Or something else?

petrosagg (approved):
LGTM! Just a minor request for the documentation of `bump_lsn`

src/sql-server-util/src/cdc.rs
ParkMyCar: It's not being used in this PR, but I included it so it would be easier to review, and I'm glad you did :)

Great call out about the comments and whether or not this communicates progress. Calling this does not modify the server state in anyways, it's just used to figure out what the next LSN value is

petrosagg: > Calling this does not modify the server state in anyways, it's just used to figure out what the next LSN value is

Then I'm still worried that we call out to the server! The comment calls out that the LSNs are opaque but at the same time we must be able to understand them and order them for the timely stream to make any sense. When we implement `PartialOrder` on the type we're asserting this fact. If we can't rely on manipulating LSNs ourselves then we may have to revisit the design of this source. Is there really any danger of doing a `+1` ourselves without calling out to the server?

ParkMyCar: I see! I have misworded the comment, we do understand the structure of the LSN ([code](https://github.com/MaterializeInc/materialize/blob/main/src/sql-server-util/src/cdc.rs#L290-L306)) and how they sort. We should be able to safely do a `+1` to the last component of the LSN (the log record number), from what I've observed this is all the SQL function does. 

The current implementation calls the server mostly because it seemed to fit into the model thus far, but we should be able to switch to manually bumping it by one.

I'm going to remove this part of the PR for now and can explore this more when it's actually needed!

PR #32094, opened 2025-04-04
timely-util: fix cloning `Column::Align`

teskje:
The implementation of `Column::clone` for the `Align` variant would first allocate a zero-filled region with `alloc_aligned_zeroed`, and then attempt to _append_ the bytes from the original region to its end using `Region::extend_from_slice`. The correct thing to do is to `copy_from_slice` instead, overwriting the zeroes.

Furthermore, the `Column` implementation calls `copy_from_slice` on a region returned by `alloc_aligned_zeroed` in a couple places. The returned region can be larger than the requested length, in which case `copy_from_slice` would panic because it expects both source and destination to have the same length. This is fixed here by slicing the destination to the source length.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9088

### Checklist

### DIFF
--- a/src/timely-util/src/containers.rs
+++ b/src/timely-util/src/containers.rs
@@ -126,12 +126,13 @@ mod container {
                 Column::Bytes(b) => {
                     assert_eq!(b.len() % 8, 0);
                     let mut alloc: Region<u64> = super::alloc_aligned_zeroed(b.len() / 8);
-                    bytemuck::cast_slice_mut(&mut alloc[..]).copy_from_slice(&b[..]);
+                    let alloc_bytes = bytemuck::cast_slice_mut(&mut alloc);
+                    alloc_bytes[..b.len()].copy_from_slice(b);
                     Self::Align(alloc)
                 }
                 Column::Align(a) => {
                     let mut alloc = super::alloc_aligned_zeroed(a.len());
-                    alloc.extend_from_slice(&a[..]);
+                    alloc[..a.len()].copy_from_slice(a);
                     Column::Align(alloc)
                 }
             }
@@ -198,7 +199,8 @@ mod container {
             } else {
                 // We failed to cast the slice, so we'll reallocate.
                 let mut alloc: Region<u64> = super::alloc_aligned_zeroed(bytes.len() / 8);
-                bytemuck::cast_slice_mut(&mut alloc[..]).copy_from_slice(&bytes[..]);
+                let alloc_bytes = bytemuck::cast_slice_mut(&mut alloc);
+                alloc_bytes[..bytes.len()].copy_from_slice(&bytes);
                 Self::Align(alloc)
             }
         }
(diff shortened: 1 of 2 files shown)


src/timely-util/src/containers.rs
```diff
@@ -474,3 +476,75 @@ mod provided_builder {
...
+
+    /// Produce some bytes that are in columnar format.
+    fn raw_columnar_bytes() -> Vec<u8> {
+        let mut raw = Vec::new();
+        raw.extend(12_u64.to_le_bytes()); // length
+        raw.extend(1_i32.to_le_bytes());
+        raw.extend(2_i32.to_le_bytes());
+        raw.extend(3_i32.to_le_bytes());
+        raw.extend([0, 0, 0, 0]); // padding
+        raw
```
teskje, line 497: I tried using `Sequence::encode` here, but it gives me a `Vec<u64>` and I don't know how to convert that into a `Vec<u8>` without unsafe, so I just went the manual way.

src/timely-util/src/containers.rs
```diff
@@ -474,3 +476,75 @@ mod provided_builder {
...
+        let column: Column<i32> = Column::from_bytes(aligned_bytes);
+        assert!(matches!(column, Column::Bytes(_)));
+        assert_eq!(column.iter().collect::<Vec<_>>(), vec![&1, &2, &3]);
+
+        let buf = vec![0; raw.len() + 8];
+        let align = buf.as_ptr().align_offset(std::mem::size_of::<u64>());
+        let mut bytes_mut = BytesMut::from(buf);
+        let _ = bytes_mut.extract_to(align + 1);
+        bytes_mut[..raw.len()].copy_from_slice(&raw);
+        let unaligned_bytes = bytes_mut.extract_to(raw.len());
```
teskje, line 544: Is there an easier way to get aligned/unaligned `Bytes`?

src/timely-util/src/containers.rs
antiguru: You could use bytemuck: https://docs.rs/bytemuck/latest/bytemuck/fn.cast_slice.html

antiguru:
Thank you! I'll check in more detail on Monday. We shouldn't use this path as we're not using the zero-copy allocator and don't enable lgalloc for columnar if I'm not mistaken.

teskje:
> We shouldn't use this path as we're not using the zero-copy allocator and don't enable lgalloc for columnar if I'm not mistaken.

The `enable_columnar_lgalloc` dyncfg defaults to `true` and we're not setting a different value in LD, so I think we do enable it. The [panic](https://materializeinc.sentry.io/issues/6401706298/?project=6780145&referrer=github_integration) only occurred two times in @jubrad's staging environment, so it seems like a path that we hit very rarely. Not sure what starts need to align for that.

src/timely-util/src/containers.rs
```diff
@@ -126,12 +126,13 @@ mod container {
                 Column::Bytes(b) => {
                     assert_eq!(b.len() % 8, 0);
                     let mut alloc: Region<u64> = super::alloc_aligned_zeroed(b.len() / 8);
-                    bytemuck::cast_slice_mut(&mut alloc[..]).copy_from_slice(&b[..]);
+                    let alloc_bytes = bytemuck::cast_slice_mut(&mut alloc);
+                    alloc_bytes[..b.len()].copy_from_slice(b);
```
antiguru, line 130: This shouldn't be necessary, right? `alloc_bytes` should have capacity for all of `b`, so it should copy the whole of `b`, even if it has more capacity.

src/timely-util/src/containers.rs
```diff
@@ -198,7 +199,8 @@ mod container {
             } else {
                 // We failed to cast the slice, so we'll reallocate.
                 let mut alloc: Region<u64> = super::alloc_aligned_zeroed(bytes.len() / 8);
-                bytemuck::cast_slice_mut(&mut alloc[..]).copy_from_slice(&bytes[..]);
+                let alloc_bytes = bytemuck::cast_slice_mut(&mut alloc);
+                alloc_bytes[..bytes.len()].copy_from_slice(&bytes);
```
antiguru, line 203: Here again, this should be needed, right?

src/timely-util/src/containers.rs
```diff
@@ -126,12 +126,13 @@ mod container {
...
                     let mut alloc: Region<u64> = super::alloc_aligned_zeroed(b.len() / 8);
-                    bytemuck::cast_slice_mut(&mut alloc[..]).copy_from_slice(&b[..]);
+                    let alloc_bytes = bytemuck::cast_slice_mut(&mut alloc);
+                    alloc_bytes[..b.len()].copy_from_slice(b);
                     Self::Align(alloc)
                 }
                 Column::Align(a) => {
                     let mut alloc = super::alloc_aligned_zeroed(a.len());
-                    alloc.extend_from_slice(&a[..]);
+                    alloc[..a.len()].copy_from_slice(a);
```
antiguru, line 135: It's a bug, thanks for catching!

We might not want to call `alloc_aligned_zeroed` here because it sets the memory twice. Instead, we'd want to allocate an uninitialized region and use `extend_from_slice` as before. This needs a new API because `alloc_aligned_zeroed` checks whether to use lgalloc or not, and we don't have an equivalent non-initializing variant.

antiguru (commented):
Left some comments inline. Thanks very much for diagnosing and fixing!

src/timely-util/src/containers.rs
teskje: `copy_from_slice` panics if the slices have different lengths ([docs](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice)). Since `alloc_bytes` can be longer than `b`, we need to truncate it to avoid the panic.

src/timely-util/src/containers.rs
teskje: Makes sense! I'd like to treat that out of scope for this PR though.

antiguru (approved):

teskje:
TFTR!

bosconi:
bugbot run

PR #32095, opened 2025-04-04
Dictionary compressed arrangements

frankmcsherry:
Enables (behind a default-off flag) dictionary compression in arrangements.

Roughly, as each `row: &[u8]` is presented, we'll rip it apart into columns, following `read_datum`, and look for the option to use otherwise unused byte patterns (tags, otherwise used for row decoding) to reference popular values in each column. Columns are encoded independently, so things can be differently popular in different columns.

### Checklist

### DIFF
--- a/misc/python/materialize/mzcompose/__init__.py
+++ b/misc/python/materialize/mzcompose/__init__.py
@@ -71,6 +71,7 @@ def get_minimal_system_parameters(
             "true" if version >= MzVersion.parse_mz("v0.132.0-dev") else "false"
         ),
         "enable_alter_swap": "true",
+        "enable_arrangement_dictionary_compression_alpha": "false",
         "enable_case_literal_transform": "true",
         "enable_cast_elimination": "true",
         "enable_coalesce_case_transform": "true",
--- a/misc/python/materialize/parallel_workload/action.py
+++ b/misc/python/materialize/parallel_workload/action.py
@@ -1582,6 +1582,9 @@ class FlipFlagsAction(Action):
             BOOLEAN_FLAG_VALUES
         )
         self.flags_with_values["enable_alter_table_add_column"] = BOOLEAN_FLAG_VALUES
+        self.flags_with_values["enable_arrangement_dictionary_compression_alpha"] = (
+            BOOLEAN_FLAG_VALUES
+        )
         self.flags_with_values["enable_compute_peek_response_stash"] = (
             BOOLEAN_FLAG_VALUES
         )
--- a/src/compute-types/src/dyncfgs.rs
+++ b/src/compute-types/src/dyncfgs.rs
@@ -359,6 +359,20 @@ pub const COMPUTE_LOGICAL_BACKPRESSURE_INFLIGHT_SLACK: Config<Duration> = Config
     "Round observed timestamps to slack.",
 );
 
+/// Enable per-column dictionary compression for row containers in arrangements.
+///
+/// The `_alpha` suffix is load-bearing: this feature is not yet considered
+/// production-ready, and the name is meant to make that unmissable at the
+/// `ALTER SYSTEM SET` call site rather than relying on out-of-band warnings.
+///
+/// Disposition: added 2026-06-09, default off; solicit feedback for one month
+/// and remove in the absence of a positive response.
+pub const ENABLE_ARRANGEMENT_DICTIONARY_COMPRESSION_ALPHA: Config<bool> = Config::new(
+    "enable_arrangement_dictionary_compression_alpha",
+    false,
+    "Enable arrangement dictionary compression (alpha; not yet production-ready).",
+);
+
 /// Whether to enable the peek response stash, for sending back large peek
 /// responses. The response stash will only be used for results that exceed
 /// `compute_peek_response_stash_threshold_bytes`.
@@ -484,6 +498,7 @@ pub fn all_dyncfgs(configs: ConfigSet) -> ConfigSet {
         .add(&ENABLE_COMPUTE_LOGICAL_BACKPRESSURE)
         .add(&COMPUTE_LOGICAL_BACKPRESSURE_MAX_RETAINED_CAPABILITIES)
         .add(&COMPUTE_LOGICAL_BACKPRESSURE_INFLIGHT_SLACK)
+        .add(&ENABLE_ARRANGEMENT_DICTIONARY_COMPRESSION_ALPHA)
         .add(&ENABLE_PEEK_RESPONSE_STASH)
         .add(&PEEK_RESPONSE_STASH_THRESHOLD_BYTES)
         .add(&PEEK_RESPONSE_STASH_BATCH_MAX_RUNS)
--- a/src/row-spine/src/lib.rs
+++ b/src/row-spine/src/lib.rs
@@ -7,19 +7,14 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-//! Packed-bytes differential dataflow spine [`Layout`]s for `Row`-valued
-//! arrangements.
+//! Types and traits in support of containers for row-encoded byte slices.
 //!
-//! The spines defined here store `Row` keys and values as concatenated bytes
-//! in a single contiguous backing region (via `mz_ore::region::Region`, which
-//! uses lgalloc when available), instead of as separately-allocated heap
-//! objects. This gives cursor lookups block locality and lets the OS evict
-//! cold pages cleanly under memory pressure.
-//!
-//! [`Layout`]: differential_dataflow::trace::implementations::Layout
+//! This includes the vanilla `bytes_container` that holds byte slices in contiguous
+//! allocations, as well as a `dictionary` encoding wrapper that is able to rewrite
+//! the byte slices to use spare tags in each column to reference common values.
 
-pub use self::container::DatumContainer;
-pub use self::container::DatumSeq;
+pub use self::dictionary::DatumContainer;
+pub use self::dictionary::DatumSeq;
 pub use self::offset_opt::OffsetOptimized;
 pub use self::spines::{
     RowBatcher, RowBuilder, RowRowBatcher, RowRowBuilder, RowRowColPagedBuilder, RowRowSpine,
@@ -27,6 +22,10 @@ pub use self::spines::{
     ValRowColPagedBuilder, ValRowSpine,
 };
 use differential_dataflow::trace::implementations::OffsetList;
+
+/// Enable per-column dictionary compression in row containers.
+pub static DICTIONARY_COMPRESSION: std::sync::atomic::AtomicBool =
+    std::sync::atomic::AtomicBool::new(false);
 
 /// Spines specialized to contain `Row` types in keys and values.
 mod spines {
@@ -36,8 +35,9 @@ mod spines {
     use differential_dataflow::trace::implementations::Layout;
     use differential_dataflow::trace::implementations::Update;
     use differential_dataflow::trace::implementations::merge_batcher::MergeBatcher;
-    use differential_dataflow::trace::implementations::ord_neu::{OrdKeyBatch, OrdKeyBuilder};
-    use differential_dataflow::trace::implementations::ord_neu::{OrdValBatch, OrdValBuilder};
+    use differential_dataflow::trace::implementations::ord_neu::{
+        OrdKeyBatch, OrdValBatch, OrdValBuilder,
+    };
     use differential_dataflow::trace::implementations::spine_fueled::Spine;
     use differential_dataflow::trace::rc_blanket_impls::RcBuilder;
     use mz_repr::Row;
@@ -53,12 +53,11 @@ mod spines {
 
     pub type RowRowSpine<T, R> = Spine<Rc<OrdValBatch<RowRowLayout<((Row, Row), T, R)>>>>;
     pub type RowRowBatcher<T, R> = KeyValBatcher<Row, Row, T, R>;
-    pub type RowRowBuilder<T, R> = RcBuilder<
-        OrdValBuilder<RowRowLayout<((Row, Row), T, R)>, ColumnationStack<((Row, Row), T, R)>>,
-    >;
+    pub type RowRowBuilder<T, R> = RcBuilder<crate::dictionary::builders::RowRowBuilder<T, R>>;
 
     /// `RowRowBuilder` variant that consumes [`Column`] chunks. Pairs with
-    /// [`Col2ValPagedBatcher`] for the spillable arrange path.
+    /// [`Col2ValPagedBatcher`] for the spillable arrange path. This is the stock
+    /// (non-dictionary) builder; dictionary-compressing the paged path is a follow-up.
     ///
     /// [`Col2ValPagedBatcher`]: mz_timely_util::columnar::Col2ValPagedBatcher
     pub type RowRowColPagedBuilder<T, R> =
@@ -66,20 +65,17 @@ mod spines {
 
     pub type RowValSpine<V, T, R> = Spine<Rc<OrdValBatch<RowValLayout<((Row, V), T, R)>>>>;
     pub type RowValBatcher<V, T, R> = KeyValBatcher<Row, V, T, R>;
-    pub type RowValBuilder<V, T, R> = RcBuilder<
-        OrdValBuilder<RowValLayout<((Row, V), T, R)>, ColumnationStack<((Row, V), T, R)>>,
-    >;
+    pub type RowValBuilder<V, T, R> =
+        RcBuilder<crate::dictionary::builders::RowValBuilder<V, T, R>>;
 
     pub type RowSpine<T, R> = Spine<Rc<OrdKeyBatch<RowLayout<((Row, ()), T, R)>>>>;
     pub type RowBatcher<T, R> = KeyBatcher<Row, T, R>;
-    pub type RowBuilder<T, R> =
-        RcBuilder<OrdKeyBuilder<RowLayout<((Row, ()), T, R)>, ColumnationStack<((Row, ()), T, R)>>>;
+    pub type RowBuilder<T, R> = RcBuilder<crate::dictionary::builders::RowBuilder<T, R>>;
 
     pub type ValRowSpine<K, T, R> = Spine<Rc<OrdValBatch<ValRowLayout<((K, Row), T, R)>>>>;
     pub type ValRowBatcher<K, T, R> = KeyValBatcher<K, Row, T, R>;
-    pub type ValRowBuilder<K, T, R> = RcBuilder<
-        OrdValBuilder<ValRowLayout<((K, Row), T, R)>, ColumnationStack<((K, Row), T, R)>>,
-    >;
+    pub type ValRowBuilder<K, T, R> =
+        RcBuilder<crate::dictionary::builders::ValRowBuilder<K, T, R>>;
 
     /// `ValRowBuilder` variant that consumes [`Column`] chunks. Pairs with
     /// `Col2ValPagedBatcher<K, Row, T, R>` for the spillable arrange path where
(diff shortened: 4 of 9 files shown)


src/row-spine/src/lib.rs
```diff
@@ -97,7 +100,7 @@ mod spines {
     }
 }
 
-/// A `Row`-specialized container using dictionary compression.
```
antiguru, line 155: :D

src/compute/src/row_spine.rs
```diff
@@ -622,3 +627,893 @@ pub(crate) fn offset_list_size(data: &OffsetList, mut callback: impl FnMut(usize
...
+            let codec = match (&cont1.codec, &cont2.codec) {
+                (Some(c1), Some(c2)) => {
+                    c1.report();
+                    c2.report();
+                    Some(ColumnsCodec::new_from([c1, c2]))
+                }
+                (None, None) => None,
+                _ => {
+                    panic!("Unclear codec configuration!")
+                }
```
antiguru: I think this case can happen when toggling the flag while we already have batches, no? We'd either need to rebuild the dictionary on the other side, or just discard and converge toward no compression.

src/compute/src/row_spine.rs
```diff
@@ -622,3 +627,893 @@ pub(crate) fn offset_list_size(data: &OffsetList, mut callback: impl FnMut(usize
...
+                        self.total,
+                        self.bytes,
+                        self.total / self.bytes
+                    );
+                    println!("COLUMNS: {:?}", self.columns.len());
+                    for column in self.columns.iter() {
+                        column.report()
+                    }
+                }
+            }
```
antiguru: This might need to change before we turn it on in prod :) I think it'd be fine to use `trace!` instead.

src/row-spine/src/lib.rs
```diff
@@ -622,3 +627,893 @@ pub(crate) fn offset_list_size(data: &OffsetList, mut callback: impl FnMut(usize
...
+                    }
+                }
+            }
+
+            fn borrow_row(row: &Row) -> Self::DecodeIter<'_> {
+                ColumnsIter {
+                    index: None,
+                    column: 0,
+                    data: row.data(),
+                }
```
antiguru, line 1469: This could call `without_codec` instead. Would require `unsafe` and the associated safety argument, but then we'd have a comment as to why this is safe.

src/compute/src/row_spine.rs
```diff
@@ -622,3 +627,893 @@ pub(crate) fn offset_list_size(data: &OffsetList, mut callback: impl FnMut(usize
...
+            // /// Allocates a Misra-Gries summary which intends to hold up to `k` examples.
+            // ///
+            // /// After `n` insertions it will contain only elements that were inserted at least `n/k` times.
+            // /// The actual memory use is proportional to `2 * k`, so that we can amortize the consolidation.
+            // pub fn with_capacity(k: usize) -> Self {
+            //     Self {
+            //         inner: Vec::with_capacity(2 * k),
+            //     }
+            // }
```
antiguru: Nit: Remove, or allow dead code.

src/compute/src/row_spine.rs
```diff
@@ -622,3 +627,893 @@ pub(crate) fn offset_list_size(data: &OffsetList, mut callback: impl FnMut(usize
...
+                    "\t{:?}v{:?}: {:?} -> {:?} + {:?} = (x{:?})",
+                    tags_used,
+                    mg.len(),
+                    self.total,
+                    self.bytes,
+                    bytes,
+                    self.total / (self.bytes + bytes),
+                )
+                // }
+            }
```
antiguru: Remove, or use tracing instead.

src/compute-types/src/dyncfgs.rs
```diff
@@ -229,6 +229,13 @@ pub const COMPUTE_LOGICAL_BACKPRESSURE_INFLIGHT_SLACK: Config<Duration> = Config
     "Round observed timestamps to slack.",
 );
 
+/// Enable per-column dictionary compression for row containers in arrangements.
+pub const ENABLE_ARRANGEMENT_DICTIONARY_COMPRESSION: Config<bool> = Config::new(
+    "enable_arrangement_dictionary_compression",
+    true,
```
antiguru, line 459:
```suggestion
    false,
```

Please disable by default before you merge.

antiguru (commented):
Left some feedback, but otherwise I like the change. We should figure out the runtime overhead, but I suspect it won't be too bad.

src/compute/src/row_spine.rs
frankmcsherry: Changed to go to `None` for any config other than the `((Some(_), Some(_))` case. I think that is more pessimistic than it needs to be, but is a better place to be than panicking.

misc/python/materialize/mzcompose/__init__.py
```diff
@@ -71,9 +71,11 @@ def get_minimal_system_parameters(
             "true" if version >= MzVersion.parse_mz("v0.132.0-dev") else "false"
         ),
         "enable_alter_swap": "true",
+        "enable_arrangement_dictionary_compression": "false",
         "enable_case_literal_transform": "true",
         "enable_cast_elimination": "true",
         "enable_coalesce_case_transform": "true",
+        "enable_compute_active_dataflow_cancelation": "true",
```
antiguru: Stray change? I think active cancellation is the only cancellation mechanism now.

misc/python/materialize/parallel_workload/action.py
```diff
@@ -1571,6 +1571,12 @@ def __init__(
             BOOLEAN_FLAG_VALUES
         )
         self.flags_with_values["enable_alter_table_add_column"] = BOOLEAN_FLAG_VALUES
+        self.flags_with_values["enable_arrangement_dictionary_compression"] = (
+            BOOLEAN_FLAG_VALUES
+        )
+        self.flags_with_values["enable_compute_active_dataflow_cancelation"] = (
+            BOOLEAN_FLAG_VALUES
+        )
```
antiguru: Dito, seems stray.

src/compute-types/src/dyncfgs.rs
```diff
@@ -296,6 +296,20 @@ pub const COMPUTE_LOGICAL_BACKPRESSURE_INFLIGHT_SLACK: Config<Duration> = Config
...
+    "Enable arrangement dictionary compression.",
+);
+
+/// Whether to use `drop_dataflow` to actively cancel dataflows.
+pub const ENABLE_ACTIVE_DATAFLOW_CANCELATION: Config<bool> = Config::new(
+    "enable_compute_active_dataflow_cancelation",
+    false,
+    "Whether to use `drop_dataflow` to actively cancel compute dataflows.",
+);
+
```
antiguru: Here again, and below: Stray change.

misc/python/materialize/mzcompose/__init__.py
frankmcsherry: Almost surely. It had a year's worth of rebases to get through, and .. never sure what is meant to be there vs not.

misc/python/materialize/mzcompose/__init__.py
```diff
@@ -112,9 +113,6 @@ def get_minimal_system_parameters(
         # End of list (ordered by name)
     }
 
-    if version < MzVersion.parse_mz("v0.163.0-dev"):
-        config["enable_compute_active_dataflow_cancelation"] = "true"
-
```
frankmcsherry, line 113: Ah, @antiguru now I have this in here. We should chat about what is meant to be true about this feature. :D Probably this gets copy/pasted in, whenever it was missed in some rebase?

misc/python/materialize/mzcompose/__init__.py
```diff
@@ -71,6 +71,7 @@ def get_minimal_system_parameters(
             "true" if version >= MzVersion.parse_mz("v0.132.0-dev") else "false"
         ),
         "enable_alter_swap": "true",
+        "enable_arrangement_dictionary_compression": "false",
```
def-: Is there a reason to have it off in tests? My assumption is that we aim to enable it in production at some point, there is more risk in the new dictionary compression logic than the non-dictionary compression logic, and hopefully not too much overhead in enabling it in CI? If we don't want to enable it in general, I'm at least running a one-off for the test and nightly pipeline to see what falls over:
- https://buildkite.com/materialize/test/builds/122304
- https://buildkite.com/materialize/nightly/builds/16318

antiguru:
Reviewed locally; structural points after a second-opinion validation pass.

**Blocking**
* No encode→decode round-trip test covers the compressed paths. `test_round_trip` runs with `DICTIONARY_COMPRESSION` defaulted off, so the `new_safe` install, `new_from`-via-seen-tags merge, and builder `seal` paths all lack unit coverage. Given the soundness contract on `SAFE_TAG_BASE` and the seen-tags bitmap, this is the gap I'd want closed before merge.

**Strong**
* `merge_capacity`'s `new_from([c1, c2])` chooses dict tags from the complement of `c1 ∪ c2` first-byte bitmaps. The invariant "no later push has a first-byte outside that union" is load-bearing for soundness but is neither documented nor asserted. A `debug_assert` in the encode raw-fallthrough path would make it self-checking.
* `Codec::encode` is being used as a stats-feed channel in `build_codec` and the `push_into` stats branch (output buffer immediately discarded). A separate `observe(...)` entry point would make the API honest.
* `STATS_THRESHOLD = 64 * 1024` and MisraGries `k = 512` are hardcoded; promoting both to dyncfgs would let production tune without a release.
* `DatumSeq.iter`, `ColumnsIter.{index,column,data}` are `pub`, exposing encoding internals to in-crate fast-paths. Methods or `pub(super)` would localize the contract.
* `partial_cmp` slow path uses `flatten()` over `&[u8]`, comparing one `&u8` per step. Hot only when one side is dict-encoded; worth measuring before assuming it doesn't matter.
* `MisraGries::tidy` allocates+sort_unstable on every trigger (`len > 2k`); under high-cardinality streams this fires per-insert. A bench would confirm the BTreeMap variant is at least neutral vs. the prior vec impl.

**Nits**
* `bytes[0]` in `DictionaryCodec::encode` is unguarded; current callers always supply non-empty slices, but a `debug_assert!(!bytes.is_empty())` is cheap insurance.
* Stray `// Stats stuff.` and two truncated `// Optional only to support borrowing owned as this` comments.
* `#[cfg(test)] pub use dictionary::SAFE_TAG_BASE;` is dead — the test reaches it via the private module path.
* `ColumnsCodec::new_from` over an empty stats iterator yields a 0-column codec silently; not wrong, undocumented.

The 10-commit history includes a self-revert ("Remove rebase dross"); worth squashing.

Codebase health: net positive — cleaner builders, removed `report` debug prints, structural test for the safe-tag boundary.

misc/python/materialize/mzcompose/__init__.py
def-: It actually looks pretty good! Only benchmark regressions in CPU, which is probably expected: https://buildkite.com/materialize/nightly/builds/16318

frankmcsherry: It is expected, and I have no idea whether we want it on in tests. We def want it *off* in production, as this is purely exploratory, so I wouldn't want to force our tests to be exercising non-prod configs without some more thought. Happy to discuss, but lmk if this is a blocker for you rather than a thought. We've run on nightly often, and it's green other than the CPU perf.

antiguru:
## Review: correctness, allocations, complexity, docs

I reviewed this critically (and had a second pass independently fact-check the findings). **No outright unsoundness** — the two tag-assignment schemes hold up, the merge invariant is sound, and the `debug_assert` in `encode` nicely makes the load-bearing "no unobserved first-byte" invariant self-checking. The notes below are robustness / perf / memory-accounting / complexity.

### Correctness & robustness

1. **`DictionaryCodec::new_from`'s decode-map construction is correct but fragile.** In `for tag in 0..=255`, a free-but-MG-exhausted tag pushes *nothing* into the positional `BytesMap`, shifting subsequent indices. This stays correct only because the `mg` filter iterator is monotonically exhausting (so every skip happens strictly after the last `Some`, and only `None`/absent entries — which decode identically to fall-through — are ever shifted). It's a non-obvious argument for a load-bearing structure. Recommend an explicit `else { decode.push(None); }` so the map is always length-256 and `decode[tag]` is trivially tag-aligned; the cost is a few `usize` offsets. (`new_safe` doesn't have this issue because it never interleaves `None` after the safe base.)

2. **`clear()` leaves `codec = Some(..)` but empties it, silently and permanently disabling compression on reuse.** `clear` calls `codec.clear()` → `columns.clear()`, leaving `Some(ColumnsCodec { columns: [] })`. On reuse, `push_into` takes the `Some(codec)` branch and re-grows fresh *default* (empty-dictionary) per-column codecs, so everything falls through to raw — sound, but compression is off for the rest of that container's life, and `stats` is not restored to re-trigger an install. If reuse-after-clear is expected, either drop the codec (`self.codec = None`) and restore `stats`, or document that clear permanently disables compression.

3. **A merge where one input lacks a codec produces an uncompressed result.** `merge_capacity` only builds a codec when *both* inputs have one (`(Some, Some) => ..., _ => None`). A single not-yet-compressed input "poisons" the merge to uncompressed, discarding the other side's dictionary. Correct, but worth a comment given how often spine batches merge.

### Performance & allocations

4. **The comparison slow path is byte-at-a-time over re-parsed columns.** `partial_cmp`'s dictionary branch does `self.iter.flatten()`, which yields `&u8` and compares byte-by-byte, with each `ColumnsIter::next` doing a dict lookup or full `read_datum` reparse per column — versus a single length check + slice `memcmp` on the raw fast path. Comparisons dominate merge/sort, so this is the dominant steady-state cost of enabling compression. Worth a slice-wise comparison, or at least a comment.

5. **Stats are fed on *every* `encode`, forever.** `DictionaryCodec::encode` unconditionally runs `stats.0.insert_ref(bytes)` (a `BTreeMap` op, plus `to_owned()` on each newly-distinct value) and the bitmap update — even for an installed, hot codec whose MG will only matter at the next merge. There's no steady state where encoding is cheap. Consider decoupling "keep the cheap tag bitmap" (needed for merge safety) from "keep the expensive MG."

6. **The stats-gathering push copies each row twice.** During the entire pre-install phase (up to `STATS_THRESHOLD = 64*1024` pushes), `push_into(DatumSeq)` builds an encoded output via `stats.encode(.., &mut staging)`, discards it (`staging.clear()`), then re-iterates `bytes_iter()` to copy raw bytes. A stats-only "observe" method (MG + bitmap, no output construction) would halve the copying.

7. **`new_safe` clones the MisraGries even though the caller owns and immediately drops it.** Install site is `let stats = self.stats.take().unwrap(); self.codec = Some(stats.new_safe());`, but `new_safe`/`DictionaryCodec::new_safe` take `&self` and `clone()` the MG internally. Taking the MG by value would drop the clone. (`new_from` legitimately needs `&` for the multi-input merge.)

8. **`heap_size` under-reports.** It forwards only to `self.inner.heap_size`, omitting the codec's `encode` `BTreeMap`, the `decode` `BytesMap`, the `staging` `Vec`, and the `stats` MisraGries (up to `2*k = 1024` *owned* `Vec<u8>` per column, unbounded by value size). Memory-based monitoring/spilling will miss all of it.

### Inconsistencies

9. **`ValRowBuilder` is not switched to the dictionary builders.** Only `RowRowBuilder`/`RowValBuilder`/`RowBuilder` are wrapped; `ValRowBuilder` stays on the plain `OrdValBuilder`/`ColumnationStack`. Since `ValRowLayout::ValContainer = DatumContainer`, its `Row` value column can only compress via the 64K threshold-install path, never via a builder `seal` (`ValRowSpine` has a live consumer in `upsert_continual_feedback_v2`). Probably intentional, but the asymmetry isn't called out.

10. **The `stats` field doc contradicts the builder path (and the field is dead there).** Doc says *"None once the codec has been installed"*, but `seal` leaves `codec = Some` and `stats = Some` simultaneously (`with_capacity` sets `stats`, `seal` only overwrites `codec`). Since `push_into` checks the codec branch first, the MG in seal-path containers is never fed and never installs anything — pure dead memory (compounding #8). Clear `stats` when setting the codec, or fix the doc.

### Complexity that could be avoided

11. **The `Codec` trait is never used generically, which forces dead code.** Nothing is bounded on `C: Codec`; `DatumContainer` holds a concrete `Option<ColumnsCodec>`, and `ColumnsIter::next` reaches into `columns[i].decode` directly rather than through `DictionaryCodec::decode`. As a result `DictionaryCodec::decode`, `DictionaryCodec::borrow_row`, and the entire `DictionaryIter` struct are dead (its `next` is a byte-identical duplicate of `ColumnsIter`'s fall-through). Dropping the trait, or splitting encode/build from decode, removes a struct + two methods and clarifies the data flow.

### Docs / minor

- **Magic numbers want rationale.** `SAFE_TAG_BASE = 122`: the `Tag` enum currently has 94 variants (discriminants 0..=93), so 122 leaves a margin of 28, guarded by `test_safe_tag_base` — a one-liner tying 122 to that count would make the budget explicit. Also `STATS_THRESHOLD` and MG `k = 512`.
- **Doc/threshold mismatch.** The `STATS_THRESHOLD` comment targets "~128 values" but MG `k = 512` (tidy at `2*k`). Reconcile, since the heavy-hitter filter (`len > 1 && count > 1`) interacts with both.
- **Test gap.** `test_codec_round_trip` (nice addition) drives the codec in isolation, but no test sets `DICTIONARY_COMPRESSION = true`, so the *container*-level compressed paths — threshold install, `merge_capacity` transcoding, and the compressed `index`/`cmp`/`eq` branches — are entirely uncovered. These are the trickiest fast/slow-path branches in the PR.
- `BytesMap::len` is `#[allow(dead_code)]`.

Happy to file any of these as inline comments if that's easier to track.

---

frankmcsherry:
Thanks for the review! Addressed in 1152daa066 (all in `src/row-spine/src/lib.rs`). Point-by-point:

**Behavior / correctness**

- **decode-map tag alignment** — `new_from` now pushes a decode slot for *every* tag `0..=255`, including the case where a tag is unused but the heavy-hitter supply is already exhausted. Previously that branch pushed nothing, so a later in-use tag's slot landed at the wrong index and `decode.get(tag)` was no longer addressable by tag.
- **`clear()` permanently disabled compression** — it left `codec = Some(empty)`, which routes all future pushes down the encode path with an empty dictionary and prevents the `STATS_THRESHOLD` install logic from ever re-engaging. `clear()` now resets to the same state as a fresh `with_capacity` (`codec = None`, stats restored when the flag is on).
- **stats-gathering copied each row twice** — the gathering branch used to `encode` into a staging buffer purely for the stats side effect, discard it, then re-copy the raw bytes. Added a stats-only `observe` method; the row is now copied once.
- **`new_safe` cloned the MisraGries summary** — it took `&self` and cloned. It now consumes the stats by value (the call site has just `take`n them and is about to drop them).
- **`heap_size` under-reported** — it only walked `inner`. It now also accounts for the staging buffer, the codec, and the stats gatherer (`heap_size` added to `ColumnsCodec` / `DictionaryCodec` / `BytesMap` / `MisraGries`; `BTreeMap` node storage approximated, key/arena bytes exact).
- **ValRow value side not dictionary-compressed** — `ValRowBuilder` used the stock `OrdValBuilder`, so the `Row` *value* container never got a seal-time codec (unlike the other Row-bearing spines). Added a dictionary `ValRowBuilder` mirroring `RowValBuilder` with the roles swapped; it builds the codec from the value column and installs it on `vals.vals`.
- **stats field doc vs. builder seal path** — the three builder seal paths now set `stats = None` after installing a codec, so the `stats: None once codec installed` invariant actually holds (and the redundant gatherer allocation is dropped).

**Dead code / abstraction**

- **`Codec` trait never used generically** — dropped the trait entirely; `ColumnsCodec`/`DictionaryCodec` methods are now inherent. Deleted the dead `DictionaryIter` struct and `DictionaryCodec::decode`/`borrow_row` (decoding goes through `ColumnsIter`, which reads the `decode` map directly). Also removed the dead `BytesMap::len`.

**Docs / clarity (no behavior change)**

- **one-sided merge → uncompressed** — commented: a codec is sound only for data whose tag usage it observed, so we can't reuse one side's codec for the other's rows; the merged container re-gathers stats and may install a fresh codec later.
- **`partial_cmp` byte-at-a-time slow path** — commented: column slice boundaries don't line up between an encoded side and a raw side, so a flat byte stream is the only directly-comparable representation; the path is cold (fast path handles the common unencoded case).
- **stats fed on every encode "forever"** — split into the cheap, soundness-critical tag bitmap (must stay accurate for the codec's whole life, since `new_from`'s dynamic-tag path relies on it) vs. the expensive MisraGries summary. We deliberately keep feeding MG on the hot path: a later merge rebuilds the merged codec from these summaries, and freezing them at install time would mean the codec can never reclaim slots as the popular set drifts under consolidation — eventually compressing values that no longer occur instead of the ones that do.
- **`SAFE_TAG_BASE = 122`** — documented: the tight bound today is 94 (the `Tag` enum has 94 variants, discriminants `0..=93`); 122 leaves headroom, and `test_safe_tag_base` pins the real invariant (every datum encodes with a first byte `< SAFE_TAG_BASE`).
- **`STATS_THRESHOLD` comment** — reconciled with the MisraGries `k`: a safe codec has at most `256 - SAFE_TAG_BASE` (134) dictionary slots per column, and MG retains up to `2*k` (1024) candidates / reduces to `k` (512), comfortably more than 134.

clippy clean, `cargo test -p mz-row-spine` green (incl. the codec round-trip and `SAFE_TAG_BASE` tests).

antiguru (approved):
Looks good!

We probably need to work on integrating columnar soon, but that can happen separately.

src/row-spine/src/lib.rs
```diff
@@ -807,3 +658,1272 @@ pub(crate) fn offset_list_size(data: &OffsetList, mut callback: impl FnMut(usize
...
+                stats,
+            }
+        }
+        fn merge_capacity(cont1: &Self, cont2: &Self) -> Self {
+            // We only build a merged codec when *both* inputs carry one. A codec is
+            // sound only for the data whose tag usage it observed, so we cannot reuse
+            // one side's codec to decode the other side's rows. When exactly one side
+            // is compressed we conservatively produce an uncompressed container rather
+            // than risk a tag collision; the merged container re-gathers stats and may
+            // install a fresh codec later via the `STATS_THRESHOLD` path.
```
def-, line 1009: The constructor sets `stats: None`, so at the moment the comment is wrong. @frankmcsherry

misc/python/materialize/mzcompose/__init__.py
```diff
@@ -71,6 +71,7 @@ def get_minimal_system_parameters(
             "true" if version >= MzVersion.parse_mz("v0.132.0-dev") else "false"
         ),
         "enable_alter_swap": "true",
+        "enable_arrangement_dictionary_compression_alpha": "false",
```
def-, line 74: Why don't we enable it in CI by default? Performance?

PR #32110, opened 2025-04-07
Eliminate a quadraticness in EQProp

ggevay:
This is Frank's https://github.com/MaterializeInc/materialize/pull/32093, with test rewrites and a feature flag.

I also had to change the `ColumnKnowledge performed work after EquivalencePropagation` error to a warning, because it became too common: specifically it's not happening at every startup. (This would have looked bad as an error when one starts up Materialize locally; one might think that there is something wrong. Also, a test framework is verifying that there are no error logs at startup.) (Note that in contrast to `EquivalencePropagation`, `ColumnKnowledge` is able to introduce knowledge about each Map expression as it goes without being quadratic, because it does less work when it introduces knowledge.)

### Motivation

  * This PR fixes a recognized bug: https://github.com/MaterializeInc/database-issues/issues/9144: This was causing 7-minute optimizations for some PowerBI queries.

### Checklist

### DIFF
--- a/src/transform/src/equivalence_propagation.rs
+++ b/src/transform/src/equivalence_propagation.rs
@@ -83,7 +83,9 @@ impl crate::Transform for EquivalencePropagation {
             let ck = crate::ColumnKnowledge::default();
             ck.transform(relation, ctx)?;
             if prior != *relation {
-                tracing::error!(
+                // This used to be tracing::error, but it became too common with
+                // dequadratic_eqprop_map.
+                tracing::warn!(
                     ?ctx.global_id,
                     "ColumnKnowledge performed work after EquivalencePropagation",
                 );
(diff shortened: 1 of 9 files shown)


src/transform/src/equivalence_propagation.rs
```diff
@@ -250,9 +252,12 @@ impl EquivalencePropagation {
...
 
-                if let Some(input_equivalences) = input_equivalences {
-                    // Clone the equivalences in case of variadic map, which will need to mutate them.
-                    let mut input_equivalences = input_equivalences.clone();
+                if let Some(input_equivalences_orig) = input_equivalences {
+                    let mut input_equivalences_cloned = None;
+                    if !ctx.features.enable_dequadratic_eqprop_map {
+                        // We mutate them for variadic Maps if the feature flag is not set.
+                        input_equivalences_cloned = Some(input_equivalences_orig.clone());
+                    }
```
frankmcsherry, line 260: It's unclear from the context and the comments why this optional cloning exists. Given the complexity the PR adds, we should also add the corresponding explanations to know how to dig ourselves back out in the future.

frankmcsherry (approved):

src/transform/src/equivalence_propagation.rs
ggevay: Ok, I've added a comment.

PR #32114, opened 2025-04-07
controller: report compute introspection updates promptly

teskje:
This PR fixes a bug where certain types of introspection updates (frontiers and wallclock lag) would be reported with a 1s delay, causing confusion particularly around comparing compute with storage frontiers.

The bug was caused by the compute controller only reporting frontier advancements to the storage controller once per second. Frontier introspection is also produced once per second, but by the per-instance controllers, and the compute controller couldn't wait on the instance controllers because it had no way to know when they finished emitting frontier updates.

The fix implemented here is to have a "compute introspection sink" task that continually reads from the compute controller-internal `introspection_rx` and pushes the introspection updates to the storage controller. This task cannot hold a reference to the storage controller, so instead it gets initialized with the sending end of the channels connected to the `CollectionManager`. This way it can push updates to the persist writer tasks directly, without going through the `StorageController`.

We need to be careful about when to start the task: During controller initialization, the storage controller doesn't yet know about the introspection collections, so attempting to get channels to send updates for them won't succeed. Instead we need to wait until bootstrapping has progressed until after the storage collections have been initialized from the catalog, at which point we can start the introspection sink task.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9149

### Tips for reviewer

Thoughts about alternative approaches:

* I was hoping that we could pass the `IntrospectionSink`, i.e. all the channels connected to the `CollectionManager`, to the instance controllers directly, and get rid of the compute controller-internal `introspection_{tx,rx}` channel. Unfortunately that doesn't work because when the compute controller and the instances are initialized the storage controller doesn't yet know about introspection collections, and so we can't get the `CollectionManager` channels.
* There is a simpler fix than the one made in this PR, which is to keep the existing `record_introspection_updates` method in the compute controller and run it every time there are updates in `introspection_rx`. For that we just need to adjust the ready/process methods and don't need to introduce a new task, or change the storage controller to provide access to the `CollectionManager` channels. I'm weary of this approach because it means added noise on the coordinator's main thread, but it's something to consider.

### Checklist

### DIFF
--- a/src/compute-client/src/controller.rs
+++ b/src/compute-client/src/controller.rs
@@ -33,7 +33,6 @@ use std::num::NonZeroI64;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
-use differential_dataflow::consolidation::consolidate;
 use futures::stream::FuturesUnordered;
 use futures::{FutureExt, StreamExt};
 use mz_build_info::BuildInfo;
@@ -54,8 +53,8 @@ use mz_ore::collections::CollectionExt;
 use mz_ore::metrics::MetricsRegistry;
 use mz_ore::now::NowFn;
 use mz_ore::tracing::OpenTelemetryContext;
-use mz_repr::{Datum, Diff, GlobalId, Row, TimestampManipulation};
-use mz_storage_client::controller::{IntrospectionType, StorageController, StorageWriteOp};
+use mz_repr::{Datum, GlobalId, Row, TimestampManipulation};
+use mz_storage_client::controller::StorageController;
 use mz_storage_types::dyncfgs::ORE_OVERFLOWING_BEHAVIOR;
 use mz_storage_types::read_holds::ReadHold;
 use mz_storage_types::read_policy::ReadPolicy;
@@ -75,6 +74,7 @@ use crate::controller::error::{
     ReplicaCreationError, ReplicaDropError,
 };
 use crate::controller::instance::{Instance, SharedCollectionState};
+use crate::controller::introspection::{spawn_introspection_sink, IntrospectionUpdates};
 use crate::controller::replica::ReplicaConfig;
 use crate::logging::{LogVariant, LoggingConfig};
 use crate::metrics::ComputeControllerMetrics;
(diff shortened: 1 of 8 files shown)


src/compute-client/src/controller.rs
```diff
@@ -1062,55 +1078,12 @@ where
...
-        let mut updates_by_type = BTreeMap::new();
-
-        for (type_, updates) in self.introspection_rx.try_iter() {
-            updates_by_type
-                .entry(type_)
-                .or_insert_with(Vec::new)
-                .extend(updates);
-        }
-        for updates in updates_by_type.values_mut() {
-            consolidate(updates);
```
antiguru, line 1080: How useful was the consolidation here? We're removing it, but I wonder why it was here in the first place.

src/compute-client/src/controller/introspection.rs
```diff
@@ -0,0 +1,114 @@
...
+pub fn spawn_introspection_sink<T: Timestamp>(
+    mut rx: mpsc::UnboundedReceiver<IntrospectionUpdates>,
+    storage_controller: &dyn StorageController<Timestamp = T>,
+) {
+    let sink = IntrospectionSink::new(storage_controller);
+
+    mz_ore::task::spawn(|| "compute-introspection-sink", async move {
+        info!("running introspection sink task");
+
+        while let Some((type_, updates)) = rx.recv().await {
```
antiguru, line 30: Would it make sense to `recv_many` here, consolidate, and send a batch of updates?

antiguru (approved):
Looks fine, thank you! Is there a another alternative where we would wait for the storage controller to create the collections and then continue working on the compute side of things to avoid the copy task?

src/compute-client/src/controller.rs
teskje: The answer is I don't know! The writer task in the `CollectionManager` also consolidates, so doing it in the compute controller only affects how much data we send over the channels. I'm not sure if I was aware of that when I wrote this code originally. I'm pretty sure I didn't add it because of a performance issue I observed, but because of an abundance of caution.

src/compute-client/src/controller/introspection.rs
teskje: I was considering that but decided against it because it would require collecting updates by `type_` here, which is a bit annoying. I think it wouldn't help too much because the individual `updates` batches should already be consolidated given how the producers work. It would help if we could merge multiple `updates` batches for the same time, but I wouldn't expect to encounter that often, given that this task immediately forwards batches and doesn't wait for them to accumulate.

teskje:
> Is there a another alternative where we would wait for the storage controller to create the collections and then continue working on the compute side of things to avoid the copy task?

Maybe, though I'm not sure how that waiting would look like in the current code. When you create a `CollectionState` in the compute controller, it wants a tx to send introspection updates on, so we'd have to delay creation of all compute collections until the storage collections are initialized. But during bootstrapping the coordinator creates cluster first, and each cluster creation implicitly causes the creation of log collection state. We might be able change the bootstrap process to initialize the introspection storage collections before creating the clusters in the controller, but that seems quite brittle and very easy to mess up later.

teskje:
TFTR!

PR #32118, opened 2025-04-07
docs: prefer mat views when using temporal filters

kay-kim:


doc/user/layouts/shortcodes/views-indexes/table-usage-pattern.html
```diff
@@ -3,6 +3,7 @@
 | Usage Pattern | General Guideline |
 |--------------------------------------------------------------------------------|--------------------|
 | View results are accessed from a single cluster only                                         | View with an [index](/sql/create-index) |
-| View results are accessed across [clusters](/concepts/clusters)                            | Materialized view  |
+| View results are accessed across [clusters](/concepts/clusters)<br>(e.g., queries are served from a cluster different from the cluster that maintains the view results) | Materialized view  |
```
kay-kim, line 6: https://preview.materialize.com/materialize/32118/concepts/indexes/#indexes-on-views-vs-materialized-views

doc/user/content/transform-data/patterns/temporal-filters.md
```diff
@@ -52,20 +52,51 @@ moves forward:
...
+
+- The `mz_now()` only accepts comparison operators `=`, `<`, `<=`, `>`, or `>=`,
+  or operators that desugar to them or a conjunction of them (for example,
+  `BETWEEN...AND...`); i.e., you cannot use `mz_now()` with date/time operators.
 
-You cannot use temporal filters in the `WHERE` clause of an [aggregate `FILTER` expression](/sql/functions/filters).
+{{< tip >}}
+
+When possible, prefer materialized views (instead of indexed views) when using
+temporal filter to take advantage of custom consolidation.
```
kay-kim, line 84: It's a little clunky here ...

doc/user/content/transform-data/patterns/temporal-filters.md
```diff
@@ -52,20 +52,51 @@ moves forward:
...
 ## Examples
 
 These examples create real objects.
 After you have tried the examples, make sure to drop these objects and spin down any resources you may have created.
 
+{{< tip >}}
+
+When possible, prefer materialized views when using temporal filter to take
+advantage of custom consolidation.
+
```
kay-kim, line 97: Same clunky tip.

doc/user/layouts/shortcodes/views-indexes/table-usage-pattern.html
antiguru: Maybe we can extract the text so the column doesn't get too wide? Or, we could change the whole block to a list. It gets hard to read if the browser doesn't have a whole screen available (or the screen is small).

antiguru (approved):
The change seems fine, and unrelated it's something we might want to fix in Materialize in the future.

PR #32128, opened 2025-04-08
Extract `BytesContainer` from `DatumContainer`

frankmcsherry:
Extract a container from `DatumContainer` that can be safely used for arbitrary byte slices, rather than only for `Row`-encoded byte slices.

### Checklist

### DIFF
--- a/src/compute/src/row_spine.rs
+++ b/src/compute/src/row_spine.rs
@@ -99,32 +99,30 @@ mod spines {
 
 /// A `Row`-specialized container using dictionary compression.
 mod container {
-    use differential_dataflow::IntoOwned;
+
     use std::cmp::Ordering;
 
     use differential_dataflow::trace::implementations::BatchContainer;
-    use mz_ore::region::Region;
-    use mz_repr::{read_datum, Datum, Row, RowPacker};
+    use differential_dataflow::IntoOwned;
     use timely::container::PushInto;
 
-    /// A slice container with four bytes overhead per slice.
+    use mz_repr::{read_datum, Datum, Row, RowPacker};
+
+    use super::bytes_container::BytesContainer;
+
+    /// Container wrapping `BytesContainer` that traffics only in `Row`-formatted bytes.
+    ///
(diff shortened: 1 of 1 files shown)


src/compute/src/row_spine.rs
```diff
@@ -134,48 +132,28 @@ mod container {
 
         fn with_capacity(size: usize) -> Self {
             Self {
-                batches: vec![DatumBatch::with_capacities(size, size)],
+                bytes: BytesContainer::with_capacity(size),
```
antiguru, line 135: Oof, what does the size even mean? Anyway, not part of this PR, so ok to leave it as-is.

src/compute/src/row_spine.rs
```diff
@@ -99,32 +99,30 @@ mod spines {
...
 mod container {
-    use differential_dataflow::IntoOwned;
+
     use std::cmp::Ordering;
 
     use differential_dataflow::trace::implementations::BatchContainer;
-    use mz_ore::region::Region;
-    use mz_repr::{read_datum, Datum, Row, RowPacker};
+    use differential_dataflow::IntoOwned;
     use timely::container::PushInto;
```
antiguru, line 108:
```suggestion
```

Nit: `use` formatting.

frankmcsherry: Will clean up in the next PR, if that's ok. Never sure where we want line break/don't (e.g. should line 104 be deleted too?).

src/compute/src/row_spine.rs
frankmcsherry: No idea, but I think we should be in a position to make it mean something more concrete in the near future (next PR should involve right-sizing the allocations).

frankmcsherry:
Thanks for the review; will address nits soon!

src/compute/src/row_spine.rs
teskje: There is a section about this in our style guidelines that may help against the uncertainty: https://github.com/MaterializeInc/materialize/blob/main/doc/developer/style.md#imports

PR #32131, opened 2025-04-08
Password auth 

DAlperin:
This is the bones of self managed password auth. This is missing integration tests and documentation which will come next. You can test this locally like so:
```shell
#this starts materialize with auth enabled on the external port but not the internal port
$ bin/environmentd --bazel --reset -- --all-features --unsafe-mode --enable-self-hosted-auth

#start a mz_system session, turn on sql support and create a user
$ psql -U mz_system -h localhost -p 6877 materialize
NOTICE:  connected to Materialize v0.139.0-dev.0
  Org ID: 1bd2c405-c638-44cc-b917-6d05dfb832ac
  Region: local/az1
  User: mz_system
  Cluster: mz_system
  Database: materialize
  Schema: public
  Session UUID: fe6d2dbe-3a94-430d-82e1-5e79cb14b91f

Issue a SQL query to get started. Need help?
  View documentation: https://materialize.com/s/docs
  Join our Slack community: https://materialize.com/s/chat
    
psql (14.17 (Homebrew), server 9.5.0)
Type "help" for help.

materialize=> alter system set enable_self_managed_auth=true;
NOTICE:  variable "enable_self_managed_auth" was updated for the system, this will have no effect on the current session
ALTER SYSTEM
materialize=> create role foo with superuser password 'bar';
CREATE ROLE

# Now connect over the port with auth enabled
$ psql -U foo -h localhost -p 6875 materialize
Password for user foo: 

```

This begins implementing https://github.com/MaterializeInc/materialize/pull/32005

### Checklist


src/auth/src/hash.rs
```diff
@@ -0,0 +1,227 @@
...
+#[derive(Debug)]
+pub enum VerifyError {
+    MalformedHash,
+    InvalidPassword,
+}
+
+/// Hashes a password using PBKDF2 with SHA256
+/// and a random salt.
+pub fn hash_password(password: &String) -> PasswordHash {
+    let mut salt = [0u8; 32];
```
DAlperin, line 55: oops, should use the const here

src/auth/src/hash.rs
```diff
@@ -0,0 +1,227 @@
...
+        iterations: hashed_password.iterations,
+        salt: hashed_password.salt,
+        server_key: server_key.as_ref().try_into().unwrap(),
+        client_key: client_key.as_ref().try_into().unwrap(),
+    }
+}
+
+fn hash_password_inner(opts: &HashOpts, password: &[u8]) -> [u8; SHA256_OUTPUT_LEN] {
+    let mut salted_password = [0u8; SHA256_OUTPUT_LEN];
+    pbkdf2::derive(
```
DAlperin, line 180: Not technically constant time, but I don't think we are really subject to timing attacks here

src/pgwire/src/protocol.rs
```diff
@@ -211,13 +216,55 @@ where
...
+            Some(FrontendMessage::Password { password }) => password,
+            _ => {
+                return conn
+                    .send(ErrorResponse::fatal(
+                        SqlState::INVALID_AUTHORIZATION_SPECIFICATION,
+                        "expected Password message",
+                    ))
+                    .await
+            }
+        };
```
DAlperin, line 233: Could potentially consolidate some of these codepaths between frontegg and self managed but probably fine for now

src/sql-parser/src/parser.rs
```diff
@@ -4178,10 +4179,15 @@ impl<'a> Parser<'a> {
                 Some(NOCREATEDB) => options.push(RoleAttribute::NoCreateDB),
                 Some(CREATEROLE) => options.push(RoleAttribute::CreateRole),
                 Some(NOCREATEROLE) => options.push(RoleAttribute::NoCreateRole),
+                Some(PASSWORD) => {
+                    //TODO(dov): allow unsetting password with null
```
DAlperin, line 4183: is this a blocker?

src/pgwire/src/protocol.rs
```diff
@@ -211,13 +216,55 @@ where
...
+            user,
+            client_ip: conn.peer_addr().clone(),
+            external_metadata_rx: None,
+            internal_user_metadata: Some(InternalUserMetadata {
+                superuser: auth_response.superuser,
+            }),
+            helm_chart_version,
+        });
+        // No frontegg check, so auth session lasts indefinitely.
+        let auth_session = pending().right_future();
```
ParkMyCar, line 258: What does Postgres do if a superuser changes the password of a currently authenticated user? Not that we need to handle this now, but seems like a place where we could periodically check for this kind of change?

src/sql-parser/src/parser.rs
ParkMyCar: IMO not for merging this PR, but we'd probably want to add this before releasing to self-hosted customers? Not having a way to unset a password feels bad!

src/sql/src/catalog.rs
```diff
@@ -485,6 +485,10 @@ pub trait CatalogSchema {
 pub struct RoleAttributes {
     /// Indicates whether the role has inheritance of privileges.
     pub inherit: bool,
+    /// The password for the role
+    pub password: Option<String>,
```
ParkMyCar, line 489: Is this the plaintext password or a hashed version? Can you also update the comment to mention this separate from the Frontegg/external auth used in the cloud product

src/sql-parser/src/parser.rs
jubrad: Does setting a password make the role a loginrole? If so we should definitely support unset.

test/sqllogictest/role_create.slt
```diff
@@ -63,7 +63,7 @@ COMPLETE 0
 simple conn=regress_role_limited_admin,user=regress_role_limited_admin
 CREATE ROLE regress_nosuch_superuser SUPERUSER;
 ----
-db error: ERROR: SUPERUSER attribute is not supported, for more information consult the documentation at https://materialize.com/docs/sql/create-role/#details
+db error: ERROR: self-managed auth is not enabled, setting password or superuser is not allowed are not supported
```
ParkMyCar, line 66: This doesn't seem right, ideally for a cloud customer we shouldn't mention anything about self-managed :)

src/adapter/src/client.rs
```diff
@@ -148,6 +150,22 @@ impl Client {
         Session::new(self.build_info, config, self.metrics().session_metrics())
     }
 
+    /// Preforms an authentication check for the given user.
+    pub async fn authenticate(
+        &self,
+        user: &String,
+        password: &String,
```
ParkMyCar, line 157: Given this is the plaintext password and it's plumbed through a number of layers, I feel pretty strongly that we should wrap it in a newtype that purposefully doesn't implement `Debug` or `Display`, we can also use the [`assert_not_impl_all!`](https://docs.rs/static_assertions/1.1.0/static_assertions/macro.assert_not_impl_all.html) macro!

FWIW I realize now that we don't do this for Frontegg App Passwords, but definitely should

src/auth/src/hash.rs
```diff
@@ -0,0 +1,227 @@
...
+    #[mz_ore::test]
+    fn test_scram256_hash() {
+        let password = "password".to_string();
+        let scram_hash = scram256_hash(&password);
+
+        let res = scram256_verify(&password, &scram_hash);
+        assert!(res.is_ok());
+        let res = scram256_verify(&"wrong_password".to_string(), &scram_hash);
+        assert!(res.is_err());
+    }
```
ParkMyCar, line 212: Turning this into a `proptest!` would be nice. Guards against any odd behavior with string handling, e.g. `$` or ` ` in the password itself

src/environmentd/src/environmentd/main.rs
```diff
@@ -209,7 +209,12 @@ pub struct Args {
     /// Frontegg arguments.
     #[clap(flatten)]
     frontegg: FronteggCliArgs,
-
+    /// Self hosted auth
+    #[clap(long, env = "ENABLE_SELF_HOSTED_AUTH")]
+    enable_self_hosted_auth: bool,
+    /// Self hosted auth over internal port
+    #[clap(long, env = "ENABLE_SELF_HOSTED_AUTH_INTERNAL")]
+    enable_self_hosted_auth_internal: bool,
```
ParkMyCar, line 217: Ignore me if this was part of a design with the Cloud team, but I'm not sure how these are intended to be used? How do these relate with the `enable_self_managed_auth` dyncfg?

ParkMyCar (commented):
Nice!! Excited to see this make progress!!

A few things that I think are missing:

1. More SLT tests. Would love to see some tests that exercise the logic of who can `ALTER` passwords for a role. Can a superuser? Can the user themselves? Can an unrelated user? Also a test that `DROP`s a role with a password, I'm not totally sure what happens here, it doesn't seem like we ever remove elements from the `role_auth` collection?
2. Would be great to add `role_auth` to our [consistency checks](https://github.com/MaterializeInc/materialize/blob/main/src/adapter/src/catalog/consistency.rs) these run after most SLT and testdrive tests to ensure relationships like `roles <-> role_auth` are correct.
3. Doesn't have to be in this PR, but a quick [platform-check](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/platform-checks.md) that creates a `ROLE` with a `PASSWORD` and later alters it would be great! Platform checks get run in all different kinds of scenarios including across restarts and upgrades so they exercise a lot of code paths.

I know we're excited to get this in so I won't block but would love to see at least [1] handled before merging :)

src/pgwire/src/protocol.rs
DAlperin: It seems to not handle it all as best I can tell. We can probably improve on this though

src/adapter/src/client.rs
ptravers: I agree and would add that this should not be merged without that.

src/adapter/src/command.rs
```diff
@@ -66,6 +66,12 @@ pub enum Command {
         notice_tx: mpsc::UnboundedSender<AdapterNotice>,
     },
 
+    AuthCheck {
```
ptravers, line 69: nit: AuthCheck -> AuthenticatePassword

src/adapter/src/coord/command_handler.rs
```diff
@@ -236,6 +244,45 @@ impl Coordinator {
         .boxed_local()
     }
 
+    #[mz_ore::instrument(level = "debug")]
+    async fn handle_auth_check(
```
ptravers, line 248: nit: handle_authenticate_password

src/environmentd/src/environmentd/main.rs
DAlperin: Probably gonna punt this to a follow up PR

src/auth/src/hash.rs
ptravers: is there another document where we outline the set of attacks we are defending against? I can't see anything in [the design](https://github.com/MaterializeInc/materialize/pull/32005/files?short_path=076932b#diff-076932baba2624ac63805005038b0708b0fe4973ca680047cde716a96bdc74a0)

src/auth/src/hash.rs
```diff
@@ -0,0 +1,227 @@
...
+// by the Apache License, Version 2.0.
+
+use std::fmt::Display;
+use std::num::NonZeroU32;
+
+use base64::prelude::*;
+use ring::digest::SHA256_OUTPUT_LEN;
+use ring::hmac::{self, Key, HMAC_SHA256};
+use ring::pbkdf2::{self, PBKDF2_HMAC_SHA256 as SHA256};
+use ring::rand::SecureRandom;
```
ptravers, line 17: we are using BoringSSL here or a repackaging of it[1]. BoringSSL is not FIPS certified[2]. in past positions we have declined to use ring/BoringSSL. it's worth noting that neither ring nor BoringSSL recommend the usage of their code. in past positions we have declined to use both these libraries. what security guarantees are we planning to give to our customers?

[1] https://crates.io/crates/ring
[2] https://boringssl.googlesource.com/boringssl/+/master/crypto/fipsmodule/FIPS.md

ptravers: I retract part of this as it looks like the algorithms in boringssl are FIPS certified. so I guess this is fine?

ptravers: spoke offline with @DAlperin we are migrating to [aws-lc-rs](https://github.com/aws/aws-lc-rs) which has passed FIPs certification with an approved lab at least for a [specific branch](https://github.com/aws/aws-lc/tree/fips-2024-09-27) of the repo.

it has **not** been approved by NIST as of 04/10/2025 for current status, see: https://csrc.nist.gov/Projects/cryptographic-module-validation-program/modules-in-process/Modules-In-Process-List

src/auth/src/hash.rs
```diff
@@ -0,0 +1,230 @@
...
+
+/// The default iteration count as suggested by
+/// https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html
+const DEFAULT_ITERATIONS: NonZeroU32 = NonZeroU32::new(600_000).unwrap();
```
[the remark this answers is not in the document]
ptravers: I'm still reading through NIST and references in that website to try to understand why they picked that value.

src/auth/src/hash.rs
```diff
@@ -0,0 +1,230 @@
...
+
+#[derive(Debug)]
+pub enum VerifyError {
+    MalformedHash,
+    InvalidPassword,
+}
+
+/// Hashes a password using PBKDF2 with SHA256
+/// and a random salt.
+pub fn hash_password(password: &String) -> PasswordHash {
```
ptravers, line 57: nit: should this also be `Password`

```suggestion
pub fn hash_password(password: &Password) -> PasswordHash {
```

src/auth/src/hash.rs
```diff
@@ -0,0 +1,230 @@
...
+    PasswordHash {
+        salt,
+        iterations: DEFAULT_ITERATIONS,
+        hash,
+    }
+}
+
+/// Hashes a password using PBKDF2 with SHA256
+/// and the given options.
+pub fn hash_password_with_opts(opts: &HashOpts, password: &String) -> PasswordHash {
```
ptravers: nit: should this also be `Password`

ParkMyCar (commented):
Overall looks good to me! Don't forget to "Squash and Merge" :)

A few comments that would be nice to address but aren't blocking. The only thing we need to do is remove "OpenSSL" from our list of approved licenses, once I see that I'll approve!!

src/auth/src/password.rs
```diff
@@ -0,0 +1,44 @@
...
+use serde::{Deserialize, Serialize};
+use std::fmt::{Debug, Display};
+
+use static_assertions::assert_not_impl_all;
+
+///Password is a String wrapper type that does not implement Display or Debug
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize, PartialOrd, Ord, Arbitrary)]
+pub struct Password(String);
+
+assert_not_impl_all!(Password: Display);
```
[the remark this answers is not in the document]
DAlperin: It does impl Debug, but we implement it manually so it always shows `Password(****)`. This is due to the fact that `Password` gets carried around in lots of structs with `#[derive(Debug)]`

PR #32132, opened 2025-04-08
[adapter] use a `SnapshotCursor` and consolidated snapshot for builtin table retractions

ParkMyCar:
When the Coordinator starts we list the entire contents of each builtin table and retract them to make table migrations easier. This PR changes the retractions to operate on a consolidated snapshot as opposed to an unconsolidated one. 

We've observed (https://github.com/MaterializeInc/database-issues/issues/9155) that after restarting Materialize a number of times in a row (>100) startup becomes very slow and we suspect this is because when Materialize is restarted in a loop we never get a chance to consolidate the builtin shards.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9155

### Checklist

### DIFF
--- a/src/adapter/src/coord.rs
+++ b/src/adapter/src/coord.rs
@@ -2484,10 +2484,10 @@ impl Coordinator {
             debug!("coordinator init: resetting system table {full_name} ({table_id})");
 
             // Fetch the current contents of the table for retraction.
-            let stream_fut = self
+            let snapshot_fut = self
                 .controller
                 .storage_collections
-                .snapshot_and_stream(system_table.table.global_id_writes(), read_ts);
+                .snapshot_cursor(system_table.table.global_id_writes(), read_ts);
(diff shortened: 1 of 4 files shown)


ParkMyCar:
Currently running the repro from https://github.com/MaterializeInc/database-issues/issues/9155 to see if this fixes the issue

bkirwi (approved):
Great!

This does increase the memory pressure a bit... but lools like we do one system table at a time here, so it ought to be fairly safe. Worth keeping an eye on in the next release I guess.

ParkMyCar:
> Great!
> 
> This does increase the memory pressure a bit... but lools like we do one system table at a time here, so it ought to be fairly safe. Worth keeping an eye on in the next release I guess.

We do actually snapshot all of the builtin tables in parallel. I'll put up a second PR that adds some concurrency limiting!

PR #32133, opened 2025-04-08
[sql_server] Initial implementation for Source Rendering

ParkMyCar:
This PR adds an MVP implementation of the `SourceRender` trait for the SQL Server source. It also adds a testdrive test that exercises replicating data from SQL Server into Materialize.

TODO(parkmycar): I need to add some more detail here

### Motivation

### Checklist

### DIFF
--- a/src/sql-server-util/src/cdc.rs
+++ b/src/sql-server-util/src/cdc.rs
@@ -51,6 +51,13 @@ pub struct CdcStream<'a> {
     capture_instances: BTreeMap<Arc<str>, Option<Lsn>>,
     /// How often we poll the upstream for changes.
     poll_interval: Duration,
+    /// How long we'll wait for SQL Server to return a max LSN before taking a snapshot.
+    ///
+    /// Note: When CDC is first enabled in an instance of SQL Server it can take a moment
+    /// for it to "completely" startup. Before starting a `TRANSACTION` for our snapshot
+    /// we'll wait this duration for SQL Server to report an LSN and thus indicate CDC is
+    /// ready to go.
+    max_lsn_wait: Duration,
 }
 
 impl<'a> CdcStream<'a> {
@@ -62,6 +69,7 @@ impl<'a> CdcStream<'a> {
             client,
             capture_instances,
             poll_interval: Duration::from_secs(1),
+            max_lsn_wait: Duration::from_secs(10),
         }
     }
 
@@ -83,6 +91,18 @@ impl<'a> CdcStream<'a> {
     /// Default is 1 second.
     pub fn poll_interval(mut self, interval: Duration) -> Self {
         self.poll_interval = interval;
+        self
+    }
+
+    /// The max duration we'll wait for SQL Server to return an LSN before taking a
+    /// snapshot.
+    ///
+    /// When CDC is first enabled in SQL Server it can take a moment before it is fully
+    /// setup and starts reporting LSNs.
+    ///
+    /// Default is 10 seconds.
+    pub fn max_lsn_wait(mut self, wait: Duration) -> Self {
+        self.max_lsn_wait = wait;
         self
     }
 
@@ -121,7 +141,7 @@ impl<'a> CdcStream<'a> {
         // the LSN from progressing, we want to wait a bit for SQL Server to
         // become ready.
         let (_client, lsn_result) = mz_ore::retry::Retry::default()
-            .max_duration(std::time::Duration::from_secs(10))
+            .max_duration(self.max_lsn_wait)
             .retry_async_with_state(&mut self.client, |_, client| async {
                 use mz_ore::retry::RetryResult;
                 let result = crate::inspect::get_max_lsn(*client).await;
--- a/src/sql-server-util/src/cdc.rs
+++ b/src/sql-server-util/src/cdc.rs
@@ -7,6 +7,25 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+//! Replicate a table from SQL Server using their Change-Data-Capture (CDC) primitives.
+//!
+//! This module provides a [`CdcStream`] type that provides the following API for
+//! replicating a table:
+//!
+//! 1. [`CdcStream::snapshot`] returns an initial snapshot of a table and the [`Lsn`] at
+//!    which the snapshot was taken.
+//! 2. [`CdcStream::into_stream`] returns a [`futures::Stream`] of [`CdcEvent`]s
+//!    optionally from the [`Lsn`] returned in step 1.
+//!
+//! Internally we get a snapshot by setting our transaction isolation level to
+//! [`TransactionIsolationLevel::Snapshot`], getting the current maximum LSN with
+//! [`crate::inspect::get_max_lsn`] and then running a `SELECT *`. We've observed that by
+//! using [`TransactionIsolationLevel::Snapshot`] the LSN remains stable for the entire
+//! transaction.
+//!
+//! After completing the snapshot we use [`crate::inspect::get_changes_asc`] which will return
+//! all changes between a `[lower, upper)` bound of [`Lsn`]s.
+
 use std::collections::{BTreeMap, BTreeSet};
 use std::fmt;
 use std::sync::Arc;
@@ -16,7 +35,6 @@ use derivative::Derivative;
 use futures::{Stream, StreamExt};
 use proptest_derive::Arbitrary;
 use serde::{Deserialize, Serialize};
-use timely::progress::Timestamp;
 
 use crate::{Client, SqlServerError, TransactionIsolationLevel};
 
@@ -161,7 +179,22 @@ impl<'a> CdcStream<'a> {
         async_stream::try_stream! {
             // Initialize all of our start LSNs.
             self.initialize_start_lsns().await?;
-            let mut max_observed_lsn: Lsn = Lsn::minimum();
+
+            // When starting the stream we'll emit one progress event if we've already observed
+            // everything the DB currently has.
+            if let Some(starting_lsn) = self.capture_instances.values().filter_map(|x| *x).min() {
+                let db_curr_lsn = crate::inspect::get_max_lsn(self.client).await?;
+                let next_lsn = db_curr_lsn.increment();
+                if starting_lsn >= db_curr_lsn {
+                    tracing::debug!(
+                        %starting_lsn,
+                        %db_curr_lsn,
+                        %next_lsn,
+                        "yielding initial progress",
+                    );
+                    yield CdcEvent::Progress { next_lsn };
+                }
+            }
 
             loop {
                 // Measure the tick before we do any operation so the time it takes
@@ -235,7 +268,7 @@ impl<'a> CdcStream<'a> {
                     }
 
                     // Increment our LSN (`get_changes` is inclusive).
-                    max_observed_lsn = db_max_lsn;
+                    //
                     // TODO(sql_server2): We should occassionally check to see how close the LSN we
                     // generate is to the LSN returned from incrementing via SQL Server itself.
                     let next_lsn = db_max_lsn.increment();
@@ -251,16 +284,6 @@ impl<'a> CdcStream<'a> {
                         let instance_lsn = instance_lsn.as_mut().expect("should be initialized");
                         // Ensure LSNs don't go backwards.
                         *instance_lsn = std::cmp::max(*instance_lsn, next_lsn);
-                    }
-                } else if curr_lsn >= db_max_lsn {
-                    // We've ingested everything the DB currently has. Emit a progress
-                    // event in case we haven't already.
-                    //
-                    // Note: Generally this only ever occurs when starting a stream and the LSN
-                    // we're starting with matches the current max of the DB.
-                    if db_max_lsn > max_observed_lsn {
-                        max_observed_lsn = db_max_lsn;
-                        yield CdcEvent::Progress { next_lsn: db_max_lsn.increment() };
                     }
                 }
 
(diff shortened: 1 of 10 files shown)


src/storage/src/source/sql_server/replication.rs
```diff
@@ -0,0 +1,337 @@
...
+                .resolve_config(
+                    &config.config.connection_context.secrets_reader,
+                    &config.config,
+                    InTask::Yes,
+                )
+                .await?;
+            let (mut client, connection) =
+                mz_sql_server_util::Client::connect(connection_config).await?;
+            // TODO(sql_server1): Move the connection into its own future.
+            mz_ore::task::spawn(|| "sql_server-connection", async move { connection.await });
```
petrosagg, line 92: What does the TODO refer to? It seems to be its own future already. Also the async block can be reduced to just `connection`: 

```suggestion
            mz_ore::task::spawn(|| "sql_server-connection", connection);
```

src/storage/src/source/sql_server/replication.rs
```diff
@@ -0,0 +1,337 @@
...
+                let snapshot_cap = data_cap_set.delayed(&snapshot_lsn);
+
+                let mut snapshot_streams = std::pin::pin!(snapshot_streams);
+                while let Some((capture_instance, data)) = snapshot_streams.next().await {
+                    let sql_server_row = data.map_err(TransientError::from)?;
+
+                    // Decode the SQL Server row into an MZ one.
+                    let Some((partition_idx, decoder)) = capture_instances.get(&capture_instance)
+                    else {
+                        let definite_error = DefiniteError::ProgrammingError(format!(
```
petrosagg, line 137: What do you mean here by `ProgrammingError`? Is this condition only hit through a bug? If yes then I wouldn't make it a definite error since that will permanently break the source.

src/storage/src/source/sql_server/replication.rs
```diff
@@ -0,0 +1,337 @@
...
+                            data_output,
+                            data_cap_set,
+                            definite_error_handle,
+                            definite_error_cap_set,
+                        )
+                        .await;
+                        return Ok(());
+                    };
+
+                    // Failing to decode data is a permanent failure.
```
petrosagg, line 152: It's not a permanent failure if the problematic row is later retracted so what we want to do here is emit the error in the output and continue as if nothing happened. The error message should contain enough data for the user to locate the problematic row (perhaps a raw representation)

src/storage/src/source/sql_server/replication.rs
```diff
@@ -0,0 +1,337 @@
...
+                            ((*partition_idx, Ok(message)), snapshot_lsn, Diff::ONE),
+                        )
+                        .await;
+                }
+
+                snapshot_lsn
+            };
+
+            // Start replicating from the LSN __after__ we took the snapshot.
+            let replication_start_lsn = snapshot_lsn.increment();
```
petrosagg, line 187: Is there any documentation to link somewhere in our code that explains exactly what happens during snapshots and whether LSNs are inclusive/exclusive?

src/storage/src/source/sql_server/replication.rs
```diff
@@ -0,0 +1,337 @@
...
+                        cdc_handle =
+                            cdc_handle.start_lsn(&output_info.capture_instance, initial_lsn);
+                    }
+                    _ => unreachable!(""),
+                }
+            }
+
+            // Off to the races! Replicate data from SQL Server.
+            let cdc_stream = cdc_handle
+                .poll_interval(std::time::Duration::from_millis(500))
```
petrosagg, line 209: This should probably become a DYNCFG

src/storage/src/source/sql_server/replication.rs
```diff
@@ -0,0 +1,337 @@
...
+                        &output_indexes[..],
+                        data_output,
+                        data_cap_set,
+                        definite_error_handle,
+                        definite_error_cap_set,
+                    )
+                    .await;
+                    return Ok(());
+                };
+                let data_cap = data_cap_set.delayed(&commit_lsn);
```
petrosagg, line 251: You don't need to call delayed here to get an exact capability. Any capability that is less than or equal to the LSN you wish to emit will do so you can just take `data_cap_set[0]` here.

More generally, when you create a capability there is a small overhead of capability tracking to be paid which becomes noticeable if done per-message

src/storage/src/source/sql_server/replication.rs
```diff
@@ -0,0 +1,337 @@
...
+                    let (sql_server_row, diff): (_, _) = match change {
+                        CdcOperation::Insert(sql_server_row)
+                        | CdcOperation::UpdateNew(sql_server_row) => (sql_server_row, Diff::ONE),
+                        CdcOperation::Delete(sql_server_row)
+                        | CdcOperation::UpdateOld(sql_server_row) => {
+                            (sql_server_row, Diff::MINUS_ONE)
+                        }
+                    };
+
+                    // Failing to decode data is a permanent failure.
```
petrosagg, line 263: Same as above, we should continue without halting here

src/sql-server-util/src/cdc.rs
```diff
@@ -310,60 +356,114 @@ pub enum CdcError {
...
+            None => panic!("overflowed Lsn, {self:?}"),
+        }
     }
 }
 
-impl PartialOrd for Lsn {
-    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        Some(self.cmp(other))
+impl fmt::Display for Lsn {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
petrosagg, line 453: Is there a more concise format than the debug view? This will be used to log frontiers and it's going to be annoying if it gets printed in such detail. In some online resources I've seen the format `format!({}:{}:{}", self.vlf_id, self.block_id, self.record_id)` which seems nice and compact. https://www.sqlshack.com/go-lsn-in-sql-server/

src/sql-server-util/src/cdc.rs
```diff
@@ -310,60 +356,114 @@ pub enum CdcError {
...
-}
 
-impl Ord for Lsn {
-    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
-        self.as_structured().cmp(&other.as_structured())
+    /// Increment this [`Lsn`].
+    ///
+    /// The returned [`Lsn`] may not exist upstream yet, but it's guaranteed to
+    /// sort greater than `self`.
+    pub fn increment(self) -> Lsn {
```
petrosagg, line 420: If you want a more concise impl, std has a nice function that can give you back the carry to add to the rest of the fields:

```rust
let (record_id, carry) = self.record_id.overflowing_add(1);
let (block_id, carry) = self.block_id.overflowing_add(carry as u32);
let (vlf_id, overflow) = self.vlf_id.overflowing_add(carry as u32);
assert!(!overflow, "overflowed Lsn, {self:?}");
Lsn { vlf_id, block_id, record_id) }
```

src/sql-server-util/src/cdc.rs
```diff
@@ -310,60 +356,114 @@ pub enum CdcError {
...
-        Lsn(bytes)
-    }
+    const SIZE: usize = 10;
 
-    /// Return the underlying byte slice for this [`Lsn`].
-    pub fn as_bytes(&self) -> &[u8] {
-        self.0.as_slice()
-    }
+    /// Interpret the provided bytes as an [`Lsn`].
+    pub fn try_from_bytes(bytes: &[u8]) -> Result<Self, String> {
```
petrosagg, line 389: If there are any links to documentation about the internal structure of this field including things like endianess it would be great to include them

src/sql-server-util/src/cdc.rs
```diff
@@ -310,60 +356,114 @@ pub enum CdcError {
 /// function.
 ///
 /// [`sys.fn_cdc_increment_lsn`](https://learn.microsoft.com/en-us/sql/relational-databases/system-functions/sys-fn-cdc-increment-lsn-transact-sql?view=sql-server-ver16)
```
petrosagg, line 358: This doc comment block needs some updates since LSNs are not treated as opaque blobs anymore and we increment them without calling this function

src/sql-server-util/src/cdc.rs
```diff
@@ -95,6 +96,30 @@ impl<'a> CdcStream<'a> {
...
         tracing::info!(?tables, "got table for capture instance");
 
+        // Before we start a transaction, ensure SQL Server is reporting an LSN.
+        //
+        // If CDC was very recently enabled (e.g. < 30 seconds ago) then it will
+        // report NULL for an LSN. Before we start a transaction, which prevents
+        // the LSN from progressing, we want to wait a bit for SQL Server to
+        // become ready.
+        let (_client, lsn_result) = mz_ore::retry::Retry::default()
+            .max_duration(std::time::Duration::from_secs(10))
```
petrosagg, line 106: Does this give 10 seconds to the query to come back? The default might be too tight, and definitely worth a dyncfg

src/sql-server-util/src/cdc.rs
```diff
@@ -209,7 +235,10 @@ impl<'a> CdcStream<'a> {
                     }
 
                     // Increment our LSN (`get_changes` is inclusive).
-                    let next_lsn = crate::inspect::increment_lsn(self.client, new_lsn).await?;
+                    max_observed_lsn = db_max_lsn;
+                    // TODO(sql_server2): We should occassionally check to see how close the LSN we
+                    // generate is to the LSN returned from incrementing via SQL Server itself.
```
petrosagg, line 240: Out of curiosity, what will we do with this information?

src/sql-server-util/src/cdc.rs
```diff
@@ -175,7 +201,7 @@ impl<'a> CdcStream<'a> {
                             self.client,
                             &*instance,
                             *instance_lsn,
-                            new_lsn,
+                            db_max_lsn,
                             RowFilterOption::AllUpdateOld,
                         )
                         // TODO(sql_server3): Make this chunk size configurable.
```
petrosagg, line 207: Not part of this PR, but the code below seems to be doing some grouping by LSNs. Timely doesn't care about the order of LSNs in the data stream as long as progress statements are correct so we can directly iterate and emit the data as they come.

That said, in which case does SQL Server send us data out of order?

src/sql-server-util/src/cdc.rs
```diff
@@ -223,6 +252,16 @@ impl<'a> CdcStream<'a> {
                         // Ensure LSNs don't go backwards.
                         *instance_lsn = std::cmp::max(*instance_lsn, next_lsn);
                     }
+                } else if curr_lsn >= db_max_lsn {
+                    // We've ingested everything the DB currently has. Emit a progress
+                    // event in case we haven't already.
+                    //
+                    // Note: Generally this only ever occurs when starting a stream and the LSN
+                    // we're starting with matches the current max of the DB.
```
petrosagg, line 260: I think this else block can be removed entirely and instead just emit a progress event at the beginning before entering the replication loop. Or don't emit it at all and let the caller assume it

src/storage/src/source/sql_server/replication.rs
```diff
@@ -0,0 +1,337 @@
...
+use timely::dataflow::{Scope, Stream as TimelyStream};
+use timely::progress::{Antichain, Timestamp};
+
+use crate::source::RawSourceCreationConfig;
+use crate::source::sql_server::{
+    DefiniteError, ReplicationError, SourceOutputInfo, TransientError,
+};
+use crate::source::types::{SignaledFuture, SourceMessage, StackedCollection};
+
+/// Used as a partition ID to determint the worker that is responsible for
```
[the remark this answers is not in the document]
ParkMyCar: Thanks!

src/storage/src/source/sql_server/replication.rs
```diff
@@ -0,0 +1,337 @@
...
+                })
+                .collect();
+
+            // Map from a SQL Server 'capture instance' to Materialize collection.
+            let capture_instances: BTreeMap<_, _> = outputs
+                .values()
+                .map(|output| {
+                    (
+                        Arc::clone(&output.capture_instance),
+                        (output.partition_index, Arc::clone(&output.decoder)),
```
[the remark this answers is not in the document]
ParkMyCar: Chatted offline, I have a local branch that does what Petros describes, I'll make this change in a follow-up PR

src/storage/src/source/sql_server/replication.rs
ParkMyCar: Normally here you would do a `.expect(...)` but in case of a bug I don't want to panic `clusterd`. Switched to a `TransientError`!

src/storage/src/source/sql_server/replication.rs
ParkMyCar: Added some documentation on the `Lsn` struct and the `mz_sql_server_util::inspect::get_changes` method.

src/storage/src/source/sql_server/replication.rs
ParkMyCar: Done!

src/storage/src/source/sql_server/replication.rs
ParkMyCar: Makes sense! Updated

src/storage/src/source/sql_server/replication.rs
ParkMyCar: Done!

src/sql-server-util/src/cdc.rs
ParkMyCar: Yeah the debug format was just easy, updated to your suggestion!

src/sql-server-util/src/cdc.rs
ParkMyCar: wow way better, thank you!!

src/sql-server-util/src/cdc.rs
ParkMyCar: Updated!

src/sql-server-util/src/cdc.rs
ParkMyCar: It does, this should only really matter when CDC is first initialized for the entire DB, but is something I ran into in tests. I made this configurable via a dyncfg and updated the max wait to 30 seconds

src/sql-server-util/src/cdc.rs
ParkMyCar: Not much, but it might be interesting to observe how the LSN of the upstream advances just for our own understanding

src/sql-server-util/src/cdc.rs
ParkMyCar: Good to know!

SQL Server shouldn't be sending data out of order but the type this Stream yields is:
```
Data {
    /// The capture instance these changes are for.
    capture_instance: Arc<str>,
    /// The LSN that this change occurred at.
    lsn: Lsn,
    /// The change itself.
    changes: Vec<Operation>,
},
```
So we're grouping by LSN just so we can emit an entire set of `Operation`s all at once instead of `(Arc<str>, Lsn, Operation)` tuples. Happy to refactor this in a follow up if you'd prefer!

src/sql-server-util/src/cdc.rs
ParkMyCar: Good thought, refactored to emit this once at the start!

src/sql-server-util/src/cdc.rs
petrosagg: Emitting a sequence of `(Arc<str>, Lsn, Operation)` is definitely more timely-like so I highly suggest that we do that, and totally fine as a followup. Separately, it would be nice if we could get rid of that `Arc<str>` in favor of the table id which I presume is some small copy-able type that doesn't need refcounts update all the time.

petrosagg (approved):
Thanks for the quick turnaround on those comments!

ParkMyCar:
> Thanks for the quick turnaround on those comments!

Thank you for the quick review!

PR #32135, opened 2025-04-09
Deny anyhow

antiguru:
Deny access to anyhow macros and some types

Deny access to the anyhow macros `anyhow` and `bail`, and the `Result` type. This covers most of our use of anyhow where we're not just passing errors around.

We could also deny `anyhow::Error`, but that would need us to silence another 1000+ warnings, which I feel has diminishing returns: we're mostly interested to finding places where we construct errors, which are covered by this change. Unfortunately, anyhow allows silent error conversion, meaning any `?` operator can convert from an error to `anyhow`, so we should eventually deny `anyhow::Error`, too.

In support of MaterializeInc/database-issues#9092

### Checklist


clippy.toml
```diff
@@ -88,9 +88,12 @@ disallowed-macros = [
     { path = "proptest::prop_oneof", reason = "use `proptest::strategy::Union::new` instead" },
     { path = "log::log", reason = "use the macros provided by `tracing` instead (database-issues#3001)" },
     { path = "tracing::instrument", reason = "use `mz_ore::instrument` instead" },
+    { path = "anyhow::anyhow", reason = "use structured errors instead" },
+    { path = "anyhow::bail", reason = "use structured errors instead" },
```
pH14, line 92: imo this would be more valuable with a fleshed out example to point people towards

misc/bazel/cargo-gazelle/src/bin/main.rs
```diff
@@ -7,6 +7,9 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+// database-issues#9092: anyhow should not be used.
```
pH14, line 10: the breadcrumb trail here should also lead folks to the right solution

pH14 (commented):
The risks outlined in https://github.com/MaterializeInc/database-issues/issues/9092 seem reasonable, but we should also make the path forward very obvious to anyone touching the code because it might be surprising that `anyhow` is to be avoided.

Following the deprecation warnings, we should be able to answer:
* If someone is writing net-new error handling code, what does it look like in absence of `anyhow`? Are there any references to model after? 
* If someone needs to update the existing code, what should they do?

petrosagg (commented):
I'm a bit concerned with merging this PR. The most likely outcome for the years to come is that all these ignore directives will stay put and we will not refactor all the errors to structured errors. And, given the pervasiveness of these warnings people might add new ones because... they are already everywhere.

I think a more targeted approach would be more suitable here. From the downsides of `anyhow` the most important one (to me) is the fact that we can end up writing them down durably. The types that we write down are known and I would be much happier if we only audited those and removed anyhow from those.

For the other problems, are they supported by enough data to cause repo wide action? The stack overflow is reported to happen in a persist test. Backtraces are expensive, but how often do we collect them? I would again prefer a more targeted approach where we find hot spots from benchmarks.

More generally, `anyhow` has its uses and not everything warrants a structured error type, so I wouldn't globally deny it. So I propose we don't merge this and instead we try to find the places where the anyhow downsides matter and fix those.

antiguru:
Not pushing this change further, we need a more structured approach here. Thank y'all for your input!

PR #32141, opened 2025-04-09
jemalloc: use GCC intrinsics for unwinding in profiling

teskje:
This is a workaround for deadlocks caused by nested unwinding. Alternative to https://github.com/MaterializeInc/materialize/pull/32139.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9159

### Checklist

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -144,6 +144,10 @@ build --@rules_rust//:extra_rustc_flag="-Ccodegen-units=64"
 # just `.rmeta` instead of the full `.rlib`.
 build --@rules_rust//rust/settings:pipelined_compilation=True
 
+# Enforce frame pointers everywhere. We rely on them for collecting backtraces.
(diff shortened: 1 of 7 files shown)


teskje:
I assume all the segfaults in CI come from us not enforcing frame pointers. We could enforce them with [force-frame-pointers](https://doc.rust-lang.org/rustc/codegen-options/index.html#force-frame-pointers), but we'd also have to do so for all libraries we depend on and apparently we risk segfaults if we miss one. The `libunwind` approach feels safer, if we can get it to work.

def- (approved):
Please ignore the upgrade test failures in Nightly, they are fixed by https://github.com/MaterializeInc/materialize/pull/32265

The only thing I'm afraid of here is that we might miss the argument in some C library and then get a crash in production as a consequence of that

PR #32147, opened 2025-04-09
Prefer slices to offsets in `Row` parsing

frankmcsherry:
The way we parse rows as `&[u8]` to form `Datum` elements uses a bunch of `offset: &mut usize` to think about where we are in the parsing. It seems like it can be replaced with slices, using `bytes: &mut &'a [u8]` as the primary argument and advancing it as we read, rather than advancing the `offset`. This simplifies a fair bit of code, and potentially removes any number of bounds checks where the slice carries the information about its validity. E.g. several `data[*offset .. *offset + len]` things turn in to `data.split_at(len)`.

### Checklist

### DIFF
--- a/src/repr/src/row.rs
+++ b/src/repr/src/row.rs
@@ -585,10 +585,7 @@ impl RowRef {
 
     /// Iterate the [`Datum`] elements of the [`RowRef`].
     pub fn iter(&self) -> DatumListIter {
-        DatumListIter {
-            data: &self.0,
-            offset: 0,
-        }
+        DatumListIter { data: &self.0 }
     }
 
     /// Return the byte length of this [`RowRef`].
@@ -621,10 +618,7 @@ impl<'a> IntoIterator for &'a RowRef {
     type IntoIter = DatumListIter<'a>;
 
     fn into_iter(self) -> DatumListIter<'a> {
-        DatumListIter {
-            data: &self.0,
-            offset: 0,
-        }
+        DatumListIter { data: &self.0 }
     }
 }
 
@@ -671,13 +665,11 @@ pub struct RowPacker<'a> {
 #[derive(Debug, Clone)]
 pub struct DatumListIter<'a> {
     data: &'a [u8],
-    offset: usize,
 }
 
 #[derive(Debug, Clone)]
 pub struct DatumDictIter<'a> {
     data: &'a [u8],
-    offset: usize,
     prev_key: Option<&'a str>,
 }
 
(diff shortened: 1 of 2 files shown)


src/repr/src/row.rs
```diff
@@ -2545,11 +2524,12 @@ impl RowPacker<'_> {
 
     /// Truncates the underlying row to contain at most the first `n` datums.
     pub fn truncate_datums(&mut self, n: usize) {
+        let prev_len = self.row.data.len();
         let mut iter = self.row.iter();
         for _ in iter.by_ref().take(n) {}
-        let offset = iter.offset;
+        let next_len = self.row.data.len();
```
antiguru, line 2530: I think this is what we need:
```suggestion
        let next_len = iter.data.len();
```

src/repr/src/row.rs
```diff
@@ -2484,8 +2462,9 @@ impl RowPacker<'_> {
         // Validate that what was written maintains the correct invariants.
         let mut actual_datums = 0;
         let mut seen = None;
-        while datum_check < self.row.data.len() {
-            let d = unsafe { read_datum(&self.row.data, &mut datum_check) };
+        let mut dataz = &self.row.data[datum_check..];
+        while !dataz.is_empty() {
+            let d = unsafe { read_datum(&mut dataz) };
```
antiguru, line 2467: You could convert this to a for-loop over a RowRef on `&self.row.data[datum_check..]`. It needs to yield exactly `expected_datums` elements. Also, remove the `datum_check`, which is now an incorrect assertion (it used to be incremented, but isn't anymore).

src/repr/src/row.rs
```diff
@@ -1004,17 +997,15 @@ fn read_byte(data: &[u8], offset: &mut usize) -> u8 {
...
 ) -> [u8; N] {
     let mut raw = [FILL; N];
-    for i in 0..length {
-        debug_assert!(i < raw.len());
-        debug_assert!(*offset + i < data.len());
-        *raw.get_unchecked_mut(i) = *data.get_unchecked(*offset + i);
+    let (prev, next) = data.split_at(length);
+    for (i, prev_byte) in prev.iter().enumerate() {
+        raw[i] = *prev_byte;
     }
```
antiguru, line 1007:
```suggestion
    let (prev, next) = data.split_at(length);
    (&mut raw[..prev.len()]).copy_from_slice(prev);
```

src/repr/src/row.rs
```diff
@@ -1004,17 +997,15 @@ fn read_byte(data: &[u8], offset: &mut usize) -> u8 {
 ///   * length <= N
 ///   * offset + length <= data.len()
 unsafe fn read_byte_array_sign_extending<const N: usize, const FILL: u8>(
```
antiguru, line 999:
```suggestion
fn read_byte_array_sign_extending<const N: usize, const FILL: u8>(
```

src/repr/src/row.rs
```diff
@@ -1025,11 +1016,10 @@ unsafe fn read_byte_array_sign_extending<const N: usize, const FILL: u8>(
 ///   * length <= N
 ///   * offset + length <= data.len()
 unsafe fn read_byte_array_extending_negative<const N: usize>(
```
antiguru, line 1018:
```suggestion
fn read_byte_array_extending_negative<const N: usize>(
```

antiguru (commented):
Left some comments inline, I think there's a bug in pushing ranges. Otherwise this seems like a good change to make; it's simpler to read and more idiomatic, too.

src/repr/src/row.rs
frankmcsherry: I deleted the check, which by its own admission would be testing undefined behavior.

antiguru (approved):
Looks great; thank you!

frankmcsherry:
Thanks for the helpful review!

PR #32163, opened 2025-04-11
Disable backtrace collection in `anyhow`

teskje:
The `anyhow` crate automatically collects backtraces when creating `anyhow::Error`s if the `RUST_BACKTRACE=1` env variable is set, and uses them to enhance the errors' Debug representation. This is a feature we don't actively rely on but that comes with significant costs:

 * Collecting backtraces significantly increases the memory and CPU costs of creating error values, which can be both expensive and surprising.
 * In rare cases, collecting backtraces can deadlock the process, if jemalloc decides to collect a profiling sample at the same time.

This PR therefore disables backtrace collection in `anyhow` by setting `RUST_LIB_BACKTRACE=0`.

This also disables backtrace collection in all other libraries that use `std::backtrace::Backtrace::capture`, but I'm not aware of any we use. Sentry does collect backtraces, but it uses the `backtrace` crate, which isn't affected by `RUST_LIB_BACKTRACE`.

### Motivation

Partially mitigates https://github.com/MaterializeInc/database-issues/issues/9159
Partially addresses the concerns raised in https://github.com/MaterializeInc/database-issues/issues/9092

### Tips for reviewer

@antiguru raised a concern that we might have persisted backtraces as part of `anyhow::Error` debug output in dataflow errors emitted to sources or MVs, in which case disabling them would change the serialized representation of the errors, making it impossible to retract them again. The concern is valid, but it's also the case that we can't expect backtraces to be stable across releases, so these error representations would already have been unstable. Switching off backtraces is no worse than having a new release that changes them.

It would be nice to be able to switch off backtraces in `anyhow` directly, instead of having to deal with the higher blast radius of `RUST_LIB_BACKTRACE`. There is no such way currently, but plenty discussion in the crate's Github issues, so we might get one eventually.

### Checklist

### DIFF
--- a/misc/images/ubuntu-base/Dockerfile
+++ b/misc/images/ubuntu-base/Dockerfile
@@ -11,6 +11,10 @@ FROM ubuntu:noble-20250404
 
 # Ensure any Rust binaries that crash print a backtrace.
 ENV RUST_BACKTRACE=1
+# Disable backtrace collection in Rust libraries, such as `anyhow`. Collecting
+# backtraces at runtime can be expensive and is known to cause deadlocks
+# (database-issues#9159).
+ENV RUST_LIB_BACKTRACE=0
 
 RUN sed -i -e 's#http://archive\.ubuntu\.com#http://us-east-1.ec2.archive.ubuntu.com#' \
            -e 's#http://security\.ubuntu\.com#http://us-east-1.ec2.archive.ubuntu.com#' \
--- a/src/balancerd/Cargo.toml
+++ b/src/balancerd/Cargo.toml
@@ -10,7 +10,7 @@ publish = false
 workspace = true
 
 [dependencies]
-anyhow = { version = "1.0.97", features = ["backtrace"] }
+anyhow = "1.0.97"
 async-trait = "0.1.88"
 axum = "0.7.5"
 bytes = "1.10.1"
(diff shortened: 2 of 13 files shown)


src/balancerd/Cargo.toml
```diff
@@ -10,7 +10,7 @@ publish = false
 workspace = true
 
 [dependencies]
-anyhow = { version = "1.0.95", features = ["backtrace"] }
+anyhow = "1.0.95"
```
teskje, line 13: Not really related to the main change of this PR, just some cleanup. The `backtrace` feature doesn't do anything for Rust versions >= 1.65, just pulls in an unnecessary dependency.

misc/images/ubuntu-base/Dockerfile
```diff
@@ -11,6 +11,10 @@ FROM ubuntu:noble-20250127
 
 # Ensure any Rust binaries that crash print a backtrace.
 ENV RUST_BACKTRACE=1
+# Disable backtrace collection in Rust libraries, such as `anyhow`. Collecting
+# backtraces at runtime can be expensive and is known to cause deadlocks
+# (database-issues#9159).
+ENV RUST_LIB_BACKTRACE=0
```
antiguru, line 17: Should we set it for things like `bin/environmentd`, too? Might be good for consistency, although not strictly required.

antiguru (commented):
:tada: Thanks for taking care of this!

misc/images/ubuntu-base/Dockerfile
teskje: I think we don't need to! `bin/environmentd` doesn't set `RUST_BACKTRACE=1`, so we shouldn't be collecting backtraces in `anyhow` in the first place.

antiguru (approved):
Thank you!

def- (approved):

teskje:
TFTRs!

PR #32173, opened 2025-04-11
[persist] Be defensive when we have structured only data

ParkMyCar:
When creating a `PartMigration` if we don't have a `schema_id` but we do have structured only data, we'll fallback to the `deprecated_schema_id` field because this indicates the `persist_record_schema_id` and `persist_batch_columnar_format` flags did not propagate properly.

### Motivation

Fix an issue at least one customer has run into.

### Checklist


bkirwi (approved):
I think this would be safe to merge, though I'd be just as happy _only_ backporting it... IIUC this codepath should never be hit in cloud, and I'm not sure how to think about whether it would be correct or not if it was.

(In retrospect it would have been wise to always set the schema id when the new format was set, though since that flag no longer exists, not much to be done about that here...)

PR #32177, opened 2025-04-11
allow orchestratord to force a rollout to promote before it's ready

doy-materialize:
### Motivation

this allows us to choose how to resolve a rollout which is stuck or taking too long - we can either cancel the rollout (by resetting the reconciliation_id to the last reconciliation id from the status) or now force it to complete by setting force_promote to the reconciliation_id value.

### Tips for reviewer

not planning on porting this to the environment controller since it we're so close to getting rid of it

### Checklist

### DIFF
--- a/src/cloud-resources/src/crd/materialize.rs
+++ b/src/cloud-resources/src/crd/materialize.rs
@@ -108,6 +108,12 @@ pub mod v1alpha1 {
         // generation rollout is automatically triggered.
         #[serde(default = "Uuid::new_v4")]
         pub request_rollout: Uuid,
+        // If force_promote is set to the same value as request_rollout, the
+        // current rollout will skip waiting for clusters in the new
+        // generation to rehydrate before promoting the new environmentd to
+        // leader.
+        #[serde(default)]
+        pub force_promote: Uuid,
         // This value will be written to an annotation in the generated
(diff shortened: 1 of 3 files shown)


src/cloud-resources/src/crd/materialize.rs
```diff
@@ -108,6 +108,12 @@ pub mod v1alpha1 {
         // generation rollout is automatically triggered.
         #[serde(default = "Uuid::new_v4")]
         pub request_rollout: Uuid,
+        // If force_promote is set to the same value as request_rollout, the
+        // current rollout will skip waiting for clusters in the new
+        // generation to rehydrate before promoting the new environmentd to
+        // leader.
+        #[serde(default)]
+        pub force_promote: Uuid,
```
alex-hunt-materialize, line 116: Does this need to be an `Option<Uuid>` to handle already existing CRs? What does K8S do in cases where a new CRD definition is incompatible with the old stored data?

The `#[serde(default)]` handles missing data on the controller side, but does K8S even allow you to upgrade the CRD in this way?

doy-materialize: as far as i can tell from testing, this works as you would want it to

alex-hunt-materialize (approved):

PR #32199, opened 2025-04-14
Allow delta join rendering from collections

frankmcsherry:
Our delta join rendering starts each path from an arrangement, and does not permit a collection stream to start .. the stream. This is because delta join planning currently requires arrangements for each input, and so there should always be an arrangement that exists, and from which it is cheapest to produce a stream of updates.

However, there are reasons to want to support delta join rendering starting from streams, specifically in service of ad-hoc queries with joins, which potentially look like `facts JOIN dim1 JOIN dim2 ..`, where we have arrangements on the dimension tables but not the fact table. When this is a bare `SELECT`, rather than a maintained dataflow, only one delta join path is active (the others can be pruned) and that one path doesn't need an arrangement for its first input (in fact it could be so large as to be counter productive).

### Checklist

### DIFF
--- a/src/compute/src/render/join/delta_join.rs
+++ b/src/compute/src/render/join/delta_join.rs
@@ -13,8 +13,6 @@
 
 #![allow(clippy::op_ref)]
 
-use std::collections::{BTreeMap, BTreeSet};
-
 use columnar::Columnar;
 use differential_dataflow::IntoOwned;
 use differential_dataflow::consolidation::ConsolidatingContainerBuilder;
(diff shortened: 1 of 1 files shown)


antiguru:
> The volume of arrangements imported into regions is higher than previously, where we went through some pains to only bring in arrangements that would be used in the region (to minimize the noise in visualization). This wouldn't be hard to add back in, but it seems to be annoying to have as the code evolves.

I think we could add a type that gives the right to enter an object into a region, but is lazy and only enters when first used. This way we could absorb the complexity to figure out which objects to import into a runtime type with interior mutability.

PR #32201, opened 2025-04-14
[sql_server] Enable platform checks

ParkMyCar:
This PR enables SQL Server in platform checks by adding a new `--features` argument that supports "azurite" and "sql_server". Then we add new workflows in CI that run SQL Server on x86 which is the only arch that is supported by Microsoft's official SQL Server Docker image.

### Motivation

Test the SQL Server source with platform-checks

### Checklist

### DIFF
--- a/ci/nightly/pipeline.template.yml
+++ b/ci/nightly/pipeline.template.yml
@@ -711,7 +711,7 @@ steps:
         plugins:
           - ./ci/plugins/mzcompose:
               composition: platform-checks
-              args: [--scenario=NoRestartNoUpgrade, "--seed=$BUILDKITE_JOB_ID", --azurite]
+              args: [--scenario=NoRestartNoUpgrade, "--seed=$BUILDKITE_JOB_ID", --features=azurite]
 
       - id: checks-restart-entire-mz
         label: "Checks + restart of the entire Mz"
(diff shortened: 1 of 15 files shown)


ci/nightly/pipeline.template.yml
```diff
@@ -769,6 +769,25 @@ steps:
               composition: platform-checks
               args: [--scenario=DropCreateDefaultReplica, --execution-mode=parallel, "--seed=$BUILDKITE_JOB_ID"]
 
+      - id: checks-parallel-drop-create-default-replica-x86_64
```
def-, line 772: It's a bit annoying that this only works on x86-64 and we mostly run our tests on aarch64. Should we maybe move half of the platform-checks over to x86-64? Then we also have some coverage for upgrades etc.

def- (approved):

ci/nightly/pipeline.template.yml
ParkMyCar: Yeah I was curious what your opinion was here. I'm happy to move half of them over, also happy to duplicate them entirely on x86? Not sure how many of self-hosted customers run x86 but it could be nice to have some move coverage there if so

def-: So far we have barely found any differences between x86-64 and aarch64, so I think duplicating is probably not worth it.

PR #32205, opened 2025-04-15
optimizer: Fix JoinImplementation's MFP lifting's inlining

ggevay:
When `JoinImplementation` lifts an MFP from a join input to after the join (to be able to reuse an arrangement), it needs to permute the join equivalences by the projection of the lifted MFP. Moreover, it can happen that a column reference that we get from the permutation refers to a column that is built by a map expression of the MFP that is being lifted. This requires further treatment, because the lifted MFP will happen _after_ the equivalences, so the equivalences can't refer to a column that is built only in the MFP. What the code does (already before this PR) is simply inline the referred Map expression in place of the problematic column reference.

But, there was a bug in the inlining logic: After substituting a problematic column reference with a new expression, we have to look at the new expression again, because it might itself be a problematic column reference. (Note that the new expression's children were already being visited, it's just that the new expression itself was not being visited again. So, the problematic case was when the inlined expression itself was again a problematic column reference.)

Example: On the reduced query in https://github.com/MaterializeInc/database-issues/issues/9156, we had the column reference `#5` in a Join equivalence before MFP lifting. The lifted MFP was
```
MapFilterProject(
  expressions:
    #7 <- case when (#4 = #4) then null else #4 end,
    #8 <- #7,
  predicates:
    <before: 2> (#1 = (#0 + #1)),
    <before: 2> (#0 = case when (#1 = #1) then null else #1 end),
    <before: 7> (#4 = ((#5 / bigint_to_double(case when (#6 = 0) then null else #6 end)) + #4)),
    <before: 7> ((#5 / bigint_to_double(case when (#6 = 0) then null else #6 end)) = case when (#4 = #4) then null else #4 end),
  projection: [0, 1, 2, 3, 4, 8]
  input_arity: 7
)
```
so the `#5` was permuted to `#8`. The inlining logic then changed the `#8` to `#7`, but this was not enough, because `#7` also happens to be built by the MFP. After the PR, we go around the added loop a second time, so that `#7` is further substituted with the `case ...` expression.

Nightly (subset): https://buildkite.com/materialize/nightly/builds/11839

(I'll do some follow-up PRs shortly: 1. use a modern visitation instead of the deprecated `visit_mut_pre_post`, 2. Log the backtrace when catching an optimizer panic, 3. catch optimizer panics that happen in `as_explain_single_plan`'s `normalize_lets` call.)

### Motivation

  * This PR fixes a recognized bug: https://github.com/MaterializeInc/database-issues/issues/9156

### Checklist

### DIFF
--- a/src/transform/src/join_implementation.rs
+++ b/src/transform/src/join_implementation.rs
@@ -925,6 +925,20 @@ fn implement_arrangements<'a>(
     )
 }
 
+/// This function continues the surgery that `implement_arrangements` started.
+///
+/// (In theory, this function could be merged into `implement_arrangements`, but it would be a bit
+/// painful, because we need to access the join's `implementation` after `implement_arrangements`,
+/// which would be somewhat convoluted after what `install_lifted_mfp` does.)
+///
+/// The given MFP should be the merged MFP from all the lifted inputs MFPs.
+///
+/// `install_lifted_mfp` mutates the given join expression as follows:
+/// - Puts the given MFP on top of the Join.
+/// - Attends to `equivalences`, which was invalidated by `implement_arrangements` if it refers to a
+///   column that was permuted or created by the given MFP.
+/// - Canonicalizes scalar expressions in maps and filters with respect to the join equivalences.
+///   See inline comment for more details.
 fn install_lifted_mfp(
     new_join: &mut MirRelationExpr,
     mfp: MapFilterProject,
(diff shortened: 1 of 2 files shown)


ParkMyCar (approved):

src/transform/src/join_implementation.rs
```diff
@@ -941,9 +955,16 @@ fn install_lifted_mfp(
                     #[allow(deprecated)]
                     expr.visit_mut_pre_post(
                         &mut |e| {
-                            if let MirScalarExpr::Column(c) = e {
+                            // This has to be a loop! This is because it can happen that the new
+                            // expression is again a column reference, in which case the visitor
+                            // wouldn't be called on it again. (The visitation continues with the
+                            // children of the new expression, but won't visit the new expression
+                            // itself again.)
+                            while let MirScalarExpr::Column(c) = e {
```
antiguru, line 963: Could we instead canonicalize the map before we visit `expr`? Not a huge concern, but it would avoid the loop here.

ggevay:
Thanks for the reviews! (We discussed this in the SQL team sync, and the long-term plan is to rewrite the entire `join_implementation.rs` at some point, because it's hard to maintain.)

PR #32209, opened 2025-04-15
Catch more panics when calling optimizer code

ggevay:
EXPLAIN (and other things) call into the optimizer at many points, and not all of these have been wrapped into `catch_unwind` so far. This PR attends to some of these. More specifically:
- `normalize_lets` is called at some places outside the normal optimizer pipelines, e.g., it's called by EXPLAIN. Therefore, the PR adds a `catch_unwind` in `normalize_lets`, so that we catch panics even when `normalize_lets` is called not through the normal optimizer pipelines. (I noticed this problem when working on https://github.com/MaterializeInc/database-issues/issues/9156: normally, the panic was caught, but sometimes it caused an actual panic in EXPLAIN commands.)
- The PR adds `catch_unwind` to the `normalize_subqueries` call. (This is also in EXPLAIN.)

I suggest reviewing commit-by-commit:
1. does a refactoring to make the 2. commit easy. I had to add the `MaybeShouldPanic` trait, because of needing to generalize over `TransformError` and `OptimizerError`.
2. attends to `normalize_lets`.
3. attends to the `normalize_subqueries` call.

### Motivation

  * We don't want to worry about EXPLAIN commands causing envd panics when impersonated into user envs.

### Checklist

### DIFF
--- a/src/sql/src/plan/explain.rs
+++ b/src/sql/src/plan/explain.rs
@@ -9,6 +9,8 @@
 
 //! `EXPLAIN` support for structures defined in this crate.
 
+use std::panic::AssertUnwindSafe;
+
 use mz_expr::explain::{ExplainContext, ExplainSinglePlan};
 use mz_expr::visit::{Visit, VisitChildren};
 use mz_expr::{Id, LocalId};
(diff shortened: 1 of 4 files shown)


src/sql/src/plan/explain.rs
```diff
@@ -53,9 +55,16 @@ impl<'a> HirRelationExpr {
         context: &'a ExplainContext<'a>,
     ) -> Result<ExplainSinglePlan<'a, HirRelationExpr>, ExplainError> {
         // unless raw plans are explicitly requested
-        // ensure that all nested subqueries are wrapped in Let blocks
+        // ensure that all nested subqueries are wrapped in Let blocks by calling
+        // `normalize_subqueries`
         if !context.config.raw_plans {
-            normalize_subqueries(self)?;
+            mz_ore::panic::catch_unwind_str(AssertUnwindSafe(|| {
```
ParkMyCar, line 61: should this use the new `catch_unwind_optimize(...)`?

ParkMyCar (approved):
Overall LGTM!

src/sql/src/plan/explain.rs
ggevay: It can't use that, because we are in `mz-sql`, and the new `catch_unwind_optimize` is in `mz-transform`, so the dependency is in the wrong way.

Also note that this doesn't need the `should_panic()` check (which is one of the things that `catch_unwind_optimize` does), because that kind of error is also an `mz-transform` thing (it can happen during `FoldConstants`, when it sees `mz_unsafe.mz_panic()`).

Btw. I'm very soon going to further refactor optimizer panic handling as [discussed in incident-512](https://materializeinc.slack.com/archives/C08NDPVT8CT/p1744752708230149?thread_ts=1744738527.172389&cid=C08NDPVT8CT), and then I will probably factor out the part that actually handles a caught panic (which is separate from the `should_panic()` check), and put that somewhere that's accessible in both crates.

PR #32211, opened 2025-04-15
design: large SELECT result size

aljoscha:
Rendered: https://github.com/aljoscha/materialize/blob/adapter-lift-max-result-size-limitation/doc/developer/design/20250415_large_select_result_size.md

Design doc for https://github.com/MaterializeInc/database-issues/issues/9180

### DIFF
--- a/doc/developer/design/20250415_large_select_result_size.md
+++ b/doc/developer/design/20250415_large_select_result_size.md
@@ -1,0 +1,200 @@
+# Large Select Result Size
+
+## Context
+
+Currently, results of SELECT queries are sent back from the cluster to
+`environmentd` via the compute protocol and these results are fully
+materialized (stored in memory) in `environmentd` before sending them out to
+the client.
+
+This has several implications:
+
+- Sending large results "clogs up" the cluster/controller communication
+- The amount of memory we want to give to `environmentd` limits the size of
+  results we can return
+
+In practice the above make it so we limit the size of results using a
+`max_result_size` parameter, and some larger customers are chafing against
+that.
+
+Today, we always materialize the results of a query in `environmentd` memory
+before sending them on to the client. This is only required for some types of
+queries, though: only queries that require _post processing_ need
+materialization while queries that don't need post processing could be streamed
+through to the client. Queries that need an ordering established on the whole
+result require post processing, which today are only those queries that have an
+ORDER BY. This is handled by `RowSetFinishing` in the code.
+
+Below, we use _streamable queries_ for queries that don't require post
+processing, and _non-streamable queries_ for those that do require it.
+
+We think it is easier to lift the result size limitation for streamable
+queries, so want to approach that first, but there are ways we can lift the
+limitation for non-streamable queries as well, for example by moving the post
+processing to the cluster side or by implementing an approach where we can sort
+and merge large results within bounded memory on the `environmentd` side.
+
+## Technical Context
+
+There are two ways of getting query results and sending them from the cluster
+back to `environmentd`:
+
+1. Extracting from an arrangement (the data structure backing indexes)
+2. Extracting directly from persist
+
+This leads to there being at least three ways a SELECT can be executed from the
+`environmentd` side:
+
+1. Fast-path SELECT: there is an existing arrangement (index) that we can
+   extract our result from.
+2. Slow-path SELECT: in order to get our result we first need to render a
+   temporary dataflow with an index that will contain the results. Then we can
+   use the same logic as a fast-path SELECT to extract the result. Afterwards
+   the temporary dataflow is torn down again.
+3. Fast-path persist SELECT: The query is simple enough that we can directly
+   extract it from a persist shard. With potentially some amount of
+   map-filter-project applied. This happens on the `clusterd` side and is
+   shipped back using the same mechanism that is used for shipping back other
+   SELECT results.
+
+1 and 2 here imply that the result must fit into memory on the cluster side,
+unless we want to, say, change how results can be extracted on the cluster
+side.
+
+## Goals
+
+The user facing goal is:
+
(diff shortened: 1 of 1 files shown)


doc/developer/design/20250415_lift_max_result_size_limitation.md
```diff
@@ -0,0 +1,113 @@
...
+that.
+
+The overall goal of this design is to lift this limitation, but there is
+nuance, especially when it comes to different query modifiers. For example,
+queries with a ORDER BY and/or LIMIT clause require results from the cluster to
+be combined somewhere, to impose the required order. Currently, this is
+happening in `environmentd`, but could happen on the cluster side. We think it
+is easier to lift result-size limitations for queries that don't require
+post-processing. We should be able to just "stream them through", without
+materializing in `environmentd`.
```
bkirwi, line 27: FWIW, I think it wouldn't be super hard to support these cases using the `Consolidator` machinery we have now. Happy to chat about this if you're interested!

doc/developer/design/20250415_lift_max_result_size_limitation.md
```diff
@@ -0,0 +1,113 @@
...
+More details need to be filled in, but the rough idea is that we a) need to
+stop sending large amounts of data through the compute protocol, and b) need to
+stop materializing large results in `environmentd` memory.
+
+On the cluster side, we "stash" query results in the blob store that is used by
+persist, send a handle to that data back via the compute protocol, and then (in
+`environmentd`) stream those results from persist back to the client.
+
+We would use a (configurable) threshold for determining whether to send a
+result inline in the compute protocol or out of band via the blob store.
```
bkirwi, line 77: Sounds like this could use the same machinery as @ParkMyCar worked up for large insert / update queries, but streaming the resulting batch into a client instead of linking it into a shard?

doc/developer/design/20250415_lift_max_result_size_limitation.md
ParkMyCar: Nice! I was thinking this might be possible, glad to know you think the same!

doc/developer/design/20250415_large_select_result_size.md
```diff
@@ -0,0 +1,174 @@
...
+Today, we always materialize the results of a query in `environmentd` memory
+before sending them on to the client. This is only required for some types of
+queries, though: only queries that require _post processing_ need
+materialization while queries that don't need post processing could be streamed
+through to the client. Queries that need an ordering established on the whole
+result require post processing, which are queries that have at least one of
+ORDER BY, LIMIT, or OFFSET. This is handled by `RowSetFinishing` in the code.
+
+Below, we will use _streamable queries_ for queries that don't require post
+processing, and _non-streamable queries_ for those that do require it.
```
antiguru, line 29: We always imply implicit post processing, i.e., we sort the data even if the user didn't specify any specific order. How would this impact your design?

Specifically, we might write down as many blobs as we have workers on the cluster replica. Each is sorted within itself. When sending the data to the client, we need to stream all blobs and merge them on-the-fly. Certainly possible, but at what resource utilization?

doc/developer/design/20250415_large_select_result_size.md
```diff
@@ -0,0 +1,174 @@
...
+- Lift result-size limitation for non-streamable queries
+- Change result-size limitations for SUBSCRIBE
+- Change how we ship around and append data to shards, when handling INSERT
+- Change how we ship around and append data for sources
+
+## Implementation
+
+Purposefully high level, but the rough idea is that we a) need to stop sending
+large amounts of data through the compute protocol, and b) need to stop
+materializing large results in `environmentd` memory.
```
antiguru, line 94: How much of a problem is it to send large amounts of data through the compute protocol? an alternative design could be implemented entirely within envd. It could receive the data from replicas, and instead of materializing it in memory, it could write to files or blob storage, and stream the results to the client from there. Just pointing out the option!

doc/developer/design/20250415_large_select_result_size.md
```diff
@@ -0,0 +1,174 @@
...
+persist shard. Though the latter is desirable in future work of decoupling
+creating and appending batches.
+
+For our user-facing goal specifically: on the cluster side, we "stash" query
+results as persist batches, send a handle to that data back via the compute
+protocol, and then (in `environmentd`) stream those results from the persist
+blob/batch store back to the client.
+
+We would use a (configurable) threshold for determining whether to send a
+result inline in the compute protocol or out-of-band via the blob store.
```
antiguru, line 108: Related: moving the task of writing data to envd would make it simple to determine whether a result is small or large.

antiguru (commented):
This reads fine, and I think would be a small incremental step towards efficient handling of large result sets. My main concern is around how we can match the current infrastructure with what might be needed for this change to work. At the moment, each cluster replica worker has a portion of the output data, and we send all partitions to envd to merge it. This step would still need to be performed to ensure consistent results. How well can persist read from many blobs and merge the outputs into one? As long as we're reading from arrangements, at least we know that the output is physically monotonic.

doc/developer/design/20250415_large_select_result_size.md
aljoscha: I would say it is a problem, and not just for peeks but also for subscribes, which you found in https://github.com/MaterializeInc/database-issues/issues/9264.

An alternative would be to add a protocol between `clusterd` and `envd` for sending back these large responses, which doesn't block the "main" protocol. But I'd say the design proposed here is already kind of that, but it uses persist blob storage as the intermediary, and I think that's easier to implement compared to introducing a new protocol.

aljoscha:
> Works for me! Thanks for following up.

tyty! 🙇‍♂️

PR #32219, opened 2025-04-15
docs: create-source-webhook: update syntax

kay-kim:
- Make `IN CLUSTER` optional and replace rr-diagram with copyable syntax form.


ParkMyCar (approved):
thanks Kay!

PR #32221, opened 2025-04-15
[sql_server] Report `ProgressStatisticsUpdate`s and cleanup committed data

ParkMyCar:
This PR does a few things:

1. In the "replication" operator, emits `ProgressStatisticsUpdate::Snapshot` as we're snapshotting tables.
2. Adds a new "progress" operator which does the following:
    a. On an interval configurable with dyncfg, probes the upstream SQL Server instance for the max LSN and emits `ProgressStatisticsUpdate::SteadyState` updates.
    b. Tracks the `resume_uppers` stream and as we learn about newly committed data, we emit `ProgressStatisticsUpdate::SteadyState` updates and call the upstream SQL Server instance to cleanup the change tables. This behavior is gated behind a dyncfg which is enabled by default but added just in case we find issues when deploying to customers.

### Motivation

### Checklist

### DIFF
--- a/src/sql-server-util/src/inspect.rs
+++ b/src/sql-server-util/src/inspect.rs
@@ -112,7 +112,16 @@ pub fn get_changes_asc(
 }
 
 /// Cleans up the change table associated with the specified `capture_instance` by
-/// deleting all entries with a `start_lsn` less than `low_water_mark`.
+/// deleting `max_deletes` entries with a `start_lsn` less than `low_water_mark`.
+///
+/// Note: At the moment cleanup is kind of "best effort".  If this query succeeds
+/// then at most `max_delete` rows were deleted, but the number of actual rows
+/// deleted is not returned as part of the query. The number of rows _should_ be
+/// present in an informational message (i.e. a Notice) that is returned, but
+/// [`tiberius`] doesn't expose these to us.
+///
+/// TODO(sql_server2): Update [`tiberius`] to return informational messages so we
+/// can determine how many rows got deleted.
 ///
 /// See: <https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sys-sp-cdc-cleanup-change-table-transact-sql?view=sql-server-ver16>.
 pub async fn cleanup_change_table(
--- a/src/sql-server-util/src/inspect.rs
+++ b/src/sql-server-util/src/inspect.rs
@@ -11,6 +11,7 @@
 
 use futures::Stream;
 use itertools::Itertools;
+use mz_ore::cast::CastFrom;
 use smallvec::SmallVec;
 use std::collections::BTreeMap;
 use std::fmt;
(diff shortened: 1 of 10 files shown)


src/sql-server-util/src/inspect.rs
```diff
@@ -110,6 +111,99 @@ pub fn get_changes_asc(
...
+SET @mz_cleanup_status_bit = 0;
+EXEC sys.sp_cdc_cleanup_change_table
+    @capture_instance = @P1,
+    @low_water_mark = @P2,
+    @threshold = @P3,
+    @fCleanupFailed = @mz_cleanup_status_bit OUTPUT;
+SELECT @mz_cleanup_status_bit;
+    ";
+
+    let max_deletes = i64::cast_from(max_deletes);
```
petrosagg, line 137: What is this limit for? I saw that SQL Server has a parameter that defaults to 5000 but as far as our source is concerned we always want to cleanup everything otherwise we risk accumulating garbage with a high throughput enough table, right?

If we want to keep this parameter with a low value to control batching, and e.g avoid a single cleanup call taking a long time, then this function should be running the cleanup function iteratively until there are no more records to cleanup.

src/storage/src/source/sql_server/progress.rs
```diff
@@ -0,0 +1,186 @@
...
+                            mz_ore::soft_panic_or_log!(
+                                "upstream SQL Server went backwards in time, current LSN: {known_lsn}, last known {prev_known_lsn}",
+                            );
+                            continue;
+                        }
+
+                        // Update any listeners with our most recently known LSN.
+                        if let Some(prev_commit_lsn) = prev_offset_committed {
+                            let known_lsn_abrv = known_lsn.abbreviate();
+                            let commit_lsn_abrv = prev_commit_lsn.abbreviate();
```
petrosagg, line 138: What kind of precision loss are we suffering here by abbreviating?

src/storage/src/source/sql_server/replication.rs
```diff
@@ -125,14 +135,44 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
+                    emit_stats(&stats_cap[0], records_known, 0);
+                }
+
+                // Begin streaming our snapshots!
                 let mut snapshot_streams = std::pin::pin!(snapshot_streams);
                 while let Some((capture_instance, data)) = snapshot_streams.next().await {
                     let sql_server_row = data.map_err(TransientError::from)?;
+                    records_total = records_total.saturating_add(1);
+
+                    if records_total % update_granularity.get() == 0 {
```
petrosagg, line 173: This will behave unpredictably depending on machine size, network speed, compression, row size, etc. I suggest we report things based on an `Instant::now()` value where we check whether some threshold time has elapsed

src/storage/src/source/sql_server/progress.rs
ParkMyCar: We lose the `record_id` field which denotes an individual operation within a log block, but AFAICT (based on the [docs](https://learn.microsoft.com/en-us/sql/relational-databases/sql-server-transaction-log-architecture-and-management-guide?view=sql-server-ver16#log-blocks) and experimentation) the `(vlf_id, log_block)` is the identifier of a single transaction, while the `record_id` denotes the the individual updates within said transaction

src/storage/src/source/sql_server/replication.rs
ParkMyCar: Totally agree that this could change depending on machine size, network speed, etc. but I'm not totally sure what your proposal is? Are you suggesting every X seconds (based on `Instant::now()` timings) we report how many updates we've ingested thus far?

Also FWIW I went with this implementation because it appears to be what both [Postgres](https://github.com/MaterializeInc/materialize/blob/main/src/storage/src/source/postgres/snapshot.rs#L489-L499) and [MySQL](https://github.com/MaterializeInc/materialize/blob/main/src/storage/src/source/mysql/snapshot.rs#L447-L457) do too. Although I could be missing something here!

petrosagg: Oh interesting. I didn't know we do this in pg and mysql and in a non-configurable way, yikes! Yes I was envisioning something like:

```rust
let mut last_report = Instant::now();
while .... {
   if last_report.elapsed() > SNAPSHOT_PROGRESS_REPORT_INTERVAL.get() {
       last_report = Instant::now();
       // reporting logic
   }
   ...
}
```

src/storage/src/source/sql_server/progress.rs
petrosagg: Ah, cool, correlating with the number of transactions sound great

src/sql-server-util/src/inspect.rs
petrosagg: My understanding is that this limit is not per LSN but per invocation of the cleanup function, right? So if we don't call it often enough we will fall behind. Since we commit to persist once per second the best case scenario is cleaning up 5000 records per second.

Until we make this an iterative process I would increase the default value of the parameter to something much much higher, like 1 million.

src/storage/src/source/sql_server/replication.rs
ParkMyCar: Ahhh I see where you're coming from now. Is your concern that we'll emit events different rates, possibly too fast? Not that the events themselves have different values? At first I was thinking about the later and was confused how a time based approach would help, but the former is definitely better for a constant rate of emitting. Updating now!

src/sql-server-util/src/inspect.rs
ParkMyCar: Updated to 1,000,000!

src/storage-types/src/sources/sql_server.rs
```diff
@@ -41,12 +41,44 @@ pub const SNAPSHOT_MAX_LSN_WAIT: Config<Duration> = Config::new(
...
     Duration::from_millis(500),
     "Interval at which we'll poll the upstream SQL Server instance to discover new changes.",
 );
 
+pub const CDC_CLEANUP_CHANGE_TABLE: Config<bool> = Config::new(
+    "sql_server_cdc_cleanup_change_table",
+    true,
+    "When enabled we'll notify SQL Server that it can cleanup the change tables \
+    as the source makes progress and commits data.",
+);
```
DAlperin, line 61: when would we want to not do this?

ParkMyCar: Should have added this as a comment, but this is a CYA flag just in case something goes wrong, e.g. we're too eagerly cleaning up data

PR #32227, opened 2025-04-16
Persist source: retain row allocation

antiguru:
At the moment, the persist source would only retain a row allocation if it had an MFP. The MFP needs to write into a different allocation, so it makes sense to recycle the allocation here. The non-MFP path didn't retain the allocation. This is suboptimal because subsequently we have to right-size another row allocation to decode into, which can involve $log n$ reallocations, where $n$ is the size of the row. Instead, clone the output and retain the probably right-sized row allocation.

### Checklist

### DIFF
--- a/src/storage-operators/src/persist_source.rs
+++ b/src/storage-operators/src/persist_source.rs
@@ -658,7 +658,9 @@ impl PendingWork {
                     } else {
                         let mut emit_time = *self.capability.time();
                         emit_time.0 = time;
-                        session.give((Ok(row), emit_time, diff.into()));
+                        // Clone row so we retain our row allocation.
+                        session.give((Ok(row.clone()), emit_time, diff.into()));
+                        row_buf.replace(SourceData(Ok(row)));
                         *work += 1;
                     }
                 }


src/storage-operators/src/persist_source.rs
```diff
@@ -658,7 +658,9 @@ impl PendingWork {
                     } else {
                         let mut emit_time = *self.capability.time();
                         emit_time.0 = time;
-                        session.give((Ok(row), emit_time, diff.into()));
+                        // Clone row so we retain our row allocation.
+                        session.give((Ok(row.clone()), emit_time, diff.into()));
+                        row_buf.replace(SourceData(Ok(row)));
```
petrosagg, line 663: Does this cause "latching" behavior in that if a huge row comes along we will permanently sit on that memory instead of recycling it through the allocator?

antiguru: It should only survive until the function returns, so not a concern. Elsewhere we have the shared row, which has the latching behavior you describe. Last I checked it was a few MiBs in production max.

petrosagg: Cool, that's good.

One more thing I noticed just now is that the proposed diff does one unecessary copy of the data. We can avoid it like so:
```rust
row_buf.replace(SourceData(Ok(Row::with_capacity(row.byte_capacity()))));
session.give((Ok(row), emit_time, diff.into()));
```

petrosagg (approved):
Approving with a strong preference of also avoiding the additional copy as described in the comment

src/storage-operators/src/persist_source.rs
antiguru: It rather not do this. This will allocate more capacity than needed if, for example, the first row is large, followed by smaller rows.

petrosagg: Ah, didn't see this comment before the merge. We can just use the length then, which will perform the same allocation as the clone method but skip the copy of the data. Will open a separate PR

PR #32229, opened 2025-04-16
*: don't report caught panics to Sentry

teskje:
Sentry's default panic handler runs before the "enhanced" panic handler we install. So if in the enhanced handler we decide that the panic should be caught instead of crashing the process, we have already reported it to Sentry. This has been causing incident reports and confusion in the past.

This commit fixes the issue by disabling Sentry's default panic integration and instead running the Sentry panic handler manually as part of the enhanced panic handler _after_ it has checked whether the panic has been caught.

### Motivation

Fixes https://github.com/MaterializeInc/incidents-and-escalations/issues/232

### Tips for reviewers

A large chunk of the diff comes from me introducing a new `panic` feature for the `mz-ore` crate, to avoid unconditionally pulling in the `sentry-panic` dependency.

### Checklist


src/adapter/src/optimize.rs
```diff
@@ -132,6 +133,12 @@ where
             }
             Ok(result) => result,
             Err(panic) => {
+                // A panic during optimization is always a bug; log an error so we learn about it.
+                tracing::error!(
+                    "caught a panic during query optimization: {panic}\n{}",
+                    Backtrace::force_capture()
+                );
+
                 let msg = format!("unexpected panic during query optimization: {panic}");
```
teskje, line 142: I didn't attempt to change the wording of this error message in this PR because that's user facing and warrants some SQL council input.

antiguru (approved):
I think this looks good! Just one ask, could you test this on staging by introducing a panic and observing that we report it as expected? We probably don't cause incidents on panic in staging, so not exactly the same, but at least something.

teskje:
I did test it locally, with a self-managed Sentry instance. Is that good enough? I can also deploy this branch to staging, but someone will have to tell me how to best induce an optimizer panic :)

antiguru:
Fine if it convinces you it does the right thing!

> I can also deploy this branch to staging, but someone will have to tell me how to best induce an optimizer panic :)

You might be able to call `mz_panic` where it will be const-folded. Requires unsafe mode ofc.

teskje:
> You might be able to call mz_panic where it will be const-folded.

I think `mz_panic` would take the `e.should_panic().is_some()` branch, where we explicitly panic again. So that would show up in Sentry as a panic (and create an incident in prod), but that's expected.

src/transform/src/lib.rs
```diff
@@ -329,6 +330,12 @@ where
         }
         Ok(result) => result.map_err(|e| e),
         Err(panic) => {
+            // A panic during optimization is always a bug; log an error so we learn about it.
+            tracing::error!(
+                "caught a panic during query optimization: {panic}\n{}",
+                Backtrace::force_capture()
```
ggevay, line 336: (This `Backtrace::force_capture` doesn't capture the backtrace of the panic, but just the backtrace of where we catch, i.e., of `catch_unwind_optimize`, right? This can also be somewhat useful, but it would be much more useful to capture the backtrace of the panic itself somehow. But it's ok if this is out of scope for this PR.)

ggevay (approved):
Thanks, LGTM!

src/transform/src/lib.rs
teskje: Ah, good catch! I think it's actually better to remove the backtrace printing here, because otherwise it could lead to confusion. People could assume that the backtrace was from the place that caused the panic, instead of the place that caught it.

I think to get the actual backtrace we'd have to collect it in the enhanced panic handler, and then smuggle it through to `catch_unwind` somehow. Probably through static variables like the `CATCHING_UNWIND` we already use. I'll leave that as a follow up.

teskje:
TFTRs!

bosconi:
bugbot run

src/ore/Cargo.toml
```diff
@@ -53,7 +53,8 @@ proptest-derive = { version = "0.5.1", optional = true }
 rand = { version = "0.8.5", optional = true }
 smallvec = { version = "1.14.0", optional = true }
 stacker = { version = "0.1.20", optional = true }
-sentry = { version = "0.37.0", optional = true, features = ["debug-images"] }
+sentry = { version = "0.37.0", optional = true, default-features = false, features = ["backtrace", "contexts", "debug-images", "transport"] }
+sentry-panic = { version = "0.37.0", optional = true }
```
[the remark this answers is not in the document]
teskje: Already fixed in https://github.com/MaterializeInc/materialize/pull/32592

src/ore/src/panic.rs
```diff
@@ -189,6 +197,9 @@ pub fn install_enhanced_handler() {
             let _ = stderr.write_all(buf.as_bytes());
         }
 
+        // Report the panic to Sentry.
+        sentry_panic::panic_handler(panic_info);
```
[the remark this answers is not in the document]
teskje: Already fixed in https://github.com/MaterializeInc/materialize/pull/32332

bosconi:
This is the output produced by asking `@claude please review https://github.com/MaterializeInc/materialize/pull/32229. But do not use the /mz-pr-review to do so.` as a comment on this PR in the `GitHub App` channel within Slack.

* To actually have it posted here (after reviewing it), follow up with `@claude please post this as a comment on the PR`
* The comment is posted as the person who invoked Claude, not as Claude, which has pros and cons.


## PR Review

**Changes**: +55/-24 across 20 files

### Summary

This PR fixes a bug where panics caught by `catch_unwind` were still being reported to Sentry, causing false incident reports. The issue was that Sentry's default panic handler runs *before* the enhanced panic handler that decides whether a panic should be caught.

### Approach

The fix disables Sentry's default panic integration and manually calls `sentry_panic::panic_handler()` inside the enhanced panic handler *after* checking if the panic should be caught.

### Code Review

**Core change in `src/ore/src/panic.rs`:**
```rust
// Report the panic to Sentry.
sentry_panic::panic_handler(panic_info);

process::abort();
```

This is the right fix — Sentry reporting now happens only for panics that will actually crash the process.

**Good decisions:**
1. **New `panic` feature for mz-ore** — Avoids unconditionally pulling in `sentry-panic` dependency
2. **Feature-gating `ore_catch_unwind`** — The method in `src/ore/src/future.rs` now requires `#[cfg(feature = "panic")]`
3. **Error logging for caught panics** — Added `tracing::error!` in `src/transform/src/lib.rs` and `src/sql/src/plan/explain.rs` so caught optimization panics are still logged

**Test updates are correct** — Switching from `mz_ore::panic::catch_unwind` to `std::panic::catch_unwind` in tests makes sense since tests don't install the enhanced handler. The `#[allow(clippy::disallowed_methods)]` annotations document this clearly.

### Minor observations

1. **Removed `DebugImagesIntegration`** from `src/ore/src/tracing.rs` — This is fine since `debug-images` feature is still enabled in the sentry dependency via Cargo.toml
2. **TODO comments** about collecting backtraces from panic sites are good follow-up items
3. **Storage-operators sentry dep** changed to `default-features = false, features = []` — Worth verifying this crate doesn't actually need any sentry features

### Verdict

Looks good. The fix is clean, well-documented, and correctly addresses the root cause. The feature flag approach avoids unnecessary dependencies, and the test changes are appropriate.

bosconi:
This is the output produced by asking `@claude please review https://github.com/MaterializeInc/materialize/pull/32229 using the /mz-pr-review skill.` as a comment on this PR in the `GitHub App` channel within Slack.

* To actually have it posted here (after reviewing it), follow up with `@claude please post this as a comment on the PR`
* The comment is posted as the person who invoked Claude, not as Claude, which has pros and cons.


# PR Review: `*: don't report caught panics to Sentry`

## Blocking

None found.

---

## Strong suggestions

### 1. Missing test coverage for the core bug fix

The PR description acknowledges this is fixing a bug (Fixes incidents-and-escalations#232), but there's no test verifying that caught panics are now correctly *not* reported to Sentry. While this is difficult to test directly (would require mocking Sentry), a test that exercises the catch-unwind flow with the enhanced handler installed could at least ensure no panics occur and the behavior is correct.

**Reasoning:** The developer guides emphasize that behavior changes should have tests. The core fix in `src/ore/src/panic.rs:200` manually calling `sentry_panic::panic_handler(panic_info)` after the catch-unwind check is untested.

### 2. Removal of `DebugImagesIntegration` from Sentry config

In `src/ore/src/tracing.rs:536`, the `DebugImagesIntegration` was removed from the Sentry client options:

```rust
-                    integrations: vec![Arc::new(DebugImagesIntegration::new())],
```

**Question:** Was this intentional? The change appears to be a side effect of restructuring dependencies. Debug images help Sentry symbolicate stack traces. If this was intentional (perhaps because it's now included via the `debug-images` feature in `sentry`'s Cargo.toml?), it would be worth a brief comment explaining why.

---

## Nits

### 1. TODO comments added

Two new `TODO(teskje)` comments were added:
- `src/sql/src/plan/explain.rs:65`: `// TODO(teskje): collect and log a backtrace from the panic site`
- `src/transform/src/lib.rs:332`: `// TODO(teskje): collect and log a backtrace from the panic site`

**nit:** Consider filing a tracking issue for these TODOs if the work is planned, to avoid them becoming stale.

### 2. Repeated `#[allow(clippy::disallowed_methods)]` pattern in tests

Multiple test files now use `std::panic::catch_unwind` with `#[allow(clippy::disallowed_methods)]` annotations (5 places). The comments explain "not using enhanced panic handler in tests" which is clear.

**nit:** If more test files need this pattern in the future, consider adding a helper function in a test utilities module, or documenting this pattern in a testing guide.

### 3. Feature name

**nit:** The new feature is named `panic` which matches the module name. A more descriptive name like `panic-handler` or `enhanced-panic` could improve clarity, but `panic` is fine.

---

## Summary

This PR is a well-structured fix for a real operational issue. The approach is sound:

1. Disable Sentry's default panic hook (by removing the `sentry-panic` integration from the `sentry` crate's features)
2. Manually invoke Sentry's panic handler in the enhanced handler *after* the catch-unwind check
3. Use a feature flag to avoid unconditional dependency on `sentry-panic`

The test changes to use `std::panic::catch_unwind` directly are reasonable since tests don't use the enhanced panic handler.

**Verdict:** This PR improves codebase health by fixing a bug that caused operational confusion. The strong suggestions above should be addressed (or at least clarified), but the overall approach is correct.

---
*Automated review by Claude Code using `/mz-pr-review`*

PR #32232, opened 2025-04-16
docs: Update namespaces docs

kay-kim:
- Various edits of namespaces page. 
- Bump up prominence of namespaces content by also including in concepts in addition to reference section where it could be easily overlooked
- Use this as an opportunity to start refactoring how we use shared content vs shortcodes.  We've been using shortcodes to single-source content ... which is fine ... but may not be optimal in some of our use cases.  For example, I think that's why the doc currently list create sinks: Redpanda in the left-hand nav but it just redirects to the kafka page - making it a weird jump experience for our users - instead of just having a single source and reusing parts.


doc/user/layouts/shortcodes/kwlist.html
```diff
@@ -1,7 +1,15 @@
-<ul>
-{{range split (readFile ("sql-grammar/keywords.txt")) "\n"}}
-  {{if and (not (hasPrefix . "#")) .}}
-    <li><code>{{.|upper}}</code></li>
+
+<table border="0" style="width:100%; border-collapse: collapse;">
```
kay-kim, line 2: Instead of a single list, make this a multi-column list

doc/user/content/concepts/namespaces.md
```diff
@@ -0,0 +1,12 @@
...
+description: "Logical namespaces for organizing objects in Materialize."
+weight: 40
+menu:
+  main:
+    parent: concepts
+    identifier: 'namespaces'
+    weight: 126
+---
+
+{{< include-md file="shared-content/namespaces-content.md" >}}
```
kay-kim, line 12: https://preview.materialize.com/materialize/32232/concepts/namespaces/

doc/user/content/sql/namespaces.md
```diff
@@ -9,91 +9,4 @@ menu:
...
-- `database.schema.source` to reach a source in another database.
-
-### Database details
-
-- Unlike PostgreSQL, Materialize allows cross-database queries.
-- By default, Materialize regions have a database named `materialize`.
-- By default, each database has a schema called `public`.
-- You can specify which database you connect to either when you connect (e.g.
-  `psql -d my_db ...`) or within SQL using `SET` (e.g. `SET DATABASE = my_db`).
+{{< include-md file="shared-content/namespaces-content.md" >}}
```
kay-kim, line 12: https://preview.materialize.com/materialize/32232/sql/namespaces/

doc/user/shared-content/namespaces-content.md
```diff
@@ -0,0 +1,120 @@
...
+    ```
+    <Source>
+    <View>
+    <Table>.<Column>
+    <View>.<Column>
+    ```
+
+## Namespace constraints
+
+- All namespaces must adhere to [identifier rules](/sql/identifiers).
```
ptravers, line 75: nit: only use bullet points if there are multiple options? but understand if this is there to be consistent between sub headings? or be clear that there is a fact here?
```suggestion
All namespaces must adhere to [identifier rules](/sql/identifiers).
```

doc/user/shared-content/namespaces-content.md
```diff
@@ -0,0 +1,120 @@
...
+
+Materialize follows SQL standard's namespace hierarchy for most objects (for the
+exceptions, see [Other objects](#other-objects)).
+
+|                           |             |
+|---------------------------| ------------|
+| 1st/Highest level:        |  **Database** |
+| 2nd level:                |  **Schema**   |
+| 3rd level:                | <table><tbody><tr><td><ul><li>**Table**</li><li>**View**</li><li>**Materialized view**</li><li>**Connection**</li></ul></td><td><ul><li>**Source**</li><li>**Sink**</li><li>**Index**</li></ul></td><td><ul><li>**Type**</li><li>**Function**</li><li>**Secret**</li></ul></td></tr></tbody></table>|
+| Lowest level:             | **Column**     |
```
ptravers, line 15: nit: do we want this to be comparable to the first entry in the table? probably would make it less legible than as currently written.
```suggestion
| 4th/Lowest level:             | **Column**     |
```

ptravers (approved):

doc/user/shared-content/namespaces-content.md
ptravers: might also read a little smoother to just not have the number at the front for Highest and Lowest?

PR #32245, opened 2025-04-17
Mz system password support

alex-hunt-materialize:
* Adds mz_system password support.
* Moves configuration of internal and external ports to a config file.

### Checklist

    For expediency, I've created https://github.com/MaterializeInc/cloud/issues/11196 rather than add it here. Nothing uses it yet, it just needs to be in before we add console support.
  
  https://github.com/MaterializeInc/cloud/pull/11161


src/cloud-resources/src/crd/materialize.rs
```diff
@@ -128,6 +128,10 @@ pub mod v1alpha1 {
         pub in_place_rollout: bool,
         // The name of a secret containing metadata_backend_url and persist_backend_url.
         pub backend_secret_name: String,
+        // The name of a secret containing a `password` key with credentials
+        // for the mz_system user.
+        // Setting this will enable password authentication.
+        pub external_login_secret_mz_system: Option<String>,
```
doy-materialize, line 134: is there a reason we're using a separate secret for this?

deny.toml
```diff
@@ -116,6 +116,9 @@ skip = [
     { name = "rand_core", version = "0.6.2" },
     { name = "reqwest", version = "0.11.24" },
     { name = "rustls-pemfile", version = "1.0.4" },
+    # Used by reqwest
+    { name = "system-configuration", version = "0.5.1" },
+    { name = "system-configuration-sys", version = "0.5.0" },
```
doy-materialize, line 121: what changed here? is there a way to tweak the dependencies to remove this duplication instead?

src/cloud-resources/src/crd/materialize.rs
alex-hunt-materialize: Separate secret from what?

deny.toml
alex-hunt-materialize: I believe this came in as part of adding the cookies feature to reqwest. We could maybe upgrade reqwest everywhere to remove the duplication there, but it seemed like a larger change.

src/cloud-resources/src/crd/materialize.rs
doy-materialize: from the one that we use to configure the backend urls and license key

alex-hunt-materialize: I think that was to leave open the possibility of having additional secrets stored for other users. I'm not sure how relevant that is now, though.

src/environmentd/src/test_util.rs
```diff
@@ -246,18 +304,111 @@ impl TestHarness {
...
 
-    pub fn with_self_hosted_auth(mut self, self_hosted_auth: bool) -> Self {
-        self.self_hosted_auth = self_hosted_auth;
-        self
-    }
-
-    pub fn with_self_hosted_auth_internal(mut self, self_hosted_auth_internal: bool) -> Self {
-        self.self_hosted_auth_internal = self_hosted_auth_internal;
+    pub fn with_password_auth(mut self) -> Self {
+        // TODO include mz_system password?
```
doy-materialize: yeah, it'd be great if we could get some tests for whether password authentication is actually working

src/orchestratord/src/controller/materialize/environmentd.rs
```diff
@@ -1419,6 +1537,214 @@ fn create_environmentd_statefulset_object(
...
+fn create_connection_info(
+    config: &super::MaterializeControllerArgs,
+    mz: &Materialize,
+    generation: u64,
+) -> ConnectionInfo {
+    let authenticator_kind = match (
+        mz.spec
+            .environmentd_extra_args
+            .as_ref()
+            .map(|args| args.iter().any(|arg| arg.starts_with("--frontegg"))),
```
doy-materialize: this feels... fragile. is there a way we can be more explicit here?

src/orchestratord/src/controller/materialize/environmentd.rs
```diff
@@ -73,6 +92,20 @@ pub enum DeploymentStatus {
...
 
+#[derive(Deserialize, Serialize)]
+pub struct LoginCredentials {
+    username: String,
+    // TODO Password type?
+    password: ByteString,
+}
+
+#[derive(Debug, Serialize)]
+pub struct ConnectionInfo {
```
doy-materialize, line 102: would it make sense to skip serializing the environmentd_url and mz_system_secret fields here? feels like the configmap contents are the only things we actually want to compare to see if there are changes to apply, and it might make things a bit easier to understand. probably not actually meaningfully different in practice though outside of debugging stuff.

doy-materialize (approved):
a couple more minor things, but the overall structure looks reasonable (i'm not super familiar with the details of like cors and tower middleware and things like that though)

src/authenticator/Cargo.toml
```diff
@@ -8,8 +8,8 @@ rust-version.workspace = true
 publish = false
 
 [dependencies]
-mz-frontegg-auth = { path = "../frontegg-auth" }
-mz-server-core = { path = "../server-core" }
+mz-adapter = { path = "../adapter", default-features = false }
+mz-frontegg-auth = { path = "../frontegg-auth", default-features = false }
```
DAlperin, line 12: what features are we trying to disable here?

src/catalog-debug/src/main.rs
```diff
@@ -655,6 +655,8 @@ async fn upgrade_check(
             enable_expression_cache_override: None,
             enable_0dt_deployment: true,
             helm_chart_version: None,
+            // TODO pass this in here?
```
DAlperin: Happy to leave this for now, but could we tag this with something like TODO(auth) for reference?

src/adapter/src/catalog/open.rs
```diff
@@ -246,6 +247,18 @@ impl Catalog {
                 };
             }
             state.create_temporary_schema(&SYSTEM_CONN_ID, MZ_SYSTEM_ROLE_ID)?;
+            if let Some(password) = config.external_login_password_mz_system {
```
DAlperin, line 250: Is this populated directly from the CLI flag? If so this means that the user can change the mz_system var just by restarting with a different password. Is this desired? Or should this only be for bootstrap?

src/environmentd/src/environmentd/main.rs
```diff
@@ -802,21 +759,32 @@ fn run(mut args: Args) -> Result<(), anyhow::Error> {
...
+                .push(HeaderValue::from_str(&format!("http://localhost:{}", port)).unwrap());
+            allowed_origins
+                .push(HeaderValue::from_str(&format!("http://127.0.0.1:{}", port)).unwrap());
+            allowed_origins.push(HeaderValue::from_str(&format!("http://[::1]:{}", port)).unwrap());
+            allowed_origins
+                .push(HeaderValue::from_str(&format!("https://localhost:{}", port)).unwrap());
+            allowed_origins
+                .push(HeaderValue::from_str(&format!("https://127.0.0.1:{}", port)).unwrap());
+            allowed_origins
+                .push(HeaderValue::from_str(&format!("https://[::1]:{}", port)).unwrap());
```
DAlperin: nit: could make these a vec! of headers than call allowed_origins.extend() with it.

src/environmentd/src/http.rs
```diff
@@ -696,21 +713,111 @@ impl IntoResponse for AuthError {
     }
 }
 
+// type Auth = (Json<LoginCredentials>, towerSessions)
```
DAlperin: errant comment?

src/environmentd/src/http.rs
```diff
@@ -696,21 +713,111 @@ impl IntoResponse for AuthError {
...
+// Simplified login handler
+pub async fn handle_login(
+    session: Option<Extension<TowerSession>>,
+    Extension(adapter_client_rx): Extension<Delayed<Client>>,
+    Json(LoginCredentials { username, password }): Json<LoginCredentials>,
+) -> impl IntoResponse {
+    let Ok(adapter_client) = adapter_client_rx.clone().await else {
+        return (
+            StatusCode::INTERNAL_SERVER_ERROR,
+            Json(json!({ "status": "false" })),
```
DAlperin: status: false/failed/success isn't a super descriptive message. could we maybe do `message: "Internal Server Error" or something if we want a response? That said, on failure we already set the headers, we would probably be fine with an empty body.

src/environmentd/src/http.rs
```diff
@@ -118,18 +124,6 @@ pub struct InternalRouteConfig {
...
-pub struct ReloadingTlsConfig {
-    pub context: ReloadingSslContext,
-    pub mode: TlsMode,
-}
-
-#[derive(Debug, Clone, Copy)]
-pub enum TlsMode {
-    Disable,
-    Require,
-}
```
DAlperin, line 131: Why did we kill this out of curiosity?

src/environmentd/src/lib.rs
```diff
@@ -243,36 +233,104 @@ impl Listeners {
...
+            AuthenticatorKind::Frontegg => Authenticator::Frontegg(
+                frontegg.expect("Frontegg args are required with AuthenticatorKind::Frontegg"),
+            ),
+            AuthenticatorKind::Password => Authenticator::Password(adapter_client.clone()),
+            AuthenticatorKind::None => Authenticator::None,
+        };
+
+        task::spawn(|| format!("{}_sql_server", label), {
+            let sql_server = mz_pgwire::Server::new(mz_pgwire::Config {
+                label,
```
DAlperin, line 277: Lets update mz_pgwire::Config to take a capital S `String` so we can avoid leaking this on the heap. You'll have to update the Metrics struct in pg_wire but thats pretty painless.

src/environmentd/src/lib.rs
```diff
@@ -243,36 +233,104 @@ impl Listeners {
...
+                dyncfg: None,
+            })
+        });
+        self.handle
+    }
+}
+
+pub struct Listeners {
+    pub http: BTreeMap<String, Listener<HttpListenerConfig>>,
+    pub sql: BTreeMap<String, Listener<SqlListenerConfig>>,
```
DAlperin, line 318: I think a BTree is probably overkill here. Is there any access pattern where we care about the order or quick searches by key?

src/environmentd/src/lib.rs
```diff
@@ -284,95 +342,86 @@ impl Listeners {
...
+                AuthenticatorKind::Password => authenticator_password_rx.clone(),
+                AuthenticatorKind::None => authenticator_none_rx.clone(),
+            };
             let metrics_registry = config.metrics_registry.clone();
-            let metrics = http::Metrics::register_into(&metrics_registry, "mz_internal_http");
-            let internal_http_server = HttpServer::new(HttpConfig {
+            let metrics_component: &'static str =
+                Box::leak(format!("mz_{}_http", &name).into_boxed_str());
+            let metrics = http::Metrics::register_into(&metrics_registry, metrics_component);
+            let source: &'static str = Box::leak(name.clone().into_boxed_str());
```
DAlperin: This doesn't feel great, but I understand why you have to do this with our current APIs :(. If we can update the call sites to take owned Strings that would be good, but I won't block this PR on it. There is no circumstance where these http servers can come down at runtime without taking the process with them, right?

src/orchestratord/src/controller/materialize/environmentd.rs
```diff
@@ -1419,6 +1537,214 @@ fn create_environmentd_statefulset_object(
...
+        (_, false) => AuthenticatorKind::None,
+    };
+
+    let external_enable_tls = issuer_ref_defined(
+        &config.default_certificate_specs.internal,
+        &mz.spec.internal_certificate_spec,
+    );
+
+    let listeners_json = serde_json::to_string(&match authenticator_kind {
+        AuthenticatorKind::Frontegg => {
```
DAlperin: If these configs are mostly the same save for the authenticator_kind, we could probably consolidate their creation a bit so we don't have to explicitly write it out twice if you wanted. But also the explicitness could be nice. Your call.

src/server-core/Cargo.toml
```diff
@@ -14,14 +14,15 @@ anyhow = "1.0.98"
...
+schemars = { version = "0.8" }
 scopeguard = "1.2.0"
 serde = { version = "1.0.219", features = ["derive"] }
 socket2 = "0.5.10"
 tokio-stream = "0.1.17"
 proxy-header = "0.1.2"
 tracing = "0.1.37"
 futures = "0.3.31"
-mz-dyncfg = { path = "../dyncfg" }
+mz-dyncfg = { path = "../dyncfg", default-features = false }
```
DAlperin, line 25: what features do we want to disable here?

DAlperin (commented):
Looks good! Just a few comments.

src/authenticator/Cargo.toml
alex-hunt-materialize: We don't need any features, but we especially want to disable workspace-hack. We don't want to include that when materialize crates are used by the cloud repo.

I actually plan to add a lint that enforces that, but haven't gotten to it yet.

src/server-core/Cargo.toml
alex-hunt-materialize: Same as above, we don't seem to need any of them, but we especially don't want workspace-hack.

src/environmentd/src/lib.rs
alex-hunt-materialize: Yeah, I agree. It feels a bit gross to leak, but these servers do last the entire lifetime of the process. This PR is already huge, and I didn't want to deal with refactoring that.

src/environmentd/src/lib.rs
alex-hunt-materialize: Consistent order is nice that when you have the same json input you'll have the same order later, like when we print out the addresses of the listeners. It isn't a big deal though.

We also do some lookups of addresses by name, mostly in the tests. These don't have to be super fast, though.

I could get away with `Vec<(String, Listener<T>)>` with slightly more annoying lookups. Does this matter?

src/environmentd/src/http.rs
alex-hunt-materialize: Consistency of interface in the json configs between HTTP and SQL listeners. There we just have a bool, which is just as easy to work with as this enum. I didn't see any reason to convert to this enum when we already have the same information in the bool. There were no other places that used it, so I deleted the unused enum.

src/environmentd/src/http.rs
alex-hunt-materialize: Agreed. I think this was a copy/paste of some dev code and I forgot to update it.

src/adapter/src/catalog/open.rs
alex-hunt-materialize: Yes, they can change the password of the mz_system user just by restarting with different flags. We discussed this earlier, and decided it was fine behavior.

src/catalog-debug/src/main.rs
alex-hunt-materialize: Removed, as this isn't needed for catalog-debug.

src/environmentd/src/test_util.rs
alex-hunt-materialize: I've just pushed a commit that exercises the mz_system password in the auth tests.

src/environmentd/src/http.rs
alex-hunt-materialize: I've removed the json response.

src/environmentd/src/lib.rs
alex-hunt-materialize: >but thats pretty painless

We pass these `&'static str`s to functions from the prometheus crate (not controlled by us). Changing this to `String` is not trivial, at least as far as I can tell.

doy-materialize: so the reason the prometheus crate works this way is because the way it is intended to be used is to always have a fixed static string as the metric name with any variable components being associated via tags/labels. are we sure we don't actually want to do that here?

alex-hunt-materialize: I think the labels also need to be static, but yes, it would be better to just have a single metric with a new label for which listener is responding. If I have time, I'll try to change that.

SangJunBak:
@alex-hunt-materialize Heyo! Wanted to call out both these still don't work! 
Websocket authentication doesn't work: https://materializeinc.slack.com/archives/D0801TD4PT2/p1745443566530209. Getting the warning `environmentd: 2025-06-04T19:32:08.724323Z  WARN mz_environmentd::http: self hosted auth only, but somehow missing session data`. CURL:
```
curl 'ws://localhost:3000/api/experimental/sql' \
  -H 'Upgrade: websocket' \
  -H 'Origin: http://localhost:3000' \
  -H 'Cache-Control: no-cache' \
  -H 'Accept-Language: en-US,en;q=0.9' \
  -H 'Pragma: no-cache' \
  -b 'hubspotutk=40ce74eab26d70198c7a9b675d02d733; intercom-device-id-r8661p0d=9d4b7039-2522-4104-8408-6304b8a13060; __stripe_mid=a644d7a6-edd2-4061-ac99-654415266d9f093869; _gcl_au=1.1.371382533.1748035557; __hssrc=1; __hstc=181257784.40ce74eab26d70198c7a9b675d02d733.1731961420516.1748975673822.1749015596388.264; ajs_user_id=c9cb3c44-51ad-490f-9d49-a7ff3302e817; ajs_anonymous_id=3c337e37-33d6-4322-9c13-8df7193fbf2a; __stripe_sid=78a33b51-be6a-47b2-8915-93dba884a0459d345a; mz_session=NKSUTvQ6YQcOJF4UM26Rew' \
  -H 'Connection: Upgrade' \
  -H 'Sec-WebSocket-Key: DnBNw0Ney9DsY2Izq1vfMg==' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36' \
  -H 'Sec-WebSocket-Version: 13' \
  -H 'Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits'
```

`www_authenticate header` being annoying as UI: https://materializeinc.slack.com/archives/D0801TD4PT2/p1745441777890019

src/cloud-resources/src/crd/materialize.rs
```diff
@@ -127,7 +129,12 @@ pub mod v1alpha1 {
         #[serde(default)]
         pub in_place_rollout: bool,
         // The name of a secret containing metadata_backend_url and persist_backend_url.
+        // It may also contain external_login_password_mz_system, which will be used as
+        // the password for the mz_system user if authenticator_kind is Password.
         pub backend_secret_name: String,
+        // How to authenticate with Materialize. Valid options are Password and None.
+        // If set to Password, the backend secret must contain external_login_password_mz_system.
+        pub authenticator_kind: AuthenticatorKind,
```
doy-materialize, line 137: does this work for existing environments without the `#[serde(default)]` annotation?

PR #32246, opened 2025-04-17
Require `RowArena::clear` to take a `&mut self`

frankmcsherry:
The `RowArena` type has unsafe internals, and in particular wraps `Vec<u8>` stuffs with a `RefCell`, but wants to return `&'a [u8]` that outlives `borrow_mut()` calls. This works as long as we don't de/re-allocate the `Vec<u8>` for the lifetime of the type, because the `'a` comes from `&'a self`.

The `RowArena::clear` method upsets this, by deallocating the memory under a `&self` reference. If we change that to be a `&mut self` reference, it seems like this should be .. more safe, at least. The mutable reference should require that all borrowed references have expired.

### Checklist

### DIFF
--- a/src/compute/src/render/context.rs
+++ b/src/compute/src/render/context.rs
@@ -903,7 +903,7 @@ where
                         let mut key_buf = Row::default();
                         let mut val_buf = Row::default();
                         let mut datums = DatumVec::new();
-                        let temp_storage = RowArena::new();
+                        let mut temp_storage = RowArena::new();
                         while let Some((time, data)) = input.next() {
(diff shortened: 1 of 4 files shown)


ParkMyCar (approved):
Chatted offline, this looks good to me!

Main points I looked at were the uses of `transmute` in `impl RowArena` to ensure that lifetimes were not getting extended such that the `&mut self` on clear wouldn't be effective. But all of our uses of transmute, change the lifetime to `'a` where `&'a self` is used as the first argument in those functions. So we look good!

PR #32248, opened 2025-04-17
Cluster spec sheet reproducer

antiguru:
Python script to run measurements for the cluster spec sheet effort

Run: `bin/mzcompose --find cluster-spec-sheet down; bin/mzcompose --find cluster-spec-sheet run default`

Export `REGION` and `ENVIRONMENT`, `NIGHTLY_MZ_USERNAME`, and `MZ_CLI_APP_PASSWORD` for your staging account.

Places a file `results_TIMESTAMP.csv` in the working directory. The file contains roughly the following columns:
* **scenario** is a name of what workload is running.
* **mode** is either weak or strong for weak and strong scaling.
* **test_name** is what we're doing specifically.
* **cluster_size** shows the size of the cluster the test ran on.
* **repetition** is a number of the repetition of the same test.
* **size_bytes** shows how much data there is in the index, either input or output, if we can determine the size.
* **time_ms** how long the test took.

The `plot` workflow accepts a results file parameter and renders the results as tables and images, both as time elapsed and credits consumed, as a function of workload and cluster size.


antiguru:
Currently running in CI: https://buildkite.com/materialize/release-qualification/builds/909#0198a2d7-9bba-496d-a181-740eb5a886bd

ci/release-qualification/pipeline.template.yml
```diff
@@ -511,3 +511,17 @@ steps:
...
         agents:
           queue: hetzner-x86-64-dedi-48cpu-192gb # 1 TB disk
+
+  - group: Cluster spec sheet
+    key: cluster-spec-sheet-group
+    steps:
+      - id: cluster-spec-sheet
+        label: Cluster spec sheet
+        depends_on: build-aarch64
+        timeout_in_minutes: 3600
```
def-, line 546:
```suggestion
        timeout_in_minutes: 3600
        concurrency: 1
        concurrency_group: 'cloud-canary'
```
I think if the Cloud Canary test will run at the same time it will collide?

ci/plugins/mzcompose/hooks/command
```diff
@@ -255,7 +255,7 @@ cleanup() {
...
+  if [ ! -s services.log ] && [ "$BUILDKITE_LABEL" != "Maelstrom coverage of persist" ] && [ "$BUILDKITE_LABEL" != "Long single-node Maelstrom coverage of persist" ] && [ "$BUILDKITE_LABEL" != "Maelstrom coverage of txn-wal" ] && [ "$BUILDKITE_LABEL" != "Mz E2E Test" ] && [ "$BUILDKITE_LABEL" != "Output consistency (version for DFR)" ] && [ "$BUILDKITE_LABEL" != "Output consistency (version for CTF)" ] && [ "$BUILDKITE_LABEL" != "QA Canary Environment Base Load" ] && [ "$BUILDKITE_LABEL" != "Parallel Benchmark against QA Canary Environment" ] && [ "$BUILDKITE_LABEL" != "Parallel Benchmark against QA Benchmarking Staging Environment" ] && [[ ! "$BUILDKITE_LABEL" =~ Terraform\ .* ]] && [ ! "$BUILDKITE_LABEL" != "Cluster spec sheet"]; then
```
def-:
```suggestion
  if [ ! -s services.log ] && [ "$BUILDKITE_LABEL" != "Maelstrom coverage of persist" ] && [ "$BUILDKITE_LABEL" != "Long single-node Maelstrom coverage of persist" ] && [ "$BUILDKITE_LABEL" != "Maelstrom coverage of txn-wal" ] && [ "$BUILDKITE_LABEL" != "Mz E2E Test" ] && [ "$BUILDKITE_LABEL" != "Output consistency (version for DFR)" ] && [ "$BUILDKITE_LABEL" != "Output consistency (version for CTF)" ] && [ "$BUILDKITE_LABEL" != "QA Canary Environment Base Load" ] && [ "$BUILDKITE_LABEL" != "Parallel Benchmark against QA Canary Environment" ] && [ "$BUILDKITE_LABEL" != "Parallel Benchmark against QA Benchmarking Staging Environment" ] && [[ ! "$BUILDKITE_LABEL" =~ Terraform\ .* ]] && [ ! "$BUILDKITE_LABEL" != "Cluster spec sheet" ]; then
```

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -0,0 +1,1187 @@
...
+# included in the LICENSE file at the root of this repository.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0.
+
+"""
+Deploy the current version on a real Staging Cloud, and run some basic
+verifications, like ingesting data from Kafka and Redpanda Cloud using AWS
+Privatelink. Runs only on main and release branches.
```
def-: Could update the description, this shows up with `./mzcompose description` (and describe)

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -0,0 +1,1187 @@
...
+        print(
+            f"--- Running {name} for {self.replica_size} with {repetitions} repetitions..."
+        )
+        for repetition in range(repetitions):
+            start_time = time.time()
+            for query_part in query:
+                self.run_query(query_part)
+            end_time = time.time()
+            if size_of_index:
+                time.sleep(2)
```
def-: I'd prefer retries until we get the expected result instead of a fixed sleep

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -0,0 +1,1187 @@
...
+import re
+import time
+from abc import abstractmethod
+from collections.abc import Callable, Hashable
+from textwrap import dedent
+from typing import Any
+
+import matplotlib.pyplot as plt
+import pandas as pd
+import pg8000.native
```
def-: I prefer psycopg, but I assume you're not transferring large amounts of data through this connection, so probably doesn't matter.

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -0,0 +1,1187 @@
...
+    def drop(self) -> list[str]:
+        pass
+
+    @abstractmethod
+    def materialize_views(self) -> list[str]:
+        pass
+
+    @abstractmethod
+    def run(self, runner: ScenarioRunner) -> None:
+        pass
```
def-:
```suggestion
```

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -0,0 +1,1187 @@
...
+        raise NotImplementedError("Subclasses should implement this method")
```
def-: I thought all methods have to be implemented if they are abstract?

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -0,0 +1,1187 @@
...
+
+
+def workflow_default(c: Composition, parser: WorkflowArgumentParser) -> None:
+    def process(name: str) -> None:
+        if name == "default":
+            return
+        with c.test_case(name):
+            c.workflow(name, *parser.args)
+
+    c.test_parts(list(c.workflows.keys()), process)
```
def-: I assume you want to run `bench` first, then `plot`? This probably works, but I'd prefer to make the order more explicit than relying on alphabetical order.

def-: If there is no parallelization of this step, then you can just skip the whole c.test_parts
```suggestion
    bench(parser)
    plot()
```
or something like that? Does it make sense to run them separately?

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -0,0 +1,1187 @@
...
+        runner.run_query(f"CREATE CLUSTER c SIZE '{replica_size}'")
+        runner.run_query("SET cluster = 'c';")
+        runner.run_query("SELECT * FROM t;")
+
+        runner.replica_size = replica_size
+
+        scenario.run(runner)
+
+
+def run_scenario_weak(
```
def-, line 1146: What is the difference between a strong and weak scenario?

test/cluster-spec-sheet/mzcompose.py
antiguru: Added a comment to explain this.

test/cluster-spec-sheet/mzcompose.py
antiguru: Fixed.

test/cluster-spec-sheet/mzcompose.py
antiguru: Fixed

test/cluster-spec-sheet/mzcompose.py
antiguru: Added a brief comment to explain the difference.

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -0,0 +1,1294 @@
...
+                scenario=TpchScenarioMV(1, replica_size_for_scale(1)),
+                results_file=f,
+                connection=conn,
+            )
+            print("+++ Running Auction strong scaling")
+            run_scenario_strong(
+                scenario=AuctionScenario(4, replica_size_for_scale(1)),
+                results_file=f,
+                connection=conn,
+            )
```
antiguru: @ggevay You can comment all but the last `run_scenario_weak`.

antiguru:
It's mostly working! https://buildkite.com/materialize/release-qualification/builds/936#_

ci/release-qualification/pipeline.template.yml
```diff
@@ -536,3 +536,19 @@ steps:
...
+        concurrency: 1
+        concurrency_group: 'cloud-canary'
+        plugins:
+          - ./ci/plugins/mzcompose:
+              composition: cluster-spec-sheet
+              run: default
```
def-, line 552:
```suggestion
              run: default
              args: [--cleanup]
```
At least the docs say you should specify --cleanup

test/cluster-spec-sheet/analysis.ipynb
def-, line 1: I don't know how to review this, so just skipping it.

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -0,0 +1,1454 @@
...
+                "SELECT * FROM t;",
+            ],
+            query=[
+                "CREATE DEFAULT INDEX ON lineitem;",
+                "SELECT count(*) > 0 FROM lineitem;",
+            ],
+            size_of_index="lineitem_primary_idx",
+        )
+
+        time.sleep(3)
```
def-: Is this to make sure the index is used?

def- (changes requested):
The plots are missing here: https://buildkite.com/materialize/release-qualification/builds/936#019910a6-2745-473f-9379-7efa58968858

> [Image artifact '/var/lib/buildkite-agent/builds/buildkite-aarch64-small-d306b64-i-03ebf87dfe0a1c133-1/materialize/release-qualification/test/cluster-spec-sheet/plots/results_1756920982/auction_arrangement_formation_strong_time_ms.png' not found]

I think it's because you use the absolute path, but buildkite cuts off the initial /. I'd just use a relative path instead.

Do we need to get the data out of buildkite? Buildkite will delete logs and artifacts after 6 months: https://buildkite.com/docs/pipelines/configure/artifacts

ci/release-qualification/pipeline.template.yml
antiguru: Correct, added.

test/cluster-spec-sheet/mzcompose.py
antiguru: Removed

test/cluster-spec-sheet/analysis.ipynb
antiguru: Removed the file

antiguru:
> I think it's because you use the absolute path, but buildkite cuts off the initial /. I'd just use a relative path instead.

Yup, I changed it. It should use the absolute path when writing the file, and a relative path to upload the artifact.

Do we need to get the data out of buildkite? Buildkite will delete logs and artifacts after 6 months: https://buildkite.com/docs/pipelines/configure/artifacts

Yes, in the long run we certainly want to record the numbers. I'd prefer to do this in a follow-up change as the PR itself is already huge. Recorded as https://github.com/MaterializeInc/database-issues/issues/9649

PR #32271, opened 2025-04-21
controller: record wallclock lag at minute boundaries

teskje:
This PR makes both controllers emit their pending wallclock lag introspection updates when time passes a minute boundary, instead of emitting them at some random moment every minute. This is meant to synchronize the two controllers, reducing the chance of inconsistent introspection data where lag spikes of dependent collections (e.g. a source and an index) are reported at different minutes.

To simplify the implementation, this PR includes two changes to reduce the complexity of the wallclock lag tracking code:
* It merges the feature flags controlling the wallclock lag recording interval (first commit).
* It factors out wallclock lag recording (as opposed to updating the in-memory tracking) into a separate method.

### Motivation

The two controllers can emit wallclock lag introspection updates at different moments within each minute, causing confusion because logically dependent lags of dependent collections can end up at different minutes in the introspection output. See [Slack thread](https://materializeinc.slack.com/archives/C08AENAP691/p1744989938902549).

### Checklist

### DIFF
--- a/src/compute-client/src/controller/instance.rs
+++ b/src/compute-client/src/controller/instance.rs
@@ -26,10 +26,7 @@ use mz_compute_types::sinks::{
     ComputeSinkConnection, ComputeSinkDesc, ContinualTaskConnection, MaterializedViewSinkConnection,
 };
 use mz_compute_types::sources::SourceInstanceDesc;
-use mz_controller_types::dyncfgs::{
-    ENABLE_WALLCLOCK_LAG_HISTOGRAM_COLLECTION, WALLCLOCK_LAG_HISTOGRAM_REFRESH_INTERVAL,
-    WALLCLOCK_LAG_HISTORY_REFRESH_INTERVAL,
-};
+use mz_controller_types::dyncfgs::{ENABLE_WALLCLOCK_LAG_HISTOGRAM_COLLECTION, WALLCLOCK_LAG_RECORDING_INTERVAL};
 use mz_dyncfg::ConfigSet;
 use mz_expr::RowSetFinishing;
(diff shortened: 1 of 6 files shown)


src/compute-client/src/controller/instance.rs
```diff
@@ -629,31 +598,94 @@ impl<T: ComputeControllerTimestamp> Instance<T> {
...
+        let interval = WALLCLOCK_LAG_RECORDING_INTERVAL.get(&self.dyncfg);
+        let now_dt = mz_ore::now::to_datetime((self.now)());
+        let now_trunc = duration_trunc(now_dt, interval).unwrap_or_else(|| {
+            soft_panic_or_log!("excessive wallclock lag recording interval: {interval:?}");
+            let default = WALLCLOCK_LAG_RECORDING_INTERVAL.default();
+            duration_trunc(now_dt, *default).unwrap()
+        });
+        if now_trunc <= self.wallclock_lag_last_recorded {
+            return;
+        }
```
antiguru, line 631: Does this risk causing workload spikes at the minute interval? It seems the previous approach smeared the work across the interval, while the new approach piles all updates on the same moment in time. This works as long as the amount of updates doesn't grow too much, but can cause problems once we track the lag of a significant number of collections.

antiguru (approved):
I think this PR is fine, but if I'm reading it correctly, it risks workload spikes at the tick of the interval. We need to ensure this wouldn't cause availability problems in the system.

A common solution is to round to the interval and then add a little bit of randomness so the work gets smoothed out across several seconds.

teskje:
TFTR!

PR #32301, opened 2025-04-21
Store active rollups in state

DAlperin:
Closes https://github.com/MaterializeInc/database-issues/issues/9193

### Checklist


src/persist-client/src/internal/state.rs
```diff
@@ -1107,6 +1107,19 @@ pub enum HollowBlobRef<'a, T> {
     Rollup(&'a HollowRollup),
 }
 
+/// A rollup that is currently being computed.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Arbitrary, Serialize)]
+pub struct ActiveRollup {
+    /// 0 if no rollup is being computed.
+    pub start_ms: u64,
```
bkirwi, line 1114: I'd much rather store an `Option<ActiveRollup>` instead of special-casing the zero value here.

Partly for general data-modelling reasons - makes it harder to forget to handle the "no rollup in progress" case - but also because I think it will round-trip with existing states a little better.

src/persist-client/src/internal/state.rs
```diff
@@ -1107,6 +1107,19 @@ pub enum HollowBlobRef<'a, T> {
     Rollup(&'a HollowRollup),
 }
 
+/// A rollup that is currently being computed.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Arbitrary, Serialize)]
+pub struct ActiveRollup {
```
bkirwi, line 1112: Would you mind stashing the seqno of the in-progress rollup here as well? Seems like useful debug information at a minumum, might be handy later on, and should be cheap...

src/persist-client/src/internal/state.rs
```diff
@@ -1241,6 +1258,31 @@ where
...
+            self.active_rollup = active_rollup.clone();
+            Continue(active_rollup)
+        } else {
+            Break(NoOpStateTransition(active_rollup))
+        }
+    }
+
+    pub fn clear_active_rollup(
+        &mut self,
+    ) -> ControlFlow<NoOpStateTransition<ActiveRollup>, ActiveRollup> {
```
bkirwi, line 1276: This return value seems never used. (And it's not clear to me what the semantics are?)

src/persist-client/src/internal/encoding.rs
```diff
@@ -845,6 +855,7 @@ impl<T: Timestamp + Lattice + Codec64> RustType<ProtoRollup> for Rollup<T> {
             ts_codec: T::codec_name(),
             diff_codec: self.state.diff_codec.into_proto(),
             last_gc_req: self.state.state.collections.last_gc_req.into_proto(),
+            active_rollup: Some(self.state.state.collections.active_rollup.into_proto()),
```
bkirwi, line 858: I think this is part of your roundtripping problem... rollups from before your change will have `None` for this field, get the default filled in, then encode as `Some`. One option would be to filter out zeros here, but better IMO to have `StateCollections::active_rollup` be an option itself.

src/persist-client/src/internal/apply.rs
```diff
@@ -506,7 +508,20 @@ where
...
+        let write_rollup = new_state.need_rollup(
+            ROLLUP_THRESHOLD.get(cfg),
+            ROLLUP_USE_ACTIVE_ROLLUP.get(cfg),
+            u64::cast_from(ROLLUP_FALLBACK_THRESHOLD_MS.get(cfg)),
+            now,
+        );
+
+        if write_rollup.is_some() && ROLLUP_USE_ACTIVE_ROLLUP.get(cfg) {
+            new_state.collections.active_rollup = Some(ActiveRollup {
+                seqno: state.seqno,
```
bkirwi, line 521: Nit: let's use the seqno contained in the `write_rollup` here instead of the current seqno. (Currently they're the same, but that way they stay in sync.)

src/persist-client/src/internal/state.rs
```diff
@@ -2415,22 +2453,43 @@ where
...
+            // we should start a new rollup.
+            if seqnos_since_last_rollup > u64::cast_from(threshold)
+                && self.active_rollup().is_none()
+            {
+                return Some(self.seqno);
+            }
+
+            // If we have an active rollup, we need to check if it has been running too long.
+            // If it has, we should start a new rollup.
+            // This is to guard against a worker dying/taking too long/etc.
```
bkirwi, line 2468: This has slightly surprising-to-me behaviour: if there's a old/timed-out rollup here, we'll kick off a new one at the current seqno even if the most recent rollup is recent enough for us.

Should only really come up if the threshold is being tuned, though, which should be rare. Happy to keep it as is if you find this behaviour more intuitive.

src/persist-client/src/internal/state.rs
```diff
@@ -2415,22 +2453,43 @@ where
...
+            if let Some(active_rollup) = self.active_rollup() {
+                if now.checked_sub(active_rollup.start_ms).unwrap_or(0) > fallback_threshold_ms {
```
bkirwi, line 2470:
```suggestion
                if now.saturating_sub(active_rollup.start_ms) > fallback_threshold_ms {
```

src/persist-client/src/internal/state.rs
```diff
@@ -89,6 +89,22 @@ pub(crate) const ROLLUP_THRESHOLD: Config<usize> = Config::new(
...
+);
+
+/// Feature flag the new active rollup tracking mechanism.
+/// We musn't enable this until we are fully deployed on the new version.
+pub(crate) const ROLLUP_USE_ACTIVE_ROLLUP: Config<bool> = Config::new(
+    "persist_rollup_use_active_rollup",
+    false,
+    "Whether to use the new active rollup tracking mechanism.",
+);
+
```
bkirwi, line 107: Good flags!

src/persist-client/src/internal/state.rs
```diff
@@ -3609,29 +3677,213 @@ pub(crate) mod tests {
...
+            ROLLUP_FALLBACK_THRESHOLD_MS,
+            NOW
+        ));
+        state.seqno = SeqNo(7);
+        assert_none!(state.need_rollup(
+            ROLLUP_THRESHOLD,
+            ROLLUP_USE_ACTIVE_ROLLUP,
+            ROLLUP_FALLBACK_THRESHOLD_MS,
+            NOW
+        ));
```
bkirwi, line 3851: Thanks for these - seems thorough!

src/persist-client/src/internal/state_serde.json
```diff
@@ -11,52 +11,288 @@
       "encoded_size_bytes": null
     }
   },
+  "active_rollup": null,
```
bkirwi, line 14: ❇️

bkirwi (approved):
A couple minor suggestions, but this looks good!

src/persist-client/src/internal/state.rs
DAlperin: Good point. I'm fairly certain it shouldn't be possible to arrive in such a situation with my understanding of the state machine here, but I will move the logic into the guard to make it clearer.

bkirwi (approved):
GC changes also look good to me!

PR #32315, opened 2025-04-23
adapter: don't allow implicit read-then-write inside transactions

aljoscha:
INSERTS of non-constant values internally get sequenced as read-then-write queries, which clear transaction ops so disable the checks we normally have in place about mixing reads and writes in transactions.

Now, we check that a transaction is in a valid state when sequencing a read-then-write.

@def- Tests didn't catch this before.

### Checklist

### DIFF
--- a/src/adapter/src/session.rs
+++ b/src/adapter/src/session.rs
@@ -1160,6 +1160,34 @@ impl<T: TimestampManipulation> TransactionStatus<T> {
         }
     }
 
+    /// Checks whether the current state of this transaction allows writes
+    /// (adding write ops).
+    /// transaction
+    pub fn allows_writes(&self) -> bool {
+        match self {
+            TransactionStatus::Started(Transaction { ops, access, .. })
+            | TransactionStatus::InTransaction(Transaction { ops, access, .. })
+            | TransactionStatus::InTransactionImplicit(Transaction { ops, access, .. }) => {
+                match ops {
+                    TransactionOps::None => access != &Some(TransactionAccessMode::ReadOnly),
+                    TransactionOps::Peeks { determination, .. } => {
+                        // If-and-only-if peeks thus far do not have a timestamp
+                        // (i.e. they are constant), we can switch to a write
+                        // transaction.
+                        !determination.timestamp_context.contains_timestamp()
+                    }
+                    TransactionOps::Subscribe => false,
+                    TransactionOps::Writes(_) => true,
+                    TransactionOps::SingleStatement { .. } => false,
+                    TransactionOps::DDL { .. } => false,
+                }
+            }
+            TransactionStatus::Default | TransactionStatus::Failed(_) => {
+                unreachable!()
+            }
+        }
+    }
+
     /// Adds operations to the current transaction. An error is produced if they cannot be merged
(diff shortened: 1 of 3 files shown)


src/adapter/src/session.rs
```diff
@@ -1160,6 +1160,44 @@ impl<T: TimestampManipulation> TransactionStatus<T> {
...
+                match ops {
+                    TransactionOps::None => {
+                        if matches!(access, Some(TransactionAccessMode::ReadOnly)) {
+                            false
+                        } else {
+                            true
+                        }
+                    }
+                    TransactionOps::Peeks { determination, .. } => {
+                        // If-and-only-of peeks thus far do not have a timestamp
```
antiguru, line 1180:
```suggestion
                        // If-and-only-if peeks thus far do not have a timestamp
```

antiguru, line 1180:
```suggestion
                        // If-and-only-if peeks thus far do not have a timestamp
```

antiguru, line 1180:
```suggestion
                        // If-and-only-if peeks thus far do not have a timestamp
```

src/adapter/src/session.rs
```diff
@@ -1160,6 +1160,44 @@ impl<T: TimestampManipulation> TransactionStatus<T> {
...
+                        // (i.e. they are constant), we can switch to a write
+                        // transaction.
+                        if !determination.timestamp_context.contains_timestamp() {
+                            true
+                        } else {
+                            false
+                        }
```
antiguru, line 1187:
```suggestion
                        !determination.timestamp_context.contains_timestamp()
```

antiguru (approved):
Maybe add a test?

antiguru:
(Github had some trouble, thus the repeated suggestion.)

aljoscha:
> Maybe add a test?

As you were reviewing I was working on the test ... 😅 pushed it

src/adapter/src/session.rs
aljoscha: duh! 😅

src/adapter/src/session.rs
```diff
@@ -1160,6 +1160,40 @@ impl<T: TimestampManipulation> TransactionStatus<T> {
...
+            TransactionStatus::Started(Transaction { ops, access, .. })
+            | TransactionStatus::InTransaction(Transaction { ops, access, .. })
+            | TransactionStatus::InTransactionImplicit(Transaction { ops, access, .. }) => {
+                match ops {
+                    TransactionOps::None => {
+                        if matches!(access, Some(TransactionAccessMode::ReadOnly)) {
+                            false
+                        } else {
+                            true
+                        }
```
antiguru, line 1177:
```suggestion
                        access != Some(TransactionAccessMode::ReadOnly)
```

;-)

test/sqllogictest/transactions.slt
```diff
@@ -627,6 +627,62 @@ SELECT INTERVAL '1 day'
 statement ok
 COMMIT
 
+# Verify that different kinds of INSERT INTO ... statements are correctly
```
def-, line 630: Thanks for adding the test. I didn't think of trying this with non-constant values!

PR #32316, opened 2025-04-23
compute: identify LIR mappings by export IDs

teskje:
This PR removes `mz_dataflow_global_ids`. This introspection source mostly duplicated `mz_compute_exports` and only existed because LIR mappings were identified by `objects_to_build` IDs, not dataflow export IDs, and the `objects_to_build` IDs are not reported anywhere in our introspection.

We can avoid the duplication, simplify a bunch of code, and make LIR mapping introspection consistent with operator hydration introspection by instead identifying LIR mappings by export IDs. All the attribution queries using the LIR mappings should continue to work the same way, they just use different IDs in their joins now.

### Checklist


src/compute/src/render/context.rs
```diff
@@ -303,6 +310,26 @@ impl HydrationLogger {
...
+
+impl ComputeEventLogger {
+    /// Log the given compute events.
+    pub fn log_many<'a>(&self, events: impl Iterator<Item = &'a ComputeEvent>) {
+        self.logger.log_many(events);
+    }
+
+    /// Log a compute event for each dataflow export.
+    pub fn log_for_exports(&self, f: impl Fn(GlobalId) -> ComputeEvent) {
+        let events: Vec<_> = self.export_ids.iter().map(|id| f(*id)).collect();
```
antiguru, line 328: This doesn't need to allocate, doesn't it?

src/compute/src/logging/compute.rs
```diff
@@ -1251,20 +1186,21 @@ impl<A: Scheduler> DemuxHandler<'_, '_, A> {
...
+        let mappings = mapping.into_iter().map(Columnar::into_owned);
+        export.lir_mapping.extend(mappings);
 
         // send the datum out
         let ts = self.ts();
-        for (lir_id, meta) in mapping.into_iter() {
+        for (lir_id, meta) in mapping {
             let datum = LirMappingDatumReference {
-                global_id,
+                global_id: export_id,
```
antiguru, line 1203: Should we rename `LirMapping::global_id`?

antiguru (commented):
Nice!

src/compute/src/render/context.rs
teskje: Probably not, but I can't figure out the types! `log_many` wants an iterator over references but here we get an iterator over owned types, and to convert that into an iterator over references, we need to stash the owned types somewhere. Maybe `log_many` could be made to work with owned types somehow?

antiguru: Ah, I see. I think you can change the signature for `log_many` to mirror that of the underlying logger, which is generic over some `PushInto`.

test/sqllogictest/introspection/attribution_sources.slt
```diff
@@ -31,58 +31,52 @@ CREATE INDEX v_idx_x ON v(x);
...
+GROUP BY export_id, lir_id, operator, parent_lir_id, nesting
+ORDER BY export_id, lir_id DESC;
 ----
-u2  5  NULL  Join::Differential␠2␠»␠4
-u2  4  5  ␠␠Arrange␠3
-u2  3  4  ␠␠␠␠Get::Collection␠u1
-u2  2  5  ␠␠Arrange␠1
-u2  1  2  ␠␠␠␠Get::Collection␠u1
 u3  7  NULL  Arrange␠6
 u3  6  7  ␠␠Get::PassArrangements␠u2
```
teskje, line 74: This might be a blocker for merging this: The attribution query output now references a dataflow `u2` that we don't show anywhere, which is confusing. We probably need to do https://github.com/MaterializeInc/database-issues/issues/8845 first.

PR #32320, opened 2025-04-24
refactor reductionmonoid prevent clone of row on clone_from

ptravers:
### Motivation

https://github.com/MaterializeInc/database-issues/issues/9230

### Tips for reviewer

see: https://github.com/MaterializeInc/materialize/blob/be719a21a61b56b4afe9ee2852b84cb2b28fbd1c/src/repr/src/row.rs#L291

### Checklist

### DIFF
--- a/src/compute/src/render/reduce.rs
+++ b/src/compute/src/render/reduce.rs
@@ -2348,9 +2348,9 @@ mod monoids {
         fn clone_from(&mut self, source: &Self) {
             use ReductionMonoid::*;
 
-            let mut row = match std::mem::replace(self, Max(Row::default())) {
+            let mut row = std::mem::take(match self {
                 Min(row) | Max(row) => row,
-            };
+            });
 
             let source_row = match source {
                 Min(row) | Max(row) => row,
--- a/src/compute/src/render/reduce.rs
+++ b/src/compute/src/render/reduce.rs
@@ -2321,7 +2321,7 @@ mod monoids {
     use serde::{Deserialize, Serialize};
 
     /// A monoid containing a single-datum row.
-    #[derive(Ord, PartialOrd, Eq, PartialEq, Debug, Clone, Serialize, Deserialize, Hash)]
+    #[derive(Ord, PartialOrd, Eq, PartialEq, Debug, Serialize, Deserialize, Hash)]
     pub enum ReductionMonoid {
         Min(Row),
         Max(Row),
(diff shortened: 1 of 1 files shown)


src/compute/src/render/reduce.rs
```diff
@@ -2336,6 +2336,25 @@ mod monoids {
...
+            use ReductionMonoid::*;
+            match self {
+                Min(row) => Min(row.clone()),
+                Max(row) => Max(row.clone()),
+            }
+        }
+
+        fn clone_from(&mut self, source: &Self) {
+            use ReductionMonoid::*;
+            match (self, source) {
```
antiguru, line 2350: This needs to make sure that the receiver inherits the correct enum variant from other, too.

src/compute/src/render/reduce.rs
```diff
@@ -2336,6 +2336,37 @@ mod monoids {
...
+                    row.clone_from(source_row);
+                }
+                // FIXME(ptravers): not sure what we want to do here we can't overwrite
+                // without having owned access to the Row in dest. I don't see a way to
+                // achieve that without a clone. Is it better to panic or clone? I would lean
+                // clone for safety but I don't know how likely it is to hit the sad case
+                // in practice?
+                (dest, src) => soft_panic_or_log!(
+                    "Mismatched monoid variants in clone_from! destination: {dest:?} source: {src:?}",
+                ),
```
ptravers, line 2365: this appears to be safe right now since the only usage of this function is in `plus_equals` which asserts that destination and source must be of the same enum variant

antiguru: Ah, I thought of something simpler, like this:
```
let mut row = match std::mem::take(self) { Min(row) | Max(row) => row };
let source_row = match source { Min(row) | Max(row) => row };
row.clone_from(source_row);
match other {
    Min(_) => *self = Min(row),
    Max(_) => *self = Max(row),
}
```

I'd prefer not to have surprising behavior built into the clone implementation, especially because it might surprise us in the future.

ptravers: ah I see! will refactor. that's much better!

src/compute/src/render/reduce.rs
```diff
@@ -2336,6 +2336,35 @@ mod monoids {
...
+            match self {
+                Min(row) => Min(row.clone()),
+                Max(row) => Max(row.clone()),
+            }
+        }
+
+        fn clone_from(&mut self, source: &Self) {
+            use ReductionMonoid::*;
+
+            let mut row = match std::mem::replace(self, Max(Row::default())) {
```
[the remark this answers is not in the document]
antiguru: Yeah, there's no sensible default for this enum. What you can instead do is to move the `take` out of the match:
```diff
diff --git i/src/compute/src/render/reduce.rs w/src/compute/src/render/reduce.rs
index a57af9d257..6574458279 100644
--- i/src/compute/src/render/reduce.rs
+++ w/src/compute/src/render/reduce.rs
@@ -2348,9 +2348,9 @@ mod monoids {
         fn clone_from(&mut self, source: &Self) {
             use ReductionMonoid::*;
 
-            let mut row = match std::mem::replace(self, Max(Row::default())) {
+            let mut row = std::mem::take(match self {
                 Min(row) | Max(row) => row,
-            };
+            });
 
             let source_row = match source {
                 Min(row) | Max(row) => row,
```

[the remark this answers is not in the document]
ptravers: ah that's brilliant, done!

antiguru (approved):
Looks great; thank you!

src/compute/src/render/reduce.rs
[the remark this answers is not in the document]
ptravers: I have been wondering how much better this will be since we will still allocate a `Row::default()` for the lifetime of the function.

PR #20834, opened 2023-07-27
use grpc http2 keepalives for storage + compute rpcs

pH14:
In https://github.com/MaterializeInc/materialize/issues/20299 we saw that `envd`'s grpc connections for storage and compute cannot detect a connection failure when there's no TCP reset. This PR adds in knobs for HTTP/2 keep-alives (and a connect timeout) to detect this scenario. @def- already wrote a test for this scenario (thank you!), which this PR now enables / passes.

The actual change is quite small, but wiring up the knobs into `StorageParameters` and `ComputeParameters` so they can be dynamically updated / accessed requires reams of new code. Unfortunately a lot of the boilerplate is to make `GrpcClientParameters` serializable over the network, even though it's only ever needed by `envd`.

### Motivation

Fixes https://github.com/MaterializeInc/materialize/issues/20299 

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered.
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] This PR includes the following [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note):
  -

PR #32330, opened 2025-04-25
Cluster transport protocol

teskje:
This PR adds the implementation of a new cluster transport protocol (CTP) that's meant to replace gRPC as the message fabric for the compute and storage protocols (and perhaps others in the future). The protocol has all the features we need from it, but the keep the scope manageable actually wiring it up in Mz is left to [a future PR](https://github.com/MaterializeInc/materialize/pull/32897).

### Motivation

Part of https://github.com/MaterializeInc/database-issues/issues/5038

### Tips for reviewer

The diff is quite large, but 50% of it consists of tests and there are plenty of code comments. Individual commits are meaningful and self-contained. If people prefer, I can split the PR into smaller ones that each only contain a subset of the commits, just lmk!

### Checklist

### DIFF
--- a/src/service/src/transport.rs
+++ b/src/service/src/transport.rs
@@ -1,0 +1,392 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! The Cluster Transport Protocol (CTP).
+//!
+//! CTP is the protocol used to transmit commands from controllers to replicas and responses from
+//! replicas to controllers. It runs on top of a reliable bidirectional connection stream, as
+//! provided by TCP or UDS, and adds message framing as well as heartbeating.
+//!
+//! CTP supports any message type that implements the serde [`Serialize`] and [`Deserialize`]
+//! traits. Messages are encoded using the [`bincode`] format and then sent over the wire with a
+//! length prefix.
+
+use std::convert::Infallible;
+use std::fmt::Debug;
+
+use anyhow::{anyhow, bail};
+use async_trait::async_trait;
+use bincode::Options;
+use futures::future;
+use mz_ore::cast::CastInto;
+use mz_ore::netio::{Listener, SocketAddr, Stream};
+use mz_ore::task::AbortOnDropHandle;
+use semver::Version;
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
+use tokio::sync::{mpsc, watch};
+use tracing::{debug, info, trace};
+
+use crate::client::{GenericClient, Partitionable, Partitioned};
+
+/// Trait for messages that can be sent over CTP.
+pub trait Message: Debug + Send + Sync + Serialize + DeserializeOwned + 'static {}
+impl<T: Debug + Send + Sync + Serialize + DeserializeOwned + 'static> Message for T {}
+
+/// A client for a CTP connection.
+#[derive(Debug)]
+pub struct Client<Out, In> {
+    conn: Connection<Out, In>,
+}
+
+impl<Out: Message, In: Message> Client<Out, In> {
+    /// Connect to the server at the given address.
+    pub async fn connect(address: &str, version: Version) -> anyhow::Result<Self> {
+        let stream = Stream::connect(address).await?;
+        info!(%address, "ctp: connected to server");
+
+        let conn = Connection::start(stream, version).await?;
+        Ok(Self { conn })
+    }
+}
+
+impl<Out, In> Client<Out, In>
+where
+    Out: Message,
+    In: Message,
+    (Out, In): Partitionable<Out, In>,
+{
+    /// Create a `Partitioned` client that connects through CTP.
+    pub async fn connect_partitioned(
+        addresses: Vec<String>,
+        version: Version,
+    ) -> anyhow::Result<Partitioned<Self, Out, In>> {
+        let connects = addresses
+            .iter()
+            .map(|addr| Self::connect(addr, version.clone()));
+        let clients = future::try_join_all(connects).await?;
+        Ok(Partitioned::new(clients))
+    }
+}
+
+#[async_trait]
+impl<Out: Message, In: Message> GenericClient<Out, In> for Client<Out, In> {
+    async fn send(&mut self, cmd: Out) -> anyhow::Result<()> {
+        self.conn.send(cmd).await
+    }
+
+    /// # Cancel safety
+    ///
+    /// This method is cancel safe.
+    async fn recv(&mut self) -> anyhow::Result<Option<In>> {
+        // `Connection::recv` is documented to be cancel safe.
+        self.conn.recv().await.map(Some)
+    }
+}
+
+/// Spawn a CTP server that serves connections at the given address.
+pub async fn serve<In, Out, H>(
+    address: SocketAddr,
+    version: Version,
+    handler_fn: impl Fn() -> H,
+) -> anyhow::Result<()>
+where
+    In: Message,
+    Out: Message,
+    H: GenericClient<In, Out> + 'static,
+{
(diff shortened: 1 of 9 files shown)


src/service/src/transport.rs
```diff
@@ -0,0 +1,553 @@
...
+    /// Receiver for errors encountered by connection tasks.
+    error_rx: watch::Receiver<String>,
+
+    /// Handles to connection tasks.
+    _tasks: [AbortOnDropHandle<()>; 2],
+}
+
+impl<Out: Message, In: Message> Connection<Out, In> {
+    /// The interval with which keepalives are emitted on idle connections.
+    const KEEPALIVE_INTERVAL: Duration = Duration::from_secs(1);
```
teskje, line 258: This is configurable for the grpc protocol, but afaik we never felt the need to change the interval, so it seems fine to hardcode here.

src/service/src/transport.rs
```diff
@@ -0,0 +1,553 @@
...
+    let mut len = 0;
+    while len == 0 {
+        len = reader.read_u64().await?;
+    }
+
+    let mut bytes = vec![0; len.cast_into()];
+    reader.read_exact(&mut bytes).await?;
+
+    wire_decode(&bytes)
+}
```
teskje: As a follow-up we should look into making these streaming. Streaming read should be relatively easy. Streaming write would require us to move from a length-prefix encoding to a chunked encoding, as we won't know the encoded message length beforehand.

teskje: On the other hand, maybe we should rather spend our time making sure no huge messages are transmitted over this protocol. We are mostly there, modulo subscribe responses.

src/service/src/transport.rs
```diff
@@ -0,0 +1,548 @@
...
+
+/// Encode a message for wire transport.
+fn wire_encode<M: Message>(msg: &M) -> anyhow::Result<Vec<u8>> {
+    let bytes = bincode::DefaultOptions::new().serialize(msg)?;
```
teskje, line 541: In an earlier version I was also applying compression here. But that caused regressions in the feature benchmarks for large peek responses. Turns out that grpc is fast because we do not switch on compression there. I decided to leave out compression for the first cut of CTP as well, to not rock the boat too much. We can consider adding it later.

teskje:
The [PR](https://github.com/MaterializeInc/materialize/pull/32897) that uses this for the cluster protocols is passing CI now, so I think this is ready to review.

src/service/src/transport.rs
```diff
@@ -0,0 +1,548 @@
...
+async fn handshake<R, W>(
+    mut reader: R,
+    mut writer: W,
+    version: Version,
+    server_fqdn: Option<String>,
+) -> anyhow::Result<()>
+where
+    R: AsyncRead + Unpin,
+    W: AsyncWrite + Unpin,
+{
```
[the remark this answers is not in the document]
teskje: Idk what you are talking about, it's just a randomly chosen number.

src/service/src/transport.rs
```diff
@@ -0,0 +1,548 @@
...
+            inbound = conn.recv() => {
+                let msg = inbound?;
+                handler.send(msg).await?;
+            },
+            // `GenericClient::recv` is documented to be cancel safe.
+            outbound = handler.recv() => match outbound? {
+                Some(msg) => conn.send(msg).await?,
+                None => bail!("client disconnected"),
+            },
+            _ = cancel_rx.as_mut().unwrap() => bail!("connection canceled"),
```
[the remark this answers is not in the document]
teskje: Removed the option!

src/service/src/transport.rs
```diff
@@ -0,0 +1,392 @@
...
+        Ok(Self {
+            msg_tx: out_tx,
+            msg_rx: in_rx,
+            error_rx,
+            _tasks: [send_task.abort_on_drop(), recv_task.abort_on_drop()],
+        })
+    }
+
+    /// Enqueue a message for sending.
+    async fn send(&mut self, msg: Out) -> anyhow::Result<()> {
```
antiguru, line 316: What about cancellation safety on the send path? The channel is bounded I think, so depending on how we call this function, we could block here.

src/service/src/transport.rs
```diff
@@ -49,14 +49,31 @@ pub struct Client<Out, In> {
...
+        let conn = Connection::start(stream, version, dest_host).await?;
         Ok(Self { conn })
     }
 }
 
+/// Helper function to extract the host part from an address string.
+///
+/// This function assumes addresses to be of the form `<host>:<port>` or `<protocol>:<host>:<port>`
+/// and yields `None` otherwise.
+fn host_from_address(address: &str) -> Option<String> {
```
antiguru, line 82: Would it make sense to have some form of URL abstraction so this logic wouldn't need to live here? Not urgent, but maybe worth recording.

src/service/src/transport.rs
antiguru: Why change it to two seconds? Just curios.

src/service/src/transport.rs
```diff
@@ -0,0 +1,549 @@
...
+    M: Message,
+{
+    let bytes = match msg {
+        Some(msg) => wire_encode(msg)?,
+        None => Vec::new(),
+    };
+
+    let len = bytes.len().cast_into();
+    writer.write_u64(len).await?;
+    writer.write_all(&bytes).await?;
```
antiguru: Maybe:
```suggestion
    let bytes = match msg {
        Some(msg) => &wire_encode(msg)?,
        None => &[],
    };

    let len = bytes.len().cast_into();
    writer.write_u64(len).await?;
    writer.write_all(bytes).await?;
```

antiguru (approved):
I think this looks fine. Some questions around cancellation safety of the send path, but maybe not relevant since we never call it from a select.

src/service/src/transport.rs
teskje: Yep, `Connection::send` is not cancel safe. But it also doesn't pretend to be and nobody relies on it being cancel safe.

src/service/src/transport.rs
teskje: Yeah, I also don't love this part. We have a similar thing in `create_sockets`. The issue is that our address format here isn't a URL format, and so parsing it with a crate like `url` doesn't work. The best I can think of is putting some parsing function into `mz_ore::netio`.

src/service/src/transport.rs
teskje: Sorry, are you asking why `MIN_TIMEOUT` is two seconds? It's because we need something that's greater than `KEEPALIVE_INTERVAL` (otherwise sending keepalives wouldn't avoid running into the timeout). And it should be greater by some margin so that the keepalive works reliably even in the face of packet drops and retransmissions. Twice the keepalive interval feels like a good value, but it might not be!

The `MIN_TIMEOUT` doesn't matter much, it's only supposed to protect against nonsensical configurations. The real timeout will be configured via dyncfg. In prod we are currently using 60s, which is maybe a bit high.

teskje: Oh, if you are instead asking why the keepalive interval here is one second while the grpc default is currently 3 seconds, the simple answer is that I didn't check the grpc default and simply selected 1s as an interval that feels reasonable. The [PR](https://github.com/MaterializeInc/materialize/pull/20834) that introduced the 3 seconds doesn't have much motivation about that value. It probably doesn't matter much either way.

teskje:
TFTR!

bosconi:
bugbot run

src/service/src/transport.rs
```diff
@@ -0,0 +1,549 @@
...
+    M: Message,
+{
+    // Skip over any empty messages (i.e. keepalives).
+    let mut len = 0;
+    while len == 0 {
+        len = reader.read_u64().await?;
+    }
+
+    let mut bytes = vec![0; len.cast_into()];
+    reader.read_exact(&mut bytes).await?;
```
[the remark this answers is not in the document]
teskje: Yes, this is a calculated risk. In exchange for keeping things simple we rely on the peer behaving according to the protocol and not sending malicious messages. Seems to be a reasonable assumption to make for the cluster protocol where we control both sides.

It would be nice to fix this, but just imposing a size limit won't do because responses can become arbitrarily large. A proper fix would require splitting messages into fixed-sized frames and re-assembling them at the receiver.

PR #32331, opened 2025-04-25
Strip AD/BC only with a leading space

frankmcsherry:
Timezones like `Asia/Baghdad` were misparsed as ending with `AD`, which was stripped resulting in `Asia/Baghd` (https://github.com/MaterializeInc/database-issues/issues/9236). This PR changes this to match only on ` AD` and ` BC`, with leading spaces.

No idea if this is within spec, but what we were doing seemed equally wrong.

For reference: postgres's docs about how they parse things: https://www.postgresql.org/docs/12/datetime-input-rules.html. We skip actual tokenization, and do a bit of ad-hoc string matching. Tokenization (and access to their "internal tables") would have treated AD/BC correctly (they would not have been tokens extracted from `Asia/Baghdad`).

### Checklist

### DIFF
--- a/src/repr/src/adt/datetime.rs
+++ b/src/repr/src/adt/datetime.rs
@@ -1874,15 +1874,6 @@ pub enum CalendarEra {
     AD,
 }
 
-impl CalendarEra {
-    fn as_str(&self) -> &'static str {
-        match self {
-            CalendarEra::BC => "BC",
-            CalendarEra::AD => "AD",
-        }
-    }
-}
-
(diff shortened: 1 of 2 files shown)


frankmcsherry:
As the tests reveal, there is more work to do. Postgres accepts `AD` and `BC` in other positions, for example without a space:
```
select '2024-01-01 00:00:00BC'::timestamptz;
```
They do a bunch of other crazy things too, for example courtesy @petrosagg:
```
postgres=# select 'allballs today bc'::timestamp;
       timestamp        
------------------------
 2025-04-25 00:00:00 BC
(1 row)
```

PR #32334, opened 2025-04-25
Mention the supported Kafka version in the docs

bkirwi:
Running older versions than this may work, but we don't test for it, and we may have to break compat in the future. Write a policy down!

### Motivation

In recent Kafka sink work, we lost some time to figuring out whether we needed to maintain support for very old Kafka versions. Having an explicit policy here should help us move faster and give users more confidence that their deployment is expected to work.

### Tips for reviewer

All our hosted service guides are for services that keep users on fairly recent versions, so I've only added the explicit version requirement on the self-hosted guide.

As usual, feel free to edit the prose if it doesn't match the overall style!

### Checklist

### DIFF
--- a/doc/user/content/ingest-data/kafka/_index.md
+++ b/doc/user/content/ingest-data/kafka/_index.md
@@ -15,6 +15,10 @@ from Kafka, you need to
 1. Create a connection that specifies access and authentication parameters.
 2. Create a source that specifies the format of the data you want to ingest.
 
+## Supported versions
+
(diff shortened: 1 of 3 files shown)


bkirwi:
We include a couple versions in the "Kafka Matrix" tests that are no longer supported by Confluent / Redpanda, respectively. Should we drop them from the test matrix? They're not really a maintenance burden AFAIK, but maybe we no longer want to spend the cycles testing versions that even upstream doesn't support -- @def-, any opinion?

kay-kim (approved):
lgtm!

def-:
The testing is pretty cheap and maybe a potential customer still runs an unsupported version, so I wouldn't remove them from the matrix test (until they break because we want to change something)

bkirwi:
Okay great! Thanks all for the feedback.

PR #32335, opened 2025-04-25
compute,adapter: add peek response stash, for arbitrary sized peek responses

aljoscha:
Currently, results of SELECT queries (aka. peeks) are sent back from the
cluster to environmentd via the compute protocol and these results are
fully materialized (stored in memory) in environmentd before sending
them out to the client.

This has several implications:

- Sending large results "clogs up" the cluster/controller communication
- The amount of memory we want to give to environmentd limits the size
  of results we can return

In practice the above make it so we limit the size of results using a
`max_result_size` parameter, and some larger customers are chafing
against that.

Specifically, peek responses are sent inline, in the response stream
from `clusterd` to `environmentd`/Controller. As a
`ComputeResponse::PeekResponse`.

With this change, we can "stash" large peek responses and only send a
handle to them back via the response stream. Large peek responses are
stored using the new persist API for "free-standing" batches (batches
that are never intended for appending to a shard) and read out using the
also new `read_batches_consolidated`.

One current limitation is that we can only stash results when the SELECT
does _NOT_ have an ORDER BY. This is left as a follow up.

We introduces a bunch of dyncfgs for controlling when and how the result
stash is used:

- Enable/disable: for staged rollout.
- Peek stash threshold: above what result size we use the stash. This is
  deliberately separate from the existing `max_result_size`
  configuration and in practice I think we want a low-ish stash
  threshold and keep the current `max_result_size`.
- Batching while writing: we don't want to block the timely worker
  thread by writing out the whole result in one go. Instead we have
  configurable batching and batch size.
- Batching while reading: we read consolidated "chunks" out of the
  batches, when serving the result to the client. Here we also make the
  batch size configurable.

Implements https://github.com/MaterializeInc/database-issues/issues/9180

### DIFF
--- a/misc/python/materialize/mzcompose/__init__.py
+++ b/misc/python/materialize/mzcompose/__init__.py
@@ -72,6 +72,7 @@ def get_minimal_system_parameters(
         # -----
         # Others (ordered by name)
         "allow_real_time_recency": "true",
+        "enable_compute_peek_response_stash": "true",
         "enable_0dt_deployment": "true" if zero_downtime else "false",
         "enable_0dt_deployment_panic_after_timeout": "true",
         "enable_0dt_deployment_sources": (
@@ -165,6 +166,13 @@ def get_variable_system_parameters(
         ),
         VariableSystemParameter(
             "compute_apply_column_demands", "true", ["true", "false"]
+        ),
+        VariableSystemParameter(
+            "compute_peek_response_stash_threshold_bytes",
+            # 1 MiB, an in-between value
+            "1048576",
+            # force-enabled, the in-between, and the production value
+            ["0", "1048576", "314572800", "67108864"],
         ),
         VariableSystemParameter(
             "disk_cluster_replicas_default", "true", ["true", "false"]
@@ -510,6 +518,11 @@ UNINTERESTING_SYSTEM_PARAMETERS = [
     "mz_metrics_lgalloc_map_refresh_interval",
     "mz_metrics_lgalloc_refresh_interval",
     "mz_metrics_rusage_refresh_interval",
+    "compute_peek_response_stash_batch_max_runs",
+    "compute_peek_response_stash_read_batch_size_bytes",
+    "compute_peek_response_stash_read_memory_budget_bytes",
+    "compute_peek_stash_num_batches",
+    "compute_peek_stash_batch_size",
 ]
 
 
--- a/misc/python/materialize/mzcompose/__init__.py
+++ b/misc/python/materialize/mzcompose/__init__.py
@@ -55,6 +55,10 @@ ADDITIONAL_BENCHMARKING_SYSTEM_PARAMETERS = {
     # This would increase the memory usage of many tests, making it harder to
     # tell small memory increase regressions
     "persist_blob_cache_scale_with_threads": "false",
+    # The peek response stash kicks in when results get larger, and it
+    # increases query latency. Which in turn makes benchmarking more
+    # unpredictable.
+    "enable_compute_peek_response_stash": "false",
 }
 
 
--- a/misc/python/materialize/parallel_workload/action.py
+++ b/misc/python/materialize/parallel_workload/action.py
@@ -1083,6 +1083,14 @@ class FlipFlagsAction(Action):
         self.flags_with_values["enable_compute_active_dataflow_cancelation"] = (
             BOOLEAN_FLAG_VALUES
         )
+        self.flags_with_values["enable_compute_peek_response_stash"] = (
+            BOOLEAN_FLAG_VALUES
+        )
+        self.flags_with_values["compute_peek_response_stash_threshold_bytes"] = [
+            "0",  # "force enabled"
+            "1048576",  # 1 MiB, an in-between value
+            "314572800",  # 300 MiB, the production value
+        ]
 
         # If you are adding a new config flag in Materialize, consider using it
         # here instead of just marking it as uninteresting to silence the
@@ -1265,6 +1273,11 @@ class FlipFlagsAction(Action):
             "mz_metrics_lgalloc_map_refresh_interval",
             "mz_metrics_lgalloc_refresh_interval",
             "mz_metrics_rusage_refresh_interval",
+            "compute_peek_stash_num_batches",
+            "compute_peek_stash_batch_size",
+            "compute_peek_response_stash_batch_max_runs",
+            "compute_peek_response_stash_read_batch_size_bytes",
+            "compute_peek_response_stash_read_memory_budget_bytes",
         ]
 
     def run(self, exe: Executor) -> bool:
--- a/src/adapter/src/coord/peek.rs
+++ b/src/adapter/src/coord/peek.rs
@@ -16,9 +16,10 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::fmt;
 use std::num::NonZeroUsize;
 use std::ops::Deref;
+use std::sync::Arc;
 
 use differential_dataflow::consolidation::consolidate;
-use futures::TryFutureExt;
+use itertools::Itertools;
 use mz_adapter_types::compaction::CompactionWindow;
 use mz_adapter_types::connection::ConnectionId;
 use mz_cluster_client::ReplicaId;
@@ -32,16 +33,21 @@ use mz_expr::explain::{HumanizedExplain, HumanizerMode, fmt_text_constant_rows};
 use mz_expr::row::RowCollection;
 use mz_expr::{
     EvalError, Id, MirRelationExpr, MirScalarExpr, OptimizedMirRelationExpr, RowSetFinishing,
-    permutation_for_arrangement,
+    RowSetFinishingIncremental, permutation_for_arrangement,
 };
 use mz_ore::cast::CastFrom;
 use mz_ore::str::{StrExt, separated};
 use mz_ore::tracing::OpenTelemetryContext;
+use mz_persist_client::Schemas;
+use mz_persist_types::codec_impls::UnitSchema;
 use mz_repr::explain::text::DisplayText;
 use mz_repr::explain::{CompactScalars, IndexUsageType, PlanRenderingContext, UsedIndexes};
-use mz_repr::{Diff, GlobalId, IntoRowIterator, RelationType, Row, RowIterator, preserves_order};
+use mz_repr::{
+    Diff, GlobalId, IntoRowIterator, RelationDesc, RelationType, Row, RowIterator, preserves_order,
+};
+use mz_storage_types::sources::SourceData;
 use serde::{Deserialize, Serialize};
-use timely::progress::Timestamp;
+use timely::progress::{Antichain, Timestamp};
 use uuid::Uuid;
 
 use crate::coord::timestamp_selection::TimestampDetermination;
@@ -368,6 +374,13 @@ pub struct PlannedPeek {
     pub plan: PeekPlan,
     pub determination: TimestampDetermination<mz_repr::Timestamp>,
     pub conn_id: ConnectionId,
+    /// The result type _after_ reading out of the "source" and applying any
+    /// [MapFilterProject](mz_expr::MapFilterProject), but _before_ applying a
+    /// [RowSetFinishing].
+    ///
+    /// This is _the_ `result_type` as far as compute is concerned and futher
+    /// changes through projections happen purely in the adapter.
+    pub intermediate_result_type: RelationType,
     pub source_arity: usize,
     pub source_ids: BTreeSet<GlobalId>,
 }
@@ -622,6 +635,7 @@ impl crate::coord::Coordinator {
             plan: fast_path,
             determination,
             conn_id,
+            intermediate_result_type,
             source_arity,
             source_ids,
         } = plan;
@@ -766,6 +780,7 @@ impl crate::coord::Coordinator {
                     index_key.len() + index_thinned_arity,
                 );
                 let map_filter_project = mfp_to_safe_plan(map_filter_project)?;
+
                 (
                     (None, timestamp, map_filter_project),
                     Some(index_id),
@@ -807,6 +822,13 @@ impl crate::coord::Coordinator {
             .insert(uuid, compute_instance);
         let (literal_constraints, timestamp, map_filter_project) = peek_command;
 
+        // At this stage we don't know column names for the result because we
+        // only know the peek's result type as a bare ResultType.
+        let peek_result_column_names =
+            (0..intermediate_result_type.arity()).map(|i| format!("peek_{i}"));
+        let peek_result_desc =
+            RelationDesc::new(intermediate_result_type, peek_result_column_names);
+
         self.controller
             .compute
             .peek(
@@ -815,18 +837,72 @@ impl crate::coord::Coordinator {
                 literal_constraints,
                 uuid,
                 timestamp,
+                peek_result_desc,
                 finishing.clone(),
                 map_filter_project,
                 target_replica,
                 rows_tx,
             )
             .unwrap_or_terminate("cannot fail to peek");
+
         let duration_histogram = self.metrics.row_set_finishing_seconds();
 
-        // Prepare the receiver to return as a response.
-        let rows_rx = rows_rx.map_ok_or_else(
-            |e| PeekResponseUnary::Error(e.to_string()),
-            move |resp| match resp {
+        // If a dataflow was created, drop it once the peek command is sent.
+        if let Some(index_id) = drop_dataflow {
+            self.remove_compute_ids_from_timeline(vec![(compute_instance, index_id)]);
+            self.drop_indexes(vec![(compute_instance, index_id)]);
+        }
+
+        let persist_client = self.persist_client.clone();
+        let peek_stash_read_batch_size_bytes =
+            mz_compute_types::dyncfgs::PEEK_RESPONSE_STASH_READ_BATCH_SIZE_BYTES
+                .get(self.catalog().system_config().dyncfgs());
+        let peek_stash_read_memory_budget_bytes =
+            mz_compute_types::dyncfgs::PEEK_RESPONSE_STASH_READ_MEMORY_BUDGET_BYTES
+                .get(self.catalog().system_config().dyncfgs());
+
+        let peek_response_stream = Self::create_peek_response_stream(
+            rows_rx,
+            finishing,
+            max_result_size,
+            max_returned_query_size,
+            duration_histogram,
+            persist_client,
+            peek_stash_read_batch_size_bytes,
+            peek_stash_read_memory_budget_bytes,
+        );
+
+        Ok(crate::ExecuteResponse::SendingRowsStreaming {
+            rows: Box::pin(peek_response_stream),
+            instance_id: compute_instance,
+            strategy,
+        })
+    }
+
+    /// Creates an async stream that processes peek responses and yields rows.
+    #[mz_ore::instrument(level = "debug")]
+    fn create_peek_response_stream(
+        rows_rx: tokio::sync::oneshot::Receiver<PeekResponse>,
+        finishing: RowSetFinishing,
+        max_result_size: u64,
+        max_returned_query_size: Option<u64>,
+        duration_histogram: prometheus::Histogram,
+        mut persist_client: mz_persist_client::PersistClient,
+        peek_stash_read_batch_size_bytes: usize,
+        peek_stash_read_memory_budget_bytes: usize,
+    ) -> impl futures::Stream<Item = PeekResponseUnary> {
+        async_stream::stream!({
+            let result = rows_rx.await;
+
+            let rows = match result {
+                Ok(rows) => rows,
+                Err(e) => {
+                    yield PeekResponseUnary::Error(e.to_string());
+                    return;
+                }
+            };
+
+            match rows {
                 PeekResponse::Rows(rows) => {
                     match finishing.finish(
                         rows,
@@ -834,25 +910,182 @@ impl crate::coord::Coordinator {
                         max_returned_query_size,
                         &duration_histogram,
                     ) {
-                        Ok((rows, _size_bytes)) => PeekResponseUnary::Rows(Box::new(rows)),
-                        Err(e) => PeekResponseUnary::Error(e),
+                        Ok((rows, _size_bytes)) => yield PeekResponseUnary::Rows(Box::new(rows)),
+                        Err(e) => yield PeekResponseUnary::Error(e),
                     }
                 }
-                PeekResponse::Canceled => PeekResponseUnary::Canceled,
-                PeekResponse::Error(e) => PeekResponseUnary::Error(e),
-            },
-        );
+                PeekResponse::Stashed(response) => {
+                    let response = *response;
 
-        // If it was created, drop the dataflow once the peek command is sent.
-        if let Some(index_id) = drop_dataflow {
-            self.remove_compute_ids_from_timeline(vec![(compute_instance, index_id)]);
-            self.drop_indexes(vec![(compute_instance, index_id)]);
-        }
+                    let shard_id = response.shard_id;
 
-        Ok(crate::ExecuteResponse::SendingRows {
-            future: Box::pin(rows_rx),
-            instance_id: compute_instance,
-            strategy,
+                    let mut batches = Vec::new();
+                    for proto_batch in response.batches.into_iter() {
+                        let batch =
+                            persist_client.batch_from_transmittable_batch(&shard_id, proto_batch);
+
+                        batches.push(batch);
+                    }
+                    tracing::trace!(?batches, "stashed peek response");
+
+                    let as_of = Antichain::from_elem(mz_repr::Timestamp::default());
+                    let read_schemas: Schemas<SourceData, ()> = Schemas {
+                        id: None,
+                        key: Arc::new(response.relation_desc.clone()),
+                        val: Arc::new(UnitSchema),
+                    };
+
+                    let mut row_cursor = persist_client
+                        .read_batches_consolidated::<_, _, _, i64>(
+                            response.shard_id,
+                            as_of,
+                            read_schemas,
+                            batches,
+                            |_stats| true,
+                            peek_stash_read_memory_budget_bytes,
+                        )
+                        .await
+                        .expect("invalid usage");
+
+                    // NOTE: Using the cursor creates Futures that are not Sync,
+                    // so we can't drive them on the main Coordinator loop.
+                    // Spawning a task has the additional benefit that we get to
+                    // delete batches once we're done.
+                    //
+                    // Batch deletion is best-effort, though, and there are
+                    // multiple known ways in which they can leak, among them:
+                    //
+                    // - ProtoBatch is lost in flight
+                    // - ProtoBatch is lost because when combining PeekResponse
+                    // from workers a cancellation or error "overrides" other
+                    // results, meaning we drop them
+                    // - This task here is not run to completion before it can
+                    // delete all batches
+                    //
+                    // This is semi-ok, because persist needs a reaper of leaked
+                    // batches already, and so we piggy-back on that, even if it
+                    // might not exist as of today.
+                    let (tx, mut rx) = tokio::sync::mpsc::channel(1);
+                    mz_ore::task::spawn(|| "read_peek_batches", async move {
+                        // We always send our inline rows first. Ordering
+                        // doesn't matter because we can only be in this case
+                        // when there is no ORDER BY.
+                        //
+                        // We _could_ write these out as a Batch, and include it
+                        // in the batches we read via the Consolidator. If we
+                        // wanted to get a consistent ordering. That's not
+                        // needed for correctness! But might be nice for more
+                        // aesthetic reasons.
+                        let result = tx.send(response.inline_rows).await;
+                        if result.is_err() {
+                            tracing::error!("receiver went away");
+                        }
+
+                        let mut current_batch = Vec::new();
+                        let mut current_batch_size: usize = 0;
+
+                        'outer: while let Some(rows) = row_cursor.next().await {
+                            for ((key, _val), _ts, diff) in rows {
+                                let source_data = key.expect("decoding error");
+
+                                let row = source_data
+                                    .0
+                                    .expect("we are not sending errors on this code path");
+
+                                let diff = usize::try_from(diff)
+                                    .expect("peek responses cannot have negative diffs");
+
+                                if diff > 0 {
+                                    let diff =
+                                        NonZeroUsize::new(diff).expect("checked to be non-zero");
+                                    current_batch_size =
+                                        current_batch_size.saturating_add(row.byte_len());
+                                    current_batch.push((row, diff));
+                                }
+
+                                if current_batch_size > peek_stash_read_batch_size_bytes {
+                                    // We're re-encoding the rows as a RowCollection
+                                    // here, for which we pay in CPU time. We're in a
+                                    // slow path already, since we're returning a big
+                                    // stashed result so this is worth the convenience
+                                    // of that for now.
+                                    let result = tx
+                                        .send(RowCollection::new(
+                                            current_batch.drain(..).collect_vec(),
+                                            &[],
+                                        ))
+                                        .await;
+                                    if result.is_err() {
+                                        tracing::error!("receiver went away");
+                                        // Don't return but break so we fall out to the
+                                        // batch delete logic below.
+                                        break 'outer;
+                                    }
+
+                                    current_batch_size = 0;
+                                }
+                            }
+                        }
+
+                        if current_batch.len() > 0 {
+                            let result = tx.send(RowCollection::new(current_batch, &[])).await;
+                            if result.is_err() {
+                                tracing::error!("receiver went away");
+                            }
+                        }
+
+                        let batches = row_cursor.into_lease();
+                        tracing::trace!(?response.shard_id, "cleaning up batches of peek result");
+                        for batch in batches {
+                            batch.delete().await;
+                        }
+                    });
+
+                    assert!(
+                        finishing.is_streamable(response.relation_desc.arity()),
+                        "can only get stashed responses when the finishing is streamable"
+                    );
+
+                    tracing::trace!("query result is streamable!");
+
+                    assert!(finishing.is_streamable(response.relation_desc.arity()));
+                    let mut incremental_finishing = RowSetFinishingIncremental::new(
+                        finishing.offset,
+                        finishing.limit,
+                        finishing.project,
+                        max_returned_query_size,
+                    );
+
+                    let mut got_zero_rows = true;
+                    while let Some(rows) = rx.recv().await {
+                        got_zero_rows = false;
+
+                        let result_rows = incremental_finishing.finish_incremental(
+                            rows,
+                            max_result_size,
+                            &duration_histogram,
+                        );
+
+                        match result_rows {
+                            Ok(result_rows) => yield PeekResponseUnary::Rows(Box::new(result_rows)),
+                            Err(e) => yield PeekResponseUnary::Error(e),
+                        }
+                    }
+
+                    // Even when there's zero rows, clients still expect an
+                    // empty PeekResponse.
+                    if got_zero_rows {
+                        let row_iter = vec![].into_row_iter();
+                        yield PeekResponseUnary::Rows(Box::new(row_iter));
+                    }
+                }
+                PeekResponse::Canceled => {
+                    yield PeekResponseUnary::Canceled;
+                }
+                PeekResponse::Error(e) => {
+                    yield PeekResponseUnary::Error(e);
+                }
+            }
         })
     }
 
--- a/src/adapter/src/coord/sequencer/inner.rs
+++ b/src/adapter/src/coord/sequencer/inner.rs
@@ -17,7 +17,7 @@ use std::time::{Duration, Instant};
 use anyhow::anyhow;
 use futures::future::{BoxFuture, FutureExt};
 use futures::stream::FuturesOrdered;
-use futures::{Future, future};
+use futures::{Future, StreamExt, future};
 use itertools::Itertools;
 use maplit::btreeset;
 use mz_adapter_types::compaction::CompactionWindow;
@@ -3021,6 +3021,7 @@ impl Coordinator {
             session,
             Default::default(),
         );
+
         self.sequence_peek(
             peek_ctx,
             plan::SelectPlan {
@@ -3039,6 +3040,7 @@ impl Coordinator {
         let strict_serializable_reads_tx = self.strict_serializable_reads_tx.clone();
         let catalog = self.owned_catalog();
         let max_result_size = self.catalog().system_config().max_result_size();
+
         task::spawn(|| format!("sequence_read_then_write:{id}"), async move {
             let (peek_response, session) = match peek_rx.await {
                 Ok(Response {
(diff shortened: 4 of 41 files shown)


def-:
Rebased because this hit an issue with Dockerhub cache: https://materializeinc.slack.com/archives/CM7ATT65S/p1749039850832839

aljoscha:
This one is now ready for review. It would be good if @teskje and/or @antiguru could review the compute-side changes. There's no one for the adapter side but I think it's good if @DAlperin ramps up on these parts, yes?

@def- Right now this is hard-coded to always use the result stash for eligible queries, so that we use it for all of CI and nightly. But I also have specific ci-settings in place. Is there anything else you want to test? This is pretty foundational, in that all tests that use SELECT would fail if this produced incorrect results, but still ... 🤷‍♂️

aljoscha:
@def- looks like this now has some other commits that appear to not be on `main`? Or github is messing up ... 🤷‍♂️

def-:
> @def- looks like this now has some other commits that appear to not be on `main`? Or github is messing up ... 🤷‍♂️

Weird, all I did was `git rebase upstream/main`. Just did it again and looks fine now.

src/compute-client/src/controller/replica.rs
```diff
@@ -297,6 +300,8 @@ where
                 if ENABLE_COMPUTE_REPLICA_EXPIRATION.get(&self.dyncfg) {
                     config.expiration_offset = self.config.expiration_offset;
                 }
+                config.peek_stash_persist_location =
+                    self.config.peek_stash_persist_location.clone();
```
teskje: This value is the same for all replicas of an instance, right? If so, I wouldn't set it in `specialize_command` but in `Instance::run`, where the `CreateInstance` command is originally created.

src/compute-client/src/protocol/command.rs
```diff
@@ -183,7 +186,7 @@ pub enum ComputeCommand<T = mz_repr::Timestamp> {
     Schedule(GlobalId),
 
     /// `AllowCompaction` informs the replica about the relaxation of external read capabilities on
-    /// a compute collection exported by one of the replica’s dataflow.
+    /// a compute collection exported by one of the replica's dataflow.
```
teskje:
```suggestion
    /// a compute collection exported by one of the replica's dataflows.
```

src/compute-client/src/protocol/command.rs
```diff
@@ -381,12 +384,26 @@ impl Arbitrary for ComputeCommand<mz_repr::Timestamp> {
...
+impl Default for InstanceConfig {
+    fn default() -> Self {
+        Self {
+            logging: Default::default(),
+            expiration_offset: Default::default(),
+            // This is only used as a placeholder and will be filled in with the
+            // real location when the command is being sent out to replicas.
+            peek_stash_persist_location: PersistLocation::new_in_mem(),
+        }
+    }
```
teskje: This `Default` impl we can also remove when you make the change to set `peek_stash_persist_location` in `Instance::run`.

src/compute-client/src/controller.rs
```diff
@@ -888,6 +913,7 @@ where
         literal_constraints: Option<Vec<Row>>,
         uuid: Uuid,
         timestamp: T,
+        intermediate_result_type: RelationType,
```
teskje: Can you add a doc comment explaining what `intermediate_result_type` means? Particularly the "intermediate" part isn't obvious.

src/compute-client/src/controller/instance.rs
```diff
@@ -1716,6 +1725,7 @@ where
             // tree to forward it on to the compute worker.
             otel_ctx,
             target: peek_target,
+            result_desc: intermediate_result_desc,
```
teskje: Any reason to not call the `Peek` field `intermediate_result_desc` as well?

src/compute-client/src/protocol/command.rs
```diff
@@ -585,6 +619,10 @@ impl PeekTarget {
 pub struct Peek<T = mz_repr::Timestamp> {
     /// Target-specific metadata.
     pub target: PeekTarget,
+    /// The relation description for the rows returned by this peek, before
+    /// applying the [RowSetFinishing] but _after_ applying the given
+    /// `map_filter_project`.
+    pub result_desc: RelationDesc,
```
teskje, line 613: I was wondering how `RowSetFinishing` can change the relation type. It is because it includes a projection, so it can remove columns, right? Do you know why we need that and can't do all the projections in the MFP instead?

aljoscha: I don't know all the cases where this is needed, but I think one of them is `INSERT INTO ... SELECT ...` with a `RETURNING` clause. Where we do the peek on workers, then the insert in envd, and we can then still project out fields before returning rows.

src/compute-client/src/controller/replica.rs
aljoscha: will do!

src/compute-client/src/controller.rs
aljoscha: Regarding this and renaming the `Peek` field to `intermediate_result_desc`. It's called "intermediate" because for me, with peeks there are three different `ResultType`/`ResultDesc`.

1. Before applying MFP: this is not visible in the code but I'd call it `source_result_desc`, where source is a very problematic name. But in the code peeks are sometimes describes as reading from a "source".
2. After applying the MFP: this is the `intermediate_result_desc`, and it's what the cluster sends back and what compute understands to be the result type of the peek.
3. After applying the `RowSetFinishing`. This is again not very visible in the code, really only the adapter frontend. But it's the `select_result_desc` (though you won't find that name in the code).

With the above, I'll rename this to just `result_desc` in the compute/controller code, because it is as far as compute is concerned. And I'll add a comment in the adapter where we derive this. Okay? And then I'd not add a docstring here.

src/compute-client/src/controller/instance.rs
aljoscha: See https://github.com/MaterializeInc/materialize/pull/32335#discussion_r2131888493

aljoscha:
@teskje I pushed commits that should address your comments. Thanks so far! 🙇‍♂️

src/compute-client/src/protocol/response.rs
```diff
@@ -662,7 +750,7 @@ mod tests {
     /// Test to ensure the size of the `ComputeResponse` enum doesn't regress.
     #[mz_ore::test]
     fn test_compute_response_size() {
-        assert_eq!(std::mem::size_of::<ComputeResponse>(), 120);
```
antiguru: Do you mind boxing the part that causes the size of `ComputeResponse` to increase? The test is to ensure its size doesn't go up.

src/compute/src/compute_state.rs
```diff
@@ -616,15 +632,19 @@ impl<'a, A: Allocate + 'static> ActiveComputeState<'a, A> {
...
 
     fn handle_cancel_peek(&mut self, uuid: Uuid) {
         if let Some(peek) = self.compute_state.pending_peeks.remove(&uuid) {
-            self.send_peek_response(peek, PeekResponse::Canceled);
+            self.send_peek_response(peek.peek(), PeekResponse::Canceled);
+        }
+        if let Some(stash_task) = self.compute_state.pending_peek_responses.remove(&uuid) {
+            // Explicitly drop the task to abort it.
+            drop(stash_task);
         }
```
antiguru: Isn't this the same as the following?
```suggestion
        let _ = self.compute_state.pending_peek_responses.remove(&uuid);
```

src/expr/src/relation.rs
```diff
@@ -3760,6 +3769,125 @@ impl RowSetFinishing {
...
+    /// # Panics
+    ///
+    /// Panics if the result is not streamable, that is it has an ORDER BY.
+    pub fn new(
+        finishing: RowSetFinishing,
+        arity: usize,
+        max_returned_query_size: Option<u64>,
+    ) -> Self {
+        assert!(finishing.is_streamable(arity));
+
```
antiguru: I think this could take `limit` and `offset` as-is. Taking a `RowSetFinishing` promises an API this cannot handle, specifically, it can only handle `RowSetFinishing`'s `limit` and `offset`, and a default projection.

src/adapter/src/coord/peek.rs
```diff
@@ -715,44 +737,245 @@ impl crate::coord::Coordinator {
...
+                                    // here, for which we pay in CPU time. We're in a
+                                    // slow path already, since we're returning a big
+                                    // stashed result so this is worth the convenience
+                                    // of that for now.
+                                    let result = tx
+                                        .send(RowCollection::new(
+                                            current_batch.drain(..).collect_vec(),
+                                            &[],
+                                        ))
+                                        .await;
```
antiguru, line 1017: Is there a benefit of forming a `RowCollection` here, other than that the channel just needs to transport a single type? Forming the collection allocates a vector, and it doesn't look like we ever merge multiple collections.

aljoscha:
> I was a bit surprised that we don't handle selects with an ordering -- this might be surprising to users, too.

True, but this was a limitation we accepted early on because what we have here is enough to unblock current users.

IMO, we might never want to support ORDER BY because these giant result sets are a bit of a niche use case.

src/compute-client/src/protocol/response.rs
aljoscha: will do!

src/compute/src/compute_state.rs
aljoscha: It is, but I wanted to be explicit about the dropping. I can also turn it into yours and add a comment that this drops/aborts the task

src/adapter/src/coord/peek.rs
aljoscha: It's because RowSetFinishingIncremental uses the same interface and internal types as `RowSetFinishing`, and it wants a `RowCollection`. I'm not too worried about the allocations because we're already in a slow path here and the bottleneck should be reading/consolidating out of persist.

You could say that we don't really need all of `RowSetFinishingIncremental`, but I wanted to keep this as similar to the other one as possible, and I have follow-up work in a branch that adds support for ORDER BY, which leans more heavily into needing this.

src/expr/src/relation.rs
aljoscha: will change!

aljoscha:
@antiguru & @teskje I pushed commits that should address your comments

antiguru (approved):

src/compute-client/src/protocol/command.rs
teskje: Ah, seems a bit strange that the `RETURNING` projection needs to be part of the peek description, rather than some post-processing that occurs after the peek. But I'm sure there is context I'm missing and it doesn't matter for this PR anyway.

src/compute-client/src/controller.rs
teskje: Sounds good!

src/compute-client/src/controller/instance.rs
```diff
@@ -1085,9 +1096,8 @@ where
             epoch: ClusterStartupEpoch::new(self.envd_epoch, 0),
         });
 
-        // Send a placeholder instance configuration for the replica task to fill in.
-        let dummy_instance_config = Default::default();
-        self.send(ComputeCommand::CreateInstance(dummy_instance_config));
+        let instance_config = InstanceConfig::new(self.peek_stash_persist_location.clone());
```
teskje: Nit but I would drop the `new` method and use the struct constructor syntax here, to make it clear that we are filling in dummy values for the other fields. Then also keep the comment saying that the replica task will fill them in later. This logic is not quite obvious, so I think the comment is helpful.

Something like:

```rust
let instance_config = InstanceConfig {
    peek_stash_persist_location: self.peek_stash_persist_location.clone(),
    // The remaining fields are replica-specific and will be set in `ReplicaTask::specialize_command`.
    logging: Default::default(),
    expiration_offset: Default::default(),
};
```

src/compute-client/src/metrics.rs
```diff
@@ -652,7 +652,7 @@ impl<M> PeekMetrics<M> {
         use PeekResponse::*;
 
         match response {
-            Rows(_) => &self.rows,
+            Rows(_) | Stashed(_) => &self.rows,
```
teskje: Mind adding a `PeekMetrics::stashed` variant instead? It might be useful to differentiate the two in metrics.

src/compute-client/src/protocol/response.rs
```diff
@@ -374,6 +380,88 @@ impl Arbitrary for PeekResponse {
...
+    pub shard_id: ShardId,
+    /// Batches of Rows, must be combined with reponses from other workers and
+    /// consolidated before sending back via a client.
+    pub batches: Vec<ProtoBatch>,
+    /// Rows that have not been uploaded to the stash, because their total size
+    /// did not go above the threshold for using the peek stash.
+    ///
+    /// We will have a mix of stashed responses and inline responses because the
+    /// result sizes across different workers can and will vary.
+    pub inline_rows: RowCollection,
```
teskje, line 405: Did you consider replacing `PeekResponse::Rows` entirely with this new response? It seems redundant to have both when one is just a specialization of the other (i.e. all rows are in `inline_rows` and none in `batches`).

src/compute-client/src/protocol/response.rs
```diff
@@ -374,6 +380,88 @@ impl Arbitrary for PeekResponse {
...
+}
+
+impl StashedPeekResponse {
+    /// Total count of [`Row`]s represented by this collection, considering a
+    /// possible `OFFSET` and `LIMIT`.
+    pub fn num_rows(&self, offset: usize, limit: Option<usize>) -> usize {
+        let num_stashed_rows: usize = usize::cast_from(self.num_rows);
+        let num_inline_rows = self.inline_rows.count(offset, limit);
```
teskje: Ah, when reading the docs of `StashedPeekResponse::num_rows` I assumed that the count includes both the rows in `batches` and in `inline_rows`. Can we either make `num_rows` include both or clearly document that it only includes rows in `batches`?

src/compute-client/src/service.rs
```diff
@@ -334,6 +354,61 @@ where
...
+                                    num_rows: other_num_rows,
+                                    encoded_size_bytes: other_encoded_size_bytes,
+                                    relation_desc: other_relation_desc,
+                                    shard_id: other_shard_id,
+                                    batches: mut other_batches,
+                                    inline_rows: other_inline_rows,
+                                } = *other;
+
+                                assert_eq!(self_shard_id, other_shard_id);
+                                assert_eq!(self_relation_desc, other_relation_desc);
```
teskje: Can we make these soft-asserts and return an "internal error" peek response if they are hit? If we have a bug, we can avoid crashing envd.

src/compute-types/src/dyncfgs.rs
```diff
@@ -236,6 +236,47 @@ pub const ENABLE_ACTIVE_DATAFLOW_CANCELATION: Config<bool> = Config::new(
     "Whether to use `drop_dataflow` to actively cancel compute dataflows.",
 );
 
+/// Whether to enable the peek response stash, for sending back large peek
+/// responses. The response stash will only be used for results that exceed
+/// `max_result_size`.
+pub const ENABLE_PEEK_RESPONSE_STASH: Config<bool> = Config::new(
+    "compute_enable_peek_response_stash",
```
teskje: Nit: Normally we call the "enable"-flags `enable_compute_*`, not `compute_enable_*`.

src/compute-types/src/dyncfgs.rs
```diff
@@ -236,6 +236,47 @@ pub const ENABLE_ACTIVE_DATAFLOW_CANCELATION: Config<bool> = Config::new(
...
+    true, /* WIP: force-enabled for CI */
```
teskje: Did you plan to remove these WIPs before merging? I think enabling in CI is good, but we should do that through the CI flags config, not the default.

src/compute-types/src/dyncfgs.rs
```diff
@@ -236,6 +236,47 @@ pub const ENABLE_ACTIVE_DATAFLOW_CANCELATION: Config<bool> = Config::new(
     "Whether to use `drop_dataflow` to actively cancel compute dataflows.",
 );
 
+/// Whether to enable the peek response stash, for sending back large peek
+/// responses. The response stash will only be used for results that exceed
+/// `max_result_size`.
```
```suggestion
/// `compute_peek_response_stash_threshold_bytes `.
```

src/compute-types/src/dyncfgs.rs
```diff
@@ -236,6 +236,47 @@ pub const ENABLE_ACTIVE_DATAFLOW_CANCELATION: Config<bool> = Config::new(
...
+pub const ENABLE_PEEK_RESPONSE_STASH: Config<bool> = Config::new(
+    "compute_enable_peek_response_stash",
+    true, /* WIP: force-enabled for CI */
+    "Whether to enable the peek response stash, for sending back large peek responses. Will only be used for results that exceed compute_peek_response_stash_threshold.",
```
```suggestion
    "Whether to enable the peek response stash, for sending back large peek responses. Will only be used for results that exceed compute_peek_response_stash_threshold_bytes.",
```

src/compute-types/src/dyncfgs.rs
```diff
@@ -236,6 +236,47 @@ pub const ENABLE_ACTIVE_DATAFLOW_CANCELATION: Config<bool> = Config::new(
...
+pub const PEEK_RESPONSE_STASH_THRESHOLD_BYTES: Config<usize> = Config::new(
+    "compute_peek_response_stash_threshold_bytes",
+    // 1024 * 1024 * 300, /* 300mb */
+    0, /* WIP: force-enabled, for CI */
+    "The threshold above which to use the peek response stash, for sending back large peek responses.",
+);
+
+/// The target size for batches of rows we read out of the peek stash.
+pub const PEEK_RESPONSE_STASH_READ_BATCH_SIZE_BYTES: Config<usize> = Config::new(
+    "adapter_peek_response_stash_read_batch_size_bytes",
```
teskje: Based on the flag name this is an adapter config. Is `mz-compute-types` the right place for it?

src/compute/src/compute_state.rs
```diff
@@ -1059,21 +1134,20 @@ pub enum PendingPeek {
...
+    let (id, peek_type) = match &peek.target {
+        PeekTarget::Index { id } => (id, logging::compute::PeekType::Index),
+        PeekTarget::Persist { id, .. } => (id, logging::compute::PeekType::Persist),
+    };
+    ComputeEvent::Peek(PeekEvent {
+        peek: logging::compute::Peek::new(*id, peek.timestamp, peek.uuid),
+        peek_type,
+        installed,
+    })
+}
```
teskje: I believe this change wouldn't be necessary if the stashed peek responses were part of the `PendingPeek`s (see ~~above~~ [below](https://github.com/MaterializeInc/materialize/pull/32335#discussion_r2139826143))?

src/compute/src/compute_state.rs
```diff
@@ -887,7 +905,43 @@ impl<'a, A: Allocate + 'static> ActiveComputeState<'a, A> {
...
+                    ControlFlow::Break(PeekStatus::NotReady) => None,
+                    ControlFlow::Break(PeekStatus::UsePeekStash) => {
+                        let _span =
+                            span!(parent: &peek.span, Level::DEBUG, "process_stash_peek").entered();
+                        let stash_task = peek_stash::StashPeekResponse::start_upload(
+                            Arc::clone(&self.compute_state.persist_clients),
+                            self.compute_state
+                                .peek_stash_persist_location
+                                .as_ref()
+                                .expect("missing persist location for peek responses"),
```
teskje: Can we check for a `peek_stash_persist_location` above together with `peek_stash_{eligible,enabled}`, instead of unwrapping here? It's true that with the current code we shouldn't ever be able to hit this expect, but that relies on non-local reasoning that can break if other code changes.

src/compute/src/compute_state.rs
```diff
@@ -887,7 +905,43 @@ impl<'a, A: Allocate + 'static> ActiveComputeState<'a, A> {
...
+
+                match peek.seek_fulfillment(
+                    upper,
+                    self.compute_state.max_result_size,
+                    peek_stash_enabled && peek_stash_eligible,
+                    peek_stash_threshold_bytes,
+                ) {
+                    ControlFlow::Continue(result) => Some(result),
+                    ControlFlow::Break(PeekStatus::NotReady) => None,
+                    ControlFlow::Break(PeekStatus::UsePeekStash) => {
```
teskje: Not sure the `ControlFlow` semantics quite fit here. From the docs:

> Used to tell an operation whether it should exit early or go on as usual.

But nothing exits early here... or maybe you could say the `Continue` and the first `Break` case exit early while the second `Break` case goes on to perform a bunch of more work. I think `Result` fits better? Or you could add another variant to `PeekStatus`.

src/compute/src/compute_state.rs
```diff
@@ -1396,27 +1478,34 @@ impl IndexPeek {
                     target = %self.peek.target.id(), diff = %copies, %error,
                     "index peek encountered negative multiplicities in error trace",
                 );
-                return Err(format!(
-                    "Invalid data in source errors, \
-                     saw retractions ({}) for row that does not exist: {}",
+                return ControlFlow::Continue(Err(format!(
+                    "Invalid data in source errors, saw retractions ({}) for row that does not exist: {}",
```
teskje: This line is now longer than 100 characters, which sometimes breaks rustfmt. Can you change it back?

src/compute/src/compute_state.rs
```diff
@@ -106,6 +112,11 @@ pub struct ComputeState {
     pub copy_to_response_buffer: Rc<RefCell<Vec<(GlobalId, CopyToResponse)>>>,
     /// Peek commands that are awaiting fulfillment.
     pub pending_peeks: BTreeMap<Uuid, PendingPeek>,
+    /// Peek responses that we are stashing in persist. We do this
+    /// asynchronously and send back a handle to the batch in the PeekResponse.
+    pub pending_peek_responses: BTreeMap<Uuid, peek_stash::StashPeekResponse>,
```
teskje: Have you considered putting this into `pending_peeks`, i.e. having an `Option<StashPeekResponse>` in `PendingPeek`? I think that would simplify some of the code handling peeks, but more importantly it prevents you from accidentally dropping an entry from `pending_peeks` but forgetting to drop it from `pending_peek_responses`. Case in point: I think there is currently a bug in compute reconciliation where exactly this happens.

src/compute/src/compute_state.rs
```diff
@@ -1348,14 +1422,16 @@ impl IndexPeek {
         &mut self,
         upper: &mut Antichain<Timestamp>,
         max_result_size: u64,
-    ) -> Option<PeekResponse> {
+        peek_stash_eligible: bool,
+        peek_stash_threshold_bytes: usize,
```
teskje, line 1437: Nit: You could consider passing a single `peek_stash_threshold_bytes: Option<usize>` instead of two fields. Or even just a single `peek_stash_threshold_bytes: usize` and set it to `usize::MAX` if stashing is not allowed.

src/compute/src/compute_state/peek_stash.rs
```diff
@@ -0,0 +1,245 @@
...
+use tokio::sync::oneshot;
+use tracing::debug;
+use uuid::Uuid;
+
+use crate::arrangement::manager::{PaddedTrace, TraceBundle};
+use crate::compute_state::peek_result_iterator;
+use crate::compute_state::peek_result_iterator::PeekResultIterator;
+use crate::typedefs::RowRowAgent;
+
+/// An async task that stashes a peek response in persist and yield a handle to
```
teskje:
```suggestion
/// An async task that stashes a peek response in persist and yields a handle to
```

src/compute/src/compute_state/peek_stash.rs
```diff
@@ -0,0 +1,245 @@
...
+    /// results from this to the `rows_tx` channel.
+    peek_iterator: Option<PeekResultIterator<PaddedTrace<RowRowAgent<Timestamp, Diff>>>>,
+    /// We can't give a PeekResultIterator to our async upload task because the
+    /// underlying trace reader is not Send/Sync. So we need to use a channel to
+    /// send result rows from the worker thread to the async background task.
+    rows_tx: Option<tokio::sync::mpsc::Sender<Result<Vec<(Row, NonZeroI64)>, String>>>,
+    /// When the async task is busy it cannot work of rows, and we don't want to
+    /// use an unbounded channel. We stash rows that we cannot send and retry
+    /// later.
+    stashed_rows: Option<Result<Vec<(Row, NonZeroI64)>, String>>,
```
teskje: We need this only because we can't put a batch of rows back into the `peek_iterator`, right? Could we just check for available space inside `rows_tx` before polling the `peek_iterator` and avoid the need for a stash? Looks like [try_reserve](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.try_reserve) lets you do that.

src/compute/src/compute_state/peek_stash.rs
```diff
@@ -0,0 +1,245 @@
...
+
+        let peek_iterator = peek_result_iterator::PeekResultIterator::new(
+            peek.target.id().clone(),
+            peek.map_filter_project.clone(),
+            peek.timestamp,
+            peek.literal_constraints.clone(),
+            oks_handle,
+        );
+
+        let task_handle = mz_ore::task::spawn(|| "compute::stash_peek_response", async move {
```
teskje: Nit: Adding the peek ID to the task name might come in handy for debugging.

src/compute/src/compute_state/peek_stash.rs
```diff
@@ -0,0 +1,245 @@
...
+        relation_desc: RelationDesc,
+        mut rows_rx: tokio::sync::mpsc::Receiver<Result<Vec<(Row, NonZeroI64)>, String>>,
+    ) -> Result<PeekResponse, String> {
+        let client = persist_clients
+            .open(persist_location)
+            .await
+            .map_err(|e| e.to_string())?;
+
+        let shard_id = format!("s{}", peek_uuid);
+        let shard_id = ShardId::try_from(shard_id).expect("can parse");
```
teskje, line 138: Huh, are we allowed to do that? :D I guess nothing goes wrong, but `ShardId` doesn't explicitly document its string format, afaict, so it feels a bit like we are relying on implementation details.

src/compute/src/compute_state/peek_stash.rs
```diff
@@ -0,0 +1,245 @@
...
+        let mut num_rows: u64 = 0;
+
+        loop {
+            let row = rows_rx.recv().await;
+            match row {
+                Some(Ok(rows)) => {
+                    assert!(
+                        rows.len() > 0,
+                        "the contract is that we only get non-empty batches"
+                    );
```
teskje: How important is this? It seems like we could just drop this assert and remove an assumption without any drawbacks?

src/compute/src/compute_state/peek_stash.rs
```diff
@@ -0,0 +1,245 @@
...
+                    .expect("missing rows_tx")
+                    .try_send(rows)
+                {
+                    Ok(_) => {
+                        // All good!
+                    }
+                    Err(TrySendError::Full(rows)) => {
+                        // Need to stash the row and try again later
+                        let prev = self.stashed_rows.replace(rows);
+                        assert!(prev.is_none());
```
teskje: Don't we need to make sure we break out of the loop here? Otherwise we'll go through it again and hit the assert here, no?

src/storage-client/src/controller.rs
```diff
@@ -293,6 +293,9 @@ pub trait StorageController: Debug {
     /// Get the current configuration, including parameters updated with `update_parameters`.
     fn config(&self) -> &StorageConfiguration;
 
+    /// Return the [PersistLocation] used by this controller.
+    fn persist_location(&self) -> PersistLocation;
```
teskje: Is this used anywhere?

misc/python/materialize/mzcompose/__init__.py
```diff
@@ -88,6 +88,9 @@ def get_default_system_parameters(
         "compute_hydration_concurrency": "2",
         "compute_replica_expiration_offset": "3d",
         "compute_apply_column_demands": "true",
+        "compute_enable_peek_response_stash": "true",
+        # always use result stash, when eligible
+        "compute_peek_response_stash_threshold_bytes": "0",
```
teskje: This seems dangerous because it completely disables testing of the inline results path in CI. Can we set this to a value that's small but not zero? Alternatively, @def- just added a way to randomize options in CI, maybe this would be a good use of that feature. https://github.com/MaterializeInc/materialize/pull/32679

aljoscha:
Follow-up issue about letting users know why the optimization wasn't used: https://github.com/MaterializeInc/database-issues/issues/9368

aljoscha:
@def- what do we want to do about tests. I've hard-coded the large-result threshold to `0`, so that we get as much coverage as possible. But we might want to tune this to some reasonable number or only use in specific tests?

def-:
Getting more coverage for the new feature that we plan to enable in prod totally makes sense. To prevent regressions in the old code paths we should add some more possible values in `FlipFlagsAction` and `get_variable_system_parameters`.

Are you worried about the performance being bad with threshold 0? In that case add a more reasonable default in `ADDITIONAL_BENCHMARKING_SYSTEM_PARAMETERS`.

src/compute-client/src/protocol/response.rs
aljoscha: I did consider it but wanted to keep it separate for now in case there's any issue with this new response type. For example, the merging logic is more involved and we go down a very different code path in the adapter for reading out the stashed rows.

We can merge this in the future, imo. If that's okay?

src/compute-client/src/protocol/response.rs
aljoscha: renaming it to `num_rows_batches` and will also update the docstring 👌

src/compute/src/compute_state/peek_stash.rs
aljoscha: Yeah, it's not pretty but it works. An alternative would have been to mint a `ShardId` in the adapter and always send that along in the `ComputeCommand::Peek`, but I thought this is such a niche code path, and so I didn't want to burden `Peek` with always having to carry that one. I can change it if you like, though?

src/compute/src/compute_state/peek_stash.rs
aljoscha: will remove, there must have been some reason but I forget what it was

src/compute/src/compute_state/peek_stash.rs
aljoscha: hrm, I think that might be correct 🙈

src/storage-client/src/controller.rs
aljoscha: good catch! this is from an earlier version when the code had to get it from the controller

misc/python/materialize/mzcompose/__init__.py
aljoscha: Yep, sorry, this was for getting maximal coverage for testing. I've been talking to Dennis about how we want to do this for real

src/compute-client/src/protocol/response.rs
teskje: Fine with me! Can you make an issue tracking this as a follow-up?

src/compute/src/compute_state/peek_stash.rs
teskje: I thought we could just mint a `ShardId` in clusterd, but that doesn't work because all workers have to agree on the same one, right? I guess fine to leave it as is if the persist folk don't have concerns.

Maybe as part of unifying peek responses, making the stashed-response more of a first-class citizen, we could also make the change and put the `ShardId` into the peek command. But I don't see any actual reason to do that except that it feels cleaner.

src/compute/src/compute_state/peek_stash.rs
teskje: If we can get rid of the `stashed_rows` entirely, this part will also become much simpler :)

aljoscha:
> Getting more coverage for the new feature that we plan to enable in prod totally makes sense. To prevent regressions in the old code paths we should add some more possible values in `FlipFlagsAction` and `get_variable_system_parameters`.

I've pushed a change that adds these and sets the dyncfg defaults to sensible values. Is that what you had in mind?

> Are you worried about the performance being bad with threshold 0? In that case add a more reasonable default in `ADDITIONAL_BENCHMARKING_SYSTEM_PARAMETERS`.

I'm not worried, both in normal CI and nightly the benchmarks seemed fine.

def-:
> > Getting more coverage for the new feature that we plan to enable in prod totally makes sense. To prevent regressions in the old code paths we should add some more possible values in `FlipFlagsAction` and `get_variable_system_parameters`.
> 
> I've pushed a change that adds these and sets the dyncfg defaults to sensible values. Is that what you had in mind?

Yep, that looks good to me.

src/compute/src/compute_state/peek_stash.rs
aljoscha: Yes, on your first part, and also the second paragraph!

src/adapter/src/coord/sequencer/inner.rs
```diff
@@ -3117,49 +3119,77 @@ impl Coordinator {
...
-                            if let Err(e) = result {
-                                warn!("internal_cmd_rx dropped before we could send: {:?}", e);
+                ExecuteResponse::SendingRowsStreaming {
+                    rows: mut rows_stream,
+                    ..
+                } => {
+                    let mut byte_size: u64 = 0;
+                    let mut diffs = Vec::new();
+                    let result = loop {
+                        match tokio::time::timeout(timeout_dur, rows_stream.next()).await {
```
DAlperin, line 3145: This is subtly different behavior. If the stream drip feeds you batches the timeout is reset per batch. This might be fine (especially considering the removed TODO, but just something to consider.

DAlperin (approved):
The adapter side looks good! The compute side seems sane but I'm not as familiar. One small note but otherwise LGTM

src/adapter/src/coord/sequencer/inner.rs
aljoscha: Good to call out! But yep, I'm also not too concerned

src/compute/src/compute_state.rs
aljoscha: true, doing just that!

src/compute/src/compute_state.rs
aljoscha: will change!

src/compute/src/compute_state/peek_stash.rs
aljoscha: Neat! I didn't know about that 🙌

aljoscha:
@teskje pushed up commits to address your (previously hidden) comments, if you want to take another look

src/compute-client/src/service.rs
```diff
@@ -334,6 +354,75 @@ where
...
+                                    .saturating_add(other_inline_rows.byte_len());
+
+                                if self_shard_id != other_shard_id {
+                                    let err = format!(
+                                        "internal error: shard IDs of stashed response do not match, got {self_shard_id} and {other_shard_id}",
+                                    );
+                                    PeekResponse::Error(err)
+                                } else if self_relation_desc != other_relation_desc {
+                                    let err = format!(
+                                        "internal error: RelationDesc of  stashed response do not match, got {:?} and {:?}",
```
teskje:
```suggestion
                                        "internal error: RelationDesc of stashed response do not match, got {:?} and {:?}",
```

misc/python/materialize/parallel_workload/action.py
```diff
@@ -1083,6 +1083,14 @@ def __init__(
...
             BOOLEAN_FLAG_VALUES
         )
+        self.flags_with_values["enable_compute_peek_response_stash"] = (
+            BOOLEAN_FLAG_VALUES
+        )
+        self.flags_with_values["compute_peek_response_stash_threshold_bytes"] = [
+            "0",  # "force enabled"
+            "1048576",  # 1 MiB, an in-between value
+            "314572800",  # 300 MiB, the production value
+        ]
```
teskje, line 1093: Should we also make these "variable system parameters" in `misc/python/materialize/mzcompose/__init__.py`, to increase the coverage of the inline path outside of parallel workload tests?

aljoscha:
@def- This is indeed failing some of the nightly benchmarks: https://buildkite.com/materialize/nightly/builds/12324#_

There's nothing much we can do: the new impl is slower for smaller result sets. Do we want to accept the regression? Or maybe add separate benchmark runs for "with result stash" and "without result stash"?

def-:
How about adding a more performant value in `ADDITIONAL_BENCHMARKING_SYSTEM_PARAMETERS`? If that isn't possible for some reason, we can add an ignore for the regressions.

aljoscha:
> How about adding a more performant value in `ADDITIONAL_BENCHMARKING_SYSTEM_PARAMETERS`? If that isn't possible for some reason, we can add an ignore for the regressions.

@def- I pushed a commit that disables the peek response stash for benchmarks that way. I'd say this is the easiest way to get predictable benchmark results. And the rest of CI/nightly ensure that it works correctly. Wdyt?

def-:
Perfect!

bosconi:
bugbot run

src/compute-client/src/protocol/response.rs
```diff
@@ -374,6 +380,90 @@ impl Arbitrary for PeekResponse {
...
+        // Consider a possible OFFSET.
+        num_rows = num_rows.saturating_sub(offset);
+
+        // Consider a possible LIMIT.
+        if let Some(limit) = limit {
+            num_rows = std::cmp::min(limit, num_rows);
+        }
+
+        num_rows
+    }
```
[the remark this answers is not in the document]
ggevay: This was a real bug, but has been fixed since then in eefc53999d08283acf0041cb4e43c80ed21d5636.

src/compute/src/compute_state/peek_stash.rs
```diff
@@ -0,0 +1,245 @@
...
+                }
+                Err(e) => {
+                    permit.send(Err(e));
+                }
+            }
+
+            num_batches -= 1;
+            if num_batches == 0 {
+                break;
+            }
```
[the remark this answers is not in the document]
ggevay: I think this is fine: `num_batches` is decremented one-by-one, so we'd break the loop before it underflows. There would be an issue only if `num_batches` is passed in to be 0 already at the beginning, but this would require a totally nonsensical value to be set for `compute_peek_stash_num_batches` in LD.

src/expr/src/relation.rs
```diff
@@ -3760,6 +3769,124 @@ impl RowSetFinishing {
...
+        // in the query history.
+        let response_size: usize = iter.clone().map(|row| row.data().len()).sum();
+
+        // Bail if we would end up returning more data to the client than they can support.
+        if let Some(max) = self.remaining_max_returned_query_size {
+            if response_size > usize::cast_from(max) {
+                let max_bytes = ByteSize::b(self.max_returned_query_size.expect("known to exist"));
+                return Err(format!("total result exceeds max size of {max_bytes}"));
+            }
+        }
```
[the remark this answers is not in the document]
teskje: https://github.com/MaterializeInc/materialize/pull/35560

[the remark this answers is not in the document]
ggevay: Nice catch! Opened a PR to fix it: https://github.com/MaterializeInc/materialize/pull/35561

src/compute/src/compute_state/peek_stash.rs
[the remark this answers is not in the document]
teskje: Still I think our code should be robust to non-sensical config values! Luckily this is an easy fix: https://github.com/MaterializeInc/materialize/pull/35562

PR #32336, opened 2025-04-25
add support for pod affinity and tolerations

doy-materialize:
### Motivation

https://github.com/MaterializeInc/database-issues/issues/9161

### Checklist

### DIFF
--- a/src/orchestrator-kubernetes/src/lib.rs
+++ b/src/orchestrator-kubernetes/src/lib.rs
@@ -20,6 +20,7 @@ use clap::ValueEnum;
 use cloud_resource_controller::KubernetesResourceReader;
 use futures::TryFutureExt;
 use futures::stream::{BoxStream, StreamExt};
+use k8s_openapi::DeepMerge;
 use k8s_openapi::api::apps::v1::{StatefulSet, StatefulSetSpec};
 use k8s_openapi::api::core::v1::{
     Affinity, Capabilities, Container, ContainerPort, EnvVar, EnvVarSource, EphemeralVolumeSource,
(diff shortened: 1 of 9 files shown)


src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -895,6 +900,16 @@ impl NamespacedOrchestrator for NamespacedKubernetesOrchestrator {
...
 
+        let mut affinity = Affinity {
+            pod_anti_affinity: anti_affinity,
+            pod_affinity,
+            node_affinity,
+            ..Default::default()
+        };
+        if let Some(service_affinity) = &self.config.service_affinity {
+            affinity.merge_from(serde_json::from_str(service_affinity)?);
+        }
```
jubrad, line 911: Is there any way we can merge with a provided anti-affinity?

I'm not sure that we'd even want to, but thought it might be worth getting the reason we don't want to written down.

jubrad (commented):
This PR seems like it'll work very well for our current requests, but I'm wondering if bringing this down a level (or two) would also make sense. On the cloud side this would mean we could more easily isolate by org, and on the self-managed (and cloud) sides the ability to isolate a workload onto specific instance groups seems relatively useful.

doy-materialize:
i'm not sure what you mean exactly - also add this to the materialize cr? or the cluster size map? i'd kiiiinda prefer to avoid adding more things to configure until we have an actual use for them, it should be reasonably easy to extend in the future

src/orchestrator-kubernetes/src/lib.rs
doy-materialize: i mean, i don't think a blanket anti-affinity configuration really makes any sense? unless you're just trying to use it as an alternative to taints, but then you can just... do that? this seems like another thing that would be better to hold off on until we have an actual use case

jubrad:
> i'm not sure what you mean exactly - also add this to the materialize cr? or the cluster size map? i'd kiiiinda prefer to avoid adding more things to configure until we have an actual use for them, it should be reasonably easy to extend in the future

I think it's partially called out [here](https://github.com/MaterializeInc/cloud/pull/11040/files#diff-f1ed19319a872081877f78a7b7e96fa3b5f440a626e1c33d112f0fc22ae24656R131), but yeah, we definitely don't need this to be in this PR.

jubrad (approved):
This looks good, although I think we'll want to find some way to override this per environment or perhaps per cluster eventually.

PR #32370, opened 2025-04-28
fix failing to parse timezones correctly

ptravers:
### Motivation

https://github.com/MaterializeInc/database-issues/issues/9236

see pg docs for full spec: https://www.postgresql.org/docs/12/datetime-input-rules.html

### Checklist

### DIFF
--- a/src/repr/src/adt/datetime.rs
+++ b/src/repr/src/adt/datetime.rs
@@ -1874,15 +1874,6 @@ pub enum CalendarEra {
     AD,
 }
 
-impl CalendarEra {
-    fn as_str(&self) -> &'static str {
-        match self {
-            CalendarEra::BC => "BC",
-            CalendarEra::AD => "AD",
-        }
-    }
-}
-
 /// Takes a 'date timezone' 'date time timezone' string and splits it into 'date
 /// {time}' and 'timezone' components then checks for a 'CalenderEra' defaulting
 /// to 'AD'.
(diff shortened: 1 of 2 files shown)


ptravers:
sure, for the first case, how would the calling function know that it's a timezone without checking the timezone table? there exist valid two character timezones.

for the second case, you're correct that is the other handled case but I don't agree that the splitting is incorrect since we only want to check for the next split if the first split is found. It makes sense to me that we would attempt to split again base `\s` when we find a remainder from the first split.

I don't really want to rewrite `split_timestamp_string` for this bug ticket. I would prefer to do it in a separate PR. I believe that this code is safe based on the tests we have written. I would like to get it out as soon as we can. I don't think that this code as written can ever result in multiples of era or timezone.

src/repr/src/adt/datetime.rs
```diff
@@ -1938,14 +1929,35 @@ pub(crate) fn split_timestamp_string(value: &str) -> (&str, &str, CalendarEra) {
...
+
+    // Covers case 3.
     match (
-        timezone_upper.strip_suffix(BC.as_str()),
-        timezone_upper.strip_suffix(AD.as_str()),
+        timezone_upper.strip_suffix(" BC"),
+        timezone_upper.strip_suffix(" AD"),
     ) {
         (Some(remainder), None) => (&timezone[..remainder.len()], BC),
         (None, Some(remainder)) => (&timezone[..remainder.len()], AD),
```
petrosagg, line 1963: It's not safe to index into a string based on a boundary obtained from another string. If the char boundaries are different it can cause a panic. The following commands cause the cluster to panic:

```sql
CREATE TABLE t1 (t text);
INSERT INTO t1 VALUES ('2025-01-01 00:00:00 Europe/Athensı AD');
SELECT t::timestamp FROM t1;
```

This is because the character `ı` is 2 bytes in lower case but one byte in upper case.

petrosagg (commented):
There is one more minor issue but otherwise we can merge that and refactor the parsing in a separate PR

src/repr/src/adt/datetime.rs
ptravers: Excellent point, will add that as a test

ptravers: done.

petrosagg (approved):

PR #32374, opened 2025-04-28
remove rc wrapper from SharedRow

ptravers:
I wasn't totally sure if the goal was to have the memory allocation be defined by the code initializing SharedRow and thereby the safety of this be up to the consumer or if we still wanted to use thread_local but with a const initialization which appears to be possible.

```rust
thread_local! {
  static SHARED_ROW: RefCell<Row> = const { RefCell::new(Row::empty()) }
}
```

### Motivation

https://github.com/MaterializeInc/database-issues/issues/9232

### Checklist

### DIFF
--- a/src/repr/src/row.rs
+++ b/src/repr/src/row.rs
@@ -144,6 +144,14 @@ impl Row {
         }
     }
 
+    /// Create an empty `Row`.
+    #[inline]
+    pub const fn empty() -> Self {
+        Self {
+            data: CompactBytes::empty(),
+        }
+    }
+
     /// Creates a new row from supplied bytes.
     ///
     /// # Safety
@@ -2807,7 +2815,7 @@ impl Default for RowArena {
     }
 }
 
-/// A thread-local row, which can be borrowed and returned.
+/// A row, which can be borrowed and returned.
 /// # Example
 ///
 /// Use this type instead of creating a new row:
@@ -2826,12 +2834,10 @@ impl Default for RowArena {
 ///
 /// [`SharedRow::get`] panics when trying to obtain multiple references to the shared row.
 #[derive(Debug)]
-pub struct SharedRow(Rc<RefCell<Row>>);
+pub struct SharedRow(RefCell<Row>);
 
 impl SharedRow {
-    thread_local! {
-        static SHARED_ROW: Rc<RefCell<Row>> = Rc::new(RefCell::new(Row::default()));
-    }
+    const SHARED_ROW: RefCell<Row> = RefCell::new(Row::empty());
 
     /// Get the shared row.
     ///
@@ -2841,7 +2847,7 @@ impl SharedRow {
     ///
     /// Panics when the row is already borrowed elsewhere.
     pub fn get() -> Self {
-        let row = Self::SHARED_ROW.with(Rc::clone);
+        let row = Self::SHARED_ROW;
         // Clear row
         row.borrow_mut().packer();
         Self(row)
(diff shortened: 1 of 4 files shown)


src/repr/src/row.rs
```diff
@@ -2826,12 +2834,10 @@ impl Default for RowArena {
...
 /// [`SharedRow::get`] panics when trying to obtain multiple references to the shared row.
 #[derive(Debug)]
-pub struct SharedRow(Rc<RefCell<Row>>);
+pub struct SharedRow(RefCell<Row>);
 
 impl SharedRow {
-    thread_local! {
-        static SHARED_ROW: Rc<RefCell<Row>> = Rc::new(RefCell::new(Row::default()));
-    }
+    const SHARED_ROW: RefCell<Row> = RefCell::new(Row::empty());
```
ptravers, line 2840: other option would be the following:

```rust
thread_local! {
  static SHARED_ROW: RefCell<Row> = const { RefCell::new(Row::empty()) }
}
```

ptravers: from thread_local documentation:

```
/// This macro supports a special `const {}` syntax that can be used
/// when the initialization expression can be evaluated as a constant.
/// This can enable a more efficient thread local implementation that
/// can avoid lazy initialization. For types that do not
/// [need to be dropped][crate::mem::needs_drop], this can enable an
/// even more efficient implementation that does not need to
/// track any additional state.
```

antiguru:
> `    const SHARED_ROW: RefCell<Row> = RefCell::new(Row::empty());`

`const` in Rust means that the value gets inlined wherever used. This is the opposite of what we want: We want to thread-local variable that stays allocated while the thread is alive, which is the current behavior. We'd like to change two things: Remove the `Rc`, and const-initialize the static thread-local variable. Moving towards const initialization reduces the cost of `SharedRow::get` to what should be a pointer dereference. In turn, we don't need the `Rc` anymore, as it just adds indirection.

petrosagg:
I happened to look into doing this last week. The easiest way to remove the`Rc` wrapper is to pay a syntactic noise cost at all use sites that will now need to use some sort of `SharedRow::with(|row| {... })` method and pass in all the logic in a closure. I checked the use sites and it's not terrible but I don't love it either.

If we don't want to pay the syntactic cost I think the next best thing is to define the thread local as `Option<RefCell<Row>>` and have the constructor take the value out and have the destructor put it back.

> `const` in Rust means that the value gets inlined wherever used.

The `thread_local` macro also accepts a `const` keyword that is not related to the normal semantics of the `const` keyword. I don't know why they did that but it has caused a bunch of confusion.

petrosagg:
Put up the `Option<_>` idea here https://github.com/MaterializeInc/materialize/pull/32375

ptravers:
@petrosagg are your concerns just syntactic with using thread_local const? from the docs it seems like it avoids lazy initialization. I haven't totally figured out how. as petros noted we would have to do `SharedRow::with(|row| {... })` at all the usage sites


```rust
thread_local! {
  static SHARED_ROW: RefCell<Row> = const { RefCell::new(Row::empty()) }
}
```

petrosagg:
> @petrosagg are your concerns just syntactic with using `thread_local` const

Yeah, our code is heavily indented due to many APIs expecting closures so if we can avoid adding another source of indentation it's a win (I think)

> from the docs it seems like it avoids lazy initialization. I haven't totally figured out how.

My understanding is that if the initial value is `const` then it has no side-effects which means it can simply put the bytes there without executing any logic and so there is no ordering issue. With non-const thread-locals the initialization logic of one might need to access another thread-local and so there has to be a "is the slot initialized" check in place

ptravers:
closing in favour of https://github.com/MaterializeInc/materialize/pull/32375

PR #32375, opened 2025-04-29
repr: store thread-local shared row inline

petrosagg:
This PR removes the Rc indirection for the shared row and simplifies the API by not requiring the caller to explicitly mutably borrow the RefCell.

The constructor of the `SharedRow` is documented to panic if more than one instance is constructed. This allows us to use a single `Cell<Option<Row>>` as the contents of the thread local that gets replaced with `None` when someone gets the shared row and is put back when the `SharedRow` value is dropped.

### Motivation

https://github.com/MaterializeInc/database-issues/issues/9232

### Checklist

### DIFF
--- a/src/repr/src/row.rs
+++ b/src/repr/src/row.rs
@@ -8,13 +8,12 @@
 // by the Apache License, Version 2.0.
 
 use std::borrow::Borrow;
-use std::cell::RefCell;
+use std::cell::{Cell, RefCell};
 use std::cmp::Ordering;
 use std::convert::{TryFrom, TryInto};
 use std::fmt::{self, Debug};
 use std::mem::{size_of, transmute};
 use std::ops::Deref;
-use std::rc::Rc;
 use std::str;
 
 use chrono::{DateTime, Datelike, NaiveDate, NaiveDateTime, NaiveTime, Timelike, Utc};
@@ -141,6 +140,14 @@ impl Row {
     pub fn with_capacity(cap: usize) -> Self {
         Self {
             data: CompactBytes::with_capacity(cap),
+        }
+    }
+
+    /// Create an empty `Row`.
+    #[inline]
+    pub const fn empty() -> Self {
+        Self {
+            data: CompactBytes::empty(),
         }
     }
 
@@ -2814,8 +2821,7 @@ impl Default for RowArena {
 /// ```
 /// use mz_repr::SharedRow;
 ///
-/// let binding = SharedRow::get();
-/// let mut row_builder = binding.borrow_mut();
+/// let mut row_builder = SharedRow::get();
 /// ```
 ///
 /// This allows us to reuse an existing row allocation instead of creating a new one or retaining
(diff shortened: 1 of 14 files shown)


src/repr/src/row.rs
```diff
@@ -2826,11 +2832,11 @@ impl Default for RowArena {
 ///
 /// [`SharedRow::get`] panics when trying to obtain multiple references to the shared row.
 #[derive(Debug)]
-pub struct SharedRow(Rc<RefCell<Row>>);
+pub struct SharedRow(Row);
 
 impl SharedRow {
     thread_local! {
-        static SHARED_ROW: Rc<RefCell<Row>> = Rc::new(RefCell::new(Row::default()));
+        static SHARED_ROW: Cell<Option<Row>> = const { Cell::new(Some(Row::empty())) }
```
ptravers, line 2839: might be worth leaving a note here, on lines 2850 and 2885 that the `Option` here is used a perf improvement to prevent `std::mem::take` from creating a new row when when it calls default in the underlying call to `std::mem::replace`. I'm not sure how obvious that would be here to people who are unfamiliar with take

src/repr/src/row.rs
```diff
@@ -2853,36 +2859,33 @@ impl SharedRow {
...
 
+impl std::ops::DerefMut for SharedRow {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+
+impl Drop for SharedRow {
+    fn drop(&mut self) {
+        Self::SHARED_ROW.set(Some(std::mem::take(&mut self.0)))
```
ptravers, line 2885: doesn't set also call drop on the inner value of the `Cell`? so wouldn't we be calling drop on `Option<SharedRow>` which would also be `self`?

better question, is it worth explicitly testing this behaviour if possible?

petrosagg: `SHARED_ROW.set(..)` does drop the value that is currently contained in the cell, but that value is always `None` when this drop function runs, so nothing is actually dropped. This is because the existence of an instance of `SharedRow` must have gotten through `SharedRow::get`, the only constructor, and that constructor *moves* the shared row contained in the thread local into the returned instance and leaves a `None` value in the cell.

What kind of test did you have in mind?

src/repr/src/row.rs
petrosagg: The `Option` here is not meant to be a performance improvement, and I would be surprised if the difference between `Cell<Option<Row>>` and `Cell<Row>` showed up in a benchmark. This is because calling calling `Row::default()` is very cheap, since it doesn't allocate.

The reason we need the `Option` here is because we want to track one bit of information, namely whether the shared row is currently borrowed. This allows us to catch any double borrows and panic whereas an `std::mem::take` would happily give back some `Row` every time.

Added a comment to clarify the structure

ptravers (approved):

src/repr/src/row.rs
ptravers: gotcha, that makes sense!

src/repr/src/row.rs
```diff
@@ -2841,9 +2851,9 @@ impl SharedRow {
     ///
     /// Panics when the row is already borrowed elsewhere.
     pub fn get() -> Self {
-        let row = Self::SHARED_ROW.with(Rc::clone);
+        let mut row = Self::SHARED_ROW.take().unwrap();
```
ptravers, line 2854: if I've understood correctly if we unwrap here then it would only fail in the case of a double borrow, correct? should this instead be `expect`?

src/repr/src/row.rs
ptravers: ahhhhhhhhh! ty! can we add that as a comment? I guess we could for a test:
```rust
let shared_row = SharedRow::get();
/// insert some data into shared_row
drop(shared_row);
assert!(SharedRow::get().is_empty());
```

also might be nice to test that we panic with a legible message when if we try to double borrow.

PR #32378, opened 2025-04-29
fix prepared statements should get different ids

ptravers:
### Motivation

https://github.com/MaterializeInc/database-issues/issues/9242

### Checklist

### DIFF
--- a/src/adapter/src/coord/statement_logging.rs
+++ b/src/adapter/src/coord/statement_logging.rs
@@ -306,7 +306,7 @@ impl Coordinator {
         let mut out = None;
 
         let uuid = match logging {
-            PreparedStatementLoggingInfo::AlreadyLogged { uuid } => uuid.clone(),
+            PreparedStatementLoggingInfo::AlreadyLogged { uuid } => *uuid,
             PreparedStatementLoggingInfo::StillToLog {
                 sql,
                 redacted_sql,
--- a/src/adapter/src/coord/statement_logging.rs
+++ b/src/adapter/src/coord/statement_logging.rs
@@ -306,9 +306,7 @@ impl Coordinator {
         let mut out = None;
 
         let uuid = match logging {
-            PreparedStatementLoggingInfo::AlreadyLogged { uuid: _ } => {
-                epoch_to_uuid_v7(&(self.now()))
-            }
+            PreparedStatementLoggingInfo::AlreadyLogged { uuid } => uuid.clone(),
             PreparedStatementLoggingInfo::StillToLog {
                 sql,
                 redacted_sql,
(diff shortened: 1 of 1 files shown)


src/adapter/src/coord/statement_logging.rs
```diff
@@ -306,7 +306,9 @@ impl Coordinator {
         let mut out = None;
 
         let uuid = match logging {
-            PreparedStatementLoggingInfo::AlreadyLogged { uuid } => *uuid,
+            PreparedStatementLoggingInfo::AlreadyLogged { uuid: _ } => {
+                epoch_to_uuid_v7(&(self.now()))
+            }
```
ggevay, line 311: I can take a more detailed look tomorrow, but I think it's not this place that needs a change. If I understand things correctly, `log_prepared_statement` is fine as it is. The problem is in `begin_statement_execution`, just after the call to `log_prepared_statement`, where `record_statement_lifecycle_event` is called. Note that these are writing to two different builtin tables: `log_prepared_statement` writes some general info about each prepared statement, so it needs to write just one row about each prepared statement, while `record_statement_lifecycle_event` records several individual lifecycle events on an _execution_ of a prepared statement.

(You can take a look at how the code was before c539850f9132d6b7e7c1d55d746345ef42b4579c: we were generating a new UUID (`ev_id`) to give to the call to `record_statement_lifecycle_event`. This was a different uuid from what `log_prepared_statement` was returning.)

ptravers: ah I see. sorry! I misunderstood your previous comment

ggevay: No problem!

ptravers: fixed, sorry about that!

src/adapter/src/coord/statement_logging.rs
```diff
@@ -719,14 +719,14 @@ impl Coordinator {
         }
         let (ps_record, ps_uuid) = self.log_prepared_statement(session, logging)?;
 
+        let now = self.now();
         let uuid = match session.qcell_rw(logging) {
-            PreparedStatementLoggingInfo::AlreadyLogged { uuid } => *uuid,
+            PreparedStatementLoggingInfo::AlreadyLogged { uuid: _ } => epoch_to_uuid_v7(&now),
```
ggevay, line 724: Is there a reason for looking at `session.qcell_rw(logging)` here? Couldn't we just unconditionally do `epoch_to_uuid_v7(&now)`? (That is roughly what the original code did before  c539850f9132d6b7e7c1d55d746345ef42b4579c.)

Note that the code as it is in the PR currently has the problem that if `log_prepared_statement` runs into throttling repeatedly and doesn't switch to `AlreadyLogged`, then we run the `StillToLog` code path here more than once for the same prepared statement. This would be a problem, because then we pick `epoch_to_uuid_v7(prepared_at)` repeatedly, so we would again log different executions with the same UUID.

ptravers: it wouldn't be the same uuid. there's a separate randomness component that prevents that in UUIDv7. but if all we care about is the id always updating that sounds fine.

ptravers: can refactor!

ggevay: Ah ok, then it's just that the code would be simpler if we don't condition on `session.qcell_rw(logging)`.

ggevay (approved):
Thank you!

src/adapter/src/coord/statement_logging.rs
ptravers: fixed! I think the original request when I talked with parker was to use prepared_at where possible since it would keep statement uuids in the order they were create?

ggevay: I think it's just that the id column's ordering should match the ordering of the timestamps that are in the same table. For the statement lifecycle table, the timestamps are the execution timestamps, so the ids' ordering should also be based on the execution timestamps. But I'm not so familiar with Persist; @ParkMyCar, can you confirm that this is the goal here?

PR #32382, opened 2025-05-01
use du for storage metrics in process orchestrator

jubrad:
### Motivation
Currently we just don't collect these which makes it hard to understand materialize performance, debug issues, and even know if lgalloc is on or not. This seems like a relatively easy way of turning disk metrics on for the emulator. Unfortunately the console still wouldn't show these in the emulator since that feature is tied to whether we're running with flexible deployment or not. We could add a db system param that that could look for instead though. 

### Checklist

### DIFF
--- a/src/orchestrator-process/src/lib.rs
+++ b/src/orchestrator-process/src/lib.rs
@@ -587,11 +587,55 @@ impl OrchestratorWorker {
                     }
                 }
             };
+
+            // Get disk usage for this process if it has a scratch directory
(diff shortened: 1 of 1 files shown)


antiguru:
`du` won't work for lgalloc because its files are unlinked and `du` can't see them. However, lgalloc already has internal reporting that is more accurate, so we should rely on this instead.

I don't know about RocksDB, the only other contributor to disk.

PR #32389, opened 2025-05-02
[persist] Postgres specific consensus queries

ParkMyCar:
This PR adds some new queries to `PostgresConsensus` that are tuned against vanilla postgres to obtain row level locks as opposed to the table level locks that the current queries get. These new queries are only used if `persist_use_postgres_tuned_queries` is set to true, __and__ we detect we're running against vanilla Postgres. 

Locks acquired on `main`
```
  pid  |       mode       | granted
-------+------------------+---------
 21004 | AccessShareLock  | t
 21004 | RowExclusiveLock | t
```

Locks acquired with the new feature enabled
```
  pid  |       mode       | granted
-------+------------------+---------
 21004 | RowShareLock     | t
 21004 | RowExclusiveLock | t
```

It's not entirely clear that this improves performance, but some simulation testing seems to indicate that it does.

### Testing

Also included in this PR are new CI workflows that run platform-checks and maelstrom against Postgres consensus, using the new Postgres queries.

### Motivation

Improve performance of self-hosted which runs against vanilla postgres

### Checklist

### DIFF
--- a/src/persist/src/postgres.rs
+++ b/src/persist/src/postgres.rs
@@ -35,7 +35,8 @@ use tracing::{info, warn};
 use crate::error::Error;
 use crate::location::{CaSResult, Consensus, ExternalError, ResultStream, SeqNo, VersionedData};
 
-pub(crate) const USE_POSTGRES_TUNED_QUERIES: mz_dyncfg::Config<bool> = mz_dyncfg::Config::new(
+/// Flag to use concensus queries that are tuned for vanilla Postgres.
+pub const USE_POSTGRES_TUNED_QUERIES: mz_dyncfg::Config<bool> = mz_dyncfg::Config::new(
     "persist_use_postgres_tuned_queries",
     false,
     "Use a set of queries for consensus that have specifically been tuned against
--- a/src/persist/src/postgres.rs
+++ b/src/persist/src/postgres.rs
@@ -399,14 +399,13 @@ impl Consensus for PostgresConsensus {
             /// minimizes possible serialization conflicts.
             static POSTGRES_CAS_QUERY: &str = "
             WITH last_seq AS (
-                SELECT sequence_number
-                FROM materialize1.consensus
+                SELECT sequence_number FROM consensus
                 WHERE shard = $1
                 ORDER BY sequence_number DESC
                 LIMIT 1
                 FOR UPDATE
             )
-            INSERT INTO materialize1.consensus (shard, sequence_number, data)
+            INSERT INTO consensus (shard, sequence_number, data)
             SELECT $1, $2, $3
             FROM last_seq
             WHERE last_seq.sequence_number = $4;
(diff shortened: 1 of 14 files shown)


ParkMyCar:
Nightlies with the feature gate set to true: https://buildkite.com/materialize/nightly/builds/11948

ParkMyCar:
> A bunch of CI uses Cockroach explicitly instead of Postgres.

Great! I saw the `CockroachOrPostgres` wrapper on a number of tests, but figured explicitly adding it would be nice too. Happy to roll back the changes to platform-checks if you want! I'd like to keep the maelstrom changes though because it seems like that was always using Cockroach?

> I'm worried about all the benchmark results we are seeing: https://buildkite.com/materialize/nightly/builds/11949

This is a great callout! I'm not too concerned because the queries are feature gated so we can always turn them off, but also these tuned queries are designed to fix a large amount of contention that we're seeing in a customers environment. So even if they're a bit slower it's better than the experience we're currently observing

test/platform-checks/mzcompose.py
```diff
@@ -82,19 +81,21 @@ def create_mzs(
...
+            metadata_store="postgres-metadata" if postgres_consensus else "cockroach",
         )
     ]
 
 
 SERVICES = [
     TestCerts(),
-    # TODO(def-): Switch to CockroachOrPostgres after we have 4 versions
-    # support Postgres as metadata store
     Cockroach(
```
def-, line 91: This should use CockroachOrPostgres to allow switching more easily, tell me if I should make the switch.

def- (approved):
Test changes are mostly fine for me, left one comment

def- (approved):

test/platform-checks/mzcompose.py
ParkMyCar: That would be great! Mind making the switch in a follow up PR? I want to get this one merged ASAP

src/persist/src/postgres.rs
```diff
@@ -424,12 +482,49 @@ impl Consensus for PostgresConsensus {
...
+            ORDER BY sequence_number ASC
+            LIMIT 1
+            FOR UPDATE
+        ),
+        to_lock AS (
+            SELECT ctid FROM consensus
+            WHERE shard = $1
+            AND sequence_number < $2
+            AND EXISTS (SELECT * FROM newer_exists)
+            ORDER BY sequence_number DESC
```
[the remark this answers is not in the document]
ParkMyCar: Added a comment! My thinking could be wrong here but AFAIU this `ORDER BY` ensure minimal conflict between operations

test/platform-checks/mzcompose.py
def-: Works for me, will take a look tomorrow.

src/persist/src/postgres.rs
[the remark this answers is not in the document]
ptravers: the order by should force the reverse index seq scan, no?

PR #32418, opened 2025-05-06
storage: document and uphold upsert correctness property

petrosagg:
This PR fixes the correctness issue observed during concurrent execution of two upsert sources on multi-replica clusters.

The issue stems from a violation of a correctness proprety of the `drain_staged_input` function, that this PR now documents.

The diff is meant to be minimal to aid quick review and backporting but it does not make any progress towards simplifying the overal logic, which will be subsequent work.

## Explanation

When `drain_staged_input` is called in `DrainStyle::ToUpper` mode it drains all the upsert commands that are exactly at the timestamp of `persist_upper` and for each key it:

1. Keeps the command with the highest offset
2. Retrieves the current state from RocksDB
3. Produces a retraction of the current value if it was set
4. Produces an insertion of the new value.

In this mode the RocksDB state is NOT modified. Therefor if the function is called twice with the same `persist_upper` and processes a command for the same key it will produce a double retraction, since the second call will not be aware of the first call.

The code was previously set up to call `drain_staged_input` once per held capability with the same `persist_upper`. Due to how the rest of the code was setup the set of held capabilities always turned out to be a singleton when a single replica was involved, therefore not triggering the bug.

With multiple replicas however one instance of the operator could receive the following in its command stream:

* cap=1, batch = `[ ({key=b, value=NULL}, ts=1), ({key=a, value=a}, ts=2) ]`
* cap=2, batch = `[ ({key=a, value=b}, ts=2) ]`
* progress event to {3}

Then another instance of the operator seeing the same stream of commands processes everything till `{2}` and manages to commit to persist.

Finally the original operator receives a persist upper progress to `{2}` and ends up calling `drain_staged_input` twice since it holds onto two capabilities and ends up processing the same key (`key=a`) twice.

Fixes MaterializeInc/database-issues#9160

### Checklist

### DIFF
--- a/src/storage/src/upsert_continual_feedback.rs
+++ b/src/storage/src/upsert_continual_feedback.rs
@@ -11,9 +11,6 @@
 //! [`upsert_inner`] for a description of how the operator works and why.
 
 use std::cmp::Reverse;
-// We don't care about the order, but we do want drain().
-#[allow(clippy::disallowed_types)]
-use std::collections::HashMap;
 use std::fmt::Debug;
 use std::sync::Arc;
 
@@ -164,7 +161,8 @@ where
     let upsert_shared_metrics = Arc::clone(&upsert_metrics.shared);
 
     let shutdown_button = builder.build(move |caps| async move {
-        let [mut output_cap, snapshot_cap, health_cap]: [_; 3] = caps.try_into().unwrap();
+        let [output_cap, snapshot_cap, health_cap]: [_; 3] = caps.try_into().unwrap();
+        drop(output_cap);
         let mut snapshot_cap = CapabilitySet::from_elem(snapshot_cap);
 
         // The order key of the `UpsertState` is `Option<FromTime>`, which implements `Default`
@@ -192,9 +190,10 @@ where
         let mut multi_get_scratch = Vec::new();
 
         // For stashing source input while it's not eligible for processing.
-        // We don't care about the order, but we do want drain().
-        #[allow(clippy::disallowed_types)]
-        let mut stash = HashMap::new();
+        let mut stash = vec![];
+        // A capability suitable for emitting any updates based on stash. No capability is held
+        // when the stash is empty.
+        let mut stash_cap: Option<Capability<G::Timestamp>> = None;
         let mut input_upper = Antichain::from_elem(Timestamp::minimum());
         let mut partial_drain_time = None;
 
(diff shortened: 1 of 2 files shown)


src/storage/src/upsert_continual_feedback.rs
```diff
@@ -342,19 +341,31 @@ where
 
                                 stage_input(
                                     &mut stash,
-                                    cap,
                                     &mut data,
                                     &input_upper,
                                     &resume_upper,
                                 );
+                                if !stash.is_empty() {
+                                    // Update the stashed capability to the minimum
```
ParkMyCar, line 349: If I understand correctly, this comment seems a little misleading because it's the "minimum" time which kind of means the "maximum" capability?

caveat: I definitely have a more limited understanding of timely, so feel free to ignore me here!

src/storage/src/upsert_continual_feedback.rs
```diff
@@ -432,23 +433,18 @@ where
             // loop. More of our stash can become eligible for draining both
             // when the source-input frontier advances or when the persist
             // frontier advances.
+            if !stash.is_empty() {
+                let cap = stash_cap.as_mut().expect("missing capability for non-empty stash");
```
ParkMyCar, line 437: AFAIU this change here seems to be the fix for the original issue? Previously because `stash` was a hashmap we could end up with the same key with two different values spread across two different `(cap, updates)` sets. Thus `drain_staged_input` could get called multiple times with the same `persist_upper` and process the same key.

Now because we process the entire `stash` at once, and we have the check on line 423, we're guaranteed to only call `drain_staged_input` with a `DrainStyle::ToUpper` with a given `persist_upper`, a single time.

If that understanding is correct, do you mind adding a comment here explaining how this is load bearing? Or maybe how these pieces work together to maintain the new correctness property documented on `drain_staged_input`?

ParkMyCar (approved):
I'm not very familiar with the `upsert_continual_feedback` but this seems to solve the described problem as far as I understand it!

src/storage/src/upsert_continual_feedback.rs
```diff
@@ -342,19 +341,31 @@ where
...
                                 );
+                                if !stash.is_empty() {
+                                    // Update the stashed capability to the minimum
+                                    stash_cap = match stash_cap {
+                                        Some(stash_cap) => {
+                                            if cap.time() < stash_cap.time() {
+                                                Some(cap)
+                                            } else {
+                                                Some(stash_cap)
+                                            }
```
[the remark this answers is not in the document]
petrosagg: No because we compare `cap.time()` but we actually want `cap` itself, and `Capability<_>` does not implement `Ord`

src/storage/src/upsert_continual_feedback.rs
petrosagg: It's true that the minimum capability is maximally flexible. I've always seen these adjectives refer to the contained time and not to the capability's flexibility though so I think "minimum" here matches the intended meaning of retaining the capability with the minimum time

src/storage/src/upsert_continual_feedback.rs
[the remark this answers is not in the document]
ParkMyCar: Yep 100%, I even commented on this difference below 🤦

def-:
Nightly looks good, thanks for running it: https://buildkite.com/materialize/nightly/builds/11974 (scalability has also recently been flaking in main)

src/storage/src/upsert_continual_feedback.rs
bkirwi: One option is to keep a `CapabilitySet` here instead... IIUC will do the equivalent minimum-finding, but maybe more explicit about intent? Would be fine as a followup in any case!

bkirwi (approved):

src/storage/src/upsert_continual_feedback.rs
petrosagg: Technically having the `HashMap` would be fine too if the managed vectors of updates were appropriately distributed by timestamp. So even though this change fixes the issue it's not load bearing in the sense that it must be this way.

To avoid going another CI round I'll merge as is and we will revisit this code soon since we want to simplify its logic further and this section will definitely be affected.

PR #32431, opened 2025-05-07
helm-chart release: Also bump terraform versions automatically

def-:
Is it that simple?

### Checklist


misc/helm-charts/publish.sh
```diff
@@ -97,6 +97,72 @@ else
...
+  echo "Failing because Helm Chart was not successfully published"
+  exit 1
+fi
+
+echo "--- Bumping terraform-helm-materialize"
+rm -rf terraform-helm-materialize
+git clone https://"$GITHUB_TOKEN"@github.com/MaterializeInc/terraform-helm-materialize.git
+cd terraform-helm-materialize
+sed -i "s/\".*\"\(.*\) # META: helm-chart version/\"$CI_HELM_CHART_VERSION\"\\1 # META: helm-chart version/" variables.tf
+sed -i "s/\".*\"\(.*\) # META: mz version/\"$CI_MZ_VERSION\"\\1 # META: mz version/" variables.tf
```
def-, line 127: I couldn't find any tool that can edit our variables.tf files cleanly, tried a bunch. Thus the META comments to make sed work.

def-:
Ready for review.

kay-kim (approved):

misc/helm-charts/publish.sh
jubrad: huh... so this is a good idea, but it probably breaks when we have two LTS versions and we need to patch the prior one. I mean it won't break break, but the newest TF module may point to the newest version of the prior LTS, which would be confusing. 

We might need to use the same name for the helm chart and LTS versions if we're going to upgrade them in lockstep?

misc/helm-charts/publish.sh
```diff
@@ -97,6 +97,72 @@ else
...
+git add variables.tf README.md
+git commit -m "Bump to helm-chart $CI_HELM_CHART_VERSION, materialize $CI_MZ_VERSION"
+# Bump the patch version by one (v0.1.12 -> v0.1.13)
+TERRAFORM_HELM_VERSION=$(git for-each-ref --sort=creatordate --format '%(refname:strip=2)' refs/tags | grep '^v' | tail -n1 | awk -F. -v OFS=. '{$NF += 1; print}')
+git tag "$TERRAFORM_HELM_VERSION"
+git push origin main "$TERRAFORM_HELM_VERSION"
+cd ..
+
+declare -A TERRAFORM_VERSION
+for repo in terraform-aws-materialize terraform-azurerm-materialize terraform-google-materialize; do
```
jubrad, line 140: I'm not sure if we can full automate these all the way through the path...It could be the case that updating the operator actually means we need (or want) updates to the TF code itself to support some new feature or create a new configmap that we're plugging into an operator chart value.

I'm wondering if we should do something like
1. update/publish opeartor helm and envd backport
2. auto generate the new docs update with operator/envd versions
3. auto create a new helm chart release pr 
4. manually (as part of the runbook or templated issue) merge and release all PRs for TF making the appropriate changes for the new feature. 


This is a lot more manually steps, though... maybe there's a flag on this script to release terraform modules that we only check if we're not going to have that additional work. I'm open to suggestions, but I'm not super sure about adding the TF module in the docs if they don't have the features implemented for the version we're trying to release.

jubrad (commented):
I have a couple of questions. I think this would work for many cases, but seems like there are some times that it could be problematic especially if we start back porting changes to a prior LTS release.

misc/helm-charts/publish.sh
def-: I think we can check if there are higher versions released already (for example v25.2.5) and then not mark the newly released v25.1.13 as the latest version etc.

def-: This is addressed in https://github.com/MaterializeInc/materialize/pull/32431/commits/2009dfb9a55bc930fad0751c4eb931641a247d0d

misc/helm-charts/publish.sh
def-: This is addressed by https://github.com/MaterializeInc/materialize/pull/32431/commits/2009dfb9a55bc930fad0751c4eb931641a247d0d (and the corresponding [trigger-ci change](https://github.com/MaterializeInc/trigger-ci/commit/ad0dc10dff3f775839fcaf4d0ff83d1d8cc253db))
Screenshot 2025-05-09 at 19 06 11

bobbyiliev (commented):
I think that this looks good. Only thing that comes to mind is that if I'm reading this correctly, it bumps all Terraform modules right after the Helm chart is released. Should we be concerned that this assumes `main` of each Terraform module is always compatible with the new chart? And also, could that be risky if `main` of any module has unreleased changes that aren't meant to go out yet?

def-:
> Should we also probably bump the Terraform CI tests to use the latest version of the modules?

Oh, right... Will fix, but in the end this is all a big circle of dependencies now, I guess I'm starting to understand using mono repos now. `materialize -> terraform-helm-materialize -> terraform-{aws,azurerm,google} -> materialize`

> Should we be concerned that this assumes main of each Terraform module is always compatible with the new chart? And also, could that be risky if main of any module has unreleased changes that aren't meant to go out yet?

Yes, except if you disable the terraform bump in trigger-ci, then you have to deal with the rest manually as is being done currently.

bobbyiliev (approved):

PR #32438, opened 2025-05-07
catalog: fix the ordering of `CONNECTION` objects when bootstrapping

ParkMyCar:
This PR fixes an issue with our `CONNECTION` objects. `CONNECTION`s can depend on one another and be `ALTER`-ed. This means that a connection can depend on another object with an ID greater than its own which breaks some of our assumptions

The fix here is to do a topological sort of `CONNECTION` objects when applying updates.

### Motivation

Fix a known panic.

### Checklist

### DIFF
--- a/src/adapter/src/catalog/apply.rs
+++ b/src/adapter/src/catalog/apply.rs
@@ -38,8 +38,9 @@ use mz_compute_types::config::ComputeReplicaConfig;
 use mz_controller::clusters::{ReplicaConfig, ReplicaLogging};
 use mz_controller_types::ClusterId;
 use mz_expr::MirScalarExpr;
+use mz_ore::collections::CollectionExt;
 use mz_ore::tracing::OpenTelemetryContext;
-use mz_ore::{instrument, soft_assert_no_log};
+use mz_ore::{assert_none, instrument, soft_assert_no_log};
 use mz_pgrepr::oid::INVALID_OID;
 use mz_repr::adt::mz_acl_item::{MzAclItem, PrivilegeMap};
 use mz_repr::role_id::RoleId;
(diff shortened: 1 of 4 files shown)


src/adapter/src/catalog/apply.rs
```diff
@@ -1935,6 +1936,71 @@ fn sort_updates_inner(updates: Vec<StateUpdate>) -> Vec<StateUpdate> {
...
 
+    /// Sort `CONNECTION` items.
+    ///
+    /// `CONNECTION`s can depend on one another, e.g. a `KAFKA CONNECTION` contains
+    /// a list of `BROKERS` and these broker definitions can themselves reference other
+    /// connections. This `BROKERS` list can also be `ALTER`ed and thus it's possible
+    /// for a connection to depend on another with an ID greater than its own.
+    fn sort_connections(connections: &mut Vec<(mz_catalog::durable::Item, Timestamp, StateDiff)>) {
+        let mut topo: BTreeMap<
+            (mz_catalog::durable::Item, Timestamp, StateDiff),
```
ptravers, line 1947: nit: could we define this as a named type?

src/adapter/src/catalog/apply.rs
```diff
@@ -1935,6 +1936,72 @@ fn sort_updates_inner(updates: Vec<StateUpdate>) -> Vec<StateUpdate> {
...
+            // dependency already exists and thus it's not in `connections`.
+            dependencies.retain(|dep| existing_connections.contains(dep));
+
+            // Be defensive and ensure we're not clobbering any items.
+            assert_none!(topo.insert((connection, ts, diff), dependencies));
+        }
+        tracing::info!(?topo, ?existing_connections, "built topological sort");
+
+        // Do a topological sort, pushing back into the provided Vec.
+        while !topo.is_empty() {
```
petrosagg, line 1973: This looks like a quadratic loop if you have a chain of connections that depend on each other when this can be linear. 

You can get a better algorithm by constructing a `todo` vector that initially contains only items with no dependencies, which you find from above. Then in this loop you repeatedly pop an item from the todo, push any dependencies it might have back in todo with a single lookup, and then push the item back in `connections`.

src/adapter/src/catalog/apply.rs
```diff
@@ -1935,6 +1936,72 @@ fn sort_updates_inner(updates: Vec<StateUpdate>) -> Vec<StateUpdate> {
...
+    /// for a connection to depend on another with an ID greater than its own.
+    fn sort_connections(connections: &mut Vec<(mz_catalog::durable::Item, Timestamp, StateDiff)>) {
+        let mut topo: BTreeMap<
+            (mz_catalog::durable::Item, Timestamp, StateDiff),
+            BTreeSet<CatalogItemId>,
+        > = BTreeMap::default();
+        let existing_connections: BTreeSet<_> = connections.iter().map(|item| item.0.id).collect();
+
+        // Initialize our set of topological sort.
+        tracing::info!(?connections, "sorting connections");
```
ptravers, line 1953: q: will this be spammy in the logs? or do we not see alter connection a lot?

petrosagg (commented):
The code looks fine to me but we've been bitten by accidentally quadratic things before so I think it's worth restructuring

ptravers (approved):

src/adapter/src/catalog/apply.rs
ptravers: do we have any data on how large the set of connections gets in practice? can we do the algorithm improvements in a follow up? no totally sure how broken flare is in prod right now either.

petrosagg: It's a simple change to make it linear so I don't think it makes sense to do a follow up with another round of review and a chance of never happening. It's just BFS :)

ptravers: fair point

ParkMyCar: @ptravers is interested in improving this algo, to get folks unblocked I'm going to merge this so I can cut a patch release :)

PR #25221, opened 2024-02-14
sql: support parameters in LIMIT expressions

benesch:
When the `enable_expressions_in_limit_syntax` feature flag is enabled, this commit teaches Materialize to allow parameters in LIMIT expressions.

The trick is to allow a `RowSetFinishing` to be generic over the type of the limit, so that the planner can hold the finishing's limit as an `HirScalarExpr` long enough to call `bind_parameters` on it.

Fix https://github.com/MaterializeInc/materialize/issues/24893.

### Motivation

  * This PR adds a known-desirable feature.

### Checklist

- [x] This PR has adequate test coverage / QA involvement has been duly considered.
- [x] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [x] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [x] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [x] This PR includes the following [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note):
  - Allow `LIMIT` expressions to contain parameters.

PR #32443, opened 2025-05-08
adapter: Support parameters in OFFSET clause

ggevay:
This PR adds support for prepared statement parameters in `OFFSET` clauses. It follows a similar approach as https://github.com/MaterializeInc/materialize/pull/25221 did for `LIMIT` expressions: We make `RowSetFinishing` generic over the type of the offset, so that the planner can hold the finishing's offset as an `HirScalarExpr` long enough to call `bind_parameters` on it.

I suggest starting the review from `plan_query_inner` in `query.rs`, since that is the first thing that runs from among all the touched code. The other important thing is in `plan_select_inner`. And then all the other things are just in service of these two things.

For the type of the `OFFSET` clause's expression, I went with `ScalarType::Int64`, i.e., `bigint` for now, as [discussed on slack](https://materializeinc.slack.com/archives/C08A62E0751/p1746711669627839?thread_ts=1746563801.896649&cid=C08A62E0751).

I've also added various TODOs to existing code. I'm planning to do those in follow-up PRs.

There is also the issue that the type of the arguments given in `EXECUTE` calls have to exactly match the expected types, i.e., it's not enough to be castable to it. This seems to be a general limitation of our prepared statements, e.g., it also occurs for the `LIMIT` clause. I'm somewhat surprised that this hasn't caused problems so far, but we can work on this separately if it turns out to be a problem. (I'll at least open an issue shortly.)

Nightly: https://buildkite.com/materialize/nightly/builds/11986

### Motivation

  * This PR adds a known-desirable feature: https://github.com/MaterializeInc/database-issues/issues/9009

### Checklist

### DIFF
--- a/src/sql/src/plan/hir.rs
+++ b/src/sql/src/plan/hir.rs
@@ -25,11 +25,13 @@ use mz_expr::AggregateFunc::{FusedWindowAggregate, WindowAggregate};
 pub use mz_expr::{
     BinaryFunc, ColumnOrder, TableFunc, UnaryFunc, UnmaterializableFunc, VariadicFunc, WindowFrame,
 };
+use mz_ore::cast::CastInto;
 use mz_ore::collections::CollectionExt;
-use mz_ore::stack;
+use mz_ore::error::ErrorExt;
 use mz_ore::stack::RecursionLimitError;
 use mz_ore::str::separated;
 use mz_ore::treat_as_equal::TreatAsEqual;
+use mz_ore::{soft_assert_or_log, stack};
 use mz_repr::adt::array::ArrayDimension;
 use mz_repr::adt::numeric::NumericMaxScale;
 use mz_repr::*;
@@ -165,10 +167,26 @@ pub enum HirRelationExpr {
         group_key: Vec<usize>,
         /// Column indices used to order rows within groups.
         order_key: Vec<ColumnOrder>,
-        /// Number of records to retain
+        /// Number of records to retain.
+        /// It is of ScalarType::Int64. (It's not entirely clear why not UInt64, see below at
+        /// `offset`.)
         limit: Option<HirScalarExpr>,
-        /// Number of records to skip
-        offset: usize,
+        /// Number of records to skip.
+        /// It is of ScalarType::Int64.
+        /// This can contain parameters at first, but by the time we reach lowering, this should
+        /// already be simply a Literal.
+        ///
+        /// TODO: It's not clear why this is Int64 instead of UInt64. If it were UInt64, we wouldn't
+        /// need to manually check non-negativity, but would just get this for free when casting to
+        /// UInt64.
+        /// There are some arguments for Postgres-compatibility, because Postgres expects an
+        /// Int64. But it's not clear whether there is an actual situation where having a UInt64
+        /// here would introduce a Postgres-incompatibility.
+        /// Another thing is that our prepared statements currently have the limitation that the
+        /// argument passed to EXECUTE must have exactly the same type as the parameter, i.e., it's
+        /// not enough if it's castable. Since UInt64 is not a standard type, various tools might
+        /// have trouble passing it.
+        offset: HirScalarExpr,
         /// User-supplied hint: how many rows will have the same group key.
         expected_group_size: Option<u64>,
     },
@@ -1665,7 +1683,7 @@ impl HirRelationExpr {
         group_key: Vec<usize>,
         order_key: Vec<ColumnOrder>,
         limit: Option<HirScalarExpr>,
-        offset: usize,
+        offset: HirScalarExpr,
         expected_group_size: Option<u64>,
     ) -> Self {
         HirRelationExpr::TopK {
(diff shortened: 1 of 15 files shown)


src/sql/src/plan/hir.rs
```diff
@@ -2718,15 +2774,20 @@ impl VisitChildren<HirScalarExpr> for HirRelationExpr {
...
                 group_key: _,
                 order_key: _,
-                limit: _,
-                offset: _,
+                limit,
+                offset,
                 expected_group_size: _,
+            } => {
+                if let Some(limit) = limit {
+                    f(limit)
```
ggevay, line 2786: It was a bug on `main` that `limit` was not visited here! (and the same is 2 more cases below)

test/sqllogictest/order_by.slt
```diff
@@ -1174,3 +1174,409 @@ Used Indexes:
...
+query error db error: ERROR: column "a" does not exist
+PREPARE p_error AS
+SELECT *
+FROM foo
+ORDER BY a, b
+OFFSET a;
+
+# TODO: This error msg is incomplete: it's missing the `integer_to_bigint\(\#\^0\{a\}\)` from the end, which is present
+# in the actual error msg. If we include that, then sqllogictest panics due to a bug in sqllogictest and/or in
+# `regex::escape`.
```
ggevay, line 1319: I will open an issue for this, and fix it separately. (Might be as simple as adding the `x` flag when working with the resulting regex, and then `\#` will be considered a valid escape sequence.)

src/sql/src/plan/lowering.rs
```diff
@@ -1686,7 +1699,10 @@ impl HirScalarExpr {
                 els: Box::new(els.lower_uncorrelated()?),
             },
             Select { .. } | Exists { .. } | Parameter(..) | Column(..) | Windowing(..) => {
-                sql_bail!("unexpected ScalarExpr in uncorrelated plan: {:?}", self);
+                sql_bail!(
+                    "Internal error: unexpected HirScalarExpr in lower_uncorrelated: {:?}",
+                    self
+                );
```
ggevay, line 1705: This was already kind of an internal error in the sense that it was totally not understandable for users. I've now made this official, by explicitly saying that this is an "Internal error".

There are many calls to this function, and I'm planning to go through all of them, to sanity check them that they avoid this error.

src/sql/src/plan/hir.rs
```diff
@@ -3423,6 +3522,58 @@ impl HirScalarExpr {
             }
         })
     }
+
+    /// Attempts to simplify this expression of [`ScalarType::Int64`] to a literal Int64 and
+    /// returns it as an usize.
```
ParkMyCar, line 3527:
```suggestion
    /// returns it as an i64.
```

src/sql/src/plan/statement/dml.rs
```diff
@@ -216,7 +217,21 @@ fn plan_select_inner(
...
-    // A top-level limit cannot be data dependent so eagerly evaluate it.
+    // OFFSET clauses in `expr` should become constants with the above binding of parameters.
+    // Let's check this and simplify them to literals.
+    expr.try_visit_mut_pre(&mut |expr| {
+        if let HirRelationExpr::TopK { offset, .. } = expr {
+            let offset_value = offset_into_value(offset.take())?;
+            *offset = HirScalarExpr::literal(Datum::Int64(offset_value), ScalarType::Int64);
+        }
+        Ok::<(), PlanError>(())
+    })?;
```
ParkMyCar, line 228: Why do we simply here for offsets but not limits? It seems like we should be simplifying too? I might be misunderstanding though!

ParkMyCar (approved):
Great work!! Thanks Gabor!!

src/sql/src/plan/statement/dml.rs
ggevay: `LIMIT` clauses in `expr` need not be simplified into a simple value to be able to lower them, see https://github.com/MaterializeInc/materialize/pull/23276
So, we can keep working with them even if they can't be simplified to a value. If they can be simplified to a value, then the optimizer will do that later.

(But even `LIMIT` clauses need to be simplified into a constant in the `RowSetFinishing`, which happens a few lines below.)

I'll add a code comment to note this.

src/sql/src/plan/hir.rs
```diff
@@ -3355,6 +3425,11 @@ impl HirScalarExpr {
         f(depth, self)
     }
 
+    /// Returns an _internal_ error if the expression contains
```
antiguru, line 3428: Comment is incorrect, function returns an option.

src/sql/src/plan/hir.rs
```diff
@@ -3423,6 +3522,58 @@ impl HirScalarExpr {
...
+            .simplify_to_literal_with_result()
+            .and_then(|row| {
+                let datum = row.unpack_first();
+                if datum.is_null() {
+                    Err(PlanError::ConstantExpressionSimplificationFailed(format!(
+                        "Expected an expression that evaluates to a non-null value, got {}",
+                        self
+                    )))
+                } else {
+                    Ok(datum.unwrap_int64().cast_into())
```
antiguru, line 3559: I think the `cast_into` is unnecessary, `unwrap_int64` returns `i64`.

test/sqllogictest/order_by.slt
ggevay: (Fixed in https://github.com/MaterializeInc/materialize/pull/32447)

src/sql/src/plan/hir.rs
ggevay: Ah, yes, copy-paste fail! Will fix in a follow-up PR.

src/sql/src/plan/hir.rs
ggevay: Indeed, thanks! This was leftover from when this used to be unwrapping a UInt64 and then returning a usize.

PR #32448, opened 2025-05-09
Correct per-crate features

antiguru:
Fixes the crate dependencies such that each crate can be compiled on its own. Before, some crates would rely on features enabled by other crates to compile, which leads to hard-to-explain errors.

This change does not add a regression test, but it would be nice to have one.


petrosagg:
Nice! Do you know why we now have a wasm `Cargo.lock` file but we didn't before?

antiguru:
> Nice! Do you know why we now have a wasm Cargo.lock file but we didn't before?

We had it for a while, but it wasn't updated, so there's a bunch of changes in it now.

petrosagg (approved):

PR #32449, opened 2025-05-09
Remove `bind_parameter` calls from statement types that don't support `PREPARE`

ggevay:
This PR is a little bit of spring cleaning: It removes `bind_parameter` calls and related wiring from the planning of statement types that can't be made into prepared statements (DDL and EXPLAIN TIMESTAMP). Note that Postgres also doesn't support preparing DDL.

However, even for these statements, we still need to check that there are no parameters in them. E.g., if the user writes
```
CREATE VIEW v AS SELECT $1
```
then we need to do _something_ about that `$1`. That something used to be `bind_parameters` complaining that `there is no parameter $1`. After the PR we print a more specific error msg: `views cannot have parameters`.

Additionally, the PR removes the `bind_parameters` fn from `AggregateExpr`, because it was unused, and I can't imagine a use for it.

### Motivation

   * Spring cleaning, but I also have a more specific motivation: I'm planning to tackle https://github.com/MaterializeInc/database-issues/issues/9266, where I'll need to pass more things into `bind_parameter`, so I didn't want to have more calls to it than necessary.

### Checklist

### DIFF
--- a/src/environmentd/tests/pgwire.rs
+++ b/src/environmentd/tests/pgwire.rs
@@ -18,6 +18,7 @@ use bytes::BytesMut;
 use fallible_iterator::FallibleIterator;
 use mz_adapter::session::DEFAULT_DATABASE_NAME;
 use mz_environmentd::test_util::{self, PostgresErrorExt};
+use mz_ore::collections::CollectionExt;
 use mz_ore::retry::Retry;
 use mz_ore::{assert_err, assert_ok};
 use mz_pgrepr::{Numeric, Record};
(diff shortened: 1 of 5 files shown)


src/environmentd/tests/pgwire.rs
```diff
@@ -127,14 +128,62 @@ fn test_bind_params() {
...
-    let err = client
-        .query_one("CREATE VIEW v AS SELECT $3", &[])
-        .unwrap_db_error();
-    // TODO(benesch): this should be `UNDEFINED_PARAMETER`, but blocked
-    // on database-issues#1031.
-    assert_eq!(err.message(), "there is no parameter $3");
-    assert_eq!(err.code(), &SqlState::INTERNAL_ERROR);
+    // Some statement types can't have parameters at all.
+    //
+    // TODO: these should be `SqlState::UNDEFINED_PARAMETER`.
```
ggevay, line 133: I'm planning to submit a quick follow-up PR that fixes this. https://github.com/MaterializeInc/materialize/discussions/5340 is from 2021, and things got better since then.

aljoscha (approved):
looks very reasonable!

src/environmentd/tests/pgwire.rs
ggevay: https://github.com/MaterializeInc/materialize/pull/32457

PR #32451, opened 2025-05-09
Convert more binary funcs to the sqlfunc macro

antiguru:
Includes:
* Decode
* Left
* LikeEscape
* MzAclItemContainsPrivilege
* ParseIdent
* PrettySql
* Right
* StartsWith
* TextConcat
* TimezoneOffset
* Trim{Leading, Trailing, }

As usual, the new functionality is tested, but not yet used.

### DIFF
--- a/src/expr/src/scalar/func.rs
+++ b/src/expr/src/scalar/func.rs
@@ -592,7 +592,7 @@ fn convert_from<'a>(a: Datum<'a>, b: Datum<'a>) -> Result<Datum<'a>, EvalError> 
     }
 }
 
-#[sqlfunc(output_type = "String", sqlname = "encode", propagates_nulls = true)]
+#[sqlfunc(output_type = "String", propagates_nulls = true)]
 fn encode<'a>(
     bytes: Datum<'a>,
     format: Datum<'a>,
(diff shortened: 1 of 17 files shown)


src/expr/src/scalar/func.rs
```diff
@@ -3027,6 +3028,18 @@ fn timezone_interval_timestamptz(a: Datum<'_>, b: Datum<'_>) -> Result<Datum<'st
     }
 }
 
+#[sqlfunc(
+    output_type_expr = "ScalarType::Record {
+                fields: [
+                    (\"abbrev\".into(), ScalarType::String.nullable(false)),
+                    (\"base_utc_offset\".into(), ScalarType::Interval.nullable(false)),
+                    (\"dst_offset\".into(), ScalarType::Interval.nullable(false)),
```
ggevay, line 3036: You could use raw string literals (`r#" ... "#`) instead, so that you don't have to escape the `"`s inside the literal.

ggevay (approved):

src/expr/src/scalar/func.rs
antiguru: Yup, that works! Thanks for the suggestion.

PR #32463, opened 2025-05-10
Automatically cast parameters of `EXECUTE`

ggevay:
This is on top of https://github.com/MaterializeInc/materialize/pull/32457.

I recommend reviewing the commits separately.

This PR makes us try to automatically cast prepared statement parameters that are passed to `EXECUTE`, the same way Postgres does (https://github.com/MaterializeInc/database-issues/issues/9266).

`plan_params` used to check for exact type equality. We now do a `plan_hypothetical_cast` in `plan_params`, then save both the actually passed type and the expected type. Then, `bind_parameters` actually plans the cast. For this, I needed to pass in the `StatementContext` and the `QueryLifetime` to `bind_params`.

### Motivation

  * This PR addresses a Postgres incompatibility: https://github.com/MaterializeInc/database-issues/issues/9266

### Checklist

### DIFF
--- a/src/sql/src/plan.rs
+++ b/src/sql/src/plan.rs
@@ -2042,7 +2042,8 @@ impl TryFrom<ClusterAlterOptionExtracted> for AlterClusterPlanStrategy {
 #[derive(Debug, Clone)]
 pub struct Params {
     pub datums: Row,
-    pub types: Vec<ScalarType>,
+    pub actual_types: Vec<ScalarType>,
+    pub expected_types: Vec<ScalarType>,
 }
 
 impl Params {
@@ -2050,7 +2051,8 @@ impl Params {
     pub fn empty() -> Params {
         Params {
             datums: Row::pack_slice(&[]),
-            types: vec![],
+            actual_types: vec![],
+            expected_types: vec![],
         }
     }
 }
--- a/src/sql/src/plan.rs
+++ b/src/sql/src/plan.rs
@@ -2041,8 +2041,11 @@ impl TryFrom<ClusterAlterOptionExtracted> for AlterClusterPlanStrategy {
 /// A vector of values to which parameter references should be bound.
 #[derive(Debug, Clone)]
 pub struct Params {
+    /// The datums that were provided in the EXECUTE statement.
     pub datums: Row,
-    pub actual_types: Vec<ScalarType>,
+    /// The types of the datums provided in the EXECUTE statement.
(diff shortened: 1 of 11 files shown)


src/sql/src/plan/hir.rs
```diff
@@ -167,24 +167,18 @@ pub enum HirRelationExpr {
...
         /// Number of records to retain.
-        /// It is of ScalarType::Int64. (It's not entirely clear why not UInt64, see below at
-        /// `offset`.)
+        /// It is of ScalarType::Int64.
+        /// (UInt64 would make sense in theory: Then we wouldn't need to manually check
+        /// non-negativity, but would just get this for free when casting to UInt64. However, Int64
+        /// is better for Postgres compat. This is because if there is a $1 here, then when external
+        /// tools `describe` the prepared statement, they discover this type. If what they find
+        /// were UInt64, then they might have trouble calling the prepared statement, because the
+        /// unsigned types are non-standard, and also don't exist even in Postgres.)
```
ggevay, line 176: See https://materializeinc.slack.com/archives/C08A62E0751/p1746884578661789?thread_ts=1746563801.896649&cid=C08A62E0751

src/sql/src/plan.rs
```diff
@@ -2042,15 +2042,17 @@ impl TryFrom<ClusterAlterOptionExtracted> for AlterClusterPlanStrategy {
 #[derive(Debug, Clone)]
 pub struct Params {
     pub datums: Row,
-    pub types: Vec<ScalarType>,
+    pub actual_types: Vec<ScalarType>,
```
ParkMyCar, line 2045: Not blocking, but what do you think about renaming this to `provided_types`? As far as I understand "actual_types" here refer to the types of the datums we were provided? The naming makes sense but I didn't realize the exact context until I saw the changes in `set_portal(...)`

ParkMyCar (approved):
Nice! This all checks out to me, thanks @ggevay!!

src/sql/src/plan.rs
ggevay: `provided_types` would be misunderstandable to mean that the types themselves were directly provided, i.e., a reader might think they refer to type hints (e.g., the `int` in `PREPARE b (int) AS SELECT $1`, which we don't support yet).

How about naming it `execute_types` + adding a comment on top of the field?

ParkMyCar: `execute_types` makes sense to me! Thanks for thinking through this :)

PR #32480, opened 2025-05-13
compute: more logging on dataflow reconciliation failures

teskje:
Reconciliation failures due to incompatible dataflow plans are sometimes hard to debug, so this PR adds a new log that dumps the full plans.

The specific motivation is debugging https://github.com/MaterializeInc/database-issues/issues/9273. If it turns out that the log is too noisy, e.g. because incompatible plans happen a lot in the wild for non-bug reasons, we may decide to remove it again.

### Motivation

  * This PR helps investigating a recognized bug.

Advances https://github.com/MaterializeInc/database-issues/issues/9273

### Checklist

### DIFF
--- a/src/compute/src/server.rs
+++ b/src/compute/src/server.rs
@@ -574,6 +574,17 @@ impl<'w, A: Allocate + 'static> Worker<'w, A> {
                                     new_as_of = ?as_of,
                                     "dataflow reconciliation failed",
                                 );
+
+                                // Dump the full dataflow plans if they are incompatible, to
+                                // simplify debugging hard-to-reproduce reconciliation failures.
+                                if !compatible {
+                                    warn!(
+                                        old = ?old_dataflow,
+                                        new = ?dataflow,
+                                        "incompatible dataflows in reconciliation",
+                                    );
+                                }
+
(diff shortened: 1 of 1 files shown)


src/compute/src/server.rs
```diff
@@ -570,10 +570,19 @@ impl<'w, A: Allocate + 'static> Worker<'w, A> {
                                     ?uncompacted,
                                     ?subscribe_free,
                                     ?dependencies_retained,
-                                    old_as_of = ?old_dataflow.as_of,
-                                    new_as_of = ?as_of,
```
antiguru, line 574: I'd leave the `as_of` printing here, too, because it might be difficult to read it from the next log line if the dataflow plan is too big.

antiguru (approved):
LGTM, but I would not change the existing warning.

src/compute/src/server.rs
teskje: Fair enough! I was thinking we could get the as-ofs from the "creating dataflow" logs so logging them again here is redundant. But the log for the old dataflow might be a couple days old so it might be hard to find it.

PR #32483, opened 2025-05-13
tests: make sqllogictest tests resilient to result ordering

aljoscha:
A side effect of the work on https://github.com/MaterializeInc/database-issues/issues/9180 is that it can change the ordering of results when no ORDER BY is specified. This is fine according to the SQL standard and all databases say you shouldn't rely on any ordering.

Some of our tests do, however, rely on a specific ordering, which is what this PR changes, as preparation for https://github.com/MaterializeInc/database-issues/issues/9180.

### Tips for reviewers

This is a bit tedious to review, but there's no changes to actual database code, so if all the tests still pass I'd say all's good. You still probably want to check my changes to `sqllogictest` around sorting, and spot-check individual test files.

### DIFF
--- a/misc/python/materialize/query_fitness/pick.py
+++ b/misc/python/materialize/query_fitness/pick.py
@@ -41,7 +41,7 @@ def dump_slt(conn: pg8000.Connection, query: str) -> None:
     print(
         f"""
 
-query {colspec}
+query {colspec} rowsort
 {query}
 ----
 9999999999 values hashing to YY
--- a/test/sqllogictest/aclitem.slt
+++ b/test/sqllogictest/aclitem.slt
@@ -287,7 +287,7 @@ SELECT makeaclitem(1, 2, 'CREATE', true)::text
 
 # aclexplode
 
-query T
+query T rowsort
 SELECT aclexplode(ARRAY[makeaclitem(1, 2, 'SELECT, INSERT, DELETE', false), makeaclitem(3, 4, 'USAGE', false)])
 ----
 (4,3,USAGE,f)
--- a/test/sqllogictest/aggregates.slt
+++ b/test/sqllogictest/aggregates.slt
@@ -71,7 +71,7 @@ SELECT a + 1 FROM t GROUP BY a + 1 HAVING sum(b) = 3
 2
 3
 
-query II
+query II rowsort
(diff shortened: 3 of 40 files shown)


def- (approved):
Nightly triggered: https://buildkite.com/materialize/nightly/builds/12004

Tell me if you need any help with finishing this up

ggevay (commented):
Looks ok, but I'm wondering, why not simply make the default be `rowsort` / `valuesort`, to reduce the boilerplate when writing tests? And when somebody does need to check the order of the result rows (which happens only when there is an `ORDER BY`), they could still use `nosort`.

aljoscha:
> Looks ok, but I'm wondering, why not simply make the default be `rowsort` / `valuesort`, to reduce the boilerplate when writing tests? And when somebody does need to check the order of the result rows (which happens only when there is an `ORDER BY`), they could still use an explicit `nosort`.

I didn't want to go down the rabbit-hole of making sure that nosort is correctly applied to tests that need it. Say a test that wants to test ORDER BY would succeed with rowsort as well, but that's not what the intention of the test is. So this explicit way seemed safer.

ggevay:
Ok, we can just merge this now to unblock you, and we can flip the default around at some other time.

Btw., is this needed due to a one-time change in result ordering due to the large SELECT work, or the large SELECT work can make result ordering non-deterministic in the sense that even on the exact same Materialize version different test runs can sometimes have different orderings? Asking, because if the latter, then we should definitely flip the default around soon, because otherwise people will add new tests where they will forget to apply `rowsort`, and then those tests will be flaky.

ggevay (approved):

aljoscha:
Thanks!

Regarding one-time vs. arbitrary: it's mixed, the ordering when using this new "large SELECT" machinery is deterministic, it happens to sort by the mz_repr ordering of all columns. But now the ordering depends on whether we use the inline compute response machinery or the large response machinery. So if the result sizes of your queries are stable and we don't change the threshold for when we use the large result machinery, you should be good.

I think in the long run it's probably good to be resilient to result-ordering changes, though, as you say.

ggevay:
Ok, created an issue for flipping the default: https://github.com/MaterializeInc/database-issues/issues/9283

PR #27239, opened 2024-05-22
design doc: MIR typechecking using representation types

mgree:
### Motivation

- This PR adds a feature that has not yet been specified.

Implicit casts (`varchar_to_text`) can impede join planning; by having MIR use a type system that corresponds to the representation we'll actually use, we can type-safely elide implicit casts.

PR #32486, opened 2025-05-14
Fix types in `reduce_list_create_list_index_literal`

ggevay:
Fixes a bug that @aljoscha [ran into](https://materializeinc.slack.com/archives/C08ACQNGSQK/p1747168112195079?thread_ts=1747159620.609509&cid=C08ACQNGSQK): When `reduce_list_create_list_index_literal` removes a literal index, it removes one layer of `ListCreate`s by partial evaluation, and thus all the outer `ListCreate`s lose one list layer, but this was not reflected in their types.

### Motivation

### Checklist

### DIFF
--- a/src/expr/src/scalar.rs
+++ b/src/expr/src/scalar.rs
@@ -1537,9 +1537,13 @@ impl MirScalarExpr {
             // element of `list_create_mut_refs` is not actually a ListCreate, then we break out of
             // the loop. When we remove a literal, we need to partial-evaluate all ListCreates
             // that are at the current level (except those that disappeared due to
-            // literals at earlier levels), and change each element in `list_create_mut_refs`
-            // to the result of the partial evaluation.
+            // literals at earlier levels), index into them with the literal, and change each
+            // element in `list_create_mut_refs` to the result.
+            // We also record mut refs to all the earlier `element_type` references that we have
+            // seen in ListCreate calls, because when we process a literal index, we need to remove
+            // one layer of list type from all these earlier ListCreate `element_type`s.
             let mut list_create_mut_refs = vec![&mut list_create_to_reduce];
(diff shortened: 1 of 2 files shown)


src/expr/src/scalar.rs
```diff
@@ -1573,6 +1577,29 @@ impl MirScalarExpr {
...
                     }
+                    // Peel one layer off of each of the earlier element types.
+                    for t in earlier_list_create_types.iter_mut() {
+                        if let ScalarType::List {
+                            element_type,
+                            custom_id: _,
+                        } = t
+                        {
+                            **t = *element_type.clone();
+                            // These are not the same types anymore, so remove custom_ids all the
```
mgree, line 1588: I don't follow this logic (but I also don't know exactly how `custom_id`s affect how we understand/use a type.

In the abstract, `[custom(a), custom(b), custom(c)][2]` should have `list[custom]` as its type, and after reducing we should have `custom(b)` with `custom` as the type... no?

mgree (approved):
Looks good to me, though I don't understand the custom type reasoning.

src/expr/src/scalar.rs
ggevay: You are right, we might need more complicated logic here. But I'd like to fix this in a follow-up PR. The current PR definitely makes the types less wrong than they are on `main`, and it seems an exotic special case when there are custom types involved.

(Also, it might be not incorrect to even just erase _all custom ids everywhere in MIR_, similarly to how we want to erase differences among various string types in the representation types work. This is because the actual Datums that we produce in `ListCreate` are not affected by the custom ids, as far as I can tell)

mgree: Oh my. 🫣

This is very reasonable, but let's make sure we actually sort out the story. Perhaps another argument for a separate notion of "representation type" https://github.com/MaterializeInc/materialize/pull/27239.

PR #32487, opened 2025-05-14
optimizer: Typecheck also in local optimization

ggevay:
This PR makes us do typechecking also in local optimization. Not doing this earlier was the reason that we did not catch https://github.com/MaterializeInc/materialize/pull/32486.

(Normally, `optimize_dataflow_relations` sets the `GlobalId` in the `TransformCtx`, but `optimize_dataflow_relations` is not involved in the local optimization.)

This should now produce errors in `list.slt` (because https://github.com/MaterializeInc/materialize/pull/32486 is not fixed on this branch), but I'm not sure if this will fail CI. Running CI with the PR in draft to find out. I'm also running all slts locally with an explicit panic instead of the error, and with https://github.com/MaterializeInc/materialize/pull/32486 cherry-picked, to see if any other type errors show up due to doing typechecking in more places now.

Edit: There are also type errors in `src/transform/tests/test_runner.rs`! Need to investigate! Edit 2: Found it, will fix it up after the meeting.

### Checklist

### DIFF
--- a/src/transform/src/lib.rs
+++ b/src/transform/src/lib.rs
@@ -144,11 +144,12 @@ impl<'a> TransformCtx<'a> {
         typecheck_ctx: &'a SharedContext,
         df_meta: &'a mut DataflowMetainfo,
         metrics: Option<&'a OptimizerMetrics>,
+        global_id: Option<GlobalId>,
     ) -> Self {
         Self {
             indexes: &EmptyIndexOracle,
             stats: &EmptyStatisticsOracle,
-            global_id: None,
+            global_id,
             features,
             typecheck_ctx,
             df_meta,
--- a/src/transform/src/lib.rs
+++ b/src/transform/src/lib.rs
@@ -186,7 +186,8 @@ impl<'a> TransformCtx<'a> {
         Arc::clone(self.typecheck_ctx)
     }
 
-    fn set_global_id(&mut self, global_id: GlobalId) {
+    /// Lets self know the id of the object that is being optimized.
+    pub fn set_global_id(&mut self, global_id: GlobalId) {
         self.global_id = Some(global_id);
     }
 
--- a/src/transform/tests/test_runner.rs
+++ b/src/transform/tests/test_runner.rs
@@ -48,27 +48,31 @@ mod tests {
 
     const TEST_GLOBAL_ID: GlobalId = GlobalId::Transient(1234567);
 
-    thread_local! {
-        static FULL_TRANSFORM_LIST: Vec<Box<dyn Transform>> = {
-            let features = OptimizerFeatures::default();
-            let typecheck_ctx = typecheck::empty_context();
-            let mut df_meta = DataflowMetainfo::default();
-            let mut transform_ctx = TransformCtx::local(&features, &typecheck_ctx, &mut df_meta, None, Some(TEST_GLOBAL_ID));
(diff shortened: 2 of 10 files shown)


ggevay:
CI is green, so I'm confused. Do we not fail CI even on the more serious kind of Typechecking error?

src/transform/tests/test_runner.rs
```diff
@@ -48,27 +48,39 @@ mod tests {
 
     const TEST_GLOBAL_ID: GlobalId = GlobalId::Transient(1234567);
 
-    thread_local! {
-        static FULL_TRANSFORM_LIST: Vec<Box<dyn Transform>> = {
-            let features = OptimizerFeatures::default();
-            let typecheck_ctx = typecheck::empty_context();
```
ggevay, line 52: The testing bug was that this `typecheck_ctx` used to get shared among different test runs, so Typecheck runs from different test runs stepped on each other's toes through the expected types.

ggevay: (This test framework is on its way to be deprecated, so don't think too hard about this code.)

src/transform/src/lib.rs
```diff
@@ -185,7 +186,8 @@ impl<'a> TransformCtx<'a> {
         Arc::clone(self.typecheck_ctx)
     }
 
-    fn set_global_id(&mut self, global_id: GlobalId) {
+    /// Lets self know the id of the object that is being optimized.
```
mgree, line 189: Since `global_id` is `pub`... why not just write the `global_id` in directly?

mgree (approved):
To confirm my understanding: we weren't setting a `global_id` for these `optimize_mir_local` calls, which meant we weren't ever checking that types were consistent pass to pass.

Looks good to me!

src/transform/tests/test_runner.rs
```diff
@@ -46,27 +46,41 @@ mod tests {
     const JSON: &str = "json";
     const TEST: &str = "test";
 
-    thread_local! {
-        static FULL_TRANSFORM_LIST: Vec<Box<dyn Transform>> = {
-            let features = OptimizerFeatures::default();
-            let typecheck_ctx = typecheck::empty_context();
-            let mut df_meta = DataflowMetainfo::default();
-            let mut transform_ctx = TransformCtx::local(&features, &typecheck_ctx, &mut df_meta, None);
+    const TEST_GLOBAL_ID: GlobalId = GlobalId::Transient(1234567);
```
mgree, line 49: I don't think we'll ever get up to 12 million built-in things, but would it be possible to dynamically allocate this rather than hardcoding something? (Or adding an `assert!` early on in this testing function that this global ID isn't already in use?)

ggevay: I think hardcoding this in tests is beneficial, because then you can grep for this specific number when you see it in an error msg.

And I'm not afraid of id clashes, because having a million global objects would put a crazy load on envd, so this is not possible even in a test.

ggevay:
> To confirm my understanding: we weren't setting a `global_id` for these `optimize_mir_local` calls, which meant we weren't ever checking that types were consistent pass to pass.

Yes, during local optimization, we were doing only that part of typechecking that is internal to an expression, but we were never comparing the final type of the expression between different parts of the local pipeline.

PR #32493, opened 2025-05-14
Docs ingest data considerations

kay-kim:
single-sourced + incorporated into the various ingest data pages to raise awareness instead of having our users find out after-the-fact. Some examples of where the info is now incorporated:
- https://preview.materialize.com/materialize/32493/ingest-data/mysql/#considerations
- https://preview.materialize.com/materialize/32493/ingest-data/postgres/cloud-sql/#considerations


nstewart (approved):
LGTM

PR #32496, opened 2025-05-15
Add pprof profile offline symbolization tool

SangJunBak:

### Motivation
Followup of https://github.com/MaterializeInc/materialize/pull/32423 and https://github.com/MaterializeInc/database-issues/issues/8908 where the debug tool scrapes the profiles and us, internally, can run this python script internally to spin up a docker container with a UI similar to pprof.me. 

To easily test:
```bash
# Start the emulator with internal http port exposed 
docker rm materialized && docker run -d --name materialized -p
 6878:6878 materialize/materialized:latest

# Download the heap profile
curl -o envd.pprof.gz http://localhost:6878/prof/heap
# login 
aws sso login --profile mz-cloud-production-engineering-on-call
export AWS_PROFILE=mz-cloud-production-engineering-on-call

# Run the script and wait for "Serving web UI on http://localhost:8080"
bin/pyactivate -m materialize.visualize_pprof_profile.visualize_pprof_profile envd.pprof.gz


```

Open to suggestions given it's my first time writing a script like this!

Note: We still need to implement CPU profiling in an unsymbolized format then backport it into LTS. We only have memory profiles.

We also have to figure out why the latest binary for the emulator is like over a gigabyte if they're supposed to be stripped 😅 

### Checklist


ci/builder/requirements.txt
```diff
@@ -36,6 +36,7 @@ pdoc==15.0.3
 # We can revert back to standard pg8000 versions once https://github.com/tlocke/pg8000/pull/161 is released
 pg8000@git+https://github.com/tlocke/pg8000@46c00021ade1d19466b07ed30392386c5f0a6b8e
 prettytable==3.16.0
+protobuf==5.29.3
```
SangJunBak, line 39: Specifically used this version since that's the version used in profile.proto

misc/python/materialize/visualize_pprof_profile/README.md
```diff
@@ -0,0 +1,53 @@
...
+- The web UI will be available at `http://localhost:<port>` (default: http://localhost:8080)
+- Initial symbolization might take a few moments - wait until you see "Serving web UI on http://localhost:8080" message
+- The Docker container continues running even after you quit the program  
+- The profile.proto file is sourced from: https://raw.githubusercontent.com/google/pprof/main/proto/profile.proto
+
+## Example
+
+```bash
+# Set up AWS credentials
+export AWS_PROFILE=mz-cloud-production-engineering-on-call
```
SangJunBak, line 47: So for me, this is the only AWS profile with access to that bucket. For some reason, `mz-core-admin` doesn't work. Let me know if I should change it but I assume this is the same for most and although it's kinda suspicious using an overprivileged profile, I'd rather have one that works than not

misc/python/materialize/visualize_pprof_profile/profile.proto
```diff
@@ -0,0 +1,233 @@
+// Copyright 2016 Google Inc. All Rights Reserved.
```
SangJunBak, line 1: As said in pprof, this is received from a CURL. Not sure if we need to do something to account for this apache license.

A potential worry is the `pprof.gz` provided doesn't match this proto, but given protobufs are supposed to be backwards compatible and because this `.proto` file (one used for pprof files) is most likely stable, I thought it'd be okay

SangJunBak: Noticing this is failing the check-copyright lint test. I can fix this via changing the header to 
```
// Copyright Materialize, Inc. and contributors. All rights reserved.
//
// Use of this software is governed by the Business Source License
// included in the LICENSE file.
//
// As of the Change Date specified in that file, in accordance with
```
but this seems quite suspicious to just override a copyright header like that. Any advice?

ci/builder/requirements.txt
def-: This will be auto-upgraded by dependabot. Is that a problem?

misc/python/materialize/visualize_pprof_profile/README.md
def-: I guess some devs won't have access to this profile. We should probably make the S3 bucket more available.

misc/python/materialize/visualize_pprof_profile/profile.proto
def-: I have added an ignore for the copyright check in the linters.

ci/deploy_mz-debug/README.md
```diff
@@ -22,13 +22,15 @@ You can manually deploy by following steps 1-2 above and running the following c
...
 export BUILDKITE_TAG=mz-debug-vx.y.z
+export AWS_PROFILE=...
 
 # macOS
 bin/pyactivate -m ci.deploy_mz-debug.macos
 
 # Linux
 bin/pyactivate -m ci.deploy_mz-debug.linux
 ```
+where AWS_PROFILE is the profile with access to the materialize-binaries S3 bucket in the Materialize Core account.
```
def-, line 33: I don't think devs have access to the Materialize Core account normally? At least I don't.

def-: What I always used to use is https://debuginfo.dev.materialize.com/buildid/d6a8b86c62ce2b7a6fd146a048ebba77/executable, which is publicly available. I think that's easier instead of using S3.

misc/python/materialize/visualize_pprof_profile/visualize_pprof_profile.py
```diff
@@ -0,0 +1,195 @@
...
+            # We need this specific path and file format in order for pprof to read the binary/debug info
+            (binary_path, f"/tmp/{build_id}/binary"),
+            (debug_info_path, "/tmp/binary.debug"),
+        ]:
+            subprocess.run(
+                ["docker", "cp", str(path), f"{container_name}:{dest}"], check=True
+            )
+
+        # Install graphviz in the container
+        subprocess.run(
```
def-, line 128: Shouldn't this just be a Dockerfile?

def- (commented):
Maybe I'm a bit confused of what the purpose of this is. Is it just doing the same as the instructions in "Analyzing pprof for a release build" section of https://www.notion.so/materialize/analyzing-pprof-for-a-release-build-3fa5a68aef994d90b3c94bca6eea4da8 ? That always worked well for me, I'm not sure if we need another tool for it.

SangJunBak:
On second thought: I think a good meet in the middle solution would be to generate a bash script per profile in the debug tool and expect users (us when debugging internally) to have pprof installed. That way users can just run the script and don't have to look up a guide.

def-:
One thing that would be nice in the script is to be able to pass all kinds of parameters to pprof that it supports.

PR #32504, opened 2025-05-15
Design doc on extending the query lifecycle tables

ggevay:
This just mostly copies over the things from https://github.com/MaterializeInc/database-issues/issues/9140#issuecomment-2876868717
plus fills in some sections of the design doc template.

### Motivation

* Design doc for https://github.com/MaterializeInc/database-issues/issues/9140#issuecomment-2876868717, so that we can have visibility into end-to-end query latency at, e.g., https://github.com/MaterializeInc/accounts/issues/3

### Checklist


ParkMyCar (commented):
Thanks for working through this @ggevay!

As far as I understand, as proposed we would have 4 different tables:

1. `mz_statement_lifecycle_history` - tracks the lifecycle of a statement from receiving the bytes to parsing to eventually binding and executing if the extended query protocol is used. You can have multiple events here for a single `statement_id`.
2. `mz_prepared_statement_history` - Tracks all of the prepared statements, the primary key of this relation would be a "statement_id". (essentially this is a rollup of `mz_statement_lifecycle_history`?)
3. `mz_statement_execution_lifecycle_history` - tracks the lifecycle of a statement as it's being executed. You can have multiple events here for a single `execution_id`.
4. `mz_statement_execution_history` - Tracks every instance of a statement execution, the primary key of this relation would be an "execution_id" and it would join against a "statement_id" from `mz_prepared_statement_history`. (essentially this is a rollup of `mz_statement_execution_lifecycle_history`?)

This this is the case, it makes sense to me! Although I propose a slightly different naming scheme:

1. `mz_statement_preparation_lifecycle_history`
2. `mz_statement_preparation_rollup_history`
3. `mz_statement_execution_lifecycle_history`
4. `mz_statement_execution_rollup_history`

I know the `_rollup_` tables aren't exactly rollups because they also include extra metadata, but conceptually that seems to be the purpose for the tables. What do you think?

doc/developer/design/20250515_query_lifecycle_extensions.md
```diff
@@ -0,0 +1,125 @@
...
+
+Note that if a user is interested only in queries submitted through 1. or 2., and they are not interested in when the parsing and other preparation ended, it will be enough to look at `mz_statement_execution_lifecycle_history` (and optionally join it with `mz_statement_execution_history` and `mz_prepared_statement_history` to get more info about the execution and/or statement). They will still be able to get the end-to-end time, because `ExecutionMessageReceived` will be part of `mz_statement_execution_lifecycle_history`.
+
+If a user is interested also in prepared statements, i.e., 3. or 4, then they would need to look at also `mz_statement_lifecycle_history. In this case, a 4-way join between `mz_statement_lifecycle_history`, `mz_statement_execution_lifecycle_history`, `mz_statement_execution_history`, and `mz_prepared_statement_history` would be needed.
+
+Also, I'm thinking to add a column to `mz_prepared_statement_history` to indicate which of 1., 2., 3., or 4. was used to submit the statement.
+
+## Examples
+
+For a query submitted through 1.:
```
ParkMyCar, line 75: Not a huge deal, but small preference to using either the `Complete` or `Finished` suffix and either the `Began` or `Started` suffix for events, but not instances of both.

doc/developer/design/20250515_query_lifecycle_extensions.md
```diff
@@ -0,0 +1,125 @@
...
+        - `ParsingComplete`
+        - `PrepareComplete`
+- `FrontendMessage::Bind` arrives in `protocol.rs`, to which
+    - `mz_statement_lifecycle_history` would get the following events:
+        - `BindReceived`
+        - `BindComplete` (whose timestamp would probably be very close to `BindReceived`, but later we might do more interesting things at bind time, in which case binding will take more time)
+- `FrontendMessage::Execute` arrives in `protocol.rs`, to which
+    - `mz_statement_lifecycle_history` would get the following events:
+        - `ExecuteReceived`
+        - `ExecuteFinished`
```
ParkMyCar, line 103: I'm a little thrown off by recording these stages so explicitly under the assumption they may change in the future. But given this just mirrors the Postgres wire protocol which will never change, I think the more info we have the better.

doc/developer/design/20250515_query_lifecycle_extensions.md
```diff
@@ -0,0 +1,125 @@
...
+    - Additionally, we should have ...received and ...finished events for each of the other msgs in the "Extended Query" flow, e.g., `Bind`, `Describe`, `Execute`.
+
+Note that if a user is interested only in queries submitted through 1. or 2., and they are not interested in when the parsing and other preparation ended, it will be enough to look at `mz_statement_execution_lifecycle_history` (and optionally join it with `mz_statement_execution_history` and `mz_prepared_statement_history` to get more info about the execution and/or statement). They will still be able to get the end-to-end time, because `ExecutionMessageReceived` will be part of `mz_statement_execution_lifecycle_history`.
+
+If a user is interested also in prepared statements, i.e., 3. or 4, then they would need to look at also `mz_statement_lifecycle_history. In this case, a 4-way join between `mz_statement_lifecycle_history`, `mz_statement_execution_lifecycle_history`, `mz_statement_execution_history`, and `mz_prepared_statement_history` would be needed.
+
+Also, I'm thinking to add a column to `mz_prepared_statement_history` to indicate which of 1., 2., 3., or 4. was used to submit the statement.
+
+## Examples
```
ParkMyCar, line 73: Thanks for writing out these examples! Helped me piece together how everything would fit together!

doc/developer/design/20250515_query_lifecycle_extensions.md
```diff
@@ -0,0 +1,125 @@
+# Extending the Query Lifecycle Tables
+
+- Associated: [#9140](https://github.com/MaterializeInc/database-issues/issues/9140)
```
ParkMyCar, line 3: Can you also link the original [statement logging doc](https://github.com/MaterializeInc/materialize/blob/4487749289f1591e7918ebaa1458e23b5e40633f/doc/developer/design/20230519_statement_logging.md) and the [query lifecycle doc](https://github.com/MaterializeInc/materialize/blob/4487749289f1591e7918ebaa1458e23b5e40633f/doc/developer/design/20231204_query_lifecycle_events_logging.md)?

doc/developer/design/20250515_query_lifecycle_extensions.md
ggevay: Indeed, it's better to be consistent. Fixed!

doc/developer/design/20250515_query_lifecycle_extensions.md
ggevay: Yes, this is intended for the long term.

doc/developer/design/20250515_query_lifecycle_extensions.md
ggevay: Done

ParkMyCar:
> I like your new names! They are more structured. This will mean slightly more renaming of existing tables than my original proposal, but I think it's worth it to make it easier to understand these 4 tables.

Great! Yeah definitely is a bit more work, but thank you for taking it on. As-is I get lost with how all of the existing tables are connected 😅

> The only thing is `rollup` sounds a bit out of place for me. "Rollup" would suggest to me that it's the same info as in the other table but just grouped in some way, when in reality it has extra metadata as you pointed out. What do you think about saying actually `metadata` instead of `rolloup` in the table names?

"metadata" works for me, what actually what do you think about `_log` as the entire suffix? So we'd have:

1. `mz_statement_preparation_lifecycle_history`
2. `mz_statement_preparation_log`
3. `mz_statement_execution_lifecycle_history`
4. `mz_statement_execution_log`

This is definitely getting into bikeshed territory and mostly what I'm concerned about is the new structure. So whatever suffixes you decide to use work for me :)

ParkMyCar (approved):

doc/developer/design/20250515_query_lifecycle_extensions.md
```diff
@@ -0,0 +1,128 @@
...
+        - For 4., when we are about to send a `ParseFinished` back to the user in `protocol.rs`.
+    - Additionally, we should have ...received and ...finished events for each of the other msgs in the "Extended Query" flow, e.g., `Bind`, `Describe`, `Execute`.
+
+Note that if a user is interested only in queries submitted through 1. or 2., and they are not interested in when the parsing and other preparation ended, it will be enough to look at `mz_statement_execution_lifecycle_history` (and optionally join it with `mz_statement_execution_history` and `mz_prepared_statement_history` to get more info about the execution and/or statement). They will still be able to get the end-to-end time, because `ExecutionMessageReceived` will be part of `mz_statement_execution_lifecycle_history`.
+
+If a user is interested also in prepared statements, i.e., 3. or 4, then they would need to look at also `mz_statement_lifecycle_history. In this case, a 4-way join between `mz_statement_lifecycle_history`, `mz_statement_execution_lifecycle_history`, `mz_statement_execution_history`, and `mz_prepared_statement_history` would be needed.
```
teskje, line 72: Have you considered not having the additional `mz_statement_lifecycle_history` table and instead denormalizing  the prepared statement events into `mz_statement_execution_lifecycle_history`? It seems like this would simplify the implementation, reduce the catalog surface, and make getting at the data easier for the Console. At least I think the Console wants to show timings for statement executions only and not for prepared statements?

A benefit of normalizing would be that safe some space by deduplicating events when the same prepared statement is executed a large amount of times. Do we have any way of knowing how frequently that happens?

doc/developer/design/20250515_query_lifecycle_extensions.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Minimal Viable Prototype
+
+I have a limited prototype that doesn't add new builtin tables or change schemas in existing ones, but just adds events to the existing `mz_statement_lifecycle_history` without regard to prepared statements:
+https://github.com/MaterializeInc/materialize/pull/32424
+The prototype shows how we can transfer information between `protocol.rs` and the already existing lifecycle logging code in the `Coordinator` through portals.
+
+## Alternatives
+
+One could imagine trying to keep things simple by just quickly implementing the events only for statements that are not prepared statements. however, it turns out that [a big user](https://github.com/MaterializeInc/accounts/issues/3) is actually using prepared statements, because PowerBI always uses the "Extended Query" flow of pgwire. And since having visibility into that user's PowerBI query latency is one of the main motivations of this work, we should do the right thing and support prepared statements.
```
teskje, line 124: Another alternative is to consider replacing the entire statement logging mechanism. The current one has proven to not scale well and is causing incidents from time to time, so we'll have to rethink it eventually. This will be a major piece of work, possibly involving the integration of some external component, so I can understand if you consider this out of scope. But there is a real risk that the query lifecycle will run into the same scalability issues the statement log experiences once we start leaning more on it.

PR #32509, opened 2025-05-15
#9275 refactor to batch compare and append for one shard in persist sink

ptravers:
### Motivation

https://github.com/MaterializeInc/database-issues/issues/9275

### Checklist

### DIFF
--- a/misc/python/materialize/mzcompose/__init__.py
+++ b/misc/python/materialize/mzcompose/__init__.py
@@ -311,7 +311,7 @@ def get_variable_system_parameters(
         ),
         VariableSystemParameter("storage_reclock_to_latest", "true", ["true", "false"]),
         VariableSystemParameter(
-            "storage_sink_write_batches_to_persist", "true", ["true", "false"]
+            "storage_sink_write_batches_to_persist", "false", ["true", "false"]
         ),
         VariableSystemParameter(
             "storage_source_decode_fuel",


src/persist-client/src/batch.rs
```diff
@@ -1294,17 +1294,6 @@ pub(crate) fn validate_truncate_batch<T: Timestamp>(
...
-    if !PartialOrder::less_equal(batch.lower(), truncate.lower())
-        || PartialOrder::less_than(batch.upper(), truncate.upper())
-    {
-        return Err(InvalidUsage::InvalidBatchBounds {
-            batch_lower: batch.lower().clone(),
-            batch_upper: batch.upper().clone(),
-            append_lower: truncate.lower().clone(),
-            append_upper: truncate.upper().clone(),
-        });
-    }
```
ptravers, line 1307: we can retain this check by instead verifying that the last batch in the set of batches has the same upper as the upper we are attempting to migrate to

ptravers:
existing implementation:

src/storage/src/render/persist_sink.rs
```diff
@@ -1360,6 +1358,8 @@ where
                             );
                             anyhow::bail!("collection concurrently modified. Ingestion dataflow will be restarted");
                         }
+
+                        break;
```
ptravers, line 1362: I am concerned that none of the above behaviour may ever be actuated by a test.

src/ore/src/collections/hash.rs
```diff
@@ -96,6 +96,12 @@ impl<K, V> HashMap<K, V> {
     pub fn clear(&mut self) {
         self.0.clear();
     }
+
+    /// See [`std::collections::HashMap::drain`].
+    #[inline]
+    pub fn drain(&mut self) -> Drain<'_, K, V> {
+        self.0.drain()
+    }
```
bkirwi: Can't do it! The idea of this wrapper is that we only expose methods that don't depend on iteration order, but this does.

src/persist-client/src/batch.rs
```diff
@@ -1294,17 +1294,19 @@ pub(crate) fn validate_truncate_batch<T: Timestamp>(
...
+    //if !PartialOrder::less_equal(batch.lower(), truncate.lower())
+    //    || PartialOrder::less_than(batch.upper(), truncate.upper())
+    //{
+    //    return Err(InvalidUsage::InvalidBatchBounds {
+    //        batch_lower: batch.lower().clone(),
+    //        batch_upper: batch.upper().clone(),
+    //        append_lower: truncate.lower().clone(),
+    //        append_upper: truncate.upper().clone(),
+    //    });
+    //}
```
bkirwi: We don't typically commit commented-out code; if we're deleting, just a comment is fine.

src/persist-client/src/write.rs
```diff
@@ -453,6 +453,13 @@ where
     ///
     /// The clunky multi-level Result is to enable more obvious error handling
     /// in the caller. See <http://sled.rs/errors.html> for details.
+
+    // NB(ptravers): we have explicitly disabled checking the expected upper and new upper
+    // as we wish to allow batches that are a mixture of empty frontier moving
+    // updates and batches of actual data. As a result we do not check that the
+    // submitted batches fit within the stated boundaries as persist does not know
+    // in the case of a pure frontier movement if there is a batch that fills a given
+    // time span.
```
bkirwi: Is this the reason? My understanding is that we wanted to be able to commit eg. a batch that goes from [A, B) and a batch that goes from [B, C) as part of a single append from [A, C)... and that doesn't necessarily involve empty batches at all.

src/storage/src/render/persist_sink.rs
```diff
@@ -1101,39 +1103,42 @@ where
             // `upper` line up.
```
bkirwi: This comment is no longer accurate, since the `reverse` is removed.

bkirwi: Though the whole process of reverse-sorting, then using the first element as the upper bound and the last element as the upper bound (if there is one) seems baroque... if we sorted normally, can't we use `done_batches.first()` and `last()` and skip the iterator stuff?

src/storage/src/render/persist_sink.rs
```diff
@@ -1101,39 +1103,42 @@ where
...
-                );
+            let mut batch_lower = if let Some(last_batch_description) = done_batches_iter.last() {
+                last_batch_description.0.clone()
+            } else {
+                first_batch_description.0.clone()
+            };
 
-                let (batch_lower, batch_upper) = done_batch_metadata;
+            let mut batches: Vec<FinishedBatch> = vec![];
+            let mut batch_metrics: Vec<BatchMetrics> = vec![];
```
bkirwi:
```suggestion
            let mut batch_metrics: BatchMetrics = BatchMetrics::default();
```

It looks like the only thing we do with this Vec is reduce it later... might as well just add up the metrics in the loop and save the allocation / complexity?

src/storage/src/render/persist_sink.rs
```diff
@@ -1271,21 +1275,27 @@ where
...
-                        metrics.row_inserts.inc_by(batch_metrics.inserts);
-                        metrics.row_retractions.inc_by(batch_metrics.retractions);
-                        metrics.error_inserts.inc_by(batch_metrics.error_inserts);
-                        metrics
-                            .error_retractions
-                            .inc_by(batch_metrics.error_retractions);
+                        if let Some(batch_metrics) = batch_metrics.drain(..).reduce(|mut l, r| {l.add_assign(&r); l}) {
+                            source_statistics
+                                .inc_updates_committed_by(batch_metrics.inserts + batch_metrics.retractions);
+                            // FIXME(ptravers): we should increment by the total number of batches in the stored batch
```
bkirwi: The idea is to fix this before merging, yeah?

src/storage/src/render/persist_sink.rs
```diff
@@ -1271,21 +1275,27 @@ where
...
+                            metrics.processed_batches.inc();
+                            metrics.row_inserts.inc_by(batch_metrics.inserts);
+                            metrics.row_retractions.inc_by(batch_metrics.retractions);
+                            metrics.error_inserts.inc_by(batch_metrics.error_inserts);
+                            metrics
+                                .error_retractions
+                                .inc_by(batch_metrics.error_retractions);
+                        }
 
+                        to_append.clear();
```
bkirwi: This is a noop, right? (This vec goes out of scope immediately anyways?)

src/storage/src/render/persist_sink.rs
bkirwi: In theory multi-replica sources should trigger this, at least sometimes! Does seem like a good thing to have test coverage on before rolling this out, esp. since we're removing a safety rail as part of this change...

src/storage/src/render/persist_sink.rs
```diff
@@ -1313,40 +1323,28 @@ where
...
-                                } else {
-                                    batch_delete_futures.push(batch.batch.delete());
-                                }
+                            for batch in batches.drain_filter_swapping(|batch| !mismatch.current.less_equal(&batch.data_ts)) {
+                                batch_delete_futures.push(batch.batch.delete());
                             }
 
+                            batch_lower = mismatch.current.clone();
+                            to_append = batches.iter_mut()
+                                .sorted_by(|l, r| Ord::cmp(&l.data_ts, &r.data_ts))
```
bkirwi: Why does the sorting matter here? (We don't seem to do it on the first attempt...)

src/ore/src/collections/hash.rs
bkirwi: OTOH seems like this might be dead code?

src/storage/src/render/persist_sink.rs
ptravers: yes

src/storage/src/render/persist_sink.rs
ptravers: good, point. I will run tests without the clear

src/storage/src/render/persist_sink.rs
ptravers: true, will refactor.

src/persist-client/src/batch.rs
ptravers: gotcha, will delete.

src/storage/src/render/persist_sink.rs
ptravers: okay, cool. I will look at a test for that.

src/ore/src/collections/hash.rs
ptravers: ah, okay. if it isn't dead I will make it so.

src/storage/src/render/persist_sink.rs
ptravers: I changed the `sort_by` to do a reverse so I think it should still hold? but I will look at refactoring today.

src/storage/src/render/persist_sink.rs
ptravers: taking a second look!

src/persist-client/src/write.rs
ptravers: that is correct. I am referencing the removal of the constraint that was checking the batches matched the new upper. since when the batches were empty previously we would not actually perform the validation check. but if you blend empty and non empty batches we now start to perform the check. I only saw the validation constraint fail when [A,B) contained a non empty batch set and [B,C) had an empty batch set.

src/storage/src/render/persist_sink.rs
ptravers: I think I added it since I assumed that filter swapping would jumble the batch.

src/persist-client/src/write.rs
bkirwi: Hmm... if neither [A, B) or [B, C) was empty, I would still expect that trying to batch them up in one big append to fail. (Since we run the check per-batch.) It makes sense that if they were both empty the validation would succeed. (Since there's nothing to check.)

Is the idea that you think that one of the ranges needs to be empty for the validation to fail, or that just happens to be the pattern that sources usually generate?

ptravers: ah sure, for context, I had just tried to write a validation that took the lowest lower and highest upper then compared them against the submitted new_upper and previous_upper. this did not work due to the issue of empty batches.

you are correct that whenever there was any data + multiple batches we would see failure.

ptravers: but also that was just the main pattern I saw in our end to end tests.

ptravers: happy to update the comment to the above description. I was thinking about code that you never saw that made me decide that it is impossible to enforce any constraint at all.

src/storage/src/render/persist_sink.rs
```diff
@@ -1125,13 +1125,14 @@ where
             };
 
             let mut batches: Vec<FinishedBatch> = vec![];
-            let mut batch_metrics: Vec<BatchMetrics> = vec![];
+            let mut batch_metrics: BatchMetrics = BatchMetrics::default();
+            let num_batches = done_batches.len().try_into();
```
ptravers: is there a better way to handle conversion to a u64?

src/storage/src/render/persist_sink.rs
ptravers: removed.

src/storage/src/render/persist_sink.rs
ptravers: fixed!

src/storage/src/render/persist_sink.rs
ptravers: refactored.

src/storage/src/render/persist_sink.rs
bkirwi: `u64::cast_from` is probably the cleanest - since we only build for 64 bit it's guaranteed lossless.

ptravers: fixed

ptravers:
nightly run for [787b0e6a2b](https://github.com/MaterializeInc/materialize/commit/787b0e6a2b3737a8f78f5f4adec89dd0fa9c7579): https://buildkite.com/materialize/nightly/builds/12218

ptravers:
nightly build: https://buildkite.com/materialize/nightly/builds/12226

src/storage/src/render/persist_sink.rs
```diff
@@ -1083,57 +1083,55 @@ where
...
-                let mut batches = batch_set.finished;
-
-                trace!(
-                    "persist_sink {collection_id}/{shard_id}: \
-                        done batch: {:?}, {:?}",
-                    done_batch_metadata,
-                    batches
-                );
+                // SAFETY(ptravers): if first exists so must last.
+                (Some(_), None) | (None, Some(_)) => unreachable!(),
```
ptravers:
```suggestion
                (Some(_), None) | (None, Some(_)) => unreachable!("the list of batches must have a first and last if it is non empty."),
```

src/storage/src/render/persist_sink.rs
ptravers: here's our multi replica tests. currently running them with a nightly build.

https://github.com/MaterializeInc/materialize/pull/32188/files

bkirwi:
Thanks for all the followup! Persist-side stuff looks good to me.

It would be good to get some sources-side signoff for this. (For example, I think this will now retry in some cases where it previously would have bailed, when `bail_on_concurrent_modification` is set - but that may be fine. A sources expert would know!)

Also makes sense to get this behaviour behind a flag. But I understand that's in the works also!

src/persist-client/src/write.rs
```diff
@@ -453,6 +453,13 @@ where
     ///
     /// The clunky multi-level Result is to enable more obvious error handling
     /// in the caller. See <http://sled.rs/errors.html> for details.
+
+    // NB(ptravers): We no longer validate that every batch covers the entire range between
```
aljoscha: Should this just be included in the docstring, without `NB(ptravers)`. It's important to point out at least the very last part

src/storage/src/render/persist_sink.rs
```diff
@@ -1313,52 +1311,55 @@ where
...
-
-                            // Re-add the new batch to the list of batches to
-                            // process.
-                            done_batches.push(new_done_batch_metadata.clone());
-
                             // Retain any batches that are still in advance of
                             // the new lower, and delete any batches that are
                             // not.
-                            //
-                            // Temporary measure: this bookkeeping is made
```
aljoscha, line 1329: Do we need to retain this comment?

src/storage/src/render/persist_sink.rs
```diff
@@ -1313,52 +1311,55 @@ where
...
 
+                            batch_lower = mismatch.current.clone();
+                            to_append = batches.iter_mut()
+                                .map(|batch| &mut batch.batch)
+                                .collect::<Vec<_>>();
+
                             // Best-effort attempt to delete unneeded batches.
                             future::join_all(batch_delete_futures).await;
+
+                            if bail_on_concurrent_modification {
```
aljoscha: Why was the bail behavior changed? I'd think that we still need to bail on mismatch when `bail_on_concurrent_modification == true`.

Side note: `bail_on_concurrent_modification` is currently always false, because we only use `upsert_continual_feedback`, which doesn't need bailing on concurrent modification... 🤷‍♂️

src/persist-client/src/write.rs
ptravers: yeah, good call.

src/storage/src/render/persist_sink.rs
ptravers: should I just remove the bail on concurrent modification?

src/storage/src/render/persist_sink.rs
ptravers: no, after talking with the persist team we do not want to use the append api. the batch may also contain data for multiple timestamps.

src/storage/src/render/persist_sink.rs
ptravers: I had to move the bail behaviour as we rely on the loop to do retries but cannot alter the state of to_append as it has already been passed to persist.

src/persist-client/src/write.rs
ptravers: done.

src/storage/src/render/persist_sink.rs
ptravers: added comments explaining the behaviour of the loops and at what point we would want to exit.

aljoscha: I see! I would be in favour of removing the bail behavior, but I'd also be in favour of removing the "old" UPSERT impl that needed it. @petrosagg was against that in the past, so I'd defer to him

src/persist-client/src/lib.rs
```diff
@@ -1171,38 +1171,43 @@ mod tests {
...
+            //        .append_batch(batch, Antichain::from_elem(2), Antichain::from_elem(3))
+            //        .await
+            //        .unwrap_err(),
+            //    InvalidUsage::InvalidBatchBounds {
+            //        batch_lower: Antichain::from_elem(3),
+            //        batch_upper: Antichain::from_elem(4),
+            //        append_lower: Antichain::from_elem(2),
+            //        append_upper: Antichain::from_elem(3),
+            //    },
+            //);
```
petrosagg: Do we intend to uncomment this code in the future? If not it should get deleted

src/persist-client/src/write.rs
```diff
@@ -453,6 +453,12 @@ where
     ///
     /// The clunky multi-level Result is to enable more obvious error handling
     /// in the caller. See <http://sled.rs/errors.html> for details.
+    /// We no longer validate that every batch covers the entire range between
+    /// the expected and new uppers, as we wish to allow combining batches that
+    /// cover different subsets of that range, including subsets of that range
+    /// that include no data at all. The caller is responsible for guaranteeing
+    /// that the set of batches provided collectively include all updates for
+    /// the entire range between the expected and new upper.
```
petrosagg, line 463: Instead of adding text describing a modification we should modify the comment to state how this API call works after this PR. Specifically the documentation should mention how the shard is modified relative to the batches passed into the call without mentions of how it worked in the past.

If I understand correctly this PR modifies the behavior such that following is true:

> After a successful call the shard upper frontier is advanced to `new_upper` and the set of shard updates that are beyond `expected_upper` and not beyond `new_upper` equal the set of updates that are beyond `expected_upper` and `new_upper` from each batch in `batches`. All other updates in `batches` are ignored.

src/persist-client/tests/machine/truncate
```diff
@@ -38,14 +38,16 @@ write-batch output=nope lower=1 upper=3
 ----
 parts=0 len=0
 
-truncate-batch-desc input=nope output=nope0 lower=0 upper=2
-----
-error: invalid batch bounds [Antichain { elements: [1] }, Antichain { elements: [3] }) for append call with [Antichain { elements: [0] }, Antichain { elements: [2] })
+# NB(ptravers): see compare_and_append_batch.
+# truncate-batch-desc input=nope output=nope0 lower=0 upper=2
+# ----
+# error: invalid batch bounds [Antichain { elements: [1] }, Antichain { elements: [3] }) for append call with [Antichain { elements: [0] }, Antichain { elements: [2] })
```
petrosagg: Do we intend to revert these commented out tests? If not we should remove them

src/storage/src/render/persist_sink.rs
```diff
@@ -932,439 +935,574 @@ where
...
+
+                            assert!(
+                                is_new,
+                                "append_batches: sink {} got more than one batch \
+                                    for a given description in-flight: {:?}",
+                                collection_id, in_flight_batches
+                            );
                         }
+
+                        continue 'consumer;
```
petrosagg: We have a different kind of API on the async handle to avoid these goto-style statements. We should `select!` on b`batch_description_input.ready()` and then inside the handler you can do a synchronous loop that will give you all ready items like so:

```rust
while let Some(..) = batch_description_input.next_sync() {
   ...
}
```

src/storage/src/render/persist_sink.rs
```diff
@@ -932,439 +935,574 @@ where
...
+                                .entry(batch_description)
+                                .or_default();
+
+                            batches.finished.push(FinishedBatch {
+                                batch: write.batch_from_transmittable_batch(batch.batch),
+                                data_ts: batch.data_ts,
+                            });
+                            batches.batch_metrics += &batch.metrics;
                         }
+                        continue 'consumer;
```
petrosagg: Same here

src/storage/src/render/persist_sink.rs
```diff
@@ -932,439 +935,574 @@ where
...
+
+        if let Err(e) = result {
+            return Err(e);
+        }
+    }
+}
 
-            while let Some(done_batch_metadata) = done_batches.pop() {
-                in_flight_descriptions.remove(&done_batch_metadata);
+async fn update_persist_with_batches(
```
petrosagg: Does this need to be a separate function? Its logic only makes sense in the context of the function body above, since they also share a bunch of mutable state (`done_batches` and `in_flight_descriptions`). We should keep this as one function or try to come up slices of the logic that can stand on their own.

src/storage/src/render/persist_sink.rs
```diff
@@ -932,439 +935,574 @@ where
...
+        // `batches_frontier` because it is held back by the writer
+        // operator as long as a) the `batch_description_frontier` did
+        // not advance and b) as long as the `desired_frontier` has not
+        // advanced to the `upper` of a given batch description.
+
+        let mut done_batches = in_flight_descriptions
+            .iter()
+            .filter(|(lower, _upper)| !PartialOrder::less_equal(&batches_frontier, lower))
+            .cloned()
+            .sorted_by(|a, b| {
```
petrosagg: This is `Itertools::sorted_by` which will create a temporary vector with all the elements of the iterator, sort them, and then pass the resulting vector as an iterator to `collect`, which will allocate a brand new vector to put the elements in.

I think what you want to do is collect first and then sort the resulting vector in a separate statement

src/storage/src/render/persist_sink.rs
```diff
@@ -932,439 +935,574 @@ where
...
+    in_flight_descriptions: &mut std::collections::HashSet<(
+        Antichain<mz_repr::Timestamp>,
+        Antichain<mz_repr::Timestamp>,
+    )>,
+) -> Result<(), anyhow::Error> {
+    let (mut batch_lower, batch_upper) = match (done_batches.first(), done_batches.last()) {
+        (Some(first), Some(last)) => (first.0.clone(), last.1.clone()),
+        (None, None) => {
+            upper_cap_set.downgrade(current_upper.borrow().iter());
+            //
```
petrosagg: Leftover comment?

src/storage/src/render/persist_sink.rs
```diff
@@ -932,439 +935,574 @@ where
...
+            return Ok(());
+        }
+        // SAFETY(ptravers): if first exists so must last.
```
petrosagg: `SAFETY` comments are reserved for documenting how unsafe code blocks are upholding whatever invariants are necessary to avoid UB. Since there is no `unsafe` here we should make this a normal comment.

src/storage/src/render/persist_sink.rs
```diff
@@ -932,439 +935,574 @@ where
...
+    }
 
-                let batch_set = in_flight_batches
-                    .remove(&done_batch_metadata)
-                    .unwrap_or_default();
+    let mut to_append = batches.iter_mut().map(|b| &mut b.batch).collect::<Vec<_>>();
+
+    // NB(ptravers): We loop here to handle competing writers to the same shard.
+    // We handle failure to write in three situations:
+    //  1. the lower of the batch is greater than the current upper of shard - persist_sink exits with an error.
```
petrosagg, line 1554: Can this happen? I thought the changes in the API call of persist allow that.

Also, we should us the "beyond" terminology when talking about comparing frontiers. Specifically we use "f1 is beyond f2" if `PartialOrder::less_equal(f2, f1)` and "f1 is not beyond f2"  if `!PartialOrder::less_equal(f2, f1)`. More rarely we use the term "f1 is strictly beyond f2" to mean `ParialOrder::less_equal(f2, f1) && f1 != f2`

So my question here is if persist refuses the API call if for any of the to-be-committed batches `batch.lower` is strictly beyond the current shard upper. My understanding is that it will not refuse it but maybe I misunderstood the comment.

src/storage/src/render/persist_sink.rs
```diff
@@ -932,439 +935,574 @@ where
...
+    //  2. the upper of the shard is less than the upper of the batch and greater
+    //     than the lower of the batch - we retry with a truncated batch.
+    //  3. all other cases - we exit the retry loop after updating the current_upper to the upper reported in the error.
```
petrosagg, line 1557: Is this a failure condition? This is listed as the third item in the "We handle failure to write in three situations" list, but this sounds like the success case to me

src/storage/src/render/persist_sink.rs
```diff
@@ -932,439 +935,574 @@ where
...
+                                ?batch_lower,
+                                ?batch_upper,
+                                ?current_upper,
+                                "persist_sink has come out of read-only mode"
+                            );
+                        }
 
-                let (batch_lower, batch_upper) = done_batch_metadata;
+                        // It's okay to write now.
+                        break 'spin_on_lock Ok(());
```
petrosagg: During the time it takes to do a `compare_and_append` more batches may be ready to commit so I think we should re-evaluate the done batches each time we retry to commit the maximum amount of work

src/storage/src/render/persist_sink.rs
ptravers: same response.

see: https://github.com/MaterializeInc/materialize/blob/367c04ad36142f617b1aeb69d830f26fd57fb3c2/src/storage/src/render/persist_sink.rs#L1060

src/storage/src/render/persist_sink.rs
ptravers: can we do that in a separate PR? I'm a little concerned about broadening the scope of this PR. this is existing behaviour. the only change I made is to annotate with a name for clarity.

see: https://github.com/MaterializeInc/materialize/blob/367c04ad36142f617b1aeb69d830f26fd57fb3c2/src/storage/src/render/persist_sink.rs#L1037

src/persist-client/tests/machine/truncate
ptravers: I wasn't totally sure if we wanted to keep them to enable testing that the feature flag works?

src/persist-client/src/lib.rs
ptravers: wasn't sure if we wanted it to test that the flag works.

src/storage/src/render/persist_sink.rs
ptravers: good call.

src/storage/src/render/persist_sink.rs
ptravers: guilty!

src/storage/src/render/persist_sink.rs
ptravers: I separated it out as a way to allow cargofmt to work again. also I found it fairly hard to follow as one large function. I will need to break out a separate function to support the feature flag I'm currently adding.

src/storage/src/render/persist_sink.rs
ptravers: good point!

src/storage/src/render/persist_sink.rs
ptravers: I believe the assertion that would fail here is that we are asserting the current upper of the shard is the lower of the batch and persist will error as the current upper is less than the batch lower.

see: https://github.com/MaterializeInc/materialize/blob/367c04ad36142f617b1aeb69d830f26fd57fb3c2/src/persist-client/src/write.rs#L644

src/storage/src/render/persist_sink.rs
ptravers: I have labeled it a failure as we would have not written anything in this case. does that make sense as a failure to you?

src/persist-client/src/lib.rs
petrosagg: Hm, I didn't see any feature flag added in this PR. Is the intention to add one?

src/storage/src/render/persist_sink.rs
petrosagg: I see. The mythical "separate PR" never comes in my experience but you're also totally right that this is not behavior introduced in this PR so yeah, we can leave this as-is.

src/storage/src/render/persist_sink.rs
petrosagg: I see. I think one confusion here is using the term "batch lower" to refer to the expected shard upper passed to the function. I got confused because in the function call we are also passing in a list of batches and the lower of *those* doesn't have any constraints, right?

src/storage/src/render/persist_sink.rs
petrosagg: Hm, a failure to me means `compare_and_append` returned an `Err(_)` instead of an `Ok(_)`. Is this third case talking about the `Ok(_)` behavior or an `Err(_)` behavior?

src/persist-client/src/write.rs
ptravers: since I'm passing in a flag now. are you okay with me amending this comment to state that we disable the checks outlined in the above comment when the `enforce_single_batch` flag is passed? I'm also very open to better names...

src/storage/src/render/persist_sink.rs
ptravers: we failed to append but the shard was moved to the correct upper? so is that a failure?

src/storage/src/render/persist_sink.rs
ptravers: that's a good point. should we do that in a follow up? and leave this to soak as is?

src/storage/src/render/persist_sink.rs
```diff
@@ -932,439 +935,850 @@ where
...
+            #[allow(clippy::disallowed_types)]
+            let mut in_flight_descriptions = std::collections::HashSet::<(
+                Antichain<mz_repr::Timestamp>,
+                Antichain<mz_repr::Timestamp>,
+            )>::new();
+
+            // In flight batches that haven't been `compare_and_append`'d yet, plus metrics about
+            // the batch.
+            let mut in_flight_batches =
+                HashMap::<(Antichain<mz_repr::Timestamp>, Antichain<mz_repr::Timestamp>), BatchSet>::new();
```
antiguru: Pointing out a potential smell here. Often we can use frontier semantics to encode what should happen when instead of relying on the antichains itself. I'm not familiar enough to suggest an alternative implementation of the algorithm below.

Also, the hash implementation of antichains is rather slow.

antiguru: Oh, I see, this exists already. Never mind then.

ptravers: I will add a TODO to improve this code path based on your suggestions.

src/storage/src/render/persist_sink.rs
ptravers: fair enough! I will add an issue to make it more likely the mythic PR arrives

https://github.com/MaterializeInc/database-issues/issues/9369

src/persist-client/src/lib.rs
ptravers: added! wasn't sure how best to add functionality and flag or what we do in terms of layering a review?

src/storage/src/render/persist_sink.rs
ptravers: this is a great question. looking into it now. I believe that the lower of all the batches in the batch would be bounded by the batch lower? but I am confirming now.

src/storage/src/render/persist_sink.rs
ptravers: sorry for clarity in all of these cases compare_and_append returned an error that there was an `UpperMismatch` so the else is handling an upper mismatch that doesn't fit 1 or 2. which should mean that the upper has advanced beyond the batches we are holding. we exit and say that we no longer have data that can be written. does that make sense?

there is a `"failure to write to persist"` rather than an error. does that make sense to you?

src/persist-client/src/write.rs
```diff
@@ -453,12 +453,21 @@ where
...
     #[instrument(level = "debug", fields(shard = %self.machine.shard_id()))]
     pub async fn compare_and_append_batch(
         &mut self,
         batches: &mut [&mut Batch<K, V, T, D>],
         expected_upper: Antichain<T>,
         new_upper: Antichain<T>,
+        enforce_single_batch: bool,
```
ptravers, line 470: @bkirwi do you mind taking a wee look at this?

ptravers:
running nightly for [b7feb76](https://github.com/MaterializeInc/materialize/pull/32509/commits/b7feb769bb5e77bb06f249e83faa93a71a64ccbc) please note this does not test the batching change but tests that the existing implementation still works.

src/storage/src/render/persist_sink.rs
ptravers: we will need to do this work to support draining the input channels when we get the write lock further down. does that make sense to you as well?

ptravers: I think we would still need to use an outer loop since we would want to consume until both channels are empty, right? I have an implementation locally. right now we get a load balancing behaviour not sure how big of an issue it would be if we changed that behaviour?

ptravers:
apologies, rebased to latest main.

ptravers:
nightly build with feature flag checking for new flag https://buildkite.com/materialize/nightly/builds/12293/steps/waterfall

ptravers:
two nightly builds:
1. random feature flags - https://buildkite.com/materialize/nightly/builds/12294/steps/waterfall
2. default flags with persist sink batching enabled - https://buildkite.com/materialize/nightly/builds/12295/steps/waterfall

src/persist-client/src/write.rs
bkirwi: Seems fine to me!

ptravers: great! thanks!

def-:
Awesome, thanks for triggering Nightly! Note that with random feature flags we currently expect to see a bunch of test failures (OoMs, perf. regressions, different plans), but not wrong results or panics. I see a bunch of scary stuff in there:
1. With random flags there are wrong results in [🏎️ testdrive with SIZE 1](https://buildkite.com/materialize/nightly/builds/12294#019760fd-11f6-4a42-a9a1-1fe7ef26c8bf) and [testdrive with :azure: blob store](https://buildkite.com/materialize/nightly/builds/12294#019760fd-11f8-411f-9f00-e143e7f1b89a): 
```
fetch-select-during-ingest.td:73:1: non-matching rows: expected:
[["123"], ["234"]]
got:
[["123"]]
Poor diff:
- 234
```
I'm not sure if this is related to your PR, but I haven't seen it on main yet! Doing some more test runs just in case though: [1](https://buildkite.com/materialize/nightly/builds/12299), [2](https://buildkite.com/materialize/nightly/builds/12298), [3](https://buildkite.com/materialize/nightly/builds/12297) Edit: Couldn't repro with those, trying locally with your PR and the specific set of flags now.
2. With default flags two tests panic: [CRDB rolling restarts](https://buildkite.com/materialize/nightly/builds/12295#019760fe-170c-472e-a02d-0e352c10e38f) & [Zero downtime](https://buildkite.com/materialize/nightly/builds/12295#019760fe-1721-4fc8-9c69-b005fe9ee414):
```
0dt-mz_new-1             | 2025-06-11T23:44:26.527793Z  thread 'persist:0009' panicked at src/persist-client/src/fetch.rs:1214:13: key=n0014700/p6dda475f-9bd9-466b-a941-fc3190577c91 inline=Description { lower: Antichain { elements: [1749685442001] }, upper: Antichain { elements: [] }, since: Antichain { elements: [0] } } registered=Description { lower: Antichain { elements: [1749685441001] }, upper: Antichain { elements: [1749685444001] }, since: Antichain { elements: [0] } }
```
3. With default flags [Zippy Postgres CDC](https://buildkite.com/materialize/nightly/builds/12295#019760fe-16cb-4aa0-9781-ce020abaa323) has an OoM:
```
Jun 11 22:38:21 hetzner-aarch64-8cpu-16gb-07668cf2 kernel: Memory cgroup out of memory: Killed process 389025 (clusterd) total-vm:28756204kB, anon-rss:12667528kB, file-rss:47520kB, shmem-rss:0kB, UID:999 pgtables:48544kB oom_score_adj:0
```
I have not seen this flaking on main, is it expected that your PR increases memory usage?

The unused dependencies and Kafka failures are also on main, I'll look into them, feel free to ignore for your PR.

ptravers:
brilliant! thanks so much for deep diving. I will take a look at the memory issue. I was not expecting a significant bump in memory usage as a result of the PR so will look into it.

def-:
A rerun of the Zippy Postgres CDC didn't OoM, so the signal is not very strong.

ptravers:
yeah, I will still triple check. by:
1. soak test - running a long running source with a limited memory footprint for the materialized.
2. pushing a large changeset through a source (write large batch to postgres table that has a source in materialized). 
 
hoping that either large amount of data in a short time or over a long time would expose a memory issue.

def-:
The wrong result in Testdrive reproduces!
```
bin/mzcompose --find testdrive down && CI_SYSTEM_PARAMETERS_SEED=019760fd-11f6-4a42-a9a1-1fe7ef26c8bf CI_SYSTEM_PARAMETERS=random bin/mzcompose --find testdrive run default --default-size=1 fetch-select-during-ingest.td
```
But also on main, I'll open a bug, not related to your PR. Edit: https://github.com/MaterializeInc/database-issues/issues/9378

src/storage/src/render/persist_sink.rs
ptravers: I don't see an issue here since we will catch the expected upper deviating from the batch lower when we attempt to write then filter based on the reported upper from persist before retrying. feel free to comment if you believe there is an issue I am missing. I would prefer to let persist tell us if there's an issue with the expected upper not matching the batch lower.

ptravers:
nice! thanks for tracking that down!

src/persist-client/src/write.rs
petrosagg: I agree we need a better name. If I just read the parameter name, "enforce_single_batch" strongly implies that the `batches` parameter must contain a single entry. But that's not what it does, you can still have many entries as long as the batch boundaries match the CaA call boundaries. Also, if this ends up being an additional argument (see below) of such foundational API we should create an enum for it. Example naming could be:

```rust
/// Configures the behavior of CaA operations when validating batch boundaries.
enum ArbitraryBatchBoundaries {
    /// All appended batches must have their lower frontier equal to the
    /// expected upper and their upper frontier equal the new upper.
    Deny,
    /// Batches can have arbitrary boundaries. Only the updates that happen at 
    /// times beyond the expected upper and not beyond the new upper are appended. 
    /// All other updates are ignored. 
    Allow,
}
```

All that said, we don't usually configure persist on a per-call basis so this doesn't seem like the right place to put this configuration option. The `PersistConfig` struct seems more appropriate for configuring this behavior https://dev.materialize.com/api/rust-private/mz_persist_client/cfg/struct.PersistConfig.html @bkirwi any thoughts?

src/storage-types/src/dyncfgs.rs
```diff
@@ -256,6 +256,13 @@ pub const STORAGE_USE_CONTINUAL_FEEDBACK_UPSERT: Config<bool> = Config::new(
     "Whether to use the new continual feedback upsert operator.",
 );
 
+/// Whether to use the new continual feedback upsert operator.
```
petrosagg, line 259: left over comment from copy pasta

src/storage-types/src/dyncfgs.rs
```diff
@@ -256,6 +256,13 @@ pub const STORAGE_USE_CONTINUAL_FEEDBACK_UPSERT: Config<bool> = Config::new(
...
+pub const STORAGE_SINK_USE_BATCHING_TO_PERSIST: Config<bool> = Config::new(
+    "storage_sink_write_batches_to_persist",
```
petrosagg, line 261: The LD parameter name should match the rust name

src/storage/src/render/persist_sink.rs
```diff
@@ -932,439 +935,804 @@ where
...
-        }
+    let (shutdown_button, errors) = append_op.build_fallible(move |caps| {
+        Box::pin(async move {
+            // This may SEEM unnecessary, but metrics contains extra
+            // `DeleteOnDrop`-wrapped fields that will NOT be moved into this
+            // closure otherwise, dropping and destroying
+            // those metrics. This is because rust now only moves the
+            // explicitly-referenced fields into closures.
+            let metrics = metrics;
+            let storage_state_configuration = storage_state_configuration;
```
petrosagg, line 946: This seems unnecessary. Does rust complain without this?

src/storage/src/render/persist_sink.rs
```diff
@@ -932,439 +935,804 @@ where
...
+            current_upper.borrow_mut().clone_from(write.upper());
+            upper_cap_set.downgrade(current_upper.borrow().iter());
+            source_statistics.initialize_snapshot_committed(write.upper());
+
+            // The current input frontiers.
+            let mut batch_description_frontier = Antichain::from_elem(Timestamp::minimum());
+            let mut batches_frontier = Antichain::from_elem(Timestamp::minimum());
+
+            'consumer: loop {
+                // Drain both input channels as of resumption and advance frontier.
```
petrosagg, line 1009: What do you mean here by "as of resumption"?

src/storage/src/render/persist_sink.rs
ptravers: removed this in the end. just fyi.

src/persist-client/src/write.rs
bkirwi: My thoughts are as follows:
- IIUC this is a "rollout" flag: we expect it to go away in a month or whenever we're confident in the new feature. For that reason I don't think it's worth investing _too_ much brainpower in choosing the ideal name or type etc. On the other hand, if someone is motivated to provide that brainpower and come up with a better name regardless, that's cool too. 
- I think it is fine to add this as an argument. Ideally we'll get everyone on the same behaviour once we've built confidence and finished the rollout, and this param will go away; in the meantime, this is a simple way to enable or disable the check on a per-call basis. (If we added it to the persist config, we'd have to turn the validations on or off globally, and we'd run the risk of a flag rollout ordering issue where the validations are enabled while we're still running the new logic in the source and we panick ourselves unnecessarily.)
- "All appended batches must have their lower frontier equal to the expected upper and their upper frontier equal the new upper." is not the current policy, and would be too strict even for the current logic; it's a <= / >= check, depending. (I know that wasn't really the point of your example code, but I figured I'd mention it anyways in case it causes a real confusion.)

petrosagg: Just to make sure we're on the same page, @bkirwi is feature flagging the *validation* something you want for the persist side? Because my understanding is that we can remove validation unconditionally and only feature flag the behavior of the persist sink to coalesce multiple batches in a single CaA call, which would avoid the need for this extra parameter.

src/storage-types/src/dyncfgs.rs
ptravers: 😬 you got me!

src/storage-types/src/dyncfgs.rs
ptravers: ah cheers!

src/storage/src/render/persist_sink.rs
ptravers: ah yeah, it was guess that I accidentally left in for an unrelated issue. thanks!

src/storage/src/render/persist_sink.rs
ptravers: I can remove it. I wrote it for myself to track why this loop exists and where it gets resumed with timely async.

src/persist-client/src/write.rs
ptravers: petros raises a good point that by doing the coallescing behaviour in the persist_sink not here we can avoid the validation ever failing.

bkirwi: Given that we were already introducing a flag for the client-side logic, it seemed like a nice touch to thread it through here to limit the scope. (While Persist doesn't seem to rely on this check for correctness, it's possible that the validation is nice for other callers; and since this PR's already fairly complex I'm happy to not worry about stuff like that until later.)

ptravers: the parameter still has value as it means that we are not changing the behaviour of the callers of this function while we roll out.

ptravers: > I agree we need a better name. If I just read the parameter name, "enforce_single_batch" strongly implies that the batches parameter must contain a single entry. But that's not what it does, you can still have many entries as long as the batch boundaries match the CaA call boundaries.

that's not quite correct you can have many batches so long as there are no frontier advancements since those present as an empty set of batches. right now frontier advancements work because we skip the validation step in the for loop since there is nothing in the list of batches. but maybe we're talking past eachother?

petrosagg: What is a frontier advancement?

My main question is whether setting that flag to true implies that `batches.len() == 1`. This is what the name "enforce single batch" suggests to me. Is this what the flag does?

PR #32513, opened 2025-05-16
persist: expose API for writing/reading "free-standing" batches

aljoscha:
Add API for writing and reading batches without creating a `WriteHandle`/`ReadHandle`. Writing batches still requires a `ShardId`, which is used for namespacing in blob storage.

Work towards
https://github.com/MaterializeInc/database-issues/issues/9180, where we want to use persist batches/blob to stash peek results when sending them back to `environmentd`.


src/persist-client/src/lib.rs
```diff
@@ -121,6 +133,9 @@ pub const BUILD_INFO: BuildInfo = build_info!();
 // Re-export for convenience.
 pub use mz_persist_types::{PersistLocation, ShardId};
 
+pub use crate::internal::encoding::Schemas;
+pub use crate::internal::state::HollowBatch;
```
aljoscha, line 137: You might object to these being exposed. We only need `HollowBatch` because we use it to sniff out the size/num_rows from stashed batches without turning them into a `Batch` and/or reading them.

src/persist-client/src/lib.rs
```diff
@@ -539,6 +554,187 @@ impl PersistClient {
...
+    /// blob storage which can then be appended to the given shard using
+    /// [Self::compare_and_append_batch] or [Self::append_batch], or which can
+    /// be read using [PersistClient::read_batches_consolidated].
+    ///
+    /// The builder uses a bounded amount of memory, even when the number of
+    /// updates is very large. Individual records, however, should be small
+    /// enough that we can reasonably chunk them up: O(KB) is definitely fine,
+    /// O(MB) come talk to us.
+    #[instrument(level = "debug", fields(shard = %shard_id))]
+    pub async fn batch_builder<K, V, T, D>(
```
aljoscha, line 567: These are basically duplicated code. I could add a `WriteHandle::batch_builder_inner` that doesn't take a `self` but all the arguments it needs. And then both `WriteHandle` and the code here could call that one.

Same for `read_batches_consolidated` below.

I didn't do this for now because I felt that the method signature was almost the largest part of these methods. But happy to change that!

ParkMyCar (approved):

src/persist-client/src/lib.rs
```diff
@@ -539,6 +554,189 @@ impl PersistClient {
...
+                // If this is a noticeable performance impact, thread the max-len in from the caller.
+                max_len: self.cfg.compaction_yield_after_n_updates,
+                max_bytes: BLOB_TARGET_SIZE.get(&self.cfg).max(1),
+            }
+        };
+
+        Ok(Cursor {
+            consolidator,
+            // WIP: What to do about this?
+            _lease: Lease::new(SeqNo(0)),
```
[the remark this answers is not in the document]
aljoscha: Also agreed! I played around with this for quite a while but didn't find a satisfying solution yet. The basic problem when workin with persist batches is that nothing can really prevent the caller from dropping if they're "malicious". Even when passing in owned `Batches`, there's nothing stopping the caller from turning the same `ProtoBatch` into a `Batch` again and then deleting it, for example.

For reference, here's the code that uses the API: https://github.com/MaterializeInc/materialize/blob/2d0800993b593d85822ab1c1a4c747e8323ec145/src/adapter/src/coord/peek.rs#L888. Specifically it's the end of the batches' lifetime where they're deleted.

The code does the right thing, but yes, nothing enforces that the batches aren't deleted before we finish reading them.

Also, another difficulty is that dropping batches requires async. So when I pass in owned batches, all I could do is add an async `clean_up_lease` method on `Cursor`, with a trait `DroppableLease` that has an async `release()` method, where's that's a no-op for the current `Lease`. But I'd add a new struct that keeps the owned batches which can delete them when they're "released". This is a very rough sketch, so sorry if it's not clear. It would add quite a bit of complexity, though, imo.

Or, some more spitballing. I could add a generic parameter to `Cursor` for the lease type, and an `into_lease()` method that allows me to get back the lease. In the case of this method here, the lease type would be a `Vec<Batches>`, so the code that reads them could change the cursor for the original vec of batches and then clean them up. What do you think about that one?

src/persist-client/src/lib.rs
bkirwi: `Batch::encoded_size_bytes` seems fine, yeah!

aljoscha:
@bkirwi I pushed an extra commit that adds the lease type to `Cursor`. And here's how it's used in the bigger PR: https://github.com/MaterializeInc/materialize/blob/3ab9627fe2e9b72b05a2a0585f8e5e2c8d073587/src/adapter/src/coord/peek.rs#L887

bkirwi (approved):
There are a few open comments from before (adding a docstring etc.) but the changes look good and I don't think anything's blocking! Thank you for your caution here!

aljoscha:
thanks for the reviews! 🙇‍♂️

PR #32563, opened 2025-05-23
Broaden types converted to jsonb numeric

frankmcsherry:
We detect a few scalar types that can convert into jsonb's numeric representation, but types we don't explicitly match are converted to strings.

Todo:
1. add tests that cover these cases. The user reported error was for `select jsonb_build_object('data', count(*)) from ..`.
2. add exhaustive matches, to avoid missing cases like this in the future.

Putting up without this work to raise for discussion. Specifically, the code comments say
```
///   * All numeric types are converted to `Float64`s, then become JSON numbers.
```
which .. isn't actually what happens (they are directly converted to `Numeric { max_scale: None }`). If it is important that something very close to that happens we should discuss, but we already convert `Numeric`, which subsumes all of our integer types, into `Numeric` rather than `Float64` or anything similar.

### Checklist

### DIFF
--- a/src/sql/src/plan/typeconv.rs
+++ b/src/sql/src/plan/typeconv.rs
@@ -911,7 +911,7 @@ pub fn to_jsonb(ecx: &ExprContext, expr: HirScalarExpr) -> HirScalarExpr {
         Bool | Jsonb | Numeric { .. } => {
             expr.call_unary(UnaryFunc::CastJsonbableToJsonb(func::CastJsonbableToJsonb))
         }
-        Int16 | Int32 | Float32 | Float64 => plan_cast(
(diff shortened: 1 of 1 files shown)


antiguru (approved):
Looks good. We might have some work to do in this area. For `select jsonb_build_object('data', 10000000000000000000000000000000000);`, Postgres and Materialize disagree:
```
postgres=# select jsonb_build_object('data', 10000000000000000::int8);
     jsonb_build_object
-----------------------------
 {"data": 10000000000000000}
(1 row)
```
Materialize:
| jsonb_build_object |
| ------------------ |
| {"data":1e+34}     |

frankmcsherry:
Merging, but to ack the comment: I don't think the PR regresses here does it? I'm imagining that number is in `numeric` territory, and that code went unchanged.

PR #32566, opened 2025-05-23
docs: tweak prepare + execute page

kay-kim:


doc/user/content/sql/execute.md
```diff
@@ -22,10 +22,18 @@ Field | Use
 
 ## Example
 
+The following example [prepares a statement](/sql/prepare/) `a` and runs it
```
kay-kim, line 25: Later on (when I update the syntax to remove the rr diagram), I'll rename `a` to something else and will mention what it actually does.

frankmcsherry (approved):

PR #32590, opened 2025-05-27
Improve ColumnType::union

antiguru:
Improve the implementation of ColumnType::union to avoid cloning if not needed. Also, make the code more readable by destructing tuples into their components. No functionality changes expected otherwise.


teskje (approved):
LGTM

antiguru:
Thanks for the reviews!

PR #32591, opened 2025-05-27
ColumnName wraps boxed string

antiguru:
ColumnName currently owns a string, but doesn't need mutable access in all but one places. This change alters it to a boxed str, which reduces the size from 24 to 16 bytes. It also adds a deref implementation, which allows us to remove some of the `as_str` calls.

No urgency to merge it, just capturing what I found in my local changes.


ggevay (approved):
LGTM, but @mgree might want to also check this when he is back.

antiguru:
Thanks for the reviews!

PR #32598, opened 2025-05-27
[persist] Lower some thresholds for persist in CI

bkirwi:
- Make the code more robust to aggressive compaction limits. (Instead of panicking, we fall back to the already-existing code for when we can't fit two runs in memory.)
- Exercise a wider range of configs in CI.
- Fix a `txn-wal` bug, which could be triggered when data had undergone compaction but wasn't globally consolidated. (Which can happen when memory limits are too small to compact all parts in one go.)

### Motivation

Turns out we're missing some test coverage in this area!

### Tips for reviewer

I'd love a review from a `txn-wal` expert, since I'm not that familiar with the code there.

### Checklist

### DIFF
--- a/misc/python/materialize/parallel_workload/action.py
+++ b/misc/python/materialize/parallel_workload/action.py
@@ -1064,6 +1064,13 @@ class FlipFlagsAction(Action):
             "16",
             "1000",
         ]
+        self.flags_with_values["persist_compaction_memory_bound_bytes"] = [
+            # 64 MiB, 1 * 128 MiB, 4 * 128 MiB, 8 * 128 MiB
+            "67108864",
+            "134217728",
+            "536870912",
+            "1073741824",
+        ]
         self.flags_with_values["persist_part_decode_format"] = [
             "row_with_validate",
             "arrow",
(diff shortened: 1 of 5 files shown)


bkirwi:
For the record, the full CI history: https://buildkite.com/materialize/nightly/builds?branch=bkirwi%3Aci-tune

The first couple runs included the CI tuning, but not the bugfix, rebased on a recent and old release. (To check that it wasn't finding a recently introduced bug.)

The next two runs include the bugfix, and the failures have disappeared.

misc/python/materialize/parallel_workload/action.py
```diff
@@ -1064,6 +1064,12 @@ def __init__(
             "16",
             "1000",
         ]
+        self.flags_with_values["persist_compaction_memory_bound_bytes"] = [
```
ParkMyCar, line 1067: Above we limit the size of blobs to 1 MiB, 16 MiB, and 128 MiB. Should those limits be reflected here too? Assuming that we're tuning these to fit a very limited number of Parts/blobs in-memory at a given time during compaction.

ParkMyCar (approved):

def- (approved):
No complaints from QA!

misc/python/materialize/parallel_workload/action.py
bkirwi: Done, thanks!

PR #32602, opened 2025-05-28
lgalloc disk usage limiter

antiguru:
This PR adds a limiter task that periodically confirms that lgalloc's current disk usage is below some configured limit and terminates the process otherwise.

The disk limit is calculated by applying a configurable factor to the process' memory limit. There is also a burst option, allowing the process to use more than the configured disk limit for a time.

The default limits are configured to match the current production reality: The disk limit is twice the memory limit and bursting is disabled.

### Motivation

Closes MaterializeInc/database-issues/issues/9306

### Checklist


src/clusterd/src/lib.rs
```diff
@@ -354,6 +355,7 @@ async fn run(args: Args) -> Result<(), anyhow::Error> {
         ComputeInstanceContext {
             scratch_directory: args.scratch_directory,
             worker_core_affinity: args.worker_core_affinity,
+            announce_memory_limit: args.announce_memory_limit,
```
teskje: I am confused about this being called `announce_memory_limit` instead of just `memory_limit`. Is there a good reason? If not, it's probably too much of a hassle to change the command-line option, but we can change the internal variable names.

src/compute-types/src/dyncfgs.rs
```diff
@@ -83,6 +83,34 @@ pub const LGALLOC_SLOW_CLEAR_BYTES: Config<usize> = Config::new(
...
+);
+
+/// Bias to the lgalloc limiter usage factor.
+pub const LGALLOC_LIMITER_USAGE_BIAS: Config<f64> = Config::new(
+    "lgalloc_limiter_usage_bias",
+    1.,
+    "Multiplicative bias to lgalloc_limiter_usage_factor.",
+);
+
+/// Bias to the lgalloc limiter usage factor.
```
```suggestion
/// Burst factor to disk limit.
```

src/compute-types/src/dyncfgs.rs
```diff
@@ -83,6 +83,34 @@ pub const LGALLOC_SLOW_CLEAR_BYTES: Config<usize> = Config::new(
...
+pub const LGALLOC_LIMITER_BURST_FACTOR: Config<f64> = Config::new(
+    "lgalloc_limiter_BURST_FACTOR",
```
```suggestion
    "lgalloc_limiter_burst_factor",
```

src/compute/src/lgalloc.rs
```diff
@@ -0,0 +1,282 @@
...
+    }
+}
+
+/// Register all metrics into the provided registry.
+///
+/// We do not recommend calling this function multiple times. It is safe to call this function,
+/// but it might delete previous metrics. If we ever want to change this, we should
+/// remove the shared static mutex and make this function return a handle to the metrics.
+///
+/// This function is async, because it needs to be called from a tokio runtime context.
```
teskje: I'm wondering how much value this pattern adds. You add some noise but needing the async keyword, the `allow`, and this comment. You get protection against calling this function outside an async (not necessarily tokio!) context, but I think you also have this protection through CI, as spawning a task would immediately panic if no tokio runtime was available.

If we want to use this pattern here, shouldn't `Limiter::create_task` also be async?

src/compute/src/lgalloc.rs
```diff
@@ -0,0 +1,282 @@
...
+#[allow(clippy::unused_async)]
+pub async fn register_metrics_into(
```
teskje: I think this should be called `run_limiter` or somesuch. A call to "register metrics" spawning a non-metrics task in the background is surprising. A call to "run limiter" running a limiter that also registers metrics into the provided registry is not.

src/compute/src/lgalloc.rs
```diff
@@ -0,0 +1,282 @@
...
+    last_update: Instant,
+}
+
+impl LimiterTask {
+    fn update(&mut self) -> Result<(), anyhow::Error> {
+        // Get lgalloc stats and obtain the disk utilization from file stats, summed across all
+        // files and size classes. Compare the disk utilization against the configured disk limit,
+        // and if it exceeds the limit, reduce the burst budget by the amount of disk utilization
+        // that exceeds the limit. If the burst budget is exhausted, we will not allow any more disk
+        // access and terminate the process.
```
teskje: Nit: This fits better in the module docstring imo.

src/compute-types/src/dyncfgs.rs
```diff
@@ -83,6 +83,34 @@ pub const LGALLOC_SLOW_CLEAR_BYTES: Config<usize> = Config::new(
...
+    0.5,
```
teskje: 0.5 seems like a too-small value, assuming it means you get "0.5 x limit byte-seconds" of burst. If a hydration takes 10 minute that's less than 0.1% of additional disk you can use.

In any case, we should probably disable bursting by default, and make switching it on a conscious decision. Then we can also decide how much bursting we need based on the use case we want to enable.

src/compute/src/lgalloc.rs
```diff
@@ -0,0 +1,282 @@
...
+                if self.burst_budget_remaining >= excess {
+                    self.burst_budget_remaining -= excess;
+                } else {
+                    // Burst budget exhausted, terminate the process.
+                    tracing::error!(
+                        "Disk utilization {} exceeded limit {} and burst budget exhausted",
+                        disk_utilization,
+                        disk_limit
+                    );
+                    std::process::exit(1);
```
teskje: We need to check what implications existing with code 1 has. It might trigger alerts in our monitoring or show the replica as crashed due to an error in the console. Ideally we want this case the look the same as an OOD scenario.

Also, we should log a warning, not an error. Errors will produce noise in Sentry we don't want.

src/compute-types/src/dyncfgs.rs
antiguru: Yep, it should be disabled by default.

src/compute/src/lgalloc.rs
```diff
@@ -0,0 +1,282 @@
...
+        // Notify tasks about updated configuration.
+        self.task.update_dyncfg(config_set);
+    }
+}
+
+enum Update {
+    Interval(Duration),
+    DiskLimit(Option<usize>),
+    BurstBudget(usize),
+}
```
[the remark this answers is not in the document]
teskje: Changed to always apply the limits config atomically.

src/clusterd/src/lib.rs
teskje: Turns out we don't need the `ComputeInstanceContext` field! Changed other instances of `announce_memory_limit` to `memory_limit`.

src/compute/src/lgalloc.rs
teskje: [Discussion in Slack.](https://materializeinc.slack.com/archives/C08ACQNGSQK/p1749654075353419)

src/compute/src/lgalloc.rs
```diff
@@ -0,0 +1,237 @@
...
+            let excess = (disk_usage - disk_limit) * usize::cast_from(elapsed);
+
+            if burst_budget_remaining >= excess {
+                self.burst_budget_remaining -= excess;
+            } else {
+                // Burst budget exhausted, terminate the process.
+                warn!(
+                    disk_usage,
+                    disk_limit, "lgalloc disk utilization exceeded configured limits",
+                );
```
teskje, line 207: rustfmt showing some bad taste here...

teskje:
Tested this in my staging env and confirmed:

* Reaching the configured disk limit terminates the replica.
* Changing the configured disk limit of a healthy replica to a value lower than its current disk usage terminates the replica.
* A replica can survive a hydration on the burst budget that it would otherwise not survive.

def-:
Rebased and retriggered nightly: https://buildkite.com/materialize/nightly/builds/12336 (since Hetzner is still struggling with aarch64 availability)

teskje:
Hm... Some feature benchmark regressions. I think we need to move the lgalloc limit checker into a separate thread so the I/O it's doing doesn't interfere with other tokio tasks.

antiguru:
> I think we need to move the lgalloc limit checker into a separate thread so the I/O it's doing doesn't interfere with other tokio tasks.

Did you have a chance to look at the metric to see how long its invocations took? We should have enough tokio tasks to not interfere, but who knows!

teskje:
> Did you have a chance to look at the metric to see how long its invocations took?

Just checked in my staging env and it seems like the max invocation time is always 512us... which is also the smallest bucket in the histogram :D The benchmark regression has also gone away after a retry, so seems fine to merge.

PR #32605, opened 2025-05-28
Assert diffs sum does not change when swapping in spine batches

DAlperin:
### Checklist

### DIFF
--- a/src/persist-client/src/internal/state_diff.rs
+++ b/src/persist-client/src/internal/state_diff.rs
@@ -794,7 +794,7 @@ fn apply_diff_map<K: Ord, V: PartialEq + Debug>(
 // This might leave state in an invalid (umm) state when returning an error. The
 // caller ultimately ends up panic'ing on error, but if that changes, we might
 // want to revisit this.
-fn apply_diffs_spine<T: Timestamp + Lattice>(
+fn apply_diffs_spine<T: Timestamp + Lattice + Codec64>(
     metrics: &Metrics,
     mut diffs: Vec<StateFieldDiff<HollowBatch<T>, ()>>,
     trace: &mut Trace<T>,
@@ -868,7 +868,7 @@ fn apply_diffs_spine<T: Timestamp + Lattice>(
         // that was generated elsewhere. Most of the time we can, though, so
         // count the good ones and fall back to the slow path below when we
         // can't.
-        if trace.apply_merge_res(&res).applied() {
+        if trace.apply_merge_res(&res, &metrics.columnar).applied() {
             // Maybe return the replaced batches from apply_merge_res and verify
             // that they match _inputs?
             metrics.state.apply_spine_fast_path.inc();
@@ -1427,7 +1427,7 @@ mod tests {
                             leader
                                 .collections
                                 .trace
-                                .apply_merge_res(&FueledMergeRes { output });
+                                .apply_merge_res(&FueledMergeRes { output }, &metrics.columnar);
                         }
                     }
                 }
--- a/src/persist-client/src/internal/trace.rs
+++ b/src/persist-client/src/internal/trace.rs
@@ -48,6 +48,8 @@
 //! [Batch::Merger]: differential_dataflow::trace::Batch::Merger
 
 use arrayvec::ArrayVec;
+use mz_persist::metrics::ColumnarMetrics;
+use mz_persist_types::Codec64;
 use std::cmp::Ordering;
 use std::collections::BTreeMap;
 use std::fmt::Debug;
@@ -562,16 +564,6 @@ impl<T: Timestamp + Lattice> Trace<T> {
         self.spine.validate()
     }
 
-    pub fn apply_merge_res(&mut self, res: &FueledMergeRes<T>) -> ApplyMergeResult {
-        for batch in self.spine.spine_batches_mut().rev() {
-            let result = batch.maybe_replace(res);
-            if result.matched() {
-                return result;
-            }
-        }
-        ApplyMergeResult::NotAppliedNoMatch
-    }
-
     /// Obtain all fueled merge reqs that either have no active compaction, or the previous
     /// compaction was started at or before the threshold time, in order from oldest to newest.
     pub(crate) fn fueled_merge_reqs_before_ms(
@@ -651,6 +643,22 @@ impl<T: Timestamp + Lattice> Trace<T> {
             }
         }
         metrics
+    }
+}
+
+impl<T: Timestamp + Lattice + Codec64> Trace<T> {
+    pub fn apply_merge_res(
+        &mut self,
+        res: &FueledMergeRes<T>,
+        metrics: &ColumnarMetrics,
+    ) -> ApplyMergeResult {
+        for batch in self.spine.spine_batches_mut().rev() {
+            let result = batch.maybe_replace(res, metrics);
+            if result.matched() {
+                return result;
+            }
+        }
+        ApplyMergeResult::NotAppliedNoMatch
     }
 }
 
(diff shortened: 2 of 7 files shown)


src/persist-client/src/internal/state_diff.rs
```diff
@@ -794,7 +794,7 @@ fn apply_diff_map<K: Ord, V: PartialEq + Debug>(
 // This might leave state in an invalid (umm) state when returning an error. The
 // caller ultimately ends up panic'ing on error, but if that changes, we might
 // want to revisit this.
-fn apply_diffs_spine<T: Timestamp + Lattice>(
+fn apply_diffs_spine<T: Timestamp + Lattice + Codec64>(
```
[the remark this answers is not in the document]
DAlperin: It plumbs down to https://github.com/MaterializeInc/materialize/pull/32605/files#diff-97c740ce5d3de94e4eb0e32af0126361287e0a4348ffc8f4a52413d4aad157b3R375

src/persist-client/src/internal/trace.rs
```diff
@@ -824,8 +853,183 @@ impl<T: Timestamp + Lattice> SpineBatch<T> {
...
+                .reduce(|a, b| match (a, b) {
+                    (Some(mut a), Some(b)) => {
+                        a.plus_equals(&b);
+                        Some(a)
+                    }
+                    _ => None,
+                })
+                .flatten();
+
+            if !res.output.parts.is_empty() {
```
bkirwi: This is vestigial, right? (This was introduced so we didn't fire on tombstones, but those go down another path now.)

If we don't need it we should definitely remove it - on this path returning an empty list of parts when the diff sum is nonzero would be a pretty gnarly bug that would be nice to catch.

src/persist-client/src/internal/trace.rs
```diff
@@ -824,8 +853,183 @@ impl<T: Timestamp + Lattice> SpineBatch<T> {
...
+        if exact_match {
+            *self = SpineBatch::merged(IdHollowBatch {
+                id: self.id(),
+                batch: Arc::new(res.output.clone()),
+            });
+            return ApplyMergeResult::AppliedExact;
+        }
+
+        if let Some((lower_idx, upper_idx, id_lower, id_upper)) = self.find_replacement_range(res) {
+            self.perform_subset_replacement(res, lower_idx, upper_idx, id_lower, id_upper)
```
bkirwi: IIUC subset replacement should never happen for a tombstone, right? This is dead code?

src/persist-client/src/internal/trace.rs
```diff
@@ -848,13 +1052,27 @@ impl<T: Timestamp + Lattice> SpineBatch<T> {
...
+        } else {
+            ApplyMergeResult::NotAppliedNoMatch
+        }
+    }
+
+    /// Find the range of parts that can be replaced by the merge result
+    fn find_replacement_range(
+        &self,
+        res: &FueledMergeRes<T>,
+    ) -> Option<(usize, usize, usize, usize)> {
```
bkirwi: This is a return type I will definitely eventually mess up... too easy to confuse all the `usize`s.

IIUC we could make this `Option<(SpineId, Range<usize>)>` without too much fuss, if that seems reasonable?

bkirwi: (Using the closed-open form of range specifically, since it'll make the math nicer elsewhere.)

src/persist-client/src/internal/compact.rs
```diff
@@ -156,7 +156,7 @@ where
     K: Debug + Codec,
     V: Debug + Codec,
     T: Timestamp + Lattice + Codec64 + Sync,
-    D: Semigroup + Ord + Codec64 + Send + Sync,
```
bkirwi: I think now that you've added `Debug` to `Codec64` some of these are no longer needed...

src/persist-client/src/internal/trace.rs
```diff
@@ -824,8 +853,183 @@ impl<T: Timestamp + Lattice> SpineBatch<T> {
...
+            return ApplyMergeResult::AppliedExact;
+        }
+
+        // Try subset replacement
+        if let Some((lower_idx, upper_idx, id_lower, id_upper)) = self.find_replacement_range(res) {
+            let old_diffs_sum = self
+                .parts
+                .iter()
+                .skip(lower_idx)
+                .take(upper_idx - lower_idx + 1)
```
bkirwi: The `skip` and `take` here would be more idiomatic as `self.parts[lower_idx..=upper_idx].iter()` or whatever!

src/persist-client/src/internal/trace.rs
```diff
@@ -824,8 +853,183 @@ impl<T: Timestamp + Lattice> SpineBatch<T> {
...
+                .flat_map(|p| p.batch.parts.iter())
+                .map(|p| p.diffs_sum::<D>(metrics))
+                .reduce(|a, b| match (a, b) {
+                    (Some(mut a), Some(b)) => {
+                        a.plus_equals(&b);
+                        Some(a)
+                    }
+                    _ => None,
+                })
+                .flatten();
```
bkirwi: There are a bunch of very similar blocks in this code now... time for a `diffs_sum<D>(parts: impl Iterator<...>) -> Option<D>` I think!

src/persist-client/src/internal/trace.rs
```diff
@@ -1808,7 +1999,7 @@ pub mod datadriven {
         let res = FueledMergeRes {
             output: DirectiveArgs::parse_hollow_batch(args.input),
         };
-        match datadriven.trace.apply_merge_res(&res) {
+        match datadriven.trace.apply_merge_res_unchecked(&res) {
```
bkirwi, line 1987: Do we have tests that make the checked version fail? 😬

bkirwi (commented):
Cool! Overall shape seems good.

There remains a bunch of almost-duplicated code with subtle differences, which makes me a bit nervous and is kind of hard to review. I made one suggestion at a small scale (summing diffs of parts) but if there are any other ideas you think have a good cost-benefit ratio I think they're probably worthwhile!

There's a fair bit of red in the nightlies, but I haven't taken a look in detail - let me know if you want a second pair of eyes there.

src/persist-client/src/internal/trace.rs
DAlperin: You would think wouldn't you! This subtle difference threw me for a loop. The apply code iterates through the SpineBatches and tries to apply. The tombstone code iterates through the HollowBatches of the parts _in_ the spinebatch. I think this is overkill and I would probably refactor the tombstone code a bit but I didn't feel like including it in this PR.

src/persist-client/src/internal/trace.rs
DAlperin: Best of my understanding we are just creating empty batches to test merging our descriptions here. There are no kvtd underlying them.

src/persist-client/src/internal/trace.rs
```diff
@@ -884,75 +1083,52 @@ impl<T: Timestamp + Lattice> SpineBatch<T> {
...
-                    parts: new_parts,
-                    active_compaction: None,
-                };
-                if new_spine_batch.len() > self.len() {
-                    return ApplyMergeResult::NotAppliedTooManyUpdates;
-                }
-                *self = new_spine_batch;
-                ApplyMergeResult::AppliedSubset
+            (Some((lower_idx, id_lower)), Some((upper_idx, id_upper))) => {
+                Some((SpineId(id_lower, id_upper), lower_idx..upper_idx))
```
bkirwi: Ah, this is much better but not quite right - `upper_idx` is inclusive, but the upper bound of this sort of range is supposed to be exclusive. Ideally we'd do `..(upper_idx+1)` here instead of in all the callers, but if we really want an inclusive upper here we should use `..=`!

(Sorry to nitpick, but this stuff compounds quickly IME.)

src/persist-client/src/internal/trace.rs
bkirwi: Mmm... so not really the code under test, and it looks like it's a hassle to call the checked version anyways. Fair enough!

bkirwi (approved):
Lots of code movement so this is a little hard to verify... but I think this is good! (And nightlies look pretty clean.) Appreciate all the cleanup.

We've verified that the fixed version of this still errors on the bad change in CI, yeah?

def-:
> We've verified that the fixed version of this still errors on the bad change in CI, yeah?

I did a run here: https://buildkite.com/materialize/nightly/builds/12213. Seeing lots of
```
platform-checks-materialized-1       | 2025-06-04T16:39:14.746851Z  thread 'persist:0008' panicked at src/persist-client/src/internal/trace.rs:1018:17: assertion `left == right` failed: merge res diffs sum (-1406225) did not match spine batch diffs sum (-919999)
```
so it seems good!

PR #32612, opened 2025-05-30
[webhooks] increase the HTTP requst body limit to 5MiB

ParkMyCar:
This PR increases the HTTP request body limit from 2MB to 5MiB. The original 2MB limit was picked arbitrarily and it should be safe to bump the limit up.

It also updates an existing test to ensure we'll get logging in prometheus if a request gets rejected for too large of a body.

### Motivation

Support larger webhook payloads

### Checklist

### DIFF
--- a/src/environmentd/src/http.rs
+++ b/src/environmentd/src/http.rs
@@ -92,7 +92,7 @@ pub use metrics::Metrics;
 pub use sql::{SqlResponse, WebSocketAuth, WebSocketResponse};
 
 /// Maximum allowed size for a request.
-pub const MAX_REQUEST_SIZE: usize = u64_to_usize(2 * bytesize::MB);
+pub const MAX_REQUEST_SIZE: usize = u64_to_usize(5 * bytesize::MIB);
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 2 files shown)


src/environmentd/src/http.rs
```diff
@@ -92,7 +92,7 @@ pub use metrics::Metrics;
 pub use sql::{SqlResponse, WebSocketAuth, WebSocketResponse};
 
 /// Maximum allowed size for a request.
-pub const MAX_REQUEST_SIZE: usize = u64_to_usize(2 * bytesize::MB);
+pub const MAX_REQUEST_SIZE: usize = u64_to_usize(5 * bytesize::MIB);
```
ptravers, line 95: only concern is switching to mebibytes. is it worth just remaining on megabytes?

ptravers (approved):
thought more about it and it makes sense to use base 2 rather than base 10 for a network request. envoy also works on mebi from memory.

src/environmentd/src/http.rs
ParkMyCar: Good call out, I updated the docs!

PR #32624, opened 2025-06-02
Add memory-request replica property

antiguru:
Add a `memory_request` property to replica allocations. While we normally set request to the limit to enforce guaranteed QoS scheduling, setting a separate request would allow us to switch to a lower QoS that would be compatible with swap.

Fixes MaterializeInc/database-issues/issues/9328

Context: https://kubernetes.io/blog/2025/03/25/swap-linux-improvements/

### DIFF
--- a/src/orchestrator-kubernetes/src/lib.rs
+++ b/src/orchestrator-kubernetes/src/lib.rs
@@ -959,8 +959,6 @@ impl NamespacedOrchestrator for NamespacedKubernetesOrchestrator {
                 image_pull_policy: Some(self.config.image_pull_policy.to_string()),
                 resources: Some(ResourceRequirements {
                     claims: None,
-                    // Set both limits and requests to the same values, to ensure a
-                    // `Guaranteed` QoS class for the pod.
                     limits: Some(limits.clone()),
                     requests: Some(requests.clone()),
                 }),
@@ -1161,8 +1159,6 @@ impl NamespacedOrchestrator for NamespacedKubernetesOrchestrator {
                     security_context: container_security_context.clone(),
                     resources: Some(ResourceRequirements {
                         claims: None,
-                        // Set both limits and requests to the same values, to ensure a
-                        // `Guaranteed` QoS class for the pod.
                         limits: Some(limits),
                         requests: Some(requests),
                     }),
--- a/src/orchestrator-kubernetes/src/lib.rs
+++ b/src/orchestrator-kubernetes/src/lib.rs
@@ -587,6 +587,7 @@ impl NamespacedOrchestrator for NamespacedKubernetesOrchestrator {
             args,
             ports: ports_in,
             memory_limit,
+            memory_request,
             cpu_limit,
             scale,
             labels: labels_in,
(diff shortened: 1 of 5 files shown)


src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -649,17 +650,32 @@ impl NamespacedOrchestrator for NamespacedKubernetesOrchestrator {
...
         if let Some(memory_limit) = memory_limit {
             limits.insert(
                 "memory".into(),
                 Quantity(memory_limit.0.as_u64().to_string()),
             );
+            requests.insert(
+                "memory".into(),
+                Quantity(memory_limit.0.as_u64().to_string()),
+            );
+        }
```
teskje, line 663: Noting that this leaves us in a bit of a weird state because the Mz replica size semantics don't match the k8s semantics anymore: In k8s if you specify a limit but no request, you get the `Burstable` QoS class. Whereas here when you specify a limit but no request, you get the `Guaranteed` QoS class. I imagine this might be confusing later.

antiguru:
Yes, the trick to enable swap is to set a request smaller than the limit.

My main concern around aligning the semantics of `memory_request` to what Kubernetes does is backwards compatibility. If someone has a replica size map for a current version of Materialize, they might expect the same behavior going forward, which is that without changing, the pods will still be scheduled as guaranteed QoS. This is what the PR currently implements.

We can break compatibility and align closer to Kubernetes' semantics by taking the `memory_request` verbatim, i.e., without specifying it we'll not request anything. Anyone who still needs guaranteed QoS needs to add the `memory_request` field at the appropriate time. I'm open to both solutions, and agree that this might come with less surprises in the future.

teskje:
What do you think about using the `cpu_exclusive` flag instead? That should be a backward compatible change, mostly. Replicas that currently set `cpu_exclusive = false` will lose the `Guaranteed` QoS class, but since they are not CPU pinning, that shouldn't be an issue. The argument I could see against it that it's too much implicit magic and we should be explicit instead.

There are a bunch of conflicting desires here:
1. We'd like to not break compatibility for existing replica sizes.
2. We'd like replica size configurations to be clear/obvious, not confusing.
3. We'd like the defaults to match the common case.

Maybe a good middle ground is to:
* Merge your PR as-is now, in line with (1) and (3).
* When/if we move to swap as a default, change the semantics of `memory_request = None` to the k8s semantics. This breaks (1) but ensures that existing replica sizes automatically support swap. If people want to keep using CPU pinning instead, they need to change their size definitions to explicitly specify `memory_request`.

src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -946,7 +962,7 @@ impl NamespacedOrchestrator for NamespacedKubernetesOrchestrator {
                     // Set both limits and requests to the same values, to ensure a
                     // `Guaranteed` QoS class for the pod.
```
teskje: Stale comment.

src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -1147,8 +1163,8 @@ impl NamespacedOrchestrator for NamespacedKubernetesOrchestrator {
                         claims: None,
                         // Set both limits and requests to the same values, to ensure a
                         // `Guaranteed` QoS class for the pod.
```
teskje: Stale comment.

src/orchestrator-kubernetes/src/lib.rs
alex-hunt-materialize: I prefer our semantics of `Guaranteed`. There are extremely few situations where `Burstable` class memory settings make sense.

alex-hunt-materialize (approved):

antiguru:
Thanks for the reviews!

PR #32629, opened 2025-06-02
build: don't disable asserts in librdkafka

teskje:
[librdkafka uses asserts to handle runtime errors.](https://github.com/confluentinc/librdkafka/issues/5099) If asserts are disabled, runtime errors can turn into memory safety bugs instead, which is something we should avoid.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9313

### Checklist

### DIFF
--- a/misc/bazel/c_deps/rust-sys/BUILD.librdkafka.bazel
+++ b/misc/bazel/c_deps/rust-sys/BUILD.librdkafka.bazel
@@ -30,6 +30,7 @@ cmake(
         # Note: casing here is important.
         "OpenSSL_ROOT": "$(execpath @openssl//:openssl_root)",
     },
+    copts = ["-UNDEBUG"],
     generate_args = [
         # Features enabled by the Rust `rdkafka-sys` crate.
         "-DRDKAFKA_BUILD_STATIC=1",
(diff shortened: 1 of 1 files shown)


teskje:
Proof that this works with a branch that intentionally triggers an assert in librdkafka: https://github.com/MaterializeInc/materialize/pull/32625

antiguru (approved):

misc/bazel/c_deps/rust-sys/BUILD.librdkafka.bazel
```diff
@@ -55,6 +56,12 @@ cmake(
         "-DCMAKE_FIND_USE_CMAKE_ENVIRONMENT_PATH=0",
         "-DCMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH=0",
         "-DCMAKE_FIND_USE_CMAKE_SYSTEM_PATH=0",
+        # When builing with `CMAKE_BUILD_TYPE=Release`, cmake automatically
+        # sets `-DNDEBUG` flags. This breaks librdkafka, which uses asserts for
+        # runtime error checking (confluentinc/librdkafka#5099). We unset
+        # `NDEBUG` by manually overwriting the cmake's `C_FLAGS` variables.
+        "-DCMAKE_C_FLAGS_RELEASE=-O3",
```
ParkMyCar, line 63: We probably also want to include `-fPIC` to compile position independent code.

FWIW if you didn't want to always set `-O3` we could create a `build_mode` flag ([example](https://github.com/MaterializeInc/materialize/blob/main/misc/bazel/platforms/BUILD.bazel#L21-L36)) and here use a `select({ ... })` to use different options for the different modes.

As-is though if this fixes the issue then let's ship it!

ParkMyCar (approved):

misc/bazel/c_deps/rust-sys/BUILD.librdkafka.bazel
teskje: > We probably also want to include `-fPIC` to compile position independent code.

Might be! Does position-independent code do anything for static libraries?

> FWIW if you didn't want to always set `-O3` we could create a `build_mode` flag (example) and here use a `select({ ... })` to use different options for the different modes.

This is a bit weird: Somehow we are always building librdkafka with `CMAKE_BUILD_TYPE=Release`, no matter what optimization level is selected in bazel. The `CMAKE_*_FLAGS_RELEASE` options only apply to release builds, so we could add a `build_mode` flag like you suggest to select a different `CMAKE_BUILD_TYPE`. But compiling seems fast enough either way, so not too worried about that atm.

PR #32634, opened 2025-06-03
tests: make more sqllogictests resilient to result ordering

aljoscha:
More work towards https://github.com/MaterializeInc/database-issues/issues/9180.

Changing the way we deliver results can change their ordering, which is okay by the SQL standard. This prepares more tests for that.


test/sqllogictest/transform/literal_constraints.slt
```diff
@@ -1351,33 +1351,8 @@ EOF
...
-# Note: This has LIMIT but no ORDER BY, so this is ok to change when row representation changes.
-query IT
-SELECT a,b
-FROM t1
-WHERE a IN (4,3,2,1)
-LIMIT 1;
-----
-1  a
+# TODO: Is it okay to remove this test? Without ORDER BY the result is really
+# arbitrary, so not deterministically testable.
```
ggevay: Ok!

ggevay (approved):

PR #32637, opened 2025-06-03
self-managed: Rename lts to self-managed internally

def-:
### Checklist

### DIFF
--- a/ci/nightly/pipeline.template.yml
+++ b/ci/nightly/pipeline.template.yml
@@ -876,8 +876,8 @@ steps:
               composition: platform-checks
               args: [--scenario=UpgradeEntireMz, "--seed=$BUILDKITE_JOB_ID"]
 
-      - id: checks-lts-upgrade
-        label: "Checks LTS upgrade, whole-Mz restart"
+      - id: checks-self-managed-upgrade
+        label: "Checks Self-Managed upgrade, whole-Mz restart"
         depends_on: build-aarch64
         timeout_in_minutes: 180
         parallelism: 3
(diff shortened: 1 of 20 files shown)


doc/user/config.toml
```diff
@@ -213,4 +213,4 @@ include = "docs/**"
 
 # Avoid deleting the LTS and self-managed docs which are deployed from the
 # lts-docs and self-managed-docs branches.
-exclude = "docs/{lts,self-managed/v25.1}/**"
+exclude = "docs/{lts,self-managed/v25.1,self-managed/v25.2}/**"
```
alex-hunt-materialize, line 216: I assume we can't remove these yet.

ci/nightly/pipeline.template.yml
```diff
@@ -1263,7 +1263,7 @@ steps:
             run: aws-temporary
             # Cleanup runs in pre-exit hook
             args: [--no-cleanup]
-      branches: "main v*.* lts-v* *aws* *tf* *terraform* *helm* *self-managed* *orchestratord*"
```
alex-hunt-materialize, line 1266: We still have these lts branches. Are they no longer in use?

Specifically we have `lts-v0.130`. Is this not where we build environmentd/clusterd patches of the current release from?

doc/user/config.toml
def-: They are from another LTS release we did long ago: https://materialize.com/docs/lts/ (v0.26.0). I guess we can remove them if we want to. I have never heard of anyone using this.

ci/nightly/pipeline.template.yml
def-: Not in use, we can remove the lts-v0.130 branch.

doc/user/config.toml
alex-hunt-materialize: Oh, v0.26.0 is the binary before we split for platform v1. We probably still have people using it, but we don't want to support them, and probably don't care about hosting docs pointing them that way. That's more of a product question, though.

alex-hunt-materialize (approved):

PR #32640, opened 2025-06-03
k8s: fetch disk usage metrics from clusterd

teskje:
This PR changes the way disk usage metrics are collected in Materialize. Instead of getting them from the k8s API, envd now connects to a new metrics endpoint exposed by the clusterd process. Clusterd collects disk usage metrics by running `statvfs` on the `/scratch` directory, which is equivalent to invoking the `df` command.

The motivation for this change is to simplify our metrics infrastructure. The k8s API providing disk usage metrics is a custom API [with quite a lot of moving parts under the hood](https://materializeinc.slack.com/archives/C07PN7KSB0T/p1741365366537109?thread_ts=1741364508.509219&cid=C07PN7KSB0T). One consequence of the complexity is that disk usage reporting does not work with self-managed Materialize, which is something this PR should fix.

### Motivation

Part of https://github.com/MaterializeInc/database-issues/issues/9243
Closes https://github.com/MaterializeInc/database-issues/issues/9217

### Tips for reviewer

I have verified that the change works in my staging env. I also verified by running self-managed in a local kind cluster. The disk usage showed up but was way too high, which I attribute to the disk setup being strange in kind.

### Checklist


src/clusterd/src/usage_metrics.rs
```diff
@@ -0,0 +1,58 @@
...
+            Ok(stat) => stat,
+            Err(err) => {
+                error!("statvfs error: {err}");
+                return None;
+            }
+        };
+
+        // `fsblkcnt_t` is a `u32` on macOS but a `u64` on Linux.
+        #[allow(clippy::useless_conversion)]
+        let used_blocks = u64::from(stat.blocks() - stat.blocks_free());
```
teskje: Something to discuss: [statvfs](https://man7.org/linux/man-pages/man3/statvfs.3.html) returns two different free block values:
* `f_bfree`: Number of free blocks
* `f_bavail`: Number of free blocks for unprivileged users

We could use `blocks_available()` here instead of `blocks_free()`. Apparently the space reserved for root is 5% by default, but that might not be the case in our k8s setup.

alex-hunt-materialize: +1 to using available.

teskje: Looks like in cloud we only reserve 16MiB, so it doesn't really matter either way:

```
materialize@mz6tylgpv9fe-cluster-u1-replica-u1-gen-2-0:/$ df /scratch
Filesystem                                                                    1K-blocks    Used Available Use% Mounted on
/dev/mapper/instance--store--vg-pvc--e3579d60--2fe8--48c2--94f3--eec2b60c4155   8154588 1334568   6803636  17% /scratch
```

I'll change to available because of Alex' vote and because it's consistent with `df`.

src/clusterd/src/usage_metrics.rs
antiguru, line 1: Consider moving this file to `mz-metrics`, so it would be accessible for envd, too. But don't block on it.

src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -1674,6 +1514,48 @@ impl OrchestratorWorker {
...
+                        .find(|p| p.name == Some("internal-http".into()))
+                })
+                .map(|p| p.port);
+            let Some(port) = internal_http_port else {
+                bail!("internal-http port missing in service spec");
+            };
+            let metrics_url = format!(
+                "http://{service_name}-{i}.{service_name}.{namespace}.svc.cluster.local:{port}\
+                 /api/usage-metrics"
+            );
```
antiguru, line 1550: I think this is fine for this PR, but we should revisit how information flows between envd and cluster replicas. In theory, we have a protocol that can handle all of it (the cluster protocol), but it's not quite enough to enable collecting these metrics. And then, we have prom metrics, regular queries, introspection subscribes, which all serve similar purposes. Do we really want to have multiple distinct interfaces, or should we restructure it to be narrower?

If you agree, do you mind filing an issue to capture this thought?

src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -1674,6 +1514,48 @@ impl OrchestratorWorker {
                 disk_usage_bytes: disk_usage,
             }
         }
+
+        async fn get_disk_usage(
```
antiguru, line 1523: Nit: Please document :)

src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -1674,6 +1514,48 @@ impl OrchestratorWorker {
...
+            let Some(port) = internal_http_port else {
+                bail!("internal-http port missing in service spec");
+            };
+            let metrics_url = format!(
+                "http://{service_name}-{i}.{service_name}.{namespace}.svc.cluster.local:{port}\
+                 /api/usage-metrics"
+            );
+
+            let http_client = reqwest::Client::builder()
+                .timeout(Duration::from_secs(10))
```
antiguru, line 1553: Does the timeout need to be configurable?

antiguru (approved):
Looks good, left some comments, but nothing that would require another round of review! Thank you

src/orchestrator-kubernetes/src/lib.rs
teskje: I actually started with transmitting this information through the compute protocol, but that approach is worse for multiple reasons:
* More parts have to be involved: The Timely workers, the protocol clients, the controller, and possibly the coordinator. This ends up being more code and complexity.
* Since the metrics are collected per process, in processes with multiple workers the workers have to either elect a leader or coordinate on who reports new metric values. That's not necessary when there is only a single endpoint per process that's queried by the orchestrator.
* Since Timely workers are involved in reporting the disk usage metrics, metrics reporting can stall if the workers are overloaded. Fetching the metrics through an HTTP endpoint is more reliable.
* If we transmit the disk usage through the compute protocol, but still let the orchestrator collect CPU and memory usage through the kube API, writing the combined data into system tables becomes more cumbersome.

I appreciate the concern about having multiple metrics path. Because of the worker overload issue it's imo a good idea to report metrics outside of the worker logic as much as we can. So I wouldn't want to report everything through introspection subscribes. But we can consider making the disk usage metrics Prometheus metrics, and doing so would allow us to delete a bunch of infrastructure in cloud I think.

src/orchestrator-kubernetes/src/lib.rs
teskje: Not sure! I cargo-culted this from a bunch of other `reqwest::Client` usages in the repo that all hardcode the timeout. I think it's mostly important that we set a timeout at all, hard to come up with a scenario in which we'd want to adjust it.

src/orchestrator-kubernetes/src/lib.rs
teskje: Opened an issue for the Prometheus metrics: https://github.com/MaterializeInc/database-issues/issues/9348

src/clusterd/src/usage_metrics.rs
teskje: From what I can tell that crate only contains Prometheus metrics so far. I'll make the move as part of https://github.com/MaterializeInc/database-issues/issues/9348.

teskje:
TFTRs!

bosconi:
bugbot run

src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -1674,6 +1514,53 @@ impl OrchestratorWorker {
...
+            service_name: &str,
+            i: usize,
+        ) -> anyhow::Result<u64> {
+            #[derive(Deserialize)]
+            pub(crate) struct Usage {
+                disk_bytes: Option<u64>,
+            }
+
+            let service = self_.service_api.get(service_name).await.unwrap();
+            let namespace = service.metadata.namespace.unwrap();
```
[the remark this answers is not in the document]
teskje: Already fixed in https://github.com/MaterializeInc/materialize/pull/32699

PR #32644, opened 2025-06-04
Add memory-request replica property #32624

antiguru:
#32624 for self-managed.


def-:
I added a the CI workaround commits, we should also add them if we cut a v0.130.14 (and a few more PRs that we'd have to pick together: https://github.com/MaterializeInc/materialize/issues?q=label%3Alts-backport-v25.1%20-label%3Alts-backport-v25.1-done)

PR #32650, opened 2025-06-04
Move statement lifecycle begin up to Adapter frontend

ggevay:
This PR implements the first part of https://github.com/MaterializeInc/materialize/pull/32504, namely the moving of `began_at` (and the corresponding `execution-began` event in the lifecycle table) up to the Adapter frontend (`protocol.rs`) from the Coordinator. For this, we pass down the timestamp from the Adapter frontend to the Coordinator through the Portal.

The `LifecycleTimestamps` struct contains only one field for now, but I plan to add more things to it later.

Testing is a bit hard for this, but I did the following:
- A minor thing is that I've added a test in `statement-logging.td` to test that `mz_statement_execution_history.began_at` is always exactly the same as `mz_statement_lifecycle_history.occurred_at` on the `execution-began` event. On main, these were close, but not always exactly the same, as they were determined in separate calls to `self.now()`. (I've verified that the test indeed doesn't always pass on main.)
- Verifying that `began_at` is taken in the Adapter frontend is a bit harder. I couldn't think of a regression test for this (I'm open to ideas), but I did some tests with extra code that is not possible to merge:
  - If I put a `panic!` in the `else` branch of the `let began_at = if let Some(lifecycle_timestamps) = lifecycle_timestamps`, the panic doesn't happen while running `statement-logging.td` or when running slts or when running queries manually.
  - I've put a sleep at the beginning of the `Some(FrontendMessage::Execute {` match arm in `protocol.rs`, and verified that `finished_at - began_at` includes that sleep on the PR, but not on `main`. (Note that Testdrive submits queries using the Extended Query flow.)

### Motivation

  * This PR adds a known-desirable feature: Works towards https://github.com/MaterializeInc/materialize/pull/32504

### Checklist


src/adapter/src/session.rs
```diff
@@ -990,6 +994,24 @@ impl InProgressRows {
...
+    /// When the query was received. More specifically, when the tokio recv returned.
+    /// For a Simple Query, this is for the whole query, for the Extended Query flow, this is only
+    /// for `FrontendMessage::Execute`. (This means that this is after parsing for the
+    /// Extended Query flow.)
+    pub received: EpochMillis,
+}
+
+impl LifecycleTimestamps {
+    /// Creates a new `LifecycleTimestamps`.
+    pub fn new(received: EpochMillis) -> Self {
```
aljoscha, line 1010: Not something you have to change now, but I think this `new()` will become unwieldy in the future, when the parameters are just a bunch of `EpochMillis`. Might be better to just always initialize using `LifecycleTimestamps { this_timestamp, that_timestamp, etc. }`.

ggevay: Ok, I'll keep this in mind when adding more fields!

ggevay:
Thanks for the quick review!

PR #32655, opened 2025-06-04
[persist] Upgrade `arrow` to fix a minor source roundtripping issue

bkirwi:
Upgrading to pick up this upstream arrow fix: https://github.com/apache/arrow-rs/pull/6953

### Motivation

https://github.com/MaterializeInc/database-issues/issues/9318

### Tips for reviewer

There were a couple minor API changes to adapt to as well, due to the upgrade, which explains most of the diff. (The largest is that `chrono` added a new method to a trait, shadowing our own method - I've adapted the code to use the upstream version instead.)

I've committed the proptest failure, though note that it's still hard to reproduce even so... the bug only appears when two floats hash to the same bucket in a hash table, which is not common. I could only repro the original bug after several hours of `cargo stress`; so while I'm pretty confident this is the fix, it's hard to prove with testing.

### Checklist


bkirwi:
Nightly run: https://buildkite.com/materialize/nightly/builds/12214

The only failure seems to already exist on `main`.

PR #32674, opened 2025-06-06
orchestratord: Don't pass in sql-listen-addr and http-listen-addr, also fix terraform test

def-:
Follow-up to https://github.com/MaterializeInc/materialize/pull/32245

Test failures seen in https://buildkite.com/materialize/nightly/builds/12232

> error: unexpected argument '--sql-listen-addr' found

### Checklist

### DIFF
--- a/src/orchestratord/src/controller/materialize/environmentd.rs
+++ b/src/orchestratord/src/controller/materialize/environmentd.rs
@@ -1381,11 +1381,6 @@ fn create_environmentd_statefulset_object(
         }
     } else {
         args.extend([
-            format!("--sql-listen-addr=0.0.0.0:{}", config.environmentd_sql_port),
-            format!(
-                "--http-listen-addr=0.0.0.0:{}",
-                config.environmentd_http_port
-            ),
             format!(
                 "--internal-sql-listen-addr=0.0.0.0:{}",
                 config.environmentd_internal_sql_port
--- a/ci/nightly/pipeline.template.yml
+++ b/ci/nightly/pipeline.template.yml
@@ -1232,7 +1232,7 @@ steps:
       plugins:
         - ./ci/plugins/mzcompose:
             composition: cloud-canary
-      branches: "main v*.*"
+      branches: "main v*.* *orchestratord*"
 
     - id: mz-e2e
       label: "Mz E2E Test"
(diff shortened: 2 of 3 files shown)


def-:
Apparently this is not enough:
```
$ kubectl apply -f -
The Materialize "12345678-1234-1234-1234-123456789012" is invalid: spec.authenticatorKind: Required value
```
@alex-hunt-materialize Do you have any idea? I guess I should just pass in `authenticatorKind: None`
Have we updated this in our documentation already?

def-:
Success: https://buildkite.com/materialize/nightly/builds/12241

src/orchestratord/src/controller/materialize/environmentd.rs
```diff
@@ -1381,11 +1381,6 @@ fn create_environmentd_statefulset_object(
         }
     } else {
         args.extend([
-            format!("--sql-listen-addr=0.0.0.0:{}", config.environmentd_sql_port),
-            format!(
-                "--http-listen-addr=0.0.0.0:{}",
-                config.environmentd_http_port
-            ),
```
jubrad, line 1388: hmmm, I think we have these changes in the cloud as well... do they need to be removed there?

After looking into this, I really think these shouldn't be removed?!

def-: I'm confused about this as well. It seems like v0.146.0 will otherwise fail to start up:
> error: unexpected argument '--sql-listen-addr' found

jubrad: yeah, there's some code that should version gate this, so anything >= V146_DEV0 shouldn't get these flags

jubrad: I was able to run ` bin/mzcompose --find terraform run aws-temporary --no-cleanup`  and it worked with just the spec: authenticatorType changes

jubrad: I'll mess around with some kind deploys to make sure!

jubrad:
subsumed by 
https://github.com/MaterializeInc/materialize/pull/32680

Not a release blocker since we reverted
https://github.com/MaterializeInc/materialize/pull/32245

alex-hunt-materialize:
These args definitely should not be deleted. They are required for supporting older versions.

PR #32676, opened 2025-06-06
storage: correct source statistics for multi-replica clusters

aljoscha:
This fixes source statistics to work okay when there are multiple
replicas. See comments in the code for how this fixes things.

Implements https://github.com/MaterializeInc/database-issues/issues/9010

WIP: @def- I still need to add a test that has sources and multi-replica clusters.


aljoscha:
@martykulma & @petrosagg how do we like this approach?

aljoscha:
Updated to also include the same implementation for sinks

def-:
Rebased to pick up recent CI autoscaler changes.

doc/user/content/sql/system-catalog/mz_internal.md
```diff
@@ -1021,13 +1021,14 @@ Note that:
...
-| `messages_staged`    | [`uint8`] | The number of messages staged but possibly not committed to the sink.                                               |
-| `messages_committed` | [`uint8`] | The number of messages committed to the sink.                                                                       |
-| `bytes_staged`       | [`uint8`] | The number of bytes staged but possibly not committed to the sink. This counts both keys and values, if applicable. |
-| `bytes_committed`    | [`uint8`] | The number of bytes committed to the sink. This counts both keys and values, if applicable.                         |
+| Field                | Type      | Meaning                                                                                                                        |
+|----------------------|-----------| --------                                                                                                                       |
+| `id`                 | [`text`]  | The ID of the sink. Corresponds to [`mz_catalog.mz_sources.id`](../mz_catalog#mz_sinks).                                       |
```
kay-kim: Should this be [`mz_catalog.mz_sinks.id`]?

kay-kim (approved):
docs change lgtm - mod one comment

src/storage-controller/src/collection_mgmt.rs
```diff
@@ -656,7 +658,7 @@ where
                 .await;
 
                 let scraper_token =
-                    statistics::spawn_statistics_scraper::<_, SinkStatisticsUpdate, _>(
+                    statistics::spawn_statistics_scraper::<_, ControllerSinkStatistics, _>(
```
martykulma: nit: I don't believe you need to specify the types here, rust should be able to infer this as above for source statistics.

doc/user/content/sql/system-catalog/mz_internal.md
aljoscha: That is exactly right! And thanks for taking a look 🙇‍♂️

src/storage-controller/src/lib.rs
```diff
@@ -2272,22 +2273,75 @@ where
...
+                                ?source_stats,
+                                "missing replica_id for source statistics update"
+                            );
+                            continue;
+                        };
+
                         for stat in source_stats {
-                            // Don't override it if its been removed.
-                            shared_stats
+                            // tracing::info!(?stat, "new stats");
```
martykulma: still needed?

aljoscha: sorry, still somewhat in progress as I make sure this passes ci/nightly 🙈

aljoscha:
@martykulma and maybe @petrosagg, this should now be ready for review. I updated the main description above to explain what I did. I didn't yet go through all the code myself, so there might be typos and some debugging leftovers. Sorry about that, but I ran out of steam for today. I'll review myself first thing Monday but wanted to present this now in case you fancy taking a look early.

src/storage-controller/src/lib.rs
```diff
@@ -2262,32 +2255,85 @@ where
...
+
+                        let replica_id = if let Some(replica_id) = replica_id {
+                            replica_id
+                        } else {
+                            tracing::error!(
+                                ?source_stats,
+                                "missing replica_id for source statistics update"
+                            );
+                            continue;
+                        };
```
martykulma, line 2275: If looks like we could move this check before `self.source_statistics.lock()` to avoid the lock.

nit: technically, you don't need to extract the `replica_id`, since it's just wrapped in `Some` a few lines down.  Then you can just pass `replica_id` instead of `Some(replica_id)`

```suggestion
                        if replica_id.is_none() {
                            tracing::error!(
                                ?source_stats,
                                "missing replica_id for source statistics update"
                            );
                            continue;
                        };
```

src/storage-controller/src/lib.rs
```diff
@@ -2262,32 +2255,85 @@ where
...
+
+                        let replica_id = if let Some(replica_id) = replica_id {
+                            replica_id
+                        } else {
+                            tracing::error!(
+                                ?sink_stats,
+                                "missing replica_id for sink statistics update"
+                            );
+                            continue;
+                        };
```
martykulma: ditto for moving this before the lock

could also give it the same `is_none()` treatment as above (but then you'd have to unwrap), or just remove the `if`
```suggestion
                        let Some(replica_id) = replica_id else {
                            tracing::error!(
                                ?sink_stats,
                                "missing replica_id for sink statistics update"
                            );
                            continue;
                        };

```

test/testdrive-old-kafka-src-syntax/indexes.td
```diff
@@ -482,3 +482,107 @@ pg_class_all_databases_ind                                  pg_class_all_databas
 pg_description_all_databases_ind                            pg_description_all_databases                 mz_catalog_server    {objoid,classoid,objsubid,description,oid_database_name,class_database_name}    ""
 pg_namespace_all_databases_ind                              pg_namespace_all_databases                   mz_catalog_server    {nspname}                                   ""
 pg_type_all_databases_ind                                   pg_type_all_databases                        mz_catalog_server    {oid}                                       ""
+
+>[version>=14800] SHOW INDEXES IN CLUSTER mz_catalog_server
+mz_active_peeks_per_worker_s2_primary_idx                   mz_active_peeks_per_worker                   mz_catalog_server    {id,worker_id}                              ""
+mz_arrangement_batches_raw_s2_primary_idx                   mz_arrangement_batches_raw                   mz_catalog_server    {operator_id,worker_id}                     ""
```
martykulma, line 488: did we just miss adding these in the past?

aljoscha: which one of "these" specifically did you mean? 😅

martykulma: lol.. well I didn't want to highlight most of them, but it looks like a lot that have nothing to do with this PR 😀

aljoscha: ahh, this is copying a whole block and slapping on the `>[version>=14800]` filter. It's a testdrive thing. Because the old version would not emit a `replica_id` column for certain builtins but the new version (after this PR) does.

Does that make sense now? 😅

src/storage-controller/src/lib.rs
aljoscha: Yeah, this whole unrapping, and then putting in a `Some` again business seems weird. But I did it because in this place we expect there to always be a replica_id. We only support `Option` keys because of webhook "sources", which don't run on a replica. I'll add a comment about that!

test/testdrive-old-kafka-src-syntax/indexes.td
martykulma: 🤦 yeah, sorry - I needed to look at the whole file to realize why the giant section!!

src/storage-controller/src/lib.rs
martykulma: Good idea, that'll definitely help out future us!! 

Sorry, I think some of what I was saying got lost because part of it is in the code snippet 😅 .  If the conditional statement calls `continue` when `replica_id.is_none()` then `replica_id` is guaranteed to be `Some()` lower and you won't need to re-wrap.

aljoscha: ahh sorry!

PR #32692, opened 2025-06-09
adapter,storage: allow webhook sources on multi-replica clusters

aljoscha:
The restriction was artificial, webhooks don't actually run on clusters.

Implements https://github.com/MaterializeInc/database-issues/issues/9181

@def- Do we want more than this in terms of testing?

### DIFF
--- a/misc/python/materialize/checks/all_checks/sink.py
+++ b/misc/python/materialize/checks/all_checks/sink.py
@@ -1268,7 +1268,8 @@ class AlterSinkWebhook(Check):
         return Testdrive(
             dedent(
                 """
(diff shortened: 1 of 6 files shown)


def- (approved):
I have a few tests where I have disabled multi-replica clusters because they contain webhook sources, will revisit them.

def-:
I have pushed the test changes directly into this PR and triggered nightly: https://buildkite.com/materialize/nightly/builds/12250 Edit: https://buildkite.com/materialize/nightly/builds/12251 (green)

aljoscha:
tyty! 🙇‍♂️

PR #32695, opened 2025-06-09
add minDomain for cluster topologySpread

jubrad:
Without a minDomain a topology spread constraint may place pods on only the subset of topology keys
where an existing node is provisioned. If we want to spread equally across all support availability zones rather than all availability zones that currently have nodes, then we need to set a minDomain value equal to the number of supported availability zones.

Perhaps a more appropriate way to handle this would be use the availability_zone's provided to environmentd. This was easier to throw together, it's more flexible, but less intuitive and easier to mess up.

There's some added trickiness to this one could imagine, where, for self-managed, this is not a global static, but a value that is dictated by the AZs available on the node_groups servicing a particular cluster or cluster size.

### Motivation
We don't quite spread evenly.

### Checklist

### DIFF
--- a/src/sql/src/session/vars/definitions.rs
+++ b/src/sql/src/session/vars/definitions.rs
@@ -1613,6 +1613,18 @@ pub mod cluster_scheduling {
         false,
     );
 
+    // `minDomains`, like maxSkew, is used to spread across a topology
+    // key. Unlike max skew, minDomains will force node creation to ensure
+    // distribution across a minimum number of keys.
+    // https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#spread-constraint-definition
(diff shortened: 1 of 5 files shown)


src/sql/src/session/vars/definitions.rs
```diff
@@ -1613,6 +1613,13 @@ pub mod cluster_scheduling {
         false,
     );
 
+    pub static CLUSTER_TOPOLOGY_SPREAD_MIN_DOMAINS: VarDefinition = VarDefinition::new(
+        "cluster_topology_spread_min_domains",
+        value!(Option<i32>; None),
+        "`minDomains` for replica topology spread constraints (Materialize).",
```
alex-hunt-materialize: Can we add to this comment that this should be equal to the number of AZs karpenter can launch materialize-instance nodes into for that region?

PR #32696, opened 2025-06-09
TEMP docs: WIP aws terraform updates

kay-kim:
Additional work will come as https://github.com/MaterializeInc/terraform-aws-materialize/pull/69 evolves; e.g.,

- Update just the links to the terraforms
- Update the annotations in the appendix installs for older versions (if deprecated/etc.)
- Do the same for the other cloud providers

Then, add to the v25.1 version of the docs as well.

Placing here to give an idea of what it could look like in terms of providing older instructions, etc.


kay-kim:
Closing as the cloud TF strategy has changed ... See https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/1

PR #32701, opened 2025-06-10
cluster: add timeouts to network calls in `create_sockets`

teskje:
In the `create_sockets` protocol, if a process goes away after a connection has been established but before the protocol handshake has finished, and the process fails to reset the connection (e.g. during a network partition), its peer can get stuck trying to finish the handshake. A process stuck that way deadlocks the whole protocol.

This commit prevents the described deadlock by introducing timeouts for all handshake network calls. Now the peer aborts the handshake and connection instead of getting stuck, so the protocol can continue to make progress.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9365

### Checklist


src/cluster/src/communication.rs
```diff
@@ -624,8 +639,8 @@ mod turmoil_tests {
             }
 
             sim.step().unwrap();
-            if sim.elapsed() > Duration::from_secs(120) {
-                panic!("simulation not finished after 120s");
+            if sim.elapsed() > Duration::from_secs(60) {
+                panic!("simulation not finished after 60s");
```
teskje, line 651: I checked and the test failure that motivated the 60s->120s bump (https://github.com/MaterializeInc/materialize/pull/32306) was caused by the same bug, so it should be safe to unbump.

src/cluster/src/communication.rs
```diff
@@ -348,10 +348,10 @@ async fn connect_lower(
             tcp.set_nodelay(true)?;
         }
 
-        s.write_u64(u64::cast_from(my_index)).await?;
-        let peer_epoch = Epoch::read(&mut s).await?;
+        timeout(s.write_u64(u64::cast_from(my_index))).await?;
```
antiguru, line 354: I'm not sure the timeouts on `write` make sense, but I might be mistaken! We're writing to a socket, which sends data asynchronously, unless the send buffer is full. In this case, the timeout could only ever be reached if the send buffer is full, which seems somewhat unlikely?

The timeout for _read_ makes more sense to me!

teskje: That's true. I just added a timeout to every `.await` around here (expect for `accept`/`connect`), to make sure I'm not forgetting anything. The timeouts for writes don't hurt so I'm inclined to leave them in. Unless you think they are more confusing than they help?

A thing we might find if we remove them is that turmoil doesn't simulate the send buffer and does wait for the peer to accept a write. In which case we'd need to timeout to avoid a flaky test. But I haven't tested that yet!

antiguru: Works for me. Do you mind leaving a comment in the code as to why we also have timeouts on the send path? Otherwise LGTM.

antiguru (approved):
Thanks!

src/cluster/src/communication.rs
teskje: Done!

teskje:
TFTR!

PR #32712, opened 2025-06-11
storage/pg: switch to a per-row round-robin to ensure even distribution

petrosagg:
We recently observed that distributing data container-by-container doesn't work well if the size of containers isn't uniform and can lead to OOMs if the distributing worker always happens to distribute a small container to itself and a large container to everyone else.

We don't yet know why the oscillation occurs but until then this PR should help with even distribution of work regardless of the size patterns.

I have also combined the exchange of data with the casting of rows in a single operator.

### Checklist


src/storage/src/source/postgres/snapshot.rs
```diff
@@ -556,31 +555,37 @@ pub(crate) fn render<G: Scope<Timestamp = MzOffset>>(
     let mut text_row = Row::default();
     let mut final_row = Row::default();
     let mut datum_vec = DatumVec::new();
+    let mut next_worker = (0..scope.peers()).cycle();
+    let round_robin = Exchange::new(move |_| u64::cast_from(next_worker.next().unwrap()));
```
martykulma: rather than casting on every call to `next()`, could cast `scope.peers()` to u64 once.

petrosagg: Sure. It's a no-op anyway but I'll move it there

petrosagg:
> Is the plan to do this in replication as well once we're sure it solves the problem in snapshot?

That was not my plan because it didn't cross my mind! I will change the replication too

PR #9302, opened 2021-11-26
testdrive: add a new `psql-execute` action

benesch:
This command will be useful for writing integration tests for the
various backslash commands in psql. This commit adds only documentation
and one sample use of the action.

### Motivation

  * This PR adds a feature that has not yet been specified.

### Checklist

- [x] This PR has adequate test coverage / QA involvement has been duly considered.
- [x] This PR adds a release note for any
  [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/user/content/release-notes.md#what-changes-require-a-release-note).

PR #32713, opened 2025-06-11
storage/pg: ignore main source output in resume upper calculation

petrosagg:
The "main source output" never contains any data and therefore its upper is irrelevant when determining at which point to resume reading the replication stream. Separately, the assertion that it always exists in the configuration isn't always true, leading to flaky CI tests.

Fixes #9302

### Checklist

### DIFF
--- a/src/storage/src/source/postgres/replication.rs
+++ b/src/storage/src/source/postgres/replication.rs
@@ -92,7 +92,6 @@ use mz_repr::{Datum, DatumVec, Diff, Row};
 use mz_sql_parser::ast::{Ident, display::AstDisplay};
 use mz_storage_types::dyncfgs::{PG_OFFSET_KNOWN_INTERVAL, PG_SCHEMA_VALIDATION_INTERVAL};
 use mz_storage_types::errors::DataflowError;
-use mz_storage_types::sources::SourceTimestamp;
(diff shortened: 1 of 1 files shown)


src/storage/src/source/source_reader_pipeline.rs
```diff
@@ -551,20 +551,17 @@ where
...
-                        Some(Probe {
-                            probe_ts: now_fn().into(),
-                            upstream_frontier: Antichain::new(),
-                        })
-                    });
+                    .await;
+                new_probe = match probe {
+                    Ok(probe) => (*probe).clone(),
+                    Err(_) => break,
+                };
```
teskje: Did you intend to include this change here?

petrosagg: Yes, it's the thing I asked you about in slack. Without this change if we transiently instantiate a source with no subsources we mint a binding for the empty frontier. And then any later added subsource gets stuck in reclocking because the frontier never advances to the empty frontier

def-:
Rebased to see if test failures persist. Edit: still failing: https://buildkite.com/materialize/test/builds/106881

petrosagg:
this was fixed by the create table from source work

PR #32717, opened 2025-06-11
Clean up type constraints in compute, introduce wrapper traits

antiguru:
Introduce `MzData` complementing other data traits, and constraining implementors to support columnation and columnar, with the right constraints on the GATs. Introduce a `MzTimestamp` extending `MzData`. This allows us to clean up and simplify some type definitions used throughout rendering.

### DIFF
--- a/src/compute/src/typedefs.rs
+++ b/src/compute/src/typedefs.rs
@@ -123,12 +123,16 @@ pub type KeyValBatcher<K, V, T, D> =
     MergeBatcher<Vec<((K, V), T, D)>, ColumnationChunker<((K, V), T, D)>, ColMerger<(K, V), T, D>>;
 
 /// Timestamp trait for rendering, constraint to support [`MzData`] and [timely::progress::Timestamp].
-pub trait MzTimestamp: MzData + timely::progress::Timestamp {}
+pub trait MzTimestamp:
+    MzData + timely::progress::Timestamp + differential_dataflow::lattice::Lattice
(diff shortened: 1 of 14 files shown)


src/compute/src/typedefs.rs
```diff
@@ -121,3 +121,31 @@ pub type RowErrBuilder<T, R> = RowValBuilder<DataflowError, T, R>;
 pub type KeyBatcher<K, T, D> = KeyValBatcher<K, (), T, D>;
 pub type KeyValBatcher<K, V, T, D> =
     MergeBatcher<Vec<((K, V), T, D)>, ColumnationChunker<((K, V), T, D)>, ColMerger<(K, V), T, D>>;
+
+/// Timestamp trait for rendering, constraint to support [`MzData`] and [timely::progress::Timestamp].
+pub trait MzTimestamp: MzData + timely::progress::Timestamp {}
```
teskje: How about adding `Lattice` too? That seems to be a common `T` bound as well.

teskje (approved):
I'm always in favor of reducing the generic types boilerplate. The only concern would be that me we're starting to have too many timestamp traits in compute (`Timestamp`, `RenderTimestamp`, `MzTimestamp`). Maybe we can merge the last two?

src/compute/src/typedefs.rs
antiguru: Added!

PR #32732, opened 2025-06-12
storage/pg: attempt exact snapshot count only when estimate is low enough

petrosagg:
We have recently seen cases where the exact count of rows is problematic when the target table contains many rows. The value of this exact count is not worth the resources spent since the result only informs the snapshot progress reporting.

This PR changes the logic to always collect the estimated number of rows first and if it's below a threshold (currently one million) attempt an exact count as well. Otherwise we just use the estimate.

I retained the flag to compeltely disable exact counting in case the current threshold is problematic.

This PR also removes the assertion that the total number of ingested rows match the number of counted rows. This assertion effectively checks that PostgreSQL transactions work, which doesn't seem very useful and due to other reasons cause CI flakes. See MaterializeInc/database-issues#9362

Fixes MaterializeInc/database-issues#9362

### Checklist

### DIFF
--- a/src/storage/src/source/postgres/snapshot.rs
+++ b/src/storage/src/source/postgres/snapshot.rs
@@ -149,6 +149,7 @@ use mz_postgres_util::{Client, PostgresError, simple_query_opt};
 use mz_repr::{Datum, DatumVec, Diff, Row};
 use mz_sql_parser::ast::{Ident, display::AstDisplay};
 use mz_storage_types::errors::DataflowError;
+use mz_storage_types::parameters::PgSourceSnapshotConfig;
 use mz_storage_types::sources::{MzOffset, PostgresSourceConnection};
 use mz_timely_util::builder_async::{
     Event as AsyncEvent, OperatorBuilder as AsyncOperatorBuilder, PressOnDropButton,
@@ -517,21 +518,12 @@ pub(crate) fn render<G: Scope<Timestamp = MzOffset>>(
             }
             *rewind_cap_set = CapabilitySet::new();
 
-            let strict_count = config.config.parameters.pg_snapshot_config.collect_strict_count;
-            if strict_count {
-                mz_ore::soft_assert_eq_or_log!(
-                    snapshot_staged,
-                    snapshot_total,
-                    "timely-{worker_id} size mimatch for snapshot of source {id}. \
-                        expected: {snapshot_total} actual: {snapshot_staged}"
-                );
-            }
             // Report the same known and staged records to signify that the snapshot is complete.
             stats_output.give(
                 &stats_cap[0],
                 ProgressStatisticsUpdate::Snapshot {
-                    records_known: snapshot_total,
-                    records_staged: snapshot_total,
+                    records_known: snapshot_staged,
+                    records_staged: snapshot_staged,
                 },
             );
 
@@ -717,14 +709,8 @@ async fn fetch_snapshot_size(
 
     let mut total = 0;
     for (table, oid, output_count) in tables {
-        let stats =
-            collect_table_statistics(client, snapshot_config.collect_strict_count, &table, oid)
-                .await?;
-        metrics.record_table_count_latency(
-            table,
-            stats.count_latency,
-            snapshot_config.collect_strict_count,
-        );
+        let stats = collect_table_statistics(client, snapshot_config, &table, oid).await?;
+        metrics.record_table_count_latency(table, stats.count_latency);
         total += stats.count * u64::cast_from(output_count);
     }
     Ok(total)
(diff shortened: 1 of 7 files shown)


src/storage/src/source/postgres/snapshot.rs
```diff
@@ -487,17 +487,17 @@ pub(crate) fn render<G: Scope<Timestamp = MzOffset>>(
...
+                    snapshot_staged += 1;
+                    // TODO(guswynn): does this 1000 need to be configurable?
+                    if snapshot_staged % 1000 == 0 {
+                        stats_output.give(
+                            &stats_cap[0],
+                            ProgressStatisticsUpdate::Snapshot {
+                                records_known: snapshot_total,
+                                records_staged: snapshot_staged,
+                            },
+                        );
```
petrosagg: @martykulma this change is debatable. We were previously counting each row as many times as there were tables for it. With the `CREATE TABLE FROM SOURCE` feature a certain upstream table can be ingested into multiple MZ tables. Should we count the rows received from the network when we report snapshot progress or the number of rows stored on our side?

martykulma: For now, I would say leave this out as there might be test fallout due to the different accounting. I have some issues assigned to me to update the metrics (https://github.com/MaterializeInc/database-issues/issues/9295, https://github.com/MaterializeInc/database-issues/issues/9361) and I can take care of this one as well.

src/storage/src/source/postgres/snapshot.rs
```diff
@@ -738,41 +723,35 @@ struct TableStatistics {
...
+    .wall_time()
+    .set_at(&mut stats.count_latency)
+    .await?;
+    stats.count = match estimate_row {
+        Some(row) => row.get("estimate_count").unwrap().parse().unwrap_or(0),
+        None => bail!("failed to get estimate count for {table}"),
+    };
+
+    // If the estimate is low enough we can attempt to get an exact count
+    if config.collect_strict_count && stats.count < 1_000_000 {
```
martykulma, line 750: This may end up running if the pg_class doesn't have a `reltuples` estimate.  How do you feel about adding `&& stats.out > 0`?

I feel like it's extremely unlikely, but if it did happen, the parties involved <strike>might</strike> will end up wasting time on it.

src/storage/src/source/postgres/snapshot.rs
petrosagg: Makes sense, will leave this out

src/storage/src/source/postgres/snapshot.rs
```diff
@@ -517,15 +517,6 @@ pub(crate) fn render<G: Scope<Timestamp = MzOffset>>(
...
-            if strict_count {
-                mz_ore::soft_assert_eq_or_log!(
-                    snapshot_staged,
-                    snapshot_total,
-                    "timely-{worker_id} size mimatch for snapshot of source {id}. \
-                        expected: {snapshot_total} actual: {snapshot_staged}"
-                );
-            }
             // Report the same known and staged records to signify that the snapshot is complete.
             stats_output.give(
```
martykulma, line 522: <strike>mmm... I may have found a reason to retract my comment about leaving out the accounting change 😅 .</strike>   In the case of using the pg_class estimate, would it make sense to have this set the values to staged (actual) vs. total (estimate)?

src/storage/src/source/postgres/snapshot.rs
petrosagg: I didn't add a special case for this on the basis that if there isn't an estimate then it must be a ~new table which would correlate with it not having many rows.

Is there some other case where there is no estimate but the table is very large?

src/storage/src/source/postgres/snapshot.rs
martykulma: GH -- not useful!

These are the lines I'm looking at
https://github.com/MaterializeInc/materialize/blob/25eb775818508c4ef3c58c5ed01dac817498b426/src/storage/src/source/postgres/snapshot.rs#L530-L536

src/storage/src/source/postgres/snapshot.rs
martykulma: Only bulk loading data or restoring via pg_restore.  Generally I've done that, or seen others do that, for test data, but it's been rare.  I don't feel strongly about it, but figured I would raise it.  Might be good to jot a comment with the assumption if nothing else.

petrosagg: Oh good point. I'll add a comment about the possibility and we still have the flag if someone wants to snapshot a huge but not yet vacuumed table

src/storage/src/source/postgres/snapshot.rs
petrosagg: Great point, fixed

PR #32738, opened 2025-06-13
cluster: optionally initialize Timely runtimes at cluster startup

teskje:
This PR makes it possible to initialize the Timely clusters at replica process startup time, as opposed to controller connection time. It introduces an (off-by-default) feature flag that, when enabled, makes replica processes on startup first initialize their Timely clusters and connect to their peer processes before starting to accept connections from the controller. The necessary Timely cluster config is now passed through the orchestrator as a command line argument to new clusterd processes.

Initializing Timely clusters at process startup simplifies controller-replica communication in two ways:

* On controller (re)connects, we don't have to run the (potentially blocking) Timely cluster setup. This simplifies the replica-side connection code, hopefully reducing the risk of deadlock bugs we had there in the past (https://github.com/MaterializeInc/database-issues/issues/8467, https://github.com/MaterializeInc/materialize/pull/29139, https://github.com/MaterializeInc/database-issues/issues/8957, https://github.com/MaterializeInc/database-issues/issues/8964).
* Once the feature is stabilized, the `CreateTimely` command can be removed, simplifying the cluster protocol and every client interacting with it.

### Motivation

### Tips for reviewer

I also have a PR that skips the feature flag and immediately makes the new scheme the only option: https://github.com/MaterializeInc/materialize/pull/31710. It's nice because we can remove a bunch of code and end up with a net-negative diff, but I prefer the feature flag approach for two reasons:

* It derisks the rollout of this change.
* We can split out the (substantial) test changes that this change necessitates into a separate PR, reducing the review burden.

### Checklist


src/clusterd/ci/entrypoint.sh
```diff
@@ -21,11 +21,15 @@ export CLUSTERD_INTERNAL_HTTP_LISTEN_ADDR=${CLUSTERD_INTERNAL_HTTP_LISTEN_ADDR:-
...
-# in Docker), this is likely not desirable.
 if [[ "${KUBERNETES_SERVICE_HOST:-}" ]]; then
+    # Pass the host's FQDN as the host to be used for GRPC request validation
+    # only when running in Kubernetes. In other contexts (like when running
+    # locally, or in Docker), this is likely not desirable.
     export CLUSTERD_GRPC_HOST=${CLUSTERD_GRPC_HOST:-$(hostname --fqdn)}
+
+    # When running in Kubernetes, pass the StatefulSet replica's ordinal index
+    # as the process index.
+    export CLUSTERD_PROCESS=${CLUSTERD_PROCESS:-${HOSTNAME##*-}}
```
teskje, line 32: The plan is to have a StatefulSet per process and supply the process index in the `TimelyConfig`s passed to each. Changing to multiple StatefulSets is a bigger change though ([draft](https://github.com/MaterializeInc/materialize/pull/31629)), so deferring it to a follow-up and doing the simple thing here.

src/controller/src/lib.rs
```diff
@@ -23,10 +23,10 @@
 
 use std::collections::btree_map::Entry;
 use std::collections::{BTreeMap, BTreeSet};
-use std::mem;
 use std::num::NonZeroI64;
 use std::sync::Arc;
 use std::time::Duration;
+use std::{mem, str};
```
antiguru: Stray?

antiguru (approved):
Looks good, thanks!

teskje:
TFTR!

PR #32752, opened 2025-06-17
feature flags: Update defaults based on Sentry production and changelog

def-:
In anticipation of self-managed v25.2
### Checklist


teskje (approved):

antiguru (approved):

src/adapter-types/src/timestamp_selection.rs
```diff
@@ -20,13 +20,13 @@ pub enum ConstraintBasedTimestampSelection {
...
     fn default() -> Self {
-        Self::Verify
+        Self::Enabled
     }
 }
 
 impl ConstraintBasedTimestampSelection {
     pub const fn const_default() -> Self {
-        Self::Verify
+        Self::Enabled
```
def-, line 29: @frankmcsherry Is this fine?

src/balancerd/src/dyncfgs.rs
```diff
@@ -39,7 +39,7 @@ pub const SIGTERM_LISTEN_WAIT: Config<Duration> = Config::new(
 /// Whether to inject tcp proxy protocol headers to downstream http servers.
 pub const INJECT_PROXY_PROTOCOL_HEADER_HTTP: Config<bool> = Config::new(
     "balancerd_inject_proxy_protocol_header_http",
-    false,
+    true,
```
def-: @jubrad Is this fine?

src/controller-types/src/dyncfgs.rs
```diff
@@ -22,7 +22,7 @@ pub const CONTROLLER_PAST_GENERATION_REPLICA_CLEANUP_RETRY_INTERVAL: Config<Dura
 
 pub const ENABLE_0DT_DEPLOYMENT_SOURCES: Config<bool> = Config::new(
     "enable_0dt_deployment_sources",
-    false,
+    true,
```
def-, line 25: @aljoscha Is this fine?

src/persist-client/src/batch.rs
```diff
@@ -371,7 +371,7 @@ pub(crate) const BATCH_DELETE_ENABLED: Config<bool> = Config::new(
 
 pub(crate) const ENCODING_ENABLE_DICTIONARY: Config<bool> = Config::new(
     "persist_encoding_enable_dictionary",
-    false,
+    true,
```
def-, line 374: @bkirwi Is this fine?

src/sql/src/session/vars/definitions.rs
```diff
@@ -2080,7 +2080,7 @@ feature_flags!(
     {
         name: enable_variadic_left_join_lowering,
         desc: "Enable joint HIR ⇒ MIR lowering of stacks of left joins",
-        default: false,
+        default: true,
```
def-, line 2083: @ggevay Is this fine?

src/sql/src/session/vars/definitions.rs
```diff
@@ -2182,13 +2182,13 @@ feature_flags!(
     {
         name: enable_join_prioritize_arranged,
         desc: "Whether join planning should prioritize already-arranged keys over keys with more fields.",
-        default: true,
+        default: false,
```
def-, line 2185: @ggevay Is this fine?

src/sql/src/session/vars/definitions.rs
```diff
@@ -2182,13 +2182,13 @@ feature_flags!(
...
         enable_for_item_parsing: false,
     },
     {
         name: enable_sql_server_source,
         desc: "Creating a SQL SERVER source",
-        default: false,
+        default: true,
```
def-, line 2191: @ptravers Is this fine?

src/sql/src/session/vars/definitions.rs
ggevay: Yes

src/sql/src/session/vars/definitions.rs
ggevay: Kind of. It's the same on cloud prod, but it _can_ lead to plan regressions when turning on. So when people upgrade self-managed from the old version, we might have some unpleasant surprises. But I guess it would be good to harmonize it between cloud and self-managed, so maybe it's worth it to risk it while we don't have many self-managed users? :badpokerface:

ggevay: What do you think @frankmcsherry ?

src/metrics/src/dyncfgs.rs
```diff
@@ -16,7 +16,7 @@ use mz_dyncfg::{Config, ConfigSet};
 /// How frequently to refresh lgalloc map stats.
 pub(crate) const MZ_METRICS_LGALLOC_MAP_REFRESH_INTERVAL: Config<Duration> = Config::new(
     "mz_metrics_lgalloc_map_refresh_interval",
-    Duration::from_secs(60),
+    Duration::from_secs(600),
```
teskje, line 19: Is this right? I still see this as 60s in LD in production.

def-: Screenshot 2025-06-18 at 10 30 00
Weird, this is what I see

teskje: Huh, I see that now too. No idea what I was looking at before!

src/sql/src/session/vars/definitions.rs
ptravers: yeah, that matches cloud prod at this point.

src/adapter-types/src/timestamp_selection.rs
frankmcsherry: I'm ~positive~ supportive, but @DAlperin might be the best person to confirm that this is low risk.

src/sql/src/session/vars/definitions.rs
frankmcsherry: I think we should assertively set the defaults to be those we want people to use, and I think we want people to use this. Aiui, self-managed folks have the ability to flip the flag back, right, in case they don't like it?

def-: Correct.

src/adapter-types/src/timestamp_selection.rs
def-: Dov confirms via Slack

src/controller-types/src/dyncfgs.rs
aljoscha: it is!

src/balancerd/src/dyncfgs.rs
jubrad: I think we really only want this on in cloud deployments, the same is probably true for network policies in general.

In self-managed we put nginx infront of our api which breaks our ability to determine client IPs in MZ, the proxy_protocol_header can only be used if there's no proxy upstream of the balancer.

src/persist-client/src/batch.rs
bkirwi: sure!

PR #32753, opened 2025-06-17
as-of-selection: work around concurrent dropping of collections

teskje:
During a 0dt upgrade, the leader env may drop storage collections while the read-only env is bootstrapping. As a result, the read-only env might see inconsistent read frontiers for storage collections, where downstream collections have a non-empty read frontier while their upstreams have empty read frontiers, making as-of selection fail.

There is no work planned to avoid this situation from occurring, so this commit takes the approach of working around the issue during as-of selection. It does so by assigning empty as-ofs to storage collections with empty frontiers, and all collections depending on them. As a result, no constraint errors are raised in as-of selection and the affected compute collections are not rendered.

To limit the blast radius of this change, which has the potential to reduce our ability to detect bugs in the system, we only apply the special pruning step when the environment starts in read-only mode.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/8836

### Checklist


src/compute-client/src/as_of_selection.rs
```diff
@@ -737,6 +755,46 @@ impl<'a, T: TimestampManipulation> Context<'a, T> {
...
+        // Remove dropped dropped storage collections.
+        let mut pruned = BTreeSet::new();
+        self.collections.retain(|id, _c| {
+            let empty = self
+                .storage_collections
+                .collection_frontiers(*id)
+                .map_or(true, |f| f.read_capabilities.is_empty());
+
+            if empty {
+                warn!(%id, "pruning collection with empty frontier");
```
antiguru: Maybe move the `warn` out of the closure and do something like this:
```
if !pruned.is_empty() {
    warn!(%pruned, "Pruned collections with empty frontiers");
}
```

antiguru (commented):
Looks fine, but waiting for out of draft.

antiguru (approved):

teskje:
TFTR!

PR #32756, opened 2025-06-17
catalog: `mz_cluster_replica_{metrics,statuses}` views

teskje:
This PR converts `mz_cluster_replica_{metrics,statuses}` from builtin tables into a views over `mz_cluster_replica_{metrics, status}_history`. Semantics stay exactly the same, except that the indexes on these views are not "retained metrics objects" anymore, i.e. they don't keep around a 30 day history anymore. For historical lookups, `mz_cluster_replica_{metric,status}_history` should be used.

### Motivation

Closes https://github.com/MaterializeInc/database-issues/issues/8403

### Checklist

### DIFF
--- a/src/adapter/src/catalog/open.rs
+++ b/src/adapter/src/catalog/open.rs
@@ -872,11 +872,28 @@ fn add_new_remove_old_builtin_items_migration(
     // the catalog.
     let mut deleted_system_objects = BTreeSet::new();
     let mut deleted_runtime_alterable_system_ids = BTreeSet::new();
-    for (_, mapping) in system_object_mappings {
+    let mut deleted_comments = BTreeSet::new();
+    for (desc, mapping) in system_object_mappings {
         deleted_system_objects.insert(mapping.description);
         if mapping.unique_identifier.fingerprint == RUNTIME_ALTERABLE_FINGERPRINT_SENTINEL {
             deleted_runtime_alterable_system_ids.insert(mapping.unique_identifier.catalog_id);
         }
+
+        let id = mapping.unique_identifier.catalog_id;
+        let comment_id = match desc.object_type {
+            CatalogItemType::Table => CommentObjectId::Table(id),
+            CatalogItemType::Source => CommentObjectId::Source(id),
+            CatalogItemType::View => CommentObjectId::View(id),
+            CatalogItemType::Sink
+            | CatalogItemType::MaterializedView
+            | CatalogItemType::Index
+            | CatalogItemType::Type
+            | CatalogItemType::Func
+            | CatalogItemType::Secret
+            | CatalogItemType::Connection
+            | CatalogItemType::ContinualTask => continue,
+        };
+        deleted_comments.insert(comment_id);
     }
     // If you are 100% positive that it is safe to delete a system object outside any of the
     // unstable schemas, then add it to this set. Make sure that no prod environments are
@@ -899,6 +916,7 @@ fn add_new_remove_old_builtin_items_migration(
         "only objects in unstable schemas can be deleted, deleted objects: {:?}",
         deleted_system_objects
     );
+    txn.drop_comments(&deleted_comments)?;
     txn.remove_items(&deleted_runtime_alterable_system_ids)?;
     txn.remove_system_object_mappings(deleted_system_objects)?;
 


def-:
This might increase the CPU utilization so much that Yugabyte fails to start up here: https://buildkite.com/materialize/test/builds/104957#019789c0-9ee9-4e29-8571-857b7ca52eab (only 4 cpus, so maybe just bump it)

Similarly cluster tests seem to have a bunch of query timeouts

teskje:
I missed making the indexes on these views non-"retained metrics" objects. Looks like after fixing this the Yugabyte tests succeed now.

teskje:
@SangJunBak Once this lands, `mz_cluster_replica_statuses` will not retain a history anymore like it currently does. The console still queries this relation, but only to get the latest replica status so it doesn't rely on the history, correct?

doc/user/content/sql/system-catalog/mz_internal.md
```diff
@@ -155,7 +155,7 @@ The `mz_cluster_schedules` table shows the `SCHEDULE` option specified for each
 
 ## `mz_cluster_replica_metrics`
 
-The `mz_cluster_replica_metrics` table gives the last known CPU and RAM utilization statistics
+The `mz_cluster_replica_metrics` view gives the last known CPU and RAM utilization statistics
```
aljoscha, line 158: Or, something like "The builtin `mz_cluster_replica_metrics` gives ...", to avoid specifying whether it's a table, view, or what have you. In case we want to change it again in the future.

But what you have is totally fine!

src/adapter/src/catalog/builtin_table_updates.rs
```diff
@@ -437,35 +435,6 @@ impl CatalogState {
         updates
     }
 
-    // TODO(jkosh44) This should not be a builtin table, it should be a builtin source.
```
aljoscha, line 440: 🙌

src/orchestrator-process/src/lib.rs
```diff
@@ -683,7 +683,7 @@ impl OrchestratorWorker {
                     listen_addrs,
                     peer_addrs: &peer_addrs,
                 });
-                args.push(format!("--process={i}"));
+                // args.push(format!("--process={i}"));
```
aljoscha: ? 🤔

test/sqllogictest/distinct_arrangements.slt
```diff
@@ -1002,7 +1002,7 @@ ArrangeBy[[Column(0),␠Column(1)]]  2
 ArrangeBy[[Column(0),␠Column(2)]]  4
 ArrangeBy[[Column(0),␠Column(3)]]  4
 ArrangeBy[[Column(0),␠Column(4)]]  1
-ArrangeBy[[Column(0)]]  155
+ArrangeBy[[Column(0)]]  157
```
aljoscha, line 1005: always love when my change make it so I have to change these, and I look up to auto-generation scripts again 😅

doc/user/content/sql/system-catalog/mz_internal.md
teskje: I'll leave it like this as it matches the text we have for other relations. Not against changing it, but if we do so we should do it consistently and I consider that out of scope.

src/orchestrator-process/src/lib.rs
teskje: Damn, glad you saw this! I commented this out during testing, though I don't remember why now

test/sqllogictest/distinct_arrangements.slt
teskje: At least you can automatically rewrite the SLTs if you know which ones are likely to need rewrites 🫠

teskje:
TFTRs!

SangJunBak:
@teskje Ah there is a feature that uses subscribe to check if a cluster OOM'd in the last 15 minutes and shows it as a tooltip in the cluster list page [[Code pointer](https://github.com/MaterializeInc/console/blob/main/src/api/materialize/cluster/useLatestOfflineReplica.ts#L15-L16)].

However I haven't seen this feature work in a while so it's debatable if it still works. Regardless, this change won't show an error state (since we use AS OF AT LEAST) and thus shouldn't block this PR. Created a followup Console ticket https://github.com/MaterializeInc/database-issues/issues/10143

PR #32763, opened 2025-06-18
k8s: retry API requests in the cloud resource controller

teskje:
This PR makes the cloud resource controller retry k8s API requests, to handle transient errors gracefully. In contrast to the `OrchestratorWorker` it doesn't retry forever but only for a limited amount of times (10 minutes currently).

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9253

### Tips for reviewer

I chose the 10 minute timeout out of fear that we'll start hanging somewhere indefinitely and that this might be worse than crashing envd (or whatever else happens when the cloud resource controller returns an error). The fear might be unfounded and maybe we should retry forever, lmk your thoughts!

### Checklist

### DIFF
--- a/src/orchestrator-kubernetes/src/cloud_resource_controller.rs
+++ b/src/orchestrator-kubernetes/src/cloud_resource_controller.rs
@@ -11,16 +11,19 @@
 
 use std::collections::BTreeMap;
 use std::sync::Arc;
+use std::time::Duration;
 
 use async_trait::async_trait;
 use chrono::Utc;
-use futures::StreamExt;
 use futures::stream::BoxStream;
+use futures::{StreamExt, TryFutureExt};
 use kube::api::{DeleteParams, ListParams, ObjectMeta, Patch, PatchParams};
 use kube::runtime::{WatchStreamExt, watcher};
 use kube::{Api, ResourceExt};
 use maplit::btreemap;
+use mz_ore::retry::Retry;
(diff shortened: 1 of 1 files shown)


src/orchestrator-kubernetes/src/cloud_resource_controller.rs
```diff
@@ -189,7 +203,21 @@ impl KubernetesResourceReader {
...
+/// Helper for calling a kube API with (limited) retry.
+async fn call_api<'a, K, F, U, T, E>(api: &'a Api<K>, f: F) -> Result<T, E>
+where
+    F: Fn(&'a Api<K>) -> U + 'a,
+    U: Future<Output = Result<T, E>>,
+    E: std::fmt::Display,
+{
+    Retry::default()
+        .clamp_backoff(Duration::from_secs(10))
+        .max_duration(Duration::from_secs(60 * 10))
```
alex-hunt-materialize, line 220: I am also not sure about whether it is better to retry forever. I assume that depends on the surrounding code. If we can't make progress on something else because of retries, giving up is better, but if we can make progress, retrying forever is better.

teskje: Agreed. Giving up eventually is the conservative choice, so let's try that initially and see how it fares.

teskje:
TFTR!

PR #32767, opened 2025-06-19
adapter: Add a tokio canary metric

ggevay:
This PR adds a simple tokio task that just sleeps for 1 sec in a loop, and records how much later than 1 sec it got woken up. This way, we'll be able to see how overloaded tokio is.

### Motivation

  * This PR adds a known-desirable feature: This is for https://github.com/MaterializeInc/database-issues/issues/9140, discussed with @aljoscha

### Checklist

### DIFF
--- a/src/adapter/src/metrics.rs
+++ b/src/adapter/src/metrics.rs
@@ -47,6 +47,7 @@ pub struct Metrics {
     pub pgwire_message_processing_seconds: HistogramVec,
     pub result_rows_first_to_last_byte_seconds: HistogramVec,
     pub pgwire_ensure_transaction_seconds: HistogramVec,
+    pub tokio_canary_delay_ms: HistogramVec,
 }
 
 impl Metrics {
@@ -206,7 +207,12 @@ impl Metrics {
                 help: "The time it takes to run `ensure_transactions` when processing pgwire messages.",
                 var_labels: ["message_type"],
                 buckets: histogram_seconds_buckets(0.001, 512.0),
-            ))
+            )),
+            tokio_canary_delay_ms: registry.register(metric!(
+                name: "mz_tokio_canary_delay_ms",
+                help: "How much later was the tokio canary task woken up than the requested time.",
+                buckets: histogram_milliseconds_buckets(1., 512000.),
+            )),
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/adapter/src/metrics.rs
```diff
@@ -206,7 +207,12 @@ impl Metrics {
                 help: "The time it takes to run `ensure_transactions` when processing pgwire messages.",
                 var_labels: ["message_type"],
                 buckets: histogram_seconds_buckets(0.001, 512.0),
-            ))
+            )),
+            tokio_canary_delay_seconds: registry.register(metric!(
```
aljoscha: should we be optimistic and call this `mz_tokio_canary_delay_ms`, and use ms as the base unit? 😅

aljoscha (approved):
Nice and simple. I had one recommendation to use ms as the base unit

src/adapter/src/metrics.rs
ggevay: Yep, I'll do that! :sweat_smile:

ggevay:
(Merging is currently blocked on some CI infrastructure issue, being discussed [here](https://materializeinc.slack.com/archives/C01LKF361MZ/p1750337918741699).)

def-:
I think the SLT failures in it are real, everything else failing in the test is just CI stuff

I checked out the PR and locally ran `bin/sqllogictest -- --verbose --verbose --verbose test/sqllogictest/*.slt` and I see no output, it just seems to hang on the first test. Without the PR I see queries running immediately

bkirwi:
Unsure if you're already aware, but we do capture a bunch of tokio metrics via: https://docs.rs/tokio/latest/tokio/runtime/struct.RuntimeMetrics.html

Not all of those are instrumented yet! But if any of those would help answer your question they'd be fun to add also.

ggevay:
Ok, as Ben and Moritz noted, we do already have some metrics on tokio. Specifically, `mz_tokio_global_queue_depth` might also give us the confirmation that we need that tokio is not too overloaded.

Edit: But that is sampled on once per minute, so for spikes of a few seconds, it might miss it. The metric in the PR is once per second. I'm not sure what would it take to also modify `mz_tokio_global_queue_depth` to be once per second.

ggevay:
Superseded by https://github.com/MaterializeInc/materialize/pull/32981

PR #32768, opened 2025-06-19
Initialize Timely runtimes at clusterd process startup by default

teskje:
This PR enables initialization of the Timely runtimes at process startup, by flipping the default value of the corresponding feature flag. It then adjusts all test code using unorchestrated replicas accordingly, by applying the Timely cluster configuration as env variables to the clusterd processes, instead of specifying them in the `CREATE CLUSTER REPLICA` command.

### Motivation

### Tips for reviewers

I strongly recommend reviewing the diff with whitespace ignored. It gets much smaller this way. For many tests I had to add additional service overwrites, which add one level of indentation.

Also note that I'm not yet removing the `STORAGE ADDRESSES`/`COMPUTE ADDRESSES`/`WORKERS` options from `CREATE CLUSTER REPLICA` commands, as they are currently still expected by Materialize and are expected to match the process command options. We'll be able to remove them once we remove support for the old way of initializing Timely runtimes.

### Checklist

### DIFF
--- a/misc/python/materialize/mzcompose/services/clusterd.py
+++ b/misc/python/materialize/mzcompose/services/clusterd.py
@@ -7,6 +7,7 @@
 # the Business Source License, use of this software will be governed
 # by the Apache License, Version 2.0.
 
+import json
 
 from materialize.mzcompose import DEFAULT_MZ_ENVIRONMENT_ID, DEFAULT_MZ_VOLUMES
 from materialize.mzcompose.service import (
@@ -29,6 +30,8 @@ class Clusterd(Service):
         stop_grace_period: str = "120s",
         scratch_directory: str = "/scratch",
         volumes: list[str] = [],
+        workers: int = 1,
+        process_names: list[str] = [],
     ) -> None:
         environment = [
             "CLUSTERD_LOG_FILTER",
@@ -41,6 +44,17 @@ class Clusterd(Service):
             environment_id = DEFAULT_MZ_ENVIRONMENT_ID
 
         environment += [f"CLUSTERD_ENVIRONMENT_ID={environment_id}"]
+
+        process_names = process_names if process_names else [name]
+        process_index = process_names.index(name)
+        compute_timely_config = timely_config(process_names, 2102, workers, 16)
+        storage_timely_config = timely_config(process_names, 2103, workers, 1337)
+
+        environment += [
+            f"CLUSTERD_PROCESS={process_index}",
+            f"CLUSTERD_COMPUTE_TIMELY_CONFIG={compute_timely_config}",
+            f"CLUSTERD_STORAGE_TIMELY_CONFIG={storage_timely_config}",
+        ]
 
         options = [f"--scratch-directory={scratch_directory}", *options]
 
@@ -74,3 +88,21 @@ class Clusterd(Service):
         )
 
         super().__init__(name=name, config=config)
+
+
+def timely_config(
+    process_names: list[str],
+    port: int,
+    workers: int,
+    arrangement_exert_proportionality: int,
+) -> str:
+    config = {
+        "workers": workers,
+        "process": 0,
+        "addresses": [f"{n}:{port}" for n in process_names],
+        "arrangement_exert_proportionality": arrangement_exert_proportionality,
+        "enable_zero_copy": False,
+        "enable_zero_copy_lgalloc": False,
+        "zero_copy_limit": None,
+    }
+    return json.dumps(config)
--- a/test/cluster-isolation/mzcompose.py
+++ b/test/cluster-isolation/mzcompose.py
@@ -236,46 +236,68 @@ $ kafka-verify-data format=avro sink=materialize.public.sink1 sort-messages=true
 def run_test(c: Composition, disruption: Disruption, id: int) -> None:
     print(f"+++ Running disruption scenario {disruption.name}")
 
-    c.up("testdrive", persistent=True)
-    c.up("materialized", "clusterd_1_1", "clusterd_1_2", "clusterd_2_1", "clusterd_2_2")
-
-    c.sql(
-        "ALTER SYSTEM SET unsafe_enable_unorchestrated_cluster_replicas = true;",
-        port=6877,
-        user="mz_system",
-    )
-
-    c.sql(
-        """
-        DROP CLUSTER IF EXISTS cluster1 CASCADE;
-        CREATE CLUSTER cluster1 REPLICAS (replica1 (
-            STORAGECTL ADDRESSES ['clusterd_1_1:2100', 'clusterd_1_2:2100'],
-            STORAGE ADDRESSES ['clusterd_1_1:2103', 'clusterd_1_2:2103'],
-            COMPUTECTL ADDRESSES ['clusterd_1_1:2101', 'clusterd_1_2:2101'],
-            COMPUTE ADDRESSES ['clusterd_1_1:2102', 'clusterd_1_2:2102']
-        ));
-        """
-    )
-
-    c.sql(
-        """
-        DROP CLUSTER IF EXISTS cluster2 CASCADE;
-        CREATE CLUSTER cluster2 REPLICAS (replica1 (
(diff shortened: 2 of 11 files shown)


test/cluster/mzcompose.py
```diff
@@ -57,10 +57,10 @@
...
     SchemaRegistry(),
     Localstack(),
-    Clusterd(name="clusterd1"),
-    Clusterd(name="clusterd2"),
-    Clusterd(name="clusterd3"),
-    Clusterd(name="clusterd4"),
+    Clusterd(name="clusterd1", workers=2),
+    Clusterd(name="clusterd2", workers=2),
+    Clusterd(name="clusterd3", workers=2),
+    Clusterd(name="clusterd4", workers=2),
```
teskje, line 63: I used 2 as the default worker count because that seemed to be what's used by the most tests. Not opposed to using 1 instead and changing the tests with overwrites.

test/cluster/mzcompose.py
```diff
@@ -525,8 +555,8 @@ def workflow_test_github_4545(c: Composition) -> None:
                 WORKERS 2
             ),
             logging_off (
-                STORAGECTL ADDRESSES ['clusterd1:2100'],
-                STORAGE ADDRESSES ['clusterd1:2103'],
+                STORAGECTL ADDRESSES ['clusterd2:2100'],
+                STORAGE ADDRESSES ['clusterd2:2103'],
```
teskje, line 559: I think this only worked before because we are not trying to use the storage cluster in this test.

test/cluster/mzcompose.py
```diff
@@ -2071,9 +2125,9 @@ def workflow_test_drop_during_reconciliation(c: Composition) -> None:
         c.up("materialized", "clusterd1", "toxiproxy")
         c.up("testdrive", persistent=True)
 
-        # Set up toxi-proxies for all clusterd endpoints.
+        # Set up toxi-proxies for clusterd GRPC endpoints.
```
teskje, line 2122: Excluding the Timely runtime endpoints from getting toxiproxied. This change is not strictly necessary here but already prepares for the removal of the `COMPUTE ADDRESSES`/`STORAGE ADDRESSES` options.

test/limits/mzcompose.py
```diff
@@ -1846,12 +1846,12 @@ def app_password(email: str) -> str:
...
     "clusterd_1_1_1",
+    "clusterd_1_1_2",
     "clusterd_1_2_1",
+    "clusterd_1_2_2",
     "clusterd_2_1_1",
-    "clusterd_2_2_1",
+    "clusterd_2_1_2",
     "clusterd_3_1_1",
-    "clusterd_3_2_1",
-    "clusterd_4_1_1",
```
teskje, line 1854: I took the liberty of adjusting the clusterd names to be `clusterd_{cluster}_{replica}_{process}`. Previously they were `clusterd_{replica}_{process}_1`.

test/limits/mzcompose.py
```diff
@@ -2006,15 +2006,57 @@ def workflow_main(c: Composition, parser: WorkflowArgumentParser) -> None:
...
+            name="clusterd_2_1_2",
+            workers=args.workers,
+            process_names=["clusterd_2_1_1", "clusterd_2_1_2"],
+        ),
+        Clusterd(
+            name="clusterd_3_1_1",
+            workers=1,
+        ),
+    ):
+        run_scenarios(c, scenarios, args.find_limit, args.workers)
```
teskje, line 2045: I mainly added the `run_scenarios` method so I don't have to indent all subsequent code by one. I think it helps readability but you might disagree!

test/limits/mzcompose.py
```diff
@@ -2081,7 +2123,7 @@ def workflow_main(c: Composition, parser: WorkflowArgumentParser) -> None:
                             password=app_password(ADMIN_USER),
                         ) as cur:
                             start_time = time.time()
-                            cur.execute(scenario.EXPLAIN)
+                            cur.execute(scenario.EXPLAIN)  # type: ignore
```
teskje: Type checking complains here because `Cursor.execute` only takes `LiteralString`, or some psycopg type that prevents SQL injection. But we do dynamically build the value of `EXPLAIN`, so I'm not sure there is anything better to do here than ignore the type check error.

teskje:
I'm taking a look at the `SmallClusters` feature benchmark. This change moves Timely initialization from after controller connection time to before controller connection time, and it's plausible that this messes with the benchmark timings because the benchmark (accidentally?) also measures how long it takes for the cluster to come up.

misc/python/materialize/mzcompose/services/clusterd.py
```diff
@@ -74,3 +88,21 @@ def __init__(
...
+    arrangement_exert_proportionality: int,
+) -> str:
+    config = {
+        "workers": workers,
+        "process": 0,
+        "addresses": [f"{n}:{port}" for n in process_names],
+        "arrangement_exert_proportionality": arrangement_exert_proportionality,
+        "enable_zero_copy": False,
+        "enable_zero_copy_lgalloc": False,
+        "zero_copy_limit": None,
```
def-, line 106: Would those be interesting to cover in tests?

test/limits/mzcompose.py
def-:
```suggestion
                            cur.execute(scenario.EXPLAIN.encode())
```
This error indicates that you should either pass a literal string or UTF8-encoded bytes.

misc/python/materialize/mzcompose/services/clusterd.py
teskje: Maybe! They are switched off everywhere in CI right now. If we want to change that I'd do it as a separate PR (cc @antiguru).

antiguru: I think we cannot toggle these flags until we have access to swap, so it's fine to keep this disabled for the moment.

test/cluster-isolation/mzcompose.py
```diff
@@ -236,46 +236,68 @@ def validate(c: Composition) -> None:
...
-    )
-
     with c.override(
         Testdrive(
             no_reset=True,
             materialize_params={"cluster": "cluster2"},
             seed=id,
-        )
+        ),
+        Clusterd(
```
antiguru, line 245: Could we add a function to `Clusterd` that returns the SQL snippet to create the cluster? That might avoid a bunch of the boilerplate in our tests. Alternatively, a function that returns the clusterd service and a sql snippet...

(Doesn't need to be part of this PR.)

antiguru (approved):

test/cluster-isolation/mzcompose.py
teskje: Ah, I'm planning to remove all the Timely-runtime SQL options anyway after this has run in prod for a release or so. I don't think it's worth spending time on removing boilerplate which we plan to rip out soon anyway?

teskje:
In the `SmallClusters` feature benchmark we see a regression of ~1s. That is consistent with the maximum backoff the controller uses when trying to connect to a replica. I think what happens here is that we are now running into this maximum backoff more often because the Timely initialization now happens before the replica starts accepting connections, instead of after.

Imo a regression of 1s in replica connection time is not an issue. In production we can expect much higher delays since pods might not be immediately scheduleable and we might need to spawn a new node first.

To make CI green, we can just ignore the benchmark regression. But I also wonder if the benchmark is even meant to test cluster startup time? If it's not, we have an opportunity to make it more reliable by excluding the cluster startup time from the measured time.

test/limits/mzcompose.py
teskje: Ah, tricky! Presumably that's only allowed because `LiteralBytes` does not exist.

def- (approved):

teskje:
TFTRs!

bosconi:
bugbot run

test/cluster/mzcompose.py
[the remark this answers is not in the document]
teskje: This test was removed in 85faa6eeead22b1f4da5e5b9f8adf8032ba3b5a0.

PR #32769, opened 2025-06-19
More flexible polar signals debug symbol uploading

antiguru:
This enhances our tool to upload debug symbols to polar signals by allowing to download symbols using http, and by overriding the arch.


def- (approved):
Looks reasonable

PR #32774, opened 2025-06-20
[persist] Fix a frontier issue in maelstrom test

bkirwi:
Chasing down a very rare flake in a maelstrom test. The cause appears to be the test harness attempting to start a listen operation at a time earlier than the since hold it's obtained on the shard... so if the since of the shard is downgraded aggressively, the listen may work at first but then fail.

IMO it would be good to change the API so that we disallow reads before the since we've obtained, but that has the risk of causing errors for calls that currently work, so for now let's add a soft assert and tweak the test to avoid this case.

### Motivation

https://github.com/MaterializeInc/database-issues/issues/9400

### Checklist

### DIFF
--- a/src/persist-client/src/read.rs
+++ b/src/persist-client/src/read.rs
@@ -24,10 +24,10 @@ use differential_dataflow::trace::Description;
 use futures::Stream;
 use futures_util::{StreamExt, stream};
 use mz_dyncfg::Config;
-use mz_ore::halt;
 use mz_ore::instrument;
 use mz_ore::now::EpochMillis;
 use mz_ore::task::{AbortOnDropHandle, JoinHandle, RuntimeExt};
+use mz_ore::{halt, soft_assert_or_log};
 use mz_persist::location::{Blob, SeqNo};
 use mz_persist_types::columnar::{ColumnDecoder, Schema};
 use mz_persist_types::{Codec, Codec64};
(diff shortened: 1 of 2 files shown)


bkirwi:
There's some red in the nightlies, but I'm pretty confident it's unrelated: the maelstrom test itself is green, and outside the test all we've done is add a soft assert, and we don't actually seem to hit that assert.

src/persist-client/src/read.rs
```diff
@@ -677,6 +677,12 @@ where
     /// `as_of` that would have been accepted.
     #[instrument(level = "debug", fields(shard = %self.machine.shard_id()))]
     pub async fn listen(self, as_of: Antichain<T>) -> Result<Listen<K, V, T, D>, Since<T>> {
+        soft_assert_or_log!(
```
DAlperin, line 680: I am trying to get an intuition for how often we think this happens if at all?

DAlperin (approved):

src/persist-client/src/read.rs
bkirwi: I think it _probably_ never happens, but we don't prevent it... so this will hopefully help confirm that hypothesis and help prevent anyone adding new code that does it.

In general we can't guarantee that a read before the handle's since will succeed, so it'd be a little weird for someone to rely on it. Conceivable that it happens to work, though.

PR #32780, opened 2025-06-21
test: reset replicas after slts

ggevay:
The following fails on the current `main`:
```
bin/sqllogictest -- -v test/sqllogictest/github-3963.slt test/sqllogictest/github-4336.slt test/sqllogictest/github-5126.slt
```

This seems to be because https://github.com/MaterializeInc/materialize/pull/32203 removed some code that used to reset the replicas of the quickstart cluster if an slt (such as `github-3963.slt`) drops it. This PR puts back the code that was removed from `reset_database`. I'm not sure why the code was removed, so it's possible that this PR does something incorrect.

I'll open a slack thread about some additional problem around whether `RunConfig.replicas` means replication factor or replica size, which we should fix in a later PR. Edit: https://materializeinc.slack.com/archives/C01LKF361MZ/p1750514085398339

### Motivation

### Checklist

### DIFF
--- a/src/sqllogictest/src/runner.rs
+++ b/src/sqllogictest/src/runner.rs
@@ -862,6 +862,36 @@ impl<'a> Runner<'a> {
                 .batch_execute("CREATE CLUSTER quickstart REPLICAS ()")
                 .await?;
         }
+        let mut needs_default_replica = true;
+        for row in inner
+            .system_client
+            .query(
+                "SELECT name, size FROM mz_cluster_replicas
+                 WHERE cluster_id = (SELECT id FROM mz_clusters WHERE name = 'quickstart')",
+                &[],
+            )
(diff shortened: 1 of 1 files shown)


ggevay:
Yes, but that seems to be hardcoded to 1 anyway, see [slack thread on the configuration problem](https://materializeinc.slack.com/archives/C01LKF361MZ/p1750514085398339). E.g. https://github.com/MaterializeInc/materialize/blob/e19a9a8db6bd648f61c62ed66f61712dd95cda63/src/sqllogictest/src/runner.rs#L1144

When we change the replication factor configuration to not be hardcoded, then we'll have to modify the code re-added in this PR to not always set exactly one replica, but set as many as configured.

def- (approved):

PR #32784, opened 2025-06-21
Refactor compaction to use a stream

DAlperin:
This is a precursor to incremental compaction https://github.com/MaterializeInc/database-issues/issues/9194

### Checklist

### DIFF
--- a/src/persist-client/src/internal/compact.rs
+++ b/src/persist-client/src/internal/compact.rs
@@ -19,7 +19,8 @@ use anyhow::anyhow;
 use differential_dataflow::difference::Semigroup;
 use differential_dataflow::lattice::Lattice;
 use differential_dataflow::trace::Description;
-use futures_util::{StreamExt, TryFutureExt};
+use futures::{Stream, pin_mut};
+use futures_util::StreamExt;
 use mz_dyncfg::Config;
 use mz_ore::cast::CastFrom;
 use mz_ore::error::ErrorExt;
@@ -112,7 +113,7 @@ pub struct Compactor<K, V, T, D> {
         Instant,
         CompactReq<T>,
         Machine<K, V, T, D>,
-        oneshot::Sender<Result<ApplyMergeResult, anyhow::Error>>,
+        oneshot::Sender<Result<(), anyhow::Error>>,
     )>,
     _phantom: PhantomData<fn() -> D>,
 }
@@ -172,7 +173,7 @@ where
             Instant,
             CompactReq<T>,
             Machine<K, V, T, D>,
-            oneshot::Sender<Result<ApplyMergeResult, anyhow::Error>>,
+            oneshot::Sender<Result<(), anyhow::Error>>,
         )>(cfg.compaction_queue_size);
         let concurrency_limit = Arc::new(tokio::sync::Semaphore::new(
             cfg.compaction_concurrency_limit,
@@ -237,14 +238,13 @@ where
                     let res = Self::compact_and_apply(&machine, req, write_schemas)
                         .instrument(compact_span)
                         .await;
-                    let res = res.map(|(res, maintenance)| {
+                    if let Ok(maintenance) = res {
                         maintenance.start_performing(&machine, &gc);
-                        res
-                    });
+                    }
 
                     // we can safely ignore errors here, it's possible the caller
                     // wasn't interested in waiting and dropped their receiver
-                    let _ = completer.send(res);
+                    let _ = completer.send(Ok(()));
 
                     // moves `permit` into async scope so it can be dropped upon completion
                     drop(permit);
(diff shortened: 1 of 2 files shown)


src/persist-client/src/internal/compact.rs
```diff
@@ -369,84 +369,56 @@ where
             compaction_schema.id,
         );
 
+        let isolated_runtime = Arc::clone(&machine.isolated_runtime);
+        let machine_clone = machine.clone();
+        let metrics_clone = Arc::clone(&machine.applier.metrics);
         let compact_span = debug_span!("compact::consolidate");
         let res = tokio::time::timeout(
             timeout,
-            // Compaction is cpu intensive, so be polite and spawn it on the isolated runtime.
```
bkirwi, line 379: Lost a comment here.

src/persist-client/src/internal/compact.rs
```diff
@@ -369,84 +369,56 @@ where
...
+        .await
+        .map_err(|e| {
+            metrics.compaction.timed_out.inc();
+            anyhow!(
+                "compaction timed out after {}s: {}",
+                timeout.as_secs_f64(),
+                e
+            )
+        })?
+        .map_err(|e| anyhow!(e))?;
```
bkirwi: Here we're replacing the old `compact` method with a composition of the three new methods on the "main" codepath, but all the tests and admin tools still hit the old method. This means that we're keeping around some almost-dead code, and we're not giving the new stuff as much coverage as we otherwise would.

IMO it would be reasonable to either allow flagging between the old and new path, or to just implement `fn compact` with calls to the new methods...

src/persist-client/src/internal/compact.rs
```diff
@@ -458,6 +430,218 @@ where
...
+                    runs,
+                    Arc::clone(&blob),
+                    Arc::clone(&metrics),
+                    Arc::clone(&shard_metrics),
+                    Arc::clone(&isolated_runtime),
+                    write_schemas.clone(),
+                )
+                .await?;
+                let (parts, run_splits, run_meta, updates) =
+                    (batch.parts, batch.run_splits, batch.run_meta, batch.len);
```
bkirwi: This code is destructuring a batch just to rebuild it from clones of all the parts, which feels odd... I think we can just pass on the batch?

src/persist-client/src/internal/compact.rs
```diff
@@ -458,6 +430,218 @@ where
...
+                assert!(
+                    (updates == 0 && parts.len() == 0) || (updates > 0 && parts.len() > 0),
+                    "updates={}, parts={}",
+                    updates,
+                    parts.len(),
+                );
+
+                if updates == 0 {
+                    continue;
+                }
```
bkirwi: This is unnecessary / has no effect currently, right?

If so let's skip it... as soon as we're doing actual incremental compactions, we'll want to treat these zero-output runs the same as every other run.

src/persist-client/src/internal/compact.rs
```diff
@@ -458,6 +430,218 @@ where
...
+    ///    number of outstanding parts we can keep in-flight to Blob.
+    pub fn compact_stream(
+        cfg: CompactConfig,
+        blob: Arc<dyn Blob>,
+        metrics: Arc<Metrics>,
+        shard_metrics: Arc<ShardMetrics>,
+        isolated_runtime: Arc<IsolatedRuntime>,
+        req: CompactReq<T>,
+        write_schemas: Schemas<K, V>,
+    ) -> impl Stream<Item = Result<CompactRes<T>, anyhow::Error>> {
```
bkirwi: Let's have this return `HollowBatch` instead of `CompactRes` for now, since it is not actually safe to apply to the spine.

bkirwi (commented):
📈!!

src/persist-client/src/internal/compact.rs
DAlperin: Good point, compact is reimplemented in terms of these new functions

src/persist-client/src/internal/compact.rs
```diff
@@ -373,84 +373,57 @@ where
...
+                timeout.as_secs_f64(),
+                e
+            )
+        })?
+        .map_err(|e| anyhow!(e))?;
 
         metrics
             .compaction
             .seconds
             .inc_by(start.elapsed().as_secs_f64());
```
bkirwi, line 413: IIUC semantics have changed here slightly due to the `?` above... this metric won't be incremented if the compaction hits an error. Should be straightforward to move the `map_err`s below this line...

src/persist-client/src/internal/compact.rs
```diff
@@ -462,6 +435,86 @@ where
...
+
+        let mut all_parts = vec![];
+        let mut all_run_splits = vec![];
+        let mut all_run_meta = vec![];
+        let mut len = 0;
+
+        while let Some(res) = stream.next().await {
+            let res = res?;
+            let (parts, updates, run_meta, run_splits) =
+                (res.parts, res.len, res.run_meta, res.run_splits);
```
bkirwi, line 453: Very non-mandatory, but if you'd like to prevent empty runs, this would be a fine place for the old:

```rust
if updates == 0 {
    continue;
}
```

check. (Dropping them in `compact_stream` caused some of the metrics anomalies last time, but that's not a risk here since we're generating one big batch no matter what.)

PR #32802, opened 2025-06-23
persist: add flag for lower bound validations on read

ptravers:
### Motivation

[9399](https://github.com/materializeinc/database-issues/issues/9399)

### Tips for reviewer

adds a feature flag to enable turning off validation of batch vs part lower bounds on read from persist.

### Checklist

### DIFF
--- a/misc/python/materialize/parallel_workload/action.py
+++ b/misc/python/materialize/parallel_workload/action.py
@@ -1078,7 +1078,7 @@ class FlipFlagsAction(Action):
         self.flags_with_values["persist_encoding_enable_dictionary"] = (
             BOOLEAN_FLAG_VALUES
         )
-        self.flags_with_values["persist_validate_part_lower_bounds_on_read"] = (
+        self.flags_with_values["persist_validate_part_bounds_on_read"] = (
             BOOLEAN_FLAG_VALUES
         )
         self.flags_with_values["compute_apply_column_demands"] = BOOLEAN_FLAG_VALUES
--- a/misc/python/materialize/parallel_workload/action.py
+++ b/misc/python/materialize/parallel_workload/action.py
@@ -1078,6 +1078,9 @@ class FlipFlagsAction(Action):
         self.flags_with_values["persist_encoding_enable_dictionary"] = (
             BOOLEAN_FLAG_VALUES
         )
+        self.flags_with_values["persist_validate_part_lower_bounds_on_read"] = (
+            BOOLEAN_FLAG_VALUES
+        )
         self.flags_with_values["compute_apply_column_demands"] = BOOLEAN_FLAG_VALUES
         self.flags_with_values["enable_alter_table_add_column"] = BOOLEAN_FLAG_VALUES
         self.flags_with_values["enable_compute_active_dataflow_cancelation"] = (
--- a/src/persist-client/src/cli/inspect.rs
+++ b/src/persist-client/src/cli/inspect.rs
@@ -36,7 +36,7 @@ use crate::async_runtime::IsolatedRuntime;
 use crate::cache::StateCache;
 use crate::cli::args::{NO_COMMIT, READ_ALL_BUILD_INFO, StateArgs, make_blob, make_consensus};
 use crate::error::CodecConcreteType;
-use crate::fetch::EncodedPart;
+use crate::fetch::{EncodedPart, FetchConfig};
 use crate::internal::encoding::{Rollup, UntypedState};
 use crate::internal::paths::{
     BlobKey, BlobKeyPrefix, PartialBatchKey, PartialBlobKey, PartialRollupKey, WriterKey,
@@ -338,6 +338,7 @@ pub async fn blob_batch_part(
     let desc = parsed.desc.clone();
 
     let encoded_part = EncodedPart::new(
+        &FetchConfig::from_persist_config(&cfg),
         metrics.read.snapshot.clone(),
         parsed.desc.clone(),
         &key.0,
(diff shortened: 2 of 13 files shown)


src/txn-wal/src/operator.rs
```diff
@@ -667,8 +669,8 @@ impl DataSubscribe {
                 (data_stream.leave(), token)
             });
             let (data, txns) = (ProbeHandle::new(), ProbeHandle::new());
-            let data_stream = data_stream.flat_map(|part| {
-                let part = part.parse();
+            let data_stream = data_stream.flat_map(move |part| {
+                let part = part.parse(persist_cfg.clone());
```
ptravers: this is all pretty gross. I wonder if it would make more sense to have the public parse just take a bool for validation enabled?

def- (approved):
Test changes lgtm

src/persist-client/src/cli/inspect.rs
```diff
@@ -369,6 +370,7 @@ pub async fn blob_batch_part(
 }
 
 async fn consolidated_size(args: &StateArgs) -> Result<(), anyhow::Error> {
+    let cfg = PersistConfig::new_default_configs(&READ_ALL_BUILD_INFO, SYSTEM_TIME.clone());
```
bkirwi: I'd use `state_versions.cfg` instead here if we can, just to avoid having to potentially-divergent configs floating around...

src/persist-client/src/fetch.rs
```diff
@@ -91,14 +92,39 @@ pub(crate) const OPTIMIZE_IGNORED_DATA_FETCH: Config<bool> = Config::new(
...
+        Self {
+            validate_lower_bounds_on_read: FETCH_VALIDATE_PART_LOWER_BOUNDS_ON_READ.get(cfg),
+        }
+    }
+
+    pub fn from_compact_config(cfg: &CompactConfig) -> Self {
+        Self {
+            validate_lower_bounds_on_read: cfg.fetch_validate_lower_bounds_on_read,
+        }
+    }
```
bkirwi, line 111: If you don't feel strongly, can we store the `FetchConfig` as a field on `CompactConfig`, like we do for the `BatchBuilderConfig`? A bit closer to the existing code / saves some duplication.

src/txn-wal/src/operator.rs
bkirwi: I think the least gross version would be to store the `FetchConfig` as a field in the `BatchFetcherConfig`, then add it as a field on `FetchedBlob`. That should prevent any changes outside the `persist-client` codebase, and it's lightly more consistent...

src/persist-client/src/fetch.rs
ptravers: yep!

src/persist-client/src/cli/inspect.rs
ptravers: yep!

src/persist-client/src/fetch.rs
ptravers: done.

src/persist-client/src/cli/inspect.rs
ptravers: done.

misc/python/materialize/parallel_workload/action.py
```diff
@@ -1078,6 +1078,9 @@ def __init__(
         self.flags_with_values["persist_encoding_enable_dictionary"] = (
             BOOLEAN_FLAG_VALUES
         )
+        self.flags_with_values["persist_validate_part_lower_bounds_on_read"] = (
+            BOOLEAN_FLAG_VALUES
+        )
```
bkirwi, line 1083: Thanks for adding these!

src/persist-client/src/fetch.rs
```diff
@@ -1211,13 +1260,15 @@ where
...
             || inline_desc.upper() != registered_desc.upper();
         if needs_truncation {
-            assert!(
-                PartialOrder::less_equal(inline_desc.lower(), registered_desc.lower()),
-                "key={} inline={:?} registered={:?}",
-                printable_name,
-                inline_desc,
-                registered_desc
-            );
+            if cfg.validate_lower_bounds_on_read {
```
bkirwi: We also need to opt out of the assert on the upper, a few lines down, yeah? (IIUC, the new append pattern in the source will have part uppers that are less than the registered batch upper, for the same reason that we'll have part lowers that are greater than the registered lower...)

src/persist-client/src/fetch.rs
```diff
@@ -631,6 +664,7 @@ pub struct FetchedBlob<K: Codec, V: Codec, T, D> {
     filter_pushdown_audit: bool,
     structured_part_audit: PartDecodeFormat,
     fetch_permit: Option<Arc<MetricsPermits>>,
+    batch_fetcher_config: BatchFetcherConfig,
```
bkirwi: We only need the `FetchConfig` here, right? (Fine like this though if you prefer it.)

src/persist-client/src/fetch.rs
bkirwi: Assuming so - you might want to rename the flag name to not explicitly mention just the lower bound. I hereby pre-approve such a change if this code changes. :)

bkirwi (approved):
This seems fine and is for sure mergeable, but possibly doesn't cover everything you'll need - see the comments.

Overall the latest take on the structure / plumbing is pretty clean - appreciate it!

src/persist-client/src/fetch.rs
ptravers: ah that's a good point. I think my head was a wee bit muddled with the release stuff as well.

src/persist-client/src/fetch.rs
ptravers: ah yeah, good point. will refactor and rename.

PR #32803, opened 2025-06-23
slt: Support multiple replicas, run with 4 replicas in nightly

def-:
Fixes: https://github.com/MaterializeInc/database-issues/issues/9405
### Checklist


ci/nightly/pipeline.template.yml
```diff
@@ -1933,18 +1933,35 @@ steps:
...
-      queue: hetzner-aarch64-16cpu-32gb
-    plugins:
-      - ./ci/plugins/mzcompose:
-          composition: sqllogictest
-          run: slow-tests
+  - group: SLT
+    key: slt
+    steps:
+      - id: sqllogictest
+        label: ":bulb: SQL logic tests (1 replica)"
```
ggevay: I'm not sure whether we need the 1-replica version. It's hard to imagine some bug occurring only with 1 replica, but not with 4 replicas. So maybe to save money, we are ok to run only the 4-replica version.

ggevay: Another thing, maybe we could have bigger replicas than in the normal CI? How about a replica size of 4. The machine has 16 CPUs, so with 4 replicas with size 4 each we still shouldn't have a big slowdown, as each replica can still have its own CPU core.

ggevay (approved):
Oh, I see this required quite a bit of fiddling with tests that were assuming 1 replica. Thank you for going through all this! Looks good, I wrote some minor comments only.

ci/nightly/pipeline.template.yml
def-: I tried with replica size 4, but many tests also have the replica size hardcoded, and I'm not sure I want to fix them all.

ggevay: Ok, thanks, then the replica size is fine as it is!

PR #32817, opened 2025-06-24
Update timely, differential, columnar dependencies

antiguru:
Update timely, differential, and columnar to their latest versions.


def- (approved):
Nightly triggered, no complaints from QA side if green: https://buildkite.com/materialize/nightly/builds/12426

teskje (approved):

PR #32819, opened 2025-06-24
adapter: in support_execute_one drop SessionClient only after receiving result rows

aljoscha:
As the comment in code explains, we would previously immediately cancel
the query before we have a chance to run the peek or read the results.

This went unnoticed because this code path is only used in our
telemetry/segment reporting, which we only do in production. I'm adding
a follow-up commit that enables at least some form of telemetry
reporting in testing.

### DIFF
--- a/src/adapter/src/client.rs
+++ b/src/adapter/src/client.rs
@@ -402,11 +402,24 @@ Issue a SQL query to get started. Need help?
         session_client
             .declare(EMPTY_PORTAL.into(), stmt, sql.to_string())
             .await?;
+
         match session_client
             .execute(EMPTY_PORTAL.into(), futures::future::pending(), None)
             .await?
         {
-            (ExecuteResponse::SendingRowsStreaming { rows, .. }, _) => Ok(rows),
+            (ExecuteResponse::SendingRowsStreaming { mut rows, .. }, _) => {
+                // We have to only drop the session client _after_ we read the
+                // result. Otherwise the peek will get cancelled right when we
+                // drop the session client. So we wrap it up in an extra stream
+                // like this, which owns the client and can return it.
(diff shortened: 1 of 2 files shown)


ggevay:
Ah wait, we have a problem: slts are hanging forever, because they expect tasks to stop themselves, but the task that `start_reporting` starts (which it didn't do before in slts) is an infinite loop.

You can repro locally with e.g.
```
bin/sqllogictest -- -v test/sqllogictest/table_func.slt
```

src/adapter/src/client.rs
```diff
@@ -402,11 +402,24 @@ Issue a SQL query to get started. Need help?
...
+
         match session_client
             .execute(EMPTY_PORTAL.into(), futures::future::pending(), None)
             .await?
         {
-            (ExecuteResponse::SendingRowsStreaming { rows, .. }, _) => Ok(rows),
+            (ExecuteResponse::SendingRowsStreaming { mut rows, .. }, _) => {
+                // We have to only drop the session client _after_ we read the
+                // result. Otherwise the peek will get cancelled right when we
+                // ddrop the session client. So we wrap it up in an extra stream
```
antiguru:
```suggestion
                // drop the session client. So we wrap it up in an extra stream
```

aljoscha:
> soft_panic

Wild! Are there no other tasks in `envd`/adapter that run forever!?!

ggevay:
> Wild! Are there no other tasks in envd/adapter that run forever!?!

It seems so, yes!

Btw. this is the same problem that Dennis and Peter debugged last week with my tokio canary PR:
https://materializeinc.slack.com/archives/CTESPM7FU/p1750363355471699?thread_ts=1750338982.832639&cid=CTESPM7FU

PR #32837, opened 2025-06-25
ci: Ignore failures for known issues with ci-ignore-failure: true

def-:
to prevent us from having to disable too many flaky issues, we can still keep track of what issues are happening and would still notice other issues in the same test

### Checklist


teskje (approved):

doc/developer/ci-regexp.md
```diff
@@ -19,3 +19,5 @@ Examples:
 ci-apply-to: SQLsmith
 ci-apply-to: sqlsmith explain
 ```
+
+When a test issue is discovered and the GitHub issue is still open, you may want the failure to not cause a test failure, so that CI does not become flaky, but still add an annotation and record the failure into https://ci-failures.dev.materialize.com. You can use `ci-ignore-failure: true` for that. Note that this will even cause us to ignore other errors happening in the same test run, since there is no way to tell the difference between a different error and a follow-up error.
```
teskje: The note at the end is not true anymore, right?

def-:
One remaining question: Do we need a separate `ci-ignore-failure: true` pseudo comment or should we consider every issue marked with the `ci-flake` tag so implicitly?

def-:
These issues are happening so often that I opted not to reenable the tests that exercise them, but at least they should all be easy to reproduce/debug from there:
- https://github.com/MaterializeInc/database-issues/issues/8636
- https://github.com/MaterializeInc/database-issues/issues/8447
- https://github.com/MaterializeInc/database-issues/issues/7775
- https://github.com/MaterializeInc/database-issues/issues/8619

PR #32838, opened 2025-06-25
Increase threshold for slow optimization warning in debug builds

ggevay:
Optimizations are much slower in debug builds, so this PR increases the slow optimization warning threshold in debug builds, to not spam people screens with the big warnings. ([slack discussion](https://materializeinc.slack.com/archives/C08ACQNGSQK/p1750844468500809?thread_ts=1750844293.433969&cid=C08ACQNGSQK))

### DIFF
--- a/src/sql/src/optimizer_metrics.rs
+++ b/src/sql/src/optimizer_metrics.rs
@@ -12,6 +12,7 @@
 use std::cell::RefCell;
 use std::time::Duration;
 
+use mz_ore::assert::soft_assertions_enabled;
 use mz_ore::metric;
 use mz_ore::metrics::MetricsRegistry;
 use mz_ore::stats::histogram_seconds_buckets;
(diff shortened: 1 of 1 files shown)


src/sql/src/optimizer_metrics.rs
```diff
@@ -66,7 +67,19 @@ impl OptimizerMetrics {
         self.e2e_optimization_time_seconds
             .with_label_values(&[object_type])
             .observe(duration.as_secs_f64());
-        if duration > self.e2e_optimization_time_seconds_log_threshold {
+        // Also log it when it's big.
+        let debug_threshold = soft_assertions_enabled();
```
def-, line 71: Wait, I think we also enable soft assertions in release mode in CI. Can‘t we check how the program was compiled?

def-: Sorry for only thinking of this after the review

ggevay: Ahh, yes! Would using `cfg!(debug_assertions)` be better? `soft_assertions_enabled` is
```
pub static SOFT_ASSERTIONS: AtomicBool = {
    let default = cfg!(debug_assertions) || crate::env::is_var_truthy("MZ_SOFT_ASSERTIONS");
    AtomicBool::new(default)
};
```

def-: yes

ggevay: PR: https://github.com/MaterializeInc/materialize/pull/32841
Thanks for spotting this!

PR #32846, opened 2025-06-25
[persist] More evenly-sized GC runs

bkirwi:
### Checklist


src/persist-client/src/internal/apply.rs
```diff
@@ -494,6 +494,17 @@ where
         let is_rollup = cmd.name == metrics.cmds.add_rollup.name;
         let is_become_tombstone = cmd.name == metrics.cmds.become_tombstone.name;
 
+        let gc_config = GcConfig {
```
DAlperin, line 497: nice this is better

src/persist-client/src/internal/state.rs
```diff
@@ -3823,18 +3849,35 @@ pub(crate) mod tests {
...
+        assert_eq!(state.seqno_since(), SeqNo(10_000));
+
+        let now = now_fn();
+        assert_eq!(
+            state.maybe_gc(true, now, GC_CONFIG),
+            Some(GcReq {
+                shard_id: state.shard_id,
+                new_seqno_since: SeqNo(900)
+            })
+        );
```
DAlperin: We could add one more test to prove that the number we advance by is equal to the max in this case but it's not required

bkirwi: I've reworked the test to make it more explicit that we're bounded by the max-versions limit here. If that doesn't satisfy I'm happy to follow up!

PR #32847, opened 2025-06-25
Fix disk reporting for local setup

SangJunBak:
Fix documentation for problem stated in https://materializeinc.slack.com/archives/C08ACQNGSQK/p1750881871059709?thread_ts=1750881519.933969&cid=C08ACQNGSQK

### Checklist

### DIFF
--- a/doc/user/layouts/shortcodes/self-managed/versions/curl-sample-files-local-install.html
+++ b/doc/user/layouts/shortcodes/self-managed/versions/curl-sample-files-local-install.html
@@ -19,3 +19,5 @@ curl -o sample-materialize.yaml https://raw.githubusercontent.com/MaterializeInc
 
 These configuration files are for local evaluation/testing purposes only and
 not intended for production use.
+
+To enable disk metrics reporting, configure the `storage.storageClass.name` field to `standard` in `sample-values.yaml`.


doc/user/layouts/shortcodes/self-managed/versions/curl-sample-files-local-install.html
```diff
@@ -19,3 +19,5 @@
 
 These configuration files are for local evaluation/testing purposes only and
 not intended for production use.
+
+To enable disk metrics reporting, configure the `storage.storageClass.name` field to `standard` in `sample-values.yaml`.
```
jubrad, line 23: uh... I think we should have more docs than this. I assume this is in the helm values for the operator?

kay-kim: Yeah .. if you want to allow people to set it for local kind/minikube ...
See https://materialize.com/docs/self-managed/v25.2/installation/configuration/#storage-parameters

jubrad: Also is there any reason we shouldn't just set this to be the default?

jubrad (approved):

doc/user/layouts/shortcodes/self-managed/versions/curl-sample-files-local-install.html
SangJunBak: >uh... I think we should have more docs than this. 

Ah like an explanation why we need to do this to enable disk metrics reporting? Yeah that makes sense to me.

SangJunBak:
Talked to @kay-kim offline about this and given this is broken for kind and minikube anyways, it's debatable if we even wanna merge this... I'll follow up to see how hard it'd be to fix this but otw, holding off on merging

Unless we have a volume per cluster replica, this won't work for local kind and minikube

PR #32849, opened 2025-06-26
cluster: don't panic on `LocalClient` send failure

teskje:
The `LocalClient` made the incorrect assumption that its receiver would never drop first. It is incorrect at least if the receiver is a compute worker. Compute workers can learn about disconnects from other workers, before their own GRPC server learns about the disconnect, in which case they'll drop their receiver before the `LocalClient`'s sender is dropped.

The fix is to not panic in the `LocalClient` on send failure but to return an error instead. The error will make the GRPC server end the connection and wait for a new one.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9414

### Checklist

### DIFF
--- a/src/service/src/local.rs
+++ b/src/service/src/local.rs
@@ -12,6 +12,7 @@
 use std::fmt;
 use std::thread::Thread;
 
+use anyhow::anyhow;
 use async_trait::async_trait;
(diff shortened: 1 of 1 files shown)


antiguru (approved):
Thank you!

PR #32851, opened 2025-06-26
Selective www authenticate header

alex-hunt-materialize:
* Redirect HTTP to HTTPS instead of returning unauthorized.
* Allow basic auth for `AuthenticatorKind::Password`. We still support sessions, but they are no longer the only way to authenticate. At present, this doesn't initiate a session.
* Selectively add the `www-authenticate` header to only some `AuthError`s. This may be an incomplete list, and it feels a bit gross to do additional path comparisons in the auth functions, but it should unblock access to the debug pages without spamming other pages with login boxes.
* Renamed `AuthError::InvalidLogin` to `AuthError::RoleDisallowed` for clarity.

### Motivation

The debug pages have no way of logging in with `AuthenticatorKind::Password`.
Discussion at https://materializeinc.slack.com/archives/C07PN7KSB0T/p1750849485776399

Redirecting HTTP to HTTPS is a better user experience.
Allowing basic auth rather than forcing a login request to create a session is a better user experience.

### Tips for reviewer

It is easier to review commit by commit.

### Checklist

### DIFF
--- a/src/environmentd/src/http.rs
+++ b/src/environmentd/src/http.rs
@@ -686,8 +686,8 @@ where
 
 #[derive(Debug, Error)]
 enum AuthError {
-    #[error("invalid username in client certificate")]
-    InvalidLogin(String),
+    #[error("role dissallowed")]
+    RoleDisallowed(String),
     #[error("{0}")]
     Frontegg(#[from] FronteggError),
     #[error("missing authorization header")]
@@ -698,6 +698,8 @@ enum AuthError {
     SessionExpired,
     #[error("failed to update session")]
     FailedToUpdateSession,
+    #[error("invalid credentials")]
+    InvalidCredentials,
 }
 
 impl IntoResponse for AuthError {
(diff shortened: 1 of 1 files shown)


src/environmentd/src/http.rs
```diff
@@ -821,7 +841,18 @@ async fn http_auth(
         None
     };
 
-    let user = auth(&authenticator, creds, allowed_roles).await?;
+    let path = req.uri().path();
+    let include_www_authenticate_header = path == "/"
+        || ["/memory", "/hierarchical-memory", "/prof/"]
```
jubrad: Can we move these into a `METRICS_API_ENDPOINTS` constant

jubrad (approved):
One small recommendation, otherwise solid

src/environmentd/src/http.rs
alex-hunt-materialize: Moving to a `PROFILING_API_ENDPOINTS` constant, as these are not related to metrics.

PR #32855, opened 2025-06-26
fix 0dt cluster reconfig

jubrad:
This limits the scope of storage collections we're checking to collections on the cluster we're attempting to alter.

### Motivation
This was previously holding the alter until all storage collections across the env were hydrated.

### Checklist

### DIFF
--- a/src/storage-controller/src/lib.rs
+++ b/src/storage-controller/src/lib.rs
@@ -398,8 +398,15 @@ where
     fn collections_hydrated_on_replicas(
         &self,
         target_replica_ids: Option<Vec<ReplicaId>>,
+        target_cluster_id: &StorageInstanceId,
         exclude_collections: &BTreeSet<GlobalId>,
     ) -> Result<bool, StorageError<Self::Timestamp>> {
+        // If an empty set of replicas is provided there can be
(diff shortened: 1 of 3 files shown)


src/storage-controller/src/lib.rs
```diff
@@ -410,14 +417,20 @@ where
             if collection_id.is_transient() || exclude_collections.contains(collection_id) {
                 continue;
             }
-            let hydrated = match &collection_state.extra_state {
+            let hydrated_or_untargeted = match &collection_state.extra_state {
```
aljoscha: We could leave this as `hydrated`, and inside check for `state.instance_id != target_cluster_id` and then do a `continue`. That's also what we do right above, and it keeps the nesting level low

jubrad: ah... that is how I had it originally. I'll go back!

aljoscha: hehe, you didn't push that change back up yet, right?

aljoscha: either way, I think we can merge as is but thought the `continue` would be a bit nicer

PR #32862, opened 2025-06-26
Support incrementally applying compaction results

DAlperin:
Towards: https://github.com/MaterializeInc/database-issues/issues/9194

### Checklist

### DIFF
--- a/src/persist-client/src/batch.rs
+++ b/src/persist-client/src/batch.rs
@@ -685,23 +685,22 @@ where
         let write_run_ids = self.parts.cfg.enable_incremental_compaction;
         let batch_delete_enabled = self.parts.cfg.batch_delete_enabled;
         let shard_metrics = Arc::clone(&self.parts.shard_metrics);
-        let runs = self.parts.finish().await; // Now returns update counts too
+        let runs = self.parts.finish().await;
 
         let mut run_parts = vec![];
         let mut run_splits = vec![];
         let mut run_meta = vec![];
         let total_updates = runs
             .iter()
-            .map(|(_, _, update_counts)| update_counts.iter().sum::<usize>())
-            .sum();
-        for (order, parts, update_counts) in runs {
+            .map(|(_, _, num_updates)| num_updates)
+            .sum::<usize>();
+        for (order, parts, num_updates) in runs {
             if parts.is_empty() {
                 continue;
             }
             if run_parts.len() != 0 {
                 run_splits.push(run_parts.len());
             }
-            let num_updates: usize = update_counts.iter().sum();
             run_meta.push(RunMeta {
                 order: Some(order),
                 schema: self.write_schemas.id,
@@ -1300,7 +1299,7 @@ impl<T: Timestamp + Codec64> BatchParts<T> {
     }
 
     #[instrument(level = "debug", name = "batch::finish_upload", fields(shard = %self.shard_id))]
-    pub(crate) async fn finish(self) -> Vec<(RunOrder, Vec<RunPart<T>>, Vec<usize>)> {
+    pub(crate) async fn finish(self) -> Vec<(RunOrder, Vec<RunPart<T>>, usize)> {
         match self.writing_runs {
             WritingRuns::Ordered(RunOrder::Unordered, run) => {
                 let completed_runs = run.finish();
@@ -1309,11 +1308,7 @@ impl<T: Timestamp + Codec64> BatchParts<T> {
                     let completed_run = completed_run.into_result().await;
                     // Each part becomes its own run for unordered case
                     for part in completed_run.parts {
-                        output.push((
-                            RunOrder::Unordered,
-                            vec![part],
-                            vec![completed_run.num_updates],
-                        ));
+                        output.push((RunOrder::Unordered, vec![part], completed_run.num_updates));
                     }
                 }
                 output
@@ -1321,11 +1316,11 @@ impl<T: Timestamp + Codec64> BatchParts<T> {
             WritingRuns::Ordered(order, run) => {
                 let completed_runs = run.finish();
                 let mut all_parts = Vec::new();
-                let mut all_update_counts = Vec::new();
+                let mut all_update_counts = 0;
                 for completed_run in completed_runs {
                     let completed_run = completed_run.into_result().await;
                     all_parts.extend(completed_run.parts);
-                    all_update_counts.push(completed_run.num_updates);
+                    all_update_counts += completed_run.num_updates;
                 }
                 vec![(order, all_parts, all_update_counts)]
             }
@@ -1334,15 +1329,8 @@ impl<T: Timestamp + Codec64> BatchParts<T> {
                 let mut output = Vec::new();
                 for (order, run) in runs {
                     let completed_run = run.into_result().await;
-                    // For compacting, each CompletedRun represents one logical run.
-                    // The update count is for the entire run, not per part.
-                    // We store the total update count only once in the first position,
-                    // and zeros for the remaining parts to avoid redundant storage.
-                    let mut update_counts = vec![0; completed_run.parts.len()];
-                    if !update_counts.is_empty() {
-                        update_counts[0] = completed_run.num_updates;
-                    }
-                    output.push((order, completed_run.parts, update_counts));
+
+                    output.push((order, completed_run.parts, completed_run.num_updates));
                 }
                 output
             }
--- a/src/persist-client/src/batch.rs
+++ b/src/persist-client/src/batch.rs
@@ -690,6 +690,10 @@ where
         let mut run_parts = vec![];
         let mut run_splits = vec![];
         let mut run_meta = vec![];
+        let total_updates = runs
+            .iter()
+            .map(|(_, _, update_counts)| update_counts.iter().sum::<usize>())
+            .sum();
         for (order, parts, update_counts) in runs {
             if parts.is_empty() {
                 continue;
@@ -717,7 +721,6 @@ where
             run_parts.extend(parts);
         }
         let desc = registered_desc;
-        let num_updates = run_meta.iter().filter_map(|m| m.len).sum();
 
         let batch = Batch::new(
             batch_delete_enabled,
@@ -725,7 +728,7 @@ where
             self.blob,
             shard_metrics,
             self.version,
-            HollowBatch::new(desc, run_parts, num_updates, run_meta, run_splits),
+            HollowBatch::new(desc, run_parts, total_updates, run_meta, run_splits),
         );
 
         Ok(batch)
--- a/src/persist-client/src/batch.rs
+++ b/src/persist-client/src/batch.rs
@@ -639,7 +639,6 @@ where
     metrics: Arc<Metrics>,
 
     write_schemas: Schemas<K, V>,
-    num_updates: Vec<usize>,
     parts: BatchParts<T>,
 
     // These provide a bit more safety against appending a batch with the wrong
@@ -667,7 +666,6 @@ where
             blob,
             metrics,
             write_schemas,
-            num_updates: vec![],
             parts,
             shard_id,
             version,
@@ -687,22 +685,19 @@ where
         let write_run_ids = self.parts.cfg.enable_incremental_compaction;
         let batch_delete_enabled = self.parts.cfg.batch_delete_enabled;
         let shard_metrics = Arc::clone(&self.parts.shard_metrics);
-        let runs = self.parts.finish().await;
+        let runs = self.parts.finish().await; // Now returns update counts too
 
         let mut run_parts = vec![];
         let mut run_splits = vec![];
         let mut run_meta = vec![];
-        let mut part_cursor = 0;
-        for (order, parts) in runs {
+        for (order, parts, update_counts) in runs {
             if parts.is_empty() {
                 continue;
             }
             if run_parts.len() != 0 {
                 run_splits.push(run_parts.len());
             }
-            let num_updates = self.num_updates[part_cursor..part_cursor + parts.len()]
-                .iter()
-                .sum();
+            let num_updates: usize = update_counts.iter().sum();
             run_meta.push(RunMeta {
                 order: Some(order),
                 schema: self.write_schemas.id,
@@ -719,11 +714,10 @@ where
                     None
                 },
             });
-            part_cursor += parts.len();
             run_parts.extend(parts);
         }
         let desc = registered_desc;
-        let num_updates = self.num_updates.iter().sum();
+        let num_updates = run_meta.iter().filter_map(|m| m.len).sum();
 
         let batch = Batch::new(
             batch_delete_enabled,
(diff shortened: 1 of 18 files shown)


src/persist-client/src/batch.rs
```diff
@@ -692,13 +692,17 @@ where
...
         for (order, parts) in runs {
             if parts.is_empty() {
                 continue;
             }
             if run_parts.len() != 0 {
                 run_splits.push(run_parts.len());
             }
+            let num_updates = self.num_updates[part_cursor..part_cursor + parts.len()]
+                .iter()
+                .sum();
```
bkirwi: This sort of thing is going to give the wrong answer when used with `BatchParts::new_compacting`, since that can change the total number of parts returned as runs get compacted together. I think we'll need to make `BatchParts` track this internally.

(It might be easiest to just have the `BatchParts` thread around a `RunMeta` for each run, instead of just the order. Whatever feels cleaner!)

bkirwi: Interesting this didn't show up in CI even with all the flags enabled... maybe we need to tune some more stuff more aggressively.

src/persist-client/src/cli/admin.rs
```diff
@@ -739,14 +743,18 @@ pub async fn dangerous_force_compaction_and_break_pushdown<K, V, T, D>(
         let (reqs, mut maintenance) = machine.spine_exert(fuel).await;
         for req in reqs {
             info!(
-                "force_compaction {} {} compacting {} batches in {} parts totaling {} bytes: lower={:?} upper={:?} since={:?}",
+                "force_compaction {} {} compacting {} batches in {} parts with {} runs totaling {} bytes: lower={:?} upper={:?} since={:?}",
```
bkirwi, line 746: 🆒

src/persist-client/src/fetch.rs
```diff
@@ -1297,10 +1300,31 @@ where
                 registered_desc
             );
         } else {
+            // Due to incremental compaction, we may have a batch that has an inline
+            // desc that has the same lower and upper as the registered desc,
+            // but an older since.
```
bkirwi, line 1305: This sounds right!

src/persist-client/src/batch.rs
```diff
@@ -831,6 +842,7 @@ impl<T: Timestamp + Codec64> BatchParts<T> {
                                         } else {
                                             None
                                         },
+                                        len: None,
```
bkirwi: And yeah, if we plumed around the `RunMeta` we could probably fill this one in...

src/persist-client/src/internal/compact.rs
```diff
@@ -385,6 +394,23 @@ where
...
+                        .all(|x| x.batch.runs().all(|(meta, _)| meta.id.is_some()));
+                    let all_runs_have_len = req
+                        .inputs
+                        .iter()
+                        .all(|x| x.batch.runs().all(|(meta, _)| meta.len.is_some()));
+
+                    let incremental_enabled = ENABLE_INCREMENTAL_COMPACTION
+                        .get(&machine_clone.applier.cfg)
+                        && all_runs_have_uuids
+                        && all_runs_have_len;
```
bkirwi, line 414: Good!

src/persist-client/src/internal/compact.rs
```diff
@@ -668,107 +784,162 @@ where
...
+                acc.entry(item.0).or_default().push(item.1);
+                acc
+            });
+
+        let all_batches_have_one_run = grouped.values().all(|runs| runs.len() <= 1);
+
+        let mut chunks = Vec::new();
+
+        // There are two cases to consider here:
+        // 1. There are as many runs as there are input batches in which case we
```
bkirwi: This comment's a little stale!

src/persist-client/src/internal/compact.rs
```diff
@@ -783,14 +954,14 @@ where
...
                             ordered_runs.push((
+                                run_id,
                                 desc,
                                 meta,
                                 Cow::Owned(vec![RunPart::Single(part?.into_owned())]),
                             ));
                         }
                     }
                 }
-                batch_runs.push_back((desc, runs));
```
bkirwi, line 793: Nit: now that this is gone, we don't need a vec deque at all. This can just be a normal iteration.

src/persist-client/src/internal/state.rs
```diff
@@ -1217,10 +1219,11 @@ pub struct HollowRollup {
 }
 
 /// A pointer to a blob stored externally.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 pub enum HollowBlobRef<'a, T> {
     Batch(&'a HollowBatch<T>),
     Rollup(&'a HollowRollup),
+    Part(&'a RunPart<T>),
```
bkirwi: Yeah, feeling the same about this on reflection!

Since `blob_deletes` only has one caller - and rollups and batches are handled differently anyways - the option that strikes me as lowest impact is to break that method up into `rollup_deletes` (which can stay the same) and `rollup_parts` (your new thing) and then we don't need to touch the type.

src/persist-client/src/internal/state.rs
```diff
@@ -4279,4 +4289,14 @@ pub(crate) mod tests {
...
+
+    #[mz_ore::test]
+    fn runid_parse() {
+        let runid = "rif00c21f0-6907-4035-84d9-0cd07ae4f8c3";
+        let parsed = RunId::from_str(runid);
+
+        println!("Parsed RunId: {:?}", parsed);
+
+        assert!(parsed.is_ok());
+    }
```
bkirwi, line 4326: This would be a good chance to test the id roundtripping that caught us up last time - re-encode and check it's equal. (Bonus points for a proptest.)

src/persist-client/src/internal/trace.rs
```diff
@@ -273,6 +282,7 @@ impl<T: Timestamp + Lattice> Trace<T> {
         }
     }
     pub(crate) fn unflatten(value: FlatTrace<T>) -> Result<Self, String> {
+        let flat_trace_clone = value.clone();
```
bkirwi: Dead!

src/persist-client/src/internal/trace.rs
```diff
@@ -358,7 +368,7 @@ impl<T: Timestamp + Lattice> Trace<T> {
                     batch = Arc::new(HollowBatch::empty(Description::new(
                         batch.desc.lower().clone(),
                         new_upper,
-                        expected_desc.since().clone(),
+                        batch.desc.since().clone(),
```
bkirwi, line 370: Is this load-bearing? You mentioned issues in flattening and unflattening, but for recent code IIUC the expected and actual desc really ought to match...

src/persist-client/src/internal/trace.rs
```diff
@@ -694,8 +704,12 @@ enum SpineLog<'a, T> {
     },
     Disabled,
 }
+/// A RunId uniquely identifies a run within a hollow batch.
+/// It is a pair of `SpineId` and an index within that batch.
```
bkirwi: This comment's a bit stale, and we should make sure to document what it means when the run id is `None`.

src/persist-client/src/internal/trace.rs
```diff
@@ -918,31 +939,190 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
+        }
+
+        Self::diffs_sum(parts.iter(), metrics)
+    }
+
+    fn construct_batch_with_runs_replaced(
+        original: &HollowBatch<T>,
+        run_ids: &[RunId],
+        replacement: &HollowBatch<T>,
+    ) -> Result<HollowBatch<T>, ApplyMergeResult> {
```
bkirwi, line 996: Still feeling this is a bit too complex for the work it's doing. We chatted about some ideas in the call, but let me know if none of them pan out...

src/persist-client/src/internal/trace.rs
```diff
@@ -918,31 +939,190 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
+        Ok(HollowBatch {
+            desc: desc.clone(),
+            len,
+            parts,
+            run_meta,
+            run_splits,
+        })
     }
 
-    // TODO: Roundtrip the SpineId through FueledMergeReq/FueledMergeRes?
```
bkirwi, line 944: ✅

src/persist-client/src/internal/trace.rs
```diff
@@ -918,31 +939,190 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
-    /// Checked variant that performs diff sum assertions
     fn maybe_replace_checked<D>(
```
bkirwi, line 1221: This also is creeping up in complexity, but I don't have immediate suggestions... I will play around with it and follow up.

src/persist-client/src/internal/trace.rs
DAlperin: It seemed to be for reasons that I cannot fathom. I will test again

src/persist-client/src/internal/datadriven.rs
```diff
@@ -190,7 +190,7 @@ mod tests {
     async fn machine(dyncfgs: ConfigUpdates) {
         use crate::internal::machine::datadriven as machine_dd;
 
-        ::datadriven::walk_async("tests/machine", |mut f| {
+        ::datadriven::walk_async("tests/adhoc", |mut f| {
```
bkirwi: 👀

src/persist-client/src/internal/state_serde.json
```diff
@@ -708,15 +713,5 @@
...
-      "since": {
-        "elements": []
-      },
-      "remaining_work": 0,
-      "active_compaction": {
-        "start_ms": 0
-      }
-    }
-  }
+  "merges": {}
```
bkirwi: Spooky! Why do we no longer include the merge here?

src/persist-client/src/batch.rs
```diff
@@ -687,18 +685,23 @@ where
         let write_run_ids = self.parts.cfg.enable_incremental_compaction;
         let batch_delete_enabled = self.parts.cfg.batch_delete_enabled;
         let shard_metrics = Arc::clone(&self.parts.shard_metrics);
-        let runs = self.parts.finish().await;
+        let runs = self.parts.finish().await; // Now returns update counts too
```
bkirwi: Nit: comment should describe the current state and not what changed.

src/persist-client/src/batch.rs
```diff
@@ -687,18 +685,23 @@ where
...
 
         let mut run_parts = vec![];
         let mut run_splits = vec![];
         let mut run_meta = vec![];
-        for (order, parts) in runs {
+        let total_updates = runs
+            .iter()
+            .map(|(_, _, update_counts)| update_counts.iter().sum::<usize>())
+            .sum();
```
bkirwi: `finish` returns a vec of usize, but this is the only use. Might as well just be a `usize`.

src/persist-client/src/batch.rs
```diff
@@ -1240,30 +1300,49 @@ impl<T: Timestamp + Codec64> BatchParts<T> {
...
-                let mut output = Vec::with_capacity(runs.len());
+                let mut output = Vec::new();
                 for (order, run) in runs {
-                    let run = run.into_result().await;
-                    output.push((order, run))
+                    let completed_run = run.into_result().await;
+                    // For compacting, each CompletedRun represents one logical run.
+                    // The update count is for the entire run, not per part.
+                    // We store the total update count only once in the first position,
+                    // and zeros for the remaining parts to avoid redundant storage.
```
bkirwi: Once `finish` no longer returns a `Vec<usize>`, this sort of thing will no longer be required!

src/persist-client/src/internal/compact.rs
```diff
@@ -631,8 +726,28 @@ where
                     batch.parts.len(),
                 );
 
+                // Set up active compaction metadata
+                let clock = SYSTEM_TIME.clone();
```
bkirwi: Can't do it! Similar to hardcoding the diff from a few PRs ago, this replaces the user-configurable now fn with a hardcoded default.

src/persist-client/src/internal/trace.rs
```diff
@@ -1012,7 +1203,213 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
+            let batch = &batch.batch;
+            let run_ids = inputs
+                .values()
+                .next()
+                .unwrap()
+                .iter()
+                .filter_map(|id| id.1)
+                .collect::<Vec<_>>();
+
+            // backwards compatibility: if the run_ids are empty, we assume we want to replace all runs
```
bkirwi: I don't think this is safe: if run ids are not present in one of the batches we're compacting, and we do an incremental / partial compaction for that batch, we will get just a subset of the runs here but treat it as though it is a replacement for the full batch since no run ids are present.

src/persist-client/src/internal/trace.rs
```diff
@@ -1012,7 +1203,213 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
+            if let (Some(old_diffs_sum), Some(new_diffs_sum)) = (old_diffs_sum, new_diffs_sum) {
+                if old_diffs_sum != new_diffs_sum {
+                    warn!(
+                        "merge res diffs sum ({:?}) did not match spine batch diffs sum ({:?})",
+                        new_diffs_sum, old_diffs_sum
+                    );
+                    warn!("merge res: {res:#?}");
+                    warn!("spine batch: {batch:#?}");
+                    warn!("run_ids: {run_ids:?}");
+                    warn!("replacement_range: {replacement_range:?}");
```
bkirwi: Instead of many warn lines, which can get split up by the logging infra, you probably want structured parameters like this: https://github.com/MaterializeInc/materialize/blob/e3c1b3b3620ea6d6c619f59efbfada11a5575ecb/src/storage/src/source/reclock/compat.rs#L281-L286

src/persist-client/src/internal/state_serde.json
DAlperin: We loosened the "is_compact" definition to only care about the number of runs rather than the number of batches, so the batches generated by the proptest no longer require merging. https://github.com/MaterializeInc/materialize/pull/32862/files. We can adjust this proptest if desired, but I think this is actually a sign we are doing what we expected.

bkirwi: Mmm... I think the new definition is a bug in that case! This spine batch has multiple runs, one per hollow batch, and we absolutely do want to compact in that case. (And you've written code to do it if it's triggered!)

src/persist-client/src/internal/trace.rs
```diff
@@ -792,9 +825,13 @@ impl<T: Timestamp + Lattice> SpineBatch<T> {
...
-        // This definition is extremely likely to change, but for now, we consider a batch
-        // "compact" if it has at most one hollow batch with at most one run.
-        self.parts.len() <= 1 && self.parts.iter().all(|p| p.batch.run_splits.is_empty())
+        // A compact batch has at most one run.
+        // This check used to be if there was at most one hollow batch with at most one run,
+        // but that was a bit too strict since introducing incremental compaction.
+        // Incremental compaction can result in a batch with a single run, but multiple empty
+        // hollow batches, which we still consider compact. As levels are merged, we
+        // will eventually clean up the empty hollow batches.
+        self.parts.iter().all(|p| p.batch.run_splits.is_empty())
```
bkirwi: Re: the other comment, the comment doesn't match the logic... if we have 10 hollow batches with one run each this will still return true.

PR #32865, opened 2025-06-27
Add missing Columnar implementations

antiguru:
Add missing columnar implementations

### DIFF
--- a/src/repr/src/timestamp.rs
+++ b/src/repr/src/timestamp.rs
@@ -74,26 +74,26 @@ mod columnar_timestamp {
 
     /// A newtype wrapper for a vector of `Timestamp` values.
     #[derive(Clone, Copy, Default, Debug)]
-    pub struct TimestampVec<T>(T);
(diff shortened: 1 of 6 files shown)


src/repr/src/timestamp.rs
```diff
@@ -67,6 +67,103 @@ impl RustType<ProtoTimestamp> for Timestamp {
...
 }
 
+mod columnar_timestamp {
+    use crate::Timestamp;
+    use columnar::Columnar;
+    use mz_ore::cast::CastFrom;
+
+    /// A newtype wrapper for a vector of `Timestamp` values.
+    #[derive(Clone, Copy, Default, Debug)]
+    pub struct TimestampVec<T>(T);
```
teskje: Bike-shedding: Since the wrapped type doesn't have to be a `Vec`, I wonder if a more generic name would avoid some confusion. How about `Timestamps`?

teskje (approved):

src/repr/src/timestamp.rs
antiguru: Works for me!

PR #32868, opened 2025-06-27
docs: Temp use 0.147.1 for the emulator (will revert when the latest …

kay-kim:
…versions works


SangJunBak:
Sad because this uses the Console version w/o the self managed auth changes (25.2.2), but when Console 25.2.3 comes out, it should fix this 🤞

SangJunBak (approved):
Good on my end! Will try to get this fixed ASAP given emulator might be missing useful features from 0.147.3

PR #32872, opened 2025-06-27
[sinks] Replace the assert in a kafka-resumption test

bkirwi:
The old version of this assert used a metric that no longer exists, and was oddly implementation-dependent. (The old implementation might have resumed from offset 3 when it had successfully imported offset 6, but there was no reason it couldn't have chosen any other offset.)

This PR removes the old assert and replaces it with an assert on the progress collection, since that's the most direct way to get access to an offset today.

### Motivation

https://github.com/MaterializeInc/database-issues/issues/6271


def- (approved):

def-:
Can we close the issue then?

bkirwi:
Closed the issue!

PR #32888, opened 2025-06-30
compute: use protocol nonces to identify controller connections

teskje:
This PR makes the compute channel adapter use protocol nonces, instead of local epochs, to identify controller connections. These nonces are the same between workers and thus using them in the channel adapter avoids that different workers associate different epochs with the same iteration of the compute protocol. This in turn avoids bugs where workers would send responses to the wrong client connections or would skip sending responses they should send. See https://github.com/MaterializeInc/database-issues/issues/9415#issuecomment-3019512619 for details.

Here we add the nonce to the cluster protocol by simply replacing the unused `CreateTimely::epoch` field. We also take the opportunity to remove some dead epoch plumbing code. The existing epoch was not useful because it isn't unique across envd restarts in read-only environments.

Note that this also adds a nonce to the storage protocol, even though it remains unused. This is mostly to keep the two protocols as similar as possible. But also I suspect that we might find a bug in the handling of storage commands and responses at some point that will require a nonce there as well.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9415

### Tips for reviewer

Lmk if you can think of alternative solutions. It would be nice to not need to keep around the `CreateTimely` command just for the nonce, but I think we can't get away with giving each of the workers an ID that uniquely identifies the connection.

### Checklist


teskje:
TFTR!

PR #32897, opened 2025-07-02
cluster: use CTP as the transport for compute and storage protocol connections

teskje:
This PR starts making use of the Cluster Transport Protocol introduced in #32330, to replace gRPC in both the compute and the storage protocol. CTP has been designed with that use case in mind, so the amount of required changes is relatively low.

To de-risk the rollout, we want to be able to fall back to gRPC for now. To this end, a feature flag, `enable_ctp_cluster_protocols` is introduced and wired through the controller. Replicas need to select the controller protocol at startup and they do so using a new clusterd command line flag, `--use-ctp`.

### Motivation

Part of https://github.com/MaterializeInc/database-issues/issues/5038.

### Checklist

### DIFF
--- a/src/compute-client/src/controller/replica.rs
+++ b/src/compute-client/src/controller/replica.rs
@@ -23,6 +23,7 @@ use mz_ore::retry::{Retry, RetryState};
 use mz_ore::task::AbortOnDropHandle;
 use mz_service::client::GenericClient;
 use mz_service::params::GrpcClientParameters;
+use mz_service::transport;
 use tokio::select;
 use tokio::sync::mpsc::error::SendError;
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender, unbounded_channel};
@@ -49,6 +50,7 @@ pub(super) struct ReplicaConfig {
     pub grpc_client: GrpcClientParameters,
     /// The offset to use for replica expiration, if any.
     pub expiration_offset: Option<Duration>,
+    pub enable_ctp: bool,
 }
 
 /// A client for a replica task.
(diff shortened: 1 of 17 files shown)


src/compute-client/src/controller/replica.rs
```diff
@@ -179,36 +183,59 @@ where
...
     async fn connect(&self) -> Client<T> {
-        let try_connect = |retry: RetryState| {
-            let addrs = &self.config.location.ctl_addrs;
-            let dests = addrs
-                .iter()
-                .map(|addr| (addr.clone(), self.metrics.clone()))
-                .collect();
+        let try_connect = async |retry: RetryState| {
             let version = self.build_info.semver_version();
             let client_params = &self.config.grpc_client;
```
teskje, line 188: Reusing the grpc params for CTP connections here. Once the gRPC protocol is removed, we should rename them.

src/compute-client/src/metrics.rs
```diff
@@ -86,7 +87,7 @@ impl ComputeControllerMetrics {
             command_message_bytes_total: metrics_registry.register(metric!(
                 name: "mz_compute_command_message_bytes_total",
                 help: "The total number of bytes sent in compute command messages.",
-                var_labels: ["instance_id", "replica_id", "command_type"],
```
teskje, line 89: Had to remove the command type from the byte count metrics because the CTP metrics don't support it. I think that's fine, we still have the command count metrics.

antiguru (approved):
Looks great; thanks!

teskje:
TFTR!

PR #32900, opened 2025-07-02
Reduce number of timely operators in source

antiguru:
This reduces the number of timely operators a bit in the source pipeline.  Eyeballing, this results in roughly 10% less operators/channels in Timely in an ingestion of 1000 Postgres tables, down to 54k IDs.

I got some tracing info by applying the following patch and then looking through the log output:

```diff

diff --git i/src/storage/src/storage_state.rs w/src/storage/src/storage_state.rs
index d48e301020..282c0fbc92 100644
--- i/src/storage/src/storage_state.rs
+++ w/src/storage/src/storage_state.rs
@@ -105,6 +105,7 @@ use mz_storage_types::sources::IngestionDescription;
 use mz_timely_util::builder_async::PressOnDropButton;
 use mz_txn_wal::operator::TxnsContext;
 use timely::communication::Allocate;
+use timely::logging::TimelyEventBuilder;
 use timely::order::PartialOrder;
 use timely::progress::Timestamp as _;
 use timely::progress::frontier::Antichain;
@@ -242,6 +243,17 @@ impl<'w, A: Allocate> Worker<'w, A> {
             server_maintenance_interval: Duration::ZERO,
         };

+        timely_worker
+            .log_register()
+            .unwrap()
+            .insert::<TimelyEventBuilder, _>("timely", |time, data| {
+                if let Some(data) = data {
+                    data.iter().for_each(|x| println!("LOG1: {:?}", x))
+                } else {
+                    println!("LOG1: Flush {time:?}");
+                }
+            });
+
         // TODO(aljoscha): We might want `async_worker` and `internal_cmd_tx` to
         // be fields of `Worker` instead of `StorageState`, but at least for the
         // command flow sources and sinks need access to that. We can refactor
```

### DIFF
--- a/src/storage/src/healthcheck.rs
+++ b/src/storage/src/healthcheck.rs
@@ -367,8 +367,6 @@ where
     let operator_name = format!("healthcheck({})", healthcheck_worker_id);
     let mut health_op = AsyncOperatorBuilder::new(operator_name, scope.clone());
 
-    // let health = health_stream.enter(scope);
-
     let mut input = health_op.new_disconnected_input(
         &health_stream,
         Exchange::new(move |_| u64::cast_from(chosen_worker_id)),
--- a/src/timely-util/src/operator.rs
+++ b/src/timely-util/src/operator.rs
@@ -956,7 +956,7 @@ where
         I: IntoIterator<Item = StreamCore<G, C>>,
         CB: ContainerBuilder + for<'a> PushInto<C::Item<'a>>,
     {
-        let mut builder = OperatorBuilder::new("Concatenate".to_string(), self.clone());
+        let mut builder = OperatorBuilder::new("ConcatenateFlatten".to_string(), self.clone());
         builder.set_notify(false);
 
         // create new input handles for each input stream.


src/storage/src/healthcheck.rs
```diff
@@ -368,10 +367,10 @@ where
     let operator_name = format!("healthcheck({})", healthcheck_worker_id);
     let mut health_op = AsyncOperatorBuilder::new(operator_name, scope.clone());
 
-    let health = health_stream.enter(scope);
+    // let health = health_stream.enter(scope);
```
martykulma: no longer needed?

src/timely-util/src/operator.rs
```diff
@@ -900,3 +902,85 @@ where
...
+where
+    G: Scope,
+    C: Container + 'static,
+{
+    fn concatenate_flatten<I, CB>(&self, sources: I) -> StreamCore<G, CB::Container>
+    where
+        I: IntoIterator<Item = StreamCore<G, C>>,
+        CB: ContainerBuilder + for<'a> PushInto<C::Item<'a>>,
+    {
+        let mut builder = OperatorBuilder::new("Concatenate".to_string(), self.clone());
```
martykulma: is the name used for debugging?  May want to use `ConcatenateFlatten` to differentiate from the timely `Concatenate`

antiguru: Yup, changed. Thanks for spotting!

PR #32914, opened 2025-07-03
WIP add sni tenant lookups to balancer pgwire

jubrad:
- Adds sni tenant lookups to the balancer pgwire sni will be the initial lookup with fe auth as the fallback.
- unifies the lookup code between http and pgwire balancers
- introduces a cache for this lookup to prevent accidental DOS
- introduces LD flags to control how pgwire tenant lookups are performed
- As a word of caution this was an exercise in using cursor

### Checklist

### DIFF
--- a/src/balancerd/src/dyncfgs.rs
+++ b/src/balancerd/src/dyncfgs.rs
@@ -43,6 +43,34 @@ pub const INJECT_PROXY_PROTOCOL_HEADER_HTTP: Config<bool> = Config::new(
     "Whether to inject tcp proxy protocol headers to downstream http servers.",
 );
 
+/// Whether to enable SNI-based tenant resolution for PgWire connections.
+pub const PGWIRE_SNI_TENANT_RESOLUTION: Config<bool> = Config::new(
+    "balancerd_pgwire_sni_tenant_resolution",
+    true,
+    "Whether to enable SNI-based tenant resolution for PgWire connections.",
+);
+
+/// Whether to enable frontegg-based tenant resolution for PgWire connections.
+pub const PGWIRE_FRONTEGG_TENANT_RESOLUTION: Config<bool> = Config::new(
+    "balancerd_pgwire_frontegg_tenant_resolution",
+    false,
+    "Whether to enable frontegg-based tenant resolution for PgWire connections.",
+);
+
+/// TTL for DNS tenant resolution cache entries.
+pub const TENANT_CACHE_TTL: Config<Duration> = Config::new(
+    "balancerd_tenant_cache_ttl",
+    Duration::from_secs(300), // 5 minutes
+    "TTL for DNS tenant resolution cache entries.",
+);
+
+/// TTL for DNS address resolution cache entries.
+pub const ADDR_CACHE_TTL: Config<Duration> = Config::new(
+    "balancerd_addr_cache_ttl",
+    Duration::from_secs(1), // 1 second for fast rollover
+    "TTL for DNS address resolution cache entries.",
+);
+
 /// Sets the filter to apply to stderr logging.
 pub const LOGGING_FILTER: Config<&str> = Config::new(
     "balancerd_log_filter",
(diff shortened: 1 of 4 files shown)


src/balancerd/src/dyncfgs.rs
```diff
@@ -43,6 +43,34 @@ pub const INJECT_PROXY_PROTOCOL_HEADER_HTTP: Config<bool> = Config::new(
     "Whether to inject tcp proxy protocol headers to downstream http servers.",
 );
 
+/// Whether to enable SNI-based tenant resolution for PgWire connections.
+pub const PGWIRE_SNI_TENANT_RESOLUTION: Config<bool> = Config::new(
```
alex-hunt-materialize, line 47: Is there any situation where we would ever want this to be disabled?

src/balancerd/src/lib.rs
```diff
@@ -599,6 +608,187 @@ impl ServerMetrics {
...
+        resolve_template: &str,
+    ) -> Option<ResolvedAddr> {
+        debug!("attempting SNI-based resolution with servername: {servername}");
+
+        // Construct address using the template
+        let addr = resolve_template.replace("{}", servername);
+        debug!("SNI address: {addr}");
+
+        // Attempt to get tenant from DNS CNAME lookup (cached)
+        let tenant = self.resolve_tenant_from_dns(dns_resolver, &addr).await?;
```
alex-hunt-materialize: Isn't the tenant only used for metrics? We previously just set `None` if it didn't exist.

src/balancerd/src/lib.rs
```diff
@@ -599,6 +608,187 @@ impl ServerMetrics {
...
+        port: u16,
+    ) -> Result<ResolvedAddr, anyhow::Error> {
+        let addr = resolve_template.replace("{}", servername);
+        debug!("https address: {addr}");
+
+        // Attempt to get tenant from DNS CNAME lookup (cached)
+        let tenant = self.resolve_tenant_from_dns(dns_resolver, &addr).await;
+
+        // Do the regular IP lookup with port (cached)
+        let addr_with_port = format!("{addr}:{port}");
```
alex-hunt-materialize: This is the only difference between this function and the one above. Could we just make it take `port: Option<u16>` instead, and combine these?

src/balancerd/src/lib.rs
```diff
@@ -599,6 +608,187 @@ impl ServerMetrics {
...
+            tenant,
+        })
+    }
+
+    /// Finds the tenant of a DNS address using CNAME lookup with caching.
+    /// Returns None if DNS lookup fails or tenant cannot be determined.
+    async fn resolve_tenant_from_dns(&self, resolver: &StubResolver, addr: &str) -> Option<String> {
+        // Check cache first
+        if let Some(cached_tenant) = self.tenant_cache.get(addr).await {
+            debug!("tenant cache hit for {addr}: {cached_tenant:?}");
```
alex-hunt-materialize: Can we add metrics for these cache hits/misses?

src/balancerd/src/lib.rs
```diff
@@ -1385,12 +1601,90 @@ mod tests {
...
+            ("myapp.staging.company.com", "myapp"),
+            // Single component - should return as-is
+            ("singlename", "singlename"),
+        ];
+
+        for (input, expected) in test_cases {
+            let result = match input.split_once('.') {
+                Some((left, _right)) => left,
+                None => input,
+            };
```
alex-hunt-materialize: Should we be calling some function used in the code here? We seem to be testing rust's `split_once` method, which doesn't seem very useful to me.

src/balancerd/src/lib.rs
```diff
@@ -1385,12 +1601,90 @@ mod tests {
...
+            assert_eq!(result, expected, "Failed for input: {}", input);
+        }
+
+        // Test that tenant resolver can be created with default configs
+        use mz_dyncfg::ConfigSet;
+        let mut configs = ConfigSet::default();
+        configs = crate::dyncfgs::all_dyncfgs(configs);
+        let _tenant_resolver = TenantResolver::new(&configs);
```
alex-hunt-materialize: Should we be doing something with these?

src/balancerd/src/lib.rs
```diff
@@ -1385,12 +1601,90 @@ mod tests {
...
+        let tenant_cache_key = "test-tenant-hostname";
+        let test_tenant = Some("test-tenant".to_string());
+
+        tenant_resolver
+            .tenant_cache
+            .insert(tenant_cache_key.to_string(), test_tenant.clone())
+            .await;
+
+        let cached_tenant = tenant_resolver.tenant_cache.get(tenant_cache_key).await;
+        assert_eq!(cached_tenant, Some(test_tenant));
```
alex-hunt-materialize: Shouldn't we be doing actual resolving here, and verifying that it gave us the cached results, rather than just calling `.get` on the caches? We're mostly just testing moka's cache type, which doesn't seem useful to me.

src/balancerd/src/dyncfgs.rs
jubrad: oh, I was thinking just to make it safer to rollout/rollback. Otherwise not really.

alex-hunt-materialize: I guess that's fair.

PR #32935, opened 2025-07-07
Operator size tracking

antiguru:
Include all operators that might have a size in the arrangement size tracking. Not ideal because it could be operators without arrangements, but better than what we have now, and I don't want to rework the operator size tracking at this point.

Best viewed with white space hidden.

Related: https://github.com/MaterializeInc/database-issues/issues/9452

### DIFF
--- a/src/catalog/src/builtin.rs
+++ b/src/catalog/src/builtin.rs
@@ -8613,7 +8613,14 @@ pub static MZ_ARRANGEMENT_SIZES_PER_WORKER: LazyLock<BuiltinView> = LazyLock::ne
             .finish(),
         column_comments: BTreeMap::new(),
         sql: "
-WITH batches_cte AS (
+WITH operators_per_worker_cte AS (
+    SELECT
+        id AS operator_id,
+        worker_id
+    FROM
+        mz_introspection.mz_dataflow_operators_per_worker
+),
+batches_cte AS (
     SELECT
         operator_id,
         worker_id,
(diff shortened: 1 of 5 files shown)


src/catalog/src/builtin.rs
```diff
@@ -8702,24 +8709,38 @@ batcher_allocations_cte AS (
...
-LEFT OUTER JOIN batcher_capacity_cte USING (operator_id, worker_id)
-LEFT OUTER JOIN batcher_allocations_cte USING (operator_id, worker_id)",
+    operator_id, worker_id, records, batches, size, capacity, allocations
+FROM combined
+WHERE
+       records     IS NOT NULL
+    OR batches     IS NOT NULL
+    OR size        IS NOT NULL
+    OR capacity    IS NOT NULL
+    OR allocations IS NOT NULL
```
teskje, line 8664: Are these checks correct? All of these should be `COALESCE`d to 0 above, so they can never be NULL, right?

antiguru: I wanted them to be NULL for undefined values, instead of forcing it to be 0. Fixed, and handled the downstream fallout in a new commit.

kay-kim (commented):
docs part lgtm

src/catalog/src/builtin.rs
```diff
@@ -6528,11 +6498,11 @@ SELECT
...
-    pg_catalog.sum(records) AS records,
-    pg_catalog.sum(batches) AS batches,
-    pg_catalog.sum(size) AS size,
-    pg_catalog.sum(capacity) AS capacity,
-    pg_catalog.sum(allocations) AS allocations
+    SUM(records)::int8 AS records,
+    SUM(batches)::int8 AS batches,
+    SUM(size)::int8 AS size,
+    SUM(capacity)::int8 AS capacity,
+    SUM(allocations)::int8 AS allocations
```
teskje, line 6505: Why `int8` instead of `uint8`?

src/catalog/src/builtin.rs
```diff
@@ -8624,7 +8534,7 @@ batches_cte AS (
     SELECT
         operator_id,
         worker_id,
-        pg_catalog.count(*) AS batches
+        COUNT(*) AS batches
```
teskje, line 8537: Do you know if there was a reason why these specified the schema before?

antiguru: In ye good old times we had to prefix it. It changed at some point, not sure when.

src/catalog/src/builtin.rs
antiguru: Because it's signed types in logging, and I'd like to guard against errors in logging to avoid crashing if we fail to retract something...

teskje: What would be the consequence if we don't guard and cast to `uint8`? Would we just get a query error? If so that seems fine (i.e. not worse than the error you get when trying to read negative multiplicities from the `mz_arrangement_records_raw` index). But this is an internal relation anyway, so probably doesn't matter much.

PR #32939, opened 2025-07-07
Limit overall memory utilization

antiguru:
Limit the memory utilization of a cluster replica process, based on RSS and swap utilization.

This change introduces a memory limiter task that periodically reads `/proc/self/status` and extracts the `VmRSS` and `VmSwap` attributes. It enforces that the sum of the two attributes is less than a configured limit, taking into account an optional burst budget. It has the same logic as the lgalloc limiter, off which it is based.

The first commit just copies the lgalloc limiter. The second changes it and includes the scaffolding to integrate the memory limiter into the cluster replica.

Fixes https://github.com/MaterializeInc/database-issues/issues/9436


src/compute-types/src/dyncfgs.rs
```diff
@@ -83,6 +83,34 @@ pub const LGALLOC_SLOW_CLEAR_BYTES: Config<usize> = Config::new(
...
+);
+
+/// Bias to the memory limiter usage factor.
+pub const MEMORY_LIMITER_USAGE_BIAS: Config<f64> = Config::new(
+    "memory_limiter_usage_bias",
+    1.,
+    "Multiplicative bias to memory_limiter_usage_factor.",
+);
+
+/// Burst factor to disk limit.
```
```suggestion
/// Burst factor to memory limit.
```

src/compute/src/memory_limiter.rs
```diff
@@ -0,0 +1,331 @@
...
+            Err(_err) => ProcStatus {
+                vm_rss: 0,
+                vm_swap: 0,
+            },
+        }
+    }
+
+    /// Perform a memory usage check, terminating the process if the configured limits are exceeded.
+    fn check(&mut self) -> Result<(), anyhow::Error> {
+        debug!("checking limiter limits");
```
teskje:
```suggestion
        debug!("checking memory limits");
```

src/compute/src/memory_limiter.rs
```diff
@@ -225,43 +244,88 @@ impl LimiterTask {
...
+                    * 1024;
+            } else if line.starts_with("VmSwap:") {
+                vm_swap = line
+                    .split_whitespace()
+                    .nth(1)
+                    .and_then(|s| s.parse().ok())
+                    .unwrap_or(0)
+                    * 1024;
+            }
+        }
```
teskje, line 330: I think we should make _some_ noise if we encounter something unexpected here:

* `VmRSS` missing (I think `VmSwap` is only present if swap is enabled)
* a parse error in the KiB values

I'd probably just return an error from this function and then log an error in the caller. Alternatively, we could continue returning 0 and only log something here.

teskje (commented):
Looks good! Thanks for splitting into two commits :)

I think we should have some test in CI to ensure that this works with swap as expected. It's maybe a bit too early to switch our entire CI from lgalloc to swap, but maybe we can come up with a single targeted test?

src/compute/src/memory_limiter.rs
antiguru: I think it now better surfaces errors. It still doesn't warn if the values aren't found, but it seems both `VmRSS` and `VmSwap` have been part of Linux since approx. forever, so I don't think it makes much sense to protect against them missing.

antiguru:
This is good for another look! Thank you

test/cluster/mzcompose.py
```diff
@@ -5745,3 +5745,119 @@ def setup_workload():
         c.testdrive("> SELECT count(*) FROM mv\n1000000")
 
         c.kill("clusterd1")
+
+
+def workflow_test_memory_limiter(c: Composition) -> None:
+    """
+    Test that the lgalloc disk usage limiter functions as expected.
```
teskje:
```suggestion
    Test that the memory limiter functions as expected.
```

test/cluster/mzcompose.py
```diff
@@ -5745,3 +5745,119 @@ def setup_workload():
...
+
+    We run a workload whose disk usage is roughly known and then assert that it
```
teskje:
```suggestion
    We run a workload whose memory usage is roughly known and then assert that it
```

test/cluster/mzcompose.py
```diff
@@ -5745,3 +5745,119 @@ def setup_workload():
...
+    does, or does not, manage to hydrate with various limiter configurations.
+    """
+
+    c.down(destroy_volumes=True)
+
+    with c.override(
+        Materialized(
+            additional_system_parameter_defaults={
+                "enable_compute_correction_v2": "true",
```
teskje: We don't need this one! Doesn't hurt to have it, but might raise questions.

test/cluster/mzcompose.py
```diff
@@ -5745,3 +5745,119 @@ def setup_workload():
...
+                "enable_lgalloc": "false",
+                "memory_limiter_interval": "100ms",
+                "unsafe_enable_unorchestrated_cluster_replicas": "true",
+            },
+        ),
+        Clusterd(
+            name="clusterd1",
+            # Announce an (unenforced) memory limit of 1GiB. Memory limits are
+            # derived from this memory limit.
+            options=["--announce-memory-limit=1073741824"],
```
teskje, line 5772: The ambiguity of the term "memory limit" is annoying. It would be nice to have a word for "memory usage including swap", but I'm not sure that exists.

test/cluster/mzcompose.py
```diff
@@ -5745,3 +5745,119 @@ def setup_workload():
...
+                    WORKERS 1
+                )
+            )
+            """
+        )
+
+        def setup_workload():
+            """
+            For our workload we use a large MV, which we obtain by performing a cross
+            join. We make sure that the rows are large, so they spill to disk well.
```
teskje: The "We make sure that the rows are large, so they spill to disk well" part is also not necessary here

test/cluster/mzcompose.py
```diff
@@ -5745,3 +5745,119 @@ def setup_workload():
...
+                DROP TABLE IF EXISTS t CASCADE;
+                CREATE TABLE t (i int, x text);
+                INSERT INTO t
+                    SELECT generate_series, repeat('a', 100) FROM generate_series(1, 1000);
+                CREATE MATERIALIZED VIEW mv IN CLUSTER test AS
+                    SELECT t1.i i1, t1.x x1, t2.i i2, t2.x x2 FROM t t1, t t2;
+                """
+            )
+
+        # Test 1: The MV should be able to hydrate with a disk limit of 1 GiB.
```
teskje:
```suggestion
        # Test 1: The MV should be able to hydrate with a memory limit of 1 GiB.
```

test/cluster/mzcompose.py
```diff
@@ -5745,3 +5745,119 @@ def setup_workload():
...
+            user="mz_system",
+        )
+        setup_workload()
+        c.up("clusterd1")
+
+        c.testdrive("> SELECT count(*) FROM mv\n1000000")
+
+        c.kill("clusterd1")
+
+        # Test 2: The MV should be unable to hydrate with a disk limit of 10 MiB.
```
teskje:
```suggestion
        # Test 2: The MV should be unable to hydrate with a memory limit of 10 MiB.
```

test/cluster/mzcompose.py
```diff
@@ -5745,3 +5745,119 @@ def setup_workload():
...
+
+        for _ in range(100):
+            time.sleep(1)
+            ps = c.invoke("ps", "clusterd1", "-a", capture=True, silent=True).stdout
+            if "Exited (167)" in ps:
+                break
+        else:
+            raise RuntimeError("replica did not exit with code 167")
+
+        # Test 3: The MV should be able to hydrate with a disk limit of 10 MiB
```
teskje:
```suggestion
        # Test 3: The MV should be able to hydrate with a memory limit of 10 MiB
```

teskje (approved):
Looks good! Just a couple copy-paste-os left in the test.

PR #32944, opened 2025-07-07
expr: also encode MSE column names when serializing with protobuf

teskje:
This PR makes the `RustType for MirScalarExpression` also serialize column names, instead of ignoring them. Having column names show up in dataflow operators can be useful, and the fact that protobuf serialization doesn't roundtrip all MSE values is surprising. Also, in the future we would like to rely on derived serialization logic, rather than a hand-rolled one, and this special case makes that harder.

The bulk of the changes is adjusting the tests. For SLTs, we just use the automatic rewriting. For testdrive that isn't available so we want to be a bit more careful to not make the testdrive output too susceptible to optimizer changes. In particular, wherever we match on the name of a dataflow-internal arrangement, we want to replace the column name because it may change if the optimizer's column name derivation logic changes. If we are matching on the name of an index export arrangement, using the plain column name is fine because it is unlikely to change as long as the index definition remains unchanged.

### Checklist


teskje:
TFTR!

PR #32947, opened 2025-07-07
[design] Design doc for Persist schema migrations

bkirwi:
### Checklist


bkirwi:
@teskje, @petrosagg - tagging you two as reviewers since IIUC you have looked at either migrations or schema changes recently, and it'd be good to have the design space settled one way or the other.

doc/developer/design/20250707_persist_schemas.md
```diff
@@ -0,0 +1,118 @@
...
+
+A Persist shard maintains an ordered sequence of schemas. This sequence is appended to with the `PersistClient::compare_and_evolve_schema` method. If the sequence is empty, this sets the initial schema of the shard. If the sequence is non-empty, this will check that the schema obeys the schema-migration rules; if it does, it will add the given schema as the latest schema of the shard. A monotonically-increasing `SchemaId` identifies a particular schema in this sequence.
```
teskje: > If the sequence is empty, this sets the initial schema of the shard.

Does this match the current implementation? `compare_and_evolve_schema` takes a `SchemaId` to compare the current schema to, so it seems like if there is no schema, it would be impossible to provide that `SchemaId`.

doc/developer/design/20250707_persist_schemas.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Tracking deleted fields
+
+To check whether it’s allowed to migrate a shard to a particular datatype, it’s not enough just to check the current datatype: we need to ensure that any added fields have never been used before in *any* of the previous schemas. IDLs like Protobuf uses “reserved fields” to track this information as part of the current schema, but Arrow datatypes have no similar concept built in.
+
+Internally, we define a new “migration type” that *does* track deleted fields: identical to an arrow datatype, but where struct fields include whether or not a field has been deleted. We can recover the current datatype by filtering out the deleted fields.
+
+### Reader schemas
+
+Backward compatibility mean that readers using any valid schema should be able to read data that was written using any previous schema. However, we do not enforce forward compatibility… so we may apply a schema change that invalidates existing readers. In that case, the reader should be “fenced”: exit with an error instead of passing along data with an unexpected shape.
```
teskje, line 58: This is not how things currently work, right? I think the 0dt upgrade logic relies on readers not getting fenced as long as no new data is written with the new schema.

* MV nullability evolution in the read-only env may make a column nullable that wasn't before, which would fence out the readers of that MV in the leader env.
* Builtin item migration, as of very recently, may perform any backward-compatible schema evolution, according ones that are not forward-compatible.

doc/developer/design/20250707_persist_schemas.md
bkirwi: The `register_schema` method (used internally / implicitly by writers) handles the other case - it registers the given schema if there aren't any existing schemas.

So: it does not match, but it has the ~same effect. Overall the design in this section is "cleaner" and more general than what's implemented (re. the section header) but I can see that being confusing when I'm eg. calling out specific methods... I can either add the caveat here or make this less specific.

doc/developer/design/20250707_persist_schemas.md
bkirwi: > This is not how things currently work, right?

Right! This is discussed a bit further down: at the moment, the only migration we support is both backwards and forwards compatible, so the caveat in this section doesn't apply. In the future, when we _do_ support a richer set of migrations, we'd have to start worrying about fencing. (Probably by shifting the schema assertion in the writer to write time instead of creation time, so we don't have to apply schema migrations in the read-only replica, which has other benefits.)

> MV nullability evolution in the read-only env may make a column nullable that wasn't before, which would fence out the readers of that MV in the leader env.

Nonobvious: at the Arrow DataType level we mark all the columns in an MV or similar nullable, because we know that even if they're not known nullable now they might be in the future.

> Builtin item migration, as of very recently, may perform any backward-compatible schema evolution, according ones that are not forward-compatible.

Since the only supported migration (adding a nullable field to the end of the column list) is also forward-compatible, this isn't a risk currently in practice.

doc/developer/design/20250707_persist_schemas.md
teskje: I'd prefer the design specifying the desired end state! Maybe just add a note in the "Minimal Viable Prototype" section below, since that talks about what's currently implemented?

doc/developer/design/20250707_persist_schemas.md
teskje: Makes sense, thanks!

> Nonobvious: at the Arrow DataType level we mark all the columns in an MV or similar nullable, because we know that even if they're not known nullable now they might be in the future.

It feels like we should note this down somewhere in this design doc, since it comes up from time to time as a source of confusion. I realize that it's an implementation detail of the `RelationDesc` schema and doesn't quite fit into a high-level design doc, but I can't think of a better place to put it where people that need to know about it will likely find it.

teskje (approved):
Thanks, very helpful!

PR #32949, opened 2025-07-08
ci: Try ci-builder min

def-:
### Checklist


.bazelrc
```diff
@@ -186,7 +186,7 @@ build:release-tagged --config=release --config=release-lto --config=release-stam
 # PRs in CI.
 #
 # Not doing a full stamp nor omitting full debug info, greatly speeds up compile times.
-build:release-dev --config=release --config=release-lto --config=debuginfo-limited
```
antiguru: One reason to keep release-lto here is to enable performance measurements on release-dev builds.

def-: I added a CI_RELEASE_LTO_BUILD flag and will expose it in trigger-ci

antiguru: Thank you!

def-:
Sorry for another huge PR, this should get us to 15-20 min test time, 20-28 min with builds (assuming no ongoing Buildkite and Hetzner incidents).

ptravers:
@def- do you mind adding a description to the PR?

ptravers:
@def- is it worth doing a full nightly run when you're ready?

misc/python/materialize/mzbuild.py
```diff
@@ -205,11 +211,61 @@ def docker_images() -> set[str]:
...
+    #                 "Accept": "application/vnd.docker.distribution.manifest.v2+json",
+    #                 "Authorization": f"Bearer {token}",
+    #             },
+    #         )
+    #
+    #     return response.status_code == 200
+    #
+    # except Exception as e:
+    #     print(f"Error checking Docker image: {e}")
+    #     return False
```
ptravers: can we put the change you're recommending in an issue instead?

ptravers:
afaict, this seems fine but slightly worried about disabling a test. I don't have much context on what other correctness guarantees we have for that code path.

misc/python/materialize/mzbuild.py
def-: The incident is resolved, I have reenabled it for now, will check if it works reliably now.

def-:
No tests were altogether disabled, some slower and rarely failing test cases were moved to Nightly. If this turns out to not be wise, we can move them back to Test pipeline.

def-:
> @def- is it worth doing a full nightly run when you're ready?

Absolutely, nightly and release qualification: https://buildkite.com/materialize/nightly/builds/12581 & https://buildkite.com/materialize/release-qualification/builds/890

ci/test/pipeline.template.yml
```diff
@@ -547,7 +399,7 @@ steps:
           - ./ci/plugins/mzcompose:
               composition: kafka-auth
         agents:
-          queue: hetzner-aarch64-4cpu-8gb
+          queue: hetzner-aarch64-8cpu-16gb
```
ptravers, line 402: do we need to be concerned about increasing CI spend with the size bump? or will the size bump net out at enough decreased usage that we should be okay?

ptravers: thought further, so long as we have something tracking CI spend. this seems fine to me. we can always rollback if our bill goes up?

ptravers (approved):

ptravers:
assuming the nightly and release qualification runs look okay.

ci/test/pipeline.template.yml
def-: I'm tracking costs in https://docs.google.com/spreadsheets/d/1QEPoy1poVPtwGwFj7mH_5E6oTl8d-ygEtLBStfRd0oU/edit?gid=780614635#gid=780614635
This PR definitely makes CI more expensive, but that's a tradeoff we are willing to make.

PR #32958, opened 2025-07-08
Omit INSERT statements after 100 bytes

SangJunBak:
Things I'm unsure about:
- Should the tests live there? Is there a simpler function I can reuse?
- Rn we just add a ' <Omitted>', including the space to not just cut it off so abruptly. Do we think this is fine? Another thing we could do is '...<OMITTED>'
- INSERTs seem like the biggest culprit but someone could potentially pollute the statement log with `UPDATE`s too. Should we include those statements or just stick with INSERTs?

### Motivation
Aids in https://github.com/MaterializeInc/database-issues/issues/9417

### Checklist


src/adapter/src/coord/statement_logging.rs
```diff
@@ -104,6 +107,27 @@ impl PreparedStatementLoggingInfo {
...
+            .map(|(i, _)| i)
+            .last()
+            .unwrap_or(0);
+
+        let omitted_str = " <OMITTED>";
+
+        let mut result = String::with_capacity(byte_limit + omitted_str.len());
+        result.push_str(&s[..(cutoff)]);
+        result.push_str(omitted_str);
+        result
```
antiguru:
```suggestion
        format!("{} <OMITTED>", &s[..cutoff])
```

src/adapter/src/coord/statement_logging.rs
```diff
@@ -104,6 +107,27 @@ impl PreparedStatementLoggingInfo {
...
         }
     }
+
+    fn insert_omitted_at_byte_limit(s: &str, byte_limit: usize) -> String {
+        if s.len() <= byte_limit {
+            return s.to_string();
+        }
+
+        // Find the character boundary just before or at `byte_limit`
+        let cutoff = s
```
antiguru: Too bad the API for this isn't stable yet: https://doc.rust-lang.org/std/primitive.str.html#method.floor_char_boundary

src/adapter/src/coord/statement_logging.rs
```diff
@@ -104,6 +107,27 @@ impl PreparedStatementLoggingInfo {
...
+            .char_indices()
+            .take_while(|(i, _)| *i <= byte_limit)
```
antiguru: This will give you up to three bytes more than `byte_limit` because the last char can start at the last position in `byte_limit`, but then continue across the boundary. Not sure if it matters, just pointing it out.

src/adapter/src/coord/statement_logging.rs
```diff
@@ -90,6 +90,9 @@ impl PreparedStatementLoggingInfo {
             Some(StatementKind::CreateSecret | StatementKind::AlterSecret) => {
                 stmt.map(|s| s.to_ast_string_redacted()).unwrap_or_default()
             }
+            // INSERT statements can contain a lot of data and is proportional to the
+            // amount of data being inserted, so at a byte limit, we omit the rest of the statement.
+            Some(StatementKind::Insert) => Self::insert_omitted_at_byte_limit(&raw_sql, 100),
```
antiguru: We should make the limit configurable.

ggevay:
> where did we land on the discussion whether we should redact the insert statements?

We record both the redacted and unredacted SQL, and surface them in different builtin views with different ACLs. (In more detail: `mz_sql_text` is the builtin source that has both, and then `mz_sql_text_redacted`, `mz_recent_sql_text_redacted`, `mz_recent_activity_log_redacted` selects only the redacted sql.)

When filling the `redacted_sql` column in the builtin source, [we call `to_ast_string_redacted` on the original statement](https://github.com/MaterializeInc/materialize/blob/e477cd55a4377437734e2a84d1daaf40bd6c3617/src/adapter/src/coord/statement_logging.rs#L98). It turns out that fortunately `to_ast_string_redacted` [already truncates after 20 rows](https://github.com/MaterializeInc/materialize/blob/b2f751929db0da73c87d692e066f9d87730b8ad8/src/sql-parser/src/ast/defs/query.rs#L820-L825).

aljoscha:
> INSERTs seem like the biggest culprit but someone could potentially pollute the statement log with UPDATEs too. Should we include those statements or just stick with INSERTs?

Yep, I'd say we want the same treatment for UPSERT!

src/adapter/src/coord/statement_logging.rs
SangJunBak: This is true, however since we're doing `[..cutoff]` and doing a slice upto that last character (exclusive), aren't we ensuring the boundary is still never crossed?

SangJunBak:
@antiguru Decided we should just redact `INSERT` and these types of statements entirely instead of some custom omit / truncation https://materializeinc.slack.com/archives/C08A62E0751/p1752076534058579?thread_ts=1751920625.811429&cid=C08A62E0751

Addressed in this commit https://github.com/MaterializeInc/materialize/pull/32958/commits/534733d0cdda5c31eb138f814f28738c3c79cf52

ggevay (approved):
LGTM! Thank you for iterating on this!

The test failure can be addressed by adding INSERT, UPDATE, EXECUTE to the if [here](https://github.com/MaterializeInc/materialize/blob/c6b942c7ecd1055bc566eeb8ec992c26bd4e48b4/src/environmentd/tests/server.rs#L309).

PR #32969, opened 2025-07-09
Introduce EQ Class that does not allow errors

DAlperin:
If EQ Classes just swallow errors, you can end up with unfortunate situations like this https://github.com/MaterializeInc/database-issues/issues/5511. This introduces an EQ class wrapper that does not allow errors to be inserted, as well as prevents eq classes from introducing erroring exprs.

### Checklist

### DIFF
--- a/ci/plugins/mzcompose/hooks/command
+++ b/ci/plugins/mzcompose/hooks/command
@@ -155,21 +155,21 @@ cleanup() {
   find cores -name 'core.*' | while read -r core; do
       exe=$(echo "$core" | sed -e "s/core\.\(.*\)\.[0-9]*/\1/" -e "s/.*\!//")
       # Core dumps can take a while to be written, so if extracting the info fails, try again later
-      bin/ci-builder run min gdb --batch -ex "bt full" -ex "thread apply all bt" -ex "quit" cores/"$exe" "$core" > "$core".txt || (sleep 2m; bin/ci-builder run min gdb --batch -ex "bt full" -ex "thread apply all bt" -ex "quit" cores/"$exe" "$core" > "$core".txt || true)
+      bin/ci-builder run "$builder" gdb --batch -ex "bt full" -ex "thread apply all bt" -ex "quit" cores/"$exe" "$core" > "$core".txt || (sleep 2m; bin/ci-builder run "$builder" gdb --batch -ex "bt full" -ex "thread apply all bt" -ex "quit" cores/"$exe" "$core" > "$core".txt || true)
       buildkite-agent artifact upload "$core".txt
   done
   # can be huge, clean up
   rm -rf cores
 
   echo "Compressing parallel-workload-queries.log"
-  bin/ci-builder run min zstd --rm parallel-workload-queries.log || true
+  bin/ci-builder run "$builder" zstd --rm parallel-workload-queries.log || true
 
   mapfile -t artifacts < <(printf "run.log\nservices.log\njournalctl-merge.log\nnetstat-ant.log\nnetstat-panelot.log\nps-aux.log\ndocker-inspect.log\n"; find . -name 'junit_*.xml'; find mz_debug_* -name '*.log')
   artifacts_str=$(IFS=";"; echo "${artifacts[*]}")
 
   echo "--- Running trufflehog to scan artifacts for secrets & uploading artifacts"
   {
-    bin/ci-builder run min trufflehog --no-update --no-verification --json --exclude-detectors=coda,dockerhub,box,npmtoken,github,snykkey,eightxeight,sumologickey,miro filesystem "${artifacts[@]}" | trufflehog_jq_filter_logs > trufflehog.log
+    bin/ci-builder run "$builder" trufflehog --no-update --no-verification --json --exclude-detectors=coda,dockerhub,box,npmtoken,github,snykkey,eightxeight,sumologickey,miro filesystem "${artifacts[@]}" | trufflehog_jq_filter_logs > trufflehog.log
   } &
 
   unset CI_EXTRA_ARGS # We don't want extra args for the annotation
@@ -181,7 +181,7 @@ cleanup() {
   } &
   wait
   echo "--- Annotating errors"
-  bin/ci-builder run min bin/ci-annotate-errors --test-cmd="$TEST_CMD" --test-desc="$TEST_DESC" --test-result="$TEST_RESULT" "${artifacts[@]}" trufflehog.log > ci-annotate-errors.log || CI_ANNOTATE_ERRORS_RESULT=$?
+  bin/ci-builder run "$builder" bin/ci-annotate-errors --test-cmd="$TEST_CMD" --test-desc="$TEST_DESC" --test-result="$TEST_RESULT" "${artifacts[@]}" trufflehog.log > ci-annotate-errors.log || CI_ANNOTATE_ERRORS_RESULT=$?
   buildkite-agent artifact upload "ci-annotate-errors.log"
 
   export_cov() {
--- a/ci/publish-helm-charts/pipeline.template.yml
+++ b/ci/publish-helm-charts/pipeline.template.yml
@@ -29,6 +29,7 @@ steps:
           composition: terraform
           run: aws-temporary
           args: ["--tag=$CI_MZ_VERSION", --no-run-testdrive-files]
+          ci-builder: stable
 
   - id: terraform-gcp-tag
     label: "Terraform + Helm Chart E2E on GCP (tagged)"
@@ -43,6 +44,7 @@ steps:
           composition: terraform
           run: gcp-temporary
           args: ["--tag=$CI_MZ_VERSION", --no-run-testdrive-files]
+          ci-builder: stable
 
   - id: terraform-azure-tag
     label: "Terraform + Helm Chart E2E on Azure (tagged)"
@@ -57,6 +59,7 @@ steps:
           composition: terraform
           run: azure-temporary
           args: ["--tag=$CI_MZ_VERSION", --no-run-testdrive-files]
+          ci-builder: stable
 
   - wait: ~
 
--- a/ci/test/pipeline.template.yml
+++ b/ci/test/pipeline.template.yml
@@ -84,10 +84,12 @@ steps:
         sanitizer: skip
 
       - id: build-wasm
-        label: Build WASM
+        label: ":rust: Build WASM"
         command: bin/ci-builder run stable bin/pyactivate -m ci.deploy.npm --no-release
         inputs:
-          - "*"
+          - "ci/deploy/npm.py"
+          - "bin/wasm-build"
+          - "misc/wasm"
         depends_on: []
         timeout_in_minutes: 20
         agents:
--- a/doc/developer/tracing.md
+++ b/doc/developer/tracing.md
@@ -61,6 +61,16 @@ Like all system variables, it is also possible to set a log filter default when 
 Note that this value is passed through to all compute and storage instances when running locally.
 In the future, it is possible to add per-service filtering, if deemed useful.
 
+### Log filtering in tests
+
+When running environmentd tests (i.e `bin/cargo-test environmentd`), the test harness is used to initialize environmentd. To configure log filtering in tests, you'll need to override the `--system-parameter-default` when initializing the test harness. For example:
+```rs
+let server = test_util::TestHarness::default()
+  .with_system_parameter_default("log_filter".to_string(), "my_interesting_module::sub_module=trace,error".to_string())
+  .start()
+  .await;
+```
+
 ## Spans
 
 **Please see [the tracing docs](https://docs.rs/tracing/latest/tracing/index.html#spans) for more information about spans**
--- a/misc/python/materialize/sqlsmith.py
+++ b/misc/python/materialize/sqlsmith.py
@@ -135,4 +135,5 @@ known_errors = [
     "exceeded recursion limit of 2048",
     "key cannot be null",  # expected, see PR materialize#25941
     "regexp_extract must specify at least one capture group",
+    "array_fill with arrays not yet supported",
 ]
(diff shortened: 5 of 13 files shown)


ggevay:
Kicked off a subset of Nightly that is interesting for the optimizer: https://buildkite.com/materialize/nightly/builds/12561

Edit: RQG has unrelated failures. Looking into it. Edit 2: https://materializeinc.slack.com/archives/C01LKF361MZ/p1752083364128639

src/transform/src/analysis/equivalences.rs
```diff
@@ -339,6 +341,111 @@ pub struct EquivalenceClasses {
     remap: BTreeMap<MirScalarExpr, MirScalarExpr>,
 }
 
+/// A wrapper struct for equivalence classes that can not contain any errors.
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Default, Debug)]
+pub struct EquivalenceClassesWithoutErrors {
+    /// The underlying equivalence classes.
+    pub equivalence_classes: EquivalenceClasses,
+    /// A map of errors that have been encountered related to their equivalence classes.
+    // pub errors: BTreeMap<MirScalarExpr, MirScalarExpr>,
```
ggevay: I guess the commented out code is not needed.

Also, the comment is outdated.

src/transform/src/equivalence_propagation.rs
```diff
@@ -491,23 +497,36 @@ impl EquivalencePropagation {
...
 
-                equivalences.clone_from(&join_equivalences.classes);
+                let extracted_equivalences =
+                    join_equivalences.extract_equivalences(input_types.as_ref().map(|x| &x[..]));
+
+                info!(
+                    ?inputs,
+                    ?extracted_equivalences,
+                    "Join equivalences extracted"
+                );
```
ggevay, line 538: This should be at `debug!` level.

src/transform/src/equivalence_propagation.rs
```diff
@@ -491,23 +497,36 @@ impl EquivalencePropagation {
...
+                equivalences.clone_from(
+                    &join_equivalences.extract_equivalences(input_types.as_ref().map(|x| &x[..])),
```
ggevay: Could this just refer to `extracted_equivalences`?

src/transform/src/analysis/equivalences.rs
```diff
@@ -339,6 +341,111 @@ pub struct EquivalenceClasses {
     remap: BTreeMap<MirScalarExpr, MirScalarExpr>,
 }
 
+/// A wrapper struct for equivalence classes that can not contain any errors.
```
ggevay: It would be good to add a bit more explanation on how errors are handled, i.e., that they are saved separately, and `extract_equivalences` returns also them.

ggevay: Also, it might be a bit confusing that the original `EquivalenceClasses` is also "without errors", in that `tidy` throws out errors. So, compared to `EquivalenceClasses`, this new struct is actually retaining errors more. So, we might be able to come up with a better name.

src/transform/src/analysis/equivalences.rs
```diff
@@ -339,6 +341,111 @@ pub struct EquivalenceClasses {
...
+        let without_errs = equivalences
+            .iter()
+            .filter(|e| !e.contains_err())
+            .cloned()
+            .collect::<Vec<_>>();
+        if let Some(expr) = without_errs.first()
+            && !with_errs.is_empty()
+        {
+            with_errs.push(expr.clone());
+            info!("Holding back equivalence class with errors {:?}", with_errs);
```
ggevay: `debug` level

src/transform/src/analysis/equivalences.rs
```diff
@@ -339,6 +341,111 @@ pub struct EquivalenceClasses {
...
+            let reducer = self.equivalence_classes.reducer();
+            for class in self.held_back_classes.iter_mut() {
+                class.iter_mut().for_each(|e| {
+                    reducer.reduce_expr(e);
+                });
+            }
+            let mut classes = self.equivalence_classes.classes.clone();
+            // info!("before held back classes: {:?}", classes);
+            classes.extend(self.held_back_classes.iter().cloned());
+            canonicalize_equivalence_classes(&mut classes);
```
ggevay, line 487: `canonicalize_equivalence_classes` is legacy code, so we'd like to avoid calling it if possible. Among other problems, it's quadratic in the number of classes.

`minimize` superseded this, so the non-error classes should be already fine even without this call. And for the error classes, we might want to just accept not minimizing them.

ggevay: Oh but then the problem would be that the `reduce_expr` call above would possibly make non-canonical classes, where e.g. the same expression is in two classes, so we can't simply omit this call.

ggevay: Maybe calling `refresh` would be safe even when there are errors? `refresh` has some smarts (UnionFind) to do the work of `canonicalize_equivalence_classes` very quickly.

ggevay: ~Or just also remove the `reduce_expr`.~ Edit: That's not a good fix, because putting together the held back and the normal classes might also make a non-canonical situation, not just `reduce_expr`. (either due to reductions in the non-error part or due to `push_equivalences` having split up a class that has multiple non-error exprs and some error exprs)

src/transform/src/analysis/equivalences.rs
```diff
@@ -696,7 +803,12 @@ impl EquivalenceClasses {
             for expr in class.iter_mut() {
                 self.remap.reduce_child(expr);
                 if let Some(columns) = columns {
+                    let orig_expr = expr.clone();
                     expr.reduce(columns);
+                    if expr.contains_err() {
+                        // Rollback to the original expression if it contains an error.
+                        *expr = orig_expr;
+                    }
```
ggevay, line 855: Could you explain the reasoning behind this change a bit? Was it needed to fix the bug on the repro? Or are we just worried in theory that `tidy` could make us lose an error after a `reduce` here reveals an error?

src/transform/src/equivalence_propagation.rs
```diff
@@ -481,7 +485,9 @@ impl EquivalencePropagation {
                         col.nullable = true;
                     }
                 }
-                join_equivalences.minimize(input_types.as_ref().map(|x| &x[..]));
+                join_equivalences
+                    .equivalence_classes
+                    .minimize(input_types.as_ref().map(|x| &x[..]));
```
ggevay, line 502: `extract_equivalences` also calls `minimize`. Is it intentional to also do it there?

src/transform/src/analysis/equivalences.rs
```diff
@@ -339,6 +341,111 @@ pub struct EquivalenceClasses {
...
+            .cloned()
+            .collect::<Vec<_>>();
+        // info!("Equivalence class with errors: {:?}", with_errs);
+        let without_errs = equivalences
+            .iter()
+            .filter(|e| !e.contains_err())
+            .cloned()
+            .collect::<Vec<_>>();
+        if let Some(expr) = without_errs.first()
+            && !with_errs.is_empty()
```
ggevay: Would this throw out errors when `without_errs` is empty? I'm not sure that's ok, because maybe there can be a join equivalence where both sides are errors, and we wouldn't want to lose them. Maybe just the
```
self.held_back_classes.push(with_errs);
```
needs to be moved out of the if block to afterwards the if block?

src/transform/src/analysis/equivalences.rs
```diff
@@ -339,6 +341,111 @@ pub struct EquivalenceClasses {
...
+    /// Create a new instance of `EquivalenceClassesWithoutErrors` with the given equivalence classes.
+    pub fn new(equivalence_classes: EquivalenceClasses) -> Self {
+        Self {
+            equivalence_classes,
+            held_back_classes: Vec::new(),
+        }
+    }
+
+    /// Push some new predicate that we expect to be true
+    pub fn push_predicate(&mut self, predicate: MirScalarExpr) {
```
ggevay: Is this function used?

src/transform/src/analysis/equivalences.rs
```diff
@@ -339,6 +341,111 @@ pub struct EquivalenceClasses {
...
+        }
+    }
+
+    /// Subject the constraints to the column projection, reworking and removing equivalences.
+    /// This method should also introduce equivalences representing any repeated columns.
+    pub fn project<I>(&mut self, output_columns: I)
+    where
+        I: IntoIterator<Item = usize> + Clone,
+    {
+        self.equivalence_classes.project(output_columns.clone());
```
ggevay, line 464: `held_back_classes` can also have column references, because `push_equivalences` puts in one non-erroring expr as well. I think this is not causing an actual bug here, because the code that calls this `project` function only refers to `equivalence_classes` afterwards. But it would be good to either document this limitation in the doc comment, or maybe just also attend to `held_back_classes`.

src/transform/src/analysis/equivalences.rs
```diff
@@ -339,6 +341,111 @@ pub struct EquivalenceClasses {
...
+            .iter()
+            .filter(|e| e.contains_err())
+            .cloned()
+            .collect::<Vec<_>>();
+        // info!("Equivalence class with errors: {:?}", with_errs);
+        let without_errs = equivalences
+            .iter()
+            .filter(|e| !e.contains_err())
+            .cloned()
+            .collect::<Vec<_>>();
```
ggevay: You could
```
let (mut with_errs, without_errs): (Vec<_>, Vec<_>) =
   equivalences.into_iter().partition(|e| e.contains_err());
```

src/transform/src/analysis/equivalences.rs
```diff
@@ -339,6 +341,111 @@ pub struct EquivalenceClasses {
...
+    /// The underlying equivalence classes.
+    pub equivalence_classes: EquivalenceClasses,
+    /// A map of errors that have been encountered related to their equivalence classes.
+    // pub errors: BTreeMap<MirScalarExpr, MirScalarExpr>,
+    pub held_back_classes: Vec<Vec<MirScalarExpr>>,
+}
+
+impl EquivalenceClassesWithoutErrors {
+    /// Create a new instance of `EquivalenceClassesWithoutErrors` with the given equivalence classes.
+    pub fn new(equivalence_classes: EquivalenceClasses) -> Self {
```
ggevay: Is this function used?

ggevay (commented):
Thanks for picking this up!

The approach might work, but I have a big open question around the `canonicalize_equivalence_classes` call, and some other minor suggestions/questions.

Also, it would be good to add the repro in an slt.

src/transform/src/analysis/equivalences.rs
DAlperin: Yeah, it was required to fix the bug in the repo. Otherwise the expr would get reduced into an error and then immediately swallowed.

src/transform/src/analysis/equivalences.rs
DAlperin: Yeah it's not clear to me what a better option is here.

src/transform/src/equivalence_propagation.rs
DAlperin: Don't we want the minimized version below when revisiting the children?

ggevay: Indeed, ok!

src/transform/src/analysis/equivalences.rs
ggevay: Ok, thx!

src/transform/src/analysis/equivalences.rs
ggevay: Yeah, it's painful, because we can't just call `refresh`, because that would `tidy` away the error again.
I would recommend stealing some part of `refresh`, where you just take that part of it that is directly relevant for ensuring the invariant that no two classes have the same expr. This would be just running the Union-Find, I think. But it's probably best to also ask @frankmcsherry what he would do here.

(Btw. my original plan before this PR would have been to give up on minimizing Join's equivalence classes, and just do the bare minimum on it: not put it into an `EquivalenceClasses`, but just run the Union-Find part of `refresh`. It's interesting that we've somehow arrived at working with Union-Find code also with the PR's alternative approach.)

ggevay:
I've triggered a new Nightly: https://buildkite.com/materialize/nightly/builds/12586

ggevay:
The test failure in `column_knowledge.spec` just needs an auto-rewrite. You can go into the file's dir, and then
```
REWRITE=1 cargo test
```

ggevay:
Could you please rebase the PR? A test issue was fixed on `main`, which is blocking RQG here.

ggevay:
Thanks! Triggered RQG: https://buildkite.com/materialize/nightly/builds/12587

ggevay:
Ah, it seems there was another Nightly issue on main at the time of the rebase. But anyhow, other tests have passed, and the code seems fine!

One last thing is that I think it would be good to guard this behind a feature flag, because changing join equivalences can potentially change join orders, which can have a huge impact on performance. The flag can be turned on by default. (It's just a precaution: if there is a newly OOMing cluster during the release rollout, then we'll try turning the flag off.) You can add the flag to `feature_flags!` and `optimizer_feature_flags!`, and then follow the compile errors. There is also a recent example here: https://github.com/MaterializeInc/materialize/pull/32978/commits/97802cf2cc379c8f5115d80b27516be4650b1039. `enable_for_item_parsing` should be false.

DAlperin:
Added feature flags!

ggevay:
Thank you for adding the flag!

The new spec test failure is because currently the test framework doesn't pick up the default flag setting. You can add the flag like how some other flags are already added [here](https://github.com/MaterializeInc/materialize/blob/3bab0b62a2239b17ebb7334176573af3d34c96ca/src/transform/tests/test_transforms.rs#L265-L266). It's on my todo list to fix this, because it's really annoying, sorry about it.

(I'm not sure what happened in the Cluster tests 1 fail.)

DAlperin:
No worries! And hopefully the cluster test failure is just a flake.

DAlperin:
Looks like the flake is fixed in main @ggevay should we merge this or pull main in here?

DAlperin:
https://github.com/MaterializeInc/materialize/pull/33021

PR #32973, opened 2025-07-09
Unbork subquery lowering

frankmcsherry:
Our subquery lowering has been incorrect for a while, as it relates to errors. It produces unguarded errors atop a filtered collection, and that isn't good enough. To behave properly, the error needs to be guarded by an `If` (a `CASE` in SQL). This changes lowering to use what I hope is a more appropriate expression.

I used `Datum::Dummy` because there is a place that a typed datum is required, but it is unclear if there are default values for types. I couldn't find such a function in the 1kLOC for one `impl` block for `ScalarType`, but perhaps it's there somewhere. Would love to remove it, so lmk where to find such a thing.

Haven't tested yet, because python doesn't seem to work correctly. Spoiled on Rust. T.T

### Checklist


ggevay:
Corresponding thread that points out the problem with the (original version of the) PR: https://materializeinc.slack.com/archives/C08ACQNGSQK/p1752111767252899?thread_ts=1752097050.500379&cid=C08ACQNGSQK
(Linking this so that if at some point later we come across this PR we won't wonder why it's still in draft.)

frankmcsherry:
Replaced by #32978

PR #32978, opened 2025-07-10
Guard subquery size with a TableFunc

frankmcsherry:
### Checklist

### DIFF
--- a/src/expr/src/relation/func.rs
+++ b/src/expr/src/relation/func.rs
@@ -3734,6 +3734,19 @@ pub enum TableFunc {
     GenerateSeriesInt64,
     GenerateSeriesTimestamp,
     GenerateSeriesTimestampTz,
+    /// Supplied with an input count,
+    ///   1. Adds an error column announcing too many results,
+    ///   2. Filters the row away if the count is only one.
+    /// This logic could nearly be achieved with map, filter, project logic,
+    /// but has been challenging to do in a way that respects the vagaries of
+    /// SQL and our semantics.
+    /// Before replacing this by an `MirScalarExpr`, quadruple check that it
+    /// would not result in misoptimizations due to expression evaluation order
+    /// being utterly undefined, and predicate pushdown trimming any fragements
+    /// that might produce columns that will not be needed.
+    GuardSubquerySize {
+        column_type: ScalarType,
+    },
     Repeat,
     UnnestArray {
         el_typ: ScalarType,
@@ -3779,6 +3792,9 @@ impl RustType<ProtoTableFunc> for TableFunc {
                 TableFunc::GenerateSeriesInt64 => Kind::GenerateSeriesInt64(()),
                 TableFunc::GenerateSeriesTimestamp => Kind::GenerateSeriesTimestamp(()),
                 TableFunc::GenerateSeriesTimestampTz => Kind::GenerateSeriesTimestampTz(()),
+                TableFunc::GuardSubquerySize { column_type } => {
+                    Kind::GuardSubquerySize(column_type.into_proto())
+                }
                 TableFunc::Repeat => Kind::Repeat(()),
                 TableFunc::UnnestArray { el_typ } => Kind::UnnestArray(el_typ.into_proto()),
                 TableFunc::UnnestList { el_typ } => Kind::UnnestList(el_typ.into_proto()),
@@ -3818,6 +3834,9 @@ impl RustType<ProtoTableFunc> for TableFunc {
             Kind::GenerateSeriesInt64(()) => TableFunc::GenerateSeriesInt64,
             Kind::GenerateSeriesTimestamp(()) => TableFunc::GenerateSeriesTimestamp,
             Kind::GenerateSeriesTimestampTz(()) => TableFunc::GenerateSeriesTimestampTz,
+            Kind::GuardSubquerySize(x) => TableFunc::GuardSubquerySize {
+                column_type: x.into_rust()?,
+            },
             Kind::Repeat(()) => TableFunc::Repeat,
             Kind::UnnestArray(x) => TableFunc::UnnestArray {
                 el_typ: x.into_rust()?,
@@ -3909,6 +3928,16 @@ impl TableFunc {
             }
             TableFunc::GenerateSubscriptsArray => {
                 generate_subscripts_array(datums[0], datums[1].unwrap_int32())
+            }
+            TableFunc::GuardSubquerySize { column_type: _ } => {
+                // We error if the count is greater than one,
+                // or produce no rows if not greater than one.
+                let count = datums[0].unwrap_int64();
+                if count > 1 {
+                    Err(EvalError::MultipleRowsFromSubquery)
+                } else {
+                    Ok(Box::new([].into_iter()))
+                }
             }
             TableFunc::Repeat => Ok(Box::new(repeat(datums[0]).into_iter())),
             TableFunc::UnnestArray { .. } => Ok(Box::new(unnest_array(datums[0]))),
@@ -4017,6 +4046,11 @@ impl TableFunc {
                 let keys = vec![vec![0]];
                 (column_types, keys)
             }
+            TableFunc::GuardSubquerySize { column_type } => {
+                let column_types = vec![column_type.clone().nullable(false)];
+                let keys = vec![];
+                (column_types, keys)
+            }
             TableFunc::Repeat => {
                 let column_types = vec![];
                 let keys = vec![];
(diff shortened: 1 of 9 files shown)


test/sqllogictest/transform/literal_lifting.slt
```diff
@@ -804,19 +804,18 @@ Explained Query:
         Project (#0{a}) // { arity: 1 }
           Join on=(#0{a} = #1) type=differential // { arity: 2 }
             implementation
-              %0:t_pk[#0{a}]UKiif » %1[#0]UKiif
+              %0:t_pk[#0{a}]UKiif » %1[#0]Kiif
```
frankmcsherry, line 807: @ggevay This missing `U` may connect to some pattern detector we have, but also I'm delighted to disappoint it if so. It may also connect to having no clue about primary keys offered by the `TableFunc` logic, but similarly no real interest in enabling the optimizer around that.

test/sqllogictest/transform/literal_lifting.slt
```diff
@@ -847,19 +846,18 @@ Explained Query:
         Project (#0{a}) // { arity: 1 }
           Join on=(#0{a} = #1) type=differential // { arity: 2 }
             implementation
-              %1[#0]UK » %0:t[#0{a}]Kiif
+              %0:t[#0{a}]Kiif » %1[#0]Kiif
```
frankmcsherry, line 849: Similarly unclear, but also a join involving a collection that is either errored or empty.

src/expr/src/relation/func.rs
```diff
@@ -4017,6 +4046,11 @@ impl TableFunc {
                 let keys = vec![vec![0]];
                 (column_types, keys)
             }
+            TableFunc::GuardSubquerySize { column_type } => {
+                let column_types = vec![column_type.clone().nullable(false)];
+                let keys = vec![];
```
frankmcsherry, line 4061: @ggevay no clue what this should be, nor what keys would even mean in this context. Is it "if the output were treated as a relation we were joining in, would it have any keys"? If so, we could give some unique keys here as the output has at most one row.

src/expr/src/relation/func.rs
```diff
@@ -3734,6 +3734,19 @@ pub enum TableFunc {
     GenerateSeriesInt64,
     GenerateSeriesTimestamp,
     GenerateSeriesTimestampTz,
+    /// Supplied with an input count,
+    ///   1. Adds an error column announcing too many results,
```
ggevay: It would be good to mention that this column would hold the subquery's result if it were not an error, to make it easier to understand how things tie out with the `Union`.

Also, the column doesn't actually hold the error yet (at least not until we have `Datum::Error` :)), so the term "error column" sounds a bit strange.

src/transform/src/canonicalization/flatmap_to_map.rs
ggevay, line 1: Would be good to also rename the file.

test/sqllogictest/transform/literal_lifting.slt
ggevay: Yes, we used to see the unique key of the `Reduce`, but now the `FlatMap` blocks the unique key derivation.

But seems fine, because, as you say elsewhere, it's a collection that is either errored or empty.

test/sqllogictest/transform/literal_lifting.slt
ggevay: Ok

test/sqllogictest/transform/literal_lifting.slt
```diff
@@ -919,8 +917,8 @@ EXPLAIN OPTIMIZED PLAN WITH(humanized expressions, arity, join implementations)
 Explained Query:
   CrossJoin type=delta // { arity: 4 }
     implementation
-      %0:t1 » %2[×]U » %1:t1[×]ef
-      %1:t1 » %2[×]U » %0:t1[×]
+      %0:t1 » %1:t1[×]ef » %2[×]ef
+      %1:t1 » %2[×]ef » %0:t1[×]
```
ggevay, line 921: (Same thing as above.)

src/expr/src/relation/func.rs
ggevay: Yeah, I think it's fine to have no keys here.

> If so, we could give some unique keys here as the output has at most one row.

0 rows actually, because we only emit errors, never any data, right? But I don't see how the optimizer could make use of this fact, so it's probably better to not try to be too smart here.

(A different thing, which would be actually valuable, is to detect keys at the Union of the subquery pattern, by noting that this input of the Union is empty, so we could just copy the keys from the other input of the Union to the Union itself. But putting the `[]` key on this input wouldn't help with this, since `[]` means 0 or 1 row, while we specifically are interested in 0 rows for this trick. Added a note [here](https://github.com/MaterializeInc/database-issues/issues/4778).)

ggevay:
`test/sqllogictest/transform/fold_vs_dataflow/*` also needs a rewrite, as per Nightly. (Nightly runs more slt files than the normal CI.)

src/expr/src/relation/func.rs
frankmcsherry: Oh you are right. In the first version of the code I tried to pack the error literal into the column, but then realized that the method allows us to return an `Err` instead. Will tidy up, thanks!

src/transform/src/canonicalization/flatmap_to_map.rs
frankmcsherry: Yeah, good point. I'll clean some stuff up tomorrow with the feature flags too.

src/expr/src/relation/func.rs
frankmcsherry: > 0 rows actually, because we only emit errors, never any data, right?

Fwiw, it's very important not to reveal this fact! Several constructions did reveal this, and if the optimizer learns about the fact it will just prune the branch.

src/transform/src/canonicalization/flatmap_to_map.rs
frankmcsherry: Actually, I may not clean this up as part of this PR, as I think "the right answer" involves new directories, stuff like that. I'm up for doing that next, but I'd love to decouple the efforts.

src/expr/src/relation/func.rs
frankmcsherry: I've changed this to be clearer, but one other thing I cleaned up (and wanted to mention, in case it is clearly wrong): the behavior is now documented as erroring on anything that is not exactly one, which includes zero and negative counts. They "shouldn't" happen, but should we ever relax that in the future, this is meant to produce an error rather than silently ignore that weirdness.

src/transform/src/canonicalization/flatmap_to_map.rs
ggevay: Ok, no big reorganization is needed, just renaming the file itself to `flatmap_elimination.rs` would be good in this PR.

src/expr/src/relation/func.rs
ggevay: > erroring on anything that is not exactly one, which includes zero and negative counts.

Zero count can happen, and it shouldn't error. It should put a null as the result of the subquery. E.g.:
```
create table t(x int);
insert into t values (7);

select *, (select x from t) from t;

select *, (select x from t where x < 5) from t;
```

You mean you have this code change locally at the moment? Asking because the PR still errors only for
`count > 1`.

ggevay: Ah ok, but it won't get an actual 0 count in that case, because this is not a SQL `count`, but an MIR `count`, which has an empty result on an empty input. Then it's fine.

src/expr/src/relation/func.rs
```diff
@@ -3910,6 +3939,16 @@ impl TableFunc {
             TableFunc::GenerateSubscriptsArray => {
                 generate_subscripts_array(datums[0], datums[1].unwrap_int32())
             }
+            TableFunc::GuardSubquerySize { column_type: _ } => {
+                // We error if the count is greater than one,
+                // or produce no rows if not greater than one.
+                let count = datums[0].unwrap_int64();
+                if count != 1 {
```
ggevay, line 3946: Could you update the comment please? Also, maybe you could explain it in the comment [why it is ok to error on 0 here.](https://github.com/MaterializeInc/materialize/pull/32978#discussion_r2200858370)

ggevay: Actually, how about creating a special error here for the `count < 1` case, and marking it as an internal error?

Or, alternatively, we could put the actual count into the error itself. This would be also helpful for users for debugging their queries, and potentially useful for us for debugging Materialize itself if we ever see a 0 or negative count show up.

frankmcsherry: I'd prefer to do none of these, and will revert the change if that is better.

ggevay: It's ok as it is. I'll tweak the error when I have time.

PR #32979, opened 2025-07-10
SQL server: consistent snapshot and lsn selection

martykulma:
The current mechanism for snapshot and LSN selection is not correct due to the asynchronous nature of MS SQL CDC and the implementation details of `SNAPSHOT` isolation.

This implementation is similar to the MySQL implementation in that it locks the tables for a brief period to establish a point in time where a consistent view of the DB can be observed. Once the `LSN` and transaction sequence number (`XSN`), the table locks are dropped, and we have a consistent version of the table(s) to read from as well as a correct `LSN` to process CDC events.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9476

### Tips for reviewer

You can look at the commits to see the history, but the actual fix is not until the end.  

I didn't squash my commits because the bug that this fixes was created in response to this PR (at the time a draft that was just a test change to show the race condition).  It will get squashed on merge!

### Checklist

### DIFF
--- a/src/sql-server-util/src/cdc.rs
+++ b/src/sql-server-util/src/cdc.rs
@@ -152,7 +152,7 @@ impl<'a> CdcStream<'a> {
         &'b mut self,
         instances: Option<BTreeSet<Arc<str>>>,
         worker_id: usize,
-        source_id: &'b GlobalId,
+        source_id: GlobalId,
     ) -> Result<
         (
             Lsn,
--- a/src/storage/src/source/sql_server/replication.rs
+++ b/src/storage/src/source/sql_server/replication.rs
@@ -150,7 +150,7 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
                 }
 
                 let (snapshot_lsn, snapshot_stats, snapshot_streams) = cdc_handle
-                    .snapshot(Some(needs_snapshot), config.worker_id, &config.id)
+                    .snapshot(Some(needs_snapshot), config.worker_id, config.id)
                     .await?;
                 let snapshot_cap = data_cap_set.delayed(&snapshot_lsn);
 
--- a/src/sql-server-util/examples/cdc.rs
+++ b/src/sql-server-util/examples/cdc.rs
@@ -71,7 +71,7 @@ async fn main() -> Result<(), anyhow::Error> {
 
     // Get an initial snapshot of the table.
     let (lsn, stats, snapshot) = cdc_handle
-        .snapshot(None, 1, &mz_repr::GlobalId::User(1))
+        .snapshot(None, 1, mz_repr::GlobalId::User(1))
         .await?;
     tracing::info!("snapshot stats: {stats:?}");
     {


src/sql-server-util/src/cdc.rs
```diff
@@ -140,13 +181,57 @@ impl<'a> CdcStream<'a> {
         // the upstream DB is ready for CDC.
         self.wait_for_ready().await?;
 
+        tracing::info!("Upstream is ready");
```
ptravers: nit: might want to add some additional context here and also are we sure we want this at info level?

src/sql-server-util/src/cdc.rs
```diff
@@ -140,13 +181,57 @@ impl<'a> CdcStream<'a> {
...
+        // as it will be just be locking the table(s).
+        let mut fencing_client = self.client.new_connection().await?;
+        let mut fence_txn = fencing_client.transaction().await?;
+
+        // TODO (maz): we should consider a timeout or a lock + snapshot per-table instead of collectively
+        for (_capture_instance, schema, table) in &tables {
+            tracing::trace!(%schema, %table, "locking table");
+            fence_txn.lock_table_exclusive(&*schema, &*table).await?;
+        }
+        tracing::info!(?tables, "Locked tables");
```
ptravers: nit: same comment as above.

src/sql-server-util/src/cdc.rs
martykulma: added a comment

src/sql-server-util/src/cdc.rs
martykulma: ditto!

src/sql-server-util/src/cdc.rs
ptravers: ah I see! would we ever see these concurrently on the same node? would it be clear which database is ready?

src/sql-server-util/src/cdc.rs
```diff
@@ -140,13 +181,63 @@ impl<'a> CdcStream<'a> {
...
+        // often, but if there are problems here, it will be much easier to troubleshoot
+        // knowing where stall/hang might be happening.
+        tracing::info!("Upstream is ready");
+
+        // The client that will be used for fencing does not need any special isolation level
+        // as it will be just be locking the table(s).
+        let mut fencing_client = self.client.new_connection().await?;
+        let mut fence_txn = fencing_client.transaction().await?;
+
+        // TODO (maz): we should consider a timeout or a lock + snapshot per-table instead of collectively
```
ptravers: what telemetry do we need to be able to make decision about the necessity of locking per table?

I would guess we need to know whether the table would have been locked for less time that way?

ptravers (approved):

src/sql-server-util/src/cdc.rs
ptravers: follow up: maybe we can batch with a configurable number of tables per batch? we might want to track with a histogram/metric roughly how long we are locking a tables.

seems like this only becomes a problem if there's a very large number of tables or a very slow lock acquisition.

src/sql-server-util/src/cdc.rs
martykulma: 🤦 - just realized i didn't include `?tables`

petrosagg: +1 on adding context. The standard format we use is `timely-{worker_id}` and also include the source id this message is for

src/sql-server-util/src/cdc.rs
```diff
@@ -140,13 +181,63 @@ impl<'a> CdcStream<'a> {
...
+        for (_capture_instance, schema, table) in &tables {
+            tracing::trace!(%schema, %table, "locking table");
+            fence_txn.lock_table_exclusive(&*schema, &*table).await?;
```
petrosagg: Do we need an exclusive lock in the end? We don't care if other people concurrently read the table with us, we only want to block writes.

src/sql-server-util/src/cdc.rs
petrosagg: same as above, let's add context about the timely worker and source id here

src/sql-server-util/src/cdc.rs
```diff
@@ -140,13 +181,63 @@ impl<'a> CdcStream<'a> {
...
+        // visible to this snapshot and the LSN of such a write will be less than
+        // or equal to the LSN captured here.
+        txn.create_savepoint(SAVEPOINT_NAME).await?;
+        tracing::info!(%SAVEPOINT_NAME, "Created savepoint");
+        let lsn = txn.get_lsn().await?;
+
+        // Once the XSN is esablished and the LSN captured, the tables no longer
+        // need to be locked.  Any writes that happen to the upstream tables
+        // will have a LSN higher than our captured LSN, and will be read from CDC.
+        fence_txn.rollback().await?;
```
petrosagg, line 239: I think we should commit this transaction instead of rolling it back. What I'm thinking is that when we wrap up the fencing transaction we actually want to ensure that the server didn't decide to cancel it or otherwise mess with it while we weren't looking, since in that case the lock we think we have might not be there.

I don't know if SQL does this too but for example pg cancels a transaction if it can't serialize it and you only learn it at commit time.

src/sql-server-util/src/lib.rs
```diff
@@ -352,10 +362,71 @@ impl<'a> Transaction<'a> {
...
+    /// Retrieve the [`Lsn`] associated with the current session.
+    ///
+    /// MS SQL Server will not assign a [`Lsn`] until a write is performed (e.g. via `SAVE TRANSACTION`).
+    pub async fn get_lsn(&mut self) -> Result<Lsn, SqlServerError> {
+        static CURRENT_LSN_QUERY: &str = "
+SELECT dt.database_transaction_begin_lsn
+FROM sys.dm_tran_database_transactions AS dt
+JOIN sys.dm_tran_session_transactions AS st
+  ON dt.transaction_id = st.transaction_id
+WHERE st.session_id = @@SPID
```
petrosagg: This is the query that the Machine produced but on a second look I don't understand why it has to go through the session id. There is `dm_tran_current_transaction` that we can join with `dm_tran_database_transactions` which seems more straightforward.

Also, I see that the `dm_tran_database_transactions` table includes both a `begin_lsn` and a `database_transaction_most_recent_savepoint_lsn`. Should we use the latter to make it clear that we're using the LSN that we caused by creating a savepoint?

src/sql-server-util/src/lib.rs
```diff
@@ -352,10 +362,71 @@ impl<'a> Transaction<'a> {
...
+    /// The lock is obtained using a `SELECT` statement that will not read the table. The lock is released
+    /// after transaction commit or rollback.
+    pub async fn lock_table_exclusive(
+        &mut self,
+        schema: &str,
+        table: &str,
+    ) -> Result<(), SqlServerError> {
+        // This query probably seems odd, but there is no LOCK command in MS SQL. Locks are specified
+        // in SELECT using the WITH keyword.  This query does not need to return any rows to lock the table,
+        // hence the 1=0, which is something short that always evaluates to false in this universe;
```
petrosagg:
```suggestion
        // hence the 1=0, which is something short that always evaluates to false in this universe.
```

src/sql-server-util/src/lib.rs
```diff
@@ -400,6 +471,7 @@ impl<'a> Transaction<'a> {
         // N.B. Mark closed _before_ running the query. This prevents us from
         // double closing the transaction if this query itself fails.
         self.closed = true;
+        self.savepoints.clear();
```
petrosagg: You don't need to clear the savepoints here since this method has a `self` receiver and the entire struct will be dropped before returning.

src/sql-server-util/src/lib.rs
```diff
@@ -410,6 +482,7 @@ impl<'a> Transaction<'a> {
...
```
petrosagg: Same here

src/storage/src/source/sql_server/replication.rs
```diff
@@ -204,7 +204,7 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
                     data_output
                         .give_fueled(
                             &snapshot_cap,
-                            ((*partition_idx, message), snapshot_lsn, Diff::ONE),
+                            ((*partition_idx, message), Lsn::minimum(), Diff::ONE),
```
petrosagg, line 208: !! Great catch

src/sql-server-util/src/cdc.rs
petrosagg: > or a lock + snapshot per-table instead of collectively

I think we know this is better, right? I would just create an issue for it so that we do it. (or, if it's not a big refactor do it now)

src/sql-server-util/src/lib.rs
```diff
@@ -336,6 +345,7 @@ pub type RowStream<'a> =
 pub struct Transaction<'a> {
     client: &'a mut Client,
     closed: bool,
+    savepoints: Vec<String>,
```
petrosagg: This field is never read. Do we need it?

src/sql-server-util/src/cdc.rs
```diff
@@ -140,13 +181,63 @@ impl<'a> CdcStream<'a> {
...
+        }
+
+        // Because the tables are exclusively locked, any write operation has either
+        // completed, or is blocked. The LSN and XSN acquired now will represent a
+        // consistent point-in-time view, such that any comitted write will be
+        // visible to this snapshot and the LSN of such a write will be less than
+        // or equal to the LSN captured here.
+        txn.create_savepoint(SAVEPOINT_NAME).await?;
+        tracing::info!(%SAVEPOINT_NAME, "Created savepoint");
+        let lsn = txn.get_lsn().await?;
```
petrosagg: We can run this statement after we have committed the fencing transaction to reduce the amount of time we hold the locks for. The LSN is already established and it's fine to read it outside the critical section

src/sql-server-util/src/cdc.rs
```diff
@@ -177,10 +268,10 @@ impl<'a> CdcStream<'a> {
                 tracing::trace!(%capture_instance, %schema_name, %table_name, "snapshot end");
             }
 
-            // Slightly awkward, but if the commit fails we need to conform to
+            // Slightly awkward, but if the rollback fails we need to conform to
             // type of the stream.
-            if let Err(e) = txn.commit().await {
-                yield ("commit".into(), Err(e));
+            if let Err(e) = txn.rollback().await {
```
petrosagg, line 275: Didn't follow this change. Is there something we want to undo here?

src/sql-server-util/src/cdc.rs
```diff
@@ -545,6 +636,40 @@ impl TryFrom<&[u8]> for Lsn {
...
+            ));
+        }
+        let mut decimal_lsn = value.int_part();
+        // LSN is composed of 4 bytes : 4 bytes : 2 bytes
+        // and MS provided the method to decode that here
+        // https://github.com/microsoft/sql-server-samples/blob/master/samples/features/ssms-templates/Sql/Change%20Data%20Capture/Enumeration/Create%20Function%20fn_convertnumericlsntobinary.sql
+
+        let vlf_id = u32::try_from(decimal_lsn / 10_i128.pow(15))
+            .map_err(|e| format!("Failed to decode vlf_id for lsn {decimal_lsn}: {e:?}"))?;
+        decimal_lsn -= i128::try_from(vlf_id).unwrap() * 10_i128.pow(15);
```
petrosagg: A direct `From` implementation exists to go from `u32` (and others) to `i128` https://doc.rust-lang.org/std/primitive.i128.html#impl-From%3Cu32%3E-for-i128

```suggestion
        decimal_lsn -= i128::from(vlf_id) * 10_i128.pow(15);
```

src/sql-server-util/src/cdc.rs
```diff
@@ -17,11 +17,49 @@
...
+//! returned from any CDC tables or CDC functions will always be stale,
+//! in relation to the source table that CDC is tracking. The system table
+//! [sys.dm_tran_database_transactions](https://learn.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-tran-database-transactions-transact-sql?view=sql-server-ver17)
+//! will contain a [`Lsn`] for any transaction that performs a write operation.
+//! Creating a savepoint using [SAVE TRANSACTION](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/save-transaction-transact-sql?view=sql-server-ver17).
+//! is sufficient to generate a [`Lsn`] in this case.
+//!
+//! Unfortunately, it isn't sufficient to just fix the [`Lsn`] and `XSN`. There exists
+//! the possibility that a write transaction may have inserted a row into one
+//! of the tables in the snapshot, but that write has not comitted. In this case,
```
petrosagg:
```suggestion
//! of the tables in the snapshot, but that write has not committed. In this case,
```

src/sql-server-util/src/cdc.rs
```diff
@@ -17,11 +17,49 @@
...
+//! the `INSERT` has already been written to the transaction log at a [`Lsn`]
+//! less than than the one captured, but the snapshot will *not* observe that INSERT
+//! because the transaction has not committed, and may not commit until after
+//! the snapshot is complete. In order to force a clear delineation of updates,
+//! the upstream tables in the snapshot must be locked. This lock only needs
+//! to exist long enough to establish the [`Lsn`] and `XSN`.
```
petrosagg: Something doesn't check out for me in this paragraph.

> There exists the possibility that a write transaction may have inserted a row into one of the tables in the snapshot, but that write has not committed. In this case, the `INSERT` has already been written to the transaction log at a [`Lsn`] less than than the one captured, but the snapshot will *not* observe that INSERT because the transaction has not committed, and may not commit until after the snapshot is complete.

This sounds totally fine? If that transaction commits after the snapshot is complete then in the CDC stream that write will be associated with an LSN that is higher that then snapshot LSN we captured and so it will be correctly applied on top of the snapshot, which as you say will not observe the write.

I think the danger, and reason why we need locks, is that we have no way establishing a transaction MVCC snapshot and an LSN boundary **atomically**. With postgres we get this API by creating a temp replication slot as the first statement of the transaction with `USE_SNAPSHOT`, which links the two together. In SQL Server we are forced to perform two independent steps, and the bad scenario is if a write transaction commits between the MVCC point and the LSN boundary point. If that happens we get the table snapshot at the MVCC point, assume it happened at the LSN boundary point, and we completely miss the in-between write during the replication phase.

src/sql-server-util/src/cdc.rs
```diff
@@ -17,11 +17,49 @@
...
+//!
+//! SQL server supports exclusive table locks, but those will only be released
+//! once the outermost transaction completes. For this reason, this module
+//! uses two connections for the snapshot process.  The first conection is used
```
petrosagg:
```suggestion
//! uses two connections for the snapshot process.  The first connection is used
```

src/sql-server-util/src/cdc.rs
```diff
@@ -17,11 +17,49 @@
...
+//! to initiate a transaction and lock the upstream tables.  While the first
+//! connection maintains the locks, the second connection starts a transaction
+//! with [`TransactionIsolationLevel::Snapshot`] isolation and creates a
+//! savepoint.  Once the savepoint is created and [`Lsn`] is captured, the first
+//! connection rolls back the transaction. The snapshot is created by the second
```
petrosagg:
```suggestion
//! connection commits the transaction. The snapshot is created by the second
```
(depends on the discussion in the other comment)

src/sql-server-util/src/cdc.rs
```diff
@@ -355,7 +446,7 @@ impl<'a> CdcStream<'a> {
             }
         }
 
-        // Ensure all of the capture instances are reporting an LSN.
+        // Ensure all of the capture instances are reporting a LSN.
```
petrosagg: I think "an LSN" is correct because the rule depends on of the pronunciation "el-es-en". Here is a similar example where it's "an LED" instead of "a LED" https://en.wikipedia.org/wiki/Light-emitting_diode

We also write "a UV lamp" even though it's "an ultraviolet lamp" (example courtesy of ChatGPT)

petrosagg (commented):
This looks great! One general question: did we confirm that the mzcompose test fails when we run it with the current code?

src/sql-server-util/src/cdc.rs
martykulma: I think it's the lesser of 2 evils, can add a comment.  There is an [UPDLOCK](https://learn.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table?view=sql-server-ver17#updlock), but it only applies to rows/pages. My understanding is we'd have to read the entire table.
If it's combined with `TABLOCK`, it ends up promoted to an exclusive table lock anyway (tested to make sure this was the case).

petrosagg: I was thinking of `TABLOCK` with `HOLDLOCK` which according to the docs [here](https://learn.microsoft.com/en-us/sql/t-sql/queries/hints-transact-sql-table?view=sql-server-ver17#tablock) will get a shared lock until the end of the transaction. From a quick test I did this allows other transactions to read the table but makes all writes block, which is what we need. The statement I used was `select * from t1 WITH (TABLOCK, HOLDLOCK)`

src/sql-server-util/src/cdc.rs
martykulma: I'm not opposed to changing it.  My reasoning for rollback is that this transaction (and the snapshot transaction) should never make any changes to the database. It prevents us accidentally committing things due to some innocuous change in the future. That's entirely based on my own battle scars 😁 

A rollback without an active transaction does fail in SQL server
```
1> rollback;
2> go
Msg 3903, Level 16, State 1, Server 41d33a81ef67, Line 1
The ROLLBACK TRANSACTION request has no corresponding BEGIN TRANSACTION.
```

In the case of PG, when a transaction is canceled, are you saying that only the `commit` returns the error?  I would expect that `commit` or `rollback` would fail because there is no transaction to operate on at that point (or it's in an error state).

petrosagg: I tried to construct a case using `KILL` but I can't, rollback hangs if I kill the session under its nose. I don't feel strongly about this, it should be fine to use rollback here if any error gets surfaced instead of being swept under the carpet.

src/sql-server-util/src/cdc.rs
martykulma: My concern with `SELECT * FROM t1 WITH (TABLOCK, HOLDLOCK)` is that it will read the entire table in the fencing path. We are currently using `SELECT * FROM t1 WITH (TABLOCKX) WHERE 1=0` which reads no rows.

If you're not opposed, I'd like to tackle the lock improvement after this PR.  Relaxing the locking will improve the user experience, but ultimately does not affect correctness. 

A quick test of using `WHERE 1=0` with `HOLDLOCK` did not prevent any writes, so we know the query must return *some* data. For example, `SELECT COUNT(*) FROM (SELECT TOP 1 * FROM t1 WITH (TABLOCK, HOLDLOCK)) AS t` should read a minimal amount of data, but I would feel better with some testing with large tables to validate that is the case.

petrosagg: Ah sorry, the `*` in there wasn't load bearing, it was just me typing in the terminal.

> A quick test of using WHERE 1=0 with HOLDLOCK did not prevent any writes

Weird, this works fine for me (writes are prevented). This is what I'm doing:

```
# terminal A
begin transaction
select * from t1 WITH (TABLOCK, HOLDLOCK) where 1=0
go  <-- returns zero rows

# terminal B
select * from t1
go  <-- works
insert into t1 values (42)
go  <-- blocks
```

> so we know the query must return some data. 

If that's the case (can't repro this myself) we might have a problem if the table in question is empty during the initial snapshot. Can you share the commands you ran in your quick test to at least record it here for when we relax the lock?

> If you're not opposed, I'd like to tackle the lock improvement after this PR.

Not opposed, let's make an issue in the epic to not forget

martykulma:
> This looks great! One general question: did we confirm that the mzcompose test fails when we run it with the current code?

yes, I cherry-picked that commit to main and it failed consistently for me (I was running repeatedly to slim down the number of updates the test makes, which started at something like 10,000 😅 )

src/sql-server-util/src/cdc.rs
martykulma: ok, I think I know why. The issue appears to be if the session that establishes the locks is using `SNAPSHOT` isolation, it doesn't behave the same way!

So it looks like we can use `(TABLOCK, HOLDLOCK)` and probably wouldn't be a bad idea to explicitly set the transaction level to `READ COMMITTED`.

I inadvertently was running the commands in the wrong sessions:
session 1 (snapshot isolation)
```
1> begin tran;
2> go
1> select * from t1 with (TABLOCK, HOLDLOCK) where 1=0;
2> go

(0 rows affected)
```

session 2 (read committed isolation, and no explicit transaction)
```
1> insert into t1 values (4,4);
2> go
(1 row affected)
```

both sessions show the same in the `select *`, and commit returns no error in session 2
```
1> select * from t1;
2> go
id  1
val 1

id  2
val 2

id  3
val 3

id  4
val 4


(4 rows affected)

```

src/sql-server-util/src/cdc.rs
martykulma: fair.. though I feel this is the same type of nonsense that leads to people sayings ATM machine! English is hard.

src/sql-server-util/src/cdc.rs
petrosagg: Oh my! Why does it have to be so subtle!? 💀 So locks are ignored in snapshot isolation. Makes total sense..

src/sql-server-util/src/lib.rs
martykulma: removing savepoints, it was added to support rolling back, but that isn't necessary now that we have 2 connections!

src/sql-server-util/src/cdc.rs
martykulma: indeed!

src/sql-server-util/src/cdc.rs
martykulma: same as the above statement with commit vs. rollback.  We do not intend to make any changes to the DB, so we rollback the transaction that read all the data.

src/sql-server-util/src/lib.rs
martykulma: no, removing it

src/sql-server-util/src/lib.rs
martykulma: Sure, I can make that change.

As to using the `most_recent_savepoint_lsn` instead, the risk seems minimal.  The savepoint is what establishes both of those LSNs, and as long as don't do something silly like create additional savepoints outside the critical section before retrieving the LSN, should be fine.

....because we can certainly make it drift! Though, I can't imagine why we would need to rollback the savepoint and try again.
```
1> begin tran;
2> save tran foo;
3> rollback tran foo;
4> save tran foo;
5> rollback tran foo;
6> save tran foo;
7> select dt.database_transaction_begin_lsn, dt.database_transaction_most_recent_savepoint_lsn from sys.dm_tran_current_transaction ct inner join sys.dm_tran_database_transactions dt on (ct.transaction_id = dt.transaction_id) where database_transaction_begin_lsn is not null;
8> go
database_transaction_begin_lsn                 220000000020800072
database_transaction_most_recent_savepoint_lsn 220000000020800077
```

src/sql-server-util/src/cdc.rs
martykulma: If possible, sure.  [Created a GA task](https://github.com/MaterializeInc/database-issues/issues/9512)! this PR has plenty of action already 😁

src/sql-server-util/src/cdc.rs
martykulma: Sorry, that was me not proofreading.  I unfortunately wrote it in parts.  I trimmed it down.  Thanks for catching that Petros!

src/sql-server-util/src/lib.rs
martykulma: removed

src/sql-server-util/src/cdc.rs
petrosagg: Sounds good!

src/sql-server-util/src/cdc.rs
```diff
@@ -134,20 +175,72 @@ impl<'a> CdcStream<'a> {
...
+        let res = txn
+            .simple_query("SELECT TOP 1 object_id FROM cdc.change_tables")
+            .await?;
+        if res.len() != 1 {
+            Err(SqlServerError::InvariantViolated(
+                "No objects found in cdc.change_tables".into(),
+            ))?
+        }
+
+        // Because the tables are exclusively locked, any write operation has either
```
petrosagg:
```suggestion
        // Because the tables are locked, any write operation has either
```

src/sql-server-util/src/cdc.rs
```diff
@@ -134,20 +175,72 @@ impl<'a> CdcStream<'a> {
...
+        // completed, or is blocked. The LSN and XSN acquired now will represent a
+        // consistent point-in-time view, such that any comitted write will be
```
petrosagg:
```suggestion
        // consistent point-in-time view, such that any committed write will be
```

src/sql-server-util/src/cdc.rs
```diff
@@ -158,29 +251,29 @@ impl<'a> CdcStream<'a> {
...
+                tracing::trace!(%source_id, %capture_instance, %schema_name, %table_name, "timely-{worker_id} snapshot start");
 
                 let snapshot = crate::inspect::snapshot(txn.client, &*schema_name, &*table_name);
                 let mut snapshot = std::pin::pin!(snapshot);
                 while let Some(result) = snapshot.next().await {
                     yield (Arc::clone(&capture_instance), result);
                 }
 
-                tracing::trace!(%capture_instance, %schema_name, %table_name, "snapshot end");
+                tracing::trace!(%source_id, %capture_instance, %schema_name, %table_name, "timely-{worker_id} tmsnapshot end");
```
petrosagg:
```suggestion
                tracing::trace!(%source_id, %capture_instance, %schema_name, %table_name, "timely-{worker_id} snapshot end");
```

src/sql-server-util/src/lib.rs
```diff
@@ -356,6 +365,71 @@ impl<'a> Transaction<'a> {
...
+    ///
+    /// MS SQL Server will not assign an [`Lsn`] until a write is performed (e.g. via `SAVE TRANSACTION`).
+    pub async fn get_lsn(&mut self) -> Result<Lsn, SqlServerError> {
+        static CURRENT_LSN_QUERY: &str = "
+SELECT dt.database_transaction_most_recent_savepoint_lsn
+FROM sys.dm_tran_database_transactions dt
+JOIN sys.dm_tran_current_transaction ct
+    ON ct.transaction_id = dt.transaction_id
+WHERE dt.database_transaction_most_recent_savepoint_lsn IS NOT NULL
+";
```
petrosagg: You can indent this string and ignore the leading whitespace [like so](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=925cbf09d65b99c7b1292631b4c021d7) (notice the `\` at the end).

```suggestion
        static CURRENT_LSN_QUERY: &str = "SELECT dt.database_transaction_most_recent_savepoint_lsn \
            FROM sys.dm_tran_database_transactions dt \
            JOIN sys.dm_tran_current_transaction ct \
                ON ct.transaction_id = dt.transaction_id \
            WHERE dt.database_transaction_most_recent_savepoint_lsn IS NOT NULL";
```

src/sql-server-util/src/lib.rs
```diff
@@ -356,6 +365,71 @@ impl<'a> Transaction<'a> {
...
+        let result = self.client.simple_query(CURRENT_LSN_QUERY).await?;
+        crate::inspect::parse_numeric_lsn(&result)
+    }
+
+    /// Lock the provided table to prevent writes but allow reads , uses `(TABLOCK, HOLDLOCK)`.
```
petrosagg:
```suggestion
    /// Lock the provided table to prevent writes but allow reads, uses `(TABLOCK, HOLDLOCK)`.
```

src/sql-server-util/src/lib.rs
```diff
@@ -356,6 +365,71 @@ impl<'a> Transaction<'a> {
...
+    ///
+    /// This will set the transaction isolation level to `READ COMMITTED` and then obtain the
+    /// lock is obtained using a `SELECT` statement that will not read and data from the table.
```
petrosagg:
```suggestion
    /// lock using a `SELECT` statement that will not read any data from the table.
```

src/sql-server-util/src/cdc.rs
```diff
@@ -114,6 +151,8 @@ impl<'a> CdcStream<'a> {
     pub async fn snapshot<'b>(
         &'b mut self,
         instances: Option<BTreeSet<Arc<str>>>,
+        worker_id: usize,
+        source_id: &'b GlobalId,
```
petrosagg: Rust forces you to put a lifetime due to async and this parameter being a reference.. You can avoid that with even more syntax but I suggest you just change this to `GlobalId` since it's a small `Copy` type

martykulma: done

martykulma:
Thanks for the reviews!

PR #32980, opened 2025-07-10
docs: Add Materialize DR characteristics v1

kay-kim:
For v1, leaving out specific callouts for Kafka and webhook sources; such as for recovery reliant upon retention policy of Kafka and webhooks SOL and for Kafka sources, a cluster repl factor > 1 can provide HA.


doc/user/content/manage/disaster-recovery/recovery-characteristics.md
```diff
@@ -0,0 +1,89 @@
+---
+title: "Materialize DR characteristics"
+description: "Reference on various failure modes and recovery characteristics of Materialize."
+menu:
+  main:
+    parent: "disaster-recovery"
+    weight: 15
+---
+
+The following provides  various failure mode impact and recovery
```
kay-kim, line 10: https://preview.materialize.com/materialize/32980/manage/disaster-recovery/recovery-characteristics/

kay-kim: If we don't like the layout, can easily change ... made the actual factoids into data files ... so that we can mix and match -- also, to keep the factoids bite sized since easy to lose sight of the component if one big file/table.

doc/user/data/disaster_recovery/cluster.yml
```diff
@@ -0,0 +1,65 @@
...
+      - If compute, rehydration time.
+      - If serving, rehydration time.
+
+      To reduce rehydration time, scale up the cluster.
+    RTO (RF2 - multiple AZs): |
+      Can be:
+
+      - 0 if only compute and serving objects are on the machine.
+
+      - Time to spin up new machine if sources or sinks are on the machine.
```
kay-kim, line 30: For kafka sources, I believe this isn't true anymore ... but, for v1, am not calling out kafka or webhook sources specifically, but lumping them.

doc/user/content/manage/disaster-recovery/recovery-characteristics.md
```diff
@@ -0,0 +1,89 @@
...
+  {{</ annotation >}}
+
+## System Environment
+
+{{< yaml-table data="disaster_recovery/system_environment_nested" >}}
+
+{{< annotation type="Key point(s)" >}}
+
+- The egress gateway runs on a single machine on a single Availability Zones
+  (AZ); other components run on multiple Availability Zones (AZs).
```
jubrad: This is definitely worth knowing for us, but I don't think this is something we want to call out in our docs. I would categorize some of this as implementation details customers shouldn't have to know. 

Also, because users don't have any control over these I might focus on the failure modes and impact than the underlying components. 

**Very loose overview**
Our system components are in charge of providing features like database consensus, durable storage, network access and security, and compute provisioning. These are deployed in highly available configurations or with automated recovery mechanisms, but can still suffer from outages due to the underlying providers.
- Single Availability Zone:
  - public source and sinkand single AZ privatelink connections issues
  - brief pgwire and https connection drops as network rebalances
 - 2 Availability Zones:
   - temporary issues with cluster provisioning and console access
 - 3+ Availability Zones:
   - limited to no database access
   - May require point in time recovery of environments
- Single Region System Resources:
  - There are metadata resources running in HA in us-east-1, an outage in us-east-1 may result in issues viewing the console for other regions, this does not effect database access, up-time or performance.
   
**Key Strategy:**
- Use privatelink when possible and configure to use multiple AZs.
- Consider deploying to a second region if you are concerned about multi-AZ outages.

doc/user/data/disaster_recovery/cluster.yml
```diff
@@ -0,0 +1,65 @@
...
+  - Failure Type: "**Single AZ failure**"
+    RPO: 0
+    RTO (RF1 - single AZ): |
+
+      *For managed clusters*
+
+      Time to spin up new machine + possible rehydration time, depending on the
+      objects on the machine:
+
+      - If sources, no additional time (i.e., does not require rehydration).
```
jubrad: I believe upsert sources do require rehydration.

There may also be a period with increased cluster usage/delays due to sources or sinks catching up... 
Scaling up the cluster may help here, but may also not be worth it.. since this should be small.

doc/user/data/disaster_recovery/egress_gateway.yml
```diff
@@ -0,0 +1,27 @@
...
+      manual failover to a new AZ(~30–60 min)
+    RTO (RF2 - multiple AZs): |
+      N/A
+
+  - Failure Type: "**Regional failure (or 2 AZ failures)**"
+    RPO: 0
+    RTO (RF1 - single AZ): |
+      High/Significant. Consider using a [regional failover strategy](/manage/disaster-recovery/#level-3-a-duplicate-materialize-environment-inter-region-resilience).
+    RTO (RF2 - multiple AZs): |
+      N/A
```
jubrad: I'm not sure that we should have this page

doc/user/data/disaster_recovery/system_environment_nested.yml
```diff
@@ -0,0 +1,38 @@
...
+      **Egress gateway**
+
+    "Failure Characteristics": |
+      The egress gateway runs on a single node in a single AZ. Although no data
+      is lost if the egress gateway becomes unavailable, sources and sinks are
+      impacted during this time.
+
+      The component has the following failure characteristics:
+
+      {{< yaml-table data="disaster_recovery/egress_gateway" >}}
```
jubrad: If we want to keep this I'd just call it networking andstate that public sources and sinks may lose connection, as well as single az privatelink connections.

doc/user/data/disaster_recovery/system_environment_nested.yml
```diff
@@ -0,0 +1,38 @@
...
+  - "Component": |
+      **Everything else**
+
+    "Failure Characteristics": |
+      The remaining components run across multiple AZs, providing high
+      availability; however, some failure events may impact the UI.
+
+      The remaining components have the following failure characteristics:
+
+      {{< yaml-table data="disaster_recovery/system_environment_everything_else" >}}
```
jubrad: Failures may also effect new cluster/replica provisioning.

doc/user/data/disaster_recovery/egress_gateway.yml
kay-kim: Removed the page

doc/user/data/disaster_recovery/system_environment_nested.yml
kay-kim: Removed

doc/user/data/disaster_recovery/system_environment_nested.yml
kay-kim: removed page

doc/user/data/disaster_recovery/cluster.yml
kay-kim: Updated.

doc/user/content/manage/disaster-recovery/recovery-characteristics.md
kay-kim: 👍 replaced the systems one with your copy.

doc/user/data/disaster_recovery/system_environment.yml
```diff
@@ -0,0 +1,27 @@
...
+      - Brief `pgwire` and `https` connection drops as network rebalances.
+
+  - Failure Type: "**Two Availability Zones**"
+    Impact: |
+      - Temporary issues with cluster provisioning.
+      - Temporary issues with Console access.
+
+  - Failure Type: "**Three or More Availability Zones**"
+    Impact: |
+      - Partial to no access to the database.
```
kay-kim, line 18: Wasn't sure if the "limited" in "limited to no database access" meant:
-  only database access is lost (but no data loss, etc) or
- partial loss

doc/user/data/disaster_recovery/system_environment.yml
```diff
@@ -0,0 +1,27 @@
+columns:
+  - column: "Failure Type"
+  - column: "Impact"
+rows:
+  - Failure Type: "**Single Availability Zone (AZ)**"
+    Impact: |
+      - Connection issues using single-AZ Privatelink and sources/sinks.
```
kay-kim, line 7: Wasn't quite sure how to parse "public source and sink and single AZ privatelink connections issues"

doc/user/data/disaster_recovery/human.yml
```diff
@@ -0,0 +1,23 @@
+columns:
+  - column: "Failure Type"
+  - column: "RPO"
+  - column: "RTO (RF1/RF2)"
+rows:
+  - Failure Type: "**Accidental source drop (and dependent objects)**"
+    RPO: |
+      100% data loss.
```
nstewart: RPO = Same as source system. Source will need to be recreated.

PR #32981, opened 2025-07-10
adapter: Add pgwire_recv_scheduling_delay_ms metric

ggevay:
With this metric, we'll get a more complete picture of query latency: now we'll also know about potential delays caused by overloaded tokio worker threads.

The interesting part is in `advance_ready`, the rest is just wiring.

### Motivation

### Checklist


ggevay:
> This is a bit clunky, with how the metrics interval get's passed into the instrumented task itself

Yeah, the instrumentation had to be inserted at the `spawn`, and then it had to be passed down the call chain. The `spawn` turned out to be a bit far away from the actual thing that we want to measure, so it's quite a bit of wiring.

> I'm curious how you inserted the sleep? Wouldn't you have to "insert" it in the time between calling `recv()` and it being polled?

I inserted it after `recv`. Since I had 100 threads/connections in my test app that was submitting the queries, the sleep caused all the tokio worker threads to be quite busy. This caused significant scheduling delays, measured by the metric. (There were some non-0 numbers even without the sleep, but much smaller, on the order of a few milliseconds.)

PR #32987, opened 2025-07-11
Add source table support for SQL server

martykulma:
Adds support for `CREATE TABLE .. FROM SOURCE` for SQL Server.

### Motivation

implements https://github.com/MaterializeInc/database-issues/issues/9202

### Checklist

### DIFF
--- a/src/sql-server-util/src/inspect.rs
+++ b/src/sql-server-util/src/inspect.rs
@@ -17,8 +17,8 @@ use smallvec::SmallVec;
 use std::collections::BTreeMap;
 use std::fmt;
 use std::sync::Arc;
-use tiberius::numeric::Numeric;
 use std::time::Duration;
+use tiberius::numeric::Numeric;
 
 use crate::cdc::{Lsn, RowFilterOption};
 use crate::desc::{SqlServerColumnRaw, SqlServerTableRaw};
--- a/src/sql-server-util/src/cdc.rs
+++ b/src/sql-server-util/src/cdc.rs
@@ -474,7 +474,7 @@ pub enum CdcEvent {
 #[derive(Debug, thiserror::Error)]
 pub enum CdcError {
     #[error(
-        "the requested LSN '{requested:?}' is less then the minimum '{minimum:?}' for `{capture_instance}'"
+        "the requested LSN '{requested:?}' is less than the minimum '{minimum:?}' for `{capture_instance}'"
     )]
     LsnNotAvailable {
         capture_instance: Arc<str>,
--- a/test/sql-server-cdc/40-data-types.td
+++ b/test/sql-server-cdc/40-data-types.td
@@ -118,7 +118,7 @@ ALTER SYSTEM SET log_filter = 'mz_storage::source::sql_server=debug,mz_sql_serve
     PASSWORD = SECRET sql_server_pass
   );
 
-# TODO - SQL server delay because it's terrible and the initial_lsn selection
+# TODO - SQL server delay because it's terrible for testing and the initial_lsn selection
 # may happen faster than CDC manages to process the above updates
 $ sleep-is-probably-flaky-i-have-justified-my-need-with-a-comment duration=5s
 


src/sql/src/pure.rs
```diff
@@ -1967,12 +1967,44 @@ async fn purify_create_table_from_source(
...
+                &retrieved_source_references,
+                &requested_references,
+                &qualified_text_columns,
+                &qualified_exclude_columns,
+                &unresolved_source_name,
+                &SourceReferencePolicy::Required,
+            )
+            .await?;
+
+            // There should be exactly one source_export returned for this statement
```
petrosagg, line 2027: You can use `source_exports.into_element()` to extract the first element and additionally assert that there isn't a second one

petrosagg: Just saw the idiom you used was prior art in the other cases. We could fix those too since we're here

src/sql/src/pure.rs
```diff
@@ -2136,11 +2168,53 @@ async fn purify_create_table_from_source(
...
+            let sql_server::SqlServerExportStatementValues {
+                columns: gen_columns,
+                constraints: gen_constraints,
+                text_columns: gen_text_columns,
+                excl_columns: gen_excl_columns,
+                details: gen_details,
+                external_reference: _,
+            } = sql_server::generate_source_export_statement_values(&scx, purified_export)?;
+
+            if let Some(text_cols_option) = with_options
```
petrosagg, line 2206: nit: if you have `if let Some(..) = opt_value` and you also include an `else` statement then it's more idiomatic to write a `match` statement with `Some(..)` and `None` branches

src/storage/src/source/sql_server/replication.rs
```diff
@@ -100,38 +100,34 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
-                        (output.partition_index, Arc::clone(&output.decoder)),
-                    )
-                })
-                .collect();
             let mut cdc_handle = client
                 .cdc(capture_instances.keys().cloned())
                 .max_lsn_wait(SNAPSHOT_MAX_LSN_WAIT.get(config.config.config_set()));
 
-            // Snapshot any instances that require it.
+            // Snapshot any instances that requires it.
```
petrosagg:
```suggestion
            // Snapshot any instance that requires it.
```

src/storage/src/source/sql_server/replication.rs
```diff
@@ -100,38 +100,34 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
+            // The LSN returned here will be the max LSN for any capture instance on the SQL server.
             let snapshot_lsn = {
```
petrosagg, line 133: If I'm reading the code correctly here we are computing a snapshot LSN that is common among all exports, and that snapshot LSN depends on the max LSN that we get from each capture instance, and capture instances are created on demand as the user requests more and more exports from the available ones.

If we put all these together I think we can end up in a situation where one replica that only knows about export `foo` computes some `snapshot_lsn1` but another replica, that say hears about two exports, `foo` and `bar` computes a higher `snapshot_lsn2 > snapshot_lsn1` due to the capture instance of `bar`.

Is this reasoning correct? If yes I think there is a possibility for double ingesting the snapshot

src/storage/src/source/sql_server/replication.rs
```diff
@@ -221,22 +222,44 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
+            // If a source is created with no subsources, there will be no inputs
+            // as each capture instance collects data for a table. Without a cdc stream
+            // to consume, this source will have no outputs either.
+            if resume_lsn.is_none() {
+                // TODO: this is only in place until we implement RLU
+                // https://github.com/MaterializeInc/database-issues/issues/9212
+                tracing::warn!(%config.id, "timely-{} no resume_lsn, waiting.", config.worker_id);
+                std::future::pending::<()>().await;
+            }
+            let resume_lsn = resume_lsn.unwrap();
```
petrosagg: nit: having this be a `let resume_lsn = match resume_lsn {...}` is more idiomatic than checking `is_none()` and then unwraping

src/storage/src/source/sql_server/replication.rs
```diff
@@ -221,22 +222,44 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
 
             // Start replicating from the LSN __after__ we took the snapshot.
             let replication_start_lsn = snapshot_lsn.increment();
+            let mut rewinds:BTreeMap<_,_> = export_ids_to_snapshot
```
petrosagg:
```suggestion
            let mut rewinds: BTreeMap<_, _> = export_ids_to_snapshot
```
rustfmt gives up in this file :(

src/storage/src/source/sql_server/replication.rs
```diff
@@ -349,7 +396,7 @@ type StackedAsyncOutputHandle<T, D> = AsyncOutputHandle<
 /// Helper method to return a "definite" error upstream.
 async fn return_definite_error(
     err: DefiniteError,
-    outputs: &[usize],
+    outputs: impl Iterator<Item = &u64>,
```
petrosagg: This function probably prefers just `u64` since it unconditionally dereferences. You can add a `.copied()` in the `.iter()` calls in the flatmap above to get a `u64` from a `&u64` iterator

src/storage/src/source/sql_server/replication.rs
martykulma: Great catch! I'll have to capture the lsn earlier on.  I think MySQL does this as well, should be able to follow the same pattern.

src/sql/src/pure.rs
martykulma: fixed!

src/sql/src/pure.rs
martykulma: This was also carrying forward existing patterns. I updated the lot.

src/storage/src/source/sql_server/replication.rs
martykulma: Now captures LSN at purification and passes it down for `resume_lsn` computation

src/storage/src/source/sql_server/replication.rs
martykulma: Since we should never continue beyond that point, I opted for `let .. else { ...; unreachable!() };` - but am happy to change it to `match` with `None => std::future::pending::<Lsn>().await` if you prefer.

src/storage/src/source/sql_server/replication.rs
martykulma: yeah it does 😂

src/storage/src/source/sql_server/replication.rs
martykulma: done

martykulma:
> It all looks great and most of my comments are nits. The only thing that I'm worried about is the selection of snapshot lsn

Thanks for the review @petrosagg! I've split review updates into separate commits. You can ignore the last couple commits, which were me rushing to fix typos and then trying to get CI to work.  The most meaningful change is in the `initial_lsn` commit.

I ran into errors where I was setting things up too quickly and there was no max LSN yet, and tests consistently required sleeps to avoid this. So I opted to include a short retry window in purification to get the LSN for a better experience, and that added a little bit of refactoring to centralize retries.

src/sql/src/pure.rs
```diff
@@ -997,6 +1000,12 @@ async fn purify_create_source(
                     replication_errors,
                 ))?;
             }
+            // If CDC is enabled for a table, there is a period where the max LSN will not be
+            // available.  Rather than return an error to the user, we retry for 5 seconds to
+            // allow the CDC job a chance to run.  By default, the job runs every 5 seconds.
+            let initial_lsn =
+                mz_sql_server_util::inspect::get_max_lsn_retry(&mut client, Duration::from_secs(5))
+                    .await?;
```
ptravers: nit: can we make the retry configurable? I guess it's pretty unlikely that someone will reconfigure the timing on the capture instance.

src/sql/src/pure.rs
```diff
@@ -1609,6 +1619,12 @@ async fn purify_alter_source_add_subsources(
                 )
                 .await?;
             let mut client = mz_sql_server_util::Client::connect(config).await?;
+            // If CDC is enabled for a table, there is a period where the max LSN will not be
+            // available.  Rather than return an error to the user, we retry for 5 seconds to
+            // allow the CDC job a chance to run.  By default, the job runs every 5 seconds.
+            let initial_lsn =
+                mz_sql_server_util::inspect::get_max_lsn_retry(&mut client, Duration::from_secs(5))
+                    .await?;
```
ptravers: nit: at a minimum would be nice to get the value we retry for into a shared constant so it's less of a magic number.

src/sql/src/pure.rs
martykulma: Yeah, good idea. I realized that we have a max lsn wait parameter already for snapshot, and there no reason I can think of why the timeouts should be different, so I've renamed it and we can use that.

src/sql-server-util/src/inspect.rs
```diff
@@ -34,6 +36,28 @@ pub async fn get_min_lsn(
...
+) -> Result<Lsn, SqlServerError> {
+    let (_client, lsn_result) = mz_ore::retry::Retry::default()
+        .max_duration(max_retry_duration)
+        .retry_async_with_state(client, |_, client| async {
+            let result = crate::inspect::get_min_lsn(client, capture_instance).await;
+            (client, map_null_lsn_to_retry(result))
+        })
+        .await;
+    let Ok(lsn) = lsn_result else {
+        tracing::warn!("database did not report a maximum LSN in time");
```
petrosagg:
```suggestion
        tracing::warn!("database did not report a minimum LSN in time");
```

src/storage/src/source/sql_server/replication.rs
```diff
@@ -100,38 +100,34 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
-                    }
-                })
-                .collect();
+            for output in outputs.values() {
+                if decoder_map.insert(output.partition_index, Arc::clone(&output.decoder)).is_some() {
+                    panic!("Multiple decoders for output index {}", output.partition_index);
+                }
+                capture_instances.entry(Arc::clone(&output.capture_instance))
+                    .and_modify(|export_ids| export_ids.push(output.partition_index))
+                    .or_insert_with(|| vec![output.partition_index]);
```
petrosagg: There is a cute API for this pattern

```suggestion
                capture_instances
                    .entry(Arc::clone(&output.capture_instance))
                    .or_default()
                    .push(output.partition_index);
```

`.or_default()` inserts the default value if it's not there and then returns a mutable reference to it so that you write your update logic once.

src/storage/src/source/sql_server/replication.rs
```diff
@@ -221,22 +222,51 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
 
             // Start replicating from the LSN __after__ we took the snapshot.
             let replication_start_lsn = snapshot_lsn.increment();
+            let mut rewinds: BTreeMap<_, _> = export_ids_to_snapshot
+                .values()
+                .flat_map(|export_ids| export_ids.iter().map(|idx| (*idx, replication_start_lsn)))
```
petrosagg: hm, I was expecting to see the initial_lsn mentioned here. Don't we want to rewind to that point instead of whatever lsn replication happens to start from?

src/storage/src/source/sql_server/replication.rs
```diff
@@ -221,22 +222,51 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
+                })
+                .min();
+
+            // If a source is created with no subsources, there will be no inputs
+            // as each capture instance collects data for a table. Without a cdc stream
+            // to consume, this source will have no outputs either.
+            let Some(resume_lsn) = resume_lsn else {
+                // TODO: this is only in place until we implement RLU
+                // https://github.com/MaterializeInc/database-issues/issues/9212
+                tracing::warn!(%config.id, "timely-{} no resume_lsn, waiting.", config.worker_id);
```
petrosagg:
```suggestion
                tracing::warn!(%config.id, "timely-{} no resume_lsn, waiting", config.worker_id);
```

src/storage/src/source/sql_server/replication.rs
petrosagg: Wait, I commented in the wrong place. The rewinds are correct since the LSN we put in there is the upper bound. But I think there is still an issue where we will rewind everything to the minimum `initial_lsn` of all tables instead of the `initial_lsn` that we recorded for that *particular* table.

def-:
I have rebased your PR to pick up some CI fixes

src/storage/src/source/sql_server/replication.rs
martykulma: Updated per our conversation.  I added an assertion that initial_lsn <= snapshot_lsn.  I was going to put it in `CdcStream::snapshot`, but that's going to conflict with changes in https://github.com/MaterializeInc/materialize/pull/32979. I added a task to move this before we copy all the data.

src/sql-server-util/src/inspect.rs
```diff
@@ -46,6 +70,37 @@ pub async fn get_max_lsn(client: &mut Client) -> Result<Lsn, SqlServerError> {
     parse_lsn(&result[..1])
 }
 
+/// Returns the maximum log sequence number for the entire database, retrying
```
petrosagg, line 78: Can we fix the doc here to mention the shortcoming of this method? Namely that the returned value will be less than the maximum LSN of the server since the value depends on the async CDC process. Maybe also link to this thread https://groups.google.com/g/debezium/c/47Yg2r166KM/m/lHqtRF2xAQAJ

petrosagg (approved):

src/sql-server-util/src/inspect.rs
martykulma: Fixed, thanks Petros!

PR #32991, opened 2025-07-11
Mention in testdrive.md that `--rewrite-results` is currently buggy

ggevay:
This is to prevent people from thinking that they are doing something wrong, see
https://materializeinc.slack.com/archives/C08AXPDBPH7/p1752158187078249
"Spent 30 minutes trying to figure out why --rewrite-results in testdrive wasn’t working. Maybe we should remove it from our testdrive readme?"

cc @SangJunBak 

(Also, fix some typos that were highlighted by my IDE.)

### Checklist

### DIFF
--- a/doc/developer/testdrive.md
+++ b/doc/developer/testdrive.md
@@ -211,11 +211,11 @@ Schema registry URL [default: http://localhost:8081]
 
 ## Resource sharing and reuse across tests
 
-By default, `testdrive` will clean up the environment and Mz prior to running each test. In certain situations, two or more testdrive invocations need to run consequtively while being able to operate on the same database objects. The options below help to ensure successful cooperation.
+By default, `testdrive` will clean up the environment and Mz prior to running each test. In certain situations, two or more testdrive invocations need to run consecutively while being able to operate on the same database objects. The options below help to ensure successful cooperation.
 
 #### `--no-reset`
 
-By default, `testdrive` will clean up its environment as much as it can before the start of the test, which includes Mz databases, etc. If two consequtive invocations of `testdrive` need to be able to operate on the same objects, e.g. database tables, the second invocation needs to run with `--no-reset`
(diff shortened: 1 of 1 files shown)


def- (approved):
I would have preferred to fix it, but haven't found the issue yet!

doc/developer/testdrive.md
```diff
@@ -243,6 +243,8 @@ Shuffle the list of tests before running them (using the value from --seed, if a
 
 #### `--rewrite-results`
 
+Note that this option is currently buggy!
```
SangJunBak, line 246: I think my complaints were:
1. in `statement_logging.td` where the results would eventually show up. So setting `--default-max-tries` to 1 would just have it rewrite as empty
2. `--default-max-tries` isn't documented in this doc or `mzcompose`. The way I found it / had to input is edit the `mzcompose` file manually. I think for most mzcompose testdrive workflows, it propagates args to testdrive, but not for the testdrive test in statement logging.

I didn't try with a different number of max tries (i.e. 3), but it might be nice to recommend a default

doc/developer/testdrive.md
```diff
@@ -243,6 +243,8 @@ Shuffle the list of tests before running them (using the value from --seed, if a
...
+
 Automatically rewrite the testdrive file with the correct results when they are not as expected. Consider setting a lower `--default-max-tries` value too to get a result faster.
```
SangJunBak, line 248: Given the points:

```suggestion
Note that this option is currently buggy!

Automatically rewrite the testdrive file with the correct results when they are not as expected. Consider setting a lower `--default-max-tries` value too to get a result faster.
```
```suggestion
Automatically rewrite the testdrive file with the correct results when they are not as expected. Consider setting  `--default-max-tries` to 3 in `testdrive` to get a result faster. Contrary to setting it to `1`, where the result may be empty
```

ggevay: The problem is that it will be different for every file what a good setting is for `--default-max-tries`. E.g., for `materialized-view-refresh-options.td` we often need to wait for several seconds for a refresh time to come, which would be many more tries than 3.

PR #32992, opened 2025-07-11
persist: add len field to RunMeta for tracking updates

DAlperin:
Towards incremental compaction

### Checklist


src/persist-client/src/internal/state.rs
```diff
@@ -800,6 +800,9 @@ pub struct RunMeta {
 
     /// If set, a UUID that uniquely identifies this run.
     pub(crate) id: Option<RunId>,
+
+    /// The number of updates in this run.
```
bkirwi:
```suggestion
    /// The number of updates in this run, or `None` if the number is unknown.
```

src/persist-client/src/batch.rs
```diff
@@ -862,7 +892,10 @@ impl<T: Timestamp + Codec64> BatchParts<T> {
                             1,
                             "compaction is guaranteed to emit a single run"
                         );
-                        output_batch.parts
+                        let total_compacted_updates: usize =
+                            runs.iter().filter_map(|(meta, _)| meta.len).sum();
```
bkirwi: Here you're using `meta.len` from above, but that's only set when incremental compacction is set, so this might calculate the wrong sum in some cases. Easiest is probably to track the total count separately!

PR #32996, opened 2025-07-11
persist: structurally diff state diffs to find removed parts

DAlperin:
Right now we only ever do full replacements of hollow batches during compaction but with incremental compaction we will want to know what specifically changed.

### Checklist


src/persist-client/src/internal/state_diff.rs
```diff
@@ -263,21 +263,41 @@ impl<T: Timestamp + Lattice + Codec64> StateDiff<T> {
...
-                Update(a, _) | Delete(a) => Some(HollowBlobRef::Batch(a)),
+                Update(a, _) | Delete(a) => Some(a.parts.iter().collect::<Vec<_>>()),
             });
-        let rollups = self
-            .rollups
+        let added = self
+            .referenced_batches()
+            .filter_map(|spine_diff| match spine_diff {
+                Insert(a) => Some(a.parts.iter().collect::<Vec<_>>()),
+                Update(_, a) => Some(a.parts.iter().collect::<Vec<_>>()),
```
bkirwi: Huh... does rust allow you to merge the update + delete case above but not the two cases here?

src/persist-client/src/internal/state_diff.rs
```diff
@@ -263,21 +263,41 @@ impl<T: Timestamp + Lattice + Codec64> StateDiff<T> {
...
+                Delete(_) => None,
+            })
+            .collect::<Vec<_>>();
```
bkirwi: Possible to use a set or similar here instead? I think this is quadratic as written, which we usually try to avoid in non-debug-only codepaths, since it tends to catch up to us eventually...

bkirwi (approved):
One hopefully-uncontroversial suggestion, but otherwise, great!

src/persist-client/src/internal/state_diff.rs
DAlperin: good call

PR #33018, opened 2025-07-14
Temporal bucket: pass through data close to frontier

antiguru:
Updates the temporal bucketing logic to only store data in the bucket chain that is likely not to be revealed soon, as far as it's possible to determine this locally. The condition we implement determines a upper frontier based on the as-of of the dataflow, the current input frontier, and a summary. It computes the join of the input frontier and the as-of, and applies the summary to the result. The summary is configurable and defaults to the equivalent of two seconds.

This results in a frontier that is always ahead of the input frontier, causing the operator to release and pass through a configurable amount of future updates, saving the cost associated with inserting and extracting the data from the bucket chain.

This addresses the performance concerns for dataflows without future retractions: most of the data will never enter the bucket chain. The initial snapshot is within the 2s summary of the as-of, so it'll be forwarded verbatim. Only if there are future updates, we'll store them in the bucket chain.

There is a new cost associated with inserting data into the bucket chain, but we hope to amortize this cost with the CPU savings associated with having fewer outstanding updates in the merge batcher. We reduce the cost from once every frontier tick to once for inserting and extracting the data in the bucket chain, plus the maintenance of data in the chain.

### DIFF
--- a/src/compute/src/extensions/temporal_bucket.rs
+++ b/src/compute/src/extensions/temporal_bucket.rs
@@ -22,7 +22,7 @@ use timely::dataflow::channels::pact::Pipeline;
 use timely::dataflow::operators::Operator;
 use timely::dataflow::{Scope, StreamCore};
 use timely::order::TotalOrder;
-use timely::progress::{Antichain, Timestamp};
+use timely::progress::{Antichain, PathSummary, Timestamp};
 use timely::{Data, PartialOrder};
 
 use crate::typedefs::MzData;
@@ -32,8 +32,12 @@ use crate::typedefs::MzData;
 /// at times between the last input frontier and the current input frontier.
 pub trait TemporalBucketing<G: Scope, O> {
     /// Construct a new stream that stores updates into a [`BucketChain`] and reveals data
-    /// not in advance of the frontier.
-    fn bucket<CB>(&self) -> StreamCore<G, CB::Container>
+    /// not in advance of the frontier. Data that is `threshold` distance within the frontier
+    /// is passed through without being stored in the chain.
+    fn bucket<CB>(
+        &self,
+        threshold: <G::Timestamp as Timestamp>::Summary,
+    ) -> StreamCore<G, CB::Container>
     where
         CB: ContainerBuilder + PushInto<O>;
 }
@@ -46,7 +50,10 @@ where
     G::Timestamp: Data + MzData + BucketTimestamp + TotalOrder,
     D: Data + MzData + Ord + std::fmt::Debug,
 {
-    fn bucket<CB>(&self) -> StreamCore<G, CB::Container>
+    fn bucket<CB>(
+        &self,
+        threshold: <G::Timestamp as Timestamp>::Summary,
+    ) -> StreamCore<G, CB::Container>
     where
         CB: ContainerBuilder + PushInto<(D, G::Timestamp, mz_repr::Diff)>,
     {
@@ -60,7 +67,16 @@ where
             let mut cap = Some(cap);
             let mut buffer = Vec::new();
             move |input, output| {
-                while let Some((_time, data)) = input.next() {
+                let threshold: Antichain<_> = input
+                    .frontier()
+                    .frontier()
+                    .iter()
+                    .flat_map(|f| threshold.results_in(f)).collect();
+                while let Some((time, data)) = input.next() {
+                    // Skip data that is about to be revealed.
+                    let extracted = data.extract_if(.., |(_, t, _)| !threshold.less_than(t));
+                    output.session_with_builder(&time).give_iterator(extracted);
+
                     // Sort data by time, then drain it into a buffer that contains data for a
                     // single bucket.
                     data.sort_by(|(_, t, _), (_, t2, _)| t.cmp(t2));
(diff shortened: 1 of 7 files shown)


src/compute/src/render.rs
```diff
@@ -462,7 +462,7 @@ pub fn build_compute_dataflow<A: Allocate>(
                 for (id, (oks, errs)) in imported_sources.into_iter() {
                     let oks = if ENABLE_TEMPORAL_BUCKETING.get(&compute_state.worker_config) {
                         oks.inner
-                            .bucket::<CapacityContainerBuilder<_>>()
+                            .bucket::<CapacityContainerBuilder<_>>(2000.into())
```
petrosagg: Should this be a dyncfg?

src/compute/src/extensions/temporal_bucket.rs
```diff
@@ -60,7 +67,16 @@ where
...
             move |input, output| {
-                while let Some((_time, data)) = input.next() {
+                let threshold: Antichain<_> = input
+                    .frontier()
+                    .frontier()
+                    .iter()
+                    .flat_map(|f| threshold.results_in(f)).collect();
+                while let Some((time, data)) = input.next() {
+                    // Skip data that is about to be revealed.
+                    let extracted = data.extract_if(.., |(_, t, _)| !threshold.less_than(t));
```
petrosagg: By using `less_than` we are making it impossible to express the previous behavior. I think this should be `less_equal` so that when supplying the boundary value `T::Summary::default()` we don't extract anything and we send everything in the chain.

src/compute/src/render.rs
antiguru: Fixed

src/compute/src/extensions/temporal_bucket.rs
antiguru: Updated to use `less_equal`.

antiguru:
> My fear is that this PR will fix the current benchmarks just because we don't exercise this case. How do we know we don't have a general regression?

It's a trade-off. At the moment, we store all future updates in the merge batcher, which incurs a linear cost on _every frontier tick_. Moving the data into a bucket introduces a cost to insert and maintain the outstanding updates, but the cost should amortize over time, reducing the cost to maintain the merge batcher.

antiguru:
New feature benchmark run: https://buildkite.com/materialize/nightly/builds/12656#0198369b-2a2b-4a7b-b1bd-20b41fd9ca6e

petrosagg:
> At the moment, we store all future updates in the merge batcher, which incurs a linear cost on every frontier tick. Moving the data into a bucket introduces a cost to insert and maintain the outstanding updates, but the cost should amortize over time, reducing the cost to maintain the merge batcher.

I understand that, but we're also a bit blind to whether we are actually doing better than the linear cost here, right? In other words we don't have a benchmark that has a temporal filter pushing many updates into the future that we can compare against

src/compute/src/extensions/temporal_bucket.rs
```diff
@@ -56,46 +66,68 @@ where
...
+
             // Cap tracking the lower bound of potentially outstanding data.
             let mut cap = Some(cap);
+
+            // Buffer for data to be inserted into the chain.
             let mut buffer = Vec::new();
+
             move |input, output| {
-                while let Some((_time, data)) = input.next() {
+                // The upper frontier is the meet of the input frontier and the `as_of` frontier,
```
petrosagg:
```suggestion
                // The upper frontier is the join of the input frontier and the `as_of` frontier,
```

src/compute/src/extensions/temporal_bucket.rs
```diff
@@ -56,46 +66,68 @@ where
...
+                    let pass_through = data.extract_if(.., |(_, t, _)| !upper.less_equal(t));
+                    output
+                        .session_with_builder(&time)
+                        .give_iterator(pass_through);
+
                     // Sort data by time, then drain it into a buffer that contains data for a
-                    // single bucket.
+                    // single bucket. We scan the data for ranges of time that fall into the same
+                    // bucket so we can push batches of data at once.
                     data.sort_by(|(_, t, _), (_, t2, _)| t.cmp(t2));
```
petrosagg: We don't care about stability here, right? We could use`sort_by_unstable` which may be faster

src/compute/src/extensions/temporal_bucket.rs
```diff
@@ -56,46 +66,68 @@ where
...
-                                let wrapper = chain.find_mut(start).expect("Must exist");
-                                wrapper.inner.push_container(&mut buffer);
-                                buffer.clear();
+                    let mut drain = data.drain(..);
+                    if let Some((datum, time, diff)) = drain.next() {
+                        let mut range = chain.range_of(&time).expect("Must exist");
+                        buffer.push((datum, time, diff));
+                        for (datum, time, diff) in drain {
+                            // If we have a range, check if the time is not within it.
+                            if time < range.0 || time >= range.1 {
```
petrosagg: A nice thing you can do for the `range_of` API is to return an actual Range object (it's in std) and here do `range.contains(&time)`

antiguru: Oh, neat. Switched to using range instead of a tuple.

src/compute/src/extensions/temporal_bucket.rs
antiguru: :100:

src/compute/src/extensions/temporal_bucket.rs
antiguru: Oops :D

antiguru:
Thanks for the reviews!

PR #33031, opened 2025-07-15
tests: fix multiple statistics.td files to be multi-replica aware

aljoscha:
Fixes https://github.com/MaterializeInc/database-issues/issues/9469

It's quite a bit more verbose now, but I think more correct, and I got to remove some of the `SUM`, which were a bit of a smell, I feel like.

### DIFF
--- a/test/mysql-cdc-old-syntax/statistics.td
+++ b/test/mysql-cdc-old-syntax/statistics.td
@@ -41,25 +41,56 @@ INSERT INTO t1 VALUES ('one');
 > SELECT COUNT(*) > 0 FROM t1;
 true
 
+# NOTE: we make sure that we have stats for a replica, otherwise the
+# set-from-sql below might fail because it doesn't do retries when a row is
+# missing.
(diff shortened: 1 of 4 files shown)


test/mysql-cdc-old-syntax/statistics.td
```diff
@@ -41,49 +41,78 @@ INSERT INTO t1 VALUES ('one');
 > SELECT COUNT(*) > 0 FROM t1;
 true
 
+# NOTE: we make sure that we have stats for a replica, otherwise the
```
aljoscha, line 44: @def- we have these everywhere now because I think `set-from-sql` doesn't do retries, at least I saw it fail quickly when there was no row and the source stats are only scraped at an interval so sometimes (at the beginning) this query has an empty result.

def-: Yes, how would `set-from-sql` know what the result should be? If you know that there should be at least one row then you could run a `> SELECT COUNT(*) > 0 ...` before the `$ set-from-sql`. Edit: I see, you're doing something similar.

PR #32955, opened 2025-07-08
[persist] Add a HollowBatch::push_run method and convert usages

bkirwi:
### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #33044, opened 2025-07-15
Persist: Add new API for run based spine replacement

DAlperin:
### Checklist

### DIFF
--- a/src/persist-client/src/internal/trace.rs
+++ b/src/persist-client/src/internal/trace.rs
@@ -52,11 +52,12 @@ use differential_dataflow::difference::Semigroup;
 use mz_persist::metrics::ColumnarMetrics;
 use mz_persist_types::Codec64;
 use std::cmp::Ordering;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::Debug;
 use std::mem;
 use std::ops::Range;
 use std::sync::Arc;
+use tracing::error;
 
 use crate::internal::paths::WriterKey;
 use differential_dataflow::lattice::Lattice;
@@ -69,7 +70,7 @@ use timely::PartialOrder;
 use timely::progress::frontier::AntichainRef;
 use timely::progress::{Antichain, Timestamp};
 
-use crate::internal::state::HollowBatch;
+use crate::internal::state::{HollowBatch, RunId};
 
 use super::state::RunPart;
 
@@ -83,6 +84,8 @@ pub struct FueledMergeReq<T> {
 #[derive(Debug)]
 pub struct FueledMergeRes<T> {
     pub output: HollowBatch<T>,
+    pub input: CompactionInput,
+    pub new_active_compaction: Option<ActiveCompaction>,
 }
 
 /// An append-only collection of compactable update batches.
@@ -651,6 +654,21 @@ impl<T: Timestamp + Lattice> Trace<T> {
 }
 
 impl<T: Timestamp + Lattice + Codec64> Trace<T> {
+    pub fn apply_merge_res_checked_classic<D: Codec64 + Semigroup + PartialEq>(
+        &mut self,
+        res: &FueledMergeRes<T>,
+        metrics: &ColumnarMetrics,
+    ) -> ApplyMergeResult {
+        for batch in self.spine.spine_batches_mut().rev() {
+            let result = batch.maybe_replace_checked_classic::<D>(res, metrics);
+            if result.matched() {
+                return result;
+            }
+        }
+        ApplyMergeResult::NotAppliedNoMatch
+    }
+
+    #[allow(dead_code)]
     pub fn apply_merge_res_checked<D: Codec64 + Semigroup + PartialEq>(
         &mut self,
         res: &FueledMergeRes<T>,
@@ -675,9 +693,9 @@ impl<T: Timestamp + Lattice + Codec64> Trace<T> {
         ApplyMergeResult::NotAppliedNoMatch
     }
 
-    pub fn apply_tombstone_merge(&mut self, res: &FueledMergeRes<T>) -> ApplyMergeResult {
+    pub fn apply_tombstone_merge(&mut self, desc: &Description<T>) -> ApplyMergeResult {
         for batch in self.spine.spine_batches_mut().rev() {
-            let result = batch.maybe_replace_with_tombstone(res);
+            let result = batch.maybe_replace_with_tombstone(desc);
             if result.matched() {
                 return result;
             }
@@ -695,7 +713,20 @@ enum SpineLog<'a, T> {
     Disabled,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum CompactionInput {
+    /// We don't know what our inputs were; this should only be used for
+    /// unchecked legacy replacements.
+    Legacy,
+    /// This compaction output is a total replacement for all batches in this id range.
+    #[allow(dead_code)]
+    IdRange(SpineId),
+    /// This compaction output replaces the specified runs in this id range.
+    #[allow(dead_code)]
+    PartialBatch(SpineId, BTreeSet<RunId>),
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct SpineId(pub usize, pub usize);
 
 impl Serialize for SpineId {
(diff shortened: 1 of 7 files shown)


src/persist-client/src/internal/trace.rs
```diff
@@ -695,7 +714,17 @@ enum SpineLog<'a, T> {
...
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+/// A RunLocation uniquely identifies a run within a hollow batch.
+/// It is a pair of `SpineId` and a RunId, where the `SpineId` identifies the
+/// hollow batch and the `RunId` identifies the run within that batch.
+/// The `RunId` can be `None` when referencing a hollow batch that was
+/// written with a version of persist that did not write run ids.
+/// We generally intepret a `RunLocation` without a `RunId` as
+/// referencing the entire hollow batch, which may contain multiple runs.
```
bkirwi: This says that `None` can refer to both a specific run without an assigned run id, or all the runs in a particular batch. We should pick just one meaning instead of leaving this ambiguous!

src/persist-client/src/internal/trace.rs
```diff
@@ -918,6 +947,34 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
+    fn diffs_sum_for_runs<D: Semigroup + Codec64>(
+        batch: &HollowBatch<T>,
+        run_ids: &[RunId],
+        metrics: &ColumnarMetrics,
+    ) -> Option<D> {
+        if run_ids.is_empty() {
+            return None;
+        }
+
+        let mut parts = Vec::new();
```
bkirwi: Nit, but you don't need the allocation here... we could filter down `batch.runs` and pass that iterator directly to `diffs_sum`.

src/persist-client/src/internal/trace.rs
```diff
@@ -935,15 +992,318 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
+        };
+        let num_batches_after = self.parts.len();
+        assert!(
+            num_batches_after <= num_batches,
+            "replacing parts should not increase the number of batches"
+        );
+        res
+    }
+
     // TODO: Roundtrip the SpineId through FueledMergeReq/FueledMergeRes?
```
bkirwi: I think this counts as done!

src/persist-client/src/internal/trace.rs
```diff
@@ -935,15 +992,318 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
 
         if let Some((id, range)) = self.find_replacement_range(res) {
-            self.perform_subset_replacement(res, id, range)
+            self.perform_subset_replacement(&res.output, id, range, None)
         } else {
             ApplyMergeResult::NotAppliedNoMatch
         }
     }
 
+    #[allow(dead_code)]
```
bkirwi, line 992: It's not really dead!

src/persist-client/src/internal/trace.rs
```diff
@@ -935,15 +992,318 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
+            // We only need to replace a single part. Here we still care about the run_indices
+            // because we only want to replace the runs that are in the merge result.
+            let batch = &self.parts[range[0]];
+            let batch = &batch.batch;
+            let run_ids = inputs
+                .values()
+                .next()
+                .unwrap()
+                .iter()
+                .filter_map(|id| id.1)
```
bkirwi: This means if I submit a `FueledMergeRes` with inputs `[RunLocation(a, None), RunLocation(a, Some(b))`, I will only attempt to replace `b` and ignore the first input. Which seems wrong I think?

(I think this is where `RunLocation` ends up biting us... it's ambiguous and allows the client to specify update sets that are impossible or nonsensical.)

bkirwi: Yeah, if you update the proptest to include batches with only some runs that have run ids, this panics!

src/persist-client/src/internal/trace.rs
```diff
@@ -935,15 +992,318 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
+            replacement.run_meta.len() <= 1,
+            "replacement must have exactly 0 or 1 runs"
+        );
+
+        let run_ids = run_ids.to_vec();
+
+        let start_id = run_ids[0];
+        let end_id = *run_ids.last().unwrap();
+
+        // 0. Find the indices of the runs in the original batch.
```
bkirwi: I've suggested the possibility before, but now I'm sure of it: this logic is too complex for what is effectively a batch build over `original.runs().filter(|(meta, _)| !run_ids.contains(id).chain(replacement.runs())`. I encourage you to simplify it! Using the higher-level methods like `.runs` instead of a bunch of vec indexing ought to help, or https://github.com/MaterializeInc/materialize/pull/32955 if you want to take a look at it.

(It also imposes a bunch of restrictions that aren't documented but would be fine if they were necessary (like the single-run restriction on our input, or that runs are contiguous) but AFAICT they're only there to not make the logic below more complicated, and I think you can probably get rid of them with a more straightforward impl.)

src/persist-client/src/internal/trace.rs
bkirwi: After reviewing the whole thing, I think you'd be better served with something like:

```
pub enum CompactionInput {
    /// This compaction output is a total replacement for all batches in this id range.
    IdRange(SpineId),
    /// This compaction output replaces the specified runs in this id range.
    PartialBatch(SpineId, BTreeSet<RunId>),
}
```

Instead of `Vec<RunLocation>`. Quite a lot of this PR is validating the input representation makes sense, but since you're introducing the input specification in the same PR, we can just pick one that doesn't allow invalid states in the first place.

src/persist-client/src/internal/trace.rs
DAlperin: > or that runs are contiguous

I had a very stupid idea in my head that this was important until you made me think about it for a second and realize it wasn't. This is why I was so stuck on this function. Sorry about that.

src/persist-client/src/internal/trace.rs
```diff
@@ -695,7 +713,20 @@ enum SpineLog<'a, T> {
     Disabled,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum CompactionInput {
+    /// We don't know what our inputs where, this should only be used for
```
bkirwi:
```suggestion
    /// We don't know what our inputs were; this should only be used for
```

src/persist-client/src/internal/trace.rs
```diff
@@ -695,7 +713,20 @@ enum SpineLog<'a, T> {
...
+    /// unchecked legacy replacements.
+    Legacy,
+    /// This compaction output is a total replacement for all batches in this id range.
+    #[allow(dead_code)]
+    IdRange(BTreeSet<SpineId>),
```
bkirwi: This could just be a `SpineId` instead of a set, since a single id is already a range. (Optional, but would save you some validation below.)

src/persist-client/src/internal/trace.rs
```diff
@@ -958,6 +1066,182 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
+        };
+        let replacement_range = i..i + 1;
+
+        let batch = &batch.batch;
+        let run_ids = runs.iter().cloned().collect::<Vec<_>>();
+
+        let old_batch_diff_sum = Self::diffs_sum::<D>(batch.parts.iter(), metrics);
+        let old_diffs_sum = Self::diffs_sum_for_runs::<D>(batch, &run_ids, metrics);
+
+        if let (Some(old_diffs_sum), Some(new_diffs_sum)) = (old_diffs_sum, new_diffs_sum) {
```
bkirwi: We should also assert that the `old_diffs_sum` is `None` when the `new_diffs_some` is `None`... for new diffs `None` should always mean 0, and it would be bad if we let that replace data with nonzero sum!

src/persist-client/src/internal/trace.rs
```diff
@@ -958,6 +1066,182 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
+    {
+        if runs.is_empty() {
+            return ApplyMergeResult::NotAppliedNoMatch;
+        }
+
+        let part = self.parts.iter().enumerate().find(|(_, p)| p.id == id);
+        let Some((i, batch)) = part else {
+            return ApplyMergeResult::NotAppliedNoMatch;
+        };
+        let replacement_range = i..i + 1;
```
bkirwi:
```suggestion
        let replacement_range = i..(i + 1);
```

src/persist-client/src/internal/trace.rs
```diff
@@ -958,6 +1066,182 @@ impl<T: Timestamp + Lattice + Codec64> SpineBatch<T> {
...
+            }
+            assert_eq!(
+                old_diffs_sum, new_diffs_sum,
+                "merge res diffs sum ({:?}) did not match spine batch diffs sum ({:?})",
+                new_diffs_sum, old_diffs_sum
+            );
+        }
+
+        let parts = &self.parts[replacement_range.clone()];
+        let id = SpineId(parts.first().unwrap().id.0, parts.last().unwrap().id.1);
```
bkirwi: This is oddly baroque... we're only targeting a single batch. Why a single-element range that we then unwrap?

src/persist-client/src/internal/trace.rs
```diff
@@ -2185,4 +2486,38 @@ pub(crate) mod tests {
...
     }
+
+    #[mz_ore::test]
+    fn construct_batch_with_runs_replaced_test() {
+        proptest!(|(
+            (batch, to_replace, range) in (any_hollow_batch::<u64>(), any_hollow_batch_with_exact_runs::<u64>(1))
+            .prop_filter("non-empty batch", |(batch, _)| batch.run_meta.len() >= 1)
+            .prop_flat_map(|(batch, to_replace)| {
+                let batch_len = batch.run_meta.len();
+                (0..batch_len).prop_flat_map(move |start| {
```
bkirwi: Nit: I'd extract this gen out to a `let` above so you get proper formatting.

src/persist-client/src/internal/trace.rs
```diff
@@ -2185,4 +2486,38 @@ pub(crate) mod tests {
...
+                let max_len = batch_len - start;
+                let batch_clone = batch.clone();
+                let to_replace_clone = to_replace.clone();
+                (1..=max_len).prop_map(move |len| {
+                    let range = start..(start + len);
+                    (batch_clone.clone(), to_replace_clone.clone(), range)
+                })
```
bkirwi, line 2489: This test still bakes in the "must replace a contiguous range of runs from the original batch" idea that the function under test no longer requires. Might as well choose a random subset instead and get better coverage?

bkirwi (approved):
Good changes! If nightlies are happy then so am I.

bkirwi:
Looks like you need to opt out the new test from `miri`... lots of similar examples in our existing proptests.

I find the scalability benchmark result sort of implausible, given the limited stuff you've changed... maybe spurious?

PR #33041, opened 2025-07-15
docs: add mz_now to the idiomatic mzsql section (part1)

kay-kim:
Initial content.  Will polish up the idiomatic section for better maintainability, etc. as a future todo item.

PR #28715, opened 2024-08-02
Optimizer fast path

ggevay:
**Edit: I've done some benchmarking in release mode now, and the performance improvement in release mode is quite small (see [here](https://materializeinc.slack.com/archives/C0761MZ3QD9/p1722630312570519?thread_ts=1722544237.348389&cid=C0761MZ3QD9)), so I'm now not sure whether the complexity that this PR is introducing is actually worth it. We might want to hold off on merging this, so taking back to draft.**

**Edit 2: Now with @antiguru's other improvements, the speedup from this might be worth it. So marking ready for review.**

This PR is about running only a tiny subset of the optimizer for fast path queries. The approach is that when we are about to call `optimize_dataflow`, we quickly check whether we already have a simple enough plan that it would go to the fast path. If yes, then we call a special, tiny optimizer instead of the whole thing.

Commits:
1. Trivial comment fixes.
2. This is not strictly necessary, but makes the optimizer more robust. See commit msg.
3. The main point of the PR.
4. At the point of the peek optimization where I added the new call to `create_fast_path_plan`, we haven't yet eliminated `mz_now()`, and `create_fast_path_plan` might choke on that. This commit makes us ignore this error and just not take the fast path optimizer.

There is a tricky reason for the change in `src/environmentd/tests/testdata/http/ws`, I'm investigating the details now. It's probably not an issue, since the actual fast path plan that we arrive at is the same, just the non-fast-path plan is slightly different, but even there the physical plan would be the same, the diff is only in the MIR plan. Edit: Figured out exactly what's happening: The reason for the diff is that in the normal optimizer pipeline we call `ProjectionPushdown`, which puts the MFP in a non-canonical state (`Project` at the bottom), which is usually canonicalized by `CanonicalizeMfp`, but here `CanonicalizeMfp` is not able to put back the `Project` at the top for this query, because it calls `extract_non_errors_from_expr_mut`, which gives up when it sees the error in this MFP. But this doesn't matter because `create_fast_path_plan` then calls a different MFP extraction method: `extract_from_expression`, so the fast path plan will have the same MFP as before. (And `ProjectionPushdown` is not doing anything useful here, because MFP optimization will take care of the projections.)

### Motivation

https://materializeinc.slack.com/archives/C0761MZ3QD9/p1722545887422389?thread_ts=1722544237.348389&cid=C0761MZ3QD9

[...]

PR #33047, opened 2025-07-16
Fix some bugs around temporal expressions and add tests

ggevay:
In one-off queries, there is no reason to restrict temporal predicates, because we just substitute the current time. (except for those mz_now() restrictions that are due to not having many operations available on `mz_timestamp`) However, @kay-kim [found](https://github.com/MaterializeInc/materialize/pull/33041#discussion_r2208299816) that we are currently rejecting the same temporal predicates for one-off queries that we reject in maintained queries. Turns out that https://github.com/MaterializeInc/materialize/pull/28715 broke this. The first commit of the PR fixes this.

It also turns out that our testing coverage for temporal predicate planning is abysmal. The 2. and 5. commits address this.

The third commit fixes an error msg: Unsupported temporal predicates were being printed as internal errors.

The fourth commit is just a precaution: It makes it more clear that `UnsafeMfpPlan` is an internal error, so that somebody doesn't accidentally write code that throws this in a place from where it would bubble up to the user.

### Motivation

### Checklist


src/adapter/src/coord/peek.rs
```diff
@@ -399,7 +399,7 @@ fn mfp_to_safe_plan(
     mfp: mz_expr::MapFilterProject,
 ) -> Result<mz_expr::SafeMfpPlan, OptimizerError> {
     mfp.into_plan()
-        .map_err(OptimizerError::Internal)?
+        .map_err(|_e| OptimizerError::UnsafeMfpPlan)?
```
aljoscha: How do we know we can always statically map to this error?

ggevay: This error is just intended to be the error that `mfp_to_safe_plan` returns if something goes wrong. So, by definition, anything that goes wrong here is this error.

But I've now added a String reason to it, see that we can see what went wrong if this error ever shows up.

PR #33049, opened 2025-07-16
SLT: Parallelize on each CI agent, remove some slow and useless junit uploads

def-:
Test pipeline: [8 agents took 9 min](https://buildkite.com/materialize/test/builds/106318#01981387-f0e4-4fbc-9b73-a4d477b97716) -> [3 agents take 5 min](https://buildkite.com/materialize/test/builds/106329#01981421-2250-4631-a39e-44bf62f62f01)
Nightly pipeline: [10 agents took 1h 26 min](https://buildkite.com/materialize/nightly/builds/12618#0198106c-e6a8-45b6-a176-346195ec1afe) -> [4 agents take 48 min](https://buildkite.com/materialize/nightly/builds/12628#019813d6-9302-4fc8-af3a-c92a3de23c23)
Proof that this works fine when a test fails: https://buildkite.com/materialize/test/builds/106368
Screenshot 2025-07-17 at 12 57 28

### Checklist

### DIFF
--- a/misc/python/materialize/mzcompose/services/sql_logic_test.py
+++ b/misc/python/materialize/mzcompose/services/sql_logic_test.py
@@ -14,7 +14,6 @@ from materialize.mzcompose import (
 from materialize.mzcompose.service import (
     Service,
 )
-from materialize.mzcompose.services.postgres import METADATA_STORE
 
 
 class SqlLogicTest(Service):
@@ -22,12 +21,13 @@ class SqlLogicTest(Service):
         self,
         name: str = "sqllogictest",
         mzbuild: str = "sqllogictest",
-        environment: list[str] = [
-            "MZ_SOFT_ASSERTIONS=1",
-        ],
+        environment: list[str] | None = None,
         volumes: list[str] = ["../..:/workdir"],
-        depends_on: list[str] = [METADATA_STORE],
     ) -> None:
+        if environment is None:
+            environment = [
+                "MZ_SOFT_ASSERTIONS=1",
+            ]
         environment += [
             "MZ_SYSTEM_PARAMETER_DEFAULT="
             + ";".join(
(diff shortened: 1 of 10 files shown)


misc/python/materialize/mzcompose/services/sql_logic_test.py
```diff
@@ -14,20 +14,20 @@
...
-            "MZ_SOFT_ASSERTIONS=1",
-        ],
+        environment: list[str] | None = None,
         volumes: list[str] = ["../..:/workdir"],
-        depends_on: list[str] = [METADATA_STORE],
     ) -> None:
+        if not environment:
+            environment = [
+                "MZ_SOFT_ASSERTIONS=1",
+            ]
```
ptravers: nit: I might be missing something but is it worth just leaving the environment default as is we are going to default it to that value for the new default value? totally possibly I'm just not getting how None works.

def-: This was actually a footgun/bug. In Python having a list as a default parameter is very dangerous, because if you modify it inside the function, you also modify the default for future calls to it.

misc/python/materialize/mzcompose/services/sql_logic_test.py
```diff
@@ -14,20 +14,20 @@
...
         name: str = "sqllogictest",
         mzbuild: str = "sqllogictest",
-        environment: list[str] = [
-            "MZ_SOFT_ASSERTIONS=1",
-        ],
+        environment: list[str] | None = None,
         volumes: list[str] = ["../..:/workdir"],
-        depends_on: list[str] = [METADATA_STORE],
     ) -> None:
+        if not environment:
```
teskje: Shouldn't this be `if environment is not None:`? Otherwise this will also evaluate as true when you pass an empty list.

teskje (approved):

PR #33052, opened 2025-07-16
storage/upsert: use RocksDB in memory

teskje:
This PR changes the upsert operator to use a RocksDB MemEnv when no file system is available, instead of falling back to `InMemoryHashMap`. Testing suggests that this results in significantly reduced memory usage, due to RocksDB's compression support.

This vastly improves the performance of upsert sources on replicas that don't get a scratch fs, like swap replicas. Based on our limited experience so far, upsert sources using MemEnv on swap replicas have roughly the same resource requirements as upsert sources using disk directly. [See Slack thread.](https://materializeinc.slack.com/archives/C08A7H11KP0/p1752782229619859)

### Motivation

Closes https://github.com/MaterializeInc/database-issues/issues/9480

### Tips for reviewer

The diff has a bunch of indentation changes, much easier to review with whitespace disabled.

The `InMemoryHashMap` backend is still used in tests so I didn't attempt to remove it. There is an opportunity for a larger refactor here where we remove the `UpsertStateBackend` abstraction, given that there is only a single used backend now.

### Checklist

### DIFF
--- a/src/storage/src/storage_state.rs
+++ b/src/storage/src/storage_state.rs
@@ -398,20 +398,18 @@ impl StorageInstanceContext {
         scratch_directory: Option<PathBuf>,
         cluster_memory_limit: Option<usize>,
     ) -> Result<Self, anyhow::Error> {
+        // If no file system is available, fall back to running RocksDB in memory.
+        let rocksdb_env = if scratch_directory.is_some() {
+            rocksdb::Env::new()?
+        } else {
+            rocksdb::Env::mem_env()?
+        };
+
         Ok(Self {
             scratch_directory,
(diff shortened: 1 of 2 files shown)


teskje:
It's not an unreasonable point you're making!

teskje:
Oh... I think auto spill is the default, so if we disable auto spill when there is no file system, we'll always use the hash map instead of MemEnv.

teskje:
I think this is rfal! @martykulma's concern has been addressed by removing the autospill backend, and all the nightly failures also appear on other branches.

teskje:
TFTR!

PR #33062, opened 2025-07-17
[persist] Run fewer background threads in tests

bkirwi:
### Motivation

We believe this is a contributing factor to https://github.com/MaterializeInc/database-issues/issues/9472, by way of https://sourceware.org/bugzilla/show_bug.cgi?id=19951.

Based on that upstream report, we believe that spawning fewer threads will make this issue less likely - both because having fewer threads means fewer chances to observe the issue, and because we may be less likely to fill the cache and trigger the issue in the first place.

It's also unnecessary to spawn so many threads in a test, especially since nextest will generally run many tests concurrently...

### Tips for reviewer

We also discussed adding a timeout to shutdown, but it turns out that the timeout only applies to the blocking thread pool, and the isolated runtime interface doesn't allow spawning blocking threads... so we wouldn't expect this config to have any effect here unfortunately.

https://github.com/tokio-rs/tokio/blob/0a3fe460862720d005b8fa39cdde30575a063785/tokio/src/runtime/runtime.rs#L430-L432

### Checklist

### DIFF
--- a/src/persist-cli/src/maelstrom/txn_list_append_multi.rs
+++ b/src/persist-cli/src/maelstrom/txn_list_append_multi.rs
@@ -444,7 +444,7 @@ impl Service for TransactorService {
             Arc::new(UnreliableConsensus::new(consensus, unreliable));
 
         // Wire up the TransactorService.
-        let isolated_runtime = Arc::new(IsolatedRuntime::default());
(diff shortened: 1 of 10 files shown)


def- (approved):

def-:
Can you also remove the ignores in the relevant persist cargo tests? I just disabled them in https://github.com/MaterializeInc/materialize/pull/33051

PR #33072, opened 2025-07-18
sql_server: add preventing excluding all columns in CREATE SOURCE

ptravers:
https://github.com/MaterializeInc/database-issues/issues/9209

### Checklist

### DIFF
--- a/src/sql-server-util/src/desc.rs
+++ b/src/sql-server-util/src/desc.rs
@@ -238,7 +238,7 @@ impl SqlServerColumnDesc {
 
     /// Check if this [`SqlServerColumnDesc`] is excluded from being replicated into Materialize.
     pub fn is_excluded(&self) -> bool {
-        self.column_type == None
+        self.column_type.is_none()
     }
 }
 


ptravers:
want to verify that everyone is on board with:
1. the `is_excluded` method makes sense and that there isn't another way for a column to have type `None`.
2. you both are okay with using the `mut Vec` + for loop approach rather than map and collect.

ptravers:
update: the failing cardinality.slt test file succeeds locally...

src/sql-server-util/src/desc.rs
```diff
@@ -235,6 +235,11 @@ impl SqlServerColumnDesc {
     pub fn exclude(&mut self) {
         self.column_type = None;
     }
+
+    /// Check if this [`SqlServerColumnDesc`] is excluded from being replicated into Materialize.
+    pub fn is_excluded(&self) -> bool {
+        self.column_type == None
```
martykulma: prefer `is_none()` (more idiomatic) over `== None` for Option type

src/sql/src/pure/error.rs
```diff
@@ -382,6 +382,8 @@ pub enum SqlServerSourcePurificationError {
         col_name: Arc<str>,
         col_type: Arc<str>,
     },
+    #[error("Table {tbl_name} had all columns excluded")]
+    NoColumnsForTable { tbl_name: Arc<str> },
```
martykulma: nit: `AllColumnsExcluded` may be a better name here as it isn't immediately clear why a table would have no columns.

ptravers: great name! ty!

src/sql-server-util/src/desc.rs
ptravers: will fix

PR #33075, opened 2025-07-18
Eliminate one of the `instance_snapshot` calls from peek sequencing

ggevay:
This is a potentially expensive operation, as it enumerates all collections in the cluster.

The second call, which this PR deletes, should give the same result as the first: The controllers are driven by the coordinator main task, so nothing could have happened in the compute controller between the two calls. (`peek_validate` happens on the main task.)

(Btw. if we really wanted to, we might be able to eliminate also the other `instance_snapshot` call. It has some trivial usages, which just ask for the `instance_id`, and there is only one non-trivial usage: https://github.com/MaterializeInc/materialize/blob/235d12a7d910c486122caa172142e5dfd061d5cf/src/adapter/src/coord/indexes.rs#L89 This seems to be mostly shadowed by the filtering that is inside the `get_indexes_on` call just above it, except that it can happen that multiple objects are added to the catalog at the same time, in which case we can use only those that have been already actually created in the controller. If we could ensure that the creation order in the controller is the same as the GlobalId order, then we could replace the `contains_collection` call with just a check on the ids, similar to the `replan` check.)

### Motivation

  * This PR slightly improves QPS.

### Checklist

### DIFF
--- a/src/adapter/src/coord/sequencer/inner/peek.rs
+++ b/src/adapter/src/coord/sequencer/inner/peek.rs
@@ -296,9 +296,6 @@ impl Coordinator {
         let optimizer = match copy_to_ctx {
             None => {
                 // Collect optimizer parameters specific to the peek::Optimizer.
-                let compute_instance = self
-                    .instance_snapshot(cluster.id())
-                    .expect("compute instance does not exist");
                 let (_, view_id) = self.allocate_transient_id();
                 let (_, index_id) = self.allocate_transient_id();
 


ggevay:
The CI failure looks like a CI issue, discussed [here](https://materializeinc.slack.com/archives/C01LKF361MZ/p1752838309238919).

teskje (approved):
LGTM!

I am surprised by this part: "This is a potentially expensive operation, as it enumerates all collections in the cluster." Enumerating the collections in a cluster doesn't seem especially expensive to me. Did you notice any performance issues?

ggevay:
TFTR!

I just meant that it scales linearly with the number of objects on the cluster, which is potentially a lot. Ideally, we shouldn't need to have that for querying just one object of the cluster.

PR #33082, opened 2025-07-18
doc/design: A Small Coordinator for a More Scalable and Isolated Materialize

aljoscha:
Rendered: https://github.com/aljoscha/materialize/blob/design-small-coordinator/doc/developer/design/20250717_a_small_coordinator_more_scalable_isolated_materialize.md


doc/developer/design/20250717_a_small_coordinator_more_scalable_isolated_materialize.md
```diff
@@ -0,0 +1,332 @@
...
+_Big Coordinator_. The Coordinator is big in terms of the types of commands it
+supports, how complex ("big") those commands are, and how much time the
+Coordinator has to spend on its main loop for processing those commands.
+Currently, the frontend sends commands of the shape EXECUTE SELECT or EXECUTE
+DDL. These are higher-level, complex commands. The alternative is a Small
+Coordinator that supports a much reduced set of simpler commands: most of the
+work would have to happen in other parts of the system and the Coordinator only
+has too be involved when absolutely necessary. A good analogy might be CISC vs
+RISC instruction sets, where CISC has fewer, more complex opcodes and RISC has
+possibly more, but simpler opcodes.
```
teskje: Nit: I think you mixed up the analogy here! CISC ISAs usually have _more_ opcodes than RISC ISAs.

I think you do want the coordinator to be "RISC" in terms of the complexity (less) and number (fewer) of commands. But RISC machines also have to execute more commands than CISC machines to implement the same logic, whereas I think we also want the coordinator to execute fewer commands than it does now.

doc/developer/design/20250717_a_small_coordinator_more_scalable_isolated_materialize.md
```diff
@@ -0,0 +1,332 @@
...
+<img src="./static/a_small_coordinator/select-metrics-message-count.png" alt="SELECT benchmark - message counts" width="50%">
+
+Around 5000 we see interesting message types that correspond with our 5000 qps:
+
+- `command-execute`: starts execution of the SELECT.
+- `command-catalog_snapshot`: gets a snapshot of the catalog, needed for
+  processing SELECT.
+- `command-commit`: finalizes a single SELECT execution.
+- `controller_ready(compute)`: the compute controller signaling that a peek
+  result is ready and the Coordinator needs to act.
```
teskje, line 176: Both controllers also become ready when frontiers advance. Probably doesn't show up here because it gets drowned out by the peek responses, but it is also something we should fix. The reason we wake up on all frontier changes is that there might have been watches registered for query lifecycle tracking and we need to check those somewhere. But we shouldn't check them on the coordinator main loop.

doc/developer/design/20250717_a_small_coordinator_more_scalable_isolated_materialize.md
```diff
@@ -0,0 +1,332 @@
...
+
+<img src="./static/a_small_coordinator/big-coord-controller.png" alt="Big Coordinator - controller processing" width="50%">
+
+To move to a Small Coordinator, we need to change the storage controller to do
+its work in its own task, similar to how we did for the compute controller in
+[PR: decoupled compute
+controller](https://github.com/MaterializeInc/materialize/pull/29559). With
+that work fully realized, both for the storage controller and the remaining
+compute controller moments a visualization of the workflow would look like
+this:
```
teskje, line 269: I think it already basically looks like this for the compute controller. At least it doesn't really do anything on `process`, just maybe sends a "run maintenance" command to the instance tasks and then returns a stashed response, if it has any.

Sending the "run maintenance" command is cheap. But we can also remove it, I think, by giving each instance task its own maintenance ticker.

doc/developer/design/20250717_a_small_coordinator_more_scalable_isolated_materialize.md
aljoscha: Yeah, I messed up that one part. CISC is actually more opcodes and the opcodes are more complicated. So the analogy works quite well.

Where before the frontend would send 1 EXECUTE SELECT, it would send the smaller GET CATALOG, GET THIS CLIENT, GET THAT CLIENT messages. But the frontend would also keep those clients around so doesn't have to run those commands for every peek.

doc/developer/design/20250717_a_small_coordinator_more_scalable_isolated_materialize.md
```diff
@@ -0,0 +1,332 @@
...
+  conclude that nothing can block the main loop unreasonably long.
+
+## Out of Scope
+
+- We want to do this work within the current singleton `environmentd` process.
+  That is, we don't yet address horizontal scalability of the query/control
+  layer.
+- We don't want to improve throughput benchmark numbers, only remove the
+  Coordinator as a bottleneck. Our work might increase throughput numbers or
+  reveal similar bottlenecks in other parts of the system.
```
ggevay, line 79: I think moving the current work of peek sequencing to the frontend could increase QPS even without horizontal scaling, because:
- Doing most of the peek sequencing work on the per-session frontend task would allow us to at least vertically scale envd inside a single machine when a particular user wants a bit more QPS.
- I imagine that the current staged execution has some overhead, which would disappear if we had straight-line code instead.

ggevay (approved):
This sounds good to me!

After further discussions on zoom [and slack](https://materializeinc.slack.com/archives/C08ACQNGSQK/p1755262701709849), I've started working on the implementation of the ["Processing SELECTs" part](https://github.com/aljoscha/materialize/blob/design-small-coordinator/doc/developer/design/20250717_a_small_coordinator_more_scalable_isolated_materialize.md#processing-selects).

PR #33083, opened 2025-07-18
[slt] Make timestamp selection more deterministic in SLT

bkirwi:
### Motivation

In the previous version, the frontiers were getting held back by the zero-replica cluster apparently. Hoist up the test so the sleep can actually do some work.


ggevay (approved):

bkirwi:
Another option would be to hold back the since with `retain history`, and keep this running on a zero-replica cluster so nothing gets advanced. That would avoid the sleep but involves more complex reasoning, I think...

PR #33085, opened 2025-07-20
mz_join: efficient linear scan through times

teskje:
This PR introduces `mz_join_core_v2`, an evolution of `mz_join_core` that's meant to resolve a long-standing performance issue around processing input batches with a large number of distinct times. `mz_join_core`'s match generation is quadratic in the number of distinct times, which often causes dataflows to struggle when they are expected to join collections with deep histories.

DD's join implements a more efficient match generation strategy that's linear in the number of distinct times. The reason `mz_join_core` doesn't employ that too is that `mz_join_core` wants to yield often to ensure responsiveness, which would require keeping self-referential state between operator invocations, a thing that's notoriously difficult to do in Rust.

`mz_join_core_v2` extends `mz_join_core` by DD's the linear scan strategy. It solves the issue of self-referential state by wrapping the match generation logic into an `async fn`, which produces self-referential `Future`s we can store in the operator state. The operator can then poll these futures to advance the pending work, until it determines the need to yield control.

We implement `mz_join_core_v2` as a new join variant, to enable a slow rollout. Once we are confident that it behaves correctly and doesn't cause performance regressions, we can remove `mz_join_core`.

### Motivation

  * This PR implements a known-desirable feature.

Closes https://github.com/MaterializeInc/database-issues/issues/6777

### Tips for reviewers

The diff is large, but the upper parts of `mz_join_core_v2.rs` can mostly be skipped because they just copy `mz_join_core.rs`. The first commit verbatim copies the latter into the former, so to see what parts actually changed, you can exclude the first commit from the diff.

### Checklist

### DIFF
--- a/src/compute/src/render/join/mz_join_core_v2.rs
+++ b/src/compute/src/render/join/mz_join_core_v2.rs
@@ -9,32 +9,24 @@
 
 //! A fork of DD's `JoinCore::join_core`.
 //!
-//! Currently, compute rendering knows two implementations for linear joins:
+//! Currently, compute rendering knows three implementations for linear joins:
 //!
 //!  * Differential's `JoinCore::join_core`
 //!  * A Materialize fork thereof, called `mz_join_core`
+//!  * Another Materialize fork thereof, called `mz_join_core_v2`
 //!
 //! `mz_join_core` exists to solve a responsiveness problem with the DD implementation.
 //! DD's join is only able to yield between keys. When computing a large cross-join or a highly
 //! skewed join, this can result in loss of interactivity when the join operator refuses to yield
 //! control for multiple seconds or longer, which in turn causes degraded user experience.
 //!
-//! `mz_join_core` currently fixes the yielding issue by omitting the merge-join matching strategy
-//! implemented in DD's join implementation. This leaves only the nested loop strategy for which it
-//! is easy to implement yielding within keys.
+//! `mz_join_core` resolves the yielding issue but compromises by being quadratic over the number
+//! of distinct times in the input. `mz_join_core_v2` is an attempt to replace the quadratic
+//! behavior by using a linear scan through times, copied from the DD implementation.
 //!
-//! While `mz_join_core` retains responsiveness in the face of cross-joins it is also, due to its
-//! sole reliance on nested-loop matching, significantly slower than DD's join for workloads that
-//! have a large amount of edits at different times. We consider these niche workloads for
-//! Materialize today, due to the way source ingestion works, but that might change in the future.
-//!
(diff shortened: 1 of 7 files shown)


src/compute/src/render/join/mz_join_core_v2.rs
```diff
@@ -518,169 +486,486 @@ where
...
+        history1.replay();
+        history2.replay();
+
+        // TODO: It seems like there is probably a good deal of redundant `advance_buffer_by`
+        //       in here. If a time is ever repeated, for example, the call will be identical
+        //       and accomplish nothing. If only a single record has been added, it may not
+        //       be worth the time to collapse (advance, re-sort) the data when a linear scan
+        //       is sufficient.
+
+        // Join the next entry in `history1`, return whether we should yield.
```
antiguru: Stale comment, closure doesn't return a value.

src/compute/src/render/join/mz_join_core_v2.rs
```diff
@@ -518,169 +486,486 @@ where
...
+        self.history2.edits.load(cursor2, storage2, &self.meet);
+
+        // If the input data is small, use the simple strategy.
+        //
+        // TODO: This conditional is taken directly from DD. We should check if it might make sense
+        //       to do something different, like using the simple strategy always when the number
+        //       of distinct times is small.
+        if self.history1.edits.len() < 10 || self.history2.edits.len() < 10 {
+            self.join_key_simple(key);
+            yield_now().await;
```
antiguru, line 739: How expensive is a call to `yield_now`?

teskje: Not very I hope! The only thing it really does is check a bool and set a bool. We pass control back to `Work::process` here, where the `yield_fn` is invoked, which is probably more expensive.

If the yielding turns out to be bad for performance, we have the option of passing the `yield_fn` into the future and checking it here. I had that version before (hence the stale comment above) but reverted to the current form because it's simpler and doesn't seem to impact performance (according to our feature benchmarks at least). Passing the `yield_fn` into the future requires wrapping it into an `Rc`, and also requires passing the `start_time` in.

bkirwi (approved):
Exciting!

I've reviewed the overall structure, but not look at the details of the new join implementation - though assuming it follows DD's version as closely as it seems to, that seems likely to be already good. Overall quite straightforward considering - I appreciate the step-by-step commits.

Any suggestions I'd have for the structure would involve taking this farther away from the shape of either of the existing join impls. From my perspective this is good to go.

teskje:
TFTRs!

PR #33094, opened 2025-07-21
adapter/controller: for controller_ready metrics, differentiate between controllers

aljoscha:
This is so we can attribute better where time on the coordinator main loop is going.

### DIFF
--- a/src/adapter/src/coord.rs
+++ b/src/adapter/src/coord.rs
@@ -108,8 +108,8 @@ use mz_compute_client::controller::error::InstanceMissing;
 use mz_compute_types::ComputeInstanceId;
 use mz_compute_types::dataflows::DataflowDescription;
 use mz_compute_types::plan::Plan;
-use mz_controller::ControllerConfig;
 use mz_controller::clusters::{ClusterConfig, ClusterEvent, ClusterStatus, ProcessId};
+use mz_controller::{ControllerConfig, Readiness};
 use mz_controller_types::{ClusterId, ReplicaId, WatchSetId};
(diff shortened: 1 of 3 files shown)


src/adapter/src/coord.rs
```diff
@@ -397,6 +410,19 @@ impl Message {
     }
 }
 
+/// Which specific controller has signaled readiness.
```
teskje: Nit: Two of the variants are not controllers, so the description is a bit misleading. I don't really have a better one though. All I got is "which part of the controller is ready", but Metrics and Internal are also not controller parts.

src/controller/src/lib.rs
```diff
@@ -164,7 +164,7 @@ pub struct Controller<T: ComputeControllerTimestamp = mz_repr::Timestamp> {
     /// The cluster orchestrator.
     orchestrator: Arc<dyn NamespacedOrchestrator>,
     /// Tracks the readiness of the underlying controllers.
-    readiness: Readiness<T>,
+    pub readiness: Readiness<T>,
```
teskje: Instead of making this `pub`, can we return the `Readiness` from `Controller::ready`? Prevents that clients can muck with the `readiness` state.

If that doesn't work for some reason, I think we should have a getter, to keep this read-only.

aljoscha: It's not easily possible because Readiness is neither `Clone` nor `Copy`, and for good reason because there might be bigger metrics or subscribe responses in there. I'll add a read-only getter.

PR #33115, opened 2025-07-22
Add webhook sources to the list of unretryable commands

bkirwi:
And make a warning when re-creating the emulator container less scary.

### Motivation

https://github.com/MaterializeInc/database-issues/issues/9503


def- (approved):
Awesome, thanks for investigating that! (and sorry for leading you the wrong way)

bkirwi:
All good! Hopefully the new logging is less misleading.

PR #33117, opened 2025-07-23
testdrive: Default transaction_isolation=serializable

def-:
Some of the failures seem to be instances of https://github.com/MaterializeInc/database-issues/issues/7611, which happens whenever we made testdrive faster.
### Checklist


ggevay:
Unfortunately, I think many tests rely on STRICT SERIALIZABLE, and finding all of them reliably would be a big effort. I don't think it's enough to run CI just a few times and see what fails, because many tests would just become low-key flaky.

For example, there are a gazillion tests that insert something into a table, and then simply expect to see the stuff in the table in a SELECT. `materialized-view-refresh-options.td` does this for sure, but there are probably many-many more.

def-:
> For example, there are a gazillion tests that insert something into a table, and then simply expect to see the stuff in the table in a SELECT.

In testdrive that is not a big problem, they would just retry. So maybe I should go back to my initial commit of this which only made the change in Testdrive and not in the other queries. I have not even been able to get https://github.com/MaterializeInc/materialize/pull/33117/commits/1101a49a3cdfb8dc76fa5874d77516bf94f410b5 green anyway.

ggevay:
> In testdrive that is not a big problem, they would just retry. 

Yeah, ok, let's hope so. I can still imagine exotic situations in which it can go wrong, but they are quite exotic, so maybe it would be fine. But I'd say only do this if it yields a substantial speedup.

def-:
Based on this Nightly run I can't see much of a difference compared to main: https://buildkite.com/materialize/nightly/builds/12668 vs https://buildkite.com/materialize/nightly/builds/12662
The extra stress on Materialize found a new panic though: https://github.com/MaterializeInc/database-issues/issues/9507

At the moment I'm against merging this, since it could indeed increase flakiness for questionable benefit.

PR #33119, opened 2025-07-23
EXPLAIN PHYSICAL PLAN without AS uses VERBOSE TEXT

mgree:
When we changed the `EXPLAIN` default output, it hijacked `EXPLAIN PHYSICAL PLAN`. You can still get the old output, but you need to specify `AS VERBOSE TEXT`. This PR makes it so simply asking for `EXPLAIN PHYSICAL PLAN` without any `AS`-format gives you the old, verbose output.

### Motivation

  https://materializeinc.slack.com/archives/CPNFV9CVB/p1753175570682119

### Tips for reviewer

Test added in `explain/default.slt`.

### Checklist


src/sql-parser/src/parser.rs
```diff
@@ -8788,6 +8788,9 @@ impl<'a> Parser<'a> {
                 None => return Err(ParserError::new(self.index, "expected a format")),
                 _ => unreachable!(),
             }
+        } else if has_stage && stage == Some(ExplainStage::PhysicalPlan) {
+            // if EXPLAIN PHSYICAL PLAN is explicitly specified without AS, default to VERBOSE TEXT
```
ggevay: typo in PHSYICAL

ggevay (approved):
Thank you!

doc/user/content/sql/explain-plan.md
```diff
@@ -415,7 +420,7 @@ EXPLAIN WITH(types) FOR
 SELECT a.id, sum(b.amount) FROM accounts a JOIN bids b ON(a.id = b.buyer) GROUP BY a.id;
 ```
 
-Explain the physical plan as text:
+Explain the physical plan as verbose text:
```
kay-kim: Because we mention "verbose" here .... could we disambiguate in the preceding snippets of "Same as above, ... verbose"?

Also, just seeing that line 405 is missing FOR 😄

src/sql-parser/src/parser.rs
mgree: 😅

doc/user/content/sql/explain-plan.md
mgree: Fixed, ty!

frankmcsherry:
Looks good to me; is it the case that `AS VERBOSE TEXT` is now vestigial, or does it cause a thing to happen in another context? My understanding is that it is undocumented, and if it is both undocumented and doesn't do anything, we could also sequence deleting it. But only if it doesn't do anything. :D

mgree:
> Looks good to me; is it the case that `AS VERBOSE TEXT` is now vestigial, or does it cause a thing to happen in another context? My understanding is that it is undocumented, and if it is both undocumented and doesn't do anything, we could also sequence deleting it. But only if it doesn't do anything. :D

It is documented (but not stable). It's also doing something: it's currently the thing that differentiates between the new default `EXPLAIN` and the old `EXPLAIN PHYSICAL PLAN`.

mgree:
Thanks for the very speedy reviews!

ggevay:
> It is documented (but not stable)

(It doesn't appear in the https://materialize.com/docs/sql/explain-plan/#syntax part.)

PR #33129, opened 2025-07-23
Write outline of what needs to be done

SangJunBak:
### Checklist


src/mz-debug/src/utils.rs
```diff
@@ -66,3 +72,38 @@ pub fn zip_debug_folder(zip_file_name: PathBuf, folder_path: &PathBuf) -> std::i
...
+
+pub async fn get_k8s_auth_mode(
+    mz_username: Option<String>,
+    mz_password: Option<String>,
+    k8s_client: &Client,
+    k8s_namespaces: &Vec<String>,
+) -> Result<AuthMode, anyhow::Error> {
+    for namespace in k8s_namespaces.iter() {
+        let materialize_cr = Api::<Materialize>::namespaced(k8s_client.clone(), namespace);
+        let object_list = materialize_cr
```
alex-hunt-materialize:
```suggestion
        let materialize_api = Api::<Materialize>::namespaced(k8s_client.clone(), namespace);
        let object_list = materialize_api
```
This is the API, not one of the objects. You are also shadowing the same name for the objects down below.

src/mz-debug/src/utils.rs
```diff
@@ -66,3 +72,38 @@ pub fn zip_debug_folder(zip_file_name: PathBuf, folder_path: &PathBuf) -> std::i
     zip_writer.finish()?;
     Ok(())
 }
+
+pub async fn get_k8s_auth_mode(
+    mz_username: Option<String>,
+    mz_password: Option<String>,
+    k8s_client: &Client,
+    k8s_namespaces: &Vec<String>,
```
alex-hunt-materialize, line 80: Having a single auth mode for multiple namespaces, or multiple Materialize instances within a single namespace, doesn't really make sense.

Perhaps this is an issue with the way the debug tool operates. It should probably only operate on a single Materialize instance, and we should probably take in both a single namespace and a single name for that instance as arguments to the debug tool.

src/mz-debug/src/utils.rs
```diff
@@ -66,3 +72,38 @@ pub fn zip_debug_folder(zip_file_name: PathBuf, folder_path: &PathBuf) -> std::i
...
+            if let (Some(mz_username), Some(mz_password)) = (&mz_username, &mz_password) {
+                return Ok(AuthMode::Password(PasswordAuthCredentials {
+                    username: mz_username.clone(),
+                    password: mz_password.clone(),
+                }));
+            } else {
+                return Err(anyhow::anyhow!(
+                    "mz_username and mz_password are required for password authentication"
+                ));
+            }
```
alex-hunt-materialize: This is mostly personal preference, but I generally find it clearer, and it is definitely safer to match on enum variants rather than doing a chain of if statements. There is technically a third AuthenticatorKind variant, which you also aren't handling (even if handling it is to just bail).
```suggestion
            match authenticator_kind {
                AuthenticatorKind::None => return Ok(AuthMode::None),
                AuthenticatorKind::Password => {
                    if let (Some(mz_username), Some(mz_password)) = (&mz_username, &mz_password) {
                        return Ok(AuthMode::Password(PasswordAuthCredentials {
                            username: mz_username.clone(),
                            password: mz_password.clone(),
                        }));
                    } else {
                        return Err(anyhow::anyhow!(
                            "mz_username and mz_password are required for password authentication"
                        ));
                    }
                }
                _ => Err(anyhow!("Unsupported authenticator kind: {}", authenticator_kind))
            }
```

src/mz-debug/src/utils.rs
```diff
@@ -66,3 +72,38 @@ pub fn zip_debug_folder(zip_file_name: PathBuf, folder_path: &PathBuf) -> std::i
...
+        }
+    }
+    Ok(AuthMode::None)
```
alex-hunt-materialize: This should probably be an error, since we didn't find an AuthenticatorKind.

src/mz-debug/src/internal_http_dumper.rs
```diff
@@ -21,27 +22,95 @@ use url::Url;
...
+    heap_profile_port_label: &'static str,
+    prom_metrics_port_label: &'static str,
+}
+
+fn get_port_labels(auth_mode: &AuthMode, service_type: &ServiceType) -> HttpPortLabels {
+    match (auth_mode, service_type) {
+        (AuthMode::None, ServiceType::Clusterd)
+        | (AuthMode::None, ServiceType::Environmentd)
+        // Even if in the password listener config, the heap profile port is specified as external, clusterd will
+        // still use the internal port.
```
alex-hunt-materialize, line 66: The listener config was only written for environmentd.

src/mz-debug/src/internal_http_dumper.rs
```diff
@@ -229,94 +315,193 @@ pub async fn dump_self_managed_http_resources(
...
+                }
+            }
+            (_, _) => {
+                return Err(anyhow::anyhow!(
+                    "Invalid port label combination: heap_profile_port_label={}, prom_metrics_port_label={}",
+                    heap_profile_port_label,
+                    prom_metrics_port_label
+                ));
+            }
+        };
```
alex-hunt-materialize: You are duplicating a ton of code just to handle if you need to spawn one port forward or two. The code doesn't indicate this quickly, so it took me a while to realize this.

I made this suggestion in the github UI, so it's probably got syntax or lifetime errors.

```suggestion
            let (heap_profile_http_connection, prom_metrics_http_connection) = {
                let maybe_heap_profile_port =
                    service_info.service_ports.iter().find_map(|port_info| {
                        find_http_port_by_label(port_info, heap_profile_port_label)
                    });

                let maybe_prom_metrics_port =
                    service_info.service_ports.iter().find_map(|port_info| {
                        find_http_port_by_label(port_info, prom_metrics_port_label)
                    });

                if let (Some(heap_profile_port), Some(prom_metrics_port)) =
                    (maybe_heap_profile_port, maybe_prom_metrics_port)
                {
                    let heap_profile_port_forwarder = KubectlPortForwarder {
                        context: self_managed_context.k8s_context.clone(),
                        namespace: service_info.namespace.clone(),
                        service_name: service_info.service_name.clone(),
                        target_port: heap_profile_port.port,
                    };

                    let heap_profile_http_connection = Arc::new(
                        heap_profile_port_forwarder
                            .spawn_port_forward()
                            .await
                            .with_context(|| {
                                format!(
                                    "Failed to spawn port forwarder for service {}",
                                    service_info.service_name
                                )
                            })?,
                    );

                    let prom_metrics_port_connection = if heap_profile_port == prom_metrics_port {
                        Arc::clone(&heap_profile_http_connection)
                    } else {
                        let prom_metrics_port_forwarder = KubectlPortForwarder {
                            context: self_managed_context.k8s_context.clone(),
                            namespace: service_info.namespace.clone(),
                            service_name: service_info.service_name.clone(),
                            target_port: prom_metrics_port.port,
                        };

                        Arc::new(
                            prom_metrics_port_forwarder
                                .spawn_port_forward()
                                .await
                                .with_context(|| {
                                    format!(
                                        "Failed to spawn port forwarder for service {}",
                                        service_info.service_name
                                    )
                                })?,
                        )
                    }
                    (heap_profile_http_connection, prom_metrics_http_connection)
                } else {
                    return Err(anyhow::anyhow!(
                        "Failed to find HTTP port for service {}, heap_profile_port_label={}, prom_metrics_port_label={}",
                        service_info.service_name,
                        heap_profile_port_label,
                        prom_metrics_port_label
                    ));
                }
```

src/mz-debug/src/utils.rs
SangJunBak: > or multiple Materialize instances within a single namespace, doesn't really make sense.

Out of curiosity, what about during a 0dt deployment? Do we count those as multiple materialize instances or just one that's going through a deployment?

>Perhaps this is an issue with the way the debug tool operates. It should probably only operate on a single Materialize instance, and we should probably take in both a single namespace and a single name for that instance as arguments to the debug tool.

So I think the reason we accept multiple namespaces is because we might want to get the logs / describe outputs of k8s resources from multiple namespaces (e.g. the operator yaml files in the `materialize` namespace). When would someone have multiple instances of Materialize across different namespaces however?

src/mz-debug/src/utils.rs
SangJunBak: Yeah you're right, this is much more readable!

src/mz-debug/src/internal_http_dumper.rs
SangJunBak: Gotchya. Were there any plans to extend that for cluster pods?

src/mz-debug/src/internal_http_dumper.rs
SangJunBak: honestly pretty 1:1 with the actual thing (just missing semicolons and wrong variable names). This is much cleaner tho 😮‍💨

def-:
I noticed a problem with the mz-debug test and fixed it in the above commit. ^

src/mz-debug/src/utils.rs
alex-hunt-materialize: >Out of curiosity, what about during a 0dt deployment? Do we count those as multiple materialize instances or just one that's going through a deployment?

Just one going through a deployment. Two `generations` exist simultaneously, but it's still only one Materialize.

>So I think the reason we accept multiple namespaces is because we might want to get the logs / describe outputs of k8s resources from multiple namespaces (e.g. the operator yaml files in the materialize namespace).

That's a fair point. I'm not sure the best way to represent this. Perhaps we want separate args for **_the_** Materialize namespace, and for additional supporting namespaces.

>When would someone have multiple instances of Materialize across different namespaces however?

Dev vs other dev vs staging vs production, etc... We actually support multiple even in the same namespace, though we don't encourage that or test that well.

src/mz-debug/src/internal_http_dumper.rs
alex-hunt-materialize: I honestly hadn't considered it. We might want to at some point, but I haven't seen a pressing need for it yet.

test/mz-debug/mzcompose.py
```diff
@@ -48,27 +47,8 @@ class TestCase:
...
-            "docker",
-            "cp",
-            f"{container_id}:/usr/local/bin/mz-debug",
-            ".",
-        ],
-        check=True,
-    )
-
+    c.up("materialized", {"name": "mz-debug", "persistent": True})
+    c.invoke("cp", "mz-debug:/usr/local/bin/mz-debug", ".")
```
alex-hunt-materialize, line 55: It feels a bit weird to be pulling this out of the container. Why is it in the materialized image in the first place?

def-: It's not in the materialized image, it's in the mz-debug image. The reason we don't build mz-debug in the test itself is that building is slow and this was the slowest test.

src/mz-debug/src/utils.rs
alex-hunt-materialize: To clarify, there are still two generations of the configmap too.

SangJunBak: >That's a fair point. I'm not sure the best way to represent this. Perhaps we want separate args for the Materialize namespace, and for additional supporting namespaces.

Tbh I like this idea. May handle in a followup PR tho. Kinda sucks that need them to specify the namespaces in the first place. Also if the additional supporting namespaces are optional, there's the risk they won't specify them (I guess still a risk rn). Why do we, in the docs, separate the Materialize instance's namespace from the materialize operator, etc.? Do we expect users' S3 service and RDS instance to be in a namespace separate from the materialize operator?

SangJunBak:
>I noticed a problem with the mz-debug test and fixed it in the above commit. ^

@def- Thank you!!!

alex-hunt-materialize (approved):
I kind of hate the way we detect the auth mode across multiple namespaces and potentially multiple Materialize CRs. Fixing that would probably require changing the args and potentially a bunch more code, though, so I won't block this PR for it.

SangJunBak:
@alex-hunt-materialize Sounds good to me! I'll make a followup that does the following:
> It should probably only operate on a single Materialize instance, and we should probably take in both a single namespace and a single name for that instance as arguments to the debug tool.

>That's a fair point. I'm not sure the best way to represent this. Perhaps we want separate args for the Materialize namespace, and for additional supporting namespaces.

Also put it down as a followup in here https://github.com/MaterializeInc/database-issues/issues/8908

PR #33132, opened 2025-07-24
materialized: Use faster connection to postgres for persist

def-:
@bkirwi Are you the right person to help me here? I want MZ_PERSIST_CONSENSUS_URL_EXTERNAL_CLUSTERD to be used only by clusterd processes that live outside of the materialized container. This should give us some performance in testing and emulator usage since connecting via a local socket is faster.

### Checklist


def-:
Ready for review

test/pg-cdc/alter-table-after-source-1.td
```diff
@@ -245,167 +245,3 @@ ALTER TABLE alter_cycle_pk ADD PRIMARY KEY(f1);
 
 ! SELECT * FROM alter_cycle_pk;
 contains:altered
-
-
-#
-# Cycle PK off (no pk, pk, no pk)
```
martykulma: Was this intentional? It looks a little suspect because the file was renamed to `alter-table-after-source-1.td` and these tests removed, but the tables for these tests are still created and populated at the top of the file.  Are these tests covered elsewhere?

def-: Oops, I split it into two files, but apparently forgot to add the second one, fixed!

PR #33134, opened 2025-07-24
compute: only initialize replica after successful connection

teskje:
This PR changes the compute controller to defer sending the initialization commands until after a replica connection has been established. Doing so avoids command pileup at the replica task and ensures that commands can be reduced without ever making it to the replica if their dataflows/peeks have been dropped/canceled before the connection succeeded.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9502
Fixes https://github.com/MaterializeInc/database-issues/issues/9413

Closes https://github.com/MaterializeInc/database-issues/issues/9483

### Tips for reviewer

Ideally we would do the same thing for the storage controller, but there it's harder because the storage controller might decide to install certain sources/sinks only on a single replica. It is not clear how that should interact with sending initialization commands only to connected replicas. Should a source only be scheduled on a replica once the connection succeeds? Should the source be moved to a different replica once the currently active one disconnects? All questions we could answer, but this change seems less important for the storage controller, given that it doesn't have the same risk of (peek) commands piling up that compute has.

### Checklist

### DIFF
--- a/src/compute-client/src/controller.rs
+++ b/src/compute-client/src/controller.rs
@@ -753,7 +753,7 @@ where
         instance.replicas.insert(replica_id);
 
         instance.call(move |i| {
-            i.add_replica(replica_id, replica_config)
+            i.add_replica(replica_id, replica_config, None)
                 .expect("validated")
         });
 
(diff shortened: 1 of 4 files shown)


teskje:
This doesn't quite solve all the issues we'd like to solve. Namely it doesn't avoid that input frontiers are needlessly held back for collections installed on replicas that don't manage to come up. To avoid that, we should also defer initialization of per-replica collection state until after the connection, but this gets hairy because a lot of things assume that the per-replica collection state gets initialized together with the global collection state.

One solution would be to get rid of per-replica read holds. [We know how to do it](https://www.notion.so/materialize/Controller-Backlog-23a13f48d37b803a9d2afdc07f188cc3?source=copy_link#24113f48d37b8034b610d25a86581c3d), it just hasn't been implemented yet.

PR #33139, opened 2025-07-24
as_of_selection: fix two bugs around upper constraints

teskje:
This PR fixes two bugs in as-of selection, where we could force the upper as-of bound to be lower than necessary. A too-low as-of bounds can cause us to select too-low as-ofs for dataflows, which in turn can cause compute reconciliation to replace dataflows, impacting availability during 0dt upgrades.

The two bugs are:

* In `apply_index_read_policy_constraints` we want to apply the read policies of index collections to the meet of the write frontiers of their transitive storage inputs. For that we need to collect the storage write frontiers and then propagate them through the compute dependency tree. The propagation was wrong as it was using the upper as-of bound of the upstream compute collections, instead of their transitive storage write frontier.
* In `apply_warmup_constraints` we want to hold back the upper as-of bounds of all collections by the meet of the write frontiers of their transitive storage inputs. For that we again need to collect the storage write frontiers and propagate them. But instead we immediately applied the storage write frontiers as upper as-of constraints of their direct dependents and then propagated the upper as-of constraints downstream. This is wrong because it also propagates upper bounds we don't mean to propagate, such as bounds enforced by downstream MVs.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9273

### Checklist

### DIFF
--- a/src/compute-client/src/as_of_selection.rs
+++ b/src/compute-client/src/as_of_selection.rs
@@ -508,27 +508,51 @@ impl<'a, T: TimestampManipulation> Context<'a, T> {
     /// hydrate immediately, but it will be able to hydrate once its inputs have sufficiently
     /// advanced.
     fn apply_warmup_constraints(&self) {
-        // Apply direct constraints from storage inputs.
+        // Collect write frontiers from storage inputs.
+        let mut write_frontiers = BTreeMap::new();
         for (id, collection) in &self.collections {
-            for input_id in &collection.storage_inputs {
-                let frontiers = self
-                    .storage_collections
-                    .collection_frontiers(*input_id)
-                    .expect("storage collection exists");
-                let upper = step_back_frontier(&frontiers.write_frontier);
-                let constraint = Constraint {
-                    type_: ConstraintType::Soft,
-                    bound_type: BoundType::Upper,
-                    frontier: &upper,
-                    reason: &format!("storage input {input_id} warmup frontier"),
-                };
-                self.apply_constraint(*id, constraint);
+            let storage_frontiers = self
+                .storage_collections
+                .collections_frontiers(collection.storage_inputs.clone())
+                .expect("storage collections exist");
+
+            let mut write_frontier = Antichain::new();
+            for frontiers in storage_frontiers {
+                write_frontier.extend(frontiers.write_frontier);
             }
+
+            write_frontiers.insert(*id, write_frontier);
         }
 
-        // Propagate constraints downstream. This transparently restores any violations of
-        // `AsOfBounds` invariant (2) that might be introduced by the propagation.
-        self.propagate_bounds_downstream(BoundType::Upper);
+        // Propagate write frontiers through compute inputs.
+        fixpoint(|changed| {
+            for (id, collection) in &self.collections {
+                let mut write_frontier = write_frontiers.remove(id).expect("inserted above");
+                for input_id in &collection.compute_inputs {
+                    let frontier = &write_frontiers[input_id];
+                    *changed |= write_frontier.extend(frontier.iter().cloned());
+                }
+                write_frontiers.insert(*id, write_frontier);
+            }
+        });
+
+        // Apply the warmup constraint.
+        for (id, write_frontier) in write_frontiers {
+            let upper = step_back_frontier(&write_frontier);
+            let constraint = Constraint {
+                type_: ConstraintType::Soft,
+                bound_type: BoundType::Upper,
+                frontier: &upper,
+                reason: &format!(
+                    "warmup frontier derived from storage write frontier {:?}",
+                    write_frontier.elements()
+                ),
+            };
+            self.apply_constraint(id, constraint);
+        }
+
+        // Restore `AsOfBounds` invariant (2).
+        self.propagate_bounds_upstream(BoundType::Upper);
     }
 
     /// Apply as-of constraints to ensure indexes contain historical data as requested by their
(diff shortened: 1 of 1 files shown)


petrosagg:
> The propagation was wrong as it was using the upper as-of bound of the upstream compute collections, instead of their transitive storage write frontier.

This begs the question, if we're willing to ignore the upper frontier of the immediate compute dependencies, why do we care about the upper frontier of the storage collections? If we choose an as-of that is beyond the upper of some of the dependencies couldn't we also just select a recent upper, and ignore all dependencies? What makes storage dependencies special?

teskje:
So the warmup constraint originally exists to make sure REFRESH MVs can hydrate before the next refresh time. Without the warmup constraint, we would set their as-of to the write frontier of the output shard, which is at the next refresh time, and so they could only start to hydrate at that refresh time. If hydration takes, say, an hour, then there would be a 1-hour lag in the MV output. We can avoid that by letting the MV already hydrate before the refresh time, and then at refresh time it just needs to process the new data that arrived at that second, which is hopefully small.

But the warmup constraint is useful in other cases as well. For example, if you have an index on a source, and the source is installed on a paused cluster and therefore has a write frontier that's an hour in the past. When we restart Materialize, we would like to hydrate the index with the data currently in the source, even if it's stale. Hydration is generally the most resource intensive part of a dataflow's lifecycle, the time where the replica is at the highest risk of ooming. If the replica is going to oom it's better to know that close to envd restart, when we are still in the 0dt upgrade and may be able to roll back, than a couple hours later when the user unpauses the source cluster. Also, a stale hydrated index can still be queried with serializable isolation and is therefore more useful than an unhydrated index waiting for its input frontiers to advance.

So how should we select the as-of bound for the warmup constraint? You suggested using a "recent upper". I'm not quite sure what you mean by that, but presumably you mean the oracle read timestamp (which we receive as the `current_time` parameter to `as_of_selection::run`)? That would probably work for the REFRESH MV case, but it wouldn't help with the index-on-paused-source case -- we would select a too-late as-of for the index to hydrate based off the stale source collection. So we need to take the frontiers of storage inputs into account.

The reason why we don't want to take frontiers of compute inputs into account is that that can break compute reconciliation. Consider the graph from the test added in this PR:

```
source [20] -> index1 --> index2
                      '-> mv [10]
```

(The numbers in brackets are the storage write frontiers we get as the input to the as-of selection process.)

For correctness, we must for `mv` select an as-of <= `10`. Which means we also must select as-of <= `10` for `index1`. Now what the code does before this PR is to also force as-of <= `10` for `index2`. But if we are reconnecting to an already running replica maintaining `index1` and `index2` it might be the case that both of them have already computed their outputs up to time `20`, because that is the latest time available in the source they are reading from. `index1` isn't allowed to compact across time `10` because it is depended on by `mv` which hasn't yet processed that time. But `index2` is allowed to compact up to time `20`. So if the controller now reconnects and reinstalls `index2` with an as-of of `10`, then the replica would have to replace and rehydrate the `index2` dataflow, which is something we're trying to avoid.

Considering only the write frontiers of storage inputs does avoid the reconciliation issue. It means that dataflows might not be able to hydrated immediately because they need to wait for their input dataflows to hydrate, but that's fine. We usually prefer sequential hydration anyway. The thing we want to avoid is that a replica sits around idle with unhydrated dataflows because we selected an as-of that prevents them from being hydrated now.

petrosagg:
Thanks for the reply! I got confused when I wrote my comment because when you said "transitive storage dependencies" I incorrectly assumed that if an index depends on an mv we wouldn't stop at that mv but continue our traversal to its dependencies etc etc until we reach either sources or tables. That made me believe that we already accepted the downside of not immediately hydrating.

Anyway, after your comment I understand what you're trying to do now and it makes sense!

teskje:
Ah, I understand the confusion. I should probably have said "storage dependencies". The "transitive" is there to make it clear that we also need to follow the graph upstream through compute dependencies, but it doesn't tell you that you should _not_ follow it through the dependencies of storage collections.

ggevay:
The Nightly slt fails should go away if you rebase on main.

ggevay:
Ah, sorry, forgot to merge [the slt PR](https://github.com/MaterializeInc/materialize/pull/33133) yesterday. I merged it now.

teskje:
Yeah, no need to rebase I think. We can just ignore any nightly failures that also occur on main, which is all of them.

src/compute-client/src/as_of_selection.rs
```diff
@@ -508,27 +508,51 @@ impl<'a, T: TimestampManipulation> Context<'a, T> {
...
+        fixpoint(|changed| {
+            for (id, collection) in &self.collections {
+                let mut write_frontier = write_frontiers.remove(id).expect("inserted above");
+                for input_id in &collection.compute_inputs {
+                    let frontier = &write_frontiers[input_id];
+                    *changed |= write_frontier.extend(frontier.iter().cloned());
+                }
+                write_frontiers.insert(*id, write_frontier);
+            }
+        });
```
ggevay, line 543: The "Collect write frontiers from storage inputs." and the "Propagate write frontiers through compute inputs." parts are now copy-pasted between `apply_warmup_constraints` and `apply_index_read_policy_constraints`. Could this be factored out into a function instead?

ggevay (approved):
LGTM, apart from some minor comments.

A minor thing is that I think it would be good to mention either in the doc comment of `apply_warmup_constraints` or at its call site that this is good not just for REFRESH MVs, i.e., the second paragraph of https://github.com/MaterializeInc/materialize/pull/33139#issuecomment-3116783376.

src/compute-client/src/as_of_selection.rs
teskje: I'm on the fence on this one. On one hand this code can be easily factored out. On the other hand, I'm not so sure it helps readability. At least I can't come up with a good name to call this factored-out function and it would probably need a sizable doc string to explain what it's doing. I think when you read the `apply_*` functions, you might have to read the code of the factored-out function anyway to be sure the logic is correct. Also, the `.expect("inserted above")` isn't as obvious anymore when the insert happened in a different function.

Let's leave the duplication as is. We should definitely duplicate if we start needing this logic a third time!

ggevay: Ok!

teskje:
TFTR!

bosconi:
bugbot run

PR #33145, opened 2025-07-25
storage/upsert: remove the autospill backend

teskje:
This PR rips out the autospill backend, which would start off using the in-memory hashmap and then start spilling to RocksDB once some size threshold was reached. We determined that this is an unnecessary optimization and we can avoid some complexity by deleting it.

This also paves the way for replacing the in-memory hashmap with RocksDB+MemEnv. Previously doing so would have been kind of weird because the autospill backend would start "spilling" to an in-memory RocksDB instance once the threshold is reached.

### Motivation

   * This PR implements a known-desirable feature.

Part of https://github.com/MaterializeInc/database-issues/issues/9480

[See discussion in Slack.](https://materializeinc.slack.com/archives/C08A7H11KP0/p1752782229619859)

### Tips for reviewer

There was some uncertainty about whether or not the autospill optimization is needed or not. We switched off autospill in production without noticing any averse affects, so most likely it was indeed not needed.

### Checklist


src/storage/src/metrics/upsert.rs
```diff
@@ -95,12 +91,6 @@ impl UpsertMetricDefs {
                     this source",
                 var_labels: ["source_id", "worker_id"],
             )),
-            rocksdb_autospill_in_use: registry.register(metric!(
-                name: "mz_storage_upsert_state_rocksdb_autospill_in_use",
-                help: "Flag to denote whether upsert state has spilled to rocksdb \
-                    or using in-memory state",
-                var_labels: ["source_id", "worker_id"],
-            )),
```
teskje, line 103: Perhaps unexpectedly, this metric is only used with the autospill backend. If the in-memory or the rocksdb backend is used, no metric values are written.

We could change the metric to track use of the in-memory/rocksdb backend, but given that we plan to remove the former one as a follow-up, that seems like a waste of time.

petrosagg (approved):
Thank you!

teskje:
TFTR!

PR #33147, opened 2025-07-25
Arrangement size logging: observe more batches

antiguru:
The arrangement size logging only records batches that are in the trace, but not the ones that have been sent down the stream and that someone might still hold on to. This change corrects this.

Note that we do not account for the size of arrangements that we currently merge, we only account for the inputs but not the output. Differential doesn't offer an API to observe the state of the newly formed batch.

### DIFF
--- a/src/compute/src/extensions/arrange.rs
+++ b/src/compute/src/extensions/arrange.rs
@@ -7,12 +7,13 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::rc::Rc;
+use std::rc::{Rc, Weak};
 
 use differential_dataflow::difference::Semigroup;
 use differential_dataflow::lattice::Lattice;
 use differential_dataflow::operators::arrange::arrangement::arrange_core;
 use differential_dataflow::operators::arrange::{Arranged, TraceAgent};
+use differential_dataflow::trace::implementations::spine_fueled::Spine;
 use differential_dataflow::trace::{Batch, Batcher, Builder, Trace, TraceReader};
 use differential_dataflow::{Collection, Data, ExchangeData, Hashable};
 use timely::Container;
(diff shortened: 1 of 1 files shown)


src/compute/src/extensions/arrange.rs
```diff
@@ -265,15 +265,35 @@ where
...
+                // and retain only those that are still referenced.
+                batches.sort_unstable_by_key(Weak::as_ptr);
+                batches.dedup_by_key(|batch| batch.as_ptr());
+                batches.retain(|batch| batch.strong_count() > 0);
+
+                let (mut size, mut capacity, mut allocations) = (0, 0, 0);
+                for batch in batches.iter().flat_map(Weak::upgrade) {
+                    let (sz, c, a) = logic(&batch);
+                    (size += sz, capacity += c, allocations += a);
+                }
```
petrosagg: We're going over all batches three times here, one for dedup, one for retain, and one for computing the size and capacity. I think if we used a `BTreeMap<*const _, _>` (so that we key on the pointer) you can avoid re-doing all the work all the time:

```rust
                trace
                    .borrow()
                    .trace
                    .map_batches(|batch| {
                        let weak = Rc::downgrade(batch);
                        batches.insert(weak.as_ptr(), weak));
                    });

                let (mut size, mut capacity, mut allocations) = (0, 0, 0);
                batches.retain(|_, weak| {
                    let Some(batch) = Weak::upgrade(weak) else {
                        return false;
                    };
                    let (sz, c, a) = logic(&batch);
                    (size += sz, capacity += c, allocations += a);
                    true
                });
```

antiguru: Changed to `BTreeMap`! Should have the same asymptotic complexity, but simpler code.

antiguru:
Thanks for the review!

PR #33151, opened 2025-07-25
Remove lgalloc

antiguru:
Remove lgalloc

Replace all uses by vectors or equivalent.

The PR should probably wait until we deprecate the t-shirt cluster replica sizes.


sjwiesman:
> The PR should probably wait until we deprecate the t-shirt cluster replica sizes.

lgalloc is disablled on t-shirt sized clusters. So it does not need to wait!

antiguru:
>> The PR should probably wait until we deprecate the t-shirt cluster replica sizes.
> lgalloc is disabled on t-shirt sized clusters. So it does not need to wait!

There are some setting diversions on persist for cc vs. non-cc size, and the PR removes them too. So let's better wait, but it seems we have a plan, so should not be a problem!

PR #33156, opened 2025-07-25
Correct replica expiration test

antiguru:
Correct the replica expiration test such that it again tests what it is supposed to test. We achieve this by increasing the amount of data to process.

The test ensures that replica expiration drops data that's further out than the replica expiration. It stopped doing so after introducing (and enabling) temporal bucketing. Temporal bucketing serves as another merge batcher. The merge batcher first stages data in a chunker, and when full, inserts the data in a chain.

We had so little data that it would remain in the chunker, causing us to never reveal it in the operator/arrangement size tracking. Increasing the data volume surfaces the data that we weren't tracking.


antiguru:
> Any plans to adjust the merge batcher to also report sizes of data in the chunker?
No. The interface is a container builder, and it would require some changes around there. I don't see the benefit as not much data should be stored in the chunker.

antiguru:
Thanks for the review!

PR #33163, opened 2025-07-27
Fix `WITH ORDINALITY`

ggevay:
DRAFT TO RUN CI

### Checklist


src/sql/src/plan/hir.rs
```diff
@@ -131,7 +133,7 @@ pub enum HirRelationExpr {
         scalars: Vec<HirScalarExpr>,
     },
     CallTable {
-        func: TableFunc,
+        func: TableFuncMaybeWithOrdinality,
         exprs: Vec<HirScalarExpr>,
     },
```
antiguru: What's the downside of adding a `CallTableWithOrdinality` HIR node? Moving the variant into the `with_ordinality` field seems a bit odd from designing the enum, because it fundamentally changes how we evaluate the table functions.

src/expr/src/relation/func.rs
```diff
@@ -4214,6 +4214,136 @@ impl fmt::Display for TableFunc {
...
+        datums: &'a [Datum<'a>],
+        temp_storage: &'a RowArena,
+    ) -> Result<Box<dyn Iterator<Item = (Row, Diff)> + 'a>, EvalError> {
+        if !self.with_ordinality {
+            return self.func.eval(datums, temp_storage);
+        }
+
+        // WITH ORDINALITY: zip 1, 2, 3, 4, ... to the output of the table function.
+        //
+        // We don't want to handle negative diffs here, similarly to how Reduce and TopK assert that
```
petrosagg: I don't see any variant producing negative diffs. What is an example table func that does it?

src/expr/src/relation/func.rs
```diff
@@ -4214,6 +4214,136 @@ impl fmt::Display for TableFunc {
...
+        let it = self
+            .func
+            .eval(datums, temp_storage)?
+            .flat_map(|(row, diff)| {
+                iter::repeat(row).take(
+                    diff.into_inner()
+                        .try_into()
+                        .expect("checked above for negative diffs"),
+                )
+            })
```
petrosagg: `std::iter::repeat` unconditionally clones the item but `repeat_n` clones `n-1` times and returns the original value for the nth time. For table functions that always produce `Diff::ONE` this means no cloning

We can also reuse the incoming row since we'll drop it anyway and also make sure it has enough capacity to hold an additional datum. I think this will lead to the fewest copies of the data.

Note: `Row::reserve` doesn't currently exist but it's a trivial function to add

```suggestion
        // 2. Do the WITH ORDINALITY.
        let mut next_ordinal: i64 = 1..;
        let it = self
            .func
            .eval(datums, temp_storage)?
            .flat_map(|(row, diff)| {
                let ordinals = next_idx..(next_ordinal + diff.into_inner());
                next_ordinal += diff.into_inner();
                // Reserve enough space in the row for the largest ordinal
                row.reserve(datum_size(Datum::Int64(next_ordinal));   
                
                let n = diff.into_inner()
                        .try_into()
                        .expect("checked above for negative diffs");
                iter::repeat_n(row, n).zip(ordinals).map(|(mut row, ordinal)| {
                    RowPacker::for_existing_row(&mut row).push(Datum::Int64(ordinal));
                    row
                )
            })
```

src/sql/src/func.rs
```diff
@@ -1657,10 +1661,45 @@ macro_rules! builtins {
...
+/// The implementation of a table function is either
+/// 1. just a `CallTable` HIR node (in which case we can put `WITH ORDINALITY` into it when we
+///    create the actual HIR node in `plan_table_function_internal`),
+/// 2. or a general HIR expression. This happens when it's implemented as SQL, i.e., by a call to
+///    `sql_impl_table_func_inner`.
+///
+/// TODO(ggevay): when a table function in 2. is used with WITH ORDINALITY or ROWS FROM, we fall
+/// back to the legacy WITH ORDINALITY implementation, which relies on the row_number window
+/// function, and is mostly broken. It can give an incorrect ordering, and also has an extreme
+/// performance problem in some cases, see
```
petrosagg, line 1677: > It can give an incorrect ordering

What is the correct ordering if the expression we're applying `WITH ORDINALITY` on is a general HIR expression? Is it that the user specifies an `ORDER BY` clause in it and `WITH ORDINALITY` is supposed to preserve that?

src/transform/src/canonicalization/flatmap_to_map.rs
```diff
@@ -45,15 +45,30 @@ impl FlatMapElimination {
     /// Turns `FlatMap` into `Map` if only one row is produced by flatmap.
     pub fn action(relation: &mut MirRelationExpr) {
         if let MirRelationExpr::FlatMap { func, exprs, input } = relation {
-            if let TableFunc::GuardSubquerySize { .. } = func {
+            if let TableFuncMaybeWithOrdinality {
```
petrosagg: note to self: go over this with Gabor

src/transform/src/canonicalization/flatmap_to_map.rs
```diff
@@ -66,7 +81,16 @@ impl FlatMapElimination {
...
                             }
-                            (Some((row, Diff::ONE)), None) => {
+                            (Some((mut row, Diff::ONE)), None) => {
+                                if func.with_ordinality {
+                                    // Extend the row by an 1.
+                                    let mut new_row = Row::empty();
+                                    let mut packer = new_row.packer();
+                                    packer.extend_by_row(&row);
+                                    packer.push(Datum::Int64(1));
+                                    row = new_row;
```
petrosagg:
```suggestion
                                    RowPacker::for_existing_row(row).push(Datum::Int64(1));
```

src/expr/src/relation/func.rs
antiguru: Note that cloning a row will right-size the allocation, i.e., the `reserve` call won't make a difference.

petrosagg: Ah, I checked if we had any right-sizing logic in `Row` itself but didn't check `CompactBytes` thinking that it would do it based on capacity. That's annoying!

antiguru: It's probably better to rewrite this as a bespoke iterator implementation that has a row, which it writes to, truncates, clones, etc. to avoid the reallocation.

petrosagg: Yeah.. I took a stub I think the following will do what we want

```rust
        // 2. Do the WITH ORDINALITY.
        let mut next_ordinal: i64 = 1;
        let it = self
            .func
            .eval(datums, temp_storage)?
            .flat_map(|(row, diff)| {
                // The ordinals that will be associated with this row
                let ordinals = next_idx..(next_ordinal + diff.into_inner());
                next_ordinal += diff.into_inner();
                // The maximum byte capacity we need for the original row and its ordinal
                let cap = row.data_len() + datum_size(Datum::Int64(next_ordinal);
                std::iter::from_fn(move || {
                    let ordinal = ordinals.next()?;
                    let row = if ordinals.is_empty() {
                        // This is the last row, so no need to clone
                        std::mem::take(&mut row);
                    } else {
                        let new_row = Row::with_capacity(cap);
                        new_row.clone_from(&row);
                        new_row
                    };
                    RowPacker::for_existing_row(&mut row).push(Datum::Int64(ordinal));
                    Some(row)
                });
            })
```

antiguru: This still reallocates, right? `row` doesn't have capacity for the ordinal.

petrosagg: The behavior of this code (unless I messed up something up) is that it wants to create `diff` copies of the incoming `row`. The first `diff-1` copies are done with an exact allocation of `cap` bytes which can hold the original `row` plus the ordinal datum. For the last copy the original `row` is reused and if it happens to have enough capacity it is not reallocated. Otherwise it is reallocated. I think this is the best you can do as far as minimizing allocations. Does that check out?

src/sql/src/plan/hir.rs
ggevay: Unfortunately it's not possible to fix the problem just by doing a change at the HIR level. The fundamental problem is that MIR and LIR operators emit unordered collections, so we have to somehow execute `WITH ORDINALITY` inside `FlatMap`, which necessitates changes at all levels (HIR, MIR, optimizer transforms, LIR, rendering).

(Btw. I'm planning to refactor the PR a bit to not introduce the new `TableFuncMaybeWithOrdinality` struct that wraps `TableFunc`, but instead create a new enum variant of `TableFunc` for `WITH ORDINALITY`, which would wrap a normal `TableFunc`. And then the code that is currently in `TableFuncMaybeWithOrdinality` will move into `TableFunc`'s implementation. This is not a fundamental change in what the PR does, but I think this will make the PR's diff shorter.)

src/expr/src/relation/func.rs
ggevay: (Discussed on zoom.) `repeat_row` can emit negative diffs, but this is in `mz_unsafe`, so users can never call it, it's only used in some tests.

src/expr/src/relation/func.rs
ggevay: This checks out to me, thank you! I went with this approach, with some minor tweaks and typo fixes.

I've also added more tests with non-1 diffs to have more coverage of this code, since it is now non-trivial.

src/sql/src/func.rs
ggevay: (Discussed on zoom.)

- (Context: The user can't specify an arbitrary SQL/HIR expression on which we'd need to apply `WITH ORDINALITY`. It's just that some of our table functions are internally defined in SQL.)
- The SQL or HIR expressions doesn't define the order, which is part of the problem.
- A function itself might have a natural order, which a user could reasonably expect. For example, it would be a natural expectation for `regexp_split_to_table` to emit the substrings in the order that they appear in the original string, or for `date_bin_hopping` to emit things in time order.

My earlier solution proposal in the code comment was to define these functions with 2 SQL expressions: one would be the traditional thing, and the other would be invoked when the function is called with `WITH ORDINALITY`.

Another solution approach that I just thought of: We could ask for an `ORDER BY` clause on the SQL definitions, and then somehow follow that when executing `WITH ORDINALITY` (and ignore it when there is no `WITH ORDINALITY`, since in that case the order is meaningless, like when there is an `ORDER BY` in a subquery with no `LIMIT` or `OFFSET`). I'll also add this as a code comment.

src/transform/src/canonicalization/flatmap_to_map.rs
ggevay: (Discussed on zoom.)

@petrosagg suggested a nice refactoring of this whole optimizer transform: instead of having bespoke implementations for several specific table functions (now sometimes separately for WITH ORDINALITY and without), we could have the following blanket implementation for all table functions:
- Check that the arguments are literals.
- If yes, then call the table function with those literals as arguments.
- If the table function emits 1 row, then substitute with `Map`.
- If the table function emits 0 rows, then substitute with empty constant.

I'm thinking that I'll do this in a follow-up PR, to not conflate this PR with a non-trivial optimizer change. (It's a non-trivial change, because this will make us apply this optimization to more table functions than before. We'll probably need to feature flag it, etc.)

src/transform/src/canonicalization/flatmap_to_map.rs
ggevay: Will do, thanks!

src/expr/src/relation/func.rs
ggevay: Btw. another change that I did nearby is to not check for negative diffs, but just assert that there are no negative diffs. This is because @petrosagg pointed out that the diffs of the FlatMap's input don't matter, as they are applied on the _output_ of the table function, not on the input. So, the only situation where we can have negative diffs here is if a table function itself produces a negative diff. This is only possible for `repeat_row`, which is not a thing that users can call. I'll add this invariant to the comment on `TableFunc`, so that we don't accidentally introduce a table function that emits negative diffs.

ggevay:
(I've addressed some of the comments, but not all of them yet. I'll ping one more time when all of them are addressed.)

ggevay:
Thanks for the comments! This is ready for a (hopefully) final review!

I've refactored the PR to not introduce the `TableFuncMaybeWithOrdinality` struct that wrapped `TableFunc`, but instead create a new enum variant of `TableFunc` for `WITH ORDINALITY`, which wraps a normal `TableFunc` (based on @petrosagg's idea). The code that is currently in `TableFuncMaybeWithOrdinality` moved into `TableFunc`'s implementation. This is not a fundamental change in what the PR does, but it made PR's diff significantly smaller: +444 lines across 15 files instead of +558 lines across 33 files.

I'll open follow-up PRs for:
1. Fix [the other WITH ORDINALITY bug](https://github.com/MaterializeInc/database-issues/issues/9551), found just yesterday by a user. Edit: https://github.com/MaterializeInc/materialize/pull/33272
2. Simplify and generalize `flatmap_to_map.rs` based on @petrosagg's idea.
3. Add `WITH ORDINALITY` to our docs. (will merge when the fix goes out)

ggevay:
Just fixed a crazy bug in the new code. Check the commit msg and the comment in the added test only if you want to see a horror story. (Found while working on the `flatmap_to_map.rs` follow-up PR.)

src/expr/src/relation/func.rs
```diff
@@ -3715,9 +3716,10 @@ fn mz_acl_explode<'a>(
     Ok(res.into_iter())
 }
 
-#[derive(
-    Arbitrary, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize, Hash, MzReflect,
-)]
+/// Don't add table functions that emit negative diffs! These are undefined with `WITH ORDINALITY`.
+/// (`Repeat` _can_ emit negative diffs, on which `WITH ORDINALITY` panics, but this is in
```
petrosagg, line 3720: I forget what we said, do we ever generate an invocation of repeat with a negative input as part of optimization? If not, does it make sense to change its implementation to produce an error when the parameter is negative?

More generally, since `TableFunc` implementation don't live in the differential world we could request that they produce an iterator of `(Row, u64)` instead of `(Row, Diff)`, and bake the requirement in the type system.

ggevay: `Repeat` never occurs in prod. (It's hard to make it appear even in staging: you have to edit envd's command-line flags or something.) However, it's sometimes useful for local testing, when you want to see how various components react to negative diffs. So, I'd like to just allow `Repeat` producing negative diffs for now.

ggevay:
TFTR!

src/expr/src/relation/func.rs
frankmcsherry: We'll need more documentation than this. I don't think `WITH ORDINALITY`'s implementation is a great reason to unilaterally rule out negative diffs from functions. At least, that is a *massive* imposition, for a mostly undocumented function (reading the doccomments on `TableFunc::WithOrdinality`, and they mainly say that it is used in implementing `WITH ORDINALITY`, but not what that means and what it does).

Does this mean that `Negate` does not commute with `FlatMap`? It seems probably like one should write whichever table functions make the most sense, and the implementation of `WithOrdinality` wrapping another table function is the moment we need to take on responsibility for "but am I even well defined?", but that's up to `WithOrdinality`, rather than the whole of all the other table functions.

frankmcsherry: I think one thing to do is document the `with_ordinality` function. It currently lacks any doccomments, which means to understand it you have to read the body. But from the point of view of "managing complexity" it's a bug if anyone has to read the body of a method to understand what it does.

I think having done this, we'll see that it is well defined even on functions with negative outputs, and the conclusion isn't "don't create functions with negative weight outputs" but "`WithOrdinality` will suppress outputs with negative weights", and we should be careful wherever we create `WithOrdinality` variants that the intent and implementation match.

ggevay: > Does this mean that `Negate` does not commute with `FlatMap`?

No, fortunately, `Negate` still commutes with `FlatMap` even after this PR, even with `TableFunc::WithOrdinality`. I also fell into the trap of worrying about `Negate` in the first version of my PR, but then Petros pointed out that `TableFunc::eval` never receives a negative diff: even if the input of the `FlatMap` has negative diffs, those diffs are factored in [_after_](https://github.com/MaterializeInc/materialize/blob/833454ccc98c1b8fda03c26448edb4959a327434/src/compute/src/render/flat_map.rs#L171) evaluating the table function.

> We'll need more documentation than this. I don't think WITH ORDINALITY's implementation is a great reason to unilaterally rule out negative diffs from functions. At least, that is a massive imposition
> ...
> It seems probably like one should write whichever table functions make the most sense, and the implementation of WithOrdinality wrapping another table function is the moment we need to take on responsibility for "but am I even well defined?", but that's up to WithOrdinality, rather than the whole of all the other table functions.

Ok, I'll adjust the documentation. How about something like this:
"If you'd like to introduce a `TableFunc` (aside from `Repeat`) that emits negative diffs, please make sure to either prevent it from getting inside `TableFunc::WithOrdinality` at planning time, or come up with a special implementation of `TableFunc::WithOrdinality` for it."

>  (reading the doccomments on TableFunc::WithOrdinality, and they mainly say that it is used in implementing WITH ORDINALITY, but not what that means and what it does)

This is the doc comment:
```
    /// This is a special table function that is used to implement the `WITH ORDINALITY` clause:
    /// When the user does
    /// `f(...) WITH ORDINALITY`,
    /// we plan this as
    /// `TableFunc::WithOrdinality { inner: f(...) }`
    ///
    /// WITH ORDINALITY means that we append an extra output column, whose value is 1,2,3,... for
    /// the output rows corresponding to a call of the table function on one input row. For example,
    /// WITH ORDINALITY numbers the elements of a list when calling unnest_list.
```

The last paragraph was meant to explain what WITH ORDINALITY means. Should I make it more precise/formal? Or what do you see missing from it? Edit: I'll add that it currently asserts that diffs are non-negative.

ggevay: > I think one thing to do is document the with_ordinality function. It currently lacks any doccomments, which means to understand it you have to read the body. But from the point of view of "managing complexity" it's a bug if anyone has to read the body of a method to understand what it does.

Yes, sorry, I'll add a doccomment. Is it ok, if I make it point to the (to be adjusted) doccomment of `TableFunc::WithOrdinality`?

> it is well defined even on functions with negative outputs, and the conclusion isn't "don't create functions with negative weight outputs" but "WithOrdinality will suppress outputs with negative weights", and we should be careful wherever we create WithOrdinality variants that the intent and implementation match.

Yes, when adding a table function that emits negative diffs, making `with_ordinality` simply drop negative diffs could be a possibility also, yes. Currently it has an `assert` against negative diffs, but this could simply be removed if needed for a new table function. I'll mention this in the adjusted doccomment of `TableFunc` and/or `TableFunc::WithOrdinality`.

frankmcsherry: > "If you'd like to introduce a TableFunc (aside from Repeat) that emits negative diffs, please make sure to either prevent it from getting inside TableFunc::WithOrdinality at planning time, or come up with a special implementation of TableFunc::WithOrdinality for it."

Imo, no. This is a `WithOrdinality` limitation, not a "the rest of the IR" limitation.

> The last paragraph was meant to explain what WITH ORDINALITY means.

I don't think the question is what `WITH ORDINALITY` means as much as what `TableFunc::WithOrdinality` means. The comment is an explainer for why the variant exists, but that's not imo as useful as explaining what the variant means or causes to happen. Except for the "panics if negative diffs are produced", it feels like it could be easy to say what it does: "evaluates the inner table function, expands its results into unary (repeating each row as many times as the diff indicates), and appends an integer corresponding to the ordinal position (starting from one)".

ggevay: > evaluates the inner table function, expands its results into unary (repeating each row as many times as the diff indicates), and appends an integer corresponding to the ordinal position (starting from one)

Sounds good, thx, I'll put this there!

ggevay: >> "If you'd like to introduce a TableFunc (aside from Repeat) that emits negative diffs, please make sure to either prevent it from getting inside TableFunc::WithOrdinality at planning time, or come up with a special implementation of TableFunc::WithOrdinality for it."

> Imo, no. This is a `WithOrdinality` limitation, not a "the rest of the IR" limitation.

You mean rather than putting a comment warning against negative diff emitting table functions, just try to somehow make `WithOrdinality` not have this limitation? Would be easy to achieve this by simply removing the non-negative diff assertion from it.

But in any case, someone adding a new table function that emits negative diffs will definitely have to think for a moment about what to do with `WithOrdinality`. So, the current behavior of `WithOrdinality` on negative diffs is just there as a guardrail against people forgetting to do this. If that happens, panicking for negative diffs honestly sounds better to me than the probably incorrect behavior of simply dropping negative diffs.

PR #33184, opened 2025-07-29
Force CDC job to run to address CI flake

martykulma:
Adds 2 new testdrive commands for sql-server
- `sql-server-stop-capture-jobs`
- `sql-server-run-cdc-scan`

The first will stop the automatic capture job run by CDC, the second executes a stored procedure that will initiate the CDC scan.  Both of these have internal retries as the CDC jobs are async and their state will affect whether these succeed or fail.

Note: this will unfortunately cause `tiberius` to log some errors, those are expected.

By disabling the automatic capture job, it requires us to run the cdc scan after every change to the upstream DB in order for changes to get captured.  But I've been having this pass consistently locally and at least 3 times in CI, so it looks fairly stable.

### Motivation

fixes https://github.com/MaterializeInc/database-issues/issues/9519

### Checklist

### DIFF
--- a/src/testdrive/src/action/sql_server/cdc.rs
+++ b/src/testdrive/src/action/sql_server/cdc.rs
@@ -60,7 +60,6 @@ pub async fn run_cdc_scan(
     let timer = Instant::now();
     static CDC_SCAN: &str = "EXEC sys.sp_cdc_scan @continuous = 0, @maxscans = 1;";
     while let Err(error) = client.execute(CDC_SCAN, &[]).await {
-        };
 
         // retry if there is another CDC scan going on
         // 22903 = another connection is already running sp_replcmds


def- (approved):

src/testdrive/src/action/sql_server/cdc.rs
```diff
@@ -0,0 +1,82 @@
...
+        .sql_server_clients
+        .get_mut(&name)
+        .ok_or_else(|| anyhow!("connection {} not found", name.quoted()))?;
+
+    let max_duration = std::cmp::max(state.default_timeout, Duration::from_secs(10));
+    let timer = Instant::now();
+    static CDC_SCAN: &str = "EXEC sys.sp_cdc_scan @continuous = 0, @maxscans = 1;";
+    loop {
+        let Err(error) = client.execute(CDC_SCAN, &[]).await else {
+            break;
```
petrosagg: This is the `while let` pattern
```suggestion
    while let Err(error) = client.execute(CDC_SCAN, &[]).await {
```

def-:
Can we reenable the SQL Server test now in ci/test/pipeline.template.yml? I guess it should be much less flaky.

src/testdrive/src/action/sql_server/cdc.rs
```diff
@@ -0,0 +1,82 @@
...
+        if timer.elapsed() < max_duration {
+            tokio::time::sleep(Duration::from_millis(100)).await;
+        } else {
+            return Err(anyhow!("ERROR: sp_cdc_stop_job failed {error}"));
+        }
+    }
+    Ok(ControlFlow::Continue)
+}
+
+pub async fn run_cdc_scan(
```
petrosagg, line 47: Can we add some doc comments on these methods explaining what they do and why do we need them in the tests?

src/testdrive/src/action/sql_server/cdc.rs
```diff
@@ -0,0 +1,82 @@
...
+    static STOP_CDC_JOBS: &str = "EXEC sys.sp_cdc_stop_job @job_type = N'capture';";
+    let max_duration = std::cmp::max(state.default_timeout, Duration::from_secs(10));
+    let timer = Instant::now();
+
+    // It's very error prone to try and catch this when it is running to stop it.
+    // We need to retry here as the stored procedure errors if the job is not running.
+    loop {
+        let Err(error) = client.execute(STOP_CDC_JOBS, &[]).await else {
+            break;
+        };
```
petrosagg: same here
```suggestion
    while let Err(error) = client.execute(STOP_CDC_JOBS, &[]).await {
```

src/testdrive/src/action/sql_server/cdc.rs
```diff
@@ -0,0 +1,82 @@
...
+    let timer = Instant::now();
+    static CDC_SCAN: &str = "EXEC sys.sp_cdc_scan @continuous = 0, @maxscans = 1;";
+    loop {
+        let Err(error) = client.execute(CDC_SCAN, &[]).await else {
+            break;
+        };
+
+        // retry if there is another CDC scan going on
+        // 22903 = another connection is already running sp_replcmds
+        // see https://learn.microsoft.com/en-us/sql/relational-databases/errors-events/database-engine-events-and-errors-22000-to-22999?view=sql-server-ver17#errors-and-events-22000-to-22999
```
petrosagg, line 66: Thanks for adding references!

petrosagg (approved):
LGTM! One general question I have is whether these commands will need to be typed on every SQL Server test to make sure we don't flake. Could we make the cdc scan job run periodically instead of continuous mode? I also things like these [job schedule related functions](https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-add-schedule-transact-sql?view=sql-server-ver17) that maybe could be used to run this frequently.

Main worry here is accidentally introducing flakes if we add a test and forget to add the calls to scan because it worked locally

martykulma:
> Can we reenable the SQL Server test now in ci/test/pipeline.template.yml? I guess it should be much less flaky.

Yes, I'll pull in the latest changes from main and re-enable them as part of this PR.

martykulma:
> LGTM! One general question I have is whether these commands will need to be typed on every SQL Server test to make sure we don't flake. Could we make the cdc scan job run periodically instead of continuous mode? I also things like these [job schedule related functions](https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-add-schedule-transact-sql?view=sql-server-ver17) that maybe could be used to run this frequently.
> 
> Main worry here is accidentally introducing flakes if we add a test and forget to add the calls to scan because it worked locally

I can try doing that, but I'm not hopeful.

Continuous mode just means that the job will process log entries until it reaches `maxtrans * maxscans` transactions.  If there are no transactions, the job waits for `pollinginterval` (by default 5s) before another log scan.  Given that we don't make thousands of transactions, the scan should be fairly quick, and most of the time should be spent in the `WAITFOR` call used by the job.  Having them scheduled as non-continuous every 5 seconds should do roughly the same amount of work, but maybe I can get all the jobs to run at the same time instead of spread out.  But eventually, I feel like I'll have a pile up of jobs running that will lead to the same type of issue that existed when I didn't disable the jobs.

I started with just adding `run_cdc_scan`, but the more tests that were run, the more errors 22903 errors `run_cdc_scan` hit.  This alone didn't fix the flakes, which is how it evolved to stop the scans first.

One of the benefits of this approach is that it becomes very unlikely to be able to make an upstream change and still have any MZ data validation pass.  When I made these changes, I missed adding the cdc scan in multiple places. Tests reliably failed until I added the scans.

Another option I can try running is `stop-capture-jobs` at the top of each TD file, before creating the DB, tables, and enabling CDC. Hopefully, this disables all capture jobs that were created in previous tests, but allows the current test to have CDC enable.

src/testdrive/src/action/sql_server/cdc.rs
```diff
@@ -0,0 +1,77 @@
...
+    let client = state
+        .sql_server_clients
+        .get_mut(&name)
+        .ok_or_else(|| anyhow!("connection {} not found", name.quoted()))?;
+
+    let max_duration = std::cmp::max(state.default_timeout, Duration::from_secs(10));
+    let timer = Instant::now();
+    static CDC_SCAN: &str = "EXEC sys.sp_cdc_scan @continuous = 0, @maxscans = 1;";
+    while let Err(error) = client.execute(CDC_SCAN, &[]).await {
+        };
```
ptravers:
```suggestion
```

petrosagg:
Recording here some context from Slack.

Marty and I discussed why this happens and it looks like our current purification code comes up with an `initial_lsn` for a table based purely on the result of `fn_cdc_get_max_lsn`. However, that function returns the most recent LSN that was processed by the background log reader process of the server, and that might be before a particular capture instance for a table is registered, making it impossible for our ingestion pipeline to read the table at that LSN. This is an issue that a user might face as well if they enable CDC on a table and then quickly create a source in Materialize.

This root problem here is that purification selected a bad `initial_lsn` for the table, since it is unreadable. If we fix purification to always select an initial LSN that will be available we will both fix our test flakes without additional syntax and also provide a better UX to the users. 

**Proposed solution**
When CDC is enabled for a table SQL server selects the initial LSN that it still start recording changes from and records it in the `cdc.change_tables` table. This happens immediately after a capture instance is enabled. This means that we can change purification to instead select the `max(start_lsn, fn_cdc_get_max_lsn())` as the initial_lsn for a table. 

> **Note:** The documentation claims that the function `sys.fn_cdc_get_min_lsn` retrieves the `start_lsn` column from the `change_tables` table but as Marty discovered this is not always true and sometimes it returns NULL. For this reason we must always read the table ourselves. Here is an example where a capture instance is enabled and then we immediately run that function and also select from the raw table:

```
1> exec sys.sp_cdc_enable_table @source_schema = 'dbo', @source_name = 't1', @role_name = 'SA'
2> select sys.fn_cdc_get_min_lsn('dbo_t1');
3> select capture_instance, start_lsn from cdc.change_tables;
4> go

-----------
0xNULL

(1 row affected)
capture_instance  start_lsn
----------------- ---------------
dbo_t1            0x0000002F000053D80033

(1 row affected)
```

martykulma:
closing this PR in favor of improving the initial LSN selection https://github.com/MaterializeInc/materialize/pull/33202 (see https://github.com/MaterializeInc/materialize/pull/33184#issuecomment-3139632241)

PR #33190, opened 2025-07-30
adapter: transactional source creation

petrosagg:
### Motivation

This PR implements transactional DDL support for `CREATE SOURCE` and `CREATE TABLE FROM SOURCE` statements. Both of these statements produce side-effects after being committed to the catalog.The first commits of this PR deal with making those side effects deferrable, which boils down to making sure they are `'static`.

Then the state of transactions is augmented to hold onto a list of deferred side-effects that need to be executed when the transaction commits.

Finally, `CREATE SOURCE` and `CREATE TABLE FROM SOURCE` statements are allowed to be executed in DDL transactions.

The last commits contains purely test changes which wrap all instances of `CREATE SOURCE foo` followed by one or more `CREATE TABLE FROM SOURCE foo` in `BEGIN`/`COMMIT`. 

### Tips for reviewer

Hide whitespace for a better diff and go commit by commit since the last one is big and is just test changes

### Checklist


src/adapter/src/coord/ddl.rs
```diff
@@ -88,7 +89,11 @@ impl Coordinator {
         side_effect: F,
     ) -> Result<(), AdapterError>
     where
-        F: AsyncFnOnce(&mut Coordinator, Option<&mut ExecuteContext>) -> () + 'static,
+        F: for<'a> FnOnce(
```
aljoscha, line 92: would have been so nice if the simple `AsyncFnOnce` had worked out ... 🙈

src/adapter/src/session.rs
```diff
@@ -1510,6 +1515,18 @@ pub enum TransactionOps<T> {
         ops: Vec<crate::catalog::Op>,
         /// In-memory state that reflects the previously applied ops.
         state: CatalogState,
+        /// A list of side effects that should be executed if this DDL transaction commits
```
aljoscha: nit: missing full stop 😅

petrosagg: fixed.

PR #33196, opened 2025-07-30
controller/compute: remove `replica_epochs` map

teskje:
This PR removes `Instance::replica_epochs`, a map that tracks the per-replica epochs. A defect with this map is that we never remove entries from it, so it leaks some state whenever a replica is dropped. We can fix this by observing that the `ReplicaState` for each replica already contains the current epoch, and there is no need to store it twice. We can thus delete the `replica_epochs` map.

### Motivation

### Checklist

### DIFF
--- a/src/compute-client/src/controller/instance.rs
+++ b/src/compute-client/src/controller/instance.rs
@@ -282,8 +282,6 @@ pub(super) struct Instance<T: ComputeControllerTimestamp> {
     response_tx: mpsc::UnboundedSender<ComputeControllerResponse<T>>,
     /// Sender for introspection updates to be recorded.
     introspection_tx: mpsc::UnboundedSender<IntrospectionUpdates>,
-    /// Numbers that increase with each restart of a replica.
-    replica_epochs: BTreeMap<ReplicaId, u64>,
     /// The registry the controller uses to report metrics.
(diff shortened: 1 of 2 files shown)


src/compute-client/src/controller/instance.rs
```diff
@@ -1210,17 +1201,15 @@ where
         &mut self,
         id: ReplicaId,
         mut config: ReplicaConfig,
+        epoch: Option<u64>,
```
petrosagg, line 1204: What do we gain from making this an option? Both callers of this method already know the correct epoch is so it feels like they could supply it.

teskje:
Yeah, same. I figured this is better than having the caller need to concern itself with what the first epoch should be, and I also didn't feel like adding a wrapper method just for this. But both would be fine too.

src/compute-client/src/controller/instance.rs
teskje: It's mostly that the replica epoch is a concept internal to the per-instance controller and I'd like to not expose it outside of that. The proper way is to add a wrapper method that skips the last argument and supplies the correct first epoch when called. But I didn't want to bother thinking about how to name it 😅

teskje:
TFTR!

PR #33202, opened 2025-07-30
Improve initial_lsn selection for SQL Server

martykulma:
In some cases, the LSN returned by `sys.fn_cdc_get_max_lsn` can be less than the minimum LSN of a capture instance, resulting in a broken source. This PR change the behavior of the `initial_lsn` to address that by choosing `initial_lsn = max(minimum_lsns, sys.fn_cdc_get_max_lsn)`;

### Motivation

Fix https://github.com/MaterializeInc/database-issues/issues/9519

### Checklist

### DIFF
--- a/src/sql/src/pure/sql_server.rs
+++ b/src/sql/src/pure/sql_server.rs
@@ -176,7 +176,9 @@ pub(super) async fn purify_source_exports(
         capture_instances.values().map(|instance| instance.as_ref()),
     )
     .await?;
+
     let max_lsn = mz_sql_server_util::inspect::get_max_lsn_retry(client, timeout).await?;
+    tracing::debug!(?initial_lsns, %max_lsn, "retrieved start LSNs");
     for lsn in initial_lsns.values_mut() {
         *lsn = std::cmp::max(*lsn, max_lsn);
     }
--- a/src/storage/src/source/sql_server/replication.rs
+++ b/src/storage/src/source/sql_server/replication.rs
@@ -261,15 +261,15 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
             export_ids_to_snapshot.clear();
 
 
-            // Resumption point is the minimum LSN that has been observed.
-            let resume_lsn = outputs
+            // Resumption point is the minimum LSN that has been observed per capture instance.
+            let resume_lsns:BTreeMap<_, _> = outputs
                 .values()
-                .flat_map(|src_info| {
+                .map(|src_info| {
                     // initial_lsn is the max lsn observed, but the resume lsn
                     // is the next lsn that should be read.  After a snapshot, initial_lsn
                     // has been read, so replication will start at the next available lsn.
                     let start_lsn = src_info.initial_lsn.increment();
-                    src_info.resume_upper
+                    let resume_lsn = src_info.resume_upper
                         .elements()
                         .iter()
                         .map(move |lsn| {
@@ -279,23 +279,29 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
                                 *lsn
                             }
                         })
+                        .min()
+                        .expect("resume_upper has at least one value");
+                    (Arc::clone(&src_info.capture_instance), resume_lsn)
                 })
-                .min();
+                .collect();
 
             // If a source is created with no subsources, there will be no inputs
             // as each capture instance collects data for a table. Without a cdc stream
             // to consume, this source will have no outputs either.
-            let Some(resume_lsn) = resume_lsn else {
+            if resume_lsns.is_empty() {
                 // TODO: this is only in place until we implement RLU
                 // https://github.com/MaterializeInc/database-issues/issues/9212
                 tracing::warn!(%config.id, "timely-{} no resume_lsn, waiting", config.worker_id);
                 std::future::pending::<()>().await;
                 unreachable!();
             };
-            tracing::info!(%config.id, "timely-{} replication starting with resume_lsn = {}", config.worker_id, resume_lsn);
 
+            tracing::info!(%config.id, ?resume_lsns, "timely-{} replication starting", config.worker_id);
             for instance in capture_instances.keys() {
-                cdc_handle = cdc_handle.start_lsn(instance, resume_lsn);
+                let resume_lsn = resume_lsns
+                    .get(instance)
+                    .expect("resume_lsn exists for capture instance");
+                cdc_handle = cdc_handle.start_lsn(instance, *resume_lsn);
             }
 
             // Off to the races! Replicate data from SQL Server.


ptravers:
if it's flake should we run the suite a bunch of times til it starts to work? I think that the bug I'm seeing in LRU work is related to the initial lsn across capture tables.

src/sql-server-util/src/inspect.rs
```diff
@@ -75,6 +75,51 @@ pub async fn get_max_lsn(client: &mut Client) -> Result<Lsn, SqlServerError> {
     parse_lsn(&result[..1])
 }
 
+/// Retrieves the minumum [`Lsn`] (start_lsn field) from `cdc.change_tables`.
+/// This is based on digging into `sys.fn_cdc_get_min_lsn` implementation, which has
+/// some additional checks that we want to bypass.
```
ptravers: might be worth noting the checks we want to bypass?

martykulma:
> if it's flake should we run the suite a bunch of times til it starts to work? I think that the bug I'm seeing in LRU work is related to the initial lsn across capture tables.

Indeed!  I have run it locally repeatedly, and created this PR to run it in CI (sometimes I can't reproduce the failures, but they happen in CI).

src/sql-server-util/src/inspect.rs
martykulma: good call, updated!

martykulma:
I've run this [repeatedly in CI](https://buildkite.com/materialize/test/builds/107222#01986156-b463-4c30-8c83-9c8df56d5819) and locally, consistently passes.

src/sql-server-util/src/inspect.rs
```diff
@@ -75,6 +75,55 @@ pub async fn get_max_lsn(client: &mut Client) -> Result<Lsn, SqlServerError> {
...
+/// This is based on the `sys.fn_cdc_get_min_lsn` implementation, which has logic
+/// that we want to bypass. Specifically, `sys.fn_cdc_get_min_lsn` returns NULL
+/// if the `start_lsn` in `cdc.change_tables` is less than or equal to the LSN
+/// returned by `sys.fn_cdc_get_max_lsn`.
+///
+/// See: <https://learn.microsoft.com/en-us/sql/relational-databases/system-tables/cdc-change-tables-transact-sql?view=sql-server-ver16>
+pub async fn get_min_lsns(
+    client: &mut Client,
+    capture_instances: impl IntoIterator<Item = &str>,
+) -> Result<Vec<Lsn>, SqlServerError> {
```
petrosagg: Is the intent here that the order in the returned vector matches the order passed in by the iterator? If yes, I don't think we get this since the SQL query below will return rows in some arbitrary order. Also, since we don't check that we got back all the capture instances we requested the returned vector might contain less items than expected.

src/sql-server-util/src/inspect.rs
```diff
@@ -75,6 +75,55 @@ pub async fn get_max_lsn(client: &mut Client) -> Result<Lsn, SqlServerError> {
...
+    let capture_instances: SmallVec<[_; 1]> = capture_instances.into_iter().collect();
+    #[allow(clippy::as_conversions)]
+    let values: SmallVec<[_; 1]> = capture_instances
+        .iter()
+        .map(|ci| ci as &dyn tiberius::ToSql)
```
petrosagg: If you want to avoid the annoying clippy lint you can do it like so:
```suggestion
        .map(|ci| {
            let ci: &dyn tiberius::ToSql = ci;
            ci
        })
```

src/sql-server-util/src/inspect.rs
```diff
@@ -75,6 +75,55 @@ pub async fn get_max_lsn(client: &mut Client) -> Result<Lsn, SqlServerError> {
...
+        .collect();
+    let args = capture_instances
+        .iter()
+        .enumerate()
+        .map(|(i, _)| format!("@P{}", i + 1))
```
petrosagg: If you only care about indices you can do it like so, which will allow you to only consume the iterator once, which will then allow you to avoid collecting the iterator on line 91.

```suggestion
    let args = (0..values.len())
        .map(|i| format!("@P{}", i + 1))
```

As a more general comment, this method is written in a style (with those `SmallVec`) that makes it seem like it's very hot. But my understanding is that it runs very rarely so it would be totally fine to request a slice/vec parameter and avoid all the `SmallVec`s and iterator args for readability.

src/sql/src/pure/sql_server.rs
```diff
@@ -164,6 +165,23 @@ pub(super) async fn purify_source_exports(
...
+    let mut possible_lsns = mz_sql_server_util::inspect::get_min_lsns(
+        client,
+        capture_instances.values().map(|instance| instance.as_ref()),
+    )
+    .await?;
+    possible_lsns.push(mz_sql_server_util::inspect::get_max_lsn_retry(client, timeout).await?);
+    let initial_lsn = possible_lsns
+        .iter()
+        .max()
+        .ok_or_else(SqlServerSourcePurificationError::NoInitialLsn)?;
```
petrosagg: This is an unreachable error since the list `possible_lsns` contains at least one element from line 179. (if that fails execution doesn't reach this point).

But also, I don't think we want to be mixing the minimum LSNs between capture instances. If one capture instance's min LSN is way ahead of `max_lsn` for some reason we don't want to set the `initial_lsn` of every other table to that point. Do we gain something by forcing everyone to start at the same point?

src/sql-server-util/src/inspect.rs
martykulma: No expectation of ordering at this time.

As far as the query returning less items, I can add a check to ensure the resulting vec length matches the number of capture instances. I didn't because:

1. we already had to query this table to get the capture instances, so they should be there
2. even if we detect that the capture instance was removed between discovery and calculating initial LSN, it has as much chance of being removed after this check and it will fail in snapshot.  I figured the chances are extremely unlikely and it would add yet another error path to test, so fall through seemed appropriate.

src/sql-server-util/src/inspect.rs
martykulma: nice - ty!

src/sql-server-util/src/inspect.rs
martykulma: good point!

src/sql-server-util/src/inspect.rs
petrosagg: > we already had to query this table to get the capture instances, so they should be there

Where do we do this? Should *that* place record the `start_lsn` then in a `struct CaptureInstance` so that we don't query twice?

src/sql/src/pure/sql_server.rs
martykulma: great point, I'll convert to expect.

Nothing really major is gained, AFAIK. To restate, the concern is that the `start_lsn` for some capture instance is arbitrarily far in the future. If `get_max_lsn` was going to return that large `start_lsn` on the very next invocation, that doesn't seem like it would be a problem.

So maybe there are some number of events at LSNs between the current LSN returned by `get_max_lsn` and the large `start_lsn`? I don't actually know how MS SQL chooses the that `start_lsn`, so it's possible that is the case. 

An example: 
I have 3 capture instances, [A, B, C] with start LSNs [10, 400, 9000], and a max LSN 33. For our example, A already exists, [B, C] are being added.

Given the current logic, we pick `initial_lsn = 9000` and `resume_lsn = 11`. We will have to issue rewind requests until we get to 9000 for both [B, C], and then we can downgrade the data capability.

If we track initial LSN separately, `resume_lsn` doesn't change and we still need to processed up to LSN 9000 to downgrade the capability. We issue less rewinds for B in this case.

src/sql-server-util/src/inspect.rs
martykulma: We could, I didn't only to stay consistent with MySQL.

When we [read source references](https://github.com/MaterializeInc/materialize/blob/f889e46faf9c4a157e8b563d8dae1c0d03e9225a/src/sql/src/pure/references.rs#L203)

Calls into [get_tables](https://github.com/MaterializeInc/materialize/blob/f889e46faf9c4a157e8b563d8dae1c0d03e9225a/src/sql-server-util/src/inspect.rs#L391-L419).

src/sql-server-util/src/inspect.rs
martykulma: fixed

src/sql-server-util/src/inspect.rs
martykulma: fixed

src/sql-server-util/src/inspect.rs
martykulma: Based on the changes for https://github.com/MaterializeInc/materialize/pull/33202#discussion_r2247930367, I modified this to return a `BTreeMap`.

Per our conversation, I'll look at refactoring to the reduce the calls to SQL server in a subsequent PR.

src/sql/src/pure/sql_server.rs
martykulma: updated!

martykulma: I'll need to give `resume_lsn` the same treatment!  https://buildkite.com/materialize/test/builds/107279#019866f2-f574-4490-9b34-b55ebc6644bb

martykulma:
don't mind the coverage test failure - i accidentally kicked off coverage pipeline and canceled it.

with `initial_lsn` and `resume_lsn` updated to be per capture instance, things are looking nice good:

https://buildkite.com/materialize/test/builds/107284
https://buildkite.com/materialize/test/builds/107285
https://buildkite.com/materialize/test/builds/107286
https://buildkite.com/materialize/test/builds/107287
https://buildkite.com/materialize/test/builds/107288
https://buildkite.com/materialize/test/builds/107289

martykulma:
Thanks for the reviews!

PR #33204, opened 2025-07-30
Add alias to current_catalog

SangJunBak:
Added aliases to be pg-compatible https://www.postgresql.org/docs/current/functions-info.html

### Motivation
https://github.com/MaterializeInc/database-issues/issues/9530

### Checklist


src/sql/src/plan/transform_ast.rs
```diff
@@ -455,9 +455,8 @@ impl<'a> FuncRewriter<'a> {
                 let ident = normalize::ident(ident[0].clone());
                 let fn_ident = match ident.as_str() {
                     "current_role" => Some("current_user"),
```
SangJunBak, line 457: Not related to this PR but I'm curious why we even need this line instead of just adding `current_role` to the union below given we already map `current_role` to `UnmaterializableFunc::CurrentUser => String` in `sql/src/func.rs`

SangJunBak, line 457: Not related to this PR but I'm curious why we even need this line instead of just adding `current_role` to the union below given we already map `current_role` to `UnmaterializableFunc::CurrentUser => String` in `sql/src/func.rs`

test/sqllogictest/current_database.slt
```diff
@@ -19,6 +19,11 @@ SELECT current_catalog() = current_database()
 ----
 true
 
+query T
+SELECT current_catalog = current_catalog()
+----
+true
+
```
ptravers: can we test all of:
```
current_catalog, current_role, current_schema, current_user, session_user, and user have special syntactic status in SQL: they must be called without trailing parentheses
```

particularly current_role

ptravers (approved):

test/sqllogictest/current_database.slt
SangJunBak: Yup! I think `current_role` is being tested here https://github.com/MaterializeInc/materialize/pull/33204/files#diff-faa33a7ee4e32a8f448b9bddb87fc4c5f2c766026971d629f9071531a7eafce7L23 but I can test it using the alias

ptravers:
looks good modulo adding additional test cases. also we should confirm that these are the only funcs that need to behave this way not just all functions without parameters. the last should not be a blocker.

SangJunBak:
> also we should confirm that these are the only funcs that need to behave this way not just all functions without parameters. the last should not be a blocker.

Judging from the spec in the description, it seems like it's only the following:

Screenshot 2025-07-30 at 6 46 25 PM

Also noticed:
> In PostgreSQL, parentheses can optionally be used with current_schema, but not with the others.

which we technically violate for some of the values like `current_role()`

SangJunBak:
Also realized there are a couple of discrepancies on how Materialize handles PG reserved func aliases. Will create a followup issue but here are the ones we should aim to be compatible with:

## Creating a table using `current_role` without quotes:
```sql
CREATE TABLE test (current_role TEXT);
```

### Postgresql
Fails
### Materialize
Succeeds

## Creating a table using `current_role` with quotes:
```sql
CREATE TABLE test ("current_role" TEXT);
INSERT INTO test VALUES('weewoo');
```

### Postgresql
```sql
SELECT current_role FROM test;
```
- If the table is empty, returns nothing
- If the table has entries, returns the current user
```sql
SELECT "current_role" FROM test;
```
- If the table has entries, returns "weewoo"

### Materialize
```sql
SELECT current_role FROM test;
```
- If the table is empty, returns nothing
- If the table has entries, returns the current user
```sql
SELECT "current_role" FROM test;
```
- If the table has entries, returns the current user

SangJunBak:
Realized adding the user keyword in the first commit actually breaks our audit log sql tests because one of the columns is named `user`. Given this compatibility bug is related to `current_catalog`, I've gotten rid of the `user`alias in this PR in the second commit and we can tackle it as a followup PR. 

One thing we'd need to look out for is any user who uses `current_catalog` as a column name will now face incorrectness. I think it'd be good to broadcast this somehow and make sure no customers are using this keyword. However, it does feel unlikely that users would have columns named `current_catalog`.

At the very least when we run `SELECT "current_role" from test;` in:

Screenshot 2025-07-30 at 7 35 47 PM

https://github.com/MaterializeInc/materialize/pull/33204#issuecomment-3138100673

, we should evaluate to "weewoo" rather than the current user.

SangJunBak:
@aljoscha Looked into how to fix
>At the very least when we run `SELECT "current_role" from test;`, we should evaluate to "weewoo" rather than the current user.

and my idea was to only transform the identifier if it's not quoted. It's a bit tricky however given the current lexer/parser treats it as the same token Ident. 

Even with this change, it'd still be good to warn users who are using `current_catalog` as a column. @ptravers had a good idea of feature flagging to users who rely on the JDBC driver to limit the blast radius, but unclear how to plumb to `transform_ast`.

SangJunBak:
Ah one thing we could do is look at the column names of all relations in everyone's environment and see if any are called current_catalog. Not sure we export that in the catalog exporter but we can always check via teleport.

We can also yolo it 👀

def-:
Cherry picked the tests from https://github.com/MaterializeInc/materialize/pull/33205 in here and triggered a nightly run (it's failing on main, should be fixed by this PR): https://buildkite.com/materialize/nightly/builds/12748

SangJunBak:
Checked in all prod accounts and none have identifier with name "current_catalog"

Link to logs: https://materializeinc.slack.com/archives/C08A62E0751/p1753972032784999?thread_ts=1753968756.764879&cid=C08A62E0751

I think it's safe to ship and followup

Edit: I'll timebox a bit more time and make sure that it still works with the newest version of metabase.

Edit: Realized it's only broken for Metabase. Gonna test on JDBC instead

def-:
This doesn't fail the java connection, but it gets one step further now:
https://buildkite.com/materialize/nightly/builds/12748#019860cf-a3ef-4df8-b4ff-06c4182feeb5
```
org.postgresql.util.PSQLException: ERROR: column "i.indnatts" does not exist
```
I'd argue we should also fix it in this PR until java can connect again.

SangJunBak:
@def- 
I agree. Fixing now!

def-:
This seems to come from https://github.com/pgjdbc/pgjdbc/blob/b1d3b710b2aabeb6055e25b28392df5e1f1fe64d/pgjdbc/src/main/java/org/postgresql/jdbc/PgDatabaseMetaData.java#L2427

ptravers (approved):

src/catalog/src/builtin.rs
```diff
@@ -6765,6 +6766,14 @@ SELECT
         WHEN class_objects.type = 'view' THEN 'v'
         WHEN class_objects.type = 'materialized-view' THEN 'm'
     END relkind,
+    COALESCE(
```
SangJunBak, line 6769: Needed to coalesce otherwise `show columns` would have this column show up as nullable

SangJunBak:
Created some followup issues:
- Querying function aliases https://github.com/MaterializeInc/database-issues/issues/9535
- COUNT subquery evaluating as nullable https://github.com/MaterializeInc/database-issues/issues/9536

sjwiesman:
🔥 Thank you all

PR #33207, opened 2025-07-31
sql_server: add checking permissions for all instances and tables in one query

ptravers:
### Motivation

  * This PR adds a known-desirable feature. https://github.com/MaterializeInc/database-issues/issues/9208

### Checklist

### DIFF
--- a/src/sql-server-util/src/inspect.rs
+++ b/src/sql-server-util/src/inspect.rs
@@ -9,6 +9,7 @@
 
 //! Useful queries to inspect the state of a SQL Server instance.
 
+use anyhow::Context;
 use futures::Stream;
 use itertools::Itertools;
 use mz_ore::cast::CastFrom;
@@ -592,4 +593,91 @@ where
             "expected 1 row, got {other:?}"
         ))),
     }
+}
+
+/// Return a Result that is empty if all tables, columns, and capture instances
+/// have the necessary permissions to and an error if any table, column,
+/// or capture instance does not have the necessary permissions
+/// for tracking changes.
+pub async fn validate_source_privileges<'a>(
+    client: &mut Client,
+    capture_instances: impl IntoIterator<Item = &str>,
+) -> Result<(), SqlServerError> {
+    let params: SmallVec<[_; 1]> = capture_instances.into_iter().collect();
+
+    if params.is_empty() {
+        return Ok(());
+    }
+
+    let params_dyn: SmallVec<[_; 1]> = params
+        .iter()
+        .map(|instance| {
+            let instance: &dyn tiberius::ToSql = instance;
+            instance
+        })
+        .collect();
+
+    let param_indexes = (1..params.len() + 1)
+        .map(|idx| format!("@P{}", idx))
+        .join(", ");
+
+    // NB(ptravers): we rely on HAS_PERMS_BY_NAME to check both table and column permissions.
+    let capture_instance_query = format!(
+            "
+        SELECT
+            SCHEMA_NAME(o.schema_id) + '.' + o.name AS qualified_table_name,
+            ct.capture_instance AS capture_instance,
+            COALESCE(HAS_PERMS_BY_NAME(SCHEMA_NAME(o.schema_id) + '.' + o.name, 'OBJECT', 'SELECT'), 0) AS table_select,
+            COALESCE(HAS_PERMS_BY_NAME('cdc.' + ct.capture_instance + '_CT', 'OBJECT', 'SELECT'), 0) AS capture_table_select
+        FROM cdc.change_tables ct
+        JOIN sys.objects o ON o.object_id = ct.source_object_id
+        WHERE ct.capture_instance IN ({param_indexes});
+            "
+        );
+
+    let rows = client
+        .query(capture_instance_query, &params_dyn[..])
+        .await?;
+
+    let mut capture_instances_without_perms = vec![];
+    let mut tables_without_perms = vec![];
+
+    for row in rows {
+        let table: &str = row
+            .try_get("qualified_table_name")
+            .context("getting table column")?
+            .ok_or_else(|| anyhow::anyhow!("no table column?"))?;
+
+        let capture_instance: &str = row
+            .try_get("capture_instance")
+            .context("getting capture_instance column")?
+            .ok_or_else(|| anyhow::anyhow!("no capture_instance column?"))?;
+
+        let permitted_table: i32 = row
+            .try_get("table_select")
+            .context("getting table_select column")?
+            .ok_or_else(|| anyhow::anyhow!("no table_select column?"))?;
+
+        let permitted_capture_instance: i32 = row
+            .try_get("capture_table_select")
+            .context("getting capture_table_select column")?
+            .ok_or_else(|| anyhow::anyhow!("no capture_table_select column?"))?;
+
+        if permitted_table == 0 {
+            tables_without_perms.push(table.to_string());
+        }
+
+        if permitted_capture_instance == 0 {
+            capture_instances_without_perms.push(capture_instance.to_string());
+        }
+    }
+
+    if !capture_instances_without_perms.is_empty() || !tables_without_perms.is_empty() {
+        return Err(SqlServerError::AuthorizationError {
+            tables: tables_without_perms.join(", "),
+            capture_instances: capture_instances_without_perms.join(", "),
+        });
+    }
+
+    Ok(())
 }
--- a/src/sql-server-util/src/lib.rs
+++ b/src/sql-server-util/src/lib.rs
@@ -638,9 +638,6 @@ impl Connection {
                 let result = client.execute(query, &params[..]).await?;
 
                 match result.rows_affected() {
-                    [] => Err(SqlServerError::InvariantViolated(
-                        "got empty response".into(),
-                    )),
                     rows_affected => {
                         let response = Response::Execute {
                             rows_affected: rows_affected.into(),
@@ -854,6 +851,13 @@ pub enum SqlServerError {
     Generic(#[from] anyhow::Error),
     #[error("programming error! {0}")]
     ProgrammingError(String),
+    #[error(
+        "insufficient permissions for tables [{tables}] or capture instances [{capture_instances}]"
+    )]
+    AuthorizationError {
+        tables: String,
+        capture_instances: String,
+    },
 }
 
 /// Errors returned from decoding SQL Server rows.
--- a/src/sql/src/pure/sql_server.rs
+++ b/src/sql/src/pure/sql_server.rs
@@ -147,8 +147,6 @@ pub(super) async fn purify_source_exports(
         }
     }
 
-    // TODO(sql_server2): Validate permissions on upstream tables.
-
     let capture_instances: BTreeMap<_, _> = requested_exports
         .iter()
         .map(|requested| {
@@ -164,6 +162,12 @@ pub(super) async fn purify_source_exports(
             (table.qualified_name(), Arc::clone(capture_instance))
         })
         .collect();
+
+    mz_sql_server_util::inspect::validate_source_privileges(
+        client,
+        capture_instances.values().map(|instance| instance.as_ref()),
+    )
+    .await?;
 
     // If CDC is freshly enabled for a table, it has been observed that
     // the `start_lsn`` from `cdc.change_tables` can be ahead of the LSN
--- a/test/sql-server-cdc/20-column-options.td
+++ b/test/sql-server-cdc/20-column-options.td
@@ -29,7 +29,6 @@ EXEC sys.sp_cdc_enable_table @source_schema = 'dbo', @source_name = 't2_columns'
 
 INSERT INTO t2_columns VALUES ('invisible');
 
-
 > CREATE SECRET IF NOT EXISTS sql_server_pass AS '${arg.default-sql-server-password}'
 
 > CREATE CONNECTION sql_server_columns_connection TO SQL SERVER (
@@ -73,11 +72,11 @@ INSERT INTO t1_columns VALUES (2.000001, '01:59:32.99901', '$0.89', 0x1122AABBCC
 2.0000010000 01:59:32.999010 ESKqu8wA3Q==
 <null> <null> <null>
 
+> DROP SOURCE t1_columns_sql_server CASCADE;
+
 ! CREATE SOURCE t2_columns_sql_server
   FROM SQL SERVER CONNECTION sql_server_columns_connection (
     EXCLUDE COLUMNS (dbo.t2_columns.c1)
   )
   FOR TABLES (dbo.t2_columns);
 contains:Table t2_columns had all columns excluded
-
-> DROP SOURCE t1_columns_sql_server CASCADE;
--- a/test/sql-server-cdc/21-privileges.td
+++ b/test/sql-server-cdc/21-privileges.td
@@ -1,0 +1,82 @@
+# Copyright Materialize, Inc. and contributors. All rights reserved.
+#
+# Use of this software is governed by the Business Source License
+# included in the LICENSE file at the root of this repository.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0.
+
+# Setup SQL Server state.
+#
+# Create a table that has CDC enabled.
+
+$ sql-server-connect name=sql-server
+server=tcp:sql-server,1433;IntegratedSecurity=true;TrustServerCertificate=true;User ID=${arg.default-sql-server-user};Password=${arg.default-sql-server-password}
+
+$ sql-server-execute name=sql-server
+USE test;
+CREATE TABLE t1_privileges (c1 int, c2 int);
+EXEC sys.sp_cdc_enable_table @source_schema = 'dbo', @source_name = 't1_privileges', @role_name = 'SA', @supports_net_changes = 0;
+
+CREATE LOGIN authorization_user_column_perms WITH PASSWORD = '${arg.default-sql-server-password}';
+CREATE USER authorization_user_column_perms FOR LOGIN authorization_user_column_perms;
+ALTER ROLE db_datareader ADD MEMBER authorization_user_column_perms;
+
+DENY SELECT ON OBJECT::dbo.t1_privileges(c2) TO authorization_user_column_perms;
+
+CREATE LOGIN authorization_user_column_perms_capture_instance WITH PASSWORD = '${arg.default-sql-server-password}';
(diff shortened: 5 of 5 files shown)


src/sql-server-util/src/lib.rs
```diff
@@ -352,6 +353,77 @@ impl Client {
...
+
+        for table in tables {
+            let schema_name = table.schema_name;
+            let table_name = table.name;
+            table_values.push_str(format!("('{schema_name}.{table_name}'),").as_str());
+        }
+
+        // FIXME(ptravers): gross way to remove the last comma.
+        table_values.pop();
+        table_values.push(')');
```
martykulma: rust slice has a join method that might be useful, for example you could do something like this:
```
        let table_values = format!(
            "VALUES({})",
            tables
                .iter()
                .map(|table| format!("('{}.{}')", table.schema_name, table.name))
                .collect::<Vec<_>>()
                .join(",")
        );
```

src/sql-server-util/src/lib.rs
```diff
@@ -352,6 +353,77 @@ impl Client {
...
+            UNION ALL
+            SELECT 
+                 name,
+                 CASE 
+                     WHEN HAS_PERMS_BY_NAME(name, 'OBJECT', 'SELECT') = 1 
+                     THEN 1
+                     ELSE 0
+                 END AS can_read
+             FROM ({table_values}) AS tables(name)
+            ",
```
martykulma: my understanding is that we are only checking to make sure the user can `SELECT` from the upstream table and the cdc table, but this does not validate that the columns are accessible.

Is the use of `CASE` to handle when `HAS_PERMS_BY_NAME` returns NULL? If that's the case (hah!), `COALESCE` will let you return a default in the case of a NULL.

An alternative would be to just pass in the capture instances (see an example of parameterized query [here](https://github.com/MaterializeInc/materialize/blob/86dd8a650724580035c0b87274a6ab050c46c209/src/sql-server-util/src/inspect.rs#L331)).  Not sure that this is better though.  The request size is smaller. 🤷 

```
SELECT
    SCHEMA_NAME(o.schema_id) + '.' + o.name as qualified_table_name,
    ct.capture_instance as capture_instance,
    COALESCE(HAS_PERMS_BY_NAME(SCHEMA_NAME(o.schema_id) + '.' + o.name, 'OBJECT', 'SELECT'), 0) AS table_select,
    COALESCE(HAS_PERMS_BY_NAME('cdc.' + ct.capture_instance + '_CT', 'OBJECT', 'SELECT'), 0) AS capture_table_select
FROM cdc.change_tables ct
    JOIN sys.objects ON o.object_id = c.source_object_id
WHERE ct.capture_instance IN (....);
```

ptravers: ah good shout. thanks!

src/sql-server-util/src/lib.rs
ptravers: ah good point. I knew there was a less java way to do this...

src/sql-server-util/src/lib.rs
ptravers: also does not validate we can get the lsn or increment the lsn

ptravers: I guess my question with the PR is how much are we trying to verify upfront?

src/sql-server-util/src/lib.rs
```diff
@@ -352,6 +353,73 @@ impl Client {
...
+            let permitted_capture_instance: i16 = row
+                .try_get(3)
+                .context("getting 3rd column")?
+                .ok_or_else(|| anyhow::anyhow!("no 3rd column?"))?;
+
+            if permitted_table == 0 || permitted_capture_instance == 0 {
+                return Err(SqlServerError::AuthorizationError {
+                    table: table.to_string(),
+                    capture_instance: capture_instance.to_string(),
+                });
```
ptravers: I am returning early at the first error but could collect all the errors. I'm just worried that the list could be massive and that it won't render for large numbers of tables. happy to refactor to return all the tables that failed permissions checks.

src/sql-server-util/src/lib.rs
ptravers: okay, so it's a story of two halves with pg and mysql:
1. for pg we check table, replication, and column permissions: https://github.com/MaterializeInc/materialize/blob/510a3a5f310d3fae55dd4d3ebf43cffbe1a06f03/src/postgres-util/src/schemas.rs#L47
2. for mysql we just check if we can select and lock the table but have an added check to see if we are a required role for obtaining cdc data:
https://github.com/MaterializeInc/materialize/blob/510a3a5f310d3fae55dd4d3ebf43cffbe1a06f03/src/mysql-util/src/privileges.rs#L19

I looked into column checks. It's a bit of a mess but what I can do is either:
1. add querying for select privileges on all columns included in the source.
2. query for any denied columns on tables included in the source.

src/postgres-util/src/schemas.rs
```diff
@@ -16,21 +16,7 @@ use tokio_postgres::Client;
...
-        .into_iter()
-        .map(|row| {
-            let oid: Oid = row.get("oid");
-            let name: String = row.get("nspname");
-            let owner: Oid = row.get("nspowner");
-            PostgresSchemaDesc { oid, name, owner }
-        })
-        .collect::<Vec<_>>())
-}
+use crate::desc::{PostgresColumnDesc, PostgresKeyDesc, PostgresTableDesc};
```
martykulma: I'm assuming that this isn't used anywhere?   It seems odd to be be refactoring code in postgres when you're working in SQL server, is there a reason to do it as part of this PR?

src/sql-server-util/src/lib.rs
martykulma: my understanding for PG and MySQL is that they grant/revoke access.  So if you have `SELECT` at the table level, revoking select at a column level doesn't have any effect.

SQL Server is different in this case because in addition to `GRANT`, it has `DENY`, which in most cases overrides `GRANT`. I would say you want to check specifically for `DENY` and could just amend your above query to do it.

src/sql-server-util/src/lib.rs
martykulma: no objections from me on just returning the first.

src/postgres-util/src/schemas.rs
ptravers: I accidentally included it sorry

src/sql-server-util/src/lib.rs
ptravers: yeah, that's correct. so I think we agree that option 2 makes sense and we can reconcile that with the column excludes during source creation.

src/sql/src/pure/sql_server.rs
```diff
@@ -193,13 +202,39 @@ pub(super) async fn purify_source_exports(
...
+                        column_names: not_excluded_denied_cols,
+                    })?
+                }
+
+                table.apply_excl_columns(excl_cols);
+            }
+            (Some(excl_cols), None) => {
+                tracing::debug!(?excl_cols, "found no denied columns");
+
+                table.apply_excl_columns(excl_cols);
```
ptravers: will remove debug logs before merge.

ptravers:
I believe that the test is failing due to:
https://github.com/MaterializeInc/database-issues/issues/9541

src/sql-server-util/src/lib.rs
```diff
@@ -352,6 +356,124 @@ impl Client {
             .collect();
         crate::cdc::CdcStream::new(self, instances)
     }
+
+    /// Return a Result that is empty if all tables and capture instances
+    /// have the necessary permissions to and an error if any table
+    /// or capture instance does not have the necessary permissions
+    /// for tracking changes.
```
martykulma: nit: May want to reword this. E.g.
```
Checks the permissions for source tables, CDC tables, and ... 

Return an empty map if all tables and capture instances have the necessary permissions, and an [`SqlServerError::AuthorizationError`] if not.
```

src/sql-server-util/src/lib.rs
```diff
@@ -352,6 +356,124 @@ impl Client {
...
+            .join(", ");
+
+        let capture_instance_query = format!(
+            "
+        WITH DeniedColumnsPerTable AS (
+            SELECT dp.major_id AS object_id, STRING_AGG(c.name, ',') AS denied_columns
+            FROM sys.database_permissions AS dp
+            JOIN sys.columns AS c ON dp.major_id = c.object_id AND dp.minor_id = c.column_id
+            WHERE dp.class_desc = 'OBJECT_OR_COLUMN' AND dp.state_desc = 'DENY' AND dp.permission_name = 'SELECT'
+            GROUP BY dp.major_id
```
martykulma: This will inadvertently pick up columns if any principal is denied SELECT on the columns.  `HAS_PERMS_BY_NAME` is doing the magic of ensuring the current principal has access to the tables.

src/sql-server-util/src/lib.rs
```diff
@@ -352,6 +356,124 @@ impl Client {
...
+        }
+
+        if !capture_instances_without_perms.is_empty() || !tables_without_perms.is_empty() {
+            return Err(SqlServerError::AuthorizationError {
+                tables: tables_without_perms.join(", "),
+                capture_instances: capture_instances_without_perms.join(", "),
+            });
+        }
+
+        Ok(denied_columns_per_table)
```
martykulma: My reading of this function is that on column permission issues, the function returns a success with map of columns that can't be read, or an `Err` with a collection of upstream / cdc tables that can't be read.

It would be really nice if this just returned an error if authorization failed, or success if things are good, rather than have to some authz checks in the function and some in the caller.

src/sql/src/pure/sql_server.rs
```diff
@@ -147,7 +148,15 @@ pub(super) async fn purify_source_exports(
...
-    // TODO(sql_server2): Validate permissions on upstream tables.
+    let capture_instances = requested_exports.iter().map(|export| {
+        export
+            .meta
+            .sql_server_capture_instance()
+            .expect("sql server source")
+            .as_ref()
+    });
+
+    let denied_cols_map = client.validate_source_privileges(capture_instances).await?;
```
martykulma: nit: if you move this down a few lines, you can avoid having the validate that the capture instance is set

```
    
    let denied_cols_map = client
        .validate_source_privileges(capture_instances.values().map(|instance| instance.as_ref()))
        .await?;
```

src/sql/src/pure/sql_server.rs
```diff
@@ -193,13 +202,33 @@ pub(super) async fn purify_source_exports(
...
+                        tbl_name: Arc::clone(&table.name),
+                        column_names: not_excluded_denied_cols,
+                    })?
+                }
+
+                table.apply_excl_columns(excl_cols);
+            }
+            (Some(excl_cols), None) => table.apply_excl_columns(excl_cols),
+            (None, Some(denied_cols)) => Err(SqlServerSourcePurificationError::ColumnsDenied {
+                tbl_name: Arc::clone(&table.name),
```
martykulma: can we include the schema as well?

test/sql-server-cdc/20-column-options.td
```diff
@@ -93,3 +99,17 @@ INSERT INTO t1_columns VALUES (2.000001, '01:59:32.99901', '$0.89', 0x1122AABBCC
...
+! CREATE SOURCE t3_columns_sql_server
+  FROM SQL SERVER CONNECTION sql_server_columns_connection
+  FOR TABLES (dbo.t3_columns);
+contains:Access denied for columns c2 on table t3_columns
+
+> CREATE SOURCE t3_columns_sql_server
+  FROM SQL SERVER CONNECTION sql_server_columns_connection (
+    EXCLUDE COLUMNS (dbo.t3_columns.c2)
+  )
+  FOR TABLES (dbo.t3_columns);
```
martykulma: Shouldn't this have failed if column `c2` could not be read?

src/sql-server-util/src/lib.rs
ptravers: very good catch...

test/sql-server-cdc/20-column-options.td
ptravers: since c2 is excluded it will succeed

src/sql/src/pure/sql_server.rs
ptravers: should we just do the fully qualified table name?

src/sql-server-util/src/lib.rs
ptravers: let me double check but I had thought this view is filtered by the current user.

src/sql-server-util/src/lib.rs
ptravers: good point!

src/sql/src/pure/sql_server.rs
martykulma: yes

src/sql/src/pure/sql_server.rs
ptravers: good point!

src/sql-server-util/src/lib.rs
ptravers: that makes sense. I will pass in a reference to the excluded columns map. the return an error from the validate function.

src/sql-server-util/src/lib.rs
```diff
@@ -352,6 +356,155 @@ impl Client {
...
+                          FROM sys.database_role_members
+                          WHERE member_principal_id IN (DATABASE_PRINCIPAL_ID(SUSER_NAME()), DATABASE_PRINCIPAL_ID(USER_NAME()))
+                          UNION ALL
+                          SELECT principal_id
+                          FROM sys.database_principals
+                          WHERE name = 'public'
+                      )
+                    )
+            GROUP BY dp.major_id
+        )
```
ptravers: note: refactor this and add a comment after CI passes repeatedly.

ptravers:
@petrosagg do I need to add the permissions check for transactional ddl?

test/sql-server-cdc/20-column-options.td
```diff
@@ -38,6 +38,20 @@ INSERT INTO t2_columns VALUES ('invisible');
...
+
+CREATE LOGIN authorization_user WITH PASSWORD = '${arg.default-sql-server-password}';
+CREATE USER authorization_user FOR LOGIN authorization_user;
+ALTER ROLE db_datareader ADD MEMBER authorization_user;
+CREATE ROLE deny_role;
+ALTER ROLE deny_role ADD MEMBER authorization_user;
+
+DENY SELECT ON OBJECT::dbo.t3_columns(c4) TO deny_role;
+DENY SELECT ON OBJECT::dbo.t3_columns(c3) TO authorization_user;
+DENY SELECT ON OBJECT::dbo.t3_columns(c2) TO public;
```
ptravers: very open to opinions on whether all of these are necessary. I think it would make sense to just have one column be denied to the user. denying a custom role and the public role isn't netting us as anything as we do not assert on anything except that the table cannot be replicated.

src/sql-server-util/src/lib.rs
```diff
@@ -638,9 +724,6 @@ impl Connection {
                 let result = client.execute(query, &params[..]).await?;
 
                 match result.rows_affected() {
-                    [] => Err(SqlServerError::InvariantViolated(
-                        "got empty response".into(),
-                    )),
```
ptravers, line 643: I don't think anything is broken by removing this invariant but I'm open to comment. empty responses are returned for permissions changes like the DENY statements in the tests.

src/sql-server-util/src/lib.rs
```diff
@@ -352,6 +354,90 @@ impl Client {
             .collect();
         crate::cdc::CdcStream::new(self, instances)
     }
+
+    /// Return a Result that is empty if all tables, columns, and capture instances
+    /// have the necessary permissions to and an error if any table, column,
+    /// or capture instance does not have the necessary permissions
+    /// for tracking changes.
+    pub async fn validate_source_privileges<'a>(
```
ptravers: should I move this to [inspect.rs](https://github.com/MaterializeInc/materialize/blob/8e5b4ed8e505135f2fea610965ae57abe28579dd/src/sql-server-util/src/inspect.rs#L4)?

ptravers:
blocked on: https://github.com/MaterializeInc/materialize/pull/33250

src/sql-server-util/src/lib.rs
martykulma: No objection.  Caller can decide if empty response is ok or not.

src/sql-server-util/src/inspect.rs
```diff
@@ -593,3 +594,89 @@ where
...
+        SELECT
+            SCHEMA_NAME(o.schema_id) + '.' + o.name AS qualified_table_name,
+            ct.capture_instance AS capture_instance,
+            COALESCE(HAS_PERMS_BY_NAME(SCHEMA_NAME(o.schema_id) + '.' + o.name, 'OBJECT', 'SELECT'), 0) AS table_select,
+            COALESCE(HAS_PERMS_BY_NAME('cdc.' + ct.capture_instance + '_CT', 'OBJECT', 'SELECT'), 0) AS capture_table_select
+        FROM cdc.change_tables ct
+        JOIN sys.objects o ON o.object_id = ct.source_object_id
+        WHERE ct.capture_instance IN ({param_indexes});
+            "
+        );
```
martykulma, line 636: Could do with a comment that we expect `HAS_PERMS_BY_NAME` to validate permission on the table and all the columns.

test/sql-server-cdc/21-privileges.td
```diff
@@ -0,0 +1,71 @@
...
+    PASSWORD = SECRET sql_server_pass
+  );
+
+# Column exclusion has no effect on rejection for column level permissions.
+! CREATE SOURCE t1_privileges_sql_server_failure
+  FROM SQL SERVER CONNECTION sql_server_privileges_connection_column (
+    EXCLUDE COLUMNS (dbo.t1_privileges.c2)
+  )
+  FOR TABLES (dbo.t1_privileges);
+contains:insufficient permissions for tables [dbo.t1_privileges] or capture instances []
```
martykulma, line 69: Should this test (or another) cover the case where we report a capture instance in the error?

src/sql-server-util/src/inspect.rs
ptravers: will add, good point.

test/sql-server-cdc/21-privileges.td
ptravers: probably, will add.

ptravers: done

src/sql-server-util/src/inspect.rs
ptravers: done

PR #33208, opened 2025-07-31
cluster: don't panic when dropping `TimelyContainer`

teskje:
The panic was originally added to alert us to concurrency bugs where we would accidentally drop the Timely runtime while trying to switch between controller connections. The surrounding code has been simplified a lot since, so the panic isn't as important as a safeguard anymore.

More importantly though, since the Timely runtime is now initialized before we start serving controller connections, it is now actually possible to drop the `TimelyContainer` without any bugs present:

 * Timely runtime is initialized, yielding a `TimelyContainer`.
 * The `TimelyContainer` is passed to `GrpcServer::serve`, which starts to listen for controller connections.
 * The listen fails, for example because the listen port is already in use.
 * `GrpcServer::serve` returns with an error, dropping the `TimelyContainer`.

In this scenario, the process exits anyway, so dropping the `TimelyContainer` is fine and we shouldn't panic. The easiest fix is to simply remove the panic.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9527

### Checklist

### DIFF
--- a/src/cluster/src/client.rs
+++ b/src/cluster/src/client.rs
@@ -72,12 +72,6 @@ impl<C: ClusterSpec> TimelyContainer<C> {
     }
 }
 
-impl<C: ClusterSpec> Drop for TimelyContainer<C> {
-    fn drop(&mut self) {
-        panic!("Timely container must never drop");
-    }
-}
-
 impl<C> ClusterClient<C>
 where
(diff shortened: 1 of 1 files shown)


antiguru (approved):

teskje:
TFTR!

PR #33213, opened 2025-07-31
misc: don't specify a memory limit for replicas by default

teskje:
We are specifying a default memory limit of 4GiB for all replicas created in mzcompose tests, even for replica sizes that don't explicitly mention a limit (like '1'). This turns out to be a bit of a footgun as tests might not expect the limit and become unreliable and/or slow by getting their replicas intermittently killed.

The default memory limit was introduced to ensure that `bin/environmentd` can be run without an unlimited license key. We can also achieve that by specifically adjusting the replica sizes passed under that command to have a default memory limit, while leaving the sizes used by mzcompose unlimited.

### Motivation

Some CI workloads are unreliable because they don't expect the default memory limit placed on replica sizes.
See [Slack thread](https://materializeinc.slack.com/archives/C01LKF361MZ/p1753964208658369).

### Checklist

### DIFF
--- a/misc/python/materialize/cli/run.py
+++ b/misc/python/materialize/cli/run.py
@@ -23,6 +23,7 @@ import tempfile
 import time
 import uuid
 from datetime import datetime, timedelta
+from typing import Any
 from urllib.parse import urlparse
 
 import psutil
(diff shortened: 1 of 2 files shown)


def- (approved):
Thanks

teskje:
I didn't merge this because it would have made some `mzcompose` tests require a license key, making for a bad developer experience. License checks are disabled for tests now, so the issue this was meant to fix should be resolved now.

def-:
Many mzcompose-based tests require a license key already: https://github.com/MaterializeInc/materialize/blob/482a795e39ab4e447a6ae3f638a89a031b403f6b/doc/developer/guide.md#running-and-connecting-to-local-materialize

PR #33214, opened 2025-07-31
test/cluster: recreate test cluster between memory limiter tests

teskje:
The memory/lgalloc limiter cluster tests were flaky because they run a replica under different limit configurations and relied on timing to ensure the new configuration has been picked up. This PR changes the tests to instead create a new cluster between changing the limit config and restarting the replica process. This will hopefully ensure that the replica always receives the correct config after restart.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9502

### Tips for reviewer

I tried implementing https://github.com/MaterializeInc/database-issues/issues/9483 to fix this, but it turns out to be more complicated than I expected. So this rewrites the test instead, to get the flake fixed quickly.

### Checklist


teskje:
@def- Looks like the flake has so far only occurred on test branches of yours. Any chance you could take this fix for a spin on one of them, to verify it really solves the issue?

def-:
The commit that caused this to fail will most likely not be merged.

teskje:
Ok, still seems like a good idea to merge this. I don't see a reason why the flake should be impossible on main today.

def- (approved):

PR #33216, opened 2025-07-31
Make `MSE::reduce` treat nullability reliably

frankmcsherry:
Our `MSE::reduce` method makes irreversible transformations based on nullability information present in the type, but this information is not durable: we may manipulate the surrounding expressions and make it no longer true, and present the transformed expression with null values in these columns. Some transformations we perform change the expression behavior in that case, e.g.
1. from not erroring due to an `If` guard, to erroring when the guard is optimized away, or
2. by pruning terms in a `COALESCE`, even though at runtime they may need to be produced.

The second one may be safe (errors seem to be the main "side effect" we can cause), but there's something about COALESCE's evaluation order that worries me, and I can't map the equivalent `CASE` misoptimization onto it.

### Checklist


src/expr/src/scalar.rs
```diff
@@ -799,7 +801,7 @@ impl MirScalarExpr {
     /// test.reduce(&input_type);
     /// assert_eq!(test, expr_t);
     /// ```
-    pub fn reduce(&mut self, column_types: &[ColumnType]) {
+    pub fn reduce<const NULLABILITY: bool>(&mut self, column_types: &[ColumnType]) {
```
petrosagg, line 804: Is this function super hot? Putting the parameter as a const will generate two copies of the function and makes it a bit awkward to call. The pattern we use in other places of the codebase is to define a tiny enum like `enum Nullability { Ignore, RelyOn }` and add a parameter.

ggevay:
(I'm heading out to a holiday for Friday and Monday, but will definitely look afterwards.)

src/expr/src/scalar.rs
```diff
@@ -849,10 +851,29 @@ impl MirScalarExpr {
...
+                            Some(vec![then, els])
+                        }
+                        MirScalarExpr::CallVariadic { func, exprs } => {
+                            if let VariadicFunc::Coalesce = func {
+                                if !exprs.is_empty() {
+                                    let len = exprs.len();
+                                    for expr in exprs.iter_mut().take(len - 1) {
+                                        expr.reduce::<false>(column_types);
+                                    }
+                                    Some(vec![&mut exprs[len - 1]])
```
ggevay, line 882: I think a code comment would be good to explain why we recurse with false into all but the last argument of `coalesce`.

Also, I'm not sure what is the intended effect of this. If it is to prevent the
```
// Find the first argument that is a literal or non-nullable
// column. All arguments after it get ignored, so throw them
// away. This intentionally throws away errors that can
// never happen.
```
throw-away in line 1246 from happening due to type information on all but the last argument (where there is nothing to throw away), I don't think it achieves this: line 861 and 1246 will happen in the same `reduce` call, so even if we recurse into the arguments with false, we'll later come back to the same `reduce` call and do line 1246 with true. In other words, the `NULLABILITY &&` check in line 1251 will have an effect on `coalesce` calls that are inside other `coalesce` calls (or `If`s), but not on outermost `coalesce` calls.

src/expr/src/scalar.rs
ggevay: Also note that it's called with false only in recursive calls; all external calls are with true. So, we could also create a wrapper function that doesn't have this parameter, and calls the actual function with constant true, and call that wrapper function in all external call sites.

frankmcsherry: It was definitely done this way to force the caller to think, especially as the `true` variant (that I imagine we'd have as the default with a wrapper) is the generally not safe thing to call. Every call site was updated in the PR, so .. it's not that many. The belief is that it is hot, yes. In many profiles we spend a lot of time here, and will until we are able to rewrite it as pattern match rules that we can seek to the relevant operators.

src/expr/src/scalar.rs
frankmcsherry: The main reason this exists is that until we have a specific justification for optimizing control flow expressions, we shouldn't. Imo, we wait until then, e.g. until we have a clearer understanding of the intended semantics, to perform potentially incorrect transforms. When that happens it may turn out that this was unnecessary, but I'm personally more comfortable with that outcome than discovering that our intuition around our semantics was imperfect in the other direction.

Happy to add that as a comment, though!

src/expr/src/scalar.rs
frankmcsherry: For example, we don't call `MSE::reduce` in `column_knowledge.rs`'s handling of `MSE::If`, but had we done so we would need to have used the `::<false>` variant, rather than auto-piloting into the use of `::<true>`.

frankmcsherry: (( point being: build interfaces not to support the way the code happens to be used, but to enforce the way it should be used. A default of `false` avoids misoptimization when one doesn't think too hard, but a default of `true` makes it easier. The goal isn't to minimize the number of characters in the codebase, but the number of errors. ))

ggevay: Ok, makes sense to me!

src/expr/src/scalar.rs
```diff
@@ -858,6 +858,12 @@ impl MirScalarExpr {
                             Some(vec![then, els])
                         }
                         MirScalarExpr::CallVariadic { func, exprs } => {
+                            // `Coalesce` conditionally evaluates its arguments based on their nullability.
+                            // The thinking is that the final term will be unconditionally evaluated, once
+                            // reached, which makes it suitable for optimization that relies on unstable
+                            // information like nullability. With further detail about the evaluation order
+                            // semantics, we may be able to simplify this logic, but until we are certain
+                            // we can avoid optimizing the terms based on an ephemeral nullability reading.
```
ggevay: > The thinking is that the final term will be unconditionally evaluated, once
> reached, 

Well, unless I'm misunderstanding something, this is true for all arguments: once an argument is reached, it is unconditionally evaluated. Maybe you meant something like "the final argument can't prevent more arguments from being evaluated"?

src/expr/src/scalar.rs
ggevay: Not optimizing control flow expressions seems like a fine precaution to me, but could you double-check and confirm that the code does what you intended it to do?

As I mentioned above, the `NULLABILITY &&` check in `coalesce` optimization will _not_ prevent any truncation on outermost `coalesce` calls. I can't imagine why it could be the intended behavior that we want truncations prevented in `coalesce` calls nested inside other `coalesce` calls or `If`s, but not in outermost `coalesce` calls.

The other effect of making the recursive call with `<false>` is that `IS NULL` checks are not optimized to `false` in `coalesce`'s non-last arguments. This one I also can't see as being the intended purpose of the code, because I don't understand why this would help with avoiding error problems. In `If` conditions, the same thing helps because bools are being evaluated there on whether they are true or false, but `coalesce` doesn't care about bool args being true or false; it just checks null-ness, which won't be affected by optimizing `IS NULL` checks to `false`.

frankmcsherry: Ah, ok so if I understand correctly, I think your point is that the highlighted code is .. whatever. Needs explanation. But the actual concern is around the lines 1256 which potentially should just unconditionally ignore nullability (in my worldview, where if we do not know it is safe to do we should not do it). Let's break those apart:

1. I agree that the `::<false>` here is not what prevents coalesce pruning. What this means to do is avoid locking in the  outcome of control flow that should only be resolved at runtime. For example, if the expressions in `COALESCE` would test for nullability, and say produce a null output on non-null input but a non-null output on a null input, this optimization would currently install that literal null and then remove it, even if the expression acted as a guard at runtime against subsequent evaluation (a null input, which we might see at runtime, would now reach a later expression and potentially error). I think there is a concrete expression I can put in the documentation here that will incorrectly error post-transformation, that we need to protect against.
2. The 1256 area should just unconditionally ignore nullability, as pruning expressions and then encountering a null at runtime can propagate a null upwards, into a type that may have been determined should be non-null (e.g. perhaps the coalesce ends with a literal non-null, but we then pruned it). We would have no idea what would happen in that case, and any number of contracts would be broken.
3. I agree that there is little to no relationship between these two hunks of code, but there was also little clarity about that in the PR.

frankmcsherry: Concretely, I think this fragment has the property that we need to avoid optimizing any but the last expression of the coalesce with nullability information that may not remain true. If the first argument is unoptimized it will not error, but if it is optimized (to a literal null) it can error. (( and if the second argument is equally optimized, it would not error, but I don't think we want to rely on that happening ))
```sql
COALESCE(
    CASE WHEN x IS NULL THEN 1 ELSE NULL END,
    CASE WHEN x IS NULL THEN 1/0 ELSE 1 END
)
```

src/expr/src/scalar.rs
frankmcsherry: I think that is what I meant. I have to admit not having the right words to describe the bad thing that happens.

src/expr/src/scalar.rs
frankmcsherry: I'll update the comments, but also push the change that disables the coalesce pruning, which .. has the potentially to introduce regressions. It should be safe if subsequent expressions cannot error, but I can imagine that we aren't precise enough about that at the moment to recover the optimizations we "unsafely" make at the moment.

frankmcsherry: Pruning removal pushed. I said above it should be safe if subsequent expressions cannot error, but I think that is incorrect. The problem is more that the pruning may introduce null outputs, and we don't have a handle on which consumers of the column may error, panic, or whatever on null values they have been promised won't appear.

frankmcsherry: Yup, lots of errors now. We should discuss to see whether we prefer the potential regressions, or the potential misoptimizations.

PR #33233, opened 2025-08-04
sql: propagate retention windows by persisting them in SQL

petrosagg:
We previously propagated source retention windows from the source to all the subsources, forcing them to be the same. In the transactional DDL and "CREATE TABLE FROM SOURCE" world this doesn't make sense. Each table from source is its own object with its own retention window.

After this commit we always use the retention window used in the `CREATE TABLE FROM SOURCE` statement and for the old syntax we propagate it during planning. The end result is that each object contains its own retention window in the catalog.

### Checklist


test/sqllogictest/retain_history.slt
```diff
@@ -163,19 +163,19 @@ JOIN mz_objects o ON o.id = h.id
...
 counter  FOR  60000
 counter_progress  FOR  1000
 idx_a  FOR  420000
 idx_b  FOR  360000
 idx_c  FOR  240000
-organizations  FOR  60000
+organizations  FOR  1000
 tab_a  FOR  1000
-users  FOR  60000
+users  FOR  1000
```
petrosagg, line 178: This test exercises `ALTER SOURCE ... RETAIN HISTORY` and expects altering the `RETAIN HISTORY` setting on the source object to propagate to all subsources. I removed this functionality since this is not how it supposed to work in the future and no one is using `RETAIN HISTORY` on sources anyway.

aljoscha (approved):
Looks easy enough, and I'm happy when the migration tests are green

PR #33247, opened 2025-08-06
Expose channel type

antiguru:
Exposes the channel type in introspection relations, as presented by Timely. The channel type is the type of the containers on the channel.

### DIFF
--- a/src/compute/src/logging/timely.rs
+++ b/src/compute/src/logging/timely.rs
@@ -14,7 +14,7 @@ use std::collections::BTreeMap;
 use std::rc::Rc;
 use std::time::Duration;
 
-use differential_dataflow::consolidation::ConsolidatingContainerBuilder;
+use columnar::Columnar;
 use differential_dataflow::containers::{Columnation, CopyRegion};
 use mz_compute_client::logging::LoggingConfig;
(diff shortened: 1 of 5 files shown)


src/compute/src/logging/timely.rs
```diff
@@ -162,23 +164,29 @@ pub(super) fn construct<G: Scope<Timestamp = Timestamp>>(
...
-                    Datum::UInt64(u64::cast_from(worker_id)),
-                    Datum::UInt64(u64::cast_from(source_node)),
-                    Datum::UInt64(u64::cast_from(source_port)),
-                    Datum::UInt64(u64::cast_from(target_node)),
-                    Datum::UInt64(u64::cast_from(target_port)),
-                ]);
-                session.give((data, time, diff));
-            },
-        );
+        // TODO: Consider adding eager consolidation once we deprecate columnation
```
teskje: Is eager consolidation blocked by removing columnation? Looks like the issue is more that we don't support it with columnar?

teskje (approved):
LGTM!

src/compute/src/logging/timely.rs
antiguru: Fixed the comment!

PR #33250, opened 2025-08-06
sql server tests: Introduce auto-updating dummy table (by Petros)

def-:
The code is by petros taken from https://github.com/MaterializeInc/database-issues/issues/9541#issuecomment-3159522021 I just made it work with Testdrive and Tiberius (simple_query) and integrated in mzcompose.py.

We now only use one test database instead of creating many, is that ok? I didn't get all tests green again, I guess related to that. So maybe we want to switch back to having multiple databases?

Fixes: https://github.com/MaterializeInc/database-issues/issues/9541

### Checklist


misc/python/materialize/mzcompose/services/sql_server.py
```diff
@@ -41,6 +41,11 @@ def __init__(
                     f"SA_PASSWORD={sa_password}",
                     *environment_extra,
                 ],
+                "healthcheck": {
```
def-: Locally sql server didn't start up fast enough for me so I couldn't test at all.

martykulma:
My understanding is that this will cause all tests to share the same DB without recreating it, is that correct?  I'd rather have each test run in a fresh environment because I have no idea what cruft might get left over from other tests.

def-:
Ok, will change it. Edit: This is actually a bit awkward since I guess the dummy table has to live in that database to have an effect. @petrosagg any ideas?

petrosagg:
> Ok, will change it. Edit: This is actually a bit awkward since I guess the dummy table has to live in that database to have an effect. @petrosagg any ideas?

My original plan was to run `setup.td` before each actual `td` file ran by mzcompose so that we always have a fresh db. Is this possible to do?

def-:
Yes, it's possible, but we have to pass in the db name to the setup.td script to make sure to create the correct database, right?

petrosagg:
> Yes, it's possible, but we have to pass in the db name to the setup.td script to make sure to create the correct database, right?

If we have tests that do meaningful things with the db name, yes. Otherwise we can introduce a convention and say all tests should assume that they operate on db name X

def-:
I was struggling with making the setup.td clean up the existing DummyTicker job when I tried that. Other than that should be possible. I'm not sure if we still need this PR or not though?

martykulma:
> I was struggling with making the setup.td clean up the existing DummyTicker job when I tried that. Other than that should be possible. I'm not sure if we still need this PR or not though?

If we're no longer seeing the long runs after https://github.com/MaterializeInc/materialize/pull/33249 merged, I would hold off.

ptravers:
@martykulma if we could solve it this way it would save me a lot of typing when porting all the tests across from Postgres.

def-:
Ok, I'll give it another try to make this work.

test/sql-server-cdc/setup/setup.td
```diff
@@ -0,0 +1,58 @@
...
+    SELECT 1
+    FROM msdb.dbo.sysjobs
+    WHERE name = N'DummyTicker'
+)
+BEGIN
+    EXEC sp_delete_job @job_name = N'DummyTicker';
+END;
+
+$ sql-server-execute name=sql-server
+CREATE DATABASE test COLLATE Latin1_General_100_CI_AI_SC_UTF8;
```
def-, line 32: Are we ok with making this the default collation? I had trouble altering the collation, it seems to mess with the dummy ticker job.

martykulma: Yes, just have to update the test to not expect `??` from materialize.

def-: also just noticed that, done. RFR

ptravers (approved):

PR #33254, opened 2025-08-07
sql server tests & testdrive: Improvements

def-:
Noticed in https://github.com/MaterializeInc/materialize/pull/33250

Less controversial than the rest of that PR, so split them out

### Checklist

### DIFF
--- a/src/testdrive/src/action/sql_server/execute.rs
+++ b/src/testdrive/src/action/sql_server/execute.rs
@@ -18,6 +18,7 @@ pub async fn run_execute(
     state: &mut State,
 ) -> Result<ControlFlow, anyhow::Error> {
     let name = cmd.args.string("name")?;
+    let split_lines = cmd.args.opt_bool("split-lines")?.unwrap_or(true);
     cmd.args.done()?;
 
     let client = state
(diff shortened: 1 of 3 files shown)


src/testdrive/src/action/sql_server/execute.rs
```diff
@@ -18,17 +18,29 @@ pub async fn run_execute(
...
+    let split_lines = cmd.args.opt_bool("split-lines")?.unwrap_or(true);
     cmd.args.done()?;
 
     let client = state
         .sql_server_clients
         .get_mut(&name)
         .ok_or_else(|| anyhow!("connection {} not found", name.quoted()))?;
 
-    for query in cmd.input {
+    if split_lines {
```
petrosagg, line 29: Do we get any benefit from sending them line by line? Sounds like we could always send the entire batch and have one less parameter to worry about

def-: I think it didn't work, but let me try again and check why, might have been something else in my other PR causing it.

def-: Indeed, what happens is when we run this for example:
```
DROP DATABASE IF EXISTS consistency_test;
CREATE DATABASE consistency_test;
USE consistency_test;
ALTER DATABASE consistency_test SET ALLOW_SNAPSHOT_ISOLATION ON;
EXEC sys.sp_cdc_enable_db;
CREATE TABLE t1 (id bigint, val bigint);
EXEC sys.sp_cdc_enable_table @source_schema = 'dbo', @source_name = 't1', @role_name = 'SA', @supports_net_changes = 0;
WITH nums AS (SELECT 1 as n UNION ALL SELECT n+1 FROM nums where n < 100) INSERT INTO t1 SELECT n, n+1000 FROM nums;
```
The `USE consistency_test` fails:
```
2025-08-07T07:31:21.221536Z ERROR tiberius::tds::stream::token: Database 'consistency_test' does not exist. Make sure that the name is entered correctly. code=911
```

PR #33260, opened 2025-08-07
Rework peek handling

antiguru:
Start using containers in more places: use a KeyContainer to store the literal rows we intend to lookup in the target arrangement instead of keeping the original rows. Not intended to change performance, just good practice, and gives us confidence that it's the right way to implement interactions with arrangements.

### DIFF
--- a/src/compute/src/compute_state.rs
+++ b/src/compute/src/compute_state.rs
@@ -1528,6 +1528,8 @@ impl IndexPeek {
         let max_result_size = usize::cast_from(max_result_size);
         let count_byte_size = size_of::<NonZeroUsize>();
 
+        // We clone `literal_constraints` here because we don't want to move the constraints
+        // out of the peek struct, and don't want to modify in-place.
         let mut peek_iterator = peek_result_iterator::PeekResultIterator::new(
             peek.target.id().clone(),
             peek.map_filter_project.clone(),
--- a/src/compute/src/compute_state/peek_result_iterator.rs
+++ b/src/compute/src/compute_state/peek_result_iterator.rs
@@ -214,6 +214,8 @@ where
         let row_item = self.cursor.val(&self.storage);
         let row = row_item.to_datum_iter();
 
+        // An optional literal that we might have added to the borrow. Needs to be declared
+        // before the borrow to ensure correct drop order.
         let maybe_literal;
         let mut borrow = self.datum_vec.borrow();
         borrow.extend(key);
@@ -224,9 +226,6 @@ where
         {
             // The peek was created from an IndexedFilter join. We have to add those columns
             // here that the join would add in a dataflow.
-            // unwrap is ok, because it could be None only if current_literals is exhauseted or if
-            // the iteration is finished. In the latter case we already exited the while
-            // loop.
             maybe_literal = literal;
             borrow.extend(maybe_literal.to_datum_iter());
         }


src/compute/src/compute_state/peek_result_iterator.rs
```diff
@@ -169,18 +214,21 @@ where
...
 
-        if self.has_literal_constraints {
+        if let Some(literals) = &mut self.literals
+            && let Some(literal) = literals.peek()
+        {
             // The peek was created from an IndexedFilter join. We have to add those columns
             // here that the join would add in a dataflow.
-            let datum_vec = borrow.deref_mut();
-            // unwrap is ok, because it could be None only if !has_literal_constraints or if
+            // unwrap is ok, because it could be None only if current_literals is exhauseted or if
```
bkirwi: IIUC this was about the `unwrap` in the following code, which has been removed? (Which is great!)

src/compute/src/compute_state/peek_result_iterator.rs
```diff
@@ -22,16 +22,71 @@ where
...
+    current_index: Option<usize>,
+}
+
+impl<Tr: TraceReader<KeyOwn: Ord>> Literals<Tr> {
+    /// Construct a new `Literals` from a mutable slice of literals. Sorts contents.
+    fn new(literals: &mut [Tr::KeyOwn], cursor: &mut Tr::Cursor, storage: &Tr::Storage) -> Self {
+        // We have to sort the literal constraints because cursor.seek_key can
+        // seek only forward.
+        literals.sort();
+        let mut container = Tr::KeyContainer::with_capacity(literals.len());
```
bkirwi, line 50: I would have guessed that `literals` was pretty short and so the container wouldn't matter terribly. Is that right / is this more from a goal of columnarizing all the things?

src/compute/src/compute_state.rs
```diff
@@ -1532,7 +1532,7 @@ impl IndexPeek {
             peek.target.id().clone(),
             peek.map_filter_project.clone(),
             peek.timestamp,
-            peek.literal_constraints.clone(),
+            peek.literal_constraints.clone().as_deref_mut(),
```
bkirwi, line 1537: (eg. If literal constraints are large enough for storage to matter, not sure if we care about how these are now copied twice.)

src/compute/src/compute_state/peek_result_iterator.rs
ggevay: Should be not more than a few thousand literals. See e.g. https://github.com/MaterializeInc/materialize/blob/833454ccc98c1b8fda03c26448edb4959a327434/test/sqllogictest/transform/literal_constraints.slt#L31

antiguru: Correct; it's mostly about moving to newer Differential APIs. I don't think this would change performance in any meaningful way.

src/compute/src/compute_state.rs
antiguru: Ideally, we'd never clone. I left this clone in place because we're retaining the raw peek here, even though we only use a subset of the fields later on. I felt this is the least confusing approach if we ever were to look at the literal constraints field in the future. I'll leave a comment to this note.

src/compute/src/compute_state/peek_result_iterator.rs
antiguru: Removed comment, thanks for the call out.

PR #33263, opened 2025-08-07
Use a unique path for rocksdb even if using mem env

martykulma:
When using `mem_env`, RocksDB still attempts to create a lock file, which can cause issues with multiple workers.

```
materialized2-1  | 2025-08-07T15:37:16.862948167Z cluster-u2-replica-u3-gen-3: 2025-08-07T15:37:16.862935Z ERROR mz_rocksdb: failed to create rocksdb at /tmp: IO error: /tmp/LOCK: lock is already held.
```

### Motivation

I just felt like it.

### Checklist

### DIFF
--- a/src/storage/src/upsert.rs
+++ b/src/storage/src/upsert.rs
@@ -254,11 +254,10 @@ where
 
     let tuning = dataflow_paramters.upsert_rocksdb_tuning_config.clone();
 
-    // When running RocksDB in memory, the file system is emulated, so the path doesn't
-    // matter. However, we still need to pick one that exists on the host because of
-    // https://github.com/rust-rocksdb/rust-rocksdb/issues/1015.  RocksDB will still
-    // create a lock file at this path, so we need to ensure that whatever path is used
-    // will be unique per worker.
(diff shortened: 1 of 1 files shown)


src/storage/src/upsert.rs
```diff
@@ -253,19 +254,19 @@ where
...
-            // matter. However, we still need to pick one that exists on the host because of
-            // https://github.com/rust-rocksdb/rust-rocksdb/issues/1015.
-            "/tmp".into()
-        }
-    };
+    // When running RocksDB in memory, the file system is emulated, so the path doesn't
+    // matter. However, we still need to pick one that exists on the host because of
+    // https://github.com/rust-rocksdb/rust-rocksdb/issues/1015.  RocksDB will still
+    // create a lock file at this path, so we need to ensure that whatever path is used
+    // will be unique per worker.
```
teskje: Should we remove the first part of the comment? Seems like it isn't true that the path isn't used and therefore doesn't matter.

martykulma: Yes, I should have just rewritten the comment.  That initial blurb was just a note to myself.  Updated now, thank you!

PR #33267, opened 2025-08-07
[persist] Fetch fewer diffs during GC

bkirwi:
While we recently switched to limiting the amount of seqnos that can be GC'd in one pass, we'd still end up fetching all the diffs in every GC run, because we didn't have a good way to get both the initial state and a set of follow-up diffs otherwise. This PR switches to grabbing the appropriate rollup from the current state... which we're already loading earlier in the method.

### Motivation

It's twice now we've seen high memory pressure due to a stuck shard suddenly getting unstuck. This helps us bound the amount of memory used in such circumstances...

### Tips for reviewer

A few refactorings as well, which hopefully speak for themselves.

I put the new behaviour behind the active-compaction flag; that flag is currently CI-only and pairs well with the changes here I think.

### DIFF
--- a/src/persist-client/src/internal/gc.rs
+++ b/src/persist-client/src/internal/gc.rs
@@ -40,8 +40,8 @@ use crate::internal::machine::{Machine, retry_external};
 use crate::internal::maintenance::RoutineMaintenance;
 use crate::internal::metrics::{GcStepTimings, RetryMetrics};
 use crate::internal::paths::{BlobKey, PartialBlobKey, PartialRollupKey};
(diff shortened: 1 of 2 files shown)


src/persist-client/src/internal/datadriven.rs
```diff
@@ -190,98 +190,110 @@ mod tests {
...
-                f
-            }
-        })
+                    })
+                    .await;
+                    f
+                }
+            },
+            |m| {
+                std::env::var("DATADRIVEN_FILTER")
```
DAlperin: nice

PR #33270, opened 2025-08-08
Fix warnings reported by Rust 1.89

antiguru:
As it says on the box, fix warnings generated by Rust 1.89. The PR doesn't upgrade to Rust 1.89 yet.

### DIFF
--- a/src/expr/src/scalar/func/snapshots/mz_expr__scalar__func__binary__test__fallible1.snap
+++ b/src/expr/src/scalar/func/snapshots/mz_expr__scalar__func__binary__test__fallible1.snap
@@ -1,6 +1,6 @@
 ---
 source: src/expr/src/scalar/func/binary.rs
-expression: "#[sqlfunc()]\nfn fallible1(a: f32, b: f32) -> Result<f32, EvalError> {\n    Ok(a + b)\n}\n"
+expression: "#[sqlfunc()]\n#[allow(dead_code)]\nfn fallible1(a: f32, b: f32) -> Result<f32, EvalError> {\n    Ok(a + b)\n}\n"
 ---
 #[derive(
     proptest_derive::Arbitrary,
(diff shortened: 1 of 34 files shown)


src/expr/src/scalar/func.rs
```diff
@@ -2723,6 +2742,7 @@ where
     sqlname = "extractts",
     propagates_nulls = true
 )]
+#[allow(dead_code)]
```
def-, line 2745: Why do we have all these dead_code markers?

src/workspace-hack/Cargo.toml
```diff
@@ -85,6 +85,7 @@ lru = { version = "0.12.5" }
 memchr = { version = "2.7.5" }
 mime_guess = { version = "2.0.5" }
 minimal-lexical = { version = "0.2.1", default-features = false, features = ["std"] }
+moka = { version = "0.12.10", features = ["future", "sync"] }
```
def-, line 88: Why do we depend on moka again now? I think we removed it a while ago for [reasons](https://github.com/MaterializeInc/database-issues/issues/5879). If it's pulled in by hashbrown's new features-usage, maybe we should reconsider if we need those features.

def- (commented):
Two comments, otherwise lgtm

src/expr/src/scalar/func.rs
antiguru: It's a work-in-progress to switch expression evaluation to these functions, but I haven't gotten around to finishing it yet.

src/workspace-hack/Cargo.toml
antiguru: `domain` 0.11 introduces this dependency, and we can't turn it off :( Not sure what to do here, 0.9 produces a warning.

antiguru:
Thanks for the review!

src/workspace-hack/Cargo.toml
def-: let's hope this won't come back to bite us, but at least we know what to look out for.

teskje: Fwiw, we were already depending on moka through the LD SDK crate. Hasn't been proven to be the source of memory corruption yet (though we suspected it during the crossbeam-channel saga).

PR #33276, opened 2025-08-08
Rewrite `FlatMapElimination`

ggevay:
This is on top of https://github.com/MaterializeInc/materialize/pull/33291.

Rewrites `FlatMapElimination` to simplify and generalize it. It used to check for several specific table functions. @petrosagg noticed that it could instead just execute any table function whose arguments are all literals, and then check whether it has at most 1 rows, and if yes, then transform away the table function. The only table function where we still need a special check is `Wrap`: we can eliminate it in some cases even when not all arguments are literals. (I've removed the `WITH ORDINALITY` handling for `Wrap`, because these never actually appear together, because `Wrap` is a thing that users don't call, but we insert it ourselves.)

The new code is simpler, and also fires for more table functions, e.g., for appropriate `generate_series` calls.

### Checklist

### DIFF
--- a/src/transform/src/canonicalization/flat_map_elimination.rs
+++ b/src/transform/src/canonicalization/flat_map_elimination.rs
@@ -7,16 +7,19 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-//! Turns `FlatMap` into `Map` if only one row is produced by flatmap.
-//!
+//! For a `FlatMap` whose args are all constants, turns it into `Map` if only 1 row is produced by
+//! the table function, or turns it into an empty constant if 0 rows are produced by the table
+//! function. Additionally, a `Wrap` whose width is larger than its number of arguments can be
+//! removed.
 
+use itertools::Itertools;
 use mz_expr::visit::Visit;
 use mz_expr::{MirRelationExpr, MirScalarExpr, TableFunc};
-use mz_repr::{Datum, Diff, ScalarType};
+use mz_repr::{Diff, Row, RowArena};
 
 use crate::TransformCtx;
 
-/// Turns `FlatMap` into `Map` if only one row is produced by flatmap.
+/// See comment at the top of the file.
 #[derive(Debug)]
 pub struct FlatMapElimination;
 
@@ -42,70 +45,49 @@ impl crate::Transform for FlatMapElimination {
 }
 
 impl FlatMapElimination {
-    /// Turns `FlatMap` into `Map` if only one row is produced by flatmap.
+    /// See comment at the top of the file.
     pub fn action(relation: &mut MirRelationExpr) {
         if let MirRelationExpr::FlatMap { func, exprs, input } = relation {
-            let (func, with_ordinality) = if let TableFunc::WithOrdinality(with_ordinality) = func {
-                // get to the actual function, but remember that we have a WITH ORDINALITY clause.
-                (&*with_ordinality.inner, true)
-            } else {
-                (&*func, false)
-            };
-
-            if let TableFunc::GuardSubquerySize { .. } = func {
-                // (`with_ordinality` doesn't matter because this function never emits rows)
-                if let Some(1) = exprs[0].as_literal_int64() {
-                    relation.take_safely(None);
-                }
-            } else if let TableFunc::Wrap { width, .. } = func {
+            // Treat Wrap specially.
+            if let TableFunc::Wrap { width, .. } = func {
                 if *width >= exprs.len() {
                     *relation = input.take_dangerous().map(std::mem::take(exprs));
-                    if with_ordinality {
-                        *relation = relation.take_dangerous().map_one(MirScalarExpr::literal(
-                            Ok(Datum::Int64(1)),
-                            ScalarType::Int64,
-                        ));
-                    }
+                    return;
                 }
-            } else if is_supported_unnest(func) {
-                let func = func.clone();
-                let exprs = exprs.clone();
-                use mz_expr::MirScalarExpr;
-                use mz_repr::RowArena;
-                if let MirScalarExpr::Literal(Ok(row), ..) = &exprs[0] {
-                    let temp_storage = RowArena::default();
-                    if let Ok(mut iter) = func.eval(&[row.iter().next().unwrap()], &temp_storage) {
-                        match (iter.next(), iter.next()) {
-                            (None, _) => {
-                                // If there are no elements in the literal argument, no output.
-                                relation.take_safely(None);
-                            }
-                            (Some((row, Diff::ONE)), None) => {
-                                assert_eq!(func.output_type().column_types.len(), 1);
-                                *relation =
-                                    input.take_dangerous().map(vec![MirScalarExpr::Literal(
-                                        Ok(row),
-                                        func.output_type().column_types[0].clone(),
-                                    )]);
-                                if with_ordinality {
-                                    *relation =
-                                        relation.take_dangerous().map_one(MirScalarExpr::literal(
-                                            Ok(Datum::Int64(1)),
-                                            ScalarType::Int64,
-                                        ));
-                                }
-                            }
-                            _ => {}
-                        }
-                    };
+            }
+            // For all other table functions, check for all arguments being literals.
+            let mut args = vec![];
+            for e in exprs {
+                match e.as_literal() {
+                    Some(Ok(datum)) => args.push(datum),
+                    // Give up if any arg is not a literal, or if it's a literal error.
+                    _ => return,
                 }
+            }
+            let temp_storage = RowArena::new();
+            let (first, second) = match func.eval(&args, &temp_storage) {
+                Ok(mut r) => (r.next(), r.next()),
+                // don't play with errors
+                Err(_) => return,
+            };
+            match (first, second) {
+                // The table function evaluated to an empty collection.
+                (None, None) => {
+                    relation.take_safely(None);
+                }
+                // The table function evaluated to a collection with exactly 1 row.
+                (Some((first_row, Diff::ONE)), None) => {
+                    let types = func.output_type().column_types;
+                    let map_exprs = first_row
+                        .into_iter()
+                        .zip_eq(types)
+                        .map(|(d, typ)| MirScalarExpr::Literal(Ok(Row::pack_slice(&[d])), typ))
+                        .collect();
+                    *relation = input.take_dangerous().map(map_exprs);
+                }
+                // The table function evaluated to a collection with more than 1 row; nothing to do.
+                _ => {}
             }
         }
     }
-}
-
-/// Returns `true` for `unnest_~` variants supported by [`FlatMapElimination`].
-fn is_supported_unnest(func: &TableFunc) -> bool {
-    use TableFunc::*;
-    matches!(func, UnnestArray { .. } | UnnestList { .. })
 }
--- a/src/transform/src/canonicalization/flat_map_elimination.rs
+++ b/src/transform/src/canonicalization/flat_map_elimination.rs
@@ -7,10 +7,13 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-//! For a `FlatMap` whose args are all constants, turns it into `Map` if only 1 row is produced by
-//! the table function, or turns it into an empty constant if 0 rows are produced by the table
-//! function. Additionally, a `Wrap` whose width is larger than its number of arguments can be
-//! removed.
+//! For a `FlatMap` where the table function's arguments are all constants, turns it into `Map` if
+//! only 1 row is produced by the table function, or turns it into an empty constant collection if 0
+//! rows are produced by the table function.
+//!
+//! It does an additional optimization on the `Wrap` table function: when `Wrap`'s width is larger
+//! than its number of arguments, it removes the `FlatMap Wrap ...`, because such `Wrap`s would have
+//! no effect.
 
 use itertools::Itertools;
 use mz_expr::visit::Visit;
@@ -19,7 +22,7 @@ use mz_repr::{Diff, Row, RowArena};
 
 use crate::TransformCtx;
 
-/// See comment at the top of the file.
+/// Attempts to eliminate FlatMaps that are sure to have 0 or 1 results on each input row.
 #[derive(Debug)]
 pub struct FlatMapElimination;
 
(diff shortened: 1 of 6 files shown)


src/transform/src/canonicalization/flatmap_to_map.rs
```diff
@@ -42,70 +43,59 @@ impl crate::Transform for FlatMapElimination {
...
+            let args: Result<Vec<Datum>, ()> = exprs
+                .iter()
+                .map(|e| {
+                    e.as_literal()
+                        // give up if one of them is not a literal
+                        .ok_or(())?
+                        // give up also if one of them is a literal error
+                        .map_err(|_eval_err| ())
+                })
+                .collect();
```
petrosagg: nit: the short-circuiting behavior of collecting a Result is a big magical, maybe the imperative version reads clearer

```rust
let args = vec![];
for expr in exprs {
    match expr.as_literal() {
        Some(Ok(datum)) => args.push(datum),
        _ => return;
    }
}
```

src/transform/src/canonicalization/flatmap_to_map.rs
```diff
@@ -42,70 +43,59 @@ impl crate::Transform for FlatMapElimination {
...
-                                if with_ordinality {
-                                    *relation =
-                                        relation.take_dangerous().map_one(MirScalarExpr::literal(
-                                            Ok(Datum::Int64(1)),
-                                            ScalarType::Int64,
-                                        ));
+                            let second = r.next();
+                            if second.is_none() {
+                                drop(ev);
+                                let (first_row, first_diff) = first.expect("checked above");
```
petrosagg: You can use `match` statements to avoid a lot of the extra things needed here to make rust happy, like this expect, the drops, the check for the diff to be `Diff::ONE`.

```rust
            let mut args = vec![];
            for e in exprs {
                match e.as_literal() {
                    Some(Ok(datum)) => args.push(datum),
                    _ => return,
                }
            }
            let temp_storage = RowArena::new();
            let (first, second) = match func.eval(&args, &temp_storage) {
                Ok(mut r) => (r.next(), r.next()),
                // don't play with errors
                Err(_) => return,
            };
            match (first, second) {
                // The table function evaluated to an empty collection.
                (None, None) => {
                    relation.take_safely(None);
                }
                // The table function evaluated to a collection with exactly 1 row.
                (Some((first_row, Diff::ONE)), None) => {
                    let types = func.output_type().column_types;
                    let map_exprs = first_row
                        .into_iter()
                        .zip_eq(types)
                        .map(|(d, typ)| MirScalarExpr::Literal(Ok(Row::pack_slice(&[d])), typ))
                        .collect();
                    *relation = input.take_dangerous().map(map_exprs);
                }
                // The table function evaluated to a collection with more than 1 row;
                // nothing to do.
                _ => {}
            }
```

test/sqllogictest/table_func.slt
```diff
@@ -1710,3 +1710,33 @@ Source materialize.public.x
...
 EOF
+
+query T multiline
+EXPLAIN OPTIMIZED PLAN WITH (NO FAST PATH) FOR
+SELECT *
+FROM x, generate_series(1, x.a)
+WHERE x.a = 1
+----
+Explained Query:
+  Project (#0, #1, #0)
```
petrosagg, line 1721: Which transform is responsible for this optimization? The transform normally generates a `Map(1)` node here, right?

Do we have any testing infrastructure that can test a specific transform in isolation?

src/transform/src/canonicalization/flatmap_to_map.rs
ggevay: Hmm, ok! I thought I'm the only one who finds this hard! But seeing that a real Rust expert also doesn't like it, I'll definitely switch to the imperative version.

src/transform/src/canonicalization/flatmap_to_map.rs
ggevay: Ahh, really nice, it's even shorter now! Thank you very much!

test/sqllogictest/table_func.slt
ggevay: > Which transform is responsible for this optimization? The transform normally generates a Map(1) node here, right?

Yes. It's `MapFilterProject::optimize` being called from `CanonicalizeMfp`. Specifically, [this part](https://github.com/MaterializeInc/materialize/blob/844ad57e4b5518f99b68c3422dc3c32398361e77/src/expr/src/linear.rs#L1017-L1052).

> Do we have any testing infrastructure that can test a specific transform in isolation?

Yes, the `.spec` tests. I'll add some tests in `flatmap_to_map.spec`, good idea!

ggevay: Done

src/transform/src/canonicalization/flat_map_elimination.rs
```diff
@@ -7,16 +7,19 @@
...
+use itertools::Itertools;
 use mz_expr::visit::Visit;
 use mz_expr::{MirRelationExpr, MirScalarExpr, TableFunc};
-use mz_repr::{Datum, Diff, ScalarType};
+use mz_repr::{Diff, Row, RowArena};
 
 use crate::TransformCtx;
 
-/// Turns `FlatMap` into `Map` if only one row is produced by flatmap.
+/// See comment at the top of the file.
```
frankmcsherry: Fwiw, I think this is not super helpful for doc comments. The e.g. hover text that you'd normally get being "see the explanation somewhere else" won't be useful. The old comments still seem accurate, unless you are worried about the zero rows case? But imo worth getting good text here (and at the top of the file).

src/transform/src/canonicalization/flat_map_elimination.rs
```diff
@@ -7,16 +7,19 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-//! Turns `FlatMap` into `Map` if only one row is produced by flatmap.
-//!
+//! For a `FlatMap` whose args are all constants, turns it into `Map` if only 1 row is produced by
+//! the table function, or turns it into an empty constant if 0 rows are produced by the table
```
frankmcsherry: "empty constant" probably should be "empty collection"?

src/transform/src/canonicalization/flat_map_elimination.rs
```diff
@@ -7,16 +7,19 @@
...
+//! function. Additionally, a `Wrap` whose width is larger than its number of arguments can be
```
frankmcsherry: I have no idea what `Wrap` does, but it does now feel like the odd function out. The explanation is a bit underdeveloped. If I were writing it, I think I'd make this a new paragraph, as it seems unrelated to the previous thought (it is not an instance of constant arguments, but rather an unrelated thing we can do with flat maps).

Also maybe just a nit, but the previous sentences are active voiced (this transform will do a thing) vs "can be removed" which .. does this transform do that or not? (( yes, but let's say that it does. ))

src/transform/src/canonicalization/flat_map_elimination.rs
```diff
@@ -42,70 +45,49 @@ impl crate::Transform for FlatMapElimination {
...
-                (&*func, false)
-            };
-
-            if let TableFunc::GuardSubquerySize { .. } = func {
-                // (`with_ordinality` doesn't matter because this function never emits rows)
-                if let Some(1) = exprs[0].as_literal_int64() {
-                    relation.take_safely(None);
-                }
-            } else if let TableFunc::Wrap { width, .. } = func {
+            // Treat Wrap specially.
```
frankmcsherry: I think this may be a good moment to record what is about to happen and why. The comment "treat wrap specially" is .. not the question a reader has at this moment, I claim, but rather "why are we treating `Wrap` specially?". Is it important that we do this and return, to ensure we do not call the rest of the function? Is it important that we don't look for `WithOrdinality` here, because it is not `Wrap` which we treat specially? Many questions that the comment doesn't address, but imo we should try and provide the context to answer in the future.

src/transform/src/canonicalization/flat_map_elimination.rs
```diff
@@ -42,70 +45,49 @@ impl crate::Transform for FlatMapElimination {
...
+            }
+            let temp_storage = RowArena::new();
+            let (first, second) = match func.eval(&args, &temp_storage) {
+                Ok(mut r) => (r.next(), r.next()),
+                // don't play with errors
+                Err(_) => return,
+            };
+            match (first, second) {
+                // The table function evaluated to an empty collection.
+                (None, None) => {
```
frankmcsherry: Small nit, but could this be `(None, _)`? Idk that we actively fuse our iterators, and I think the only thing that matters is that the first call to `next()` returns a `None`. Lightly weird to call `next()` a second time after getting a `None` back, which I think we do to fit into `match`, though .. I can't tell if that is helpful or not. Not a big deal either way; just something that struck me while trying to read this.

src/transform/src/canonicalization/flat_map_elimination.rs
```diff
@@ -42,70 +45,49 @@ impl crate::Transform for FlatMapElimination {
...
                 }
+                // The table function evaluated to a collection with exactly 1 row.
+                (Some((first_row, Diff::ONE)), None) => {
+                    let types = func.output_type().column_types;
+                    let map_exprs = first_row
+                        .into_iter()
+                        .zip_eq(types)
+                        .map(|(d, typ)| MirScalarExpr::Literal(Ok(Row::pack_slice(&[d])), typ))
+                        .collect();
+                    *relation = input.take_dangerous().map(map_exprs);
```
frankmcsherry: Maybe premature defensive coding, but we might want a `return;` after setting the relation? The method should not continue once we swap the relation out, and .. tbh probably Rust will defend against that, but just like with the `Wrap` case we don't want to let execution continue after this point.

src/transform/src/canonicalization/flat_map_elimination.rs
```diff
@@ -42,70 +45,49 @@ impl crate::Transform for FlatMapElimination {
...
+                }
+            }
+            // For all other table functions, check for all arguments being literals.
+            let mut args = vec![];
+            for e in exprs {
+                match e.as_literal() {
+                    Some(Ok(datum)) => args.push(datum),
+                    // Give up if any arg is not a literal, or if it's a literal error.
+                    _ => return,
                 }
```
frankmcsherry: I think I saw a comment that the `collect()` syntax is unclear, and don't want to fight that but I do actually like it! :D At least, if it ends up being as simple as
```rust
// If all arguments are literals, we may be able to remove the `FlatMap`.
if let Some(args) = exprs.iter().map(|e| e.as_literal()).collect() {
    ...
}
```
But don't let me make things more complicated (and .. maybe `rustfmt` just makes it impossible to read; that's another reason not to do it).

src/transform/src/canonicalization/flat_map_elimination.rs
```diff
@@ -7,16 +7,19 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-//! Turns `FlatMap` into `Map` if only one row is produced by flatmap.
-//!
+//! For a `FlatMap` whose args are all constants, turns it into `Map` if only 1 row is produced by
```
frankmcsherry: Nit but I think args should either be `FlatMap::args` or be spelled out. There is an `args` member of the `FlatMap` stage, and that is what we  are talking about, but you need to know that and not worry about the fact that the main argument of the flatmap, the relation input, doesn't have to be a constant.

frankmcsherry (approved):
Generally good! I left some comments that are a mix of nits and good ideas about documentation, but .. I've left it a mystery which are which (tbh, not sure and I'd personally do them all, but maybe some are too pedantic).

src/transform/src/canonicalization/flat_map_elimination.rs
petrosagg: `as_literal()` returns a result and the original formulation was dealing with that like so:

```rust
let args: Result<Vec<Datum>, ()> = exprs
    .iter()
    .map(|e| {
        e.as_literal()
            // give up if one of them is not a literal
            .ok_or(())?
            // give up also if one of them is a literal error
            .map_err(|_eval_err| ())
    })
    .collect();
match args {
    Err(()) => return,
    Ok(args) => {
```

But your suggestion also works if modified like so:

```rust
// If all arguments are literals, we may be able to remove the `FlatMap`.
if let Some(args) = exprs.iter().map(|e| e.as_literal().ok()).collect() {
    ...
}
```

I don't have a strong opinion between this suggestion and the imperative version

src/transform/src/canonicalization/flat_map_elimination.rs
petrosagg: It could be `(None, _)`, yes.

>  Lightly weird to call next() a second time after getting a None back, which I think we do to fit into match, though .. I can't tell if that is helpful or not.

Slightly helpful. The table func iterator holds a lifetime that originates from the current MIR node so by calling `r.next()` inside the previous match statement makes sure `r` is dropped by the time we reach this point and we want to replace the current MIR node. It's possible to hold on to it but then you have to insert manual drop statements to make Rust happy

src/transform/src/canonicalization/flat_map_elimination.rs
ggevay: > if it ends up being as simple as
> ```
> // If all arguments are literals, we may be able to remove the `FlatMap`.
> if let Some(args) = exprs.iter().map(|e| e.as_literal()).collect() {
>     ...
> }
> ```

Unfortunately, it's not so simple, because `as_literal`'s return type has two layers of failure: `Option<Result<Datum<'_>, &EvalError>>`.

> But your suggestion also works if modified like so:
> ```
> // If all arguments are literals, we may be able to remove the `FlatMap`.
> if let Some(args) = exprs.iter().map(|e| e.as_literal().ok()).collect() {
>    ...
> }
> ```

This doesn't work either, because `.ok()` would work on a `Result`, but the `Result` is inside an `Option` here.

The following works:
```
if let Some(args) = exprs
    .iter()
    .map(|e| e.as_literal().map(|eval_err| eval_err.ok()))
    .flatten()
    .collect::<Option<Vec<_>>>()
{
```
which
- uses `map` after the `as_literal` to get to the `Result`;
- uses `flatten()` to get rid of the double `Option`;
- needed to specify `collect`'s generic arg, because otherwise the compiler wanted to collect into something other than a Vec, and failed.

But this doesn't look particularly readable, so I've factored out the flattening of the two failure modes of `as_literal` into a new `as_literal_non_error` function:
```
    pub fn as_literal_non_error(&self) -> Option<Datum<'_>> {
        self.as_literal().map(|eval_err| eval_err.ok()).flatten()
    }
```
Using this, I wrote:
```
            if let Some(args) = exprs
                .iter()
                .map(|e| e.as_literal_non_error())
                .collect::<Option<Vec<_>>>()
            {
```
which is close to Frank's desired version from above. (But it still needs the turbofish.)

src/transform/src/canonicalization/flat_map_elimination.rs
ggevay: I've modified it to "For a `FlatMap` where the table function's arguments are all constants", because unfortunately the relevant member of the `FlatMap` stage is named `exprs`, which is also not very clear that it's the args to the table function.

src/transform/src/canonicalization/flat_map_elimination.rs
ggevay: Added a `return`.

src/transform/src/canonicalization/flat_map_elimination.rs
ggevay: Yes, my original code had manual `drop`s, but this version looks cleaner.

I've modified it to have `(None, _)`.
(I could also add a `fuse` call if we are worried about some iterator panicking on the second call. But this is hopefully unnecessary.)

src/transform/src/canonicalization/flat_map_elimination.rs
ggevay: Yes, improved the comments.

The `return` was there to satisfy the borrow-checker. But I've now realized that the rest of the function can sometimes do a useful thing even when the special Wrap code couldn't do a thing: when the `width` condition doesn't hold, but all args to Wrap are literals. So, now I've pulled out the `Wrap` handling from the big
```
if let MirRelationExpr::FlatMap { func, exprs, input } = relation {
```
to avoid the borrow-checker issue, and removed the `return`. (And added a test for this case.)

src/transform/src/canonicalization/flat_map_elimination.rs
ggevay: Right, thanks!

src/transform/src/canonicalization/flat_map_elimination.rs
ggevay: Changed to "empty constant collection"

src/transform/src/canonicalization/flat_map_elimination.rs
ggevay: done

Tbh, the redundancy that this introduces with the comments at the top of the file always seemed weird to me, see many other transforms. But if this is The Way, I'll just keep doing it.

ggevay:
Thanks for the reviews!

src/transform/src/canonicalization/flat_map_elimination.rs
ggevay: Ah, clippy is complaining of an "unneeded `return` statement", so I've removed it.

src/transform/src/canonicalization/flat_map_elimination.rs
frankmcsherry: I'm sorry I didn't mean to re-open a can of worms. It looks good; tyvm!

src/transform/src/canonicalization/flat_map_elimination.rs
frankmcsherry: Yeah, clippy alternates between not understanding software engineering and having helpful thoughts about it.

src/transform/src/canonicalization/flat_map_elimination.rs
frankmcsherry: I agree, and .. if we end up with some clarity here (e.g. mod docs for X, struct docs for Y, method docs for Z) I'll be delighted too. It's weirder because we have a structure that ostensibly exists only to call a function, and what we really have is a `Box<dyn Transform>` but we use a struct, in its own mod, etc to do this.

PR #33281, opened 2025-08-11
ci: Introduce CI_UNPARALLELIZE

def-:
This helps with reproducing some bugs for example. In addition I sometimes like to run CI with very low priority and without parallelization on a PR when I'm just testing something unimportant and very slow, like a regular coverage or sanitizer run.
### Checklist

### DIFF
--- a/ci/mkpipeline.py
+++ b/ci/mkpipeline.py
@@ -200,6 +200,7 @@ so it is executed.""",
     permit_rerunning_successful_steps(pipeline)
     set_retry_on_agent_lost(pipeline)
     set_default_agents_queue(pipeline)
+    unparallelize(pipeline)
     set_parallelism_name(pipeline)
     check_depends_on(pipeline, args.pipeline)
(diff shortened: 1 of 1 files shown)


ci/mkpipeline.py
```diff
@@ -1057,5 +1058,18 @@ def remove_mz_specific_keys(pipeline: Any) -> None:
             )
 
 
+def unparallelize(pipeline: Any) -> None:
+    """Removes parallelism in the test, which will run longer, but exposes some interesting parallelism in some tests."""
```
ggevay, line 1062: What does "exposes" mean here? That some tests will still run in parallel due to some other kind of parallelization that is not removed here?

def-: Some tests will create lots of objects, or run checks in parallel. By having multiple agents we have fewer objects and fewer parallel checks in each individual test, which reduces the chance of running into some race conditions.

PR #33283, opened 2025-08-11
upsert: fix correctness bug due to typo

petrosagg:
### Motivation

This PR is fixing a subtle bug that can result in a source missing deletions for keys. The fix was due to a typo on a single line and it is included as a separate PR. While working on this I realized that we use `Option<FromTime>` as the type we instantiate the order parameter of upsert state backends and that we additionally record such order field for non-provisional values. Having these non-provisional orders isn't necessary and as seen from the diff they were only ever instantiated with the `Default::default()` values. Since I was in the area I cleaned up the `Value` enum such that only the provisional values carry an order.

### Checklist


src/storage/src/upsert/types.rs
```diff
@@ -721,7 +668,6 @@ impl<T: Eq, O: Default> StateValue<T, O> {
                         );
                         *self = Self::Value(Value::FinalizedValue(
                             bincode_opts.deserialize(value).unwrap(),
-                            Default::default(),
```
petrosagg, line 724: This is where the order was set for the finalized values, and it's always `Default::default()`. This makes sense, the finalized value all come from feeding back updates from persist, at which point order information is lost. The order is only relevant to sort *intra*timestamp updates. Once a timestamp is finalized the order is irrelevant.

src/storage/src/upsert_continual_feedback.rs
```diff
@@ -1052,4 +1050,197 @@ mod test {
...
+            // ..and send the next *out of order* upsert command that deletes the key. Here msg4
+            // happens at offset 4 and the operator should rememeber that.
+            input_handle.send(msg4);
+            input_handle.flush();
+            // Run the worker for enough steps to process these events. We can't guide the
+            // execution with the probe here since the frontier does not advance, only provisional
+            // updates are produced.
+            for _ in 0..5 {
+                // Commenting out this line makes the test pass because the conflicting commands
+                // are observed as part of a signle operator step and are correctly deduped.
```
petrosagg: TODO: remove this comment after CI passes

src/storage/src/upsert.rs
```diff
@@ -53,15 +53,25 @@ use types::{
 
 #[cfg(test)]
 pub mod memory;
-mod rocksdb;
+pub mod rocksdb;
```
martykulma: nit: could be `pub(crate)`

src/storage/src/upsert_continual_feedback.rs
```diff
@@ -1052,4 +1050,195 @@ mod test {
...
+                                    )
+                                },
+                            ));
+                            let rocksdb_cleanup_tries = 5;
+                            let tuning = RocksDBConfig::new(Default::default(), None);
+                            crate::upsert::rocksdb::RocksDB::new(
+                                mz_rocksdb::RocksDBInstance::new(
+                                    rocksdb_dir.path(),
+                                    mz_rocksdb::InstanceOptions::new(
+                                        Env::new().unwrap(),
```
martykulma: nit: could use `Env::mem_env()` as this test is tiny

src/storage/src/upsert/types.rs
```diff
@@ -458,93 +444,54 @@ impl<T: Eq, O> StateValue<T, O> {
...
+            Self::Value(Value::Tombstone) => None,
             Self::Value(Value::ProvisionalValue {
                 finalized_value: _,
                 provisional_value: (_, provisional_ts, provisional_order),
             }) if provisional_ts == ts => Some(provisional_order),
-            Self::Value(Value::ProvisionalValue {
-                finalized_value,
-                provisional_value: _,
-            }) => finalized_value.as_ref().map(|v| &v.1),
+            Self::Value(Value::ProvisionalValue { .. }) => None,
```
martykulma, line 450: Comment is no longer valid

PR #33287, opened 2025-08-11
cluster,orchestrator-k8s: report swap usage

teskje:
This PR adds swap usage reporting to the cluster and kubernetes orchestrator, that works the same way as the existing disk usage reporting. In the orchestrator, both disk and swap usage are simply summed up. We expect replicas to either use an explicit disk or swap, so no information is lost this way.

As a result of this change, swap usage is now reported through `mz_cluster_replica_metrics`. Note that the swap limit is not yet written to `mz_cluster_replica_sizes`, so percentage based utilization is still broken for swap replicas.

Needs a change in the console to not break the UI in environments that run swap replicas: https://github.com/MaterializeInc/console/pull/3866

### Motivation

Part of https://github.com/MaterializeInc/database-issues/issues/9366

### Checklist

### DIFF
--- a/src/orchestrator-kubernetes/src/lib.rs
+++ b/src/orchestrator-kubernetes/src/lib.rs
@@ -215,8 +215,6 @@ impl Orchestrator for KubernetesOrchestrator {
 #[derive(Clone, Copy)]
 struct ServiceInfo {
     scale: u16,
-    disk: bool,
-    disk_limit: Option<DiskLimit>,
 }
 
 struct NamespacedKubernetesOrchestrator {
@@ -1228,14 +1226,10 @@ impl NamespacedOrchestrator for NamespacedKubernetesOrchestrator {
             },
         });
 
-        self.service_infos.lock().expect("poisoned lock").insert(
-            id.to_string(),
-            ServiceInfo {
-                scale,
-                disk,
-                disk_limit,
-            },
-        );
+        self.service_infos
+            .lock()
+            .expect("poisoned lock")
+            .insert(id.to_string(), ServiceInfo { scale });
 
         Ok(Box::new(KubernetesService { hosts, ports }))
     }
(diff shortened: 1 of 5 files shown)


src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -1516,21 +1495,6 @@ impl OrchestratorWorker {
...
-            // reported metric. In that case, we report the minimum of the usage
-            // and the limit, which means we can report 100% usage temporarily
-            // if a replica is sized down.
-            let disk_usage = match (disk_usage, disk_limit) {
-                (Some(disk_usage), Some(DiskLimit(disk_limit))) => {
-                    Some(std::cmp::min(disk_usage, disk_limit.0))
-                }
-                _ => None,
-            };
-
```
teskje, line 1533: I removed this because (a) with swap we don't necessarily have a disk limit and (b) I don't understand why it exists. Bit of a Chesterton's fence situation, so lmk if anyone has more context!

antiguru: Maybe a leftover from 54aa75edf4dc802e510b28bdb117767c9d06add5?

teskje: You mean we could have removed it in that change already? I think that's true, I'm just not sure why we needed it before. Sounds like it was sometimes possible to receive a disk usage value from a previous incarnation of the replica that doesn't match the current disk configuration. If that's the case then that can indeed not happen anymore now.

src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -215,8 +215,6 @@ impl Orchestrator for KubernetesOrchestrator {
 #[derive(Clone, Copy)]
 struct ServiceInfo {
     scale: u16,
-    disk: bool,
-    disk_limit: Option<DiskLimit>,
 }
```
antiguru, line 218: (Just pondering) We could remove the `ServiceInfo` struct, but would still need a place to store private information. Maybe the orchestrator could return an opaque handle to store private information, and functions that currently take an ID would use the handle instead?

src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -1584,16 +1549,21 @@ impl OrchestratorWorker {
...
+            let Usage {
+                disk_bytes,
+                swap_bytes,
+            } = resp.json().await?;
+
+            let bytes = match (disk_bytes, swap_bytes) {
+                (Some(disk), Some(swap)) => Some(disk + swap),
+                (Some(bytes), None) | (None, Some(bytes)) => Some(bytes),
+                (None, None) => None,
+            };
```
antiguru:
```suggestion
            let bytes = disk_bytes.into_iter().chain(swap_bytes).fold(None, |r, i| Some(r.unwrap_or(0) + i))
```
Maybe?

antiguru (approved):
Looks good, thanks.

src/orchestrator-kubernetes/src/lib.rs
teskje: We could do that, but I'm not sure it improves things. We'll remove some state from the orchestrator, but we'll have to create that state in the controller (or at a higher level again). Currently the controller constructs the ID from `(cluster_id, replica_id, generation)`. It would instead need to keep a map relating these IDs to service handles, which would look a lot like the `service_infos` map in the orchestrator now.

src/orchestrator-kubernetes/src/lib.rs
teskje: I find the match clearer! It is a bit verbose though, I'll replace it with

```
if let (Some(disk), Some(swap)) = (disk_bytes, swap_bytes) {
    Some(disk + swap)
} else {
    disk_bytes.or(swap_bytes)
}
```

It would sure be nice if there was an `Option` combinator with the same semantics...

alex-hunt-materialize (approved):

teskje:
TFTRs!

PR #33291, opened 2025-08-11
Refactor `TableFunc::WithOrdinality`

ggevay:
to anticipate future TableFuncs that might not make sense with WITH ORDINALITY due to negative diffs.

This addresses @frankmcsherry's comments on the WITH ORDINALITY PR (https://github.com/MaterializeInc/materialize/pull/33163) + Slack DM discussion with him.

### Motivation

### Checklist

### DIFF
--- a/src/expr/src/relation/func.rs
+++ b/src/expr/src/relation/func.rs
@@ -3716,9 +3716,8 @@ fn mz_acl_explode<'a>(
     Ok(res.into_iter())
 }
 
-/// Don't add table functions that emit negative diffs! These are undefined with `WITH ORDINALITY`.
-/// (`Repeat` _can_ emit negative diffs, on which `WITH ORDINALITY` panics, but this is in
-/// mz_unsafe, so not callable by users.)
+/// When adding a new `TableFunc` variant, please consider adding it to
+/// `TableFunc::with_ordinality`!
 #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize, Hash, MzReflect)]
 pub enum TableFunc {
     AclExplode,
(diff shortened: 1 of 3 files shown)


src/expr/src/relation/func.rs
```diff
@@ -3785,18 +3784,59 @@ pub enum TableFunc {
...
+            | TableFunc::UnnestList { .. }
+            | TableFunc::UnnestMap { .. }
+            | TableFunc::Wrap { .. }
+            | TableFunc::GenerateSubscriptsArray
+            | TableFunc::TabletizedScalar { .. }
+            | TableFunc::RegexpMatches => Some(TableFunc::WithOrdinality(WithOrdinality {
+                inner: Box::new(inner),
+            })),
+            // IMPORTANT: Before adding a new table function here, consider negative diffs:
+            // `WithOrdinality::eval` currently asserts that the inner table function never emits
```
frankmcsherry: Light nit, but "currently asserts that" could be strengthened to "will panic if". I think we know that the former implies the latter, but .. might as well just say it.

src/expr/src/relation/func.rs
```diff
@@ -4279,51 +4326,56 @@ impl fmt::Display for TableFunc {
...
-            })
-        });
-    Ok(Box::new(it))
+impl WithOrdinality {
+    fn eval<'a>(
+        &'a self,
+        datums: &'a [Datum<'a>],
+        temp_storage: &'a RowArena,
+    ) -> Result<Box<dyn Iterator<Item = (Row, Diff)> + 'a>, EvalError> {
+        // zip 1, 2, 3, 4, ... to the output of the table function. We need to blow up non-1 diffs,
```
frankmcsherry: Nit but "blow up" could be clearer. How about instead "We need to expand rows with non-1 diffs into the corresponding number of rows with unit diffs, because the ordinality column ..."

src/expr/src/relation/func.rs
```diff
@@ -4279,51 +4326,56 @@ impl fmt::Display for TableFunc {
...
-                    new_row.clone_from(&row);
-                    new_row
-                };
-                RowPacker::for_existing_row(&mut row).push(Datum::Int64(ordinal));
-                Some((row, Diff::ONE))
-            })
-        });
-    Ok(Box::new(it))
+impl WithOrdinality {
+    fn eval<'a>(
```
frankmcsherry, line 4347: The doccomment for this function is a great place to explain what it will and will not do. We should do that here, both to be useful, but also to lock in a clear understanding of what the method does without needing to read line by line to see how it does it. For example, this is where one would normally have a `# Panics` section to explain that if a negative diff is encountered the function will panic.

src/expr/src/relation/func.rs
ggevay: done

src/expr/src/relation/func.rs
ggevay: done

src/expr/src/relation/func.rs
ggevay: done

PR #33297, opened 2025-08-12
tests: Reenable 2 disabled tests since 7611 is fixed

def-:
See https://github.com/MaterializeInc/database-issues/issues/7611

Follow-up to https://github.com/MaterializeInc/materialize/pull/33277

### Checklist

### DIFF
--- a/test/pg-cdc-old-syntax/mzcompose.py
+++ b/test/pg-cdc-old-syntax/mzcompose.py
@@ -352,10 +352,6 @@ def workflow_default(c: Composition, parser: WorkflowArgumentParser) -> None:
         if name in ("default", "migration"):
             return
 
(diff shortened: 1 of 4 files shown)


def-:
@bkirwi Anything we can do to make these reliable? Does something similar to https://github.com/MaterializeInc/materialize/pull/33277 work for them?

bkirwi:
I'm not sure. I think these were distinct issues... that test asserts on the Kafka output, while these assert on statuses, and I'm not sure error statuses are "eventually consistent" / convergent in the same way.

I see these two tests assert that the error must be about the publication being dropped. Do you know what the error is when the test flakes? A lot of the old logs are out of buildkite already...

def-:
I'd suggest we merge reenabling them and catch them flaking again, then we can look into it.

PR #33299, opened 2025-08-12
upsert: refactor `UpsertValue` struct and box error to reduce memory usage

petrosagg:
#### Enum to struct

The previous definition of `UpsertValue` was emulating a struct of two fields, one of which is optional, with an enum. It seems clearer to express directly with a struct, so this is the first commit of this PR.

#### Struct size reduction
`DataflowError` instances are heap allocated since these values are rarer and we don't want to be paying the cost of storing them for each value.

The upsert operator was previosuly unpacking the general `DataflowError` into a local `UpsertError` variant that was not boxed, leading to large structs.

This commit re-boxes the error variant bringing the size of `UpsertValue` down significantly.

Fixes https://github.com/MaterializeInc/database-issues/issues/9567

### Checklist


def- (commented):
Awesome! Nightly triggered: https://buildkite.com/materialize/nightly/builds/12836
Edit: Please ignore the platform checks + cloudtest failures, they are being fixed in https://github.com/MaterializeInc/materialize/pull/33298
I am mostly concerned about this failure:
```
snapshot-source-statistics.td:129:1: non-matching rows: expected:
[["upsert", "true", "9", "9"]]
got:
[["upsert", "false", "9", "9"]]
Poor diff:
+ upsert false 9 9
- upsert true 9 9
```

src/storage/src/upsert/types.rs
```diff
@@ -477,12 +433,7 @@ impl<T: Eq, O> StateValue<T, O> {
...
     pub fn into_finalized_value(self) -> Option<UpsertValue> {
         match self {
-            Self::Value(Value::FinalizedValue(value)) => Some(value),
-            Self::Value(Value::Tombstone) => None,
-            Self::Value(Value::ProvisionalValue {
-                finalized_value,
-                provisional_value: _,
-            }) => finalized_value,
+            Self::Value(v) => v.finalized,
             _ => panic!("called `order` without calling `ensure_decoded`"),
```
martykulma: may want to include an update to this panic string while you're in here 😁

src/storage/src/upsert/types.rs
```diff
@@ -219,20 +219,28 @@ impl<T, O> std::fmt::Debug for StateValue<T, O> {
...
+
+/// A provisional value emitted for a timestamp. This struct contains enough information to
+#[derive(Clone, serde::Serialize, serde::Deserialize, Debug)]
+pub struct ProvisionalValue<T, O> {
+    /// The timestamp at which this provisional value occured at
+    pub timestamp: T,
+    // The order of this upsert command *within* the timestamp. Commands that happen at the same
+    // timestamp with lower order get ignored. Commands with higher order override this one. If
+    // there a case of equal order then the value itself is used as a tie breaker.
+    pub order: O,
```
martykulma: should these be doc comments (///)?

s/there/there is/ on 239

def-:
I think this will solve a bunch of flakes in CI, are we planning to merge it soon?

petrosagg:
I hit retry on the statistics test to see if it was a flake. These statistics are produced by another component of the ingestion pipeline so I'm inclined to merge this even if they fail again. We'll know in a few minutes

PR #33308, opened 2025-08-12
compute: always replace copy-to dataflows in reconciliation

teskje:
Compute cannot correctly reconcile copy-to dataflows for the same reason it cannot reconcile subscribes: The `copy_to_response_buffer` is shared among all running copy-tos and so we can't selectively drop/replace some and keep others. Thus we do the same thing we do for subscribes: Replace all copy-tos unconditionally during reconciliation.

### Motivation

Compute reconciliation is likely incorrect for copy-to dataflows.

I stumbled over this while investigating https://github.com/MaterializeInc/database-issues/issues/9566. I'm not sure that issue is caused by the reconciliation bug though. At least I couldn't explain how it would lead to duplicate responses.

### Checklist

### DIFF
--- a/src/compute/src/metrics.rs
+++ b/src/compute/src/metrics.rs
@@ -280,11 +280,21 @@ impl WorkerMetrics {
     ///
     /// Reconciliation is recorded as successful if the given properties all hold. Otherwise it is
     /// recorded as unsuccessful, with a reason based on the first property that does not hold.
+    ///
+    /// The properties are:
+    ///  * compatible: The old and new dataflow descriptions are compatible.
+    ///  * uncompacted: Collections currently installed for the dataflow exports have not been
+    ///                 allowed to compact beyond that new dataflow as-of.
+    ///  * subscribe_free: The dataflow does not export a subscribe sink.
+    ///  * copy_to_free: The dataflow does not export a copy-to sink.
+    ///  * dependencies_retained: All local inputs to the dataflow were retained by compute
+    ///                           reconciliation.
     pub fn record_dataflow_reconciliation(
         &self,
         compatible: bool,
         uncompacted: bool,
         subscribe_free: bool,
+        copy_to_free: bool,
         dependencies_retained: bool,
     ) {
         if !compatible {
@@ -301,6 +311,11 @@ impl WorkerMetrics {
             self.metrics
                 .reconciliation_replaced_dataflows_count_total
                 .with_label_values(&[&self.worker_label, "subscribe"])
+                .inc();
+        } else if !copy_to_free {
+            self.metrics
+                .reconciliation_replaced_dataflows_count_total
+                .with_label_values(&[&self.worker_label, "copy-to"])
                 .inc();
         } else if !dependencies_retained {
             self.metrics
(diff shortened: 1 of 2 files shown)


src/compute/src/metrics.rs
```diff
@@ -285,6 +285,7 @@ impl WorkerMetrics {
         compatible: bool,
         uncompacted: bool,
         subscribe_free: bool,
+        copy_to_free: bool,
```
antiguru, line 297: Consider adding documentation for the parameters, just reading this function it's not clear what they indicate.

src/compute/src/server.rs
```diff
@@ -675,18 +682,19 @@ impl<'w, A: Allocate + 'static> Worker<'w, A> {
...
                     collection.sink_token = None;
                 }
             }
 
-            // We must drop the subscribe response buffer as it is global across all subscribes.
-            // If it were broken out by `GlobalId` then we could drop only those of dataflows we drop.
+            // We must drop the response buffers as they are global across all subscribe/copy-tos.
+            // If they were broken out by `GlobalId` then we could drop only those of dataflows we drop.
             compute_state.subscribe_response_buffer = Rc::new(RefCell::new(Vec::new()));
+            compute_state.copy_to_response_buffer = Rc::new(RefCell::new(Vec::new()));
```
antiguru, line 698: Should we convert the response buffers to something like a `BTreeMap<GlobalId, Rc<RefCell<...>>>`? Not part of this PR though!

src/compute/src/server.rs
```diff
@@ -675,18 +682,19 @@ impl<'w, A: Allocate + 'static> Worker<'w, A> {
...
+                // subscribe/copy-tos dropped during reconciliation.
                 if !retained {
                     collection.sink_token = None;
                 }
             }
 
-            // We must drop the subscribe response buffer as it is global across all subscribes.
-            // If it were broken out by `GlobalId` then we could drop only those of dataflows we drop.
+            // We must drop the response buffers as they are global across all subscribe/copy-tos.
+            // If they were broken out by `GlobalId` then we could drop only those of dataflows we drop.
```
antiguru:
```suggestion
            // If they were broken out by `GlobalId` then we could drop only the response buffers of dataflows we drop.
```

antiguru (approved):
Looks good, left some comments around documentation. Thank you!

src/compute/src/server.rs
teskje: The comment says that this should work, but I'm actually not sure.

For subscribes:
* There might be batches in the response buffer that are before the subscribe's new as-of. I'm pretty sure the controller would be confused seeing these. This could be fixed by making the controller more lenient, or by filtering enqueued batches during reconciliation.
* Clients might expect to get the snapshot from the new subscribe. The introspection subscribes do, at least. To get the snapshot we'll always need to recreate the subscribe dataflow.

For copy-tos:
* I _think_ we can get into a state where the copy-to dataflow has not yet been dropped from the replica state, but the response has already been sent to the controller. In this state, if we don't recreate the copy-to dataflow, we would drop it without ever sending a response over the new controller connection, so the controller would be left waiting forever.

teskje:
TFTR!

PR #33309, opened 2025-08-12
Remove optional types from LIR

frankmcsherry:
Our LIR types had space for optional types that were meant to inform specialization. There is no specialization and no concrete plans to implement such from type information. In particular, the type system itself may evolve, and until it stabilizes it makes some sense to extract these and re-introduce them once we have a clearer sense of what they would be for (vs maintaining them).

### Checklist

### DIFF
--- a/src/compute-types/src/plan.rs
+++ b/src/compute-types/src/plan.rs
@@ -25,7 +25,7 @@ use mz_proto::{IntoRustIfSome, ProtoType, RustType, TryFromProtoError};
 use mz_repr::explain::text::text_string_at;
 use mz_repr::explain::{DummyHumanizer, ExplainConfig, ExprHumanizer, PlanRenderingContext};
 use mz_repr::optimize::OptimizerFeatures;
-use mz_repr::{ColumnType, Diff, GlobalId, Row};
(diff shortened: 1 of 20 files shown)


src/compute-types/src/plan.rs
```diff
@@ -108,7 +104,7 @@ impl RustType<ProtoAvailableCollections> for AvailableCollections {
         ProtoAvailableCollections {
             raw: self.raw,
             arranged: self.arranged.into_proto(),
-            types: self.types.into_proto(),
+            types: None::<Vec<ColumnType>>.into_proto(),
```
antiguru: We can remove the `types` field from the proto message, too. No backwards compatibility required for LIR.

antiguru (approved):
Looks good, but I would remove the `types` from the proto, too.

src/compute-types/src/plan.rs
frankmcsherry: Ah, ok. I wasn't sure what was going on there so I thought I was clever. Nvm!

frankmcsherry: Pushed a commit with this. Seems plausibly right, but ended up deleting more than a field.

PR #33311, opened 2025-08-12
Fix cargo docs following new nightly version

def-:
Seen in https://buildkite.com/materialize/deploy/builds/19309#_

Follow-up to https://github.com/MaterializeInc/materialize/pull/33290

Reproducer: `rustup run nightly bin/doc --document-private-items`

### Checklist

### DIFF
--- a/src/fivetran-destination/src/lib.rs
+++ b/src/fivetran-destination/src/lib.rs
@@ -7,8 +7,6 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-#![allow(dead_code)]
-
 mod crypto;
 mod destination;
 mod error;
--- a/src/fivetran-destination/src/lib.rs
+++ b/src/fivetran-destination/src/lib.rs
@@ -7,6 +7,8 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+#![allow(dead_code)]
+
 mod crypto;
 mod destination;
 mod error;
--- a/src/interchange/src/avro/decode.rs
+++ b/src/interchange/src/avro/decode.rs
@@ -7,16 +7,13 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::cell::RefCell;
 use std::collections::BTreeMap;
-use std::rc::Rc;
 
 use anyhow::{Context, Error};
 use mz_avro::error::{DecodeError, Error as AvroError};
 use mz_avro::{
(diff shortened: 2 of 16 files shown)


src/sql-parser/src/ast/defs/name.rs
```diff
@@ -275,7 +275,7 @@ impl Ident {
     }
 
     /// An identifier can be printed in bare mode if
-    ///  * it matches the regex [a-z_][a-z0-9_]* and
+    ///  * it matches the regex \[a-z_\]\[a-z0-9_\]* and
```
teskje: This is less readable now, and somewhat misleading if you don't view the rendered markdown. Can we instead put the regex in a code block?
```suggestion
    ///  * it matches the regex `[a-z_][a-z0-9_]*` and
```

src/ore/src/cast.rs
```diff
@@ -60,10 +60,12 @@ macro_rules! cast_from {
...
-            /// Casts [`$from`] to [`$to`].
-            ///
-            /// This is equivalent to the [`crate::cast::CastFrom`] implementation but is
-            /// available as a `const fn`.
+            #[doc = concat!(
+                "Casts [`", stringify!($from), "`] to [`", stringify!($to), "`].\n",
+                "\n",
+                "This is equivalent to the [`crate::cast::CastFrom`] implementation but is\n",
+                "available as a `const fn`."
+            )]
```
teskje: I think this isn't needed. We should just change the original docstring to say "Casts `from` to `to`." (i.e. drop the brackets and dollar signs.
```suggestion
            /// Casts `from` to `to`.
            ///
            /// This is equivalent to the [`crate::cast::CastFrom`] implementation but is
            /// available as a `const fn`.
```

src/interchange/src/avro/decode.rs
```diff
@@ -133,6 +133,7 @@ impl Decoder {
     }
 }
 
+#[allow(dead_code)]
```
teskje: Can we just remove this type instead?

src/fivetran-destination/src/lib.rs
```diff
@@ -7,6 +7,8 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+#![allow(dead_code)]
```
teskje: I think we want to avoid disabling the lint for the whole crate. Can we place the allow somewhere more targeted? What's it complaining about?

def-: Weird, I just removed it and am not seeing any warning. Will try a clean rebuild, but maybe we can merge with this removed.

def-: No failures, so good to go!

teskje (approved):
LGTM!

PR #33316, opened 2025-08-13
adapter: remove non-0dt code paths

aljoscha:
These have not been used in production for close to a year now. This also tries and removes tests that uses the old code paths. Let's see what ci/nightly has to say.


test/legacy-upgrade/mzcompose.py
```diff
@@ -222,13 +159,10 @@ def test_upgrade_from_version(
...
     priors: list[MzVersion],
     filter: str,
-    zero_downtime: bool,
     force_source_table_syntax: bool,
     self_managed_upgrade: bool = False,
 ) -> None:
-    print(
-        f"+++ Testing {'0dt upgrade' if zero_downtime else 'regular upgrade'} from Materialize {from_version} to current_source."
-    )
+    print(f"+++ Testing 0dt upgrade from Materialize {from_version} to current_source.")
```
antiguru:
```suggestion
    print(f"+++ Testing upgrade from Materialize {from_version} to current_source.")
```

test/legacy-upgrade/mzcompose.py
```diff
@@ -335,12 +264,9 @@ def test_upgrade_from_version(
 
             system_parameter_defaults = get_default_system_parameters(
                 version=version,
-                zero_downtime=zero_downtime,
             )
 
-            print(
-                f"'{'0dt-' if zero_downtime else ''}Upgrading to in-between version {version}"
-            )
+            print(f"''0dt-Upgrading to in-between version {version}")
```
antiguru:
```suggestion
            print(f"''Upgrading to in-between version {version}")
```

test/legacy-upgrade/mzcompose.py
```diff
@@ -360,25 +286,18 @@ def test_upgrade_from_version(
...
+                    DeploymentStatus.READY_TO_PROMOTE, mz_service
+                )
+                c.promote_mz(mz_service)
+                c.await_mz_deployment_status(DeploymentStatus.IS_LEADER, mz_service)
+                mz_service = (
+                    "materialized2" if mz_service == "materialized" else "materialized"
+                )
+                deploy_generation += 1
+
+    print("0dt-Upgrading to final version")
```
antiguru:
```suggestion
    print("Upgrading to final version")
```

aljoscha:
@def- I tried updating the tests but you might want to have to look to see if this matches what you expect. Plus I think we're seeing some failures in nightly that are very likely caused by me messing this up

aljoscha:
@def- Huh? How was that small change all that's needed? Also, thank you!

def-:
We previously split the 0dt and non-0dt upgrades into two parts and ran them in parallel. Since you now removed the non-0dt upgrades the test complained that nothing happened (services.log is empty).

aljoscha:
Duh! 🙈

PR #33317, opened 2025-08-13
docs: Add `WITH ORDINALITY`

ggevay:
As it says on the tin.

Our support for `WITH ORDINALITY` used to have serious issues, so it was better that it was not mentioned in our docs up to now. But we've fixed these issues recently (https://github.com/MaterializeInc/materialize/pull/33163, https://github.com/MaterializeInc/materialize/pull/33291, https://github.com/MaterializeInc/materialize/pull/33272), so it's time to add it to the docs.

### Checklist

### DIFF
--- a/doc/user/content/sql/functions/table-functions.md
+++ b/doc/user/content/sql/functions/table-functions.md
@@ -8,66 +8,90 @@ menu:
 
 ## Overview
 
-[Table functions]((/sql/functions/#table-functions)) return multiple rows from one input row. They are typically used in the `FROM` clause, where their arguments are allowed to refer to columns of earlier tables in the `FROM` clause.
+[Table functions](/sql/functions/#table-functions) return multiple rows from one
+input row. They are typically used in the `FROM` clause, where their arguments
+are allowed to refer to columns of earlier tables in the `FROM` clause.
 
-For example, the `unnest` table function expands a list into a collection of rows, where each row will contain one item of the list. For example, if we have the following input data:
+For example, consider the following table whose rows consist of lists of
+integers:
+
 ```mzsql
-CREATE TABLE t(l int list);
-INSERT INTO t VALUES (LIST[5, 7, 8]), (LIST[3, 3]);
+CREATE TABLE quizzes(scores int list);
+INSERT INTO quizzes VALUES (LIST[5, 7, 8]), (LIST[3, 3]);
 ```
-then
+
+Query the `scores` column from the table:
+
 ```mzsql
-SELECT l
-FROM t;
+SELECT scores
+FROM quizzes;
 ```
-gives us one list per row:
+
+The query returns two rows, where each row is a list:
+
 ```
-    l
+ scores
 ---------
  {3,3}
  {5,7,8}
 (2 rows)
 ```
-but
+
+Now, apply the [`unnest`](/sql/functions/#unnest) table function to expand the
+`scores` list into a collection of rows, where each row contains one list item:
+
 ```mzsql
-SELECT l, e
+SELECT scores, score
 FROM
-  t,
-  unnest(l) AS e;
+  quizzes,
+  unnest(scores) AS score; -- In Materialize, shorthand for AS t(score)
 ```
-gives us one list item per row:
+
+The query returns 5 rows, one row for each list item:
+
 ```
-    l    | e
----------+---
- {3,3}   | 3
- {3,3}   | 3
- {5,7,8} | 5
- {5,7,8} | 7
- {5,7,8} | 8
+ scores  | score
+---------+-------
+ {3,3}   |     3
+ {3,3}   |     3
+ {5,7,8} |     5
+ {5,7,8} |     7
+ {5,7,8} |     8
 (5 rows)
 ```
-(The `l` column, which contains as many copies of each input list as the list length, is included in the above `SELECT`'s output columns for demonstration purposes only; in real queries, you would typically omit the original list from the `SELECT` to avoid blowing up the data size.)
+
+{{< tip >}}
+
+For illustrative purposes, the original `scores` column is included in the
+results (i.e., query projection). In practice, you generally would omit
+including the original list to minimize the return data size.
+
+{{</ tip >}}
 
 ## `WITH ORDINALITY`
 
-You can also add `WITH ORDINALITY` after a table function call in a `FROM`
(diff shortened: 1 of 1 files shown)


doc/user/data/sql_funcs.yml
```diff
@@ -718,7 +718,11 @@
     url: /sql/types/jsonb#to_jsonb
 
 - type: Table
-  description: Table functions evaluate to a set of rows, rather than a single expression.
+  description: |
+    Table functions evaluate to a set of rows, rather than a single expression.
+    You can also add `WITH ORDINALITY` after the closing parenthesis of a table function call in a `FROM` clause.
+    This causes an `ordinality` column to appear, which is a 1-based numbering of the output rows of each call of
+    the table function. For example, `unnest(list) WITH ORDINALITY` numbers the unnested elements of each input list.
```
kay-kim: Maybe?

```suggestion
    Table functions evaluate to a set of rows rather than a single expression.

    {{< annotation type="WITH ORDINALITY" >}}

    You can also add `WITH ORDINALITY` after the table function call in a `FROM`
    clause. This causes an `ordinality` column to appear, which is a 1-based
    numbering of the output rows of each call of the table function. 

    For example, `SELECT * FROM unnest(LIST[5,10]) WITH ORDINALITY AS
    list_item;` (or `SELECT * FROM unnest(LIST[5,10]) AS list_item WITH
    ORDINALITY;`) includes the `ordinality` column containing the 1-based
    numbering of the unnested elements.

    {{</ annotation >}}
```

kay-kim (approved):
lgtm -- left a suggestion.

doc/user/data/sql_funcs.yml
ggevay: Thank you! I've made some further adjustments:
- "after the table function call" -> "after a table function call", because there can be multiple table function calls in a FROM clause.
- I've dropped the "(or `SELECT * FROM unnest(LIST[5,10]) AS list_item WITH
    ORDINALITY;`)", because this is not a syntax that I'd like people to use. I think no other system allows this. We used to allow only this syntax, but that was just a parser glitch.
- "of the unnested elements" -> "of the unnested items", to align the wording with the above `list_item` column name.

kay-kim: Works for me!

ggevay:
Thank you for the review!

ggevay:
Not merging yet. I'll merge it, when the fixes go out next week Aug 21.

ggevay:
I've just realized that `ROWS FROM` was also undocumented. I've added one more commit to the PR to document this as well. Could you check it @kay-kim ? I'm not sure whether this `{{< annotation type="ROWS FROM" >}}` box is appropriate for this one as well, because this one is bigger than the other box on WITH ORDINALITY.

doc/user/data/sql_funcs.yml
```diff
@@ -718,7 +718,60 @@
...
+    clause. This causes an `ordinality` column to appear, which is a 1-based
+    numbering of the output rows of each call of the table function.
+
+    For example, `SELECT * FROM unnest(LIST[5,10]) WITH ORDINALITY AS
+    list_item;` includes the `ordinality` column containing the 1-based
+    numbering of the unnested items.
+
+    {{</ annotation >}}
+
+    {{< annotation type="ROWS FROM" >}}
```
kay-kim: Oh ... I wonder if most of the content should be actually an idiomatic SQL thing for ROWS FROM and we just link to that from here instead of a whole annotation thing.

ggevay: Well, the idiomatic SQL pages are for when there is a particular preferred way of expressing a thing in Materialize, and it is possibly different from other systems' preferred way. `ROWS FROM` is not like that: you can simply write it the same way in Materialize as in other systems.

Postgres documents both `WITH ORDINALITY` and `ROWS FROM` in the SELECT page, at the `FROM` clause (at `function_name`):
https://www.postgresql.org/docs/current/sql-select.html#SQL-FROM
We could do the same, i.e., expand `table_expr` in [our SELECT page](https://materialize.com/docs/sql/select/).

kay-kim: Ah, got it.  In that case, 100% agree with your suggestion to move these to our select page and just link to there.

ggevay: Ok, I'll do that move now.

ggevay: Actually, how about we create a separate page just for table function calls? The problem is that both `table_expr`/`table_ref` in the SELECT page and `table_ref` in the JOIN page would need to allow WITH ORDINALITY and ROWS FROM, but we don't want to repeat this on both pages. So, we could create a separate page, and just link to there from the SELECT page and JOIN page (and table function list) where we explain `table_expr`. (`table_ref` should be `table_expr` everywhere.)

ggevay: Having a separate page would be similar also to how the various extra things around aggregation function calls are explained in separate pages, e.g. 
- The FILTER clause: https://materialize.com/docs/sql/functions/filters/
- ORDER BY inside an aggregation: https://materialize.com/docs/sql/functions/list_agg/

kay-kim: That works.  Redoing the SELECT/JOIN/CTE pages are on my todo list; so, keeping this change clean works for me.

ggevay: Could you please check again @kay-kim ?

I've pushed a commit for the above:
- Created a new table functions page, moved the WITH ORDINALITY and ROWS FROM content to it, plus added some more content, and linked it in a few places.
- fixed some minor things:
  - The SELECT page had `table_ref` and `table_expr` inconsistently. Now it has `table_expr` in both places. (It would have been good to also modify the same `table_ref` to `table_expr` in the JOIN page, but I didn't do that, because I would have needed to touch the grammar, but `sql-grammar/generate.sh` was not really working for me.)
  - I've fixed the most pressing part of https://github.com/MaterializeInc/database-issues/issues/8311#issuecomment-3201977113

kay-kim: Thank you 😍 😍😍 ... Added a patch with some minor tweaks.

ggevay:
Thank you!

I'll merge it on Thursday after the release is rolled out.

PR #33321, opened 2025-08-13
[optimizer] MIR representation types

mgree:
Separates the notion of SQL types from the actual types of our representation.

### Motivation

     https://github.com/MaterializeInc/materialize/pull/27239

### Checklist


src/repr/src/relation.rs
```diff
@@ -259,6 +259,24 @@ impl RustType<ProtoKey> for Vec<usize> {
     }
 }
 
+#[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize, Hash, MzReflect)]
+pub struct ReprColumnType {
+    /// The underlying representation scalar type (e.g., Int32 or String) of this column.
+    pub scalar_type: ReprScalarType,
+    /// Whether this datum can be null.
+    #[serde(default = "return_true")]
+    pub nullable: bool,
```
ggevay, line 268: I'm wondering why we have the `#[serde(default = "return_true")]`. It's also present on the old `ColumnType`, but I don't know why. It's making me slightly nervous, because we do rely on this serde derivation for serious stuff nowadays, so we better not lose any nullabilities: https://materializeinc.slack.com/archives/C08A62E0751/p1755592398171209?thread_ts=1755590706.213029&cid=C08A62E0751

I think this annotation just specifies what to do when the field is not present in the serialized representation, but it's not clear why it wouldn't be present. Could we just try removing it and see whether all tests pass?

ggevay:
Nightly: https://buildkite.com/materialize/nightly/builds/12895

src/repr/src/scalar.rs
```diff
@@ -3728,6 +3732,118 @@ impl Arbitrary for SqlScalarType {
...
 }
 
+/// The type of a [`Datum`] as it is represented.
+///
+/// Each variant here corresponds to one or more variants of [`SqlScalarType`].
+///
+/// There is a direct correspondence between `Datum` variants and `ScalarType`
+/// variants.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Ord, PartialOrd, Hash, MzReflect)]
+pub enum ReprScalarType {
```
ggevay, line 3879: >
```
/// There is a direct correspondence between `Datum` variants and `ScalarType`
/// variants.
```

`ReprScalarType` variants?

ggevay: Also, maybe "direct correspondence" could be expanded a bit. E.g., it's not a bijection, because of `Datum::False` and `Datum::True` are both `ReprScalarType::Bool`. Maybe this is the only exception from being a bijection?

src/repr/src/scalar.rs
```diff
@@ -3728,6 +3732,118 @@ impl Arbitrary for SqlScalarType {
...
+    Float64,
+    Numeric,
+    Date,
+    Time,
+    Timestamp {
+        precision: Option<TimestampPrecision>,
+    },
+    TimestampTz {
+        precision: Option<TimestampPrecision>,
+    },
```
ggevay, line 3898: I'm wondering whether `precision` could also be omitted. (But it's also ok to consider this out of scope from this PR, if it's not immediately clear.)

src/repr/src/relation.rs
ggevay: Oh, maybe it's some `lowertest`/`mzreflect` thing? But then could it be `#[mzreflect(ignore)] ` instead?

ggevay (commented):
Looks good to me, wrote some minor comments.

Also, I couldn't figure out the reason for the new parameter to `arb_datum`. In my IDE, it looks like as if it were only called with `true`, but maybe my IDE missed some usage.

mgree:
https://github.com/MaterializeInc/materialize/blob/c9c52d33d4aa63b1213d15134032a05b196b71d6/src/repr/src/scalar.rs#L4704-L4711 uses `arb_datum(false)` to make sure we don't generate any `Datum::Dummy` when trying `is_instance_of` on data that may be invalid (since we'll panic on a dummy).

src/repr/src/relation.rs
mgree: Absolutely no idea here. Blame is four years old. Will fire up a PR and see if any tests yelp.

src/repr/src/scalar.rs
mgree: It's not immediately clear to me. `base_eq` seems to ignore the precision field, but precision matters for some operations, for casting to JSON ([`build_row_schema_field_type`](https://github.com/mgree/materialize/blob/ae822d2354e316926e2930fe02ec00035363b6a6/src/interchange/src/json.rs#L296-L302)), for avro schemas ([`validate_schema_2`](https://github.com/mgree/materialize/blob/ae822d2354e316926e2930fe02ec00035363b6a6/src/interchange/src/avro/schema.rs#L214-L219)), and DB interop with mysql/postgres/sqlserver. It _seems_ a bit like fixed lengths on strings (affects semantics of operations, but not types, morally speaking), but I'm not 100% on it. Maybe leave it out of scope, for now?

ggevay: Ok

src/repr/src/relation.rs
ggevay: Ok, thanks! If it's only mzreflect tests that complain, that will be also good to know that it's specifically for that.

mgree: https://github.com/MaterializeInc/materialize/pull/33382, expected to fail (`cargo test` failed locally): mzreflect needs this to parse nullable types:

```
tests/testdata/union-fusion:10:
(defsource x [int64 int64])

expected:
ok

actual:
error: String while serializing: missing field `nullable` at line 1 column 71
Original JSON: {"Defsource":{"name":"x","typ":{"column_types":[{"scalar_type":"Int64"},{"scalar_type":"Int64"}]}}}
```

ggevay: Ok, I see! Thanks for trying! I've now also tried `#[mzreflect(ignore)]`, and we get the same error with that as well. We can just leave it as it is then. (I'll still check back on the PR later to see that mzreflect stuff was the only failure. Edit: Done, that's exactly what happened.)

ggevay (approved):

PR #33322, opened 2025-08-14
Cancel postgresql snapshot queries on drop

martykulma:
Capture the `tokio_postgres::Connection` as part of Client so it can be terminated
when Client is dropped. 

### Motivation

Fixes one of the issues in https://github.com/MaterializeInc/database-issues/issues/9548

### Checklist

### DIFF
--- a/src/postgres-util/src/tunnel.rs
+++ b/src/postgres-util/src/tunnel.rs
@@ -15,14 +15,13 @@ use std::time::Duration;
 use mz_ore::future::{InTask, OreFutureExt};
 use mz_ore::netio::DUMMY_DNS_PORT;
 use mz_ore::option::OptionExt;
-use mz_ore::task::{self, AbortOnDropHandle, JoinHandle};
+use mz_ore::task::{self, AbortOnDropHandle};
 use mz_proto::{RustType, TryFromProtoError};
 use mz_repr::CatalogItemId;
 use mz_ssh_util::tunnel::{SshTimeoutConfig, SshTunnelConfig};
 use mz_ssh_util::tunnel_manager::SshTunnelManager;
 use proptest_derive::Arbitrary;
 use serde::{Deserialize, Serialize};
-use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::net::TcpStream as TokioTcpStream;
 use tokio_postgres::config::{Host, ReplicationMode};
 use tokio_postgres::tls::MakeTlsConnect;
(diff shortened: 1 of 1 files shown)


martykulma:
Having the connection abort by default caused a huge amount of test failures: https://buildkite.com/materialize/test/builds/107881

def-:
Why are the tests timing out?

martykulma:
> Why are the tests timing out?

I assume you mean for the original implementation that would always forcefully close the connection. For the ones I have checked, the issues ends up being a stalled source because the connection is closed.  
```
Broadcasting suspend-and-restart command because of Some((Postgres, Stalled { error: "sql client error: connection closed", hint: None, should_halt: true })) after 5s delay
```

I'm hoping to have a followup PR that makes this general, but for now the main issue is the snapshots as they can be stuck reading a ton of data.

martykulma:
> I'm hoping to have a followup PR that makes this general, but for now the main issue is the snapshots as they can be stuck reading a ton of data.

Sorry @def- , I just realized could have put that in the description!  Updated with an example of what I *think* is causing it

martykulma:
I ran into issues testing locally, so I opened another [PR](https://github.com/MaterializeInc/materialize/pull/33340) to test in CI.  It does look like there's a single source for all the test failures 🤦

martykulma:
Sorry for the eager PR, when I looked at the failures the first time, I thought I had multiple regressions on my hands.  The latest commit is more surface area, but tests seem to have run here and for https://github.com/MaterializeInc/materialize/pull/33340.

I'm not sure it's worth implementing `Drop` just to log that we are dropping the connection, but I'm open to suggestions. Without that logging, there's not really anything to test for in automation because I couldn't even catch the `COPY` in `pg_stat_activity` prior to adding the drop handling.  I'd tried locking the table, setting up `pg_cron`, and even a python thread that just kept querying `pg_stat_activity` while the test was running.

petrosagg (approved):
Looks great! That one line made the tests explode :D

The wrapper's `Client::new` doesn't seem to be pulling its weight. It's only there to get the server version.. We could delete it and construct a client struct in the three constructors.

petrosagg:
> there's not really anything to test for in automation because I couldn't even catch the COPY in pg_stat_activity prior to adding the drop handling.

Would it help if you wrote a test drive tests that first connected to pg and grabbed an exclusive lock on a table, then create a source to read that table which will block, verify that you see the copy in pg internal tables, drop the source, and verify that the query is gone?

martykulma:
> Looks great! That one line made the tests explode :D
> 
> The wrapper's `Client::new` doesn't seem to be pulling its weight. It's only there to get the server version.. We could delete it and construct a client struct in the three constructors.

Thanks Petros. Yeah, that was it. Such drama!

Good point. `server_version` appears to only be used to determine if it's PG or YB, and it will get removed per https://github.com/MaterializeInc/database-issues/issues/8572#issuecomment-2981590799.





> Would it help if you wrote a test drive tests that first connected to pg and grabbed an exclusive lock on a table, then create a source to read that table which will block, verify that you see the copy in pg internal tables, drop the source, and verify that the query is gone?

I tried `LOCK TABLE .. IN ACCESS EXCLUSIVE MODE`, but it unfortunately blocks the query we use to get the publication info -  https://github.com/MaterializeInc/materialize/blob/3cc2eb8662351feb31dc57bd9d52273f0b18d51f/src/postgres-util/src/schemas.rs#L47

Specifically, it's the pg_publication_tables view that can't be read.  I suspect it's trying to get a sharelock on the table, and the exclusive lock blocks it.

PR #33323, opened 2025-08-14
adapter: exclude crash-looping clusters from 0dt caught up checks

aljoscha:
We already have logic in the caught-up check that tries to exclude collections whose frontier is too far behind current wall-clock time, but that doesn't catch cases where a replica is "limping" along and sometimes makes process. This has made the release tedious at times, where we had to wait a long time for cutting over.

This is a more explicit approach where we straightaway ignore clusters that only have crash-looping replicas. We don't consider and wait for them before cutting over in a 0dt deployment.

@ggevay and @teskje , you might be interested to review, I think you have most acutely felt the pain of this recently. 🕊️

### DIFF
--- a/src/adapter/src/coord/caught_up.rs
+++ b/src/adapter/src/coord/caught_up.rs
@@ -11,16 +11,21 @@
 //! deployment.
 
 use std::collections::{BTreeMap, BTreeSet};
+use std::time::Duration;
 
 use differential_dataflow::lattice::Lattice as _;
+use futures::StreamExt;
 use itertools::Itertools;
 use mz_adapter_types::dyncfgs::{
-    ENABLE_0DT_CAUGHT_UP_CHECK, WITH_0DT_CAUGHT_UP_CHECK_ALLOWED_LAG,
-    WITH_0DT_CAUGHT_UP_CHECK_CUTOFF,
+    ENABLE_0DT_CAUGHT_UP_CHECK, ENABLE_0DT_CAUGHT_UP_REPLICA_STATUS_CHECK,
+    WITH_0DT_CAUGHT_UP_CHECK_ALLOWED_LAG, WITH_0DT_CAUGHT_UP_CHECK_CUTOFF,
 };
-use mz_catalog::builtin::MZ_CLUSTER_REPLICA_FRONTIERS;
+use mz_catalog::builtin::{MZ_CLUSTER_REPLICA_FRONTIERS, MZ_CLUSTER_REPLICA_STATUS_HISTORY};
 use mz_catalog::memory::objects::Cluster;
+use mz_controller_types::ReplicaId;
+use mz_orchestrator::OfflineReason;
 use mz_ore::channel::trigger::Trigger;
+use mz_ore::now::EpochMillis;
 use mz_repr::{GlobalId, Timestamp};
 use timely::PartialOrder;
 use timely::progress::{Antichain, Timestamp as _};
@@ -139,6 +144,18 @@ impl Coordinator {
 
         let now = self.now();
 
+        // Something might go wrong with querying the status collection, so we
+        // have an emergency flag for disabling it.
+        let replica_status_check_enabled =
+            ENABLE_0DT_CAUGHT_UP_REPLICA_STATUS_CHECK.get(self.catalog().system_config().dyncfgs());
+
+        // Analyze replica statuses to detect crash-looping or OOM-looping replicas
+        let problematic_replicas = if replica_status_check_enabled {
+            self.analyze_replica_looping(now).await
+        } else {
+            BTreeSet::new()
+        };
+
         let compute_caught_up = self
             .clusters_caught_up(
                 allowed_lag.into(),
@@ -146,6 +163,7 @@ impl Coordinator {
                 now.into(),
                 &live_collection_frontiers,
                 &ctx.exclude_collections,
+                &problematic_replicas,
             )
             .await;
 
@@ -176,6 +194,7 @@ impl Coordinator {
         now: Timestamp,
         live_frontiers: &BTreeMap<GlobalId, Antichain<Timestamp>>,
         exclude_collections: &BTreeSet<GlobalId>,
+        problematic_replicas: &BTreeSet<ReplicaId>,
     ) -> bool {
         let mut result = true;
         for cluster in self.catalog().clusters() {
@@ -187,6 +206,7 @@ impl Coordinator {
                     now.clone(),
                     live_frontiers,
                     exclude_collections,
+                    problematic_replicas,
                 )
                 .await;
 
@@ -229,8 +249,23 @@ impl Coordinator {
         now: Timestamp,
         live_frontiers: &BTreeMap<GlobalId, Antichain<Timestamp>>,
         exclude_collections: &BTreeSet<GlobalId>,
+        problematic_replicas: &BTreeSet<ReplicaId>,
     ) -> Result<bool, anyhow::Error> {
         if cluster.replicas().next().is_none() {
+            return Ok(true);
+        }
+
+        // Check if all replicas in this cluster are crash/OOM-looping. As long
+        // as there is at least  one healthy replica, the cluster is okay-ish.
+        let cluster_has_only_problematic_replicas = cluster
+            .replicas()
+            .all(|replica| problematic_replicas.contains(&replica.replica_id));
+
+        if cluster_has_only_problematic_replicas {
+            tracing::info!(
+                "ALL replicas of cluster {} crash/OOM-looping, ignoring for caught-up checks",
+                cluster.id
+            );
             return Ok(true);
         }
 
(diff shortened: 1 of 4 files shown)


def- (commented):
Nightly failures (so far) are unrelated, please ignore. Keeping this always on in CI makes sense.

src/adapter/src/coord/caught_up.rs
```diff
@@ -385,4 +419,139 @@ impl Coordinator {
...
+            .await
+            .expect("can't read mz_cluster_replica_status_history");
+
+        let mut replica_problem_counts: BTreeMap<ReplicaId, u32> = BTreeMap::new();
+
+        while let Some((source_data, _ts, diff)) = replica_statuses_stream.next().await {
+            // Only process inserts (positive diffs)
+            if diff <= 0 {
+                continue;
+            }
```
ggevay, line 478: `snapshot_and_stream`'s doc comment says "The output is __not__ consolidated.", so I think we should consolidate it ourselves to not need to handle negative diffs here. (I guess we could even soft_assert that there are no negative diffs after consolidation.)

src/adapter/src/coord/caught_up.rs
```diff
@@ -385,4 +419,139 @@ impl Coordinator {
...
+
+            // Extract the Row from SourceData
+            let row = match source_data.0 {
+                Ok(row) => row,
+                Err(_) => continue, // Skip rows with errors
```
ggevay: If there is a dataflow error, then all the non-error data is considered invalid (because some error checks are on a different branch of a dataflow from the normal data processing path, so it happens that the dataflow does emit also some data even when there is an error; this happens for example in the scalar subquery results size check). So, it would be better to return with an empty set in this case. (And maybe even report it to Sentry, because it's weird if this collection has errors, right?)

src/adapter/src/coord/caught_up.rs
```diff
@@ -385,4 +419,139 @@ impl Coordinator {
             ctx.trigger.fire();
         }
     }
+
+    /// Analyzes replica status history to detect replicas that are
+    /// crash-looping or OOM-looping.
+    ///
+    /// A replica is considered problematic if it has multiple failures in a
+    /// short time window. This includes crashes, OOM kills, or other non-ready
+    /// states that repeat frequently.
```
ggevay: I think the comment is out of sync with `is_problematic_status`.

src/adapter/src/coord/caught_up.rs
```diff
@@ -385,4 +419,139 @@ impl Coordinator {
...
+        result
+    }
+
+    /// Determines if a replica status indicates a problematic state that could
+    /// indicate looping.
+    fn is_problematic_status(_status: &str, reason: Option<&str>) -> bool {
+        // For now, we only look at the reason, but we could change/expand this
+        // if/when needed.
+        if let Some(reason) = reason {
+            return reason == "oom-killed";
```
ggevay: We could factor out this string into a constant, and use it also at https://github.com/MaterializeInc/materialize/blob/563ac22eaaa8318eb38ad2fe396f32855d62c7b8/src/orchestrator/src/lib.rs#L108

src/adapter/src/coord/caught_up.rs
```diff
@@ -385,4 +419,139 @@ impl Coordinator {
...
+            let reason_datum = iter.next().expect("missing reason");
+            let reason = if reason_datum.is_null() {
+                None
+            } else {
+                Some(reason_datum.unwrap_str().to_string())
+            };
+            let occurred_at = iter
+                .next()
+                .expect("missing occurred_at")
+                .unwrap_timestamptz();
```
ggevay, line 518: This is in `mz_internal`, so the schema might change. You could add a comment at the definition of `MZ_CLUSTER_REPLICA_STATUS_HISTORY` calling out that this code also needs to be changed if somebody wants to change the schema.

ggevay (commented):
I wrote some comments.

Btw. I'm wondering whether it would have been a viable alternative approach to fire off a peek that has most of the logic written in SQL instead of manually reading the collection. Manually reading the collection seems a bit delicate (consolidation, error handling). Also, the SQL logic could be encapsulated in a (non-indexed) builtin view, which would have the advantage that it's easy to query it also outside of a 0dt procedure, and also would immediately panic on startup if there is an incompatible schema change.

src/adapter/src/coord/caught_up.rs
aljoscha: I didn't do that on purpose so that we can stream through the data here, without having to materialize in memory. If there's a problematic status with a positive diff inside our time window, that's already enough signal, IMO. And materializing in memory and consolidating wouldn't give us anything.

src/adapter/src/coord/caught_up.rs
aljoscha: This can't have errors because it's only us writing to it, so I should at least log a warning or error. But even if there are errors, the signal of the problematic statuses is still valid.

src/adapter/src/coord/caught_up.rs
aljoscha: I'll change the code here to directly use `OfflineReason::OomKilled`, okay?

ggevay: Sounds good!

src/adapter/src/coord/caught_up.rs
aljoscha: I was hoping tests would highlight that, and I don't know how likely it is that the person would read this comment. IMO it's more likely they will notice when this code fails.

src/adapter/src/coord/caught_up.rs
aljoscha: will fix!

aljoscha:
> Btw. I'm wondering whether it would have been a viable alternative approach to fire off a peek that has most of the logic written in SQL instead of manually reading the collection. Manually reading the collection seems a bit delicate (consolidation, error handling). Also, the SQL logic could be encapsulated in a (non-indexed) builtin view, which would have the advantage that it's easy to query it also outside of a 0dt procedure, and also would immediately panic on startup if there is an incompatible schema change.

I thought about that as well, but in the end the code here is trying to use the most basic primitives and not rely on other machinery of the adapter/controller/etc. working. Plus there's already other code in here that directly reads from a collection this way.

src/adapter/src/coord/caught_up.rs
ggevay: Well, yes, some Nightly upgrade check would hopefully catch it.

src/adapter/src/coord/caught_up.rs
ggevay: It's just that it breaks an abstraction boundary to have to think about what we might find here if we don't consolidate. But if you feel it's safe, then I'm ok.

src/adapter/src/coord/caught_up.rs
ggevay: Yeah, again, it breaks an abstraction boundary to have to think about what it would mean for the non-error data if this had errors. In other words, it couples the internals of this code with the internals of the code that produces this collection. But if you feel it's safe, then I'm ok.

src/adapter/src/coord/caught_up.rs
ggevay: (Btw. I'd really hope that if somebody sets out to modify the schema of `MZ_CLUSTER_REPLICA_STATUS_HISTORY`, they would at least read a comment that is directly on top of the thing.)

ggevay (approved):
The code looks good to me!

I'm just a little bit worried about a possible failure scenario: 
- A user has a production incident where their prod cluster starts OOM-looping on a Wednesday (let's say due their input data increasing).
- It's fixed within a few hours, after OOMing 3 times (the threshold of this PR).
- Then we roll out a release on Thursday, and cause another production incident, because we don't wait for their cluster to catch up before cutting over, because we still consider it OOM looping, even though it was fixed a day before.

An additional safeguard could be to also check whether the cluster is still unhealthy at the current moment, by checking whether there is any collection on it that's not caught up. (In the Fubo issue last Thursday, there were different sets of collections caught up in the old and new envs, but both envs had _some_ collections not caught up.)

aljoscha:
@teskje and @ggevay, what I could do as an alternative, is to change the caught up check to only ignore clusters that have only "problematic replicas" AND have at least one collection that is too far behind the cutoff (https://github.com/MaterializeInc/materialize/blob/a6aa88d465024f82576e5ccad370d98c39b85cf6/src/adapter-types/src/dyncfgs.rs#L68). That cutoff is currently set at 2 hours, but with the additional inclusion of the problematic replica thing, I would tune that way down, to say 10 minutes.

Right now, the way it works is that if a collection's upper is to far behind the cutoff, it is ignored for the checks, but we still check other collections on the new cluster.

Summary of the above, we ignore a cluster for checks when both of these are true (AND):
- cluster has only "problematic replicas"
- cluster has at least one collection that is to far behind the (much reduced) cutoff

The goal here is to to work around the scenario Gábor mentioned above: A cluster that _had_ failures but is now healthy again.

ggevay:
This sounds good to me!

teskje:
SGTM as well!

aljoscha:
@ggevay and @teskje I pushed an extra commit that does the change. Does this match your understanding? 🙏

PR #33335, opened 2025-08-15
Move some timeline fns from Coordinator to Catalog

ggevay:
Works towards the [Small Coordinator vision](https://github.com/aljoscha/materialize/blob/design-small-coordinator/doc/developer/design/20250717_a_small_coordinator_more_scalable_isolated_materialize.md#processing-selects), specifically https://github.com/MaterializeInc/database-issues/issues/9593

These methods don't need to be on `Coordinator`, because they only work with the `Catalog`. This PR moves them to `Catalog`, making them easier to call outside the `Coordinator`.

This PR is purely code movement, no behavior change should happen from it.

### Motivation

### Checklist

### DIFF
--- a/src/adapter/src/catalog.rs
+++ b/src/adapter/src/catalog.rs
@@ -40,15 +40,15 @@ use mz_catalog::durable::{
 use mz_catalog::expr_cache::{ExpressionCacheHandle, GlobalExpressions, LocalExpressions};
 use mz_catalog::memory::error::{Error, ErrorKind};
 use mz_catalog::memory::objects::{
-    CatalogCollectionEntry, CatalogEntry, CatalogItem, Cluster, ClusterReplica, ContinualTask,
-    Database, MaterializedView, NetworkPolicy, Role, RoleAuth, Schema, View,
+    CatalogCollectionEntry, CatalogEntry, CatalogItem, Cluster, ClusterReplica, Database,
+    NetworkPolicy, Role, RoleAuth, Schema,
 };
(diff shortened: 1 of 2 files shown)


src/adapter/src/catalog.rs
```diff
@@ -1529,6 +1531,231 @@ impl Catalog {
         let builtin_table_updates = self.state.apply_updates(updates)?;
         Ok(builtin_table_updates)
     }
+
```
antiguru: Nit: The `catalog.rs` file is already very large, so I think it'd make sense to have it split across several files. This seems like an opportunity to do so?

aljoscha: Yeah, I'd say we can create a new module `catalog/timeline` and move the methods there. Wdyt?

aljoscha (approved):
Excellent! But would be good to move to a new module as Moritz suggested before merging.

src/adapter/src/catalog.rs
ggevay: Done!

ggevay: Btw., I've split off only the newly moved code now, but should I try to split off more stuff into separate files in a follow-up PR? Ideas:
- `CatalogPlans` and related methods (~200 lines)
- expression cache stuff
- `ExprHumanizer` (~150 lines)
- `ConnCatalog`
- ...

ggevay:
Merging! Can do more `catalog.rs` splitting later.

PR #9573, opened 2021-12-13
expr: concrete function definition for `pg_get_constraintref`

petrosagg:
### Motivation

Apart from the possibility of implementing it in the future, this aligns the function with the effort of migrating all functions away from the raw enum definition style.

### Checklist

- [x] This PR has adequate test coverage / QA involvement has been duly considered.

PR #33341, opened 2025-08-16
[sql_server] refactor lock and snapshot tables one by one

ptravers:
I'm cleaning this up and removing all the clones. just wanted a sanity check by running all the tests in CI.

### Motivation

    https://github.com/MaterializeInc/database-issues/issues/9512

### Checklist

### DIFF
--- a/src/sql-server-util/examples/cdc.rs
+++ b/src/sql-server-util/examples/cdc.rs
@@ -40,6 +40,8 @@ use mz_ore::future::InTask;
 use mz_sql_server_util::cdc::CdcEvent;
 use mz_sql_server_util::config::TunnelConfig;
 use mz_sql_server_util::{Client, Config};
+use std::collections::BTreeMap;
+use std::sync::Arc;
 use tracing_subscriber::EnvFilter;
 
 #[tokio::main]
@@ -69,21 +71,36 @@ async fn main() -> Result<(), anyhow::Error> {
     let mut client_2 = Client::connect(mz_config).await?;
     tracing::info!("connection 2 successful!");
 
-    // Get an initial snapshot of the table.
-    let (lsn, stats, snapshot) = cdc_handle
-        .snapshot(None, 1, mz_repr::GlobalId::User(1))
-        .await?;
-    tracing::info!("snapshot stats: {stats:?}");
-    {
+    let tables = mz_sql_server_util::inspect::get_tables_for_capture_instance(
+        &mut client_2,
+        capture_instances,
+    )
+    .await?;
+    let mut instance_to_lsn = BTreeMap::new();
+
+    for table in tables {
+        // Get an initial snapshot of the table.
+        let (lsn, stats, snapshot) = cdc_handle
+            .snapshot(&table, 1, mz_repr::GlobalId::User(1))
+            .await?;
+
+        tracing::info!("snapshot stats: {stats:?}");
+
+        instance_to_lsn.insert(Arc::clone(&table.capture_instance), lsn);
+
         let mut snapshot = std::pin::pin!(snapshot);
-        while let Some((capture_instance, result)) = snapshot.next().await {
+        while let Some(result) = snapshot.next().await {
             let row = result?;
-            tracing::info!("snapshot: {capture_instance} {row:?}");
+            tracing::info!("snapshot: {} {row:?}", &table.capture_instance);
         }
     }
 
     // Initialize all capture instances at the LSN we just snapshotted.
     for instance in capture_instances {
+        let lsn = instance_to_lsn
+            .remove(instance)
+            .expect("table must have instance");
+
         cdc_handle = cdc_handle.start_lsn(instance, lsn);
     }
     // Get a stream of changes from the table.
--- a/src/sql-server-util/src/cdc.rs
+++ b/src/sql-server-util/src/cdc.rs
@@ -61,7 +61,7 @@
 //! After completing the snapshot we use [`crate::inspect::get_changes_asc`] which will return
 //! all changes between a `[lower, upper)` bound of [`Lsn`]s.
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::BTreeMap;
 use std::fmt;
 use std::sync::Arc;
 use std::time::Duration;
@@ -73,6 +73,7 @@ use proptest_derive::Arbitrary;
 use serde::{Deserialize, Serialize};
 use tiberius::numeric::Numeric;
 
+use crate::desc::SqlServerTableRaw;
 use crate::{Client, SqlServerError, TransactionIsolationLevel};
 
 /// A stream of changes from a table in SQL Server that has CDC enabled.
@@ -143,68 +144,35 @@ impl<'a> CdcStream<'a> {
         self
     }
 
-    /// Takes a snapshot of the upstream table that the specified `capture_instance` is
-    /// replicating changes from.
-    ///
-    /// An optional `instances` parameter can be provided to only snapshot the specified instances.
+    /// Takes a snapshot of the upstream table that the specified `table` represents.
     pub async fn snapshot<'b>(
         &'b mut self,
-        instances: Option<BTreeSet<Arc<str>>>,
+        table: &SqlServerTableRaw,
         worker_id: usize,
         source_id: GlobalId,
     ) -> Result<
         (
             Lsn,
-            BTreeMap<Arc<str>, usize>,
-            impl Stream<Item = (Arc<str>, Result<tiberius::Row, SqlServerError>)> + use<'b, 'a>,
+            usize,
+            impl Stream<Item = Result<tiberius::Row, SqlServerError>>,
         ),
         SqlServerError,
     > {
         static SAVEPOINT_NAME: &str = "_mz_snap_";
 
-        // Determine what table we need to snapshot.
-        let instances = self
-            .capture_instances
-            .keys()
-            .filter(|i| match instances.as_ref() {
-                // Only snapshot the instance if the filter includes it.
-                Some(filter) => filter.contains(i.as_ref()),
-                None => true,
-            })
-            .map(|i| i.as_ref());
-        let tables =
-            crate::inspect::get_tables_for_capture_instance(self.client, instances).await?;
-        tracing::info!(%source_id, ?tables, "timely-{worker_id} got table for capture instance");
-
-        // Before starting a transaction where the LSN will not advance, ensure
-        // the upstream DB is ready for CDC.
-        self.wait_for_ready().await?;
-
-        // Intentionally logging this at info for debugging. This section won't get entered
-        // often, but if there are problems here, it will be much easier to troubleshoot
-        // knowing where stall/hang might be happening.
-        tracing::info!(%source_id, "timely-{worker_id} upstream is ready");
-
         // The client that will be used for fencing does not need any special isolation level
         // as it will be just be locking the table(s).
         let mut fencing_client = self.client.new_connection().await?;
         let mut fence_txn = fencing_client.transaction().await?;
-
-        // TODO improve table locking: https://github.com/MaterializeInc/database-issues/issues/9512
-        for (_capture_instance, schema, table) in &tables {
-            tracing::trace!(%source_id, %schema, %table, "timely-{worker_id} locking table");
-            fence_txn.lock_table_shared(&*schema, &*table).await?;
-        }
-
-        // So we know that we locked that tables and roughly how long that took based on the time diff
-        // from the last message.
-        tracing::info!(%source_id, "timely-{worker_id} locked tables");
+        fence_txn
+            .lock_table_shared(&table.schema_name, &table.name)
+            .await?;
+        tracing::info!(%source_id, %table.schema_name, %table.name, "timely-{worker_id} locked table");
 
         self.client
             .set_transaction_isolation(TransactionIsolationLevel::Snapshot)
             .await?;
         let mut txn = self.client.transaction().await?;
-
         // Creating a savepoint forces a write to the transaction log, which will
         // assign an LSN, but it does not force a transaction sequence number to be
         // assigned as far as I can tell.  I have not observed any entries added to
@@ -223,17 +191,17 @@ impl<'a> CdcStream<'a> {
             ))?
         }
 
-        // Because the tables are locked, any write operation has either
+        // Because the table is locked, any write operation has either
         // completed, or is blocked. The LSN and XSN acquired now will represent a
         // consistent point-in-time view, such that any committed write will be
         // visible to this snapshot and the LSN of such a write will be less than
         // or equal to the LSN captured here. Creating the savepoint sets the LSN,
         // we can read it after rolling back the locks.
         txn.create_savepoint(SAVEPOINT_NAME).await?;
-        tracing::info!(%source_id, %SAVEPOINT_NAME, "timely-{worker_id} created savepoint");
+        tracing::info!(%source_id, %table.schema_name, %table.name, %SAVEPOINT_NAME, "timely-{worker_id} created savepoint");
 
-        // Once the XSN is esablished and the LSN captured, the tables no longer
-        // need to be locked.  Any writes that happen to the upstream tables
+        // Once the savepoint is created (which establishes the XSN and captures the LSN),
+        // the table no longer needs to be locked. Any writes that happen to the upstream table
         // will have an LSN higher than our captured LSN, and will be read from CDC.
         fence_txn.rollback().await?;
 
@@ -241,42 +209,26 @@ impl<'a> CdcStream<'a> {
 
         tracing::info!(%source_id, ?lsn, "timely-{worker_id} starting snapshot");
 
-        // Get the size of each table we're about to snapshot.
-        //
-        // TODO(sql_server3): To expose a more "generic" interface it would be nice to
-        // make it configurable about whether or not we take a count first.
-        let mut snapshot_stats = BTreeMap::default();
-        for (capture_instance, schema, table) in &tables {
-            tracing::trace!(%capture_instance, %schema, %table, "snapshot stats start");
-            let size = crate::inspect::snapshot_size(txn.client, &*schema, &*table).await?;
-            snapshot_stats.insert(Arc::clone(capture_instance), size);
-            tracing::trace!(%source_id, %capture_instance, %schema, %table, "timely-{worker_id} snapshot stats end");
-        }
+        tracing::trace!(%source_id, %table.capture_instance.name, %table.schema_name, %table.name, "timely-{worker_id} snapshot stats start");
+        let size =
+            crate::inspect::snapshot_size(txn.client, &table.schema_name, &table.name).await?;
+        tracing::trace!(%source_id, %table.capture_instance.name, %table.schema_name, %table.name, "timely-{worker_id} snapshot stats end");
+        let schema_name = &*table.schema_name;
+        let table_name = &*table.name;
+        let rows = async_stream::try_stream! {
+            {
+                let snapshot_stream = crate::inspect::snapshot(txn.client, &*schema_name, &*table_name);
+                tokio::pin!(snapshot_stream);
 
-        // Run a `SELECT` query to snapshot the entire table.
-        let stream = async_stream::stream! {
-            // TODO(sql_server3): A stream of streams would be better here than
-            // returning the name with each result, but the lifetimes are tricky.
-            for (capture_instance, schema_name, table_name) in tables {
-                tracing::trace!(%source_id, %capture_instance, %schema_name, %table_name, "timely-{worker_id} snapshot start");
-
-                let snapshot = crate::inspect::snapshot(txn.client, &*schema_name, &*table_name);
-                let mut snapshot = std::pin::pin!(snapshot);
-                while let Some(result) = snapshot.next().await {
-                    yield (Arc::clone(&capture_instance), result);
+                while let Some(row) = snapshot_stream.next().await {
+                    yield row?;
                 }
-
-                tracing::trace!(%source_id, %capture_instance, %schema_name, %table_name, "timely-{worker_id} snapshot end");
             }
 
-            // Slightly awkward, but if the rollback fails we need to conform to
-            // type of the stream.
-            if let Err(e) = txn.rollback().await {
-                yield ("rollback".into(), Err(e));
-            }
+            txn.rollback().await?
         };
 
-        Ok((lsn, snapshot_stats, stream))
+        Ok((lsn, size, rows))
     }
 
     /// Consume `self` returning a [`Stream`] of [`CdcEvent`]s.
@@ -438,7 +390,7 @@ impl<'a> CdcStream<'a> {
     /// This method runs a retry loop that waits for the upstream DB to report good
     /// values. It should be called before taking the initial [`CdcStream::snapshot`]
     /// to ensure the system is ready to proceed with CDC.
-    async fn wait_for_ready(&mut self) -> Result<(), SqlServerError> {
+    pub async fn wait_for_ready(&mut self) -> Result<(), SqlServerError> {
         // Ensure all of the capture instances are reporting an LSN.
         for instance in self.capture_instances.keys() {
             crate::inspect::get_min_lsn_retry(self.client, instance, self.max_lsn_wait).await?;
--- a/src/sql-server-util/src/inspect.rs
+++ b/src/sql-server-util/src/inspect.rs
@@ -355,12 +355,58 @@ SELECT @mz_cleanup_status_bit;
     }
 }
 
-/// Returns the `(capture_instance, schema_name, table_name)` for the tables
-/// that are tracked by the specified `capture_instance`s.
+// Retrieves all columns in tables that have CDC (Change Data Capture) enabled.
+//
+// Returns metadata needed to create an instance of ['SqlServerTableRaw`].
+//
+// The query joins several system tables:
+// - sys.tables: Source tables in the database
+// - sys.schemas: Schema information for proper table identification
+// - sys.columns: Column definitions including nullability
+// - sys.types: Data type information for each column
+// - cdc.change_tables: CDC configuration linking capture instances to source tables
+// - information_schema views: To identify primary key constraints
+//
+// For each column, it returns:
+// - Table identification (schema_name, table_name, capture_instance)
+// - Column metadata (name, type, nullable, max_length, precision, scale)
+// - Primary key information (constraint name if the column is part of a PK)
+static GET_COLUMNS_FOR_TABLES_WITH_CDC_QUERY: &str = "
+SELECT
+    s.name as schema_name,
+    t.name as table_name,
+    ch.capture_instance as capture_instance,
+    ch.create_date as capture_instance_create_date,
+    c.name as col_name,
+    ty.name as col_type,
+    c.is_nullable as col_nullable,
+    c.max_length as col_max_length,
+    c.precision as col_precision,
+    c.scale as col_scale,
+    tc.constraint_name AS col_primary_key_constraint
+FROM sys.tables t
+JOIN sys.schemas s ON t.schema_id = s.schema_id
+JOIN sys.columns c ON t.object_id = c.object_id
+JOIN sys.types ty ON c.user_type_id = ty.user_type_id
+JOIN cdc.change_tables ch ON t.object_id = ch.source_object_id
+LEFT JOIN information_schema.key_column_usage kc
+    ON kc.table_schema = s.name
+    AND kc.table_name = t.name
+    AND kc.column_name = c.name
+LEFT JOIN information_schema.table_constraints tc
+    ON tc.constraint_catalog = kc.constraint_catalog
+    AND tc.constraint_schema = kc.constraint_schema
+    AND tc.constraint_name = kc.constraint_name
+    AND tc.table_schema = kc.table_schema
+    AND tc.table_name = kc.table_name
+    AND tc.constraint_type = 'PRIMARY KEY'
+";
+
+/// Returns the table metadata for the tables that are tracked by the specified `capture_instance`s.
 pub async fn get_tables_for_capture_instance<'a>(
     client: &mut Client,
     capture_instances: impl IntoIterator<Item = &str>,
-) -> Result<Vec<(Arc<str>, Arc<str>, Arc<str>)>, SqlServerError> {
+) -> Result<Vec<SqlServerTableRaw>, SqlServerError> {
     // SQL Server does not have support for array types, so we need to manually construct
     // the parameterized query.
     let params: SmallVec<[_; 1]> = capture_instances.into_iter().collect();
@@ -383,37 +429,14 @@ pub async fn get_tables_for_capture_instance<'a>(
         .join(", ");
 
     let table_for_capture_instance_query = format!(
-        "
-SELECT c.capture_instance, SCHEMA_NAME(o.schema_id) as schema_name, o.name as obj_name
-FROM sys.objects o
-JOIN cdc.change_tables c
-ON o.object_id = c.source_object_id
-WHERE c.capture_instance IN ({param_indexes});"
+        "{GET_COLUMNS_FOR_TABLES_WITH_CDC_QUERY} WHERE ch.capture_instance IN ({param_indexes});"
     );
 
     let result = client
         .query(&table_for_capture_instance_query, &params_dyn[..])
         .await?;
-    let tables = result
-        .into_iter()
-        .map(|row| {
-            let capture_instance: &str = row.try_get("capture_instance")?.ok_or_else(|| {
-                SqlServerError::ProgrammingError("missing column 'capture_instance'".to_string())
-            })?;
-            let schema_name: &str = row.try_get("schema_name")?.ok_or_else(|| {
-                SqlServerError::ProgrammingError("missing column 'schema_name'".to_string())
-            })?;
-            let table_name: &str = row.try_get("obj_name")?.ok_or_else(|| {
-                SqlServerError::ProgrammingError("missing column 'schema_name'".to_string())
-            })?;
 
-            Ok::<_, SqlServerError>((
-                capture_instance.into(),
-                schema_name.into(),
-                table_name.into(),
-            ))
-        })
-        .collect::<Result<_, _>>()?;
+    let tables = deserialize_table_columns_to_raw_tables(result)?;
 
     Ok(tables)
 }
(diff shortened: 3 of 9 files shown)


src/sql-server-util/src/cdc.rs
```diff
@@ -147,17 +147,27 @@ impl<'a> CdcStream<'a> {
...
+    /// a [`SnapshotEvent::Metadata`] event followed by N [`SnapshotEvent::Row`].
+    ///
+    /// We use the outer stream to ensure there will only ever have two connections open at a time.
+    /// Assuming that the consumer will consume each inner snapshot stream in sequence.
     pub async fn snapshot<'b>(
         &'b mut self,
         instances: Option<BTreeSet<Arc<str>>>,
         worker_id: usize,
         source_id: GlobalId,
     ) -> Result<
```
petrosagg, line 153: I think this PR will be better of if you changed this method signature to handle a single capture instance:

```rust
pub async fn snapshot<'b>(
    &'b mut self,
    instance: &str,
    worker_id: usize,
    source_id: GlobalId,
) -> Result<
    (Lsn, usize, impl Stream<Item = Result<tiberius::Row, SqlServerError> + use<'b, 'a>),
    SqlServerError
>
```

The caller can then organize how they want to call it and which tables to snapshot. The current design also hides some structure from the type system, namely that we expect a single metadata event followed by many data events, which forces you do add invariant checking in a bunch of cases:

```rust
let Some(result) = snapshot.next().await else {
    return Err(TransientError::from(SqlServerError::InvariantViolated(format!("snapshot stream empty for {capture_instance}"))));
};
let SnapshotEvent::Metadata { lsn: snapshot_lsn, size } = result.map_err(TransientError::from)? else {
    return Err(TransientError::from(SqlServerError::InvariantViolated(format!("first element row not metadata in snapshot stream for {capture_instance}"))));
};
```

```rust
let SnapshotEvent::Row(data) = snapshot_event.map_err(TransientError::from)? else {
    return Err(TransientError::from(SqlServerError::InvariantViolated(format!("Metadata found after first element of stream in snapshot stream for {capture_instance}"))));
};
```

None of these are needed if you work with a function signature that deals with a single instance.

> We use the outer stream to ensure there will only ever have two connections open at a time.

If we want to have this optimization we might want to add a parameter with the additional clients that this method needs or create a intermediate struct that is responsible for snapshots and creates the two clients and then we call `.snapshot(..)` on that.

ptravers: sure thing.

ptravers: would you object if we did table discovery outside the snapshot function? it would be helpful for schema checking in [#9573](https://github.com/MaterializeInc/database-issues/issues/9573)

ptravers: put up a possible solution following that pattern. in running local tests there may be a bug in it. so running tests on it in CI to try to ensure this isn't a local thing.

src/sql-server-util/Cargo.toml
```diff
@@ -9,9 +9,6 @@ publish = false
 [lints]
 workspace = true
 
-[[example]]
-name = "cdc"
-
```
ptravers: not sure how people feel about just removing this code.

cc: @martykulma @petrosagg

src/sql-server-util/src/desc.proto
```diff
@@ -20,6 +20,7 @@ message ProtoSqlServerTableDesc {
   string name = 1;
   string schema_name = 2;
   repeated ProtoSqlServerColumnDesc columns = 3;
+  string capture_instance = 5;
```
ptravers: do we care that this is a breaking change? is it acceptable to tell people using the private preview to drop and recreate? do we want to make this optional and add handling logic?

src/storage/src/source/sql_server/replication.rs
```diff
@@ -141,75 +149,88 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
+                        if last_report.elapsed() > report_interval.get() {
+                            last_report = Instant::now();
+                            emit_stats(&stats_cap[0], records_known, records_total);
+                        }
+
+                        // Decode the SQL Server row into an MZ one.
+                        let mut mz_row = Row::default();
+                        let arena = RowArena::default();
+
+                        let partition_indexes = export_ids_to_snapshot.get(&table.capture_instance)
```
ptravers: found the bug. the capture instance I'm pulling out when loading the tables doesn't match the capture instance we get from the outputs. bit surprising.

src/storage/src/source/sql_server/replication.rs
```diff
@@ -219,7 +242,7 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
                 );
                 emit_stats(&stats_cap[0], records_known, records_total);
 
-                snapshot_lsn
+                max_lsn
```
martykulma: If we only return a single LSN here, is that still valid for all tables?  If there are multiple tables, it seems this would result in a negative accumulation when rewinds are generated.

src/sql-server-util/src/desc.rs
```diff
@@ -59,6 +59,8 @@ pub struct SqlServerTableDesc {
     pub name: Arc<str>,
     /// Columns for the table.
     pub columns: Box<[SqlServerColumnDesc]>,
+    /// The capture instance for used for CDC for this table.
+    pub capture_instance: Arc<str>,
```
martykulma: Would it make more sense to just compose a new struct to hold this rather than leak it into the table description?

If you feel strongly that we should keep the capture_instance in here, then we might want to remove it from the structs that reference both a capture instance and the table description, e.g. `PurifiedExportDetails`, `ReferenceMetadata`, `SourceExportStatementDetails`, etc.

src/storage/src/source/sql_server/replication.rs
```diff
@@ -141,75 +151,88 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
+                                    Err(DataflowError::DecodeError(Box::new(DecodeError {
+                                        kind,
+                                        raw: raw.as_bytes().to_vec(),
+                                    })))
+                                }
+                            };
+                            data_output
+                                .give_fueled(
+                                    &snapshot_cap,
+                                    ((*partition_idx, message), snapshot_lsn, Diff::ONE),
```
martykulma: this is a bug that got re-introduced after https://github.com/MaterializeInc/materialize/pull/32979


these should be emitted at `Lsn::minimum()`!

ptravers: !!! will fix!

src/sql-server-util/src/desc.rs
ptravers: I could also just return a tuple with the capture instance and table description? @petrosagg was a fan of adding it to the table desc.

src/storage/src/source/sql_server/replication.rs
ptravers: discussed over zoom. this is correct. pushed a fix. still working on a test.

test/sql-server-cdc/70-lsn-updated-during-snapshot.td
```diff
@@ -0,0 +1,40 @@
...
+$ sql-server-connect name=sql-server
+server=tcp:sql-server,1433;IntegratedSecurity=true;TrustServerCertificate=true;User ID=${arg.default-sql-server-user};Password=${arg.default-sql-server-password};Database=test
+
+$ sql-server-execute name=sql-server
+CREATE TABLE t1_pk (key_col VARCHAR(20) PRIMARY KEY, val_col VARCHAR(1024));
+EXEC sys.sp_cdc_enable_table @source_schema = 'dbo', @source_name = 't1_pk', @role_name = 'SA', @supports_net_changes = 0;
+
+INSERT INTO t1_pk VALUES ('churro', 'peanut'), ('friskies extra gravy', 'minnie'), ('cereal milk', 'mischa'), ('whiskers', 'archie');
+
+$ sql-server-execute name=sql-server split-lines=false
```
ptravers, line 33: accidentally pushed, fixing now.

src/storage/src/source/sql_server/replication.rs
ptravers: fixed!

src/sql-server-util/src/cdc.rs
```diff
@@ -241,42 +217,27 @@ impl<'a> CdcStream<'a> {
...
-            // returning the name with each result, but the lifetimes are tricky.
-            for (capture_instance, schema_name, table_name) in tables {
-                tracing::trace!(%source_id, %capture_instance, %schema_name, %table_name, "timely-{worker_id} snapshot start");
-
-                let snapshot = crate::inspect::snapshot(txn.client, &*schema_name, &*table_name);
-                let mut snapshot = std::pin::pin!(snapshot);
-                while let Some(result) = snapshot.next().await {
-                    yield (Arc::clone(&capture_instance), result);
+        tracing::trace!(%source_id, %table.capture_instance, %table.schema_name, %table.name, "timely-{worker_id} snapshot stats start");
+        // Establish the consistent point-in-time and release the lock.
```
petrosagg: This comment doesn't match the code below. The consistent point and lock release have already happened above.

src/sql-server-util/src/cdc.rs
```diff
@@ -241,42 +217,27 @@ impl<'a> CdcStream<'a> {
...
+        let size =
+            crate::inspect::snapshot_size(txn.client, &table.schema_name, &table.name).await?;
```
petrosagg, line 214: Not for this PR, but we have had already issues with metadata queries being too expensive. Similarly to pg source we should put the `SELECT COUNT(*)` approach for getting the snapshot size behind a feature flag and if there is a way to get the approximate size from statistics, prefer it. Here is what the logic looks like in pg: https://github.com/MaterializeInc/materialize/blob/main/src/storage/src/source/postgres/snapshot.rs#L750-L763

src/sql-server-util/src/desc.proto
petrosagg: We can ask whoever is testing this to recreate their sources

src/sql-server-util/src/desc.rs
petrosagg: The thing I was a fan of was storing the capture instance in the source metadata instead of discovering it. It doesn't have to be in the table desc struct specifically, but it doesn't seem out of place either. Deferring to Marty

Regarding storing vs discovering the capture instance, what is the standard operating practice with those in SQL Server? Are users expected to create one and never touch it or do they have legitimate reasons to recreate them? If it's the latter case then maybe we should be discovering the capture instance of a given table and ensuring that it has the LSNs we need.

src/sql-server-util/src/inspect.rs
```diff
@@ -354,12 +354,42 @@ SELECT @mz_cleanup_status_bit;
     }
 }
 
+static GET_TABLES_QUERY: &str = "
```
petrosagg: Let's add some documentation on what information this query tries to fetch

src/sql-server-util/src/inspect.rs
```diff
@@ -444,35 +452,16 @@ pub async fn ensure_snapshot_isolation_enabled(client: &mut Client) -> Result<()
...
-";
+    let result = client.simple_query(&format!("{GET_TABLES_QUERY};")).await?;
+
+    let tables = deserialize_to_raw_tables(result)?;
+
+    Ok(tables)
+}
+
+fn deserialize_to_raw_tables(
+    rows: SmallVec<[tiberius::Row; 1]>,
```
petrosagg: This method doesn't benefit anymore from receiving a list of rows since the batching happens elsewhere. I suggest you move this to an inherent `SqlServerTableRaw::from_row` constructor and call it from the other places appropriately.

src/sql-server-util/src/inspect.rs
```diff
@@ -444,35 +452,16 @@ pub async fn ensure_snapshot_isolation_enabled(client: &mut Client) -> Result<()
...
-    ON tc.constraint_catalog = kc.constraint_catalog
-    AND tc.constraint_schema = kc.constraint_schema
-    AND tc.constraint_name = kc.constraint_name
-    AND tc.table_schema = kc.table_schema
-    AND tc.table_name = kc.table_name
-    AND tc.constraint_type = 'PRIMARY KEY';
-";
+    let result = client.simple_query(&format!("{GET_TABLES_QUERY};")).await?;
+
+    let tables = deserialize_to_raw_tables(result)?;
```
petrosagg: See the other comment for the `SqlServerTableRaw::from_row` suggestion

```suggestion
    result.iter().map(SqlServerTableRaw::from_row).collect()
```

src/storage/src/source/sql_server/replication.rs
```diff
@@ -100,12 +101,14 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
                 .await?;
             let mut client = mz_sql_server_util::Client::connect(connection_config).await?;
 
+            let worker_id = config.worker_id;
+
             // The decoder is specific to the export, and each export pulls data from a specific capture instance.
             let mut decoder_map: BTreeMap<_, _> = BTreeMap::new();
             // Maps the 'capture instance' to the output index for only those outputs that this worker will snapshot
-            let mut export_ids_to_snapshot: BTreeMap<_, Vec<_>> = BTreeMap::new();
+            let mut export_ids_to_snapshot: BTreeMap<Arc<str>, Vec<_>> = BTreeMap::new();
```
petrosagg: Something is off here. `export_id` usually refers to the `GlobalId` of a particular export but here it seems to be referring to a capture instance?

src/storage/src/source/sql_server/replication.rs
```diff
@@ -124,13 +127,20 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
                 }
             }
 
+            // TODO(sql_server3): refactor to pass table description as output of upstream.
```
petrosagg: I didn't understand what this means. What is the output of upstream?

src/storage/src/source/sql_server/replication.rs
```diff
@@ -141,75 +151,84 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
                     stats_output.give(cap, update);
                 };
 
-                let capture_instances_to_snapshot: BTreeSet<_> = export_ids_to_snapshot.keys().cloned().collect();
-                tracing::debug!(?config.id, ?capture_instances_to_snapshot, "starting snapshot");
                 // Eagerly emit an event if we have tables to snapshot.
-                if !capture_instances_to_snapshot.is_empty() {
+                if tables.len() > 0 {
```
petrosagg: nit: using `is_empty()` is a bit clearer. Normally there is a clippy lint for this https://rust-lang.github.io/rust-clippy/master/index.html#/is_empty but we don't have it enabled for the materialize repo

src/storage/src/source/sql_server/replication.rs
```diff
@@ -141,75 +151,84 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
-                                })))
-                            }
-                        };
-                        data_output
-                            .give_fueled(
-                                &snapshot_cap,
-                                ((*partition_idx, message), snapshot_lsn, Diff::ONE),
-                            )
-                            .await;
+                    snapshot_lsns.insert(Arc::clone(&table.capture_instance), Arc::new(snapshot_lsn));
```
petrosagg: `snapshot_lsn` is `Copy` and tiny so not worth putting it on the heap.

```suggestion
                    snapshot_lsns.insert(Arc::clone(&table.capture_instance), snapshot_lsn);
```

src/storage/src/source/sql_server/replication.rs
```diff
@@ -141,75 +151,84 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
+                        records_total = records_total.saturating_add(1);
+
+                        if last_report.elapsed() > report_interval.get() {
+                            last_report = Instant::now();
+                            emit_stats(&stats_cap[0], records_known, records_total);
+                        }
+
+                        // Decode the SQL Server row into an MZ one.
+                        let mut mz_row = Row::default();
+                        let arena = RowArena::default();
```
petrosagg: Let's move this arena outside both loops (i.e before the `for table in tables`) to reuse it for everything

src/storage/src/source/sql_server/replication.rs
```diff
@@ -141,75 +151,84 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
+
+                        let partition_indexes = export_ids_to_snapshot.get(&table.capture_instance)
+                            .ok_or_else(|| {
+                                let msg = format!("no snapshot outputs in known capture instances [{}] for capture instance: '{}'", export_ids_to_snapshot.keys().join(","), table.capture_instance);
+                                TransientError::ProgrammingError(msg)
+                            })?;
```
petrosagg: It's fine to just panic for these kind of invariants

```suggestion
                        let partition_indexes = export_ids_to_snapshot[&table.capture_instance];
```

src/storage/src/source/sql_server/replication.rs
```diff
@@ -141,75 +151,84 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
+
+                        for (partition_idx, _) in partition_indexes {
+                            let decoder = decoder_map.get(partition_idx).expect("decoder for output");
+                            // Try to decode a row, returning a SourceError if it fails.
```
petrosagg, line 230: Let's move the definition of the `mz_row` in here so that we can put it in `SourceMessage` without cloning. It doesn't get reused so we don't need to be paying the cost of copying the data.

```suggestion
                            // Try to decode a row, returning a SourceError if it fails.
                            let mut mz_row = Row::default();
```

src/storage/src/source/sql_server/replication.rs
```diff
@@ -237,12 +256,13 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
-                .values()
-                .flat_map(|export_ids|
+                .iter()
+                .flat_map(|(capture_instance, export_ids)|{
+                    let snapshot_lsn = snapshot_lsns.get(capture_instance).expect("snapshot lsn must be collected for capture instance");
                     export_ids
                         .iter()
-                        .map(|(idx, initial_lsn)| (*idx, (*initial_lsn, snapshot_lsn)))
-                ).collect();
+                        .map(|(idx, initial_lsn)| (*idx, (*initial_lsn, Arc::clone(snapshot_lsn))))
```
petrosagg: same here
```suggestion
                        .map(|(idx, initial_lsn)| (*idx, (*initial_lsn, snapshot_lsn)))
```

test/sql-server-cdc/70-lsn-updated-during-snapshot.td
```diff
@@ -0,0 +1,86 @@
...
+# Test that if we snapshot while writes are happening we will not lose data.
+# We want to guarantee that there were writes to t2 while t0 and t1 are being
+# snapshotted so we write 3 new rows every 10ms to t2.
+# We then halt the job writing rows to t2. Count the number of rows in SQL Server
+# then assert that the final number of rows in t2 are the same.
+#
+# Thus we verify that t2 picks the correct LSN to rewind from after snapshotting.
+#
+# Note: we chose 10ms as that was found to be the lowest interval that SQL Server would
+# reliably execute a job.
```
petrosagg: Is this test validating something more than `workflow_snapshot_consistency`?

src/sql-server-util/src/cdc.rs
ptravers: created issue for post ga https://github.com/MaterializeInc/database-issues/issues/9602

src/sql-server-util/src/cdc.rs
ptravers: will fix

src/sql-server-util/src/inspect.rs
ptravers: can do

src/sql-server-util/src/inspect.rs
ptravers: I'm not sure how that would work since one table would be derived from multiple rows. I believe if I have understood the query correctly we would get a row per column.

src/storage/src/source/sql_server/replication.rs
ptravers: I was wondering if we could pass the table description in the outputs

src/storage/src/source/sql_server/replication.rs
ptravers: sure

src/storage/src/source/sql_server/replication.rs
ptravers: sure, thing

src/storage/src/source/sql_server/replication.rs
ptravers: will do

test/sql-server-cdc/70-lsn-updated-during-snapshot.td
ptravers: yes, `workflow_snapshot_consistency` validates for one table. here we are trying to force the lsn to be incremented while t3 and t4 are being snapshotted. I updated the comment and to make it clearer and to actually test what I set out to test yesterday. we are forcing the snapshot_lsn and rewinds to be selected per table rather than for all tables.

src/storage/src/source/sql_server/replication.rs
ptravers: will fix

src/storage/src/source/sql_server/replication.rs
ptravers: sure thing

src/storage/src/source/sql_server/replication.rs
ptravers: is there any value in having a clearer error message here? I did find it helpful when debugging it to see the list of capture instances with the capture instance we were trying to retrieve.

src/sql-server-util/src/inspect.rs
ptravers: same comment as above.

src/sql-server-util/src/desc.rs
ptravers: sorry. to misrepresent you! the arguments I can see for discovering might be that we are more resilient to end user chaos. we did remove the cdc cleanup feature. customers might want to drop and recreate the cdc job periodically to decrease the size of the table? if there's a very high write workload on the table there may be a ton of LSNs?

I have no insights here. I will ask product/field engineering to see if we can get some insight from a SQL Server DBA.

test/sql-server-cdc/70-lsn-updated-during-snapshot.td
ptravers: but I'm also happy to reimplement into that workflow using python to be more consistent.

src/sql-server-util/src/desc.rs
martykulma: some overlap with https://github.com/MaterializeInc/database-issues/issues/9206?   Seems like there is a potential use case for customer doing blue-green where one MZ table reads from capture instance A, and another from capture instance B.

ptravers: yeah, I wondered about that too.

src/storage/src/source/sql_server/replication.rs
ptravers: do you mind if I file a follow up to fix that? I feel like this PR is starting to snowball.

src/sql-server-util/src/inspect.rs
ptravers: I documented and renamed the query lmk if that makes more sense.

src/storage/src/source/sql_server/replication.rs
ptravers: done

src/storage/src/source/sql_server/replication.rs
ptravers: done

src/storage/src/source/sql_server/replication.rs
ptravers: done

src/storage/src/source/sql_server/replication.rs
ptravers: done

src/storage/src/source/sql_server/replication.rs
ptravers: done

ptravers:
interestingly, `correctness-property-2.td` passes locally. I will pull that out of this PR.

src/storage/src/source/sql_server/replication.rs
ptravers: issue: https://github.com/MaterializeInc/database-issues/issues/9603

src/sql-server-util/src/inspect.rs
ptravers: I updated the explanation and name of the query to make it clearer that the result set is a row per column not table.

src/storage/src/source/sql_server/replication.rs
ptravers: removed.

ptravers:
nightly run of resumption tests for sql server: https://buildkite.com/materialize/nightly/builds/12903

ptravers:
nightly run of resumption tests for sql server: https://buildkite.com/materialize/nightly/builds/12906

ptravers:
nightly run of resumption and old syntax tests for sql server: https://buildkite.com/materialize/nightly/builds/12908

src/sql-server-util/src/desc.rs
ptravers: replaced with `SqlServerTableRaw` following slack conversation.

src/sql-server-util/src/inspect.rs
ptravers: also refactored to `deserialize_table_columns_to_raw_tables`.

test/sql-server-cdc/70-lsn-updated-during-snapshot.td
```diff
@@ -0,0 +1,113 @@
...
+  FOR ALL TABLES;
+
+$ sql-server-execute name=sql-server split-lines=false
+USE msdb;
+EXEC msdb.dbo.sp_stop_job N'FastTicker';
+# Wait for 5 seconds to ensure the job has stopped and that we stopped writing to t2.
+# We want to ensure that the count we take below is after the job has stopped to guarantee
+# that replication hasn't lost rows but then appeared to catch up due to new rows being
+# written.
+WAITFOR DELAY '00:00:05';
```
petrosagg: Can we use the testdrive sleep command here so that we can easily grep all the places in the tests where we sleep?

src/sql-server-util/Cargo.toml
petrosagg: What's the motivation behind the deletion? It's there, and shows a simple way to use the lib, does it hurt to keep it?

src/sql-server-util/src/inspect.rs
petrosagg: I see, makes sense

src/storage/src/source/sql_server/replication.rs
petrosagg: I don't think we need to refactor anything, just give it a better name to reflect its contents. Something like `capture_instance_outputs`

src/storage/src/source/sql_server/replication.rs
petrosagg: You can use `.expect()` to give a message. My main point is that `TransientError` is used for things like network errors or other errors where restarting the dataflow may resolve them. In this case it's simply a bug if the capture instance doesn't exist in the map and we want to panic so that it also shows up in Sentry and we look into it asap

src/sql-server-util/src/cdc.rs
petrosagg: Thank you!

test/sql-server-cdc/70-lsn-updated-during-snapshot.td
ptravers: ah yeah, sure

src/storage/src/source/sql_server/replication.rs
ptravers: ah I see, thanks. that makes a lot of sense.

src/storage/src/source/sql_server/replication.rs
ptravers: ah I see, sure. will rename.

src/storage/src/source/sql_server/replication.rs
ptravers: I refactored to unwrap_or_else with a panic so I could create an error message with params. hope that's okay?

src/sql-server-util/Cargo.toml
ptravers: I wasn't sure if anyone was using or found it helpful. I re-added it since you have found it helpful. I had assumed it was an artifact of the original spike and thereby might just be dead code we're maintaining.

src/storage/src/source/sql_server/replication.rs
ptravers: done, tried some other names but I think that makes the most sense. we do oscillate between calling the values export ids and partition ids. not sure what's more idiomatic/greppable.

test/sql-server-cdc/70-lsn-updated-during-snapshot.td
ptravers: done.

src/sql-server-util/src/cdc.rs
```diff
@@ -143,39 +144,21 @@ impl<'a> CdcStream<'a> {
...
-                None => true,
-            })
-            .map(|i| i.as_ref());
-        let tables =
-            crate::inspect::get_tables_for_capture_instance(self.client, instances).await?;
-        tracing::info!(%source_id, ?tables, "timely-{worker_id} got table for capture instance");
-
         // Before starting a transaction where the LSN will not advance, ensure
         // the upstream DB is ready for CDC.
         self.wait_for_ready().await?;
```
martykulma: as a follow on, we can move `wait_for_ready` out of this function since it really only needs to be run before the first snapshot.

src/sql-server-util/src/cdc.rs
```diff
@@ -223,60 +199,44 @@ impl<'a> CdcStream<'a> {
...
-                let snapshot = crate::inspect::snapshot(txn.client, &*schema_name, &*table_name);
-                let mut snapshot = std::pin::pin!(snapshot);
-                while let Some(result) = snapshot.next().await {
-                    yield (Arc::clone(&capture_instance), result);
+        tracing::trace!(%source_id, %table.capture_instance, %table.schema_name, %table.name, "timely-{worker_id} snapshot stats start");
+        let size =
+            crate::inspect::snapshot_size(txn.client, &table.schema_name, &table.name).await?;
+        tracing::trace!(%source_id, %table.capture_instance, %table.schema_name, %table.name, "timely-{worker_id} snapshot stats end");
+        let schema_name = Arc::clone(&table.schema_name);
+        let table_name = Arc::clone(&table.name);
```
martykulma: nit: shouldn't need to clone here, can just use `&*table.schema_name` and `&*table.name`

src/storage/src/source/sql_server/replication.rs
```diff
@@ -100,12 +101,14 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
                 .await?;
             let mut client = mz_sql_server_util::Client::connect(connection_config).await?;
 
+            let worker_id = config.worker_id;
+
             // The decoder is specific to the export, and each export pulls data from a specific capture instance.
             let mut decoder_map: BTreeMap<_, _> = BTreeMap::new();
             // Maps the 'capture instance' to the output index for only those outputs that this worker will snapshot
-            let mut export_ids_to_snapshot: BTreeMap<_, Vec<_>> = BTreeMap::new();
+            let mut capture_instance_to_outputs: BTreeMap<Arc<str>, Vec<_>> = BTreeMap::new();
```
martykulma: hilariously, this list and the list below it are both mappings of capture instances to outputs.  My vote would be to leave `snapshot` in the name just because we have the snapshot code intermingled with the steady state replication code.

src/storage/src/source/sql_server/replication.rs
```diff
@@ -117,20 +120,26 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
                         .or_default()
                         .push((output.partition_index, output.initial_lsn));
                 }
             }
 
+            let instances = capture_instance_to_outputs
+                    .keys()
+                    .map(|i| i.as_ref());
+
+            let tables = mz_sql_server_util::inspect::get_tables_for_capture_instance(&mut client, instances).await?;
```
martykulma: ditto for these variables

src/storage/src/source/sql_server/replication.rs
```diff
@@ -141,75 +150,83 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
+
+                        if last_report.elapsed() > report_interval.get() {
+                            last_report = Instant::now();
+                            emit_stats(&stats_cap[0], records_known, records_total);
+                        }
+
+                        let partition_indexes = capture_instance_to_outputs.get(&table.capture_instance)
+                            .unwrap_or_else(|| {
+                                panic!("no snapshot outputs in known capture instances [{}] for capture instance: '{}'", capture_instance_to_outputs.keys().join(","), table.capture_instance);
+                            });
```
martykulma: this lookup can now be done outside the `snapshot.next().await` loop!

src/sql-server-util/src/cdc.rs
ptravers: on it

src/sql-server-util/src/cdc.rs
ptravers: good catch

src/storage/src/source/sql_server/replication.rs
ptravers: can do!

src/storage/src/source/sql_server/replication.rs
ptravers: will fix

src/storage/src/source/sql_server/replication.rs
ptravers: fair point, will fix.

src/sql-server-util/src/cdc.rs
ptravers: had to make `wait_for_ready` `pub` hope that's okay?

src/sql-server-util/examples/cdc.rs
```diff
@@ -40,6 +40,8 @@ use mz_ore::future::InTask;
 use mz_sql_server_util::cdc::CdcEvent;
 use mz_sql_server_util::config::TunnelConfig;
 use mz_sql_server_util::{Client, Config};
+use std::collections::BTreeMap;
+use std::sync::Arc;
```
petrosagg: Imports from `std` go in a separate import block above the crate imports. Can you move these two upwards?

src/sql-server-util/src/inspect.rs
```diff
@@ -445,36 +468,18 @@ pub async fn ensure_snapshot_isolation_enabled(client: &mut Client) -> Result<()
...
+        .simple_query(&format!("{GET_COLUMNS_FOR_TABLES_WITH_CDC_QUERY};"))
+        .await?;
+
+    let tables = deserialize_table_columns_to_raw_tables(result)?;
+
+    Ok(tables)
+}
+
+fn deserialize_table_columns_to_raw_tables(
+    rows: SmallVec<[tiberius::Row; 1]>,
```
petrosagg: This function doesn't need ownership of the rows nor does it need to take a `SmallVec` specifically. Switching to a slice is more idiomatic

```suggestion
    rows: &[tiberius::Row],
```

src/storage/src/source/sql_server.rs
```diff
@@ -74,8 +74,6 @@ pub enum TransientError {
     ReplicationEOF,
     #[error(transparent)]
     SqlServer(#[from] SqlServerError),
-    #[error("programming error: {0}")]
-    ProgrammingError(String),
```
petrosagg, line 78: nice!

src/storage/src/source/sql_server/replication.rs
```diff
@@ -140,76 +149,92 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
...
+                                    let raw = format!("{sql_server_row:?}");
+                                    Err(DataflowError::DecodeError(Box::new(DecodeError {
+                                        kind,
+                                        raw: raw.as_bytes().to_vec(),
+                                    })))
+                                }
+                            };
+                            data_output
+                                .give_fueled(
+                                    &snapshot_cap,
```
petrosagg: This capability is created for `snapshot_lsn` but is used to emit data at `Lsn::minimum()`. Differential Dataflow requires each update triplet is emitted with a capability that is less than or equal to the time field of the triplet. So here you either need to make snapshot_cap be a capability for `Lsn::minimum()` or use `&data_cap_set[0]` which has not been downgraded yet and therefore is already at the minimum timestamp.

```suggestion
                                    &data_cap_set[0],
```

src/sql-server-util/src/inspect.rs
ptravers: gotcha, will fix!

src/sql-server-util/examples/cdc.rs
ptravers: will do!

src/storage/src/source/sql_server/replication.rs
ptravers: I meant to leave a comment asking. why did this work?

petrosagg: You mean why it doesn't explode if it's an expectation? Operators don't validate that this property holds but may depend on it. It can be that either the operators we're using are not depending on it, or they do and the condition for a wrong result is subtle and haven't hit it in CI.

PR #33350, opened 2025-08-17
Docs: clean up manage rbac section + add guidance  (1st round)

kay-kim:
Round one covers the granting privileges/default privileges/changing ownership/ + grant privilege syntax.  If we think the content works for us, then can continue on with the revokes and the sql ref pages.


doc/user/content/manage/access-control/manage-roles.md
```diff
@@ -1,69 +1,661 @@
...
+{{< /annotation >}}
+
+## Required privileges for managing roles
+
+{{< note >}}
+
+With their **superuser** privileges, [**Organization
+admins**](/manage/users-service-accounts/#organization-roles) can manage roles
+(including overriding ownership requirements when granting privileges on various
+objects).
```
kay-kim, line 33: Was of two minds about just having a blanket statement of An Org admin can manage roles ... and assume that all commands (except where explicitly stated) are run as the admin to get around ownership/hierarchical usage permissions ... because it does seem that it is most straightforward in practice.

doc/user/content/manage/access-control/manage-roles.md
```diff
@@ -1,69 +1,661 @@
...
+{{< include-md file="shared-content/rbac/role-name-restrictions.md" >}}
+{{</ tip >}}
+
+For example, the following creates:
+- A role for users who need to perform compute/transform operations in the
+  compute/transform.
+- A role for users who need to manage indexes on the serving cluster(s).
+- A role for users who need to read results from the serving cluster.
+
+{{< tabs >}}
```
kay-kim, line 80: I thought about having a source creator role here as well ... but, as with the above statement about role management being just done by the org admin, wasn't sure if in practice we do just only want org admins to create sources.

doc/user/content/sql/grant-privilege.md
```diff
@@ -1,100 +1,258 @@
...
+{{</ tabs >}}
+
 ## Privileges
 
 The privileges required to execute this statement are:
 
 {{< include-md file="shared-content/sql-command-privileges/grant-privilege.md"
 >}}
 
+## Examples
```
kay-kim, line 238: I just move the existing examples down here.
If we're okay with the examples in the manage access page, can single source and include here instead.

doc/user/content/sql/grant-privilege.md
```diff
@@ -1,100 +1,258 @@
 ---
 title: "GRANT PRIVILEGE"
-description: "`GRANT` grants privileges on a database object."
+description: "Grant privileges on a database object."
```
kay-kim, line 3: https://preview.materialize.com/materialize/33350/sql/grant-privilege/#syntax

PR #33358, opened 2025-08-18
collection_mgmt: hold lock for while sending writes

teskje:
Previously, the `update_tx.send` calls could fail because we were not holding the `{differential,append_only}_collections` lock anymore when performing them, making it possible for the writer task to be shut down in the time between the sender acquired the channel and the send. This PR avoids this by making sure the corresponding lock is held across each `send` call.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9590

### Checklist

### DIFF
--- a/src/storage-controller/src/collection_mgmt.rs
+++ b/src/storage-controller/src/collection_mgmt.rs
@@ -394,10 +394,17 @@ where
             panic!("attempting blind write to {} while in read-only mode", id);
         }
 
-        if !updates.is_empty() {
-            let update_tx = self.append_only_write_sender(id);
-            let (tx, _rx) = oneshot::channel();
-            update_tx.send((updates, tx)).expect("rx hung up");
+        if updates.is_empty() {
+            return;
+        }
+
+        let collections = self.append_only_collections.lock().expect("poisoned");
+        match collections.get(&id) {
+            Some((update_tx, _, _)) => {
+                let (tx, _rx) = oneshot::channel();
+                update_tx.send((updates, tx)).expect("rx hung up");
+            }
+            None => panic!("missing append-only collection: {id}"),
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/storage-controller/src/collection_mgmt.rs
```diff
@@ -412,7 +414,9 @@ where
         if !op.is_empty_append() {
             let update_tx = self.differential_write_sender(id);
             let (tx, _rx) = oneshot::channel();
-            update_tx.send((op, tx)).expect("rx hung up");
+
+            // Sending can fail if the writer task has been shut down concurrently.
```
bkirwi: Is the above comment still valid?
`# Panics [...] If the collection closed.`

I'm not actually sure what "collection closed" means - at a glance I can't find it documented - but I suspect it's referring to the panic that you've just removed.

teskje: Ah, I think you are right. "Collection closed" probably means that its upper advanced to the empty frontier, which is a second reason (besides receiving a shutdown signal) why the writer task would stop. Although if I'm reading things correctly, it doesn't stop when making the append to the empty frontier, but only on the next write attempt when it finds that it can't make a new append anymore. So the method didn't always panic when writing to an empty collection.

So yeah, we don't panic anymore if the target collection is closed. This does reduce our ability to find bugs, so I wonder if we should instead change this code to keep the lock until we've completed the `send`? We'd have to inline `differential_write_sender`, causing a bit of code duplication, but it's not too bad.

aljoscha: Couldn't we instead differentiate between:
- task went away
- can't write because collection is "closed", that is it has an empty upper

teskje: I'm not sure how. Both will look like a send failure from this side of the channel. Are you suggesting we add a new channel to communicate this information?

teskje: Changed to instead hold the locks across the channel send calls.

teskje:
TFTRs!

PR #33359, opened 2025-08-18
docs: clarifies that sink schema/clusters shouldn't be included in de…

kay-kim:
…ployment config for b/g


kay-kim:
closing --- was just to test.

PR #33361, opened 2025-08-18
Fix ROWS FROM + WITH ORDINALITY with many columns

ggevay:
Fixes https://github.com/MaterializeInc/database-issues/issues/9592

There was a column indexing bug in `plan_rows_from`'s `with_ordinality` case: `scope.items.len()` refers to the last column _after_ the manipulations in the preceding `for` loop, but actually, `columns` should have indexes that point to the original column indexes from _before_ the `for` loop's manipulations.

The bug was not caught before because the original code happens to work if we happen to point to an ordinality column that has enough elements, which happened in all the earlier tests. Also note that the bug can even cause panics if the type of the wrongly pointed column is not the same as expected for a `WITH ORDINALITY` column.

The first commit adds relevant tests with wrong results (plus a parser test for the empty `ROWS FROM` case), and the second commit fixes the bug and rewrites the tests to the correct results and adds some more tests.

### Motivation

  * This PR fixes a recognized bug: https://github.com/MaterializeInc/database-issues/issues/9592

### Checklist

### DIFF
--- a/src/sql/src/plan/query.rs
+++ b/src/sql/src/plan/query.rs
@@ -3096,7 +3096,7 @@ fn plan_rows_from(
     // If `WITH ORDINALITY` was specified, include the coalesced ordinality
     // column. Otherwise remove it from the scope.
     if with_ordinality {
-        columns.push(scope.items.len());
+        columns.push(offset);
     } else {
         scope.items.pop();
     }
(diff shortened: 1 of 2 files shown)


petrosagg (approved):

ggevay:
Thank you!

ggevay:
I hope it will be an easy backport, as it's a 1-liner code change. A self-managed user wants to use (implicit) `ROWS FROM`: https://materializeinc.slack.com/archives/C07PN7KSB0T/p1756133216915359?thread_ts=1756129863.315519&cid=C07PN7KSB0T

PR #33366, opened 2025-08-19
sql: remove replica disk feature flags

teskje:
This PR removes the feature flags controlling whether replicas get a disk assigned:
* enable_disk_cluster_replicas
* disk_cluster_replicas_default
* cluster_always_use_disk

These feature flags exist because we wanted to roll out disk slowly, but today there is no reason to not assign a disk to every replica.

### Checklist


test/cloudtest/test_compute.py
```diff
@@ -126,16 +126,9 @@ def get_node_selector(
...
-    mz.environmentd.sql(
-        "ALTER SYSTEM SET cluster_always_use_disk = false;",
-        port="internal",
-        user="mz_system",
-    )
-
     for value in ("true", "false"):
         mz.environmentd.sql(
-            f"CREATE CLUSTER disk_{value} MANAGED, SIZE = 'scale=1,workers=2,nodisk', DISK = {value}"
+            f"CREATE CLUSTER disk_{value} MANAGED, SIZE = 'scale=1,workers=2,legacy', DISK = {value}"
```
ggevay, line 131: How is this test passing? Isn't the `else` branch of the below `if value == "true":` intended to check whether we indeed don't have a disk if `disk` is set to false? But with the PR, we should be always getting a disk.

teskje: No, we still don't get a disk if `DISK = false` is explicitly set. It's only the next PR that makes that a no-op (and also changes this test).

ggevay: Ah, yes, ok!

ggevay (approved):

teskje:
TFTR!

ggevay:
One more thing: After the PR is rolled out to all cloud users, what do we do with these flags in LD? Should we "archive" them?

teskje:
Yeah, the process is that after they haven't been queried by any environment for more than 7d, we can archive and delete them. We'll need to wait until this change has rolled out to all envs, which can take a while, especially for staging envs. But there is no need to rush either.

PR #33369, opened 2025-08-19
Persist sink: only evaluate validate_part_bounds_on_write once

DAlperin:
This solves (I think) a rare race condition during parallel workloads. If the FlipFlags action changes persist flags in between when we decide to use bulk writing and when we decide to compare the batch a spurious InvalidBatchBounds error could be emitted.

Instead of reevaluating the flag at each step, we evaluate it once, store it in a variable and use it for the lifetime of the append_batches invocation.

Fixes: https://github.com/MaterializeInc/database-issues/issues/9585

### Checklist

### DIFF
--- a/src/storage/src/render/persist_sink.rs
+++ b/src/storage/src/render/persist_sink.rs
@@ -1259,10 +1259,7 @@ where
                                 &mut to_append[..],
                                 batch_lower.clone(),
                                 batch_upper.clone(),
-                                //NB(ptravers): we get the current state of the flag to fail
-                                // fast if we re-enable bounds checks before a batch is written.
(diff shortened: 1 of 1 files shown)


src/storage/src/render/persist_sink.rs
```diff
@@ -1259,10 +1259,7 @@ where
                                 &mut to_append[..],
                                 batch_lower.clone(),
                                 batch_upper.clone(),
-                                //NB(ptravers): we get the current state of the flag to fail
-                                // fast if we re-enable bounds checks before a batch is written.
-                                // We will see an InvalidUsage error and the sink will restart(?).
-                                write.validate_part_bounds_on_write(),
```
ptravers, line 1265: 🤦

PR #33374, opened 2025-08-19
Add temporal filter to mz_session_history

SangJunBak:
For some customers, they were using more memory than expected for mz_recent_activity_log_thinned_ind and from running EXPLAIN ANALYZE, it's because we weren't putting a temporal filter on mz_session_history which in their environment, had about 3GB. This change should drastically reduce the memory usage of their catalog server cluster.

### Checklist


SangJunBak:
Discussed in the weekly sync that ultimately we still want to throttle recording of sessions. Created a followup issue here https://github.com/MaterializeInc/database-issues/issues/9604.

ggevay (commented):
The plan looks good: we properly push down the new temporal filter so that the arrangement on `mz_session_history` will be affected by the filter.

But I'm wondering about long sessions that started shortly before 1 day ago: Before this PR, `mz_recent_activity_log_thinned` showed queries that began within 1 day, but now it will only show a query whose session was also started within 1 day. If there was an hours-long session started just before 1 day ago, then we might lose queries from a non-trivial part of the last 1 day. How about extending the new temporal filter a bit? Even something like 2 days would be a vast improvement on the current ∞ days, and would keep the semantics of `mz_recent_activity_log_thinned` mostly clean, because I guess there aren't many sessions that take longer than a day.

SangJunBak:
@ggevay that's a good point! I'll add a day and a comment

ggevay (approved):
Thank you, LGTM!

PR #33375, opened 2025-08-20
tests: Nightly fixes (2025-08-20)

def-:
Based on https://buildkite.com/materialize/nightly/builds/12890
### Checklist

### DIFF
--- a/src/sql/src/func.rs
+++ b/src/sql/src/func.rs
@@ -1671,10 +1671,10 @@ pub struct TableFuncPlan {
 /// 2. or a general HIR expression. This happens when it's implemented as SQL, i.e., by a call to
 ///    `sql_impl_table_func_inner`.
 ///
-/// TODO(ggevay): when a table function in 2. is used with WITH ORDINALITY or ROWS FROM, we fall
-/// back to the legacy WITH ORDINALITY implementation, which relies on the row_number window
-/// function, and is mostly broken. It can give an incorrect ordering, and also has an extreme
-/// performance problem in some cases, see
+/// TODO(ggevay, database-issues#9598): when a table function in 2. is used with WITH ORDINALITY or
+/// ROWS FROM, we fall back to the legacy WITH ORDINALITY implementation, which relies on the
+/// row_number window function, and is mostly broken. It can give an incorrect ordering, and also
+/// has an extreme performance problem in some cases. Discussed in
 /// <https://github.com/MaterializeInc/database-issues/issues/4764#issuecomment-2854572614>
 ///
 /// These table functions are somewhat exotic, and WITH ORDINALITY / ROWS FROM are also somewhat
--- a/test/sqllogictest/regex.slt
+++ b/test/sqllogictest/regex.slt
@@ -493,7 +493,7 @@ ORDER BY 1
 {hello,world}
 {hello,world}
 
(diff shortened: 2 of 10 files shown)


test/sqllogictest/regex.slt
```diff
@@ -493,9 +493,8 @@ ORDER BY 1
 {hello,world}
 {hello,world}
 
-# TODO: Correct ordering when database-issues#4764 is fixed.
 query T
-SELECT foo FROM regexp_split_to_table('the quick brown fox jumps over the lazy dog', '\s+') AS foo ORDER BY 1
+SELECT foo FROM regexp_split_to_table('the quick brown fox jumps over the lazy dog', '\s+') WITH ORDINALITY AS t(foo, n) ORDER BY n
```
def-: @ggevay The order still doesn't look correct to me! Should we reopen the issue?

src/sql/src/func.rs
```diff
@@ -1674,7 +1674,7 @@ pub struct TableFuncPlan {
 /// TODO(ggevay): when a table function in 2. is used with WITH ORDINALITY or ROWS FROM, we fall
 /// back to the legacy WITH ORDINALITY implementation, which relies on the row_number window
 /// function, and is mostly broken. It can give an incorrect ordering, and also has an extreme
-/// performance problem in some cases, see
+/// performance problem in some cases. Discussed in
```
def-: @ggevay `see <` in a single line would not trigger the closed issue detector, but the link is too long for a single line, so I changed the wording instead.

test/sqllogictest/regex.slt
ggevay: This is due to this issue: https://github.com/MaterializeInc/materialize/blob/73075adf987b59f5668366360dd009b691ab6c7b/src/sql/src/func.rs#L1674

My plan is to disable `WITH ORDINALITY` for table functions in this category for now. I've added a Sentry error for when people use one of these functions with `WITH ORDINALITY`. If it will show no usage when it's rolled out this Thursday, I'll open a PR on Friday to disable `WITH ORDINALITY` for these table functions.

src/sql/src/func.rs
ggevay: Ok, thanks!

test/sqllogictest/regex.slt
def-: Should we open a new issue to track that or is it never planned to be fixed?

ggevay: So, we shouldn't add tests with these table functions together with `WITH ORDINALITY`.

src/sql/src/func.rs
ggevay: I'll open a separate issue for this, and reference that instead.

test/sqllogictest/regex.slt
ggevay: I'll open an issue for it. I would guess that it will come up sooner or later, but maybe not this year.

ggevay: Issue: https://github.com/MaterializeInc/database-issues/issues/9598

src/sql/src/func.rs
ggevay: Opened the issue: https://github.com/MaterializeInc/database-issues/issues/9598

ggevay (approved):

PR #33377, opened 2025-08-20
adapter: remove replica disk flag

teskje:
This PR removes the `disk` flag associated with replica definitions. Whether a replica gets a disk now solely depends on the `disk_limit` configuration in its replica size: If the `disk_limit` is 0, no disk is attached, otherwise the replica gets a disk.

The change is intended to simplify the logic around disk assignment, which previously was complicated for historical reasons.

For backward compatibility reasons, the `CREATE CLUSTER REPLICA ... WITH (DISK = {true,false})` syntax still works for non-cc clusters, but has no effect now, except from showing a deprecation notice.

### Motivation

Simplifies the logic around disk allocation, preparing for introducing swap as a new special case.

### Checklist


src/catalog-protos/protos/objects.proto
ggevay, line 1: I thought that `objects.proto` changes always require the full catalog migration mechanism (`objects_v76.proto` and co.). Or is that not the case when there are only field removals, and simply totally forgetting the field's value is what we want? But I'm not sure I understand, because 83340b29b36f1abdb55a50c244854081d2ab457d also looks like a simple field removal at first glance, and it had a full migration. Or was there something there that was not a simple field removal?

src/catalog-protos/src/audit_log.rs
```diff
@@ -527,6 +528,47 @@ impl RustType<crate::objects::audit_log_event_v1::CreateClusterReplicaV3>
...
+            cluster_id: proto.cluster_id,
+            cluster_name: proto.cluster_name,
+            replica_id: proto.replica_id.map(|id| id.inner),
+            replica_name: proto.replica_name,
+            logical_size: proto.logical_size,
+            billed_as: proto.billed_as,
+            internal: proto.internal,
+            reason: proto
+                .reason
+                .into_rust_if_some("DropClusterReplicaV3::reason")?,
```
ggevay: V4

src/catalog-protos/protos/objects.proto
```diff
@@ -893,11 +903,11 @@ message AuditLogEventV1 {
   StringWrapper user = 4;
   EpochMillis occurred_at = 5;
 
-  // next-id: 40
   oneof details {
     CreateClusterReplicaV1 create_cluster_replica_v1 = 6;
     CreateClusterReplicaV2 create_cluster_replica_v2 = 33;
     CreateClusterReplicaV3 create_cluster_replica_v3 = 41;
+    CreateClusterReplicaV4 create_cluster_replica_v4 = 28;
```
ggevay: Nit: I think it's safer to not fill holes. The hole is probably due to a field removal, and the danger is that if a field was removed and then a new field with the same id is added, then some code might confuse the removed and the added field. It's probably ok here, if we don't really rely on protobuf backcompat, but do full migrations at every change. Except that I'm not sure whether we really do a full migration at every change.

src/catalog-protos/protos/objects.proto
teskje: Did you perhaps miss the migration in the fourth commit? 😄 

The truth is I am not entirely sure. I haven't tried to run CI without a migration, but I seem to remember that one is needed.

ggevay: Oh, nvm, I see that it's in the next commit!

ggevay: Yeah, I did! And then I also missed your msg :sweat_smile: 
Maybe I should continue this review on Monday with a fresh mind :sweat_smile:

misc/python/materialize/checks/all_checks/cluster.py
```diff
@@ -66,9 +66,12 @@ def validate(self) -> Testdrive:
                 ! SHOW CREATE CLUSTER create_cluster1;
                 contains: SHOW CREATE for unmanaged clusters not yet supported
 
-                > SHOW CREATE CLUSTER create_cluster2;
+                >[version<15400] SHOW CREATE CLUSTER create_cluster2;
```
ggevay: Unfortunately all of these will need to be modified when rebasing to `main`, because we bumped to 155 today.

src/catalog-protos/src/audit_log.rs
teskje: Also `Create` instead of `Drop` 🙈

src/catalog-protos/protos/objects.proto
teskje: Yeah, we don't need backwards compact, and we do exclude these proto files from compat checking in `buf.yaml`. Having smaller numbers is cheaper with the variable-length encoding, though I don't think performance matters much here. Changed to use the next tag ID (43) instead.

misc/python/materialize/checks/all_checks/cluster.py
teskje: Yep, done!

teskje:
TFTR!

PR #33378, opened 2025-08-20
adapter: lazily clone and update the state

petrosagg:
Some catalog transactions don't cause any changes to the state but they would advance the catalog revision, causing any concurrent DDL transactions to fail. Until we have better transaction support (e.g with retries) we can avoid the issue by only advancing the catalog revision if the state actually changed.

Fixes MaterializeInc/database-issues#9597

### Checklist

### DIFF
--- a/src/adapter/src/catalog/transact.rs
+++ b/src/adapter/src/catalog/transact.rs
@@ -474,7 +474,7 @@ impl Catalog {
     }
 
     /// Performs the transaction described by `ops` and returns the new state of the catalog, if
-    /// has chagned. If `ops` don't result in a change in the state this method returns `None`.
+    /// it has changed. If `ops` don't result in a change in the state this method returns `None`.
     ///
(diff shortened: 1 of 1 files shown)


src/adapter/src/catalog/transact.rs
```diff
@@ -472,7 +473,8 @@ impl Catalog {
         })
     }
 
-    /// Performs the transaction described by `ops`.
+    /// Performs the transaction described by `ops` and returns the new state of the catalog, if
+    /// has chagned. If `ops` don't result in a change in the state this method returns `None`.
```
aljoscha:
```suggestion
    /// Performs the transaction described by `ops` and returns the new state of the catalog, if
    /// it has changed. If `ops` don't result in a change in the state this method returns `None`.
```

aljoscha (approved):
neat!

PR #33381, opened 2025-08-20
terraform tests: Default to enableLicenseKeyChecks: true

def-:
Seen in https://materializeinc.slack.com/archives/C07PN7KSB0T/p1755682291053539

@doy-materialize Is it fine to change this globally? The terraform test just uses the default values.yaml without modifying it. If this is not fine, I'll have to pass it in specifically for the test.

### Checklist

### DIFF
--- a/ci/nightly/pipeline.template.yml
+++ b/ci/nightly/pipeline.template.yml
@@ -1390,6 +1390,7 @@ steps:
             args: [--no-cleanup]
             ci-builder: stable
       branches: "main v*.* *aws* *tf* *terraform* *helm* *self-managed* *orchestratord*"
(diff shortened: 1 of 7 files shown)


doy-materialize:
we don't want to actually change this in the helm chart until the next real self-managed release, in case we need customers to use an updated helm chart or make a patch release or whatever - i think doing it in the test for now will be better.

def-:
Done

def-:
This works fine on Azure and GCP, but on the AWS terraform module I couldn't find a way to set `enableLicenseKeyChecks: true` while keeping everything else. I believe it's this issue: https://github.com/MaterializeInc/terraform-aws-materialize/issues/71

PR #33386, opened 2025-08-20
Change k8s namespace UI for debug tool

SangJunBak:
### Motivation

Before we'd just loop through all the namespaces and find the necessary services for scraping via the first one. Issue with doing so is:
a) It doesn't make it explicit that we need the user to target the k8s namespace of the materialize deployment
b) If a user has multiple materialize instances in the same namespace, it's quite random which materialize instance we choose

Thus we:
- Separate CR namespace with other k8s namespaces via additional-k8s-namespace
-  Add a materialize instance argument, but by default we choose the last created Materialize instance

Followup to https://github.com/MaterializeInc/materialize/pull/33129#discussion_r2242022949

### Checklist

### DIFF
--- a/doc/user/content/integrations/mz-debug/self-managed.md
+++ b/doc/user/content/integrations/mz-debug/self-managed.md
@@ -64,34 +64,37 @@ Each resource type directory also contains a `describe.txt` file with the output
 
 {{% integrations/mz-debug/memory-profiles %}}
 
-## Examples
-### Viewing Materialize instance names in the Kubernetes namespace `materialize-environment`
+
+
+## Prerequisite: Get the Materialize instance name
+
+To use `mz-debug`, you need to specify the <a href="#k8s-namespace">Kubernetes namespace (`--k8s-namespace`)</a> and the <a href="#mz-instance-name">Materialize instance name (`--mz-instance-name`)</a>. To retrieve the Materialize instance name, you can use kubectl. For example, the following retrieves the name of the Materialize instance(s) running in the Kubernetes namespace `materialize-environment`:
 ```
 kubectl --namespace materialize-environment get materializes.materialize.cloud
 ```
-The output should look something like:
+The command should return the NAME of the Materialize instance(s) in the namespace:
 ```
 NAME
 12345678-1234-1234-1234-123456789012
 ```
 
-### Debugging a Materialize instance that lives in the Kubernetes namespace `materialize-environment` with Materialize instance `12345678-1234-1234-1234-123456789012`
+## Examples
+
+### Debug a Materialize instance running in a namespace
+
+The following example uses `mz-debug` to collect debug information for the Materialize instance (`12345678-1234-1234-1234-123456789012` obtained in the Prerequisite) running in the Kubernetes namespace `materialize-environment`:
 
 ```shell
-mz-debug self-managed --k8s-namespace materialize-environment --mz-instance-name 12345678-1234-1234-1234-123456789012
+mz-debug self-managed --k8s-namespace materialize-environment \
+--mz-instance-name 12345678-1234-1234-1234-123456789012
 ```
 
-### Debugging a Materialize instance with supporting infrastructure in other namespaces (e.g. `materialize`)
+### Include information from additional kubernetes namespaces
+
+When debugging a Materialize instance, you can also include information from other namespaces via <a href="#additional-k8s-namespace">`--additional-k8s-namespace`</a>. The following example collects debug information for the Materialize instance running in the Kubernetes namespace `materialize-environment` as well as debug information for the namespace `materialize`:
 
 ```shell
-mz-debug self-managed --k8s-namespace materialize-environment --mz-instance-name 12345678-1234-1234-1234-123456789012 --additional-k8s-namespace materialize
-```
-
-### Debug namespaces without automatic port-forwarding
-
-```shell
-mz-debug self-managed \
-    --k8s-namespace materialize-environment \
-    --mz-instance-name 12345678-1234-1234-1234-123456789012 \
-    --mz-connection-url 'postgres://root@127.0.0.1:6875/materialize?sslmode=disable'
+mz-debug self-managed --k8s-namespace materialize-environment \
+--mz-instance-name 12345678-1234-1234-1234-123456789012 \
+--additional-k8s-namespace materialize
 ```
--- a/doc/user/data/mz-debug/self_managed_options.yml
+++ b/doc/user/data/mz-debug/self_managed_options.yml
@@ -5,16 +5,14 @@ rows:
   - Option: "`--k8s-namespace <NAMESPACE>`"
     Description: |
 
-      <a name="k8s-namespace"></a> The Kubernetes namespace of the Materialize instance. This is
-      required.
+      <a name="k8s-namespace"></a> **Required**. The Kubernetes namespace of the Materialize instance.
 
   - Option: "`--mz-instance-name <MZ_INSTANCE_NAME>`"
     Description: |
 
-      <a name="mz-instance-name"></a> The Materialize instance to target. This is
-      required.
+      <a name="mz-instance-name"></a> **Required**. The Materialize instance to target.
 
-  - Option: "`--dump-k8s`"
+  - Option: "`--dump-k8s <boolean>`"
     Description: |
 
       <a name="dump-k8s"></a>
@@ -34,7 +32,7 @@ rows:
 
       Defaults to the `KUBERNETES_CONTEXT` environment variable.
 
-  - Option: "`--k8s-dump-secret-values`"
+  - Option: "`--k8s-dump-secret-values <boolean>`"
     Description: |
 
       <a name="k8s-dump-secret-values"></a> If `true`, include unredacted secrets in the


doc/user/content/integrations/mz-debug/self-managed.md
```diff
@@ -67,26 +66,22 @@ Each resource type directory also contains a `describe.txt` file with the output
...
 
-### Debug the `materialize-environment` namespace
+### Debugging a Materialize instance that lives in the Kubernetes namespace `materialize-environment`
 
 ```shell
 mz-debug self-managed --k8s-namespace materialize-environment
 ```
 
-### Debug the `materialize` namespace
+### Debugging Kubernetes namespace `materialize` that does not contain Materialize instances
```
alex-hunt-materialize: Maybe rephrase this to something like `Debugging a Materialize instance with supporting infrastructure in other namespaces`?

doc/user/data/mz-debug/self_managed_options.yml
```diff
@@ -2,6 +2,19 @@ columns:
...
+
+      <a name="k8s-namespace"></a> The Kubernetes namespace of the Materialize instance. This is
+      required.
+
+  - Option: "`--mz-instance-name <MZ_INSTANCE_NAME>`"
+    Description: |
+
+      <a name="mz-instance-name"></a> The Materialize instance to target.
+
+      Defaults to the latest Materialize instance.
```
alex-hunt-materialize: Can we just make this required? The concept of "latest" doesn't make much sense here. Does this mean last updated? Does it mean last created? What if the latest is marked for deletion?

src/mz-debug/src/kubectl_port_forwarder.rs
```diff
@@ -174,38 +180,72 @@ pub async fn find_environmentd_service(
...
+            let Some(name) = service.metadata.name.clone() else {
+                return None;
+            };
+            let Some(spec) = service.spec.clone() else {
+                return None;
+            };
+            let Some(selector) = spec.selector else {
+                return None;
+            };
+            let Some(ports) = spec.ports else { return None };
```
alex-hunt-materialize: Why the change from the `?` operator?

src/mz-debug/src/main.rs
```diff
@@ -52,15 +52,22 @@ pub struct SelfManagedDebugModeArgs {
...
+
+    /// The k8s namespace that the Materialize instance is running in. This is necessary to interact
+    /// with the k8s API for gathering logs, port forwarding information, and information about the user's
+    /// environment.
+    #[clap(long)]
+    k8s_namespace: String,
+    /// The name of the Materialize instance to target. By default, the tool will target the first
+    /// Materialize instance in the namespace.
+    #[clap(long)]
+    mz_instance_name: Option<String>,
```
alex-hunt-materialize: Lets just make this required.

src/mz-debug/src/kubectl_port_forwarder.rs
```diff
@@ -174,38 +180,72 @@ pub async fn find_environmentd_service(
...
+                .and_then(|labels| labels.get("environmentd.materialize.cloud/service-id"))
+                .map(|s| s.to_string())
+        })
+        .collect();
+
+    let services: Api<Service> = Api::namespaced(client.clone(), k8s_namespace);
+    let services = services
+        .list(&ListParams::default())
+        .await
+        .with_context(|| format!("Failed to list services in namespace {}", k8s_namespace))?;
```
alex-hunt-materialize: If we're going to be getting all the services anyway, we might as well filter them based on their owner reference to the Materialize CR instead. That will save us from collecting this list of pods.

In fact, we can do that for all objects created by orchestratord.

src/mz-debug/src/utils.rs
```diff
@@ -77,43 +79,97 @@ pub async fn get_k8s_auth_mode(
...
+    let materialize_cr = object_list
+        .items
+        .into_iter()
+        .find(|item| item.metadata.name.as_ref() == Some(mz_instance_name))
+        .with_context(|| {
+            format!(
+                "Could not find Materialize CR with name: {}",
+                mz_instance_name
+            )
+        })?;
```
alex-hunt-materialize: Why not just use a get? There can only be one with a specific namespace and name.

src/mz-debug/src/kubectl_port_forwarder.rs
SangJunBak: Ah so the reason why I did it this way was the `ownerReference` for Cluster services only link to the environmentd statefulset while the environmentd `ownerReference`points to the Materialize CR. e.g.

**mzxco8zi0o48-cluster-u1-replica-u1-gen-1:**
```
name: mzxco8zi0o48-cluster-u1-replica-u1-gen-1
  namespace: materialize-environment
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: StatefulSet
    name: mzxco8zi0o48-environmentd-1
    uid: 7214f7a9-e6fa-45f8-b2ee-f3d78a2af55b
  resourceVersion: "780"
  uid: 587c6791-f0fe-49cf-acc6-31d1d59e5a72
```

**mzxco8zi0o48-environmentd-1:**
```
ownerReferences:
  - apiVersion: materialize.cloud/v1alpha1
    blockOwnerDeletion: true
    kind: Materialize
    name: 12345678-1234-1234-1234-123456789012
    uid: 9cc32546-2421-4911-bc2c-234d842c3c67
```

Given the pods already have the `materialize.cloud/organization-name` label that maps to the Materialize CR, thought it'd be less code to just filter the services on the set of cluster pods with that label value. If you think it's more correct to:
```
1. Get list of tuples: (cluster_service, cluster_service.owner.statefulset, cluster_service.owner.statefulset.materialize_cr)
2. Filter on the tuples where cluster_service.owner.statefulset.materialize_cr == mz_instance_name
```

I can make the change!

src/mz-debug/src/kubectl_port_forwarder.rs
SangJunBak: Ah for some reason I thought the `?` operator short circuited the entire function rather than just the closure. My bad and good catch!

src/mz-debug/src/utils.rs
SangJunBak: Good call. Changed!

doc/user/data/mz-debug/self_managed_options.yml
SangJunBak: Moreso last created but that's a good point. Mainly made it optional for backwards compatibility and less friction to get started. But it's probably better to be explicit.

src/mz-debug/src/kubectl_port_forwarder.rs
```diff
@@ -131,41 +131,47 @@ pub struct ServiceInfo {
...
+
+    // Find the first sql service that contains balancerd
+    let maybe_service =
+        services
+            .iter()
+            .find_map(|service| match (&service.metadata.name, &service.spec) {
+                (Some(service_name), Some(spec)) => {
+                    if !service_name.to_lowercase().contains("environmentd") {
```
alex-hunt-materialize, line 158: Not something to fix in this PR, but I am realizing now that this could match both the generation service and the globally active one. It is unclear which we are looking for in this function, but I would guess we want the active one.

src/mz-debug/src/kubectl_port_forwarder.rs
```diff
@@ -174,38 +180,81 @@ pub async fn find_environmentd_service(
...
+                //     return None;
+                // };
+                // let Some(selector) = spec.selector else {
+                //     return None;
+                // };
+                // let Some(ports) = spec.ports else { return None };
+
+                // // Filter by cluster services
+                // if let Some(service_id) = selector.get("environmentd.materialize.cloud/service-id") {
+                // if !service_ids_to_scrape.contains(service_id) {
```
alex-hunt-materialize: Leftover comments?

src/mz-debug/src/kubectl_port_forwarder.rs
```diff
@@ -174,38 +180,81 @@ pub async fn find_environmentd_service(
...
+        .with_context(|| format!("Failed to list pods in namespace {}", k8s_namespace))?;
+
+    // Create a set of service IDs from pod labels
+    let service_ids_to_scrape: std::collections::BTreeSet<String> = pods
+        .iter()
+        .filter_map(|pod| {
+            pod.metadata
+                .labels
+                .as_ref()
+                .and_then(|labels| labels.get("environmentd.materialize.cloud/service-id"))
```
alex-hunt-materialize: I found a bug in our service labeling! We don't have any labels identifying which materialize instance a service should point to. This may cause services to go to the wrong materialize instance!
I just filed https://github.com/MaterializeInc/cloud/issues/11527

This also means you should probably change to getting the services with an owner reference pointing at environmentd.

SangJunBak: Oh neat! And with that we can skip fetching the pods

doc/user/data/mz-debug/self_managed_options.yml
```diff
@@ -2,6 +2,18 @@ columns:
   - column: "Option"
   - column: "Description"
 rows:
+  - Option: "`--k8s-namespace <NAMESPACE>`"
+    Description: |
+
+      <a name="k8s-namespace"></a> The Kubernetes namespace of the Materialize instance. This is
```
kay-kim: Could we maybe start with  **Required.** in front of the "The Kubernetes..." that way, it's more skimmable and catches the eye.

doc/user/data/mz-debug/self_managed_options.yml
```diff
@@ -2,6 +2,18 @@ columns:
...
+      required.
+
+  - Option: "`--mz-instance-name <MZ_INSTANCE_NAME>`"
+    Description: |
+
+      <a name="mz-instance-name"></a> The Materialize instance to target. This is
```
kay-kim: same here.  And link to the example that shows how to get the name via kubectl? or include in here as well.

doc/user/content/integrations/mz-debug/self-managed.md
```diff
@@ -66,27 +65,33 @@ Each resource type directory also contains a `describe.txt` file with the output
 {{% integrations/mz-debug/memory-profiles %}}
 
 ## Examples
+### Viewing Materialize instance names in the Kubernetes namespace `materialize-environment`
```
kay-kim: Since this is more about kubectl ... maybe make it a prereq?  like (spitballing ... )(alternatively, you can make the 3 examples into 1 example with steps where 1st step is to get the instance name. ...)

### Prereq: Get the Materialize instance names

To use mz-debug, you need to specify the [Kubernetes environment](#k8s-namespace) and the [Materialize instance name](#mz-instance-name).  To retrieve the Materialize instance, you can use `kubectl`.  For example, the following retrieves the name of the Materialize instance(s) running in the Kubernetes namespace `materialize-environment`:

doc/user/content/integrations/mz-debug/self-managed.md
```diff
@@ -66,27 +65,33 @@ Each resource type directory also contains a `describe.txt` file with the output
...
+```
+kubectl --namespace materialize-environment get materializes.materialize.cloud
+```
+The output should look something like:
```
kay-kim: Instead of "look something like", we may want to possibly future proof, in case we allow for names that are not UUIDs to a more generic:

The command should return the `NAME` of the Materialize instance(s) in the namespace:

doc/user/content/integrations/mz-debug/self-managed.md
```diff
@@ -66,27 +65,33 @@ Each resource type directory also contains a `describe.txt` file with the output
...
+```
+NAME
+12345678-1234-1234-1234-123456789012
+```
 
-### Debug the `materialize-environment` namespace
+### Debugging a Materialize instance that lives in the Kubernetes namespace `materialize-environment` with Materialize instance `12345678-1234-1234-1234-123456789012`
```
kay-kim: This title is more of a descriptive sentence. Can me make the heading slightly more generic and move the current title to be the descriptive sentence? Something like (spitballing)
Debug a Materialize instance running in a namespace
The following example uses mz-debug to collect debug information for the Materialize instance (`12345678-1234-1234-1234-123456789012` obtained in the Prereq) running in the Kubernetes namespace `materialize-environment`:

doc/user/content/integrations/mz-debug/self-managed.md
```diff
@@ -66,27 +65,33 @@ Each resource type directory also contains a `describe.txt` file with the output
...
 
 ```shell
-mz-debug self-managed --k8s-namespace materialize-environment
+mz-debug self-managed --k8s-namespace materialize-environment --mz-instance-name 12345678-1234-1234-1234-123456789012
 ```
 
-### Debug the `materialize` namespace
+### Debugging a Materialize instance with supporting infrastructure in other namespaces (e.g. `materialize`)
```
kay-kim: Same comment.  Move it into the descriptive sentence:

### Include information on additional namespaces 

When debugging a Materialize instance, you can also include information from other namespaces ([--additional-k8s-namespace](#...)).  The following example collects debug information for the Materialize instance running in the Kubernetes namespace `materialize-environment` (`12345678-1234-1234-1234-123456789012` obtained in the Prereq) as well as debug information for the namespace `materialize`:

src/mz-debug/src/kubectl_port_forwarder.rs
alex-hunt-materialize: This matching both is more problematic now that we're properly getting the clusterd pods. If we select an inactive environmentd service, we're only going to find inactive clusterd pods too.

I won't block this PR for this, but it should at least be a fast follow up.

doc/user/content/integrations/mz-debug/self-managed.md
SangJunBak: I believe kubectl is already a requirement!

PR #33390, opened 2025-08-21
adapter: support showing swap space in `mz_cluster_replica_sizes`

teskje:
This PR makes it possible to show the swap space limit in `mz_cluster_replica_sizes.disk_bytes`. There are two parts to this (included as separate commits):

* Introducing a new `swap_enabled` flag to the replica size definitions. This frees us from the need to always specify `disk_limit = 0` for swap-enabled replica sizes and allows us to instead supply the physical swap limit.
* Taking into account the memory limiter config when writing `mz_cluster_replica_sizes`, as that config can further reduce the actual swap size a replica is allowed to use.

### Motivation

Closes https://github.com/MaterializeInc/database-issues/issues/9366

### Tips for reviewers

Commits are meaningful and have meaningful descriptions.

### Checklist


src/controller/src/clusters.rs
```diff
@@ -76,9 +77,6 @@ pub struct ReplicaConfig {
 pub struct ReplicaAllocation {
     /// The memory limit for each process in the replica.
     pub memory_limit: Option<MemoryLimit>,
-    /// The memory limit for each process in the replica.
-    #[serde(default)]
-    pub memory_request: Option<MemoryLimit>,
```
antiguru, line 81: This seems like a risky change. Once we deploy this version to cloud, we won't be able to start without also changing the pulumi files. I think we need a version that supports both `memory_request` and `swap_enabled`.

teskje: I've deployed this version to my staging env and it was fine. Why do you think it wouldn't be able to start? Note that serve-json doesn't error on unknown fields, it simply ignores them.

teskje:
> When I introduced the `memory_request` property I decided against naming swap in the replica size configuration. The reason for this were that I wanted to have flexibility of changing parameters we pass to Kubernetes, and no logic changes in Materialize to change the parameters. This change is a departure from this, and I'm not sure which of the alternatives is better. I personally do not mind more information in the replica size definition.

We had a discussion about this [in Slack](https://materializeinc.slack.com/archives/C08ACQNGSQK/p1755003627178899). Basically the issue here is that we need some way to know that the replica is going to use swap, since that's configured by the external orchestration, not by the database. Having an explicit flag was deemed better than the alternative for checking for node selectors (which we don't control in self-managed). Removing the `memory_request` property is not necessary, but Alex reported that configuring it for self-managed is a hassle, so it seems not having it simplifies things a bit.

src/controller/src/clusters.rs
teskje: The one thing that will break are the `*-swap` sizes, as they won't get a Burstable QoS class anymore. That seems fine, we only use them internally anyway. If we merge a cloud PR to update them before the staging/canary roll, there shouldn't be any disruption either.

antiguru (approved):
Thank you for the clarification, I retract my concerns!

teskje:
TFTRs!

PR #33399, opened 2025-08-21
[controller] Stream through retractions instead of snapshotting

bkirwi:
### Motivation

We've observed high memory use from the collection data that's temporarily fully snapshotted in memory, and there's no reason we can't just stream out an updated batch as we go.

Part of https://github.com/MaterializeInc/incidents-and-escalations/issues/286

### Tips for reviewer

My first version used the unconsolidated stream, but that could cause exponential growth in the data size in some bad cases - if we called this method frequently without having compaction run in the background. Using the consolidated snapshot should use no more memory than the current version, but typically less, and doesn't have the risk of exponential data growth.


teskje (approved):
LGTM!

It'd be great to give the same treatment to `partially_truncate_status_history` as a follow-up, if you're up for it!

aljoscha:
🙌

bkirwi:
Thanks for the review!

PR #33400, opened 2025-08-21
sql_server: add deterministic capture instance selection

ptravers:
### Motivation

    https://github.com/MaterializeInc/database-issues/issues/9607

### Tips for Reviewer
There are a variety of fields on `cdc.change_tables`. we want to pick one that is strictly increasing so that we can pick the newest change table/capture instance. we want to pick the newest change table when a source is created to make sure we are getting the latest schema changes. change tables in SQL server reflect the schema of the source table when the change table was created. notably though change tables do not afaict have any of the constraints of source tables.

fields:
- object_id - the id of the change table is not guaranteed to be strictly increasing.
- create_date - should be, assuming monotonic clock, strictly increasing. but if clock skews then we no longer have a strictly increasing sequence.
- start_lsn - appears to be strictly increasing when subjected to enabling CDC in parallel for multiple tables.

went with start_lsn as it appears to be the safest option. or at least I haven't made it break yet. I did look at using a combo of start_lsn and create_date but in the end it didn't seem worth the added complexity.

### Checklist

### DIFF
--- a/src/sql-server-util/src/desc.rs
+++ b/src/sql-server-util/src/desc.rs
@@ -38,6 +38,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::BTreeSet;
 use std::sync::Arc;
 
+use crate::cdc::Lsn;
 use crate::{SqlServerDecodeError, SqlServerError};
 
 include!(concat!(env!("OUT_DIR"), "/mz_sql_server_util.rs"));
@@ -159,6 +160,8 @@ pub struct SqlServerTableRaw {
     pub name: Arc<str>,
     /// The capture instance replicating changes.
     pub capture_instance: Arc<str>,
+    /// The start lsn of the capture instance was created replicating changes.
+    pub capture_instance_start_lsn: Lsn,
     /// Columns for the table.
     pub columns: Arc<[SqlServerColumnRaw]>,
 }
@@ -985,6 +988,7 @@ impl SqlServerRowDecoder {
 mod tests {
     use std::collections::BTreeSet;
 
+    use crate::cdc::Lsn;
     use crate::desc::{
         SqlServerColumnDecodeType, SqlServerColumnDesc, SqlServerTableDesc, SqlServerTableRaw,
         tiberius_numeric_to_mz_numeric,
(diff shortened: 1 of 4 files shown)


src/sql-server-util/src/desc.rs
```diff
@@ -159,6 +160,10 @@ pub struct SqlServerTableRaw {
     pub name: Arc<str>,
     /// The capture instance replicating changes.
     pub capture_instance: Arc<str>,
+    /// The date the capture instance was created replicating changes.
+    pub capture_instance_create_date: NaiveDateTime,
```
ptravers: I don't think this is valuable. can anyone convince me otherwise?

ptravers: removing it.

src/sql/src/pure/references.rs
```diff
@@ -246,16 +247,26 @@ impl<'a> SourceReferenceClient<'a> {
...
+                    unique_tables
+                        .entry(key)
+                        .and_modify(|chosen_table| {
+                            // When multiple capture instances exist for the same table,
+                            // we prefer the one with the highest LSN because it represents
+                            // the most recent point in the transaction log where CDC began
+                            // tracking changes.
+                            if chosen_table
+                                .capture_instance_start_lsn
+                                .lt(&table.capture_instance_start_lsn)
```
petrosagg, line 264:
```suggestion
                            if chosen_table
                                .capture_instance_start_lsn
                                < &table.capture_instance_start_lsn
```

src/sql-server-util/src/desc.rs
```diff
@@ -159,6 +160,8 @@ pub struct SqlServerTableRaw {
     pub name: Arc<str>,
     /// The capture instance replicating changes.
     pub capture_instance: Arc<str>,
+    /// The start lsn of the capture instance was created replicating changes.
+    pub capture_instance_start_lsn: Lsn,
```
petrosagg, line 164: should we make a `struct CaptureInstance { name: Arc<str>, start_lsn: Lsn }`?

petrosagg (approved):
Nice!

src/sql-server-util/src/desc.rs
ptravers: maybe? I don't have strong opinions

PR #33401, opened 2025-08-22
storage: downgrade dependency since based on dependent's upper

petrosagg:
### Motivation

The storage dataflows that maintain sources and sinks don't need to hold back their inputs to the *since* frontier of their outputs. This is because all we need to resume ingestion/sinking is to be able to read the inputs at `upper - 1`, so that we produce the correct diffs for `upper` onwards.

For a long time the storage controller would propagate the overall since frontier of a collection to its dependencies but that is overly pessimistic. This commit changes this so that dependency read holds follow the implied capability of the dependent, which is always `upper - 1`.

Fixes MaterializeInc/database-issues#8897

### Tips for reviewer

The first commit is a tiny refactor I stumbled upon while working on this.

### Checklist

### DIFF
--- a/src/storage-controller/src/lib.rs
+++ b/src/storage-controller/src/lib.rs
@@ -76,11 +76,10 @@ use mz_storage_types::instances::StorageInstanceId;
 use mz_storage_types::oneshot_sources::{OneshotIngestionRequest, OneshotResultCallback};
 use mz_storage_types::parameters::StorageParameters;
 use mz_storage_types::read_holds::ReadHold;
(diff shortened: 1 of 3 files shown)


src/storage-controller/src/lib.rs
```diff
@@ -2989,31 +2997,6 @@ where
                     debug!(id = %key, ?frontier, "downgrading ingestion read holds!");
```
teskje: This log line is now outdated! Also the comment on the for loop above.

PR #33409, opened 2025-08-22
sql_server: handle upstream restores

martykulma:
Adds detection and handling for SQL Server restore.

If MZ detect the database configured as a source has been restored on the upstream SQL Server via `msdb.dbo.restorehistory`, emit a DefiniteError for all exports.

```
2025-08-22T19:23:02.468617115Z cluster-u1-replica-u1-gen-0: 2025-08-22T19:23:02.468606Z  INFO mz_storage::healthcheck: Health transition for source u5: Some(Starting) -> Some(Stalled { error: "sql-server: Source error: source must be dropped and recreated due to failure: Restore history id changed from None to Some(1)", hints: {"retracting the errored value may resume the source"}, namespaced_errors: {SqlServer: "Source error: source must be dropped and recreated due to failure: Restore history id changed from None to Some(1)"} })
```

### Motivation

Implements https://github.com/MaterializeInc/database-issues/issues/9591

### Checklist

### DIFF
--- a/src/sql-server-util/src/inspect.rs
+++ b/src/sql-server-util/src/inspect.rs
@@ -439,11 +439,11 @@ pub async fn ensure_database_cdc_enabled(client: &mut Client) -> Result<(), SqlS
 pub async fn get_latest_restore_history_id(
     client: &mut Client,
 ) -> Result<Option<i32>, SqlServerError> {
-    static DATABASE_CDC_ENABLED_QUERY: &str = "SELECT TOP 1 restore_history_id \
+    static LATEST_RESTORE_ID_QUERY: &str = "SELECT TOP 1 restore_history_id \
         FROM msdb.dbo.restorehistory \
         WHERE destination_database_name = DB_NAME() \
         ORDER BY restore_history_id DESC;";
-    let result = client.simple_query(DATABASE_CDC_ENABLED_QUERY).await?;
+    let result = client.simple_query(LATEST_RESTORE_ID_QUERY).await?;
 
     match &result[..] {
         [] => Ok(None),
--- a/src/storage/src/source/sql_server/progress.rs
+++ b/src/storage/src/source/sql_server/progress.rs
@@ -107,7 +107,7 @@ pub(crate) fn render<G: Scope<Timestamp = Lsn>>(
             // emit a definite error at the max LSN, but we also have to terminate the RLU probes
             // to ensure that the error propogates to downstream consumers, otherwise it will
             // wait in reclock as the server LSN will always be less than the LSN of the definite
-            // error. 
+            // error.
             let current_restore_history_id = get_latest_restore_history_id(&mut client).await?;
             if current_restore_history_id != extras.restore_history_id {
                 tracing::error!("Restore happened, exiting");


ptravers:
ran nightlies to check the resumption test: https://buildkite.com/materialize/nightly/builds/12951

martykulma:
> ran nightlies to check the resumption test: https://buildkite.com/materialize/nightly/builds/12951

thanks! I forgot those those don't run as part of the normal tests.

I'll have to dig into the CI failure, I've run this a few times locally, and it passes.
```
2025-08-22T23:36:54.695937Z  WARN mz_testdrive::action::consistency: No Catalog state on disk, skipping consistency check
--- verify-source-failed.td
> SELECT COUNT(*) > 0 FROM mz_internal.mz_source_statuses WHERE status = 'stalled';
rows match; continuing at ts 1755905814.8820739, took 0.011190417s
> SELECT * FROM t1;
query error matches; continuing
2025-08-22T23:36:55.098297Z  WARN mz_testdrive::action::consistency: No Catalog state on disk, skipping consistency check
==> mzcompose: test case workflow-backup-restore succeeded
```

martykulma:
One of the issues appears to be that we are emitting definite errors at `Lsn::minimum()`.  The other sources pick what appears to be the maximum `FromTime`.

https://github.com/MaterializeInc/materialize/blob/42ed12d3a030494269546404bb9a8328f1e96ee1/src/storage/src/source/sql_server/replication.rs#L451-L452


After fixing that, I ran into a different issue locally. The source was reporting the error correctly, but `SELECT * FROM t1` would return successfully with no results. 

```
2025-08-23T00:58:18.568589Z  WARN mz_testdrive::action::consistency: No Catalog state on disk, skipping consistency check
--- verify-source-failed.td
> SELECT COUNT(*) > 0 FROM mz_internal.mz_source_statuses WHERE status = 'stalled';
rows match; continuing at ts 1755910699.3959823, took 0.719109573s
> SELECT * FROM t1;
query succeeded, but expected error containing "Source error"
query error didn't match; sleeping to see if dataflow produces error shortly 🕑 50ms
```

It looks like as long as the SQL progress operator is running, the definite error isn't recognized downstream.  If I force the progress operator to return, the test passes again.

martykulma:
We have a [passing test](https://buildkite.com/materialize/nightly/builds/12957#0198e19e-29bc-4786-9616-61b3cb17cc61), though it feels a bit odd since it has to do check for a possible restore in 2 places (replication.rs and progress.rs).

I believe I could restructure the SQL Server render to do the restore check before calling [replication::render](https://github.com/MaterializeInc/materialize/blob/7d4843f33bb12fdc6e7aaa1e09c55ce48d084b16/src/storage/src/source/sql_server.rs#L161), and just return mostly empty `Stream`s with the DefiniteError.

Alternatively, I suppose I could also just create an mpsc channel that `replication::render` publishes to and `progress::render` consumes. I didn't see a way to propagate errors between operators, but maybe there is a pattern we use already?

Edit: We may also want to block the progress operator until the snapshot is completed, otherwise the new table locking strategy could result in a large number of bindings being minted.  I'm starting to favor the last option (or something along that line).  We seem to have a similar technique with PG with the `slot_ready` stream

@petrosagg do you have any thoughts?

src/sql-server-util/src/inspect.rs
```diff
@@ -430,6 +430,30 @@ pub async fn ensure_database_cdc_enabled(client: &mut Client) -> Result<(), SqlS
...
+/// - <https://learn.microsoft.com/en-us/sql/relational-databases/system-tables/restorehistory-transact-sql?view=sql-server-ver17>
+/// - <https://learn.microsoft.com/en-us/sql/t-sql/statements/create-table-transact-sql-identity-property?view=sql-server-ver17>
+pub async fn get_latest_restore_history_id(
+    client: &mut Client,
+) -> Result<Option<i32>, SqlServerError> {
+    static DATABASE_CDC_ENABLED_QUERY: &str = "SELECT TOP 1 restore_history_id \
+        FROM msdb.dbo.restorehistory \
+        WHERE destination_database_name = DB_NAME() \
+        ORDER BY restore_history_id DESC;";
+    let result = client.simple_query(DATABASE_CDC_ENABLED_QUERY).await?;
```
ptravers: looks like a copy paste holdover with the name of the query

martykulma: 🤣 I just really like that name - nice catch!

ptravers:
I think I agree with you. it would make the most sense to me to pass an output/channel from replication to progress and have progress wait til replication signals that snapshotting is complete. where the result can be either a snapshot success or some error.

petrosagg:
> I believe I could restructure the SQL Server render to do the restore check before calling [replication::render](https://github.com/MaterializeInc/materialize/blob/7d4843f33bb12fdc6e7aaa1e09c55ce48d084b16/src/storage/src/source/sql_server.rs#L161), and just return mostly empty Streams with the DefiniteError.

We can't do async work during dataflow rendering so I think what you have in this PR is the way to go.

> Edit: We may also want to block the progress operator until the snapshot is completed, otherwise the new table locking strategy could result in a large number of bindings being minted.

I'm too worried about this, there are some upcoming changes in the reclock operator that should make it robust to this scenario.

> We seem to have a similar technique with PG with the `slot_ready` stream

We currently do that not to avoid reclock bindings but to avoid emitting replication updates that would get "stuck" in the persist sink operator until the frontier advances. The persist sink operator has the not great behavior of creating one batch per unique timestamp and each batch builder only offloads to S3 if enough data exists at that timestamp. So if a table's replication stream has data on every timestamp but not enough to trigger S3 offload you end up holding onto all of those in memory until the frontier advances. Enough of that leads to an OOM kill. This is something that can be improved in the persist_sink implementation so that it's robust to receiving a large amount of future updates.

martykulma:
TFTR!!!

PR #33410, opened 2025-08-22
mysql: add support for integer enums

ptravers:
Note: snapshotting the table and replicating the table result in different formats. we snapshot as a string representation by `select c1 from t1` then see `c1` replicated in the binlog as an `int`. so we had to update the snapshotting logic to snapshot as an int. I looked at different ways to convert enum to int one way recommended online was to do `c1+0` but this resulted in a Double being outputted due to our usage of backticks when constructing the snapshot query, plus it's pretty gross. so we are using an explicit cast to unsigned int which appears to be okay.

~I will check this afternoon if we can explicitly cast to uint16 or what casting to a generic unsigned entails.~
checked: I don't see a way to go below a u64 in mysql when casting. as far as I can tell `unsigned integer` just goes to `unsigned bigint`.
see: http://dev.mysql.com/doc/refman/8.4/en/cast-functions.html

**Note (09/01/25):** I will squash all commits into one commit before merging to enable cherry picking into a patch release.

### Motivation

    https://github.com/MaterializeInc/database-issues/issues/9615

### Checklist

### DIFF
--- a/src/mysql-util/src/desc.rs
+++ b/src/mysql-util/src/desc.rs
@@ -228,7 +228,9 @@ impl MySqlColumnDesc {
                             .all(|(self_val, other_val)| self_val == other_val)
                 } else {
                     // For the `UInt16` enum representation we support addition of new enums but
-                    // no other changes.
+                    // no other changes. We therefore weaken the assertion to allow for more variants
+                    // in the other `MySqlColumnMeta` which represents a schema newer than the schema
+                    // we snapshotted at.
                     self_enum.values.len() <= other_enum.values.len()
                         && self_enum
                             .values
--- a/src/mysql-util/src/desc.rs
+++ b/src/mysql-util/src/desc.rs
@@ -15,7 +15,7 @@ use proptest_derive::Arbitrary;
 use serde::{Deserialize, Serialize};
 
 use mz_proto::{ProtoType, RustType, TryFromProtoError};
-use mz_repr::ColumnType;
+use mz_repr::{ColumnType, ScalarType};
 
 use self::proto_my_sql_column_desc::Meta;
 
@@ -173,38 +173,6 @@ pub enum MySqlColumnMeta {
     Bit(u32),
 }
 
-impl IsCompatible for Option<MySqlColumnMeta> {
-    fn is_compatible(&self, other: &Option<MySqlColumnMeta>) -> bool {
-        match (self, other) {
-            (None, None) => true,
-            (Some(_), None) => false,
-            (None, Some(_)) => false,
-            (Some(MySqlColumnMeta::Enum(self_enum)), Some(MySqlColumnMeta::Enum(other_enum))) => {
-                // so as long as `self.values` is a compatible prefix of `other.values`, we can
-                // ignore extra values from `other.values`.
-                self_enum.values.len() <= other_enum.values.len()
-                    && self_enum
-                        .values
-                        .iter()
-                        .zip(other_enum.values.iter())
-                        .all(|(self_val, other_val)| self_val == other_val)
-            }
-            (Some(MySqlColumnMeta::Json), Some(MySqlColumnMeta::Json)) => true,
-            (Some(MySqlColumnMeta::Year), Some(MySqlColumnMeta::Year)) => true,
-            (Some(MySqlColumnMeta::Date), Some(MySqlColumnMeta::Date)) => true,
-            // Timestamps are compatible as long as we don't lose precision
-            (
-                Some(MySqlColumnMeta::Timestamp(precision)),
-                Some(MySqlColumnMeta::Timestamp(other_precision)),
-            ) => precision <= other_precision,
-            // We always cast bit columns to u64's and the max precision of a bit column
-            // is 64 bits, so any bit column is always compatible with another.
-            (Some(MySqlColumnMeta::Bit(_)), Some(MySqlColumnMeta::Bit(_))) => true,
-            _ => false,
-        }
-    }
-}
-
 #[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, Arbitrary)]
 pub struct MySqlColumnDesc {
     /// The name of the column.
@@ -278,8 +246,41 @@ impl IsCompatible for MySqlColumnDesc {
                 (Some(_), None) => false,
                 (None, Some(_)) => false,
             }
-            // Ensure any column metadata is compatible
-            && self.meta.is_compatible(&other.meta)
+            // Ensure any column metadata is compatible.
+            && match (&self.meta, &other.meta) {
+                (None, None) => true,
+                (Some(_), None) => false,
+                (None, Some(_)) => false,
+                (Some(MySqlColumnMeta::Enum(self_enum)), Some(MySqlColumnMeta::Enum(other_enum))) => {
+                    let uses_enum_string_representation: bool = self.column_type
+                            .as_ref()
+                            .map(|col_type| col_type.scalar_type != ScalarType::UInt16)
+                            .unwrap_or(false);
+
+                    if uses_enum_string_representation {
+                        self_enum.values.len() <= other_enum.values.len()
+                            && self_enum
+                                .values
+                                .iter()
+                                .zip(other_enum.values.iter())
+                                .all(|(self_val, other_val)| self_val == other_val)
+                    } else {
+                        true
+                    }
+                }
+                (Some(MySqlColumnMeta::Json), Some(MySqlColumnMeta::Json)) => true,
+                (Some(MySqlColumnMeta::Year), Some(MySqlColumnMeta::Year)) => true,
+                (Some(MySqlColumnMeta::Date), Some(MySqlColumnMeta::Date)) => true,
+                // Timestamps are compatible as long as we don't lose precision
+                (
+                    Some(MySqlColumnMeta::Timestamp(precision)),
+                    Some(MySqlColumnMeta::Timestamp(other_precision)),
+                ) => precision <= other_precision,
+                // We always cast bit columns to u64's and the max precision of a bit column
+                // is 64 bits, so any bit column is always compatible with another.
+                (Some(MySqlColumnMeta::Bit(_)), Some(MySqlColumnMeta::Bit(_))) => true,
+                _ => false,
+            }
     }
 }
 
--- a/src/mysql-util/src/desc.rs
+++ b/src/mysql-util/src/desc.rs
@@ -153,10 +153,6 @@ impl RustType<ProtoMySqlColumnMetaEnum> for MySqlColumnMetaEnum {
     }
 }
 
-trait IsCompatible {
-    fn is_compatible(&self, other: &Self) -> bool;
-}
-
 #[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize, Arbitrary)]
 pub enum MySqlColumnMeta {
     /// The described column is an enum, with the given possible values.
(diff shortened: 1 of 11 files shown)


src/mysql-util/src/desc.rs
```diff
@@ -278,8 +246,41 @@ impl IsCompatible for MySqlColumnDesc {
...
+                // Timestamps are compatible as long as we don't lose precision
+                (
+                    Some(MySqlColumnMeta::Timestamp(precision)),
+                    Some(MySqlColumnMeta::Timestamp(other_precision)),
+                ) => precision <= other_precision,
+                // We always cast bit columns to u64's and the max precision of a bit column
+                // is 64 bits, so any bit column is always compatible with another.
+                (Some(MySqlColumnMeta::Bit(_)), Some(MySqlColumnMeta::Bit(_))) => true,
+                _ => false,
+            }
```
martykulma: Out of curiosity, why delete `impl IsCompatible for Option<MySqlColumnMeta>` and move this all under `impl IsCompatible for MySqlColumnDesc`?

src/mysql-util/src/desc.rs
```diff
@@ -278,8 +246,41 @@ impl IsCompatible for MySqlColumnDesc {
...
+                    if uses_enum_string_representation {
+                        self_enum.values.len() <= other_enum.values.len()
+                            && self_enum
+                                .values
+                                .iter()
+                                .zip(other_enum.values.iter())
+                                .all(|(self_val, other_val)| self_val == other_val)
+                    } else {
+                        true
+                    }
```
martykulma: I would leave the comment explaining that the prefix is sufficient and add one why non-string enums always return true here.

src/mysql-util/src/desc.rs
```diff
@@ -278,8 +246,41 @@ impl IsCompatible for MySqlColumnDesc {
...
+            // Ensure any column metadata is compatible.
+            && match (&self.meta, &other.meta) {
+                (None, None) => true,
+                (Some(_), None) => false,
+                (None, Some(_)) => false,
+                (Some(MySqlColumnMeta::Enum(self_enum)), Some(MySqlColumnMeta::Enum(other_enum))) => {
+                    let uses_enum_string_representation: bool = self.column_type
+                            .as_ref()
+                            .map(|col_type| col_type.scalar_type != ScalarType::UInt16)
+                            .unwrap_or(false);
```
martykulma: Since this is an `Option` you can us [is_some_and](https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some_and)

test/testdrive/force-source-tables.td
```diff
@@ -349,6 +361,7 @@ mysql_table public {a,b,c}
 mysql_table_new public {foo,bar}
 
 > SHOW TABLES ON mysql_source;
+mysql_table_string ""
```
martykulma: where does this table name come from?

test/mysql-cdc/types-enum.td
```diff
@@ -30,20 +29,36 @@ SET SESSION sql_mode='';
...
+1 val1
+2 val2
+0 not-present
+
+> DROP TABLE enum_type;
+
+$ mysql-execute name=mysql
+DROP TABLE enum_type;
+CREATE TABLE enum_type (f1 ENUM ('val1', 'val2'), f2 TEXT);
+INSERT INTO enum_type VALUES ('val1', 'val1'), ('val2', 'val2'), ('not-present', 'not-present');
```
martykulma: I vote we drop this section and create another table in MZ that doesn't use `WITH (TEXT COLUMNS..)`.  The validations of the 2 tables (one with text enum, one with int enum) should just be done in tandem.

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -541,12 +541,29 @@ fn build_snapshot_query(outputs: &[SourceOutputInfo]) -> String {
...
         .desc
         .columns
         .iter()
-        .map(|col| quote_identifier(&col.name))
+        .map(|col| {
+            let is_uint16 = col
+                .column_type
+                .as_ref()
+                .map(|col_type| col_type.scalar_type == ScalarType::UInt16)
+                .unwrap_or(false);
```
martykulma: another place for `is_some_and()`

src/mysql-util/src/schemas.rs
```diff
@@ -456,6 +456,10 @@ fn parse_data_type(
             };
             return Ok((ScalarType::UInt64, Some(MySqlColumnMeta::Bit(precision))));
         }
+        "enum" => {
+            let meta = MySqlColumnMeta::Enum(MySqlColumnMetaEnum { values: vec![] });
+            return Ok((ScalarType::UInt16, Some(meta)));
```
martykulma: Good opportunity for a comment as to why this is `Some(..)` vs. `None`

src/mysql-util/src/desc.rs
ptravers: will do!

src/mysql-util/src/desc.rs
ptravers: ah sorry, I should comment. we can't use is_compatible because we rely on state not inside `SqlColumnMeta`. namely the scalar type of the column.

src/mysql-util/src/desc.rs
ptravers: gotcha, thanks!

test/testdrive/force-source-tables.td
ptravers: missed a spot when refactoring the test. we delete that table now further up.

test/mysql-cdc/types-enum.td
ptravers: can do.

src/storage/src/source/mysql/snapshot.rs
ptravers: cheers.

src/mysql-util/src/schemas.rs
ptravers: fair point.

src/mysql-util/src/desc.rs
```diff
@@ -153,10 +153,6 @@ impl RustType<ProtoMySqlColumnMetaEnum> for MySqlColumnMetaEnum {
     }
 }
 
-trait IsCompatible {
-    fn is_compatible(&self, other: &Self) -> bool;
-}
-
```
ptravers, line 159: I removed it as it was only used in one place and we do not appear to leverage it in test either.

src/mysql-util/src/schemas.rs
ptravers: done.

src/storage/src/source/mysql/snapshot.rs
ptravers: done.

test/mysql-cdc/types-enum.td
ptravers: done, will also add similar validations for the int table

test/testdrive/source-tables.td
```diff
@@ -266,27 +266,39 @@ INSERT INTO mysql_table VALUES ('var1', 1234), ('var0', 0);
...
+1 1
+2 5678
+
+> SELECT * FROM mysql_table_1_string;
 var0 0
 var1 1234
 var0 1
 var1 5678
 
+> DROP TABLE mysql_table_1_string;
```
ptravers, line 300: looking at removing this line.

test/testdrive/force-source-tables.td
```diff
@@ -269,27 +269,39 @@ INSERT INTO mysql_table VALUES ('var1', 1234), ('var0', 0);
...
+1 1
+2 5678
+
+> SELECT * FROM mysql_table_1_string;
 var0 0
 var1 1234
 var0 1
 var1 5678
 
+> DROP TABLE mysql_table_1_string;
```
ptravers: looking at removing this line.

test/mysql-cdc-old-syntax/mysql-cdc.td
```diff
@@ -679,6 +662,38 @@ DELETE FROM mixED_CAse;
...
+
+> SELECT * FROM enum_table;
+1
+2
+
+> SELECT * FROM another_enum_table;
+1
+2
+
+> DROP SOURCE mz_source_both_enum;
```
ptravers: looking at removing this line.

test/mysql-cdc-old-syntax/mysql-cdc.td
```diff
@@ -679,6 +662,38 @@ DELETE FROM mixED_CAse;
...
+  FROM MYSQL CONNECTION mysql_conn
+  FOR TABLES (
+    public."enum_table"
+  );
+
+> SELECT * FROM enum_table;
+1
+2
+
+> DROP SOURCE mz_source_enum CASCADE;
```
martykulma: this looks like a repeat of the test just below it, probably could just drop it.

test/mysql-cdc-old-syntax/types-enum.td
```diff
@@ -71,10 +82,21 @@ val1 val1
...
 INSERT INTO enum_type VALUES ('val1', 'val1');
 
 ! SELECT * FROM enum_type;
 contains:incompatible schema change: column f1 in table enum_type has been altered
+
+> SELECT * FROM enum_type_int;
+1 val1
+2 val2
+2 val1
+1 val1
```
martykulma: This actually looks like a correctness issue.

Given the `ALTER TABLE` above. 
```
   1       2       3       4
('val2', 'val1', 'val3', 'val4')
```

I would expect to see
```
2 val1
1 val2
2 val1
2 val1
```

test/mysql-cdc/types-enum.td
```diff
@@ -30,78 +29,164 @@ SET SESSION sql_mode='';
...
+1 val1
+2 val2
+2 not-present
+1 val1
+2 val2
+2 not-present
+1 val1
+2 val1
+2 not-present
+2 val1
```
martykulma: Ditto for these.  f2 `not-present` has f1 = `val2`, which is 1 after the `ALTER TABLE`

ptravers: looking

test/mysql-cdc-old-syntax/types-enum.td
ptravers: looking

test/mysql-cdc-old-syntax/mysql-cdc.td
ptravers: will do

ptravers: done

test/mysql-cdc-old-syntax/types-enum.td
ptravers: done

test/mysql-cdc/types-enum.td
ptravers: done

src/mysql-util/src/desc.rs
```diff
@@ -173,21 +169,74 @@ pub enum MySqlColumnMeta {
...
+                    self_enum.values.len() <= other_enum.values.len()
+                        && self_enum
+                            .values
+                            .iter()
+                            .zip(other_enum.values.iter())
+                            .all(|(self_val, other_val)| self_val == other_val)
+                } else {
+                    // For the `UInt16` enum representation we support addition of new enums but
+                    // no other changes.
+                    let other_enum_vals = other_enum.values.iter().take(self_enum.values.len());
```
martykulma: My understanding is that this will also pass if the enum starts with 3 values, `(a, b, c)` and the user drops one upstream, e.g. `(a, b)`.

If other_enum has 2 elements and self_enum has 3, other_enum_vals will have 2 (take yields N or fewer elements, if the underlying iterator doesn't have enough).

When that gets put into the `iter().zip` below, it will only compare the first 2 elements as zip will complete as soon as one iterator returns None.

Can both text/uint use the same validation?

src/mysql-util/src/schemas.rs
```diff
@@ -456,6 +456,22 @@ fn parse_data_type(
             };
             return Ok((ScalarType::UInt64, Some(MySqlColumnMeta::Bit(precision))));
         }
+        "enum" => {
+            // We must include the meta to ensure that we know during snapshot that
+            // the column must be cast to an integer.
```
martykulma: comment needs update

src/storage/src/source/mysql/replication.rs
```diff
@@ -337,6 +337,10 @@ pub(crate) fn render<G: Scope<Timestamp = GtidPartition>>(
                 let event_data = event.read_data()?;
                 metrics.total.inc();
 
+                if let Some(event_data) = &event_data {
+                    tracing::info!(%id, ?event_data, "timely-{worker_id} received binlog event");
+                }
+
```
martykulma: still needed?

src/mysql-util/src/desc.rs
ptravers: right, that was intentional but was debating if we wanted that behaviour. given that you would need to recreate if you added a new enum variant in position `c` let's say making the enum `(a,b,d)`. from experimentation in mysql you can't remove an enum variant without removing all instances of that variant.

but technically, you wouldn't get negative retractions if you went `(a,b,c)` -> `(a,b)` -> `(a,b,d)`. but due to inability to update the catalog we can't support that flow.

src/storage/src/source/mysql/replication.rs
ptravers: sorry, should have waited to request review! I was still debugging.

src/mysql-util/src/schemas.rs
ptravers: can do.

src/mysql-util/src/desc.rs
ptravers: to answer your question if we want to just support addition  then we should assert self_enum.len() <= other_enum.len()

src/storage/src/source/mysql/replication.rs
ptravers: fixed

ptravers:
nightly run of last code change: https://buildkite.com/materialize/nightly/builds/13035

src/mysql-util/src/desc.rs
ptravers: refactored to assert that we will accept more enum variants than we saw at snapshot. worth noting that by allowing addition of enum variants we do allow the customer to do the following:
1. create table with enum `foo` containing `(a)`.
2. add enum variants to `foo`: `(a)` -> `(a,b,x)`.
3. change ordering of `b` and `x`: `(a,b,x)` -> `(a,x,b)`.
4. see negative retractions.

ptravers: if we don't allow adding or removing enum variants then we are not offering any improvement at this point over the existing solution.

ptravers: if we don't allow adding or removing enum variants then we are not offering any improvement at this point over the existing solution.

src/mysql-util/src/schemas.rs
ptravers: done.

src/mysql-util/src/desc.rs
martykulma: With this change, users can at least add enum values without having the source error.  I completely agree that it isn't a complete solution.  It's still possible for a user to create a situation where things go awry.

Unfortunately, MySQL updates the rows of the table and doesn't send RowEvent in binlog replication (per your testing). PG has a similar problem.  There are options to address that, but it isn't clear which is the least onerous.

I think the guidance to customers needs to be:
1. enums must be append only
2. once the new enum is created, customers should run `UPDATE` to update the rows to use the new enum values.

As for this check, after some amount of time adding enums, I agree that we won't be able to detect delete/reorder. That significantly diminishes the need for it. 

@petrosagg, any objects to a relaxed check for int enum?

ptravers:
we decided not to move forward with this work due to safety concerns.

PR #33421, opened 2025-08-25
[controller] Stream status updates like we do for metrics

bkirwi:
### Motivation

Requested in https://github.com/MaterializeInc/materialize/pull/33399 - see that issue for context and rationale.

### DIFF
--- a/src/storage-controller/src/collection_mgmt.rs
+++ b/src/storage-controller/src/collection_mgmt.rs
@@ -1609,7 +1609,7 @@ where
             let as_of = f.step_back().unwrap();
 
             storage_collections
-                .snapshot(id, as_of)
+                .snapshot_cursor(id, as_of)
                 .await
                 .expect("snapshot succeeds")
         }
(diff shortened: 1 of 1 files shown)


src/storage-controller/src/collection_mgmt.rs
```diff
@@ -1659,29 +1663,35 @@ where
...
+                        //
+                        // TODO(guswynn): unpack these into health-status objects and use
+                        // their `Ord` impl.
+                        entries.push(Reverse((timestamp, row.clone())));
+
+                        // Retain some number of entries, using pop to mark the oldest entries for
+                        // deletion.
+                        while entries.len() > n {
+                            if let Some(Reverse((_, r))) = entries.pop() {
+                                let _ = deletions
```
antiguru: Is it OK to ignore the result of `add`? I'd expect a `expect` or so, to avoid silently eating the error. Or, maybe leave a comment to explain why its fine?

antiguru (approved):

src/storage-controller/src/collection_mgmt.rs
bkirwi: Yeah, better to do the expected thing here - updated.

PR #33422, opened 2025-08-26
docs: tweak authentication

kay-kim:
https://preview.materialize.com/materialize/33422/self-managed/v25.2/manage/authentication/


doc/user/content/manage/authentication.md
```diff
@@ -9,38 +9,77 @@ menu:
     weight: 8
 ---
 
-## Configuring Authentication
+## Configuring Authentication Type
```
kay-kim, line 12: https://preview.materialize.com/materialize/33422/self-managed/v25.2/manage/authentication/

PR #33449, opened 2025-08-27
docs: Add prereq section to manage roles

kay-kim:


doc/user/content/manage/access-control/manage-roles.md
```diff
@@ -110,15 +110,27 @@ Privileges by command](/manage/access-control/appendix-command-privileges/).
 
 ## Manage current privileges for a role
 
-{{< note >}}
+### Example prerequisites
```
kay-kim, line 113: Alternatively, I can pop this into an appendix page and leave the note that links to there ... but figured, for this first cut, maybe having it be prominent won't hurt too much.

antiguru (approved):

PR #33460, opened 2025-08-28
[cluster] Easier redaction in debug formatting

bkirwi:
Adds a helper to redact certain possibly-sensitive data from the debug representation - since that can easily end up in logs.

This also uses the new helper to redact a couple of datum types. (We'd probably want to do more, assuming folks like this idea.)

### Motivation

- This helps avoid future bugs where a panic message or log accidentally contains sensitive data.
- There are cases where we avoid logging interesting data since it might contain sensitive strings in deeply nested structure. If we redacted those strings reliably, we could log more and more interesting context. (For example, I could debug-log an MFP and be confident that any sensitive data in literals in that MFP would be redacted.)

### Tips for reviewer

- The current version uses the "alternate" flag to request unredacted data. This seemed reasonable to me - for debug data, that tends to only be used for pretty-printed data, which we typically don't include in logs - but there are a few other options.
- Alternatives to this include a new / custom trait or a non-log location to stash these messages. Those may be a good idea but this seems easier!


src/repr/src/scalar.rs
```diff
@@ -168,6 +168,44 @@ pub enum Datum<'a> {
...
+            Datum::UInt64(i) => f.debug_tuple("UInt64").field(i).finish(),
+            Datum::Float32(n) => f.debug_tuple("Float32").field(n).finish(),
+            Datum::Float64(n) => f.debug_tuple("Float64").field(n).finish(),
+            Datum::Date(d) => f.debug_tuple("Date").field(d).finish(),
+            Datum::Time(t) => f.debug_tuple("Time").field(t).finish(),
+            Datum::Timestamp(t) => f.debug_tuple("Timestamp").field(t).finish(),
+            Datum::TimestampTz(t) => f.debug_tuple("TimestampTz").field(t).finish(),
+            Datum::Interval(x) => f.debug_tuple("Interval").field(x).finish(),
+            Datum::Bytes(x) => f.debug_tuple("Bytes").field(x).finish(),
+            Datum::String(x) => f.debug_tuple("String").field(&redact(x)).finish(),
```
bkirwi, line 198: eg.! I imagine we might want to redact many more of these datum variants if we think this overall approach is a good idea.

src/ore/src/str.rs
```diff
@@ -436,4 +484,25 @@ mod tests {
...
+        // Don't pick up the soft-assert flag in this test, to emulate prod behaviour.
+        pub fn redact(s: impl Debug) -> impl Debug {
+            Redacting {
+                value: s,
+                debug_mode: false,
+            }
+        }
+
+        assert_eq!(
+            r#"<"XXXX_XXXXXX">"#,
```
bkirwi, line 499: This felt like a plausible compromise - preserving some structure while redacting the majority of the content - but there are lots of other choices we could make here!

ggevay (commented):
This sounds like a good idea to me! Cuts right to the heart of the problem by going for `Datum` itself.

I've verified that the panic msg in https://github.com/MaterializeInc/database-issues/issues/9632 gets correctly redacted (if I add the `redact` call to the appropriate Datum variant). 

(I also opened https://github.com/MaterializeInc/materialize/pull/33469, which just attends to some debug printing of `MirRelationExpr`. We can also merge that one, because that PR, additionally to redaction, also makes the error messages more readable by doing pretty-printing instead of Debug printing.)

bkirwi:
Thanks for taking a look! I've updated the PR to apply `redact` to all data-carrying debug variants, and kicked off a subset of nightly just in case.

bkirwi:
Nightly found one test where we asserted on the unredacted data. (I guess there's one test that doesn't set soft asserts? Useful!) Anyways I think this is good for another look.

src/repr/src/scalar.rs
```diff
@@ -168,6 +168,50 @@ pub enum Datum<'a> {
     // other datum variants should appear before `Null`.
 }
 
+impl Debug for Datum<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use mz_ore::str::*;
+        match self {
+            Datum::False => write!(f, "False"),
+            Datum::True => write!(f, "True"),
```
ggevay, line 176: Well, maybe we should redact even these? I can imagine certain columns whose True or False value is a sensitive thing. I'm not sure.

bkirwi: We could! But the existing method wouldn't do a very good job of it... we'd end up with `<XXXXX>` vs. `<XXXX>`, but those are still distinct. And even preserving the constructor names discloses data in the case of JSON, for example. It's a tradeoff between preserving some structural info for debugging but avoiding a meaningful level of disclosure...

If you have a preference one way or the other, happy to change this! I don't feel strongly about the exact tradeoffs, but I prefer to merge _something_, since the current alternative is that everything gets emitted in the clear.

ggevay: Ok, then let's merge the current PR version for now!

PR #33468, opened 2025-08-30
sasl/scram

DAlperin:
This implements the SASL/SCRAM [protocol](https://datatracker.ietf.org/doc/html/rfc5802) as implemented by postgres. The goal is to be compatible with postgres clients that support/expect SASL/SCRAM. 

### Checklist

### DIFF
--- a/src/adapter/src/coord/command_handler.rs
+++ b/src/adapter/src/coord/command_handler.rs
@@ -10,6 +10,7 @@
 //! Logic for  processing client [`Command`]s. Each [`Command`] is initiated by a
 //! client via some external Materialize API (ex: HTTP and psql).
 
+use base64::prelude::*;
 use differential_dataflow::lattice::Lattice;
 use mz_adapter_types::dyncfgs::ALLOW_USER_SESSIONS;
 use mz_auth::password::Password;
@@ -59,13 +60,16 @@ use tokio::sync::{mpsc, oneshot};
 use tracing::{Instrument, debug_span, info, warn};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 
-use crate::command::{AuthResponse, CatalogSnapshot, Command, ExecuteResponse, StartupResponse};
+use crate::command::{
+    AuthResponse, CatalogSnapshot, Command, ExecuteResponse, SASLChallengeResponse,
+    SASLVerifyProofResponse, StartupResponse,
+};
 use crate::coord::appends::PendingWriteTxn;
 use crate::coord::{
     ConnMeta, Coordinator, DeferredPlanStatement, Message, PendingTxn, PlanStatement, PlanValidity,
     PurifiedStatementReady, validate_ip_with_policy_rules,
 };
-use crate::error::AdapterError;
+use crate::error::{AdapterError, AuthenticationError};
 use crate::notice::AdapterNotice;
 use crate::session::{Session, TransactionOps, TransactionStatus};
 use crate::util::{ClientTransmitter, ResultExt};
@@ -118,6 +122,31 @@ impl Coordinator {
                 } => {
                     self.handle_authenticate_password(tx, role_name, password)
                         .await;
+                }
+
+                Command::AuthenticateGetSASLChallenge {
+                    tx,
+                    role_name,
+                    nonce,
+                } => {
+                    self.handle_generate_sasl_challenge(tx, role_name, nonce)
+                        .await;
+                }
+
+                Command::AuthenticateVerifySASLProof {
+                    tx,
+                    role_name,
+                    proof,
+                    mock_hash,
+                    auth_message,
+                } => {
+                    self.handle_authenticate_verify_sasl_proof(
+                        tx,
+                        role_name,
+                        proof,
+                        auth_message,
+                        mock_hash,
+                    );
                 }
 
                 Command::Execute {
@@ -246,6 +275,133 @@ impl Coordinator {
         .boxed_local()
     }
 
+    fn handle_authenticate_verify_sasl_proof(
+        &self,
+        tx: oneshot::Sender<Result<SASLVerifyProofResponse, AdapterError>>,
+        role_name: String,
+        proof: String,
+        auth_message: String,
+        mock_hash: String,
+    ) {
+        let role = self.catalog().try_get_role_by_name(role_name.as_str());
+        let role_auth = role.and_then(|r| self.catalog().try_get_role_auth_by_id(&r.id));
+
+        let login = role
+            .as_ref()
+            .map(|r| r.attributes.login.unwrap_or(false))
+            .unwrap_or(false);
+
+        let real_hash = role_auth
+            .as_ref()
+            .and_then(|auth| auth.password_hash.as_ref());
+        let hash_ref = real_hash.map(|s| s.as_str()).unwrap_or(&mock_hash);
+
+        let role_present = role.is_some();
+        let make_auth_err = |role_present: bool, login: bool| {
+            AdapterError::AuthenticationError(if role_present && !login {
+                AuthenticationError::NonLogin
+            } else {
+                AuthenticationError::InvalidCredentials
+            })
+        };
+
+        match mz_auth::hash::sasl_verify(hash_ref, &proof, &auth_message) {
+            Ok(verifier) => {
+                // Success only if role exists, allows login, and a real password hash was used.
+                if login && real_hash.is_some() {
+                    let role = role.expect("login implies role exists");
+                    let _ = tx.send(Ok(SASLVerifyProofResponse {
+                        verifier,
+                        auth_resp: AuthResponse {
+                            role_id: role.id,
+                            superuser: role.attributes.superuser.unwrap_or(false),
+                        },
+                    }));
+                } else {
+                    let _ = tx.send(Err(make_auth_err(role_present, login)));
+                }
+            }
+            Err(_) => {
+                let _ = tx.send(Err(AdapterError::AuthenticationError(
+                    AuthenticationError::InvalidCredentials,
+                )));
+            }
+        }
+    }
+
+    #[mz_ore::instrument(level = "debug")]
+    async fn handle_generate_sasl_challenge(
+        &mut self,
+        tx: oneshot::Sender<Result<SASLChallengeResponse, AdapterError>>,
+        role_name: String,
+        client_nonce: String,
+    ) {
+        let role_auth = self
+            .catalog()
+            .try_get_role_by_name(&role_name)
+            .and_then(|role| self.catalog().try_get_role_auth_by_id(&role.id));
+
+        let nonce = match mz_auth::hash::generate_nonce(&client_nonce) {
+            Ok(n) => n,
+            Err(e) => {
+                let msg = format!(
+                    "failed to generate nonce for client nonce {}: {}",
+                    client_nonce, e
+                );
+                let _ = tx.send(Err(AdapterError::Internal(msg.clone())));
+                soft_panic_or_log!("{msg}");
+                return;
+            }
+        };
+
+        // It's important that the mock_nonce is deterministic per role, otherwise the purpose of
+        // doing mock authentication is defeated. We use a catalog-wide nonce, and combine that
+        // with the role name to get a per-role mock nonce.
+        let send_mock_challenge =
+            |role_name: String,
+             mock_nonce: String,
+             nonce: String,
+             tx: oneshot::Sender<Result<SASLChallengeResponse, AdapterError>>| {
+                let opts = mz_auth::hash::mock_sasl_challenge(&role_name, &mock_nonce);
+                let _ = tx.send(Ok(SASLChallengeResponse {
+                    iteration_count: mz_ore::cast::u32_to_usize(opts.iterations.get()),
+                    salt: BASE64_STANDARD.encode(opts.salt),
+                    nonce,
+                }));
+            };
+
+        match role_auth {
+            Some(auth) if auth.password_hash.is_some() => {
+                let hash = auth.password_hash.as_ref().expect("checked above");
+                match mz_auth::hash::scram256_parse_opts(hash) {
+                    Ok(opts) => {
+                        let _ = tx.send(Ok(SASLChallengeResponse {
+                            iteration_count: mz_ore::cast::u32_to_usize(opts.iterations.get()),
+                            salt: BASE64_STANDARD.encode(opts.salt),
+                            nonce,
+                        }));
+                    }
+                    Err(_) => {
+                        send_mock_challenge(
+                            role_name,
+                            self.catalog().state().mock_authentication_nonce(),
+                            nonce,
+                            tx,
+                        );
+                    }
+                }
+            }
+            _ => {
+                send_mock_challenge(
+                    role_name,
+                    self.catalog().state().mock_authentication_nonce(),
+                    nonce,
+                    tx,
+                );
+            }
+        }
+    }
+
     #[mz_ore::instrument(level = "debug")]
     async fn handle_authenticate_password(
         &mut self,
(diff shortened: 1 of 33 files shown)


src/auth/src/hash.rs
```diff
@@ -79,6 +79,17 @@ pub fn hash_password(password: &Password) -> Result<PasswordHash, HashError> {
     })
 }
 
+pub fn generate_nonce(client_nonce: &str) -> String {
+    let mut nonce = vec![0u8; client_nonce.len()];
```
jasonhernandez: Let's set a minimum server nonce length of 24 bytes or higher

src/auth/src/hash.rs
```diff
@@ -114,8 +125,84 @@ pub fn scram256_verify(password: &Password, hashed_password: &str) -> Result<(),
...
+        .collect();
+
+    // Decode provided proof
+    let provided_client_proof = BASE64_STANDARD
+        .decode(proof)
+        .map_err(|_| VerifyError::InvalidPassword)?;
+
+    // Compare
+    //FIXME: constant time comparison
+    if expected_client_proof == provided_client_proof {
```
jasonhernandez: This obviously needs to be fixed before merging.

src/adapter/src/coord/command_handler.rs
```diff
@@ -246,6 +269,81 @@ impl Coordinator {
...
     }
 
+    async fn handle_authenticate_verify_sasl_proof(
+        &mut self,
+        tx: oneshot::Sender<Result<SASLVerifyProofResponse, AdapterError>>,
+        role_name: String,
+        proof: String,
+        nonce: String,
+    ) {
+        if let Some(role) = self.catalog().try_get_role_by_name(role_name.as_str()) {
```
jasonhernandez: We should return a challenge regardless of whether a user id is valid - otherwise we're effectively disclosing what user ids are / are not valid.

jasonhernandez: 1. Always return a challenge (even for invalid users)
2. Use a deterministic dummy salt based on the username hash
3. Fail authentication only during proof verification
4. Add artificial delays to normalize timing

src/adapter/src/coord/command_handler.rs
```diff
@@ -246,6 +269,81 @@ impl Coordinator {
...
+                return;
+            }
+            if let Some(auth) = self.catalog().try_get_role_auth_by_id(&role.id) {
+                if let Some(hash) = &auth.password_hash {
+                    match mz_auth::hash::scram256_parse_opts(&hash) {
+                        Ok(opts) => {
+                            let _ = tx.send(Ok(SASLChallengeResponse {
+                                iteration_count: opts.iterations.get() as usize,
+                                salt: BASE64_STANDARD.encode(opts.salt),
+                                nonce: mz_auth::hash::generate_nonce(&client_nonce),
```
jasonhernandez: We should check the length of the client nonce. 64 bytes is a reasonable maximum.

src/adapter/src/coord/command_handler.rs
```diff
@@ -246,6 +269,81 @@ impl Coordinator {
...
+                info!("role is not allowed to login");
+                // The user is not allowed to login.
+                let _ = tx.send(Err(AdapterError::AuthenticationError));
+                return;
+            }
+            if let Some(auth) = self.catalog().try_get_role_auth_by_id(&role.id) {
+                if let Some(hash) = &auth.password_hash {
+                    info!("role auth found: {:?}", auth);
+                    info!("hash={:?},proof={:?},nonce={:?}", hash, proof, nonce);
+                    match mz_auth::hash::sasl_verify(&hash, &proof, &nonce) {
```
jasonhernandez: We should check the length of all of these inputs - proof can be max 48 bytes, auth message can be max 800 bytes, and role name max 128 bytes.
A bit higher limits are also ok, but we should avoid trying to verify extremely long inputs.

jasonhernandez (changes requested):
I did a preliminary review and wanted to give some guidance on timing attacks and input validation. I assume some of these are already on your mental to-do list even if not called out.

SangJunBak:
Sorry I just saw this! Looking over it now

src/environmentd/src/http.rs
```diff
@@ -1002,6 +1002,12 @@ async fn auth(
                 });
             }
         },
+        Authenticator::Sasl(_) => {
+            // Not supported for HTTP
+            return Err(AuthError::MissingHttpAuthentication {
```
SangJunBak, line 1009: Might wanna use the `HttpsRequired` error

SangJunBak: Wouldn't this be an issue given some customers regularly uses the Console?

src/server-core/src/listeners.rs
```diff
@@ -19,6 +19,8 @@ pub enum AuthenticatorKind {
     Frontegg,
     /// Authenticate users using internally stored password hashes.
     Password,
+    /// Authenticate users using SASL.
+    Sasl,
```
SangJunBak, line 23: nit: It feels weird to have "Password" and "SASL" both be modes given SASL/PLAIN still uses a username/password. I wonder if `SaslScramSha256` would be better or if it's a bit too verbose?

Edit: Ah just realized this supports future mechanisms. Probably fine then

src/pgwire/src/protocol.rs
```diff
@@ -180,7 +187,19 @@ where
...
+                        _ => {
+                            return conn
+                                .send(ErrorResponse::fatal(
+                                    SqlState::INVALID_AUTHORIZATION_SPECIFICATION,
+                                    "expected Password message",
+                                ))
+                                .await;
+                        }
+                    }
+                }
```
SangJunBak, line 202: nit: Isn't line 207 now "expected RawAuthentication message"?

src/pgwire/src/protocol.rs
```diff
@@ -229,7 +248,19 @@ where
...
```
SangJunBak, line 263: Same "expected Password message" nit as above

src/pgwire/src/protocol.rs
DAlperin: RawAuthentication is a bit of in implementation detail specific to mz than the pg wire protocol. All of the Authentication* messages over the postgres wire protocol have the same id so are impossible to parse without knowing the context. RawAuthentication stores the raw message bytes so that we can decode them into the appropriate messages based on config. So in fact here we _are_ expecting the Password message, by way of the RawAuthentication thing

src/auth/src/hash.rs
```diff
@@ -79,6 +80,14 @@ pub fn hash_password(password: &Password) -> Result<PasswordHash, HashError> {
     })
 }
 
+pub fn generate_nonce(client_nonce: &str) -> String {
+    let mut nonce = [0u8; 24];
+    openssl::rand::rand_bytes(&mut nonce).unwrap();
```
SangJunBak: Not sure where we do a catch_unwind but may want to return an error instead?

src/environmentd/src/http.rs
DAlperin: Really this should be an unrepresentable state, you just shouldn't be allowed to specify a Sasl Authenticator for http endpoints

src/adapter/src/coord/command_handler.rs
```diff
@@ -246,6 +275,132 @@ impl Coordinator {
...
+        &self,
+        tx: oneshot::Sender<Result<SASLVerifyProofResponse, AdapterError>>,
+        role_name: String,
+        proof: String,
+        auth_message: String,
+        mock_hash: String,
+    ) {
+        let role = self.catalog().try_get_role_by_name(role_name.as_str());
+        let role_auth = role
+            .as_ref()
```
SangJunBak: nit: I think the `as_ref`s here, line 292, and line `297` are redundant

src/adapter/src/coord/command_handler.rs
```diff
@@ -246,6 +275,132 @@ impl Coordinator {
...
+
+        let login = role
+            .as_ref()
+            .map(|r| r.attributes.login.unwrap_or(false))
+            .unwrap_or(false);
+
+        let real_hash = role_auth
+            .as_ref()
+            .and_then(|auth| auth.password_hash.as_ref());
+        let hash_ref: &str = real_hash.map(|s| s.as_str()).unwrap_or(&mock_hash);
```
SangJunBak: nit: I think the explicit type annotation is redundant

src/auth/src/hash.rs
```diff
@@ -102,20 +119,114 @@ pub fn scram256_hash(password: &Password) -> Result<String, HashError> {
...
+            .map_err(|e| VerifyError::Hash(HashError::Openssl(e)))?;
+        let mut signer =
+            openssl::sign::Signer::new(openssl::hash::MessageDigest::sha256(), &signing_key)
+                .map_err(|e| VerifyError::Hash(HashError::Openssl(e)))?;
+        signer
+            .update(auth_message.as_bytes())
+            .map_err(|e| VerifyError::Hash(HashError::Openssl(e)))?;
+        let server_signature = signer
+            .sign_to_vec()
+            .map_err(|e| VerifyError::Hash(HashError::Openssl(e)))?;
```
SangJunBak: nit: Can probably de-duplicate this code w/ the client signature code by making it a helper function or anonymous function

src/pgwire/src/message.rs
```diff
@@ -69,6 +79,19 @@ pub struct FieldDescription {
     pub format: mz_pgwire_common::Format,
 }
 
+#[derive(Debug)]
+pub enum SASLServerFinalMessageKinds {
+    Verifier(String),
+    // The spec specifies an Error kind here but PG just uses
+    // it's own error handling
```
SangJunBak: nit: its*

src/adapter/src/coord/command_handler.rs
```diff
@@ -246,6 +275,132 @@ impl Coordinator {
...
+                        let _ = tx.send(Ok(SASLChallengeResponse {
+                            iteration_count: mz_ore::cast::u32_to_usize(opts.iterations.get()),
+                            salt: BASE64_STANDARD.encode(opts.salt),
+                            nonce,
+                        }));
+                    }
+                    Err(_) => {
+                        send_mock_challenge(
+                            role_name,
+                            self.catalog().state().mock_authentication_nonce(),
```
SangJunBak, line 387: It's not obvious the purpose of `mock_authentication_nonce`. We should probably document this above the definition of `send_mock_challenge` and here as well

src/auth/src/hash.rs
```diff
@@ -248,4 +359,14 @@ mod tests {
...
+
+    #[mz_ore::test]
+    #[cfg_attr(miri, ignore)]
+    fn test_mock_sasl_challenge() {
+        let username = "alice";
+        let mock = "cnonce";
+        let opts1 = mock_sasl_challenge(username, mock);
+        let opts2 = mock_sasl_challenge(username, mock);
+        assert_eq!(opts1, opts2);
+    }
```
SangJunBak, line 371: I'd actually like to see more unit tests for `sasl_verify`! Especially since it's a pure function. Maybe something like:
- Positive case
- Client proof doesn't match

src/pgwire/src/protocol.rs
```diff
@@ -266,6 +297,160 @@ where
...
+                                initial_response.client_first_message_bare_raw,
+                                server_first_message_raw,
+                                response.client_final_message_bare_raw
+                            );
+                            match adapter_client
+                                .verify_sasl_proof(
+                                    &user,
+                                    &response.proof,
+                                    &auth_message,
+                                    &mock_hash,
```
SangJunBak, line 421: I'm guessing `mock_hash` is the hash generated from the `mock_nonce` and `mock_sasl_challenge`? Not immediately obvious the purpose of it and where it comes from, especially since the control flow goes jumps back and forth between coord messages. Should probably document where it comes from and the purpose it serves!

src/pgwire/src/codec.rs
```diff
@@ -476,7 +506,220 @@ fn decode_terminate(mut _buf: Cursor) -> Result<FrontendMessage, io::Error> {
...
+//
+//   channel-binding = "c=" base64
+//                     ;; base64 encoding of cbind-input.
+//   client-final-message-without-proof =
+//                     channel-binding "," nonce [","
+//                     extensions]
+//
+//   client-final-message =
+//                     client-final-message-without-proof "," proof
+pub fn decode_sasl_response(mut buf: Cursor) -> Result<FrontendMessage, io::Error> {
```
SangJunBak, line 678: I really think we should have some datadriven tests for these messages, if for at the very least documentation. At least for the happy path. I'm pretty sure we can use the `datadriven` test framework. See `portals.pt` for reference

Maybe like a:
- Send appropriate messages to authenticate then issue a select 1
- Get an error message if auth is incorrect

src/pgwire/src/codec.rs
```diff
@@ -476,7 +506,220 @@ fn decode_terminate(mut _buf: Cursor) -> Result<FrontendMessage, io::Error> {
...
+    while let Ok(b) = buf.peek_byte() {
+        if b == b',' {
+            break;
+        }
+        v.push(buf.read_byte()?);
+    }
+    Ok(v)
+}
+
+//   extensions = attr-val *("," attr-val)
```
SangJunBak, line 542: FYI: Noticed this comment came from https://github.com/postgres/postgres/blob/a48d1ef58652229521ba4b5070e19f857608b22e/src/backend/libpq/auth-scram.c#L933. Idk where the company stands on this but I would recommend doing a lil informal "citation" to give more context / information for future developers. We do this a bunch (i.e. `mz_acl_item.rs`)

src/pgwire/src/codec.rs
```diff
@@ -476,7 +506,220 @@ fn decode_terminate(mut _buf: Cursor) -> Result<FrontendMessage, io::Error> {
...
+//                     ;; don't include that flag).
+//   client-first-message-bare =
+//                     [reserved-mext ","]
+//                     username "," nonce ["," extensions]
+//
+//   client-first-message =
+//                     gs2-header client-first-message-bare
+pub fn decode_sasl_client_first_message(mut buf: Cursor) -> Result<SASLInitialResponse, io::Error> {
+    // 1) GS2 cbind flag
+    let cbind_flag = match buf.read_byte()? {
```
SangJunBak, line 572: At least from this code review, it seems like we don't support channel binding? It also seems like we don't really do anything with the gs2_header. I feel like at the very least, we should error out on `p` given `p` says the channel requires channel binding

SangJunBak: Also maybe making it explicitly clear that we don't support channel binding, maybe in the ChannelBinding enum?

SangJunBak (approved):
Overall the code looks great, just recommendations on documentation and testing. Nits are optional but I think we should do the comments related to testing. Lemme know if you want me to clarify anything in this review

SangJunBak:
I think the next step is rigging up the orchestratord changes. Not sure if you'd like to do that or delegate that to Alex. Regardless should prolly keep the self managed peeps updated!

src/adapter/src/coord/command_handler.rs
DAlperin: added at protocol layer

jasonhernandez (approved):
Thanks for addressing my comments / review. I think this looks ok from a security standpoint. More comments / test cases in line with @SangJunBak's recommendations are nice to add but not blocking re: security.

src/pgwire/src/codec.rs
DAlperin: Added a comment above pointing to spec.

src/pgwire/src/codec.rs
DAlperin: Discussed OOB, datadriven doesn't support testing pre auth messages. Will add a follow up issue.

PR #33471, opened 2025-08-30
Refactor how we describe FETCH, and add support for PREPARE FETCH

ggevay:
This PR is in the line of work of fixing https://github.com/MaterializeInc/database-issues/issues/9634, but it doesn't fix it yet.

The first commit does a refactoring: weirdly, `describe_fetch` used to return no `RelationDesc` in the `StatementDesc`. This PR makes `describe_fetch` return a proper `StatementDesc` that includes a `RelationDesc`. This is made possible by having `SessionCatalog` nowadays, plus the PR adding a field and a method to `SessionCatalog`/`ConnCatalog` for retrieving portal descriptions. This is similar to the already-existing field and method for retrieving prepared statements.

Note that this is just a refactoring so far, in that `describe_fetch` not returning a `RelationDesc` was not causing a bug: There are 3 code paths that call `describe_fetch` (through `describe`), and each of these had its own reasons for being ok with `describe_fetch` not returning a `RelationDesc`:
1. The `describe` in `util.rs` used to do its own describing logic specifically for fetches. This code was now moved into `describe_fetch`, so we no longer have this special casing here.
2. `describe_declare` only looks at the parameter types of the `StatementDesc`, not the `RelationType`.
3. `plan_prepare` did not have to deal with `Fetch`, because `PREPARE FETCH` was not allowed.

I'd say this code is cleaner after the PR, in that `describe_fetch` is now simply correct, so we now don't have to think hard in each of the above three code paths about how we are barely skirting potential bugs.

The second commit adds support for `PREPARE FETCH`, simply by allowing `FETCH` in `parse_prepare`. This relies on the above 3. code path now getting a real `StatementDesc` from `describe`/`describe_fetch`. Note that you could already prepare a `FETCH` if you were using the Extended Query protocol, i.e., adding support for `PREPARE FETCH` just brings the `PREPARE` SQL statement into parity with the Extended Query protocol's prepared statement creation. (Preparing a fetch through the Extended Query protocol relies on the 1. code path from above.)

The second commit also adds tests for `PREPARE FETCH` and for the first commit. Note that the first commit is also tested by the existing `cursors.pt` and `portals.pt`.

The PR also adds an slt test for demonstrating https://github.com/MaterializeInc/database-issues/issues/9634, which I will fix in the next PR (https://github.com/MaterializeInc/materialize/pull/33474). This test relies on `PREPARE FETCH`: without this, it was not possible to test for https://github.com/MaterializeInc/database-issues/issues/9634 in slt, because it was not possible to make a fetch statement into a prepared statement in slt. In fact, trying to add an slt test for https://github.com/MaterializeInc/database-issues/issues/9634 was what made me go down the path that led to this PR.

Nightly: Run on https://github.com/MaterializeInc/materialize/pull/33474, which also includes these commits.

### Motivation

  * This PR adds a feature that has not yet been specified. (`PREPARE FETCH`)

### Checklist

### DIFF
--- a/doc/user/content/sql/prepare.md
+++ b/doc/user/content/sql/prepare.md
@@ -15,7 +15,7 @@ menu:
 Field | Use
 ------|-----
 name | A name for this particular prepared statement that you can later use to execute or deallocate a statement. The name must be unique within a session.
-statement  |  Any `SELECT`, `INSERT`, `UPDATE`, or `DELETE` statement.
+statement  |  Any `SELECT`, `INSERT`, `UPDATE`, `DELETE`, or `FETCH` statement.
 
 ## Details
 
--- a/src/sql-parser/src/parser.rs
+++ b/src/sql-parser/src/parser.rs
@@ -9083,7 +9083,8 @@ impl<'a> Parser<'a> {
             Statement::Select(_)
                 | Statement::Insert(_)
                 | Statement::Delete(_)
-                | Statement::Update(_),
+                | Statement::Update(_)
+                | Statement::Fetch(_),
         ) {
             return parser_err!(self, pos, "unpreparable statement").map_no_statement_parser_err();
         }
--- a/test/sqllogictest/transactions.slt
+++ b/test/sqllogictest/transactions.slt
@@ -1353,6 +1353,104 @@ type_mod false  integer The␠packed␠type␠identifier␠of␠the␠column.
 statement ok
 COMMIT;
 
+statement ok
+CREATE TABLE t1(x text, y text);
+
+statement ok
+INSERT INTO t1 VALUES (5, 6), (7, 8);
+
+# Test PREPARE FETCH.
+simple
+BEGIN;
+DECLARE c1 CURSOR FOR (
+  SELECT count(1) FROM t1
+  UNION ALL
+  SELECT count(1) FROM t1
+);
+PREPARE p1 AS FETCH 1 c1;
+EXECUTE p1;
+ROLLBACK;
+----
+COMPLETE 0
+COMPLETE 0
+COMPLETE 0
+2
+COMPLETE 1
+COMPLETE 0
+
+# Now declare a new cursor with the same name in a different transaction, and test that we realize that the meaning of
+# p1 has changed.
+# TODO: Currently, we have a bug here, see: https://github.com/MaterializeInc/database-issues/issues/9634
+simple
+BEGIN;
+DECLARE c1 CURSOR FOR (
+  SELECT x, y FROM t1
+);
+EXECUTE p1;
+ROLLBACK;
+----
+error parsing response from server: failed to fill whole buffer
+
+# PREPARE FETCH with a SUBSCRIBE
+simple
+BEGIN;
+DECLARE c1 CURSOR FOR SUBSCRIBE (
+  SELECT count(1) FROM t1
+  UNION ALL
+  SELECT count(1) FROM t1
+);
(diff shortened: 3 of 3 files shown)


src/adapter/src/util.rs
```diff
@@ -222,46 +222,20 @@ pub fn index_sql(
...
-            {
-                Some(mut desc) => {
-                    // Parameters are already bound to the portal and will not be accepted through
-                    // FETCH.
-                    desc.param_types = Vec::new();
-                    Ok(desc)
-                }
-                None => Err(AdapterError::UnknownCursor(name.to_string())),
-            }
-        }
```
ggevay, line 253: This is the code that is moved into `describe_fetch`, so that it's no longer a special case here.

src/adapter/src/util.rs
```diff
@@ -222,46 +222,20 @@ pub fn index_sql(
...
-/// through the session.
 pub fn describe(
     catalog: &Catalog,
     stmt: Statement<Raw>,
     param_types: &[Option<ScalarType>],
     session: &Session,
 ) -> Result<StatementDesc, AdapterError> {
-    match stmt {
-        // FETCH's description depends on the current session, which describe_statement
-        // doesn't (and shouldn't?) have access to, so intercept it here.
```
ggevay, line 237: I think this comment pre-dates `SessionCatalog`/`ConnCatalog`.

src/adapter/src/catalog.rs
```diff
@@ -1796,6 +1797,12 @@ impl SessionCatalog for ConnCatalog<'_> {
             .flatten()
     }
 
+    fn get_portal_desc_unverified(&self, portal_name: &str) -> Option<&StatementDesc> {
+        self.portals
+            .map(|portals| portals.get(portal_name).map(|portal| &portal.desc))
+            .flatten()
```
antiguru: Isn't this:
```suggestion
            .and_then(|portals| portals.get(portal_name).map(|portal| &portal.desc))
```

ggevay: Yes, thanks!

src/adapter/src/error.rs
```diff
@@ -890,7 +890,10 @@ impl From<mz_sql::catalog::CatalogError> for AdapterError {
 
 impl From<PlanError> for AdapterError {
     fn from(e: PlanError) -> AdapterError {
-        AdapterError::PlanError(e)
+        match e {
+            PlanError::UnknownCursor(name) => AdapterError::UnknownCursor(name),
+            _ => AdapterError::PlanError(e),
+        }
```
antiguru, line 896: I think we handle this in `AdapterError::code` elsewhere. Might be good to do it consistently instead of coercing the error here.

ggevay: I did it in a different way for this error because `PlanError::UnknownCursor` has an equivalent `AdapterError::UnknownCursor`. In contrast, `PlanError::InvalidSchemaName`, `PlanError::ColumnAlreadyExists`, etc. (whose error codes are handled in `AdapterError::code`) don't have equivalent `AdapterError`s. (Both `PlanError::UnknownCursor` and `AdapterError::UnknownCursor` has to exist, because we sometimes run into this error in planning, and sometimes in other Adapter code.)

Also note that there is precedent for turning lower-level errors into equivalent higher-level errors in a `From` impl:
`impl From<OptimizerError> for AdapterError {`
does this e.g. when turning `OptimizerError::RecursionLimitError` into `AdapterError::RecursionLimit`.

PR #33472, opened 2025-08-30
*: disallow using `Iterator::zip`

petrosagg:
### Motivation

As discussed recently with the cluster team, `Iterator::zip` can be a footgun when early stopping isn't anticipated or desirable. This PR marks the method as disallowed and switches all uses to `Itertools::zip_eq` which panics if the two iterators are not the same length.

During this process one bug was found! The fix is in the first commit.

The subsequent commits are in descending order of "boring", with the last two being simply adding the required `use itertools::Itertools` statements and adding the deny rule. Those are also quite a bit of the diff so I highly recommend reviewing commit by commit.

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5535,6 +5535,7 @@ dependencies = [
  "chrono",
  "dec",
  "half 2.6.0",
+ "itertools 0.14.0",
  "mz-ore",
  "mz-repr",
  "num-traits",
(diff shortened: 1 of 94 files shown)


def- (approved):
Test changes lgtm

antiguru (approved):
Looks good, and has surprisingly few special cases!

PR #33474, opened 2025-08-31
Check session state in `verify_statement_revision`

ggevay:
This is on top of https://github.com/MaterializeInc/materialize/pull/33471, because of using `PREPARE FETCH` for easier testing.

This fixes https://github.com/MaterializeInc/database-issues/issues/9634: It adds a `state_revision` field to `Session`, and uses this in `verify_statement_revision` in a similar way as `catalog_revision`. We do this for both prepared statements and portals, as the bug can manifest for both, see the added tests.

Note that the original bug was for prepared statements, but portals have the same problem. This PR fixes both. It was natural to fix in the same PR, because `verify_statement_revision` overlaps between the two code paths (`verify_prepared_statement` and `verify_portal`).

A somewhat inelegant part is `get_portal_unverified_mut`, where I needed to add a warning to callers on how not to use the returned mut ref. An alternative would be to introduce interior mutability: if we made those fields of `Portal` mutable that are currently being manipulated by the callers of `get_portal_unverified_mut`, then all these callers could work with a non-mut ref, which would make it impossible to accidentally change the sensitive fields of `Portal` (`stmt` and `desc`). I am open to making this change if a reviewer things it should be done.

The PR also adds some sanity checks in the last commit, so that we catch problems similar to the above bug before we send out malformed pgwire responses.

I've also run @antiguru's repro (the cluster-spec-sheet). What happens after the PR is that we error out those statements that become invalid, and then the client library prints "Retryable error: cached plan must not change result type, reconnecting...", and continues normally.

Nightly: https://buildkite.com/materialize/nightly/builds/13024

### Checklist

### DIFF
--- a/src/adapter/src/client.rs
+++ b/src/adapter/src/client.rs
@@ -584,12 +584,14 @@ impl SessionClient {
 
         let desc = Coordinator::describe(&catalog, self.session(), stmt.clone(), param_types)?;
         let now = self.now();
+        let state_revision = self.session().state_revision();
         self.session().set_prepared_statement(
             name,
             stmt,
             sql,
             desc,
             catalog.transient_revision(),
+            state_revision,
             now,
         );
         Ok(())
@@ -611,6 +613,7 @@ impl SessionClient {
         let result_formats = vec![mz_pgwire_common::Format::Text; desc.arity()];
         let now = self.now();
         let logging = self.session().mint_logging(sql, Some(&stmt), now);
+        let session_state_revision = self.session().state_revision();
         self.session().set_portal(
             name,
             desc,
@@ -619,6 +622,7 @@ impl SessionClient {
             params,
             result_formats,
             catalog.transient_revision(),
+            session_state_revision,
         )?;
         Ok(())
     }
--- a/src/adapter/src/client.rs
+++ b/src/adapter/src/client.rs
@@ -56,7 +56,7 @@ use crate::error::AdapterError;
 use crate::metrics::Metrics;
 use crate::optimize::{self, Optimize};
 use crate::session::{
-    EndTransactionAction, PreparedStatement, Session, SessionConfig, TransactionId,
+    EndTransactionAction, PreparedStatement, Session, SessionConfig, StateRevision, TransactionId,
 };
 use crate::statement_logging::{StatementEndedExecutionReason, StatementExecutionStrategy};
 use crate::telemetry::{self, EventDetails, SegmentClientExt, StatementFailureType};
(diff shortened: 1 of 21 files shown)


src/adapter/src/client.rs
```diff
@@ -584,12 +584,14 @@ impl SessionClient {
...
         let desc = Coordinator::describe(&catalog, self.session(), stmt.clone(), param_types)?;
         let now = self.now();
+        let state_revision = self.session().state_revision();
         self.session().set_prepared_statement(
             name,
             stmt,
             sql,
             desc,
             catalog.transient_revision(),
+            state_revision,
```
antiguru: It seems all uses of `set_prepared_statement`/`set_portal` use the current state revision of the session. Could we remove the parameter and just let the session handle the revision internally?

src/adapter/src/coord/sql.rs
```diff
@@ -150,28 +153,33 @@ impl Coordinator {
...
         session: &Session,
         stmt: Option<&Statement<Raw>>,
         desc: &StatementDesc,
-        catalog_revision: u64,
-    ) -> Result<Option<u64>, AdapterError> {
-        let current_revision = catalog.transient_revision();
-        if catalog_revision != current_revision {
+        old_catalog_revision: u64,
+        old_session_state_revision: u64,
+    ) -> Result<Option<(u64, u64)>, AdapterError> {
```
antiguru: I'd lean toward wrapping the `(u64, u64)` in a struct. It's easy to accidentally swap the values.

src/adapter/src/client.rs
ggevay: Hmm, yes, will do!

src/adapter/src/session.rs
```diff
@@ -749,6 +768,10 @@ impl<T: TimestampManipulation> Session<T> {
     /// Retrieves a mutable reference to the specified portal.
     ///
     /// If there is no such portal, returns `None`.
+    ///
```
aljoscha, line 767: Yeah, this does feel a bit brittle, as you mentioned. 🙈 I imagine it's hard to get rid of the mutable accessor altogether? Say, change all call sites to use purpose-built methods to modify or get portal state?

src/adapter/src/client.rs
ggevay: Actually, it turns out that it was a bug that I was always passing the current session revision to `set_portal`! This is because the portals are sometimes based on a prepared statement, and in that case we need to take the revision of that statement, rather than the current revision, because that is the revision that the statement's desc was verified, and we are taking the statement's and putting it on the portal. (This is also what we do for the catalog revision.)

For `set_prepared_statement`, we could indeed remove the parameter, but now I'm scared that maybe we add such a call to it in the future where using the current session state revision won't be correct. So, I think it's better to make the state revision explicit, i.e., to not remove the parameter also here.

src/adapter/src/coord/sql.rs
ggevay: Done. Indeed, the code looks cleaner this way! Btw. it also makes the [above](https://github.com/MaterializeInc/materialize/pull/33474#discussion_r2313928049) bug quite obvious, because I was taking the two kinds of revisions from different places.

src/adapter/src/session.rs
ggevay: > I imagine it's hard to get rid of the mutable accessor altogether? 

It would be a non-trivial refactoring, and would also make the code slightly slower, because of needing to do portal lookups a few more times.

But I think I've found a good solution to this dilemma with just a small code change, see the "Control mutable access to portal state" commit.

ggevay:
TFTR!

PR #33492, opened 2025-09-02
MALLOC_CONF for clusterd

antiguru:
Exposes a feature flag `clusterd_malloc_conf` that conditionally sets the `MALLOC_CONF` environment variable for clusterd. An empty value maps to `None`, so no env symbol will be set.

This allows us to experiment with interesting settings, specifically passing `thp:always` to enable hughe pages.


def- (commented):
Since you've set thp:always in CI, I have triggered a nightly run to see what effect it has: https://buildkite.com/materialize/nightly/builds/13056

PR #33496, opened 2025-09-02
[persist] Configurable logging / auditing for filter pushdown

bkirwi:
### Motivation

Trying to narrow down the location of a filter-pushdown issue.

### Tips for reviewer

These logs would have been pretty hazardous to produce before, but ought to be okay with the new redaction.

### Checklist


bkirwi:
> Makes me wonder if we have any mechanisms to make sure dyncfgs move in lockstep with each other?

There are - three, I guess:
1. We'll often express the relationship in the code - only enabling a behaviour if a flag _and_ its prerequisite are set.
2. LaunchDarkly has a concept of prerequisites, apparently. I've never used it.
3. Sometimes we'll just be careful.

Since I could imagine a case where I'd want to enable this flag independently, I decided to go with strategy 3 here. But I think all of them are plausible for different situations...

PR #33499, opened 2025-09-03
TopK: avoid mapping errors when limit is constant

antiguru:
We used to always render an operator to map the error collection, even if we could determine that the limit expression could never error. This change adjusts rendering to only render error handling if the expression is not a literal, or evaluates to an error itself.


antiguru:
Thanks for the review!

PR #33500, opened 2025-09-03
Avoid deconstructing temporal filters in ANF

ggevay:
Fixes https://github.com/MaterializeInc/database-issues/issues/9643

draft

### Motivation

  * This PR fixes a recognized bug: https://github.com/MaterializeInc/database-issues/issues/9643

### Checklist


src/expr/src/linear.rs
```diff
@@ -1913,7 +1935,15 @@ pub mod plan {
...
                     }
+                    soft_assert_or_log!(
+                        is_temporal_filter,
+                        "MfpPlan::create_from and is_temporal_filter should agree"
+                    );
                 } else {
+                    soft_assert_or_log!(
+                        !is_temporal_filter,
+                        "MfpPlan::create_from and is_temporal_filter should agree"
+                    );
```
ggevay: I'm also open to refactoring these into one method if deemed worthwhile.

src/expr/src/scalar.rs
```diff
@@ -728,6 +728,39 @@ impl MirScalarExpr {
...
+            if matches!(func, Eq | Gt | Gte | Lt | Lte) {
+                let expr1_mz_now = matches!(
+                    **expr1,
+                    MirScalarExpr::CallUnmaterializable(UnmaterializableFunc::MzNow)
+                );
+                let expr2_mz_now = matches!(
+                    **expr2,
+                    MirScalarExpr::CallUnmaterializable(UnmaterializableFunc::MzNow)
+                );
+                let expr1_contains_temporal = expr1.contains_temporal();
```
mgree: `MRE::contains_temporal` fully traverses the term---we can save some of this traversal and simplify the code with something like:

```rust
if MirScalarExpr::CallUnmaterializable(UnmaterializableFunc::MzNow) == **expr1 && !expr2.contains_temporal() {
  return Some(expr2);
}
if MirScalarExpr::CallUnmaterializable(UnmaterializableFunc::MzNow) == **expr2 && !expr1.contains_temporal() {
  return Some(expr1);
}
// fall through to `None` below
```

src/expr/src/linear.rs
```diff
@@ -1913,7 +1935,15 @@ pub mod plan {
                             ));
                         }
                     }
+                    soft_assert_or_log!(
+                        is_temporal_filter,
```
mgree: This is the only one that's not negated, possibly worth calling attention with a comment.

src/expr/src/linear.rs
mgree: Eh, seems fine as is. Maybe worth adding a comment that temporal filters are necessarily binary operations?

mgree (approved):
Minor nits---looks good to me!

src/expr/src/scalar.rs
ggevay: Thanks, changed to the following:
```
                let mz_now = MirScalarExpr::CallUnmaterializable(UnmaterializableFunc::MzNow);
                if **expr1 == mz_now && !expr2.contains_temporal() {
                    return Some(expr2);
                }
                if **expr2 == mz_now && !expr1.contains_temporal() {
                    return Some(expr1);
                }
```
which is indeed significantly shorter than the original version.

src/expr/src/linear.rs
ggevay: Done

src/expr/src/linear.rs
ggevay: > Eh, seems fine as is.

Ok, I'm relieved :)

> Maybe worth adding a comment that temporal filters are necessarily binary operations?

Below the `let is_temporal_filter = ` line, there is the following old comment:
```
// Supported temporal predicates are exclusively binary operators.
```
Or did you mean in the doc comment of the `fn is_temporal_filter`? That kinda has it already, albeit somewhat implicitly, right?

frankmcsherry:
To discuss: why not use `MapFilterProject::extract_temporal`, and ultimately `SafeMfpPlan` to go from MFPs containing temporal operators to those with and without. (( i.e. I can't tell why the approach here is the right solution, vs using existing methods ))

src/expr/src/scalar.rs
```diff
@@ -728,6 +728,27 @@ impl MirScalarExpr {
         )
     }
 
+    /// Checks whether the expression is a standard temporal filter, that is:
+    /// - mz_now() by itself on one side of an Eq | Gt | Gte | Lt | Lte comparison, and
+    /// - no mz_now() anywhere inside the other side of the comparison.
+    ///
+    /// If yes, then it returns the other side of the comparison.
+    pub fn is_temporal_filter(&mut self) -> Option<&mut MirScalarExpr> {
```
frankmcsherry: Nit, but naming-wise I would expect `is_foo(&mut self) -> ...` to return a bool and not be `&mut`. Perhaps instead `as_mut_temporal_filter_expr`? That's a bit horrid, but probably clearer when used and less surprising.

src/expr/src/linear.rs
```diff
@@ -1863,6 +1877,10 @@ pub mod plan {
                             BinaryFunc::Gt => BinaryFunc::Lt,
                             BinaryFunc::Gte => BinaryFunc::Lte,
                             x => {
+                                soft_assert_or_log!(
+                                    !is_temporal_filter,
+                                    "MfpPlan::create_from and is_temporal_filter should agree"
+                                );
```
frankmcsherry: I don't understand the intent of this test. We are returning an error due to an unsupported temporal operation; why does that case need a test? It seems like maybe this is just getting some test coverage in for the correctness of the `is_temporal_filter()` logic, but .. it feels like sprinkling that in seemingly unrelated branches is just confusing.

src/expr/src/scalar.rs
frankmcsherry: Reading the other changes, I think what we actually want here is
```
/// If `self` expresses a temporal filter, normalize it to start with `mz_now()` and return references.
/// 
/// A temporal filter is an expression of the form `mz_now() <BINOP> <EXPR>`,
/// for a restricted set of `BINOP` and `EXPR` that do not themselves contain `mz_now()`.
/// Expressions may conform to this once their expressions are swapped.
///
/// If the expression is not a temporal filter it will be unchanged.
pub fn as_mut_temporal_filter(&mut self) -> Option<(&BinaryFunc, &mut MirScalarExpr)>
```
which I think centralizes the logic in one place, rather than have two flavors and tests to make sure they stay the same.

ggevay: Yes, ok, this is what I meant in https://github.com/MaterializeInc/materialize/pull/33500/files#r2318981605
If you deem this refactoring worthwhile, then I'll do it.

src/expr/src/linear.rs
ggevay: (I ended up factoring this out, due to Frank's comment.)

src/expr/src/scalar.rs
ggevay: Done, see the latest commit.

src/expr/src/linear.rs
ggevay: This was now removed as part of the refactoring.

ggevay:
Thanks for the review! (The Nightly fail looks unrelated.)

PR #33507, opened 2025-09-03
sql: Unicode normalization

ohbadiah:
Mirrors the postgres normalize function. Provides a unary function defaulting to the NFC form, as well as a binary variant taking the form to use (one of NFC, NFD, NFKC, or NFKD).

For details, see:
- [Unicode docs](https://unicode.org/reports/tr15/#Norm_Forms)
- [Postgres](https://www.postgresql.org/docs/current/functions-string.html)

### Motivation

See [issue 7930](https://github.com/MaterializeInc/database-issues/issues/7930). Postgres has this, and we want it too.


ohbadiah:

src/sql-parser/src/parser.rs
```diff
@@ -1054,6 +1055,44 @@ impl<'a> Parser<'a> {
         }))
     }
 
+    // Parse calls to normalize(), which can take the form:
+    // - normalize('string')
+    // - normalize('string', NFC)
+    // - normalize('string', NFD)
+    // - normalize('string', NFKC)
+    // - normalize('string', NFKD)
```
antiguru:
```suggestion
    /// Parse calls to normalize(), which can take the form:
    /// - normalize('string')
    /// - normalize('string', NFC)
    /// - normalize('string', NFD)
    /// - normalize('string', NFKC)
    /// - normalize('string', NFKD)
```

src/pgrepr-consts/src/oid.rs
```diff
@@ -110,6 +110,7 @@ pub const FUNC_PG_DIGEST_STRING: u32 = 12001;
 pub const FUNC_PG_DIGEST_BYTES: u32 = 12002;
 pub const FUNC_PG_HMAC_STRING: u32 = 12003;
 pub const FUNC_PG_HMAC_BYTES: u32 = 12004;
+pub const FUNC_NORMALIZE_OID: u32 = 4350;
```
ggevay: The comment above this block says

> We get to choose whatever OIDs we like for these builtins.

But I think this is not true for `normalize`. I see that you chose the same OID that Postgres has for this, so the number is correct. It's just that it should be placed somewhere else.

src/sql-parser/src/parser.rs
```diff
@@ -1054,6 +1055,44 @@ impl<'a> Parser<'a> {
...
+        let args = if self.consume_token(&Token::Comma) {
+            // Parse the normalization form (NFC, NFD, NFKC, or NFKD)
+            let form = match self.expect_one_of_keywords(&[NFC, NFD, NFKC, NFKD]) {
+                Ok(kw) => kw.as_str(),
+                Err(_) => {
+                    return self.expected(
+                        self.peek_pos(),
+                        "one of NFC, NFD, NFKC, NFKD",
+                        self.peek_token(),
+                    );
```
ggevay: You can just do
```
let form = self.expect_one_of_keywords(&[NFC, NFD, NFKC, NFKD])?.as_str();
```
which produces a very similar error automatically.

src/expr/src/scalar.proto
```diff
@@ -596,6 +603,7 @@ message ProtoBinaryFunc {
     google.protobuf.Empty position = 114;
     google.protobuf.Empty right = 115;
     google.protobuf.Empty repeat_string = 116;
+    google.protobuf.Empty normalize = 300;
```
ggevay: nit: Is there a reason for not simply adding it at the end of the list, and with the next available number (200)?

ggevay (approved):
Looks good to me! I have only some very minor comments.

Also, CI is failing. You can fix the cargo test fail by
```
REWRITE=1 METADATA_BACKEND_URL=postgres://root@localhost:26257 cargo test test_http_sql
```
and fix the slt fails by
```
bin/sqllogictest -- --rewrite-results test/sqllogictest/oid.slt test/sqllogictest/mz_catalog_server_index_accounting.slt
```

src/pgrepr-consts/src/oid.rs
ohbadiah: Good catch, thanks.

src/sql-parser/src/parser.rs
ohbadiah: Yes, that's an improvement!

src/expr/src/scalar.proto
ohbadiah: Much better! Though this problem of picking the next integer in an unordered list is a little bothersome. I don't suppose there's an easy thing to run to get the next int? Perhaps just looking through the git log is sufficient.

src/sql-parser/src/parser.rs
ohbadiah: Thanks

src/expr/src/scalar.proto
ggevay: I think we often add new variants at the end of the list. If this were always the case, then it would be enough to look at the end of the list. But sometimes we add it elsewhere, and in that case `git blame` can indeed help!

PR #33508, opened 2025-09-03
Persist: clarify `reduced` semantics to refer only to spine structure

DAlperin:
The exert function is called exclusively by the catalog force compaction code path. In the case that there was only one spine batch with multiple hollow batches, the exert function would incorrectly introduce empty SpineBatches. The behavior of the `reduced` function was always using the number of HollowBatches as a proxy for the structural shape of the spine. With incremental compaction, there might be only one run with a number of empty HollowBatches. This is totally legal, and not a reason to try and force structural merges.

### Checklist

### DIFF
--- a/src/persist-client/src/internal/trace.rs
+++ b/src/persist-client/src/internal/trace.rs
@@ -1744,16 +1744,22 @@ impl<T: Timestamp + Lattice> Spine<T> {
         self.introduce_batch(batch, usize::cast_from(index.trailing_zeros()), log);
     }
 
-    /// True iff there is at most one HollowBatch in `self.merging`.
+    /// True iff there is at most one SpineBatch across all layers.
     ///
-    /// When true, there is no maintenance work to perform in the trace, other
-    /// than compaction. We do not yet have logic in place to determine if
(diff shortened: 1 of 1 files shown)


src/persist-client/src/internal/trace.rs
```diff
@@ -1744,16 +1744,22 @@ impl<T: Timestamp + Lattice> Spine<T> {
...
-        self.spine_batches()
-            .flat_map(|b| b.parts.as_slice())
-            .count()
-            < 2
+        // Consider the trace "reduced" when there is at most one SpineBatch across all
+        // layers, regardless of how many inner hollow parts that batch contains. Inner
+        // parts can be cleaned up by compaction, which we intentionally ignore here to
+        // avoid repeatedly injecting empty batches when only structure-level compaction
+        // remains.
+        self.spine_batches().count() < 2
```
bkirwi: Thinking about this again from scratch, it feels like the "real criteria" for whether there's any real work to do would be whether there are <2 _runs_, like we do in `is_compact`.

Anyways I think this is probably an improvement, but consider that change pre-approved if it tests well!

PR #33509, opened 2025-09-03
Persist: only use PartialBatch CompactionInput when really required

DAlperin:
If we are replacing a single batch, with a single run, we can just replace the whole thing. No need to take the (slightly) more expensive partial batch replacement codepath.

### Checklist

### DIFF
--- a/src/persist-client/src/internal/compact.rs
+++ b/src/persist-client/src/internal/compact.rs
@@ -695,10 +695,21 @@ where
                         .collect::<BTreeSet<_>>();
                     match batch_ids.iter().exactly_one().ok() {
                         Some(batch_id) => {
-                            CompactionInput::PartialBatch(
-                                *batch_id,
-                                run_ids
-                            )
+                            // We are compacting runs from exactly one batch. Decide whether this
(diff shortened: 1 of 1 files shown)


src/persist-client/src/internal/compact.rs
```diff
@@ -694,13 +694,13 @@ where
...
                     match batch_ids.iter().exactly_one().ok() {
-                        Some(batch_id) => {
+                        Some(batch_id) if run_ids.len() > 1 => {
                             CompactionInput::PartialBatch(
                                 *batch_id,
                                 run_ids
                             )
                         }
-                        None => input_id_range(batch_ids),
+                        _ => input_id_range(batch_ids),
```
bkirwi: Mmmm... IIUC it's only safe to punt down to this codepath for whole batches, but our check above only asserts that the current chunk has 1 run or less.

Is the idea that chunks with one run or less must be full batches? That may be true currently but seems a bit delicate...

DAlperin: You're right. I was relying on some invariants of chunk_runs but I can imagine some cases where that might fall apart in the future. I've replaced the logic with a more explicit check

PR #33513, opened 2025-09-04
adapter,ore: some smaller cleanups

aljoscha:


src/ore/src/id_gen.rs
```diff
@@ -326,6 +326,15 @@ mod internal {
         stored: u32,
     }
 
+    impl<T: Debug, A: IdAllocatorInner> Debug for IdHandleInner<T, A> {
+        fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+            f.debug_struct("IdHandleInner")
+                .field("id", &self.id)
+                .field("stored", &self.stored)
+                .finish()
```
antiguru:
```suggestion
                .finish_non_exhaustive()
```
Maybe?

antiguru (approved):

src/ore/src/id_gen.rs
aljoscha: 👌

aljoscha:
tyty! 🙇‍♂️

PR #33520, opened 2025-09-04
parallel-workload: Use more interesting expressions in SELECT, views and conditions; support COPY FROM STDIN; prepared statements; run 10x faster

def-:
Motivated by https://materializeinc.slack.com/archives/CU7ELJ6E9/p1756852503595199

I always knew we'd need more general expressions, so finally got around to implementing them (and removing all the TODO comments).

That reproduces the original issue easily in 4/9 runs: https://buildkite.com/materialize/nightly/builds/13133

> +++ [worker_8] Query failed: INSERT INTO "db-pw-1757078559-0"."s-0"."t-1" ("c-0-bigint", "c-1-mz_timestamp") VALUES (-1154873474221047172, MZ_TIMESTAMP '19317-9-27'), (-9223372036854775808, MZ_TIMESTAMP '66822-9-13'), (9223372036854775807, MZ_TIMESTAMP '17235-10-21'), (-4665962949847538769, MZ_TIMESTAMP '69332-6-15'), (8444606562024686967, MZ_TIMESTAMP '71526-12-6'), (6100859744477175015, MZ_TIMESTAMP '96494-10-12'), (-9223372036854775808, MZ_TIMESTAMP '59117-4-1'), (119013262017748147, MZ_TIMESTAMP '78394-6-22'), (-6428999613294955046, MZ_TIMESTAMP '69507-9-2'), (214639980754110942, MZ_TIMESTAMP '70046-11-5'), (-9223372036854775808, MZ_TIMESTAMP '93346-6-14') RETURNING 94, abs(-22.491346699927163), abs(6921), current_timestamp(), TIME '6:7:10.991542', '{-93}'::int[], -36.12842768411257, '{"key0": "-33", "key1": "-71", "key2": "84", "key3": "-43", "key4": "-1", "key5": "85", "key6": "8", "key7": "13", "key8": "68", "key9": "6"}'::jsonb, cast((DATE '58327-7-20') as text), 1523333707; HTTP XX000: internal error: cannot evaluate unmaterializable function: CurrentTimestamp

This also found an unrelated bug: https://github.com/MaterializeInc/database-issues/issues/9656

New run with https://github.com/MaterializeInc/materialize/pull/33500 included should show us if it's fixed: https://buildkite.com/materialize/nightly/builds/13136
### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
[...]

PR #33531, opened 2025-09-06
adapter: Don't panic when COPY FROM's target table is dropped

ggevay:
It was panicking if the table was dropped at a bad moment (either while COPY FROM is reading from stdin or before the `ExecuteResponse::CopyFrom` comes back from the Coordinator).

Writing an automated test specifically for this seems complicated. I'm open to ideas, but maybe the manual testing that I did is enough, given that this is a small and straightforward change. Also, Parallel Workload would hopefully notice this again, if it regresses in the future.

(I've also manually tested (with sleeps) what happens if the table goes away after taking that catalog snapshot that the `try_get_item` looks into. In this case, the `try_get_item` still succeeds, but then we [later throw a graceful error from](https://github.com/MaterializeInc/materialize/blob/cc1d7672ff899f233badb4db65fc00168d94bb69/src/adapter/src/coord/sequencer/inner.rs#L2442) `sequence_end_transaction_inner`, which is fine. (The error msg could be improved, because now it just says `ERROR:  unknown catalog item 'u1'`, but it's probably not worth the hassle for now.))

@def-, should I add an ignore for this error msg in Parallel Workload? Or that test would fail only for a panic, and any graceful error is fine?

(Tagged Aljoscha for review, because we've been talking about similar things in the context of the frontend peek work.)

### Motivation

  * This PR fixes a recognized bug: https://github.com/MaterializeInc/database-issues/issues/9659

### Checklist

### DIFF
--- a/src/adapter/src/client.rs
+++ b/src/adapter/src/client.rs
@@ -774,8 +774,8 @@ impl SessionClient {
     /// ones.
     pub async fn insert_rows(
         &mut self,
-        id: CatalogItemId,
-        name: String,
+        target_id: CatalogItemId,
+        target_name: String,
         columns: Vec<ColumnIndex>,
         rows: Vec<Row>,
         ctx_extra: ExecuteContextExtra,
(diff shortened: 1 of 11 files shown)


def-:
> @def-, should I add an ignore for this error msg in Parallel Workload? Or that test would fail only for a panic, and any graceful error is fine?

I'll add the ignore in the PR itself: https://github.com/MaterializeInc/materialize/pull/33520 Edit: Done: https://github.com/MaterializeInc/materialize/pull/33520/commits/dacb48bcbf9311bd543e262bb02c3a61bf2d936e

ggevay:
The name we don't have here. `resolve_full_name` is the first thing we do on the catalog item if the `try_get_item` succeeds, but if the `try_get_item` fails, then we have no name.

Putting the ID in the error message would be easy, but of little use: it's unlikely the user knows what the ID means offhand, and they can't even look it up at this point, because the object is gone from places like `mz_objects`.

But I could wire the name through to this function if it's deemed valuable. It would involve adding it as a parameter to 5 function calls between `ExecuteResponse::CopyFrom` and this function, adding it to `ExecuteResponse::CopyFrom`, passing it there from `Plan::CopyFrom`, adding it to `Plan::CopyFrom`, and filling that in `plan_copy_from`.

But I'm thinking that it should be pretty clear what table we are talking about: the error msg says
"COPY FROM's target table was dropped",
and there is only one table mentioned in a COPY FROM command. So, even if the user doesn't remember which table they targeted, they can just use the up arrow to look at their command, and it should be immediately clear what was the target table. Or the name would still be valuable because of scripts doing COPY FROM? (Note that this code path handles only COPY FROM STDIN; The half-finished COPY FROM S3 stuff uses a completely different code path, so this error won't appear in that case.) What do you think? Edit: Went ahead an added the table name.

ggevay:
I went ahead and added the table name to the error msg. I suddenly remembered that I often get annoyed when an error msg leaves out some interesting information, so I don't want to be that guy who produces such an error msg :sweat_smile: 

Plus a bonus refactoring in a separate commit: I renamed id and name to target_id and target_name in the COPY FROM code.

def-:
I have included the change in https://github.com/MaterializeInc/materialize/pull/33520, no further occurrences!

PR #33533, opened 2025-09-06
`prep_relation_expr` earlier in peek sequencing

ggevay:
This is fixing https://github.com/MaterializeInc/database-issues/issues/9645 by moving unmaterializable function evaluation (`prep_relation_expr`/`prep_scalar_expr`) earlier in peek sequencing. It also does the same move in COPY TO sequencing. Even though fast path is not a thing there, so the above issue can't occur, we'd still like to avoid unnecessary divergence between the peek sequencing and the COPY TO sequencing codes.

Also note that even independently from https://github.com/MaterializeInc/database-issues/issues/9645, it seems generally better to do `mz_now()` (and other unmaterializable func) evaluation before optimization, because constants are easier to work with during optimization than function calls. This is demonstrated by a few (somewhat contrived) tests flipping from slow path to fast path peeks.

The cause of https://github.com/MaterializeInc/database-issues/issues/9645 was a somewhat weird interaction between various things:
1. `create_fast_path_plan` is able to handle an MFP on top of a Get, but is not able to handle an MFP on top of a Constant. This limitation is understandable, give that we usually expect FoldConstants to make an MFP on top of a Constant disappear.
2. In the above issue, an MFP on top of a Constant was not disappearing, because it was containing an unmaterializable function call, which FoldConstants can't handle.
3. We have a special fast path optimizer, which we use during peek sequencing when a peek looks like a fast path peek between local and global optimization, so that we don't have to run the full global optimization pipeline. The soft assert that was tripped in the above issue is meant to verify that if a peek looked like a fast path peek before running the fast path optimizer, then we continue to be able to plan the peek as fast path also after running the fast path optimizer. The reason this got violated here is because the plan was an MFP on top of a Get before running the fast path optimizer, but was an MFP on top of a Constant after running the fast path optimizer.

After the PR moved unmaterializable function evaluation before (global or fast path) optimization, the fast path optimizer now doesn't leave the MFP on top of the Constant, because the MFP no longer has an unmaterializable function call at that point, so FoldConstants folds away the MFP. (The MFP's result is folded into a new Constant node.)

(Philosophically, the above 1. is a _non-monotonicity_ in the optimizer, which is not good: Generally, it's good if optimization code has the behavior that if `plan'` is a better plan than `plan`, then `optimize(plan')` is a better plan than `optimize(plan)`. As per 1., `create_fast_path_plan` is an optimization that doesn't have this property: an MFP on top of a constant is a better plan than an MFP on top of a Get, but `create_fast_path_plan` makes the former into a worse plan than the latter. Such non-monotonicities in the optimizer often lead to unpleasant surprises: if optimization `O2` has a non-monotonicity, and we have an optimizer pipeline `O1 -> O2`, and we make a change to `O1`, then even if that change is a strict improvement _locally_ for `O1`, the change can still regress the `O1 -> O2` pipeline. So, it would be good to eliminate the non-monotonicity here by making `create_fast_path_plan` able to handle an MFP on top of a Constant, but I consider this out of scope for this PR. I opened a separate issue: https://github.com/MaterializeInc/database-issues/issues/9665)

This PR has a slightly annoying side-effect on slt testing: if an slt has an EXPLAIN OPTIMIZED PLAN on a peek that involves mz_now, the explain output will now show the evaluation result of the mz_now. This is a problem in tests, because mz_now changes at every slt run. I worked around this problem in two ways:
1. Made some tests create a materialize view instead of a peek. `mz_now` is not evaluated in materialized view plans. (`filter-pushdown.slt`)
2. Added a division to `mz_now` by 100000000000000 to make it always 0. I used this workaround instead of 1. in `persist-fast-path.slt`, because it was essential here for the test to remain a peek, since persist fast path is not relevant for materialized views.

@bkirwi may I ask you to check the test changes in `filter-pushdown.slt` and `persist-fast-path.slt` (explained in the previous paragraph) that I didn't accidentally invalidate any tests there?

@aljoscha, I tagged you for the overall PR review, because it's a peek sequencing thing. But if you'd like to avoid delving into optimization stuff at this moment, then let me know, and then I'll tag Michael instead.

Nightly: https://buildkite.com/materialize/nightly/builds/13207
New Nightly after fixing the problem in the 4-replica slt: https://buildkite.com/materialize/nightly/builds/13209

### Motivation

  * This PR fixes a recognized bug: https://github.com/MaterializeInc/database-issues/issues/9645

### Checklist

### DIFF
--- a/src/adapter/src/optimize/copy_to.rs
+++ b/src/adapter/src/optimize/copy_to.rs
@@ -300,6 +300,25 @@ impl<'s> Optimize<LocalMirPlan<Resolved<'s>>> for Optimizer {
         // Set the `as_of` and `until` timestamps for the dataflow.
         df_desc.set_as_of(timestamp_ctx.antichain());
 
+        // Get the single timestamp representing the `as_of` time.
+        let as_of = df_desc
+            .as_of
+            .clone()
+            .expect("as_of antichain")
+            .into_option()
+            .expect("unique as_of element");
+
+        // Resolve all unmaterializable function calls including mz_now().
+        let style = ExprPrepStyle::OneShot {
+            logical_time: EvalTime::Time(as_of),
+            session,
+            catalog_state: self.catalog.state(),
+        };
+        df_desc.visit_children(
+            |r| prep_relation_expr(r, style),
+            |s| prep_scalar_expr(s, style),
+        )?;
+
         // Use the opportunity to name an `until` frontier that will prevent
         // work we needn't perform. By default, `until` will be
         // `Antichain::new()`, which prevents no updates and is safe.
@@ -335,25 +354,6 @@ impl<'s> Optimize<LocalMirPlan<Resolved<'s>>> for Optimizer {
             // Collect the list of indexes used by the dataflow at this point.
             trace_plan!(at: "global", &df_meta.used_indexes(&df_desc));
         }
-
-        // Get the single timestamp representing the `as_of` time.
-        let as_of = df_desc
-            .as_of
-            .clone()
-            .expect("as_of antichain")
-            .into_option()
-            .expect("unique as_of element");
-
-        // Resolve all unmaterializable function calls including mz_now().
-        let style = ExprPrepStyle::OneShot {
-            logical_time: EvalTime::Time(as_of),
-            session,
-            catalog_state: self.catalog.state(),
-        };
-        df_desc.visit_children(
-            |r| prep_relation_expr(r, style),
-            |s| prep_scalar_expr(s, style),
-        )?;
 
         // Ensure all expressions are normalized before finalizing.
         for build in df_desc.objects_to_build.iter_mut() {
--- a/src/adapter/src/optimize/peek.rs
+++ b/src/adapter/src/optimize/peek.rs
@@ -295,6 +295,25 @@ impl<'s> Optimize<LocalMirPlan<Resolved<'s>>> for Optimizer {
         // Set the `as_of` and `until` timestamps for the dataflow.
         df_desc.set_as_of(timestamp_ctx.antichain());
 
+        // Get the single timestamp representing the `as_of` time.
+        let as_of = df_desc
+            .as_of
+            .clone()
+            .expect("as_of antichain")
+            .into_option()
+            .expect("unique as_of element");
+
+        // Resolve all unmaterializable function calls including mz_now().
+        let style = ExprPrepStyle::OneShot {
+            logical_time: EvalTime::Time(as_of),
+            session,
+            catalog_state: self.catalog.state(),
+        };
+        df_desc.visit_children(
+            |r| prep_relation_expr(r, style),
+            |s| prep_scalar_expr(s, style),
+        )?;
+
         // Use the opportunity to name an `until` frontier that will prevent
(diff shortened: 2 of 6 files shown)


aljoscha:
Thanks for asking, but I'll try and stay away from optimizer business 😅

ggevay:
Ok, then tagging @mgree instead for the overall review :sweat_smile:

def-:
I have included the change in https://github.com/MaterializeInc/materialize/pull/33520, no further occurrences!

bkirwi:
Persist test changes seem fine to me - thanks for the explanation!

src/adapter/src/optimize/copy_to.rs
```diff
@@ -300,6 +300,25 @@ impl<'s> Optimize<LocalMirPlan<Resolved<'s>>> for Optimizer {
...
+        // Get the single timestamp representing the `as_of` time.
+        let as_of = df_desc
+            .as_of
+            .clone()
+            .expect("as_of antichain")
+            .into_option()
+            .expect("unique as_of element");
+
+        // Resolve all unmaterializable function calls including mz_now().
+        let style = ExprPrepStyle::OneShot {
```
mgree, line 312: Could it be the case that `prep_relation_expr` or `prep_scalar_expr` could care about `until`? If so, we should move the following block before this.

src/adapter/src/optimize/peek.rs
```diff
@@ -295,6 +295,25 @@ impl<'s> Optimize<LocalMirPlan<Resolved<'s>>> for Optimizer {
...
+        let style = ExprPrepStyle::OneShot {
+            logical_time: EvalTime::Time(as_of),
+            session,
+            catalog_state: self.catalog.state(),
+        };
+        df_desc.visit_children(
+            |r| prep_relation_expr(r, style),
+            |s| prep_scalar_expr(s, style),
+        )?;
+
```
mgree, line 316: Same concern about `until` here.

test/sqllogictest/persist-fast-path.slt
```diff
@@ -195,15 +195,15 @@ EOF
 # an arbitrary number of records to find one that matches.
 
 query T multiline
-EXPLAIN OPTIMIZED PLAN WITH (humanized expressions) AS VERBOSE TEXT FOR SELECT * from numbers WHERE value > mz_now() LIMIT 10;
+EXPLAIN OPTIMIZED PLAN WITH (humanized expressions) AS VERBOSE TEXT FOR SELECT * from numbers WHERE value > mz_now()::text::int8 / 100000000000000 LIMIT 10;
```
mgree, line 198: Took me a second to notice the case here inducing the type change below. Looks good.

mgree (approved):
Just one question about timestamps, otherwise LGTM!

src/adapter/src/optimize/copy_to.rs
ggevay: Fortunately, that can't happen, because we don't pass the `DataflowDescription` to `prep_relation_expr` or `prep_scalar_expr`, so they have no way of observing what `until` we have set on the `DataflowDescription`.

ggevay:
Thanks for the review!

PR #33546, opened 2025-09-09
docs: single-source postgres replication slots considerations

kay-kim:
https://preview.materialize.com/materialize/33546/ingest-data/postgres/#replication-slots
https://preview.materialize.com/materialize/33546/sql/create-source/postgres/#postgresql-replication-slots

### DIFF
--- a/doc/user/shared-content/postgres-considerations.md
+++ b/doc/user/shared-content/postgres-considerations.md
@@ -1,60 +1,12 @@
-### Schema changes
+{{< include-md file="shared-content/postgres-known-limitations.md" >}}
 
-{{< include-md file="shared-content/schema-changes-in-progress.md" >}}
+### Replication slots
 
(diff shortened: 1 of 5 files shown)


doc/user/shared-content/postgres-considerations.md
```diff
@@ -1,60 +1,12 @@
-### Schema changes
+{{< include-md file="shared-content/postgres-known-limitations.md" >}}
```
kay-kim, line 1: Moved the known-limitations into its own file.

doc/user/shared-content/postgres-known-limitations.md
```diff
@@ -0,0 +1,60 @@
+### Schema changes
```
kay-kim, line 1: Unchanged content -- just moved from the considerations.

PR #33552, opened 2025-09-10
Switch to binary func trait for some variants

antiguru:
Switch some variants of `BinaryFunc` to use the `LazyBinaryFunc` trait's implementation.

Adds missing `negate` properties to negatable binary functions.

### DIFF
--- a/src/expr/src/scalar/func.rs
+++ b/src/expr/src/scalar/func.rs
@@ -93,112 +93,94 @@ pub fn jsonb_stringify<'a>(a: Datum<'a>, temp_storage: &'a RowArena) -> Datum<'a
 
 #[sqlfunc(
     is_monotone = "(true, true)",
-    output_type = "i16",
     is_infix_op = true,
     sqlname = "+",
     propagates_nulls = true
 )]
-fn add_int16<'a>(a: i16, b: i16) -> Result<Datum<'a>, EvalError> {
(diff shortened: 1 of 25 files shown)


src/expr/src/scalar/func.rs
```diff
@@ -4063,6 +4092,14 @@ impl BinaryFunc {
...
+            BinaryFunc::AddInt32(s) => s.negate(),
+            BinaryFunc::AddInt64(s) => s.negate(),
+            BinaryFunc::AddUInt16(s) => s.negate(),
+            BinaryFunc::AddUInt32(s) => s.negate(),
+            BinaryFunc::AddUInt64(s) => s.negate(),
+            BinaryFunc::AddFloat32(s) => s.negate(),
+            BinaryFunc::AddFloat64(s) => s.negate(),
             BinaryFunc::Eq => Some(BinaryFunc::NotEq),
             BinaryFunc::NotEq => Some(BinaryFunc::Eq),
             BinaryFunc::Lt => Some(BinaryFunc::Gte),
```
ggevay, line 4093: These are done manually rather than using the `negate` annotation because their `sqlfunc`-based definitions have not yet been "activated" yet?

antiguru: Yes, that's correct.

PR #33553, opened 2025-09-10
Switch import frontier compute logging to columnar

antiguru:
Switches the import frontier logging from vectors to columnar.


antiguru:
Thanks for the review!

PR #33556, opened 2025-09-10
iceberg sink: parse and plan create iceberg sinks statements

DAlperin:
### Checklist


DAlperin:
closes https://github.com/MaterializeInc/database-issues/issues/9577

doc/user/content/sql/system-catalog/mz_catalog.md
```diff
@@ -309,6 +309,21 @@ This table was previously in the `mz_internal` schema. All queries previously re
 | `group_id_prefix`      | [`text`]       | The value of the `GROUP ID PREFIX` connection option.                                                     |
 | `topic          `      | [`text`]       | The name of the Kafka topic the source is reading from.                                                              |
 
+{{< if-released "do-not-render-until-iceberg-is-released" >}}
```
petrosagg: Does this `"do-not-render-until-iceberg-is-released"` work? I haven't seen anything similar but also not very familiar with our docs

src/catalog/src/builtin.rs
```diff
@@ -2717,9 +2735,9 @@ pub static MZ_SINKS: LazyLock<BuiltinTable> = LazyLock::new(|| {
             .with_column("envelope_type", ScalarType::String.nullable(true))
             // This `format` column is deprecated and replaced by the `key_format` and `value_format` columns
             // below. This should be removed in the future.
-            .with_column("format", ScalarType::String.nullable(false))
+            .with_column("format", ScalarType::String.nullable(true))
             .with_column("key_format", ScalarType::String.nullable(true))
-            .with_column("value_format", ScalarType::String.nullable(false))
+            .with_column("value_format", ScalarType::String.nullable(true))
```
petrosagg: You should check with the console people in case they have any asserts that this is not null somewhere

src/catalog/src/memory/objects.rs
```diff
@@ -1305,28 +1305,35 @@ impl Sink {
...
     /// Output a combined format string of the sink. For legacy reasons
     /// if the key-format is none or the key & value formats are
     /// both the same (either avro or json), we return the value format name,
     /// otherwise we return a composite name.
-    pub fn combined_format(&self) -> Cow<'_, str> {
-        let StorageSinkConnection::Kafka(connection) = &self.connection;
-        connection.format.get_format_name()
+    pub fn combined_format(&self) -> Option<Cow<'_, str>> {
+        match &self.connection {
+            StorageSinkConnection::Kafka(connection) => Some(connection.format.get_format_name()),
```
petrosagg, line 1340: Shouldn't we return the string "iceberg" or "parquet" or something here? We are encoding the data before sending them out right?

src/sql-parser/src/ast/defs/ddl.rs
```diff
@@ -670,6 +671,9 @@ impl AstDisplay for SinkEnvelope {
             Self::Debezium => {
                 f.write_str("DEBEZIUM");
             }
+            Self::Append => {
+                f.write_str("APPEND");
```
petrosagg: What are the semantics of `ENVELOPE APPEND`? Does the sink error if it sees a negative diff? Do we require that only monotonic collections are sinked? Also, the equivalent envelope name on the sources side is called `ENVELOPE NONE`. Does the append envelope do something different than envelope none?

src/sql-parser/src/ast/defs/ddl.rs
```diff
@@ -1489,18 +1536,37 @@ impl<T: AstInfo> AstDisplay for CreateSinkConnection<T> {
...
+                f.write_node(connection);
+                if !options.is_empty() {
+                    f.write_str(" (");
+                    f.write_node(&display::comma_separated(options));
+                    f.write_str(")");
+                }
+                f.write_str(" USING AWS CONNECTION ");
+                f.write_node(aws_connection);
+                if let Some(key) = key.as_ref() {
+                    f.write_node(key);
```
petrosagg, line 1553: Hm, I see that the `AstDisplay` for `SinkKey` produces a space by itself but that's not how it should be. Since you're here can you make `SinkKey` print itself without leading/trailing spaces and adjust the spaces here accordingly?

src/sql-parser/src/parser.rs
```diff
@@ -3687,12 +3749,55 @@ impl<'a> Parser<'a> {
...
+            options
+        } else {
+            vec![]
+        };
+
+        self.expect_keywords(&[USING, AWS, CONNECTION])?;
+        let aws_connection = self.parse_raw_name()?;
+
+        let key = if self.peek_keyword(KEY) {
+            let _ = self.expect_keyword(KEY);
```
petrosagg:
```suggestion
        let key = if self.parse_keyword(KEY) {
```

src/sql-parser/src/parser.rs
```diff
@@ -3687,12 +3749,55 @@ impl<'a> Parser<'a> {
...
+            let key_columns = self.parse_parenthesized_column_list(Mandatory)?;
+
+            let not_enforced = if self.peek_keywords(&[NOT, ENFORCED]) {
+                let _ = self.expect_keywords(&[NOT, ENFORCED])?;
+                true
+            } else {
+                false
+            };
```
petrosagg:
```suggestion
            let not_enforced = self.parse_keywords(&[NOT, ENFORCED]);
```

src/sql/src/plan/statement/ddl.rs
```diff
@@ -3343,69 +3347,68 @@ fn plan_sink(
...
-                let mut uniq = BTreeSet::new();
-                for col in key_columns.iter() {
-                    if !uniq.insert(col) {
-                        sql_bail!("duplicate column referenced in KEY: {}", col);
-                    }
+    let key = match &connection {
+        CreateSinkConnection::Kafka { key, .. } => key,
+        CreateSinkConnection::Iceberg { key, .. } => key,
+    };
+    let key_indices = if let Some(key) = key.clone() {
```
petrosagg: nit: this can just be:
```suggestion
    let key_indices = match &connection {
        CreateSinkConnection::Kafka { key: Some(key), .. }
        | CreateSinkConnection::Iceberg { key: Some(key), .. } => {
           // the old code
        },
        CreateSinkConnection::Kafka { key: None, .. }
        | CreateSinkConnection::Iceberg { key: None, .. } => None,
    };
```

src/sql/src/pure/error.rs
```diff
@@ -213,6 +214,30 @@ impl KafkaSinkPurificationError {
     }
 }
 
+#[derive(Debug, Clone, thiserror::Error)]
+pub enum IcebergSinkPurificationError {
+    #[error("catalog connection errored")]
```
petrosagg, line 207: I can't remember, will this automatically include the inner error's description when printed?

petrosagg: actually.. I've seen many times in MZ where the error string is printed multiple times. Maybe it's because we have enums like this where we manually include the inner error's string..

src/storage/src/sink/kafka.rs
```diff
@@ -1518,6 +1518,9 @@ fn encode_collection<G: Scope>(
                                 dbz_format(&mut row_buf.packer(), value);
                                 Some(row_buf.clone())
                             }
+                            SinkEnvelope::Append => {
+                                panic!("Kafka sink does not support ENVELOPE APPEND, this should have been checked by the planner")
```
petrosagg: maybe change this to an `unreachable!()`

doc/user/content/sql/system-catalog/mz_catalog.md
DAlperin: yeah I added an explicit hide macro for now

src/catalog/src/memory/objects.rs
DAlperin: The only place this is used is marked deprecated so I'm not sure we want to do something here? We are encoding it but "key" and "value" encodings don't quite map.
```
            // This `format` column is deprecated and replaced by the `key_format` and `value_format` columns
            // below. This should be removed in the future.`
```

src/sql-parser/src/ast/defs/ddl.rs
DAlperin: In Append only mode deletes will be a no-op. Up in the air if v0 will support this but it feels distinct from envelope none.

PR #33567, opened 2025-09-11
storage: quality of life refactors (part 1)

petrosagg:
### Motivation

This PR refactors some things in storage that have become complicated over time due to us supporting both old syntax sources and new sources syntax. It creates an explicit `OldSyntaxIngestionDescription` so that we know from the type with which case we're dealing with. Additionally the `IngestionDescription` struct is changed so that it more accurately captures the shape of the ingestion. Namely that there is always a remap shard that we write to and zero or more exports. This allows us to remove the optional `remap_shard` field from `CollectionMetdata` and makes sure the correct `CollectionMetadata` is passed everywhere. The latter change results in a lot of `remap_shard.unwrap()` calls to be removed.

### Checklist

### DIFF
--- a/src/adapter/src/coord/ddl.rs
+++ b/src/adapter/src/coord/ddl.rs
@@ -252,11 +252,11 @@ impl Coordinator {
                                     }
                                     CatalogItem::Source(source) => {
                                         sources_to_drop.push((*id, source.global_id()));
-                                        if let DataSourceDesc::Ingestion {
-                                            ingestion_desc, ..
-                                        } = &source.data_source
+                                        if let DataSourceDesc::Ingestion { desc, .. }
+                                        | DataSourceDesc::OldSyntaxIngestion { desc, .. } =
+                                            &source.data_source
                                         {
-                                            match &ingestion_desc.desc.connection {
+                                            match &desc.connection {
                                                 GenericSourceConnection::Postgres(conn) => {
                                                     let conn = conn.clone().into_inline_connection(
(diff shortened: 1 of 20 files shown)


src/storage-client/src/storage_collections.rs
```diff
@@ -2389,7 +2358,6 @@ where
                 relation_desc: collection_desc.desc.clone(),
                 data_shard,
                 // TODO(alter_table): Support changes to sources.
-                remap_shard: None,
```
martykulma, line 2392: is the comment above this still necessary?

src/adapter/src/coord/ddl.rs
```diff
@@ -252,11 +252,11 @@ impl Coordinator {
                                     }
                                     CatalogItem::Source(source) => {
                                         sources_to_drop.push((*id, source.global_id()));
-                                        if let DataSourceDesc::Ingestion {
-                                            ingestion_desc, ..
-                                        } = &source.data_source
+                                        if let DataSourceDesc::Ingestion { desc, .. }
+                                        | DataSourceDesc::OldSyntaxIngestion { desc, .. } =
+                                            &source.data_source
```
martykulma, line 257: TIL!

src/storage-client/src/storage_collections.rs
petrosagg: probably not, I'll remove in the PR at the top of the stack to avoid another CI run

PR #33569, opened 2025-09-11
add swap to mz_cluster_replicas table

jubrad:
### Checklist

### DIFF
--- a/src/adapter/src/catalog/builtin_table_updates.rs
+++ b/src/adapter/src/catalog/builtin_table_updates.rs
@@ -15,7 +15,6 @@ use std::time::Duration;
 use bytesize::ByteSize;
 use ipnet::IpNet;
 use mz_adapter_types::compaction::CompactionWindow;
-use mz_adapter_types::dyncfgs::FORCE_SWAP_FOR_CC_SIZES;
 use mz_audit_log::{EventDetails, EventType, ObjectType, VersionedEvent, VersionedStorageUsage};
 use mz_catalog::SYSTEM_CONN_ID;
 use mz_catalog::builtin::{
@@ -376,7 +375,7 @@ impl CatalogState {
         let id = cluster.replica_id(name).expect("Must exist");
         let replica = cluster.replica(id).expect("Must exist");
 
-        let (size, disk, az, internal, pending, mut swap, is_cc) = match &replica.config.location {
+        let (size, disk, az, internal, pending, swap) = match &replica.config.location {
             // TODO(guswynn): The column should be `availability_zones`, not
             // `availability_zone`.
             ReplicaLocation::Managed(ManagedReplicaLocation {
@@ -393,7 +392,6 @@ impl CatalogState {
                 *internal,
                 *pending,
                 allocation.swap_enabled,
-                allocation.is_cc,
             ),
             ReplicaLocation::Managed(ManagedReplicaLocation {
                 size,
@@ -409,15 +407,11 @@ impl CatalogState {
                 *internal,
                 *pending,
                 allocation.swap_enabled,
-                allocation.is_cc,
             ),
-            _ => (None, None, None, false, false, false, false),
+            _ => (None, None, None, false, false, false),
         };
 
         let dyncfg = self.system_config().dyncfgs();
-        if FORCE_SWAP_FOR_CC_SIZES.get(dyncfg) && is_cc {
-            swap = true;
-        }
 
         let cluster_replica_update = BuiltinTableUpdate::row(
             &*MZ_CLUSTER_REPLICAS,
--- a/src/adapter/src/catalog/open.rs
+++ b/src/adapter/src/catalog/open.rs
@@ -189,6 +189,41 @@ impl Catalog {
         assert!(!updates.is_empty(), "initial catalog snapshot is missing");
         let mut txn = storage.transaction().await?;
 
+        // Force cc sizes to use swap, if requested.
+        if FORCE_SWAP_FOR_CC_SIZES.get(state.system_configuration.dyncfgs()) {
+            info!("force-enabling swap for cc replica sizes");
+
+            for size in state.cluster_replica_sizes.0.values_mut() {
+                if size.is_cc {
+                    size.swap_enabled = true;
+                    size.cpu_exclusive = false;
+                    size.selectors.remove("materialize.cloud/scratch-fs");
+                    size.selectors
+                        .insert("materialize.cloud/swap".into(), "true".into());
+                }
+            }
+
+            // The size definitions were copied into the in-memory replica configs during
+            // `CatalogState` initialization above. We can't move this code before that because we
+            // need to wait for the `system_configuration` to be fully initialized. Instead we also
+            // patch the replica configs here.
+            for cluster in state.clusters_by_id.values_mut() {
+                for replica in cluster.replicas_by_id_.values_mut() {
+                    if let ReplicaLocation::Managed(loc) = &mut replica.config.location {
+                        let alloc = &mut loc.allocation;
+                        if alloc.is_cc {
+                            alloc.swap_enabled = true;
+                            alloc.cpu_exclusive = false;
+                            alloc.selectors.remove("materialize.cloud/scratch-fs");
+                            alloc
+                                .selectors
+                                .insert("materialize.cloud/swap".into(), "true".into());
+                        }
+                    }
+                }
+            }
+        }
+
         // Migrate/update durable data before we start loading the in-memory catalog.
         let (migrated_builtins, new_builtin_collections) = {
             migrate::durable_migrate(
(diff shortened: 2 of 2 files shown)


src/catalog/src/builtin.rs
```diff
@@ -3513,6 +3513,7 @@ pub static MZ_CLUSTER_REPLICAS: LazyLock<BuiltinTable> = LazyLock::new(|| Builti
         .with_column("availability_zone", ScalarType::String.nullable(true))
         .with_column("owner_id", ScalarType::String.nullable(false))
         .with_column("disk", ScalarType::Bool.nullable(true))
+        .with_column("swap_enabled", ScalarType::Bool.nullable(true))
```
ggevay, line 3516: It can't be null, right? Or maybe we want to actually make it null for unmanaged replicas, because we just don't have info about whether they use swap?

jubrad: Yeah, this is also where I was a bit stuck so I just copied what we did for disk.

Also I'm not sure what an unmanaged replica is... it looks like the code works fine for replicas on unmanaged clusters.

src/adapter/src/catalog/builtin_table_updates.rs
```diff
@@ -405,10 +408,17 @@ impl CatalogState {
...
+                allocation.is_cc,
             ),
-            _ => (None, None, None, false, false),
+            _ => (None, None, None, false, false, false, false),
         };
 
+        let dyncfg = self.system_config().dyncfgs();
+        if FORCE_SWAP_FOR_CC_SIZES.get(dyncfg) && is_cc {
+            swap = true;
+        }
```
ggevay: Oh, can this get out of sync with the actual setting on the cluster? We set swap only in `initialize_state`, which happens only at startup, right? So, if somebody flips this LD flag later, then it would get out of sync.

Without this, was the problem that builtin table updates are generated in `initialize_state` before the patching of `swap_enabled` on the allocations?

ggevay: There is the comment
```
            // The size definitions were copied into the in-memory replica configs during
            // `CatalogState` initialization above. We can't move this code before that because we
            // need to wait for the `system_configuration` to be fully initialized. Instead we also
            // patch the replica configs here.
```
which explains why the patching happens late in startup. But maybe we could move it just a little  bit earlier, enough to make it happen before generating the builtin table updates?  `builtin_table_updates` starts being populated in line 265, but `system_configuration` should already be fine by line 134, right? And then we wouldn't need to look at FORCE_SWAP_FOR_CC_SIZES when generating builtin table updates, because the `allocation` would be already fine.

ggevay: (Also, we should add some tests that look at the value of the `swap` column, because now I'm scared. Like, both for replicas that exist at startup, and for newly created replicas, and with varying the feature flag value maybe both before startup and after.)

antiguru:
Just pondering, but do we actually need this column? It's a departure from the other information in the table, which is all static configuration data. Swap is a runtime setting, and toggling the flag won't update the table.

ggevay:
> Just pondering, but do we actually need this column? It's a departure from the other information in the table, which is all static configuration data. Swap is a runtime setting, and toggling the flag won't update the table.

What do you mean by "static configuration data"? Maybe you were thinking about `mz_cluster_replica_sizes` instead of `mz_cluster_replicas`?

src/adapter/src/catalog/open.rs
```diff
@@ -189,6 +189,41 @@ impl Catalog {
...
+                    size.selectors.remove("materialize.cloud/scratch-fs");
+                    size.selectors
+                        .insert("materialize.cloud/swap".into(), "true".into());
+                }
+            }
+
+            // The size definitions were copied into the in-memory replica configs during
+            // `CatalogState` initialization above. We can't move this code before that because we
+            // need to wait for the `system_configuration` to be fully initialized. Instead we also
+            // patch the replica configs here.
```
ggevay, line 209: Could you please also extend the comment with the new constraint that this has to be before the builtin table updates?

src/adapter/src/catalog/builtin_table_updates.rs
ggevay: Ok, I see you have moved the `// Force cc sizes to use swap, if requested.` part up, thanks! In that case, I think checking `FORCE_SWAP_FOR_CC_SIZES` is not needed here, because the `allocation` should already have the right swap setting, as far as I understand. (And checking `FORCE_SWAP_FOR_CC_SIZES` here would still have the problem that the table would get out of sync with reality, because whether the replica size has swap is fixed at bootup, as far as I understand.)

ggevay:
(The `upload-debug-symbols` CI fails are unrelated, and are being investigated [here](https://materializeinc.slack.com/archives/C01LKF361MZ/p1757925875365999?thread_ts=1757922760.772449&cid=C01LKF361MZ).)

src/adapter/src/catalog/builtin_table_updates.rs
jubrad: Yeah, totally agree, I moved it without changing the other part because this is how I know to get another build I can deploy to staging. 

Something might still off even with this moved up. I actually managed to corrupt my mz_cluster_replicas table since it saw a retraction for a cluster with `swap=true` but that cluster size had `swap=false`. 

I agree though, I'll remove the check, and do a bit more validation to ensure that a prior deploy hadn't gotten me into this weird state.

ggevay: > Something might still off even with this moved up. I actually managed to corrupt my mz_cluster_replicas table since it saw a retraction for a cluster with swap=true but that cluster size had swap=false.

Yes, I think that is expected if we have the
```
        if FORCE_SWAP_FOR_CC_SIZES.get(dyncfg) && is_cc {
            swap = true;
        }
```
What I imagine to have happened is that the row was added with `swap=false`, but then we tried to remove it with `swap=true` due to the above line, because the flag was flipped between adding and removing the row.

teskje:
Have we thought about the compatibility implications of this? Since this is a stable catalog table, we have to keep it backward compatible forever.

* We won't be able to remove this column again later, so we should have a good motivation for adding it. The PR doesn't include a motivation though.
* Are we allowed to add columns to system relations? Do we need to check if anybody is running `SELECT * FROM mz_cluster_replicas` queries that would break after this change?

src/adapter/src/catalog/builtin_table_updates.rs
jubrad: Yeah, there's something really annoying here. 

We need to wait till system params are loaded, but that happens well after we initialize builtin clusters. The current code that modifies clusters based on `FORCE_SWAP_FOR_CC_SIZES` does so only in the in memory catalog state outside of transactions that would lead to pack_cluster_replica_size.

I think this PR can just wait till we remove the `FORCE_SWAP_FOR_CC_SIZES`.

jubrad:
@teskje great point!

FWIW if I recall correctly we're moderately ok with adding new columns to system tables, but are committed to not adding breaking changes. 

It would probably make more sense to add this sort of configuration into `mz_cluster_replica_sizes` then just do

```sql
select r.*, is_swap from mz_cluster_replica_sizes as s right join mz_cluster_replicas as r on s.size = r.size;
```

We certainly have to have the ability to change the shape of cluster replica sizes. 

If this sounds good I can revamp this PR to add swap there.

teskje:
Well, I think `mz_cluster_replica_sizes` has exactly the same constraints as `mz_cluster_replicas` in that it is a stable system relation and we are thus limited in what changes we can make to it. You could maybe argue that it's less likely for users to query the former than the latter, so changing the former would be less risky.

> We certainly have to have the ability to change the shape of cluster replica sizes.

We do have the ability to change the replica size definitions, but that's different from changing `mz_cluster_replica_sizes`. The way we reconcile this is by accepting that not every field in the replica size definition shows up in `mz_cluster_replica_sizes`. And by setting columns to NULL/0 if we remove fields from the replica size definitions.

jubrad:
> Well, I think mz_cluster_replica_sizes has exactly the same constraints as mz_cluster_replicas in that it is a stable system relation and we are thus limited in what changes we can make to it

Yeah, this confuses me a bit, I feel like we should somewhere have tables that accurately represent the state of the cluster sizes or the replica configuration somewhere. It's possible we should just add a new internal table if our schemas are too unstable to report in system tables.

teskje:
Personally I think not exposing all the nitty-gritty detail about our replica sizes in SQL is fine. That gives us the freedom to add flags like `is_cc` and not be too embarrassed about using that for sizes not named `*cc` as well. I agree having `mz_cluster_replica_sizes` an unstable relation would be more convenient, but that ship has sailed. Adding a second relation for sizes would be possible but also feels pretty bad :/

Anyhow, I've since learned about the motivation for wanting a `swap_enabled` flag somewhere in the catalog (namely making UI decisions in the Console), which is hard to argue with. I'd still prefer adding the flag to `mz_cluster_replica_sizes` instead of `mz_cluster_replicas`, since it's a per-size configuration, not a per-replica configuration. Given the self-managed situation it seems like we'll have to support non-swap configurations for a while still, so adding a new stable column feels less bad than it would if it was only needed for a couple months.

teskje:
> I've since learned about the motivation for wanting a swap_enabled flag somewhere in the catalog (namely making UI decisions in the Console), which is hard to argue with.

Actually, I may have found a way to argue with it! It occurred to me that in self-managed it will be possible to have a replica that has both a `/scratch` disk and swap enabled. So at least the simple logic I imagined, `if swap_enabled { show_unified_ui() } else { show_memory_plus_disk_ui() }` won't work.

Instead we could always show the unified UI ("Memory Utilization") and only show the disk graph if there is a `/scratch` disk attached to the cluster. I.e. `show_unified_ui(); if disk_attached { show_disk_ui() }`. The unified UI should be applicable to clusters without swap as well, so we don't need to make it conditional.

At the moment the console is calculating `memory + disk` in order to get usage and limits for the unified UI, so my proposed approach doesn't quite work. But I think after the changes for https://github.com/MaterializeInc/database-issues/issues/9692 it should work. The unified UI would just use `heap_bytes`/`heap_limit`, and the disk UI would just use `disk_bytes`.

jubrad:
> Instead we could always show the unified UI ("Memory Utilization") and only show the disk graph if there is a /scratch disk attached to the cluster. I.e. show_unified_ui(); if disk_attached { show_disk_ui() }. The unified UI should be applicable to clusters without swap as well, so we don't need to make it conditional.

^yes, this sounds good. 

Personally I think it'd be nice to have a more consumable representation of our cluster sizes, even if it's like an internal table that had the raw json, but that's hardly a good reason for adding the column. @leedqin can weigh in on our suggestion and I'm happy to close this pr if it sounds good.

PR #33570, opened 2025-09-11
Envd ed25519

jubrad:
### Motivation
I think this will improve CPS and lower ENVD CPU. Since public comms should go through balancer this shouldn't have compatibility issues. 

### Checklist

### DIFF
--- a/src/orchestratord/src/controller/materialize/tls.rs
+++ b/src/orchestratord/src/controller/materialize/tls.rs
@@ -20,6 +20,8 @@ pub fn create_certificate(
     cert_name: String,
     secret_name: String,
     additional_dns_names: Option<Vec<String>>,
+    algorithm: CertificatePrivateKeyAlgorithm,
+    size: Option<i64>,
 ) -> Option<Certificate> {
     let default_spec = default_spec.unwrap_or_else(MaterializeCertSpec::default);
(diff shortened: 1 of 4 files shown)


src/orchestratord/src/controller/materialize/tls.rs
```diff
@@ -52,10 +52,10 @@ pub fn create_certificate(
             duration: mz_cert_spec.duration.or(default_spec.duration),
             issuer_ref,
             private_key: Some(CertificatePrivateKey {
-                algorithm: Some(CertificatePrivateKeyAlgorithm::Rsa),
+                algorithm: Some(CertificatePrivateKeyAlgorithm::Ed25519),
```
alex-hunt-materialize: This function is also used for balancers and the console, so this will cause compatibility issues.

jubrad:
> This seems fine, but I highly doubt that the algorithm we use will have any significant impact on how many connections we can handle.

Initial investigations are showing that envd and potentially balanced are restricted by cpu with significant time spent in cryptography, since we scale balancerd based on cpu they should be able to handle more load or can be pre-scaled if we know of a high connection event. Bumping up envd cpu gave a pretty big boost in CPS, and local testing done by @SangJunBak indicates a pretty significant CPS bump using cloud-kind.

PR #32317, opened 2025-04-23
WIP dyncfg and system params from file

jubrad:
### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #33571, opened 2025-09-12
Add system_param configmap to mz crd

jubrad:
Follow up of https://github.com/MaterializeInc/materialize/pull/32317

### Motivation
Adds the ability to manage system params using
file dyncfg and a k8s configmap

configmap sync is painfully slow, I wasn't sure if this was working at first.

### Checklist

### DIFF
--- a/src/orchestratord/src/controller/materialize/environmentd.rs
+++ b/src/orchestratord/src/controller/materialize/environmentd.rs
@@ -77,14 +77,14 @@ static V154_DEV0: LazyLock<Version> = LazyLock::new(|| Version {
     major: 0,
     minor: 154,
     patch: 0,
-    pre: Prerelease::new("dev.0").expect("dev.0 is valid prerelease"),
+    pre: Prerelease::new("").expect("dev.0 is valid prerelease"),
     build: BuildMetadata::new("").expect("empty string is valid buildmetadata"),
 });
 pub const V161: Version = Version::new(0, 161, 0);
 
-static V26_0_0_DEV0: LazyLock<Version> = LazyLock::new(|| Version {
-    major: 0,
-    minor: 165,
+static V26_1_0: LazyLock<Version> = LazyLock::new(|| Version {
+    major: 26,
+    minor: 1,
     patch: 0,
     pre: Prerelease::new("dev.0").expect("dev.0 is valid prerelease"),
     build: BuildMetadata::new("").expect("empty string is valid buildmetadata"),
@@ -1243,7 +1243,7 @@ fn create_environmentd_statefulset_object(
     // Add system_param configmap
     // This feature was enabled in 0.163 but did not have testing until after 0.164.
     // 0.165 should work with anything greater than 0.164 including v26 and v25.
-    if mz.meets_minimum_version(&V26_0_0_DEV0) {
+    if mz.meets_minimum_version(&V26_1_0) {
         if let Some(ref name) = mz.spec.system_parameter_configmap_name {
             volumes.push(Volume {
                 name: "system-params".to_string(),
--- a/test/orchestratord/mzcompose.py
+++ b/test/orchestratord/mzcompose.py
@@ -845,16 +845,19 @@ class SystemParamConfigMap(Modification):
             configmap_name = "system-params-test"
 
             # First create the configmap in the cluster
-            configmap_yaml = f"""
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: {configmap_name}
-  namespace: materialize-environment
-data:
-  system-params.json: |
-    {json.dumps(self.value)}
-"""
+            configmap_yaml = yaml.safe_dump(
+                {
+                    "apiVersion": "v1",
+                    "kind": "ConfigMap",
+                    "metadata": {
+                        "name": configmap_name,
+                        "namespace": "materialize-environment",
+                    },
+                    "data": {
+                        "system-params.json": json.dumps(self.value),
+                    },
+                }
+            )
             # Apply the configmap (will be done in init/run)
             definition["system_params_configmap"] = configmap_yaml
             # Set the configmap name in the Materialize spec


alex-hunt-materialize (dismissed):

misc/python/materialize/mzcompose/services/materialized.py
```diff
@@ -322,6 +323,12 @@ def __init__(
             volumes.append(f"{listeners_config_path}:/listeners_config")
             environment.append("MZ_LISTENERS_CONFIG_PATH=/listeners_config")
 
+        if config_sync_file_path is not None:
+            # assert os.path.exists(str(config_sync_file_path))
+            volumes.append(f"{config_sync_file_path}:/config_sync.json")
+            environment.append("MZ_CONFIG_SYNC_FILE_PATH=/config_sync.json")
```
alex-hunt-materialize, line 329: You use config_sync.json here, but materialize.json in the docs for the field in the CRD.

src/cloud-resources/src/crd/materialize.rs
```diff
@@ -194,6 +194,17 @@ pub mod v1alpha1 {
...
+        // The name of a configmap containing system_parameters in json format
+        // NOTE: The configmap must contain a `materialize.json` key whose value
+        // is a valid json containing valid system parameters.
+        // ex:
+        // data:
+        //   materialize.json: |
+        //     {
+        //       "max_connections": 10
+        //     }
+        pub system_parameter_configmap_name: Option<String>,
```
alex-hunt-materialize: See above, this comment doesn't match what you are doing above in the mzcompose materialized.py service file.

test/dyncfg/mzcompose.py
```diff
@@ -0,0 +1,88 @@
...
+SERVICES = [
+    Testdrive(),
+    Materialized(),
+]
+
+
+def workflow_default(c: Composition) -> None:
+    # Create config file in MZ_ROOT directory
+    mz_root = Path(os.environ.get("MZ_ROOT", Path(__file__).parent.parent.parent))
+    config_file = mz_root / "dyncfg-test-config.json"
```
alex-hunt-materialize, line 28: This also doesn't match the path in the either the CRD docs or the materialized.py mzcompose service file.

src/orchestratord/src/controller/materialize/environmentd.rs
```diff
@@ -1233,6 +1232,30 @@ fn create_environmentd_statefulset_object(
     // The `materialize` user used by clusterd always has gid 999.
     args.push("--orchestrator-kubernetes-service-fs-group=999".to_string());
 
+    // Add system_param configmap
+    if let Some(ref name) = mz.spec.system_parameter_configmap_name {
```
alex-hunt-materialize: This needs a version gate.

alex-hunt-materialize (commented):
It's a bit confusing that we are using three different names for the file in three different places.

Can we also get a version gate and a test for orchestratord?

src/orchestratord/src/controller/materialize/environmentd.rs
jubrad: good catch!

misc/python/materialize/mzcompose/services/materialized.py
jubrad: mzcompose is not using configmaps meaning we don't really care what the file name is as long as we pass in the correct path here. I don't really see any reason to use `materialize.json`.

test/dyncfg/mzcompose.py
jubrad: Yeah, this doesn't have to match the CRD, arguably it's a better test if it doesn't. We pass this config file path through to the Materialize mzcompose service and it sets the flag (via env vars) directly. As long as the file is accessible in the container and the path is correct everything should be good.

src/orchestratord/src/controller/materialize/environmentd.rs
```diff
@@ -83,6 +82,14 @@ static V154_DEV0: LazyLock<Version> = LazyLock::new(|| Version {
 });
 pub const V161: Version = Version::new(0, 161, 0);
 
+static V165_DEV0: LazyLock<Version> = LazyLock::new(|| Version {
+    major: 0,
+    minor: 165,
```
alex-hunt-materialize: Should this be v26.0.0-dev.0? I guess v0.165.0-dev.0 would meet that version, so maybe this is fine?

alex-hunt-materialize (commented):
When I said orchestratord test, I meant adding a `Modification` to test/orchestratord/mzcompose.py. Right now, we have no tests running an actual orchestratord that exercise this, so we have no tests of the volume and mount that it sets up.

I also still hate that we have three different names for the file, and I also don't like any of the names chosen. The `materialize.json` is the worst of the bunch. That one is the one that end users are going to see, so I'd prefer it to be something descriptive. Maybe `system-params.json`? I won't block this PR for this, but the current naming feels bad.

jubrad:
@alex-hunt-materialize I added a test, between this test and the dyncfg I think there's adequate coverage, but I do think in the fullness of time we'd want a psql connection to validate the settings are getting synced. 

I can circle back with testing there and documentation. Given that this will miss the v26, it may be worth waiting for at least the latter to join this PR before merging.

test/orchestratord/mzcompose.py
```diff
@@ -718,6 +713,90 @@ def check_replicas():
...
+            configmap_yaml = f"""
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {configmap_name}
+  namespace: materialize-environment
+data:
+  system-params.json: |
+    {json.dumps(self.value)}
+"""
```
alex-hunt-materialize: Doing string interpolation like this in yaml is unsafe.

Untested suggestion:
```suggestion
            configmap_yaml = yaml.safe_dumps({
                "apiVersion": "v1",
                "kind": "ConfigMap",
                "metadata": {
                    "name": configmap_name,
                    "namespace": "materialize-environment",
                },
                "data": {
                    "system-params.json": json.dumps(self.value),
                },
            })
```

test/orchestratord/mzcompose.py
```diff
@@ -718,6 +713,90 @@ def check_replicas():
...
+            ), f"Expected to find system-params volume mount in {volume_mounts}"
+
+            # Check that the config-sync-file-path arg is present
+            args = environmentd["items"][0]["spec"]["containers"][0]["args"]
+            expected_arg = "--config-sync-file-path=/etc/materialize/system-params/system-params.json"
+            assert any(
+                expected_arg in arg for arg in args
+            ), f"Expected {expected_arg} in environmentd args, but only found: {args}"
+            # TODO, validate that via SQL that the values have been updated
+            # this won't work due to a known issue with psycopg connect failing in CI.
```
alex-hunt-materialize: Do we have a link to the ticket for this issue?

test/orchestratord/mzcompose.py
```diff
@@ -718,6 +713,90 @@ def check_replicas():
...
+  name: {configmap_name}
+  namespace: materialize-environment
+data:
+  system-params.json: |
+    {json.dumps(self.value)}
+"""
+            # Apply the configmap (will be done in init/run)
+            definition["system_params_configmap"] = yaml.load(
+                configmap_yaml, Loader=yaml.Loader
+            )
```
alex-hunt-materialize: Wait, why are we even creating yaml if we're just going to load it back into a python dict and drop the yaml?

src/orchestratord/src/controller/materialize/environmentd.rs
```diff
@@ -82,6 +82,14 @@ static V154_DEV0: LazyLock<Version> = LazyLock::new(|| Version {
 });
 pub const V161: Version = Version::new(0, 161, 0);
 
+static V26_0_0_DEV0: LazyLock<Version> = LazyLock::new(|| Version {
+    major: 0,
+    minor: 165,
```
alex-hunt-materialize: Nitpick, you're calling it `V26_0_0_DEV0`, but actually checking against `v0.165.0`. We should be consistent here. You could also just point at `v26.0.0` now, as that is now released.

PR #33574, opened 2025-09-12
docs/developer: add life-of-a-query.md

aljoscha:
Rendered: https://github.com/aljoscha/materialize/blob/docs-life-of-a-query/doc/developer/life-of-a-query.md

### DIFF
--- a/doc/developer/life-of-a-query.md
+++ b/doc/developer/life-of-a-query.md
@@ -124,10 +124,10 @@ sending results back to the client.
 
 Query planning generates a
 [Plan](https://github.com/MaterializeInc/materialize/blob/b3475aaaf96f9fae4a0d68cbb5202c224d9ce15b/src/sql/src/plan.rs#L133)
-from the AST. Glossing over some details, this binds referenced names based on
-the Catalog as of planning time and determines an execution plan. The entry
-point is
-[plan](https://github.com/MaterializeInc/materialize/blob/b3475aaaf96f9fae4a0d68cbb5202c224d9ce15b/src/sql/src/plan/statement.rs#L274).
+from the AST. Glossing over some details, this first binds referenced names
+based on the Catalog as of planning time. Then
+[plan](https://github.com/MaterializeInc/materialize/blob/b3475aaaf96f9fae4a0d68cbb5202c224d9ce15b/src/sql/src/plan/statement.rs#L274)
+determines an execution plan from the resolved AST.
 
 One of the differences between the user (SQL) commands that are input to the
 adapter and commands for the rest of the system is the use of user-defined and


aljoscha:
@teskje @SangJunBak @ggevay could you review the adapter side of this? Which is most of it actually... :sweat_smile:

aljoscha:
The intention is also that other folks can add to this doc, and it's evolving as our architecture evolves.

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+The
+[adapter](https://github.com/MaterializeInc/materialize/tree/f641a29d4aad9fcfeb2de535ff54706a1f1d38c4/src/adapter)
+is named such because it translates user commands into commands that the other
+internal systems understand. It currently understands SQL but is intended as a
+generic component that isolates the other components from needing to know
+details about the front-end commands (SQL).
+
+A core component is the
+[Coordinator](https://github.com/MaterializeInc/materialize/blob/9682eb11f31e2eb8005cb3eae687f81a1bce21bb/src/adapter/src/coord.rs#L1633).
+It mediates access to durable environment state, kept in a durable catalog and
```
ggevay, line 66: > It mediates access to durable environment state, kept in a durable catalog

Wouldn't the stuff in Persist also considered "durable environment state"? Edit: I see this is treated later in the "Durable Storage" section.

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+> (which can do work for different connections/sessions concurrently) and the
+> controllers for the other components. See [Design Doc: A Small
+> Coordinator](/doc/developer/design/20250717_a_small_coordinator_more_scalable_isolated_materialize.md)
+> for details.
+
+## Query Processing
+
+Query processing follows these steps:
+
+parsing & describing → logical planning → timestamp selection → optimization & physical planning → execution
```
ggevay: By "logical" and "physical`, did you mean HIR and LIR? I wouldn't use the terms "logical" and "physical" in this way, because the code doesn't use them for these stages. In fact, the optimizer code uses these terms to mean something else: specific parts of the MIR optimization pipeline (see the `logical_optimizer` and `physical_optimizer` functions).

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+sending results back to the client.
+
+## Logical Planning
+
+Logical planning generates a
+[Plan](https://github.com/MaterializeInc/materialize/blob/b3475aaaf96f9fae4a0d68cbb5202c224d9ce15b/src/sql/src/plan.rs#L133)
+from the AST. Glossing over some details, this binds referenced names based on
+the Catalog as of planning time and determines a logical execution plan. The
+entry point is
+[plan](https://github.com/MaterializeInc/materialize/blob/b3475aaaf96f9fae4a0d68cbb5202c224d9ce15b/src/sql/src/plan/statement.rs#L274).
```
ggevay: The text mentions name resolution and then links the `plan` function, but actually name resolution happens before the linked `plan` function. One can see this from the parameter `stmt: Statement<Aug>`, where `Aug` means that names have already been resolved. (For unresolved, it would be `Statement<Raw>`.) (Name resolution is done by `mz_sql::names::resolve`.)

ggevay: Btw. there is also `transform_ast` between parsing and the `plan` function, but maybe the "Glossing over some details" was intended to handle this.

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+snippet](https://github.com/MaterializeInc/materialize/blob/bdf9573960cece48e7811db7ef777b973d355fe6/src/adapter/src/coord/sequencer/inner/peek.rs#L568).
+
+The final result of these stages will depend on the type of query we're
+optimizing, but for certain types of SELECT and permanent objects it will
+include a
+[DataflowDescription](https://github.com/MaterializeInc/materialize/blob/53acc93eee1bbbf418fde681389aec0419db8954/src/compute-types/src/dataflows.rs#L40).
+Which is a physical execution plan that can be given to the compute layer, to
+execute on a cluster.
+
+TODO: Expand this section on optimization, if/when needed.
```
ggevay: Or just link to `101-query-compilation.md`. This unfortunately currently in a very outdated state, but I could update it, if there is interest.

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+
+## Execution on a Cluster
+
+For SELECT, which is internally called PEEK, there are three different execution scenarios:
```
ggevay, line 178: > `PEEK`

nit: It's neither an abbreviation nor an SQL keyword, so no need to write it in all-caps. (And one more occurrence below.)

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+  shape that allows us to read a result right out of a storage collection (in
+  persist).
+
+The result of optimization will indicate which of these scenarios we're in, and
+the adapter will now have to talk to the compute controller to implement
+execution of the query.
+
+Ultimately, for all of these scenarios the adapter will make a call into the
+compute controller to read out a peek result. For slow-path peeks it will first
+create the dataflow, but the functionality for reading our the result is the
```
ggevay, line 196: "reading our the result" -> "reading out the result"

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+same for fast path and slow path after that. The entry point for setting of the
+actual peek is
+[peek](https://github.com/MaterializeInc/materialize/blob/4a84902b5bed3bbd605d7d165fa6e0823c88c102/src/compute-client/src/controller.rs#L862).
```
ggevay, line 199: (This function might get refactored away soon, when we delete the old peek sequencing, because the frontend peek sequencing wants to call `instance::peek` directly. The frontend peek sequencing currently still calls the old `controller::peek` for slow-path peeks, but this will probably change when we delete the old peek sequencing.)

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+And the entry point for creating a dataflow is
+[create_dataflow](https://github.com/MaterializeInc/materialize/blob/4a84902b5bed3bbd605d7d165fa6e0823c88c102/src/compute-client/src/controller.rs#L759).
+
+The adapter will pass the sending endpoint of a channel to the compute
+controller, for sending back the results, and then setup up an async task that
```
ggevay, line 204: "setup up"

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+Clusters run dataflow computations, and within those dataflows, arrangements
+are the structure that data is being kept in. Arrangements are used both for
+keeping data internal to the dataflow but also for incrementally maintained
+results. For example, an index is a dataflow where at the end we have an
+arrangement that a peek can read out of.
+
+On the more technical side, arrangements are multi-versioned indexes. The
+[arrangements documentation](/doc/developer/arrangements.md) describes them as
+an indexed representation of a stream of update triples `(data, time, diff)`,
+organized by key for efficient lookups.
```
ggevay: You could also link the "Shared Arrangement" paper: https://arxiv.org/pdf/1812.02639

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+Results can be sent back in two ways: inline in the protocol via
+`ComputeResponse`, or using the _peek result stash_. The latter allows for
+sending back larger results but there are certain restrictions on what shapes
+of queries are eligible to use the stash. This has to do with post processing,
+which we will describe below. Roughly, queries that don't require post
+processing can use the peek stash, so queries that have OFFSET, LIMIT, or ORDER
+BY _cannot_ use the large result stash.
+
+For all queries, we start out trying to use the inline result protocol but then
+fall back to the stash when we notice the result size getting to large and a
```
ggevay: "getting to large"

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+document calls post processing but which the code calls "finishing" (see
+[RowSetFinishing](https://github.com/MaterializeInc/materialize/blob/ddc1ff8d2d201c263b5e1ff40effccceb153a767/src/expr/src/relation.rs#L3640)).
+We apply any ORDER BY, LIMIT, or OFFSET clauses only _after_ results get back
+from the clusters, before sending the result out to the client. This happens
+within the async task that routes results from the controller to the front end,
+in [this
+code](https://github.com/MaterializeInc/materialize/blob/53acc93eee1bbbf418fde681389aec0419db8954/src/adapter/src/coord/peek.rs#L910).
+As mentioned above, the presence or not of a Finishing is what determines the
+eligibility of a query for the large result stash feature. Queries that require
+a Finishing cannot use the stash.
```
ggevay: `RowSetFinishing::is_streamable` allows some finishings to use the stash. (But I'm not sure if we want to go into those details in this document. If not, then we can leave it more vague in which cases a finishing allows the use of the stash.)

ggevay: Also, we seem to have the same assert twice here: https://github.com/MaterializeInc/materialize/blob/deba7d44fd6c2511116d3eec9353626b984ecb5b/src/adapter/src/coord/peek.rs#L1051-L1058

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+- The `ComputeResponse::PeekResponse` sent from the cluster to the controller
+- The result sent back from the adapter front end to the client
+
+## Details
+
+Details on components and concepts that we didn't explain above.
+
+### Durable Storage
+
+At the coarsest level, a Materialize Environment has to types of durable state:
```
ggevay: "to types"

doc/developer/life-of-a-query.md
```diff
@@ -0,0 +1,444 @@
...
+
+- The _Catalog_, which is the metadata about what objects exists, settings,
```
ggevay: exist

aljoscha:
@ggevay pushed some changes, could you double check whether that fixes your remarks? 🙏

ggevay:
Thank you, looks good, just a very minor thing still:

> this binds referenced names based on
the Catalog as of planning time and determines an execution plan. The entry
point is
[plan](https://github.com/MaterializeInc/materialize/blob/b3475aaaf96f9fae4a0d68cbb5202c224d9ce15b/src/sql/src/plan/statement.rs#L274)

The linked `plan` function does not include name resolution, i.e., the "binds referenced names" happens before this `plan` function. So, if the "entry point" is meant to say that that function does all these things, then it's not strictly true. (See also [here](https://github.com/MaterializeInc/materialize/pull/33574/changes/453e2cc7b484a07966b7de100460966a0757327d..b4c1d341aa57a7f3463a04b1a6808f06257d6375#r2742769613).)

PR #33589, opened 2025-09-15
storage,0dt: include source exports (aka. subsources) in hydration checks

aljoscha:
Before, we were erroneously only including the "root" ingestion objects and not any of their subsource/tables in hydration checks.

@def-  I believe this should fix https://github.com/MaterializeInc/database-issues/issues/9687, but I haven't yet tried running that using the `--release` nightly config.

### DIFF
--- a/src/storage-client/src/controller.rs
+++ b/src/storage-client/src/controller.rs
@@ -359,8 +359,10 @@ pub trait StorageController: Debug {
     /// capabilties.
     fn active_collection_metadatas(&self) -> Vec<(GlobalId, CollectionMetadata)>;
 
-    /// Returns the IDs of all active ingestions for the given storage instance.
-    fn active_ingestions(
+    /// Returns the IDs of ingestion exports running on the given instance. This
+    /// includes the ingestion itself, if any, and running source tables (aka.
+    /// subsources).
+    fn active_ingestion_exports(
         &self,
         instance_id: StorageInstanceId,
     ) -> Box<dyn Iterator<Item = &GlobalId> + '_>;
--- a/src/storage-controller/src/instance.rs
+++ b/src/storage-controller/src/instance.rs
@@ -254,9 +254,17 @@ where
         }
     }
 
-    /// Returns the ingestions running on this instance.
+    /// Returns ingestions running on this instance. This _only_ includes the
+    /// "toplevel" ingestions, not any of their source tables (aka. subsources).
     pub fn active_ingestions(&self) -> impl Iterator<Item = &GlobalId> {
         self.active_ingestions.keys()
+    }
+
+    /// Returns ingestion exports running on this instance. This includes the
+    /// ingestion itself, if any, and running source tables (aka. subsources).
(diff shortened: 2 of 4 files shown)


src/storage-controller/src/instance.rs
```diff
@@ -254,9 +254,19 @@ where
         }
     }
 
-    /// Returns the ingestions running on this instance.
+    /// Returns ingestions running on this instance. This _only_ includes the
+    /// "toplevel" ingestions, not any of their source tables (aka. subsources).
     pub fn active_ingestions(&self) -> impl Iterator<Item = &GlobalId> {
-        self.active_ingestions.keys()
+        let ingestion_exports = self.ingestion_exports.keys();
+        self.active_ingestions.keys().chain(ingestion_exports)
```
martykulma: Copy-pasta? If this is supposed to be just the top level ingestion, should it chain ingestion_exports?

aljoscha: Copy-pasta indeed! 🙈

def- (approved):
Test is green, good to know it actually found something of interest :)

Edit: One question I'm wondering is why it shows up now, are we just (un)lucky with the timing?

aljoscha:
> Test is green, good to know it actually found something of interest :)
> 
> Edit: One question I'm wondering is why it shows up now, are we just (un)lucky with the timing?

Might be unlucky, because you're saying that this also doesn't happen with the `--optimized` build...

@def- Did you double-check that I correctly set my nightly build to use release builds? Yours was on a branch that had a bug, so I started one once I fixed that.

def-:
Yes, it was using the Bazel+LTO build.

src/storage-client/src/controller.rs
```diff
@@ -359,8 +359,10 @@ pub trait StorageController: Debug {
     /// capabilties.
     fn active_collection_metadatas(&self) -> Vec<(GlobalId, CollectionMetadata)>;
 
-    /// Returns the IDs of all active ingestions for the given storage instance.
-    fn active_ingestions(
+    /// Returns the IDs of ingestion dataflows running on the given instance.
+    /// This includes the dataflow of the ingestion itself, if any, and running
+    /// ingestion dataflows for source tables (aka. subsources).
+    fn active_ingestion_dataflows(
```
teskje: This description doesn't match my mental model, but not sure which of them is wrong. My understanding was that there is a single dataflow for each ingestion, and if there are multiple subsources then that dataflow will have multiple persist_sinks at the end. But the description suggests there is actually a dataflow per subsource?

teskje (commented):
The idea seems reasonable to me, but I'm not sure about calling subsources "ingestion dataflows". Might be I don't understand storage well enough!

src/storage-client/src/controller.rs
aljoscha: Pushed a rename to `active_ingestion_exports`. In my head I often call fragments or subgraphs of a dataflow a dataflow as well, but export might be the more official name. 😅

PR #33590, opened 2025-09-16
Balancerd sni first

jubrad:
### Motivation
This gives a fast path to tenant resolution if the requester provides SNI in their pgwire TLS connection. It should be significantly faster to initiate a connection.

### Checklist


src/balancerd/src/lib.rs
```diff
@@ -350,7 +350,7 @@ impl BalancerService {
             });
         }
         {
-            let Some((addr, port)) = self.cfg.https_addr_template.split_once(':') else {
+            let Some((addr, port)) = self.cfg.https_sni_addr_template.split_once(':') else {
```
alex-hunt-materialize, line 353: I wonder if we should instead have separate variables for the port vs the address template. That way, we could have a single `sni_addr_template` and two port fields (`http_port` and `pgwire_port`).

test/balancerd/mzcompose.py
```diff
@@ -663,3 +664,21 @@ def workflow_webhook(c: Composition) -> None:
     """
         )
     )
+
+def workflow_pgwire_without_sni(c: Composition) -> None:
```
alex-hunt-materialize: Do we have one that explicitly uses SNI? If not, can we get one? Maybe just ensure frontegg-mock isn't running, so if it calls out to frontegg we know it's broken.

alex-hunt-materialize (commented):
Overall looks ok.

src/balancerd/src/lib.rs
jubrad: Yeah, I had thought about that, and done this in one of the failed branches. The main reason I like this more is that it is more backwards compatible.

test/balancerd/mzcompose.py
jubrad: I stole this from paul's PR 😄 

I think all other connections, as long as they're using TLS will be using SNI. 
We could maybe fetch the metrics endpoint to validate this. I had also done some gross stuff in my failed pr to write a mockable resolver that I can introduce call backs into to validate they're being used and mock out to provide tenant resolution when running the balancer rust tests.

I'll try one with FE mock turned off.

jubrad: Good suggestion for the test!

I don't know that it offers a ton since all of the other connections should be using SNI, but this ensures that we're not touching frontegg.

src/balancerd/src/lib.rs
```diff
@@ -1269,48 +1268,97 @@ impl<T: AsyncRead + AsyncWrite + Unpin + Send> ClientStream for T {}
 #[derive(Debug)]
 pub enum Resolver {
     Static(String),
-    Frontegg(FronteggResolver),
+    MultiTenant(FronteggResolver, StubResolver, Option<(String, u16)>),
```
alex-hunt-materialize: I know from context that the `Option<(String, u16)>` is the SNI template and port, but it likely wouldn't be clear to anyone just glancing at this. Maybe we should move that to its own type.

```rust
struct SniResolver {
    template: String,
    port: u16,
}
```

src/balancerd/src/bin/balancerd.rs
```diff
@@ -231,17 +241,31 @@ pub async fn run(args: ServiceArgs, tracing_handle: TracingHandle) -> Result<(),
...
-                }),
+                Resolver::MultiTenant(
+                    FronteggResolver {
+                        auth,
+                        addr_template,
+                    },
+                    StubResolver::new(),
+                    args.pgwire_sni_resolver_template
+                        .map(|t| {
+                            t.split_once(':').map(|(t, p)| {
```
alex-hunt-materialize: Nitpick
```suggestion
                            t.rsplit_once(':').map(|(t, p)| {
```

test/balancerd/mzcompose.py
```diff
@@ -663,3 +666,15 @@ def workflow_webhook(c: Composition) -> None:
...
         )
     )
+
+
+def workflow_pgwire_with_sni(c: Composition) -> None:
+    c.up(
+        "balancerd",
+        "materialized",
+        Service("testdrive", idle=True),
+    )
```
alex-hunt-materialize, line 676: Can we get an explicit down for frontegg-mock?

alex-hunt-materialize (approved):
Looks good, except for a few nitpicks.

test/balancerd/mzcompose.py
jubrad: I don't know?

PR #33592, opened 2025-09-16
Update to latest Timely and Differential

antiguru:
Update to timely 0.24 and differential 0.17.


def- (commented):
Nightly run as usual: https://buildkite.com/materialize/nightly/builds/13404

PR #33602, opened 2025-09-16
Fix case where role modification can wipe out role password

DAlperin:
### Checklist

### DIFF
--- a/src/catalog/src/durable/transaction.rs
+++ b/src/catalog/src/durable/transaction.rs
@@ -30,14 +30,15 @@ use mz_repr::network_policy_id::NetworkPolicyId;
 use mz_repr::role_id::RoleId;
 use mz_repr::{CatalogItemId, Diff, GlobalId, RelationVersion};
 use mz_sql::catalog::{
-    CatalogError as SqlCatalogError, CatalogItemType, ObjectType, RoleAttributes, RoleMembership,
-    RoleVars,
+    CatalogError as SqlCatalogError, CatalogItemType, ObjectType, PasswordAction,
+    RoleAttributesRaw, RoleMembership, RoleVars,
 };
 use mz_sql::names::{CommentObjectId, DatabaseId, ResolvedDatabaseSpecifier, SchemaId};
 use mz_sql::plan::NetworkPolicyRule;
 use mz_sql_parser::ast::QualifiedReplica;
 use mz_storage_client::controller::StorageTxn;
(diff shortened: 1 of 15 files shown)


src/sqllogictest/src/runner.rs
```diff
@@ -1832,12 +1858,30 @@ impl<'a> RunnerInner<'a> {
...
+                        expected_output: output,
+                        actual_raw_output: vec![],
+                        actual_output: actual,
+                        location,
+                    });
+                } else {
+                    return Ok(Outcome::Success);
+                }
+            }
+        };
```
SangJunBak: Maybe we move lines `1885->1918` to the `Ok(client) =>` match to not duplicate the early return and change `let client =` to a `let actual =`?

test/sqllogictest/role.slt
```diff
@@ -275,3 +275,125 @@ query TT
...
+----
+test_user
+COMPLETE 1
+
+# Clean up - revoke privileges and drop roles
+simple conn=mz_system,user=mz_system
+REVOKE USAGE ON DATABASE materialize FROM test_user
+----
+COMPLETE 0
+
```
SangJunBak, line 385: We should maybe add a test for granting a password role to a non password role. I wonder what the behavior for this is?

src/catalog/src/durable/transaction.rs
```diff
@@ -1467,21 +1469,43 @@ impl<'a> Transaction<'a> {
...
     ///
     /// Runtime is linear with respect to the total number of items in the catalog.
     /// DO NOT call this function in a loop, implement and use some `Self::update_roles` instead.
     /// You should model it after [`Self::update_items`].
-    pub fn update_role(&mut self, id: RoleId, role: Role) -> Result<(), CatalogError> {
+    pub fn update_role(
+        &mut self,
+        id: RoleId,
+        role: Role,
+        new_password: Option<Password>,
```
SangJunBak: Nit: While we're updating `update_role`, instead of separately calling `update_role` and `clear_role_password`, maybe we should move `clear_role_password` into `update_role` and convert `new_password` into an enum like `NewPassword(String) | RemovePassword` then wrap the enum in an Option?

test/sqllogictest/role.slt
DAlperin: The password shouldn't be inherited, I'll add a test though.

DAlperin:
tftr!

PR #33603, opened 2025-09-17
Unconditionally enable reduce reduction

antiguru:
Removes the feature flag and the reduce collation implementation that is now unreachable.

### DIFF
--- a/src/compute-types/src/plan/reduce.rs
+++ b/src/compute-types/src/plan/reduce.rs
@@ -271,8 +271,6 @@ pub enum BasicPlan {
 /// this aggregation.
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq, Ord, PartialOrd)]
 pub struct SingleBasicPlan {
-    /// The index in the set of inputs that we are aggregating over.
-    pub index: usize,
     /// The aggregation that we should perform.
     pub expr: AggregateExpr,
     /// Whether we fused a `FlatMap UnnestList` with this aggregation.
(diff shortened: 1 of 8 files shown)


frankmcsherry (approved):
We discussed and it seems good, but we also discussed a slightly slower roll-out where we put a soft assert here first, so that we can watch for a release or two and see if the code is as dead as we believe it to be, for production use cases. As @antiguru mentions, we are unable to be certain of the unreachability through types, in that if we can still be presented with a heterogeneous reduce if for any reason the transform fails to remove it.

antiguru:
Added the warning in #33726.

src/compute-types/src/plan/reduce.rs
```diff
@@ -371,7 +356,7 @@ impl ReducePlan {
     /// of the total list of requested aggregations.
```
ggevay: ("and are a subsequence" could be deleted)

src/compute-types/src/plan/reduce.rs
```diff
@@ -330,38 +320,33 @@ impl ReducePlan {
...
         // a subsequence of the overall sequence of aggregates.
-        for index in 0..aggregates.len() {
-            let typ = reduction_type(&aggregates[index].func);
-            let aggregates_list = reduction_types.entry(typ).or_insert_with(Vec::new);
-            aggregates_list.push((index, aggregates[index].clone()));
-        }
-
-        // Convert each grouped list of reductions into a plan.
-        let Some(plan) = reduction_types.pop_first().map(|(typ, aggregates_list)| {
+        let mut aggregates_list = Vec::with_capacity(aggregates.len());
```
ggevay, line 322: (Here also "subsequence" is outdated, in that we just take the whole list, right?)

src/compute-types/src/plan/reduce.rs
```diff
@@ -447,8 +414,8 @@ impl ReducePlan {
             ReductionType::Basic => {
                 if aggregates_list.len() == 1 {
                     ReducePlan::Basic(BasicPlan::Single(SingleBasicPlan {
-                        index: aggregates_list[0].0,
-                        expr: aggregates_list[0].1.clone(),
+                        index: 0,
+                        expr: aggregates_list[0].clone(),
```
ggevay: Could `SingleBasicPlan::index` be removed? Seems to be always 0.

antiguru: Done

antiguru:
Thanks for the reviews!

PR #33607, opened 2025-09-17
derive replica capacity limit from replica size

teskje:
To support replica sizes with different swap:memory ratios, this PR changes the way the replica memory limiter is configured. Instead of deriving the base limit from the memory limit and a dyncfg, the limit is now derived from the replica size and passed to the clusterd process as a commandline flag.

The exact logic for deriving the memory limiter limit (referred to as "capacity" here) is as follows: If swap is enabled, and the replica size definition specifies both a memory and a disk limit, the capacity limit is set to memory_limit + disk_limit. If swap is not enabled, or one of the two limits isn't set in the replica size definition, the memory limiter remains disabled.

Changing the memory limit at runtime is still possible through the remaining `memory_limiter_*` dyncfgs.

**(Note: Hold off merging until lgalloc is disabled everywhere in production.)**

### Motivation

Closes https://github.com/MaterializeInc/database-issues/issues/9691

### Checklist

### DIFF
--- a/src/clusterd/src/lib.rs
+++ b/src/clusterd/src/lib.rs
@@ -41,7 +41,7 @@ use mz_storage_types::connections::ConnectionContext;
 use mz_txn_wal::operator::TxnsContext;
 use tokio::runtime::Handle;
 use tower::Service;
-use tracing::{error, info, warn};
+use tracing::{error, info};
 
 mod usage_metrics;
 
@@ -138,9 +138,21 @@ struct Args {
     #[clap(long, env = "SCRATCH_DIRECTORY", value_name = "PATH")]
     scratch_directory: Option<PathBuf>,
 
-    /// Optional memory limit (bytes) of the cluster replica
+    /// Memory limit (bytes) of the cluster replica, if known.
+    ///
+    /// The limit is expected to be enforced by the orchestrator. The clusterd process only uses it
+    /// to inform configuration of backpressure mechanism.
     #[clap(long)]
     announce_memory_limit: Option<usize>,
+
+    /// Heap limit (bytes) of the cluster replica.
+    ///
+    /// A process heap usage is calculated as the sum of its memory and swap usage.
+    ///
+    /// In contrast to `announce_memory_limit`, this limit is enforced by the clusterd process. If
+    /// the limit is exceeded, the process terminates itself with a 167 exit code.
+    #[clap(long)]
+    heap_limit: Option<usize>,
 
     /// Whether this size represents a modern "cc" size rather than a legacy
     /// T-shirt size.
@@ -212,10 +224,10 @@ async fn run(args: Args) -> Result<(), anyhow::Error> {
     mz_alloc::register_metrics_into(&metrics_registry).await;
     mz_metrics::register_metrics_into(&metrics_registry, mz_dyncfgs::all_dyncfgs()).await;
 
-    if let Some(memory_limit) = args.announce_memory_limit {
-        mz_compute::memory_limiter::start_limiter(memory_limit, &metrics_registry);
+    if let Some(heap_limit) = args.heap_limit {
+        mz_compute::memory_limiter::start_limiter(heap_limit, &metrics_registry);
     } else {
-        warn!("no memory limit announced; disabling memory limiter");
+        info!("no heap limit announced; disabling memory limiter");
     }
 
     let secrets_reader = args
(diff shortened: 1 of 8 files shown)


src/clusterd/src/lib.rs
```diff
@@ -138,10 +138,22 @@ struct Args {
...
     announce_memory_limit: Option<usize>,
 
+    /// Capacity limit (bytes) of the cluster replica.
+    /// 
+    /// A process's capacity is the sum of its available memory and swap space.
+    ///
+    /// In contrast to `announce_memory_limit`, this limit is enforced by the clusterd process. If
+    /// the limit is exceeded, the process terminates itself with a 167 exit code.
+    #[clap(long)]
+    capacity_limit: Option<usize>,
```
antiguru: `heap_limit`? Capacity is a bit too generic unfortunately...

antiguru: What about passing a `disk_limit`, and letting the memory limiter calculate the capacity itself? This way you wouldn't need to add the logic to the controller.

teskje: Hm, it's a bit weird because memory limiter doesn't actually limit the disk, it limits the sum of memory and disk. So I think specifying a `--disk-limit` would lead to (even more) confusion.

Maybe "space limit"? Similarly generic to "capacity" though... I'm fine with `heap_limit` if you think it's good.

teskje: Changed to `heap_limit`.

src/clusterd/src/lib.rs
```diff
@@ -212,11 +224,10 @@ async fn run(args: Args) -> Result<(), anyhow::Error> {
...
     mz_metrics::register_metrics_into(&metrics_registry, mz_dyncfgs::all_dyncfgs()).await;
 
-    if let Some(memory_limit) = args.announce_memory_limit {
-        mz_compute::lgalloc::start_limiter(memory_limit, &metrics_registry);
-        mz_compute::memory_limiter::start_limiter(memory_limit, &metrics_registry);
+    if let Some(heap_limit) = args.heap_limit {
+        mz_compute::memory_limiter::start_limiter(heap_limit, &metrics_registry);
     } else {
-        warn!("no memory limit announced; disabling lgalloc limiter");
+        warn!("no heap limit announced; disabling memory limiter");
```
antiguru: Maybe we should change this to be less than `warn`? It spams tests this way, but I also see why we want it to warn...

src/compute-types/src/dyncfgs.rs
```diff
@@ -105,53 +105,18 @@ pub const MEMORY_LIMITER_INTERVAL: Config<Duration> = Config::new(
...
-    2.,
-    "Factor of the memory limit that the process will use before terminating the process.",
-);
-
 /// Bias to the memory limiter usage factor.
 pub const MEMORY_LIMITER_USAGE_BIAS: Config<f64> = Config::new(
     "memory_limiter_usage_bias",
     1.,
-    "Multiplicative bias to memory_limiter_usage_factor.",
+    "Multiplicative bias to the memory limiter's capacity limit.",
```
antiguru:
```suggestion
    "Multiplicative bias to the memory limiter's heap limit.",
```

src/compute-types/src/dyncfgs.rs
```diff
@@ -105,53 +105,18 @@ pub const MEMORY_LIMITER_INTERVAL: Config<Duration> = Config::new(
...
-    1.,
-    "Multiplicative bias to lgalloc_limiter_usage_factor.",
-);
-
-/// Burst factor to disk limit.
-pub const LGALLOC_LIMITER_BURST_FACTOR: Config<f64> = Config::new(
-    "lgalloc_limiter_burst_factor",
-    0.,
-    "Multiplicative burst factor to disk limit.",
+    "Multiplicative burst factor to the memory limiter's capacity limit.",
```
antiguru:
```suggestion
    "Multiplicative burst factor to the memory limiter's heap limit.",
```
Maybe?

antiguru (approved):
Looks good, delta maybe more consistent use of "heap". Thank you!

src/clusterd/src/lib.rs
teskje: Good catch! I think we actually don't want `warn` because it will just confuse/scare self-managed users. Will change to `info`.

src/compute-types/src/dyncfgs.rs
teskje: Changed to just "limit".

teskje:
TFTR!

PR #33608, opened 2025-09-17
Remove most protos from mz-expr

antiguru:
As it says on the box, remove protos and their Rust bindings from mz-expr. Also remove some `Arbitrary` implementations that aren't used.

We can't remove all `Arbitrary` implementations because we construct expressions in many tests.

Part of https://github.com/MaterializeInc/database-issues/issues/5038


teskje (approved):
🎉  Thanks!

PR #33609, opened 2025-09-17
Add handling for schema changes and schema mismtach

martykulma:
Adds schema checks during source creation and steady state.

An upstream table cannot have columns that the CDC table does not have, as that will result in negative accumulations during rewinds. We also want to ensure that the column types are compatible between the two tables.

During steady state, we will also inspect DDL changes made to the upstream table via `cdc.ddl_history`.  If columns are altered or dropped, this will cause the source to stall.  We can refine these further, if there is need.

### Motivation

fixes https://github.com/MaterializeInc/database-issues/issues/9642

### Tips for reviewer

I refactored the main processing loop in `src/storage/src/source/sql_server/replication.rs` to extract the handling of data events just to make that easier to read, the main addition to that loop is just handling the `CdcEvent::SchemaUpdate`

### Checklist

### DIFF
--- a/src/sql-server-util/src/desc.rs
+++ b/src/sql-server-util/src/desc.rs
@@ -149,7 +149,7 @@ pub struct SqlServerQualifiedTableName {
 
 impl ToString for SqlServerQualifiedTableName {
     fn to_string(&self) -> String {
-        format!("{}.{}", self.schema_name, self.table_name)
+        format!("[{}].[{}]", self.schema_name, self.table_name)
     }
 }
 
--- a/src/sql-server-util/src/inspect.rs
+++ b/src/sql-server-util/src/inspect.rs
@@ -537,6 +537,8 @@ impl DDLEvent {
     ///
     /// See <https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-ver17>
     pub fn is_compatible(&self) -> bool {
+        // TODO (maz): This is currently a basic check that doesn't take into account type changes.
+        // At some point, we will need to move this to SqlServerTableDesc and expand it.
         let mut words = self.ddl_command.split_ascii_whitespace();
         match (
             words.next().map(str::to_ascii_lowercase).as_deref(),


test/sql-server-cdc-old-syntax/alter-table-after-source.td
```diff
@@ -126,10 +126,8 @@ ALTER TABLE remove_column DROP COLUMN f2;
 
 INSERT INTO remove_column VALUES (3);
 
-# TODO: This should fail
-#> SELECT * from remove_column;
-#2
-#3
+! SELECT * from remove_column;
+container:Incompatible schema change
```
def-:
```suggestion
contains:Incompatible schema change
```

def- (changes requested):

def- (approved):
Test changes lgtm. Relevant nightlies: https://buildkite.com/materialize/nightly/builds/13448

martykulma:
moving back to draft as I missed some test updates

martykulma:
nightlies be happy now: https://buildkite.com/materialize/nightly/builds/13462

src/sql-server-util/src/desc.rs
```diff
@@ -147,6 +147,12 @@ pub struct SqlServerQualifiedTableName {
     pub table_name: Arc<str>,
 }
 
+impl ToString for SqlServerQualifiedTableName {
+    fn to_string(&self) -> String {
+        format!("{}.{}", self.schema_name, self.table_name)
```
petrosagg: Do we need to do any escaping here?

src/sql-server-util/src/inspect.rs
```diff
@@ -478,6 +522,111 @@ pub async fn get_latest_restore_history_id(
...
+}
+
+impl DDLEvent {
+    /// Returns true if the DDL event is a compatible change, or false if it is not.
+    /// This performs a naive parsing of the DDL command looking for modification of columns
+    ///  1. ALTER TABLE .. ALTER COLUMN
+    ///  2. ALTER TABLE .. DROP COLUMN
+    ///
+    /// See <https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-ver17>
+    pub fn is_compatible(&self) -> bool {
```
petrosagg, line 539: Just a note, we don't have to do anything in this PR. I think we will be forced to eventually move this under a different struct or accept the current schema as an argument. This is to support compatible schema changes like integer narrowing (int8 -> int4) or similar kind of changes.

src/sql-server-util/src/inspect.rs
```diff
@@ -478,6 +522,111 @@ pub async fn get_latest_restore_history_id(
...
+        let mut words = self.ddl_command.split_ascii_whitespace();
+        match (
+            words.next().map(str::to_ascii_lowercase).as_deref(),
+            words.next().map(str::to_ascii_lowercase).as_deref(),
+        ) {
+            (Some("alter"), Some("table")) => {
+                let mut peekable = words.peekable();
+                let mut compatible = true;
```
petrosagg, line 549: Bold! Given the simplicity of our heuristic bellow I would make this mark `ALTER TABLE` statements as incompatible unless proven otherwise. In the documentation link for `ALTER TABLE` you provided (thank you for that!) I see more changes that could be problematic, like making a column nullable.

src/sql-server-util/src/inspect.rs
martykulma: I'll add a comment.

src/sql-server-util/src/desc.rs
martykulma: Not using it in any SQL statements, but just to prevent future issues, I can!

src/sql-server-util/src/inspect.rs
petrosagg: I think I mis-read the documentation. On a second read I think this heuristic catches all cases where a column is altered/dropped.

petrosagg (approved):
Approving since only minor comments remain, thanks!

martykulma:
TFTR!!

PR #33618, opened 2025-09-17
kafka sink: examine partionby function types eval

DAlperin:
### Checklist


src/sql/src/plan/statement/ddl.rs
```diff
@@ -3875,6 +3875,13 @@ fn kafka_sink_builder(
             )?;
             let expr = expr.lower_uncorrelated()?;
 
+            let function_type = expr.typ(&value_desc.typ().column_types);
+
+            if function_type.nullable {
+                scx.catalog
+                    .add_notice(PlanNotice::KafkaSinkNullablePartitionByFunction);
+            }
+
```
antiguru: Determining nullability is difficult, so we can sometimes conclude an expr is nullable, if it is not in reality. I think with the restricted set of queries that are allowed here we should be fine.

src/storage/src/sink/kafka.rs
```diff
@@ -1578,16 +1578,11 @@ fn evaluate_partition_by(partition_by: &MirScalarExpr, row: &[Datum]) -> u64 {
...
     // user-facing `CREATE SINK` docs.
     let temp_storage = RowArena::new();
-    match partition_by.eval(row, &temp_storage) {
-        Ok(hash) => match hash {
-            Datum::Int32(i) => i.try_into().unwrap_or(0),
-            Datum::Int64(i) => i.try_into().unwrap_or(0),
-            Datum::UInt32(u) => u64::from(u),
-            Datum::UInt64(u) => u,
-            _ => unreachable!(),
-        },
```
antiguru, line 1588: Are there any Kafka sinks where the partition-by expression evaluates to one of the non-uint64 types in any catalog? Then this would be a regression vs. the current state as we'd partition to 0.

DAlperin: It would be very surprising to me if so, I looked at the original PR for this feature and there has always been an explicit cast to UInt64 https://github.com/MaterializeInc/materialize/pull/33618/files#diff-c840a7fbb6a642d6cbf71dddd92ec1b6dfa86310022ead5673e8fa509ba73dc4R3871-R3874

src/sql/src/plan/notice.rs
```diff
@@ -102,6 +106,9 @@ impl fmt::Display for PlanNotice {
             PlanNotice::ReplicaDiskOptionDeprecated => {
                 write!(f, "the DISK option is deprecated and has no effect")
             }
+            PlanNotice::KafkaSinkNullablePartitionByFunction => {
+                write!(f, "the PARTITION BY function may evaluate to NULL")
```
ggevay: "PARTITION BY function" -- Is it just a function, or can it be a general expression? If the latter, then better to say "the PARTITION BY expression may evaluate to NULL". (and also above)

src/storage/src/sink/kafka.rs
ggevay: You could maybe put a `soft_assert_or_log` in the else branch that if it is `Ok`, then it is `Datum::Null`. Or maybe just extend the above code comment to mention that null is also a possibility here.

ggevay (commented):
> Perhaps the more correct solution here is to impose a constraint during planning that the expression must not be nullable, but just looking at our own test cases I am confident this will break a lot of peoples setups. I've opted to add a notice instead.

Yes, the nullability determination is unfortunately not stable enough to error based on it. Whether we can determine that a certain expression is not nullable can totally regress between releases. So, just a notice is the best we can do.

src/storage/src/sink/kafka.rs
DAlperin: good idea

src/sql/src/plan/statement/ddl.rs
mgree: Is it possible to instead dynamically detect that a `PARTITION BY` has evaluated to NULL (or otherwise generated a single partition)? To put it another way: can we introspect on what partitioning we actually ended up with?

DAlperin: We can, and we do log it later on, but I feel like its nice to warn users of the potential pitfalls.

DAlperin:
I pulled the notice for now to take some more time to investigate what the real world state of play is there.

PR #33622, opened 2025-09-18
Disable lgalloc by default

antiguru:
This aligns the default with the current configuration.


teskje:
This will have implications for self-managed, right? We'll either need to set the flag somewhere in the self-managed tooling, or hold off with making this change until swap is the default in self-managed.

PR #33633, opened 2025-09-18
Update RocksDB to 0.24

antiguru:
Update rocksdb to 0.24


ggevay (approved):
Thank you! I've verified that now I can finally build without the hacky
```
export CXXFLAGS="$CXXFLAGS -include cstdint"
```
workaround on my machine.

PR #33636, opened 2025-09-19
Make `REFRESH` scheduling handle it gracefully when the cluster is in a graceful reconfig

ggevay:
This fixes  https://github.com/MaterializeInc/database-issues/issues/9710. It does the 2. approach from the issue.

(We could make these feature interact in a nicer way, but REFRESH is probably getting deprecated, so I'd like to avoid putting more effort into this now.)

Btw., @jubrad, is it expected that the error does _not_ occur with `wait for`; it only occurs with `wait until`. (Also, there is no test for this error.)

### Motivation

  * This PR fixes a recognized bug: https://github.com/MaterializeInc/database-issues/issues/9710

### Checklist

### DIFF
--- a/src/adapter/src/coord/cluster_scheduling.rs
+++ b/src/adapter/src/coord/cluster_scheduling.rs
@@ -7,7 +7,6 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use crate::coord::{Coordinator, Message};
 use itertools::Itertools;
 use mz_audit_log::SchedulingDecisionsWithReasonsV2;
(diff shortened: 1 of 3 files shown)


jubrad (approved):
Thanks!

jubrad:
> Btw., @jubrad, is it expected that the error does not occur with wait for; it only occurs with wait until. Also, there seems to be no tests for this error, as far as I can see. I just tested this manually. Should I add a test?

I believe it should occur with both 🤷 

If there's an easy way to add a test that might be nice, but probably a big deal.

ggevay:
Hmm, then we seems to have a bug. Opened an issue: https://github.com/MaterializeInc/database-issues/issues/9718

PR #33637, opened 2025-09-19
Binary function cleanup

antiguru:
Changes the names of binary function names to be canonical wrt the structs implementing the functions. Add `From` implementations for functions based on the `sqlfunc` macro. Fix some doc tests.

### DIFF
--- a/src/expr/src/scalar/func.rs
+++ b/src/expr/src/scalar/func.rs
@@ -2954,9 +2954,9 @@ pub enum BinaryFunc {
     AddInt16(AddInt16),
     AddInt32(AddInt32),
     AddInt64(AddInt64),
-    AddUInt16(AddUint16),
-    AddUInt32(AddUint32),
-    AddUInt64(AddUint64),
+    AddUint16(AddUint16),
+    AddUint32(AddUint32),
+    AddUint64(AddUint64),
     AddFloat32(AddFloat32),
     AddFloat64(AddFloat64),
     AddInterval,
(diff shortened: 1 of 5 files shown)


src/expr/src/scalar/func/binary.rs
```diff
@@ -514,55 +529,6 @@ mod test {
         //   which works because most don't look at the type. We should fix this
         //   and pass expected column types.
 
-        check(
```
antiguru, line 517: Removing the tests because they're now testing the generated implementation on both sides, i.e., there is no value in the tests.

src/expr/src/scalar/func.rs
```diff
@@ -2954,9 +2954,9 @@ pub enum BinaryFunc {
     AddInt16(AddInt16),
     AddInt32(AddInt32),
     AddInt64(AddInt64),
-    AddUInt16(AddUint16),
-    AddUInt32(AddUint32),
-    AddUInt64(AddUint64),
+    AddUint16(AddUint16),
+    AddUint32(AddUint32),
+    AddUint64(AddUint64),
```
teskje, line 2959: According to a quick grep, we are using `UInt` in more places than `Uint` (1929 vs 717), so perhaps we should canonicalize to the former instead?

antiguru: It's the sql types that use the `UInt` pattern. Unary functions use `Uint`. I'd prefer the latter because it avoid special-casing the name in the sqlfunc macro.

teskje: (`Datum` also uses `UInt`.)

Why do we need to special case the name with `UInt`? Are we converting CamelCase to snake_case there?

antiguru: We're converting snake_case to CamelCase, using the function name as input. So we'd need to rename functions like `add_u_int16` or so, which is equally disturbing ;-)

PR #33653, opened 2025-09-19
Hoist the sqlfunc types into BinaryFunc

antiguru:
This is a massive change, sorry. It absorbs all currently defined `LazyBinaryFunc`s into `BinaryFunc`, gettin much closer to defining it automatically. The changes are the corresponding fallout. Where we could name a variant directly in the past, it is now a struct, so we need to construct the `BinaryFunc` from the struct instead of just naming a variant.


frankmcsherry (approved):
We discussed in the office with @mgree, and concluded that the direction is good, if there are errors CI will be great at catching them, and that if there is any further work it should be easy to move this approach forward.

PR #33668, opened 2025-09-22
CI: Nightly fixes (2025-09-22)

def-:
See individual commits

Test run: https://buildkite.com/materialize/nightly/builds/13497
### Checklist

### DIFF
--- a/ci/mkpipeline.py
+++ b/ci/mkpipeline.py
@@ -873,10 +873,10 @@ def add_cargo_test_dependency(
         return
 
     for step in steps(pipeline):
-        if step.get("id") == "cargo-test":
-            step["depends_on"] = "build-x86_64"
-        if step.get("id") == "miri-test":
-            step["depends_on"] = "build-aarch64"
+        if step.get("id") in ("cargo-test", "miri-test"):
+            step["depends_on"] = (
+                "build-x86_64" if "x86" in step["agents"]["queue"] else "build-aarch64"
+            )
 
 
 def remove_dependencies_on_prs(


misc/python/materialize/mzcompose/composition.py
```diff
@@ -1571,7 +1571,7 @@ def await_mz_deployment_status(
         timeout: int | None = None,
         sleep_time: float | None = 1.0,
     ) -> None:
-        timeout = timeout or (1800 if ui.env_is_truthy("CI_COVERAGE_ENABLED") else 900)
+        timeout = timeout or (2400 if ui.env_is_truthy("CI_COVERAGE_ENABLED") else 1200)
```
def-: @aljoscha This started pretty reliably requring > 900 s now, might be interesting enough to look into? (my current interpretation is that we just have too many objects in the specific test, and they take a while)

misc/python/materialize/mzcompose/services/polaris.py
```diff
@@ -50,7 +50,8 @@ def __init__(
         self,
         name: str = "polaris",
         image: str = "apache/polaris",
-        tag: str = "latest",
+        # Fails with 1.1.0-incubating
+        tag: str = "1.0.1-incubating",
```
def-, line 54: @DAlperin Do we care about this failing with 1.1.0 as an Mz-compatibility problem? My interpretation was that we'd just have to change the test harness to account for changes instead, so I didn't open an issue. In general we want tests to use specific versions so that they don't start failing in CI suddenly.

ggevay (approved):

PR #33669, opened 2025-09-22
test: try to unflake `divergent-dataflow-cancellation.td`

teskje:
This test is sometimes flaky, presumably because the queries that are supposed to ensure that all Timely dataflow state has been dropped get confused by the ad-hoc dataflows created for said queries.

Joining against other introspection sources to find and filter ad-hoc dataflows is unreliable, because the different introspection sources are not necessarily in sync. For example, it can happen that a dataflow's entries in `mz_arrangement_batches_raw` stay around for longer than its entries in `mz_dataflow_operators_per_worker`, making it impossible to identify the dataflow these arrangement batches belong to.

Instead this PR attempts to fix the flaking by making sure all introspection queries performed in the test are fast-path queries. Those don't create ad-hoc dataflows, so no confusion should occur.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9720

### Tips for reviewer

I couldn't reproduce the flake locally, so it's possible that there is another cause for it. If that's the case, the output from the changed queries should be more useful for debugging than the current output.

### Checklist

### DIFF
--- a/test/testdrive/divergent-dataflow-cancellation.td
+++ b/test/testdrive/divergent-dataflow-cancellation.td
@@ -67,73 +67,33 @@ contains: canceling statement due to statement timeout
 
 > COMMIT
 
-# Confirm that all dataflows are now cancelled
+# Confirm that all dataflows are now cancelled.
+#
+# NOTE: It is important that the below queries are all fast-path queries.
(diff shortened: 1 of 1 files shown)


teskje:
TFTR!

PR #33672, opened 2025-09-22
adapter: filter our non-active storage collections in caught_up checks

aljoscha:
This filters out collections that are already "done", meaning they have an empty write frontier.

Fixes https://github.com/MaterializeInc/database-issues/issues/9721


src/adapter/src/coord/caught_up.rs
```diff
@@ -257,12 +257,30 @@ impl Coordinator {
...
+            .filter(|id| {
+                let frontiers = active_storage_collections.get(id);
+                match frontiers {
+                    Some(frontiers) => !frontiers.write_frontier.is_empty(),
+                    None => {
+                        // Not "active", so we don't care here.
+                        false
+                    }
+                }
+            })
```
teskje: Is it a bug that `controller.storage.active_ingestion_exports()` returns exports that have an empty frontier? The methods documentation says "Returns ingestion exports running on this instance." So I guess the question is whether an export that has advanced to the empty frontier is considered "running"?

aljoscha:
@teskje I pushed a change, as you suggested, PTAL 🙏

PR #33674, opened 2025-09-22
orchestratord test: Set version properly

def-:
As noticed by @alex-hunt-materialize : https://materializeinc.slack.com/archives/C07PN7KSB0T/p1758548841204599

Test run: https://buildkite.com/materialize/nightly/builds/13504
### Checklist

### DIFF
--- a/test/orchestratord/mzcompose.py
+++ b/test/orchestratord/mzcompose.py
@@ -52,8 +52,8 @@ SERVICES = [
 LOCAL_TAG = "orchestratord-local"
 
 
-def get_image(image: dict[str, Any], tag: str | None) -> str:
-    return f'{image["image"].split(":")[0]}:{tag or LOCAL_TAG}'
+def get_image(image: str, tag: str | None) -> str:
+    return f'{image.rsplit(":", 1)[0]}:{tag or LOCAL_TAG}'
 
 
 def get_orchestratord_data() -> dict[str, Any]:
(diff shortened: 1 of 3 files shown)


alex-hunt-materialize (approved):

test/orchestratord/mzcompose.py
```diff
@@ -24,29 +24,36 @@
...
-            time.sleep(10)
-    return tag
+# We can't use the real mzbuild tag because it has a different fingerprint for
+# environmentd/clusterd/balancerd and the orchestratord depends on them being
+# identical.
+LOCAL_TAG = "orchestratord-local"
+
+
+def get_image(image: dict[str, Any], tag: str | None) -> str:
+    return f'{image["image"].split(":")[0]}:{tag or LOCAL_TAG}'
```
alex-hunt-materialize: Nitpick, it feels weird to pass the whole `service` dict to this function, especially with the parameter called `image`. Perhaps it would be better to call it something else (`service`?) or change to only passing in the image field as a `str`.

Additionally, we probably want to use `.rsplit(":", 1)[1]` instead, since image urls can contain a port. For example, the code as written would behave incorrectly for `127.0.0.1:5000/myimage:v1.2.3`

test/orchestratord/mzcompose.py
```diff
@@ -711,7 +745,9 @@ def workflow_default(c: Composition, parser: WorkflowArgumentParser) -> None:
         definition["secret"] = materialize_setup[1]
         definition["materialize"] = materialize_setup[2]
 
-    definition["operator"]["operator"]["tag"] = get_tag(args.tag)
+    definition["operator"]["operator"]["image"]["tag"] = get_image(
+        c.compose["services"]["orchestratord"], args.tag
+    ).split(":")[1]
```
alex-hunt-materialize: Same here about using rsplit.

test/orchestratord/mzcompose.py
def-: Done

PR #33682, opened 2025-09-22
Docs: remove clock load generator source (self-managed v25.2)

materializekatrina:
### Motivation
The clock load generator has been causing confusion and is no longer recommended, as the mz_now() approach is the preferred strategy. Additionally, a breaking change is planned for the clock load generator in the near future. To avoid further issues for users, we’re removing it from the documentation for now, with the option to revisit if the feature becomes more stable. See [Slack thread](https://materializeinc.slack.com/archives/C08A7H11KP0/p1758549596045639) for additional context.


kay-kim (approved):
LGTM!

PR #33690, opened 2025-09-23
testdrive: bump kafka timeouts to 60s

martykulma:
The default timeout used to be on the order of minutes, but was moved down to 20s as part of improving the speed of testing. For kafka in the testdrive tests, it might be too low as we're seeing a lot of deadline expired failures leading to missing results.

Alternatively, I could modify the min timeout set in `kafka-verify-data` (which is 15s).  Much broader scope and I'm not sure if we want to necessarily make the minimum timeout a high value.

### Motivation

address test failures due to kafka-verify-timeout https://github.com/MaterializeInc/database-issues/issues/9688

### Checklist

### DIFF
--- a/src/testdrive/src/action/kafka/verify_data.rs
+++ b/src/testdrive/src/action/kafka/verify_data.rs
@@ -166,11 +166,11 @@ pub async fn run_verify_data(
         .context("subscribing to kafka topic")?;
 
     let (mut stream_messages_remaining, stream_timeout) = match partial_search {
-        Some(size) => (size, state.default_timeout),
+        Some(size) => (size, state.timeout),
         None => (expected_messages.len(), Duration::from_secs(15)),
     };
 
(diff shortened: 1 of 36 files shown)


martykulma:
while not a guarantee of fix, the runs I kicked off were all successful.  I think we won't know for sure until we make the change and let normal CI operations run for a while.

https://buildkite.com/materialize/nightly/builds/13542
https://buildkite.com/materialize/nightly/builds/13541
https://buildkite.com/materialize/nightly/builds/13540
https://buildkite.com/materialize/nightly/builds/13539
https://buildkite.com/materialize/nightly/builds/13538

martykulma:
> How about setting `$ set-sql-timeout duration=60s` in the relevant .td files instead? I'd be ok with doing it in `*kafka*.td`

Can do. I think I have to modify `kafka-verify-data` as it looks like it uses `state.default_timeout` and `$ set-sql-timeout` sets `state.timeout`.

PR #33714, opened 2025-09-25
Auth: migrate hash string to be directly compatible with pg/pgbouncer

DAlperin:
When I originally wrote auth I used a postgres inspired string format to
store the hashes, but didn't index on making it 1:1 compatible since we
had originally decided not to surface the hash to the user... That
didn't last.

Notion (and perhaps others) want to be able to use pgbouncer. It would
be easy enough to give instructions/a script to make the existing hash
format into the one pgbouncer wants in its config file but... we could
just not do that.

(This PR also adds platform-checks for passwords to ensure functionality
across versions)

### Checklist


def- (commented):
Thanks for adding tests!
Nightly triggered: https://buildkite.com/materialize/nightly/builds/13582

DAlperin:
oops had to rebase https://buildkite.com/materialize/nightly/builds/13593

DAlperin:
@def- I've had to do some unfortunate hacks to get around the lack of auth in older versions. Can you take a look when you get a chance and see if there is a better way to do what I've done?

misc/python/materialize/checks/all_checks/roles.py
```diff
@@ -107,3 +109,95 @@ def validate(self) -> Testdrive:
...
+            dedent(
+                """
+                # Validate all auth users exist
+                > SELECT name FROM mz_roles WHERE name LIKE 'auth_user%' ORDER BY name;
+                auth_user1
+                auth_user2
+                auth_user3
+                auth_user_nopass
+
+                # Test connection with user2
```
def-: We should test with all users during the validate phase.

misc/python/materialize/checks/mzcompose_actions.py
```diff
@@ -70,6 +70,14 @@ def execute(self, e: Executor) -> None:
         image = f"materialize/materialized:{self.tag}" if self.tag is not None else None
         print(f"Starting Mz using image {image}, mz_service {self.mz_service}")
 
+        listeners_config_path = (
+            f"{MZ_ROOT}/src/materialized/ci/listener_configs/no_auth.json"
+        )
+        if (self.tag and self.tag >= MzVersion.parse("v0.159.0-dev")) or not self.tag:
+            listeners_config_path = (
+                f"{MZ_ROOT}/src/materialized/ci/listener_configs/mixed_auth.json"
+            )
```
def-, line 80: This might conflict with some changes @SangJunBak is working on, but I guess we can merge it.

misc/python/materialize/mzcompose/services/materialized.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, image: str | None = None):
 
         config: ServiceConfig = {
             "mzbuild": name,
-            "ports": [6875, 6876, 6877, 6878, 26257],
+            "ports": [6875, 6876, 6877, 6878, 6885, 6895, 26257],
```
def-: Comment would be nice about what these ports do. Do they always exist? If not, have a `ports` parameter and pass it in for platform-checks and other tests that require these extra ports.

misc/shlib/shlib.bash
```diff
@@ -268,6 +268,10 @@ trufflehog_jq_filter_common() {
       .Raw != "postgresql://materialize:AbC123dEf@ep-cool-darkness-123456.us-east-2.aws.neon.tech:5432" and
       .Raw != "d3aa325086974cdfb3912f28e5a8c168" and
       .Raw != "jdbc:postgresql://postgres:5432/postgres" and
+      (.Raw | test("^postgres://auth_user[0-9]:password[0-9]@[^:]+:68$") | not) and
```
def-: Ending on 68 is weird, but I guess trufflehog was complaining like that?

DAlperin:
Stripped the test changes out since @SangJunBak landed the same thing here https://github.com/MaterializeInc/materialize/pull/33736

src/auth/src/hash.rs
```diff
@@ -378,29 +377,41 @@ mod tests {
...
-        // Format: SCRAM-SHA-256$<iterations>:<salt>$<client_key>:<server_key>
+        // Format: SCRAM-SHA-256$<iterations>:<salt>$<stored_key>:<server_key>
         let parts: Vec<&str> = hashed_password.split('$').collect();
         assert_eq!(parts.len(), 3);
         let key_parts: Vec<&str> = parts[2].split(':').collect();
         assert_eq!(key_parts.len(), 2);
-        let client_key = BASE64_STANDARD
+        let stored_key = BASE64_STANDARD
             .decode(key_parts[0])
             .expect("decode client key");
```
SangJunBak: nit: should be "decode stored key"

src/auth/src/hash.rs
```diff
@@ -267,7 +263,8 @@ struct ScramSha256Hash {
     /// The server key
     server_key: [u8; SHA256_OUTPUT_LEN],
     /// The client key
-    client_key: [u8; SHA256_OUTPUT_LEN],
+    // client_key: [u8; SHA256_OUTPUT_LEN],
```
SangJunBak: nit: Should remove the bottom comment and change the top to "The stored key"

src/catalog/src/durable/upgrade/v77_to_v78.rs
```diff
@@ -0,0 +1,85 @@
...
+    let Ok(first_key) = BASE64_STANDARD.decode(first_key_b64) else {
+        return None;
+    };
+    if first_key.len() != 32 {
+        return None;
+    }
+
+    let stored_key = openssl::sha::sha256(&first_key);
+    let stored_key_b64 = BASE64_STANDARD.encode(stored_key);
+
```
SangJunBak, line 41: I wonder if it's possible to add logging whenever any of these steps fail

SangJunBak:
Sorry for the late review! 

For testing, we should probably test the sasl version (I think we only test for basic auth rn).
For now, we can also just change Password to Sasl in `testdrive.json`, and change the base_version of PasswordAuth(Check). I can create a followup issue and do it to get the test for basic auth!

DAlperin:
@SangJunBak I think for now we should actually do the inverse. We can't easily switch the mode to SASL since the older versions running the upgrade checks won't parse that config file.

misc/python/materialize/checks/mzcompose_actions.py
```diff
@@ -211,13 +221,15 @@ def __init__(
     def execute(self, e: Executor) -> None:
         c = e.mzcompose_composition()
 
+        print(c.compose)
```
def-: Should be removed?

misc/python/materialize/checks/mzcompose_actions.py
```diff
@@ -211,13 +221,15 @@ def __init__(
...
         # Don't fail since we are careful to explicitly kill and collect logs
         # of the services thus started
         with c.override(Materialized(name=self.mz_service), fail_on_new_service=False):
             c.kill(self.mz_service, wait=True)
 
             if self.capture_logs:
                 c.capture_logs(self.mz_service)
+        print(c.compose)
```
def-: same

misc/python/materialize/checks/all_checks/password_auth.py
```diff
@@ -53,3 +53,42 @@ def validate(self) -> Testdrive:
...
+                """,
+            ]
+        ]
+
+    def validate(self) -> Testdrive:
+        return Testdrive(
+            dedent(
+                """
+                $ postgres-execute connection=postgres://user2:password2@${testdrive.materialize-sasl-sql-addr}
+                SELECT * FROM materialize.schema2.t1
```
def-, line 91: What is this validating? Just that the connection works, but not the result I guess?

DAlperin: yeah

DAlperin: yeah

DAlperin: it mirrors the test above which does the same thing

PR #33716, opened 2025-09-25
adapter: set `mz_cluster{_replica}s.disk` to false with swap

teskje:
This PR changes the meaning of the existing `disk` field in `mz_clusters` and `mz_cluster_replicas` from "whether this cluster/replica has a disk or swap" to "whether this cluster/replica has a disk". I.e. the change is that the field shows `false` for swap-enabled clusters/replicas.

The motivation is two-fold:
* Whether a replica gets swap or not depends on the node it gets scheduled on. Pretending that swap-enabled replicas always get swap would be lying.
* More importantly, the console needs a way to know which clusters still use the old way of spilling to disk, to adjust its metrics UI accordingly. This change provides that way.

### Motivation

Part of https://github.com/MaterializeInc/database-issues/issues/9692

See [Slack thread](https://materializeinc.slack.com/archives/C06GZ7GBKB5/p1758719170479669).

### Checklist

### DIFF
--- a/src/adapter/src/catalog/state.rs
+++ b/src/adapter/src/catalog/state.rs
@@ -2339,12 +2339,17 @@ impl CatalogState {
 
     /// Return whether the given replica size requests a disk.
     ///
+    /// Note that here we treat replica sizes that enable swap as _not_ requesting disk. For swap
+    /// replicas, the provided disk limit is informational and mostly ignored. Whether an instance
+    /// has access to swap depends on the configuration of the node it gets scheduled on, and is
+    /// not something we can know at this point.
+    ///
     /// # Panics
     ///
     /// Panics if the given size doesn't exist in `cluster_replica_sizes`.
     pub(crate) fn cluster_replica_size_has_disk(&self, size: &str) -> bool {
         let alloc = &self.cluster_replica_sizes.0[size];
-        alloc.disk_limit != Some(DiskLimit::ZERO)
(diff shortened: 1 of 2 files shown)


src/adapter/src/catalog/state.rs
```diff
@@ -2339,12 +2339,17 @@ impl CatalogState {
...
+    /// has access to swap depends on the configuration of the node it gets scheduled on, and is
+    /// not something we can know at this point.
+    ///
     /// # Panics
     ///
     /// Panics if the given size doesn't exist in `cluster_replica_sizes`.
     pub(crate) fn cluster_replica_size_has_disk(&self, size: &str) -> bool {
         let alloc = &self.cluster_replica_sizes.0[size];
-        alloc.disk_limit != Some(DiskLimit::ZERO)
+        !alloc.swap_enabled && alloc.disk_limit != Some(DiskLimit::ZERO)
```
ggevay, line 2352: The [linked slack msg](https://materializeinc.slack.com/archives/C06GZ7GBKB5/p1758719170479669) says

> A nice thing about this is that it will also work for self-managed clusters that have both swap and a disk.

but wouldn't the `!alloc.swap_enabled &&` make us not show the disk graph for clusters that have both a disk and swap? Maybe I'm misunderstanding what the "will also work" exactly refers to in the above sentence?

teskje: Ah, yes that comment is dumb. We don't support swap and disk at the same time. Fixed!

teskje:
TFTR!

PR #33726, opened 2025-09-26
Complain when encountering collation plans

antiguru:
We're representing collation reductions as a join of reductions. If we encounter one of the plans after enabling the reduce reduction, it is at least unexpected.


frankmcsherry (approved):
Looks great, and thanks!

PR #33729, opened 2025-09-28
storage: fix running status for source exports

teskje:
This PR fixes a bug where source exports (aka. subsources, aka. source tables) would get stuck in "starting" status when they weren't producing any data after installation. This caused confusion, both for users and for our 0dt caught-up checks.

The fix implemented here is to make sure the various source operators always emit a "running" status for all source exports, not just the root ingestion. The approach works, but is quite verbose and adds a new peculiarity one has to keep in mind when implementing a new source. Ideally this would somehow be handled automatically by the source ingestion pipeline instead. Perhaps something to pick up as part of https://github.com/MaterializeInc/database-issues/issues/5986.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9721

### Checklist

### DIFF
--- a/src/storage/src/source/source_reader_pipeline.rs
+++ b/src/storage/src/source/source_reader_pipeline.rs
@@ -365,30 +365,6 @@ where
 
                 while let Some((cap, data)) = input.next() {
                     for (message, _, _) in data.iter() {
-                        let status = match &message {
-                            Ok(_) => HealthStatusUpdate::running(),
-                            // All errors coming into the data stream are definite.
-                            // Downstream consumers of this data will preserve this
-                            // status.
-                            Err(error) => HealthStatusUpdate::stalled(
-                                error.to_string(),
-                                Some(
-                                    "retracting the errored value may resume the source"
-                                        .to_string(),
-                                ),
-                            ),
-                        };
-
-                        let status = HealthStatusMessage {
-                            id: Some(id),
-                            namespace: C::STATUS_NAMESPACE.clone(),
-                            update: status,
-                        };
-                        if last_status.as_ref() != Some(&status) {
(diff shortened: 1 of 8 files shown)


src/storage/src/source/source_reader_pipeline.rs
```diff
@@ -365,30 +365,6 @@ where
 
                 while let Some((cap, data)) = input.next() {
                     for (message, _, _) in data.iter() {
-                        let status = match &message {
-                            Ok(_) => HealthStatusUpdate::running(),
```
teskje, line 369: This is where we would previously transition an export to "running" status once the first data update flowed through. The source operators now emit the "running" statuses for all exports, so afaict this code is not needed anymore.

Btw, I initially tried a smaller fix in this operator, where I would emit a "running" status in response to frontier updates instead of data updates. Unfortunately this doesn't work because we are still in the "source time" part of the dataflow, so when the upstream has no activity, there won't be frontier updates either.

test/testdrive/status-history.td
teskje, line 1: I totally rewrote this test. It now also covers the previously broken "running" reporting behavior and is hopefully more readable.

test/testdrive/status-history.td
```diff
@@ -9,189 +9,448 @@
...
+kafka          paused
+kafka_tbl      paused
+pg             paused
+pg_tbl         paused
+mysql          paused
+mysql_tbl      paused
+auction_progress  running
+kafka_progress    running
+pg_progress       running
+mysql_progress    running
```
teskje, line 112: These show up in `mz_source_statuses` and are always "running". Seems like another bug maybe?

petrosagg (approved):
LGTM!

PR #33736, opened 2025-09-29
Add testdrive self-managed password authentication tests on upgrade

SangJunBak:
Add self-managed password authentication tests

Add tests to test self managed password auth for past patch releases upgrading to the latest self managed patch release

### Motivation
Followup for incident 697

### Checklist

### DIFF
--- a/misc/python/materialize/checks/scenarios_upgrade.py
+++ b/misc/python/materialize/checks/scenarios_upgrade.py
@@ -9,6 +9,7 @@
 
 
 from dataclasses import dataclass
+
 from materialize.checks.actions import Action, Initialize, Manipulate, Sleep, Validate
 from materialize.checks.checks import Check
 from materialize.checks.executors import Executor


ci/nightly/pipeline.template.yml
```diff
@@ -391,6 +391,14 @@ steps:
...
               args: [-m=long, test/cloudtest/test_full_testdrive.py, --no-test-parallelism]
         sanitizer: skip
+      - id: self-managed-password-auth-testdrive
+        label: "Self managed password auth testdrive"
+        depends_on: build-aarch64
+        timeout_in_minutes: 180
+        parallelism: 2
+        agents:
+          queue: linux-aarch64-small
+        sanitizer: skip
```
def-: Does this have to run in a real self-managed environment? Otherwise can use hetzner instead of linux-aarch64-small agent.

Also needs something like this:
```
        plugins:
          - ./ci/plugins/mzcompose:
              composition: password-auth
```

SangJunBak:
@def- SQLLogictests failed bc. `enabled_password_auth` creates a code path that handles role creation differently. **Not sure if I should adjust these tests or disable password auth for them.** I think role.slt and role_create.slt can just be updated since all it tests is that the `LOGIN`, `SUPERUSER` syntax is invalid for cloud. However `pg_catalog_user.slt` explictly tests against a code path for email domains.

test/platform-checks/mzcompose.py
```diff
@@ -62,6 +62,11 @@ def create_mzs(
     external_metadata_store: bool = True,
     external_blob_store: bool = True,
 ) -> list[TestdriveService | Materialized]:
+
+    # TODO (SangJunBak): Make the number of Materialized services dynamic
```
SangJunBak: Hypothetically, is there any cost to just creating a hundred `mz_{i}` services? 

Another thought I had is instead of manually labelling every "mz_service" in these actions (i.e. Validate, Manipulate, etc.) for upgrade tests, we can create a higher level action that does it for us. We can maybe dynamically create a Materialized service and put it in `self.compositions` in this higher level action

def-: Yes, there is quite some overhead in having Docker spin them all up. We had that previously in some tests for testdrive services and I removed it as part of making CI faster.

def-:
> Not sure if I should adjust these tests or disable password auth for them

Would be great to keep the tests, so disabling password auth for them sounds good.

misc/python/materialize/checks/scenarios_upgrade.py
```diff
@@ -509,3 +510,69 @@ def actions(self) -> list[Action]:
...
+            ]
+
+        actions = [
+            StartMz(
+                self,
+                tag=self.base_version(),
+                mz_service="mz_1",
+            ),
+            Initialize(self, mz_service="mz_1"),
+            Manipulate(self, phase=1, mz_service="mz_1"),
```
def-: It would be great if we could run upgrades inbetween the Initialize/Manipulate1/Manipulate2 phases too.

misc/python/materialize/mzcompose/services/materialized.py
```diff
@@ -347,7 +348,7 @@ def __init__(
             {
                 "depends_on": depends_graph,
                 "command": command,
-                "ports": [6875, 6876, 6877, 6878, 26257],
+                "ports": [6875, 6876, 6877, 6878, 6880, 26257],
```
def-, line 351: Is this relevant for all tests or only platform-checks?

test/platform-checks/mzcompose.py
```diff
@@ -62,6 +62,11 @@ def create_mzs(
...
+    # based on the number of generations we want to use in a platform check.
+    mzs = [f"mz_{i}" for i in range(1, 12)]
```
def-: Why do we need 12? We can have 2 and switch between them. mz_1 -> mz_2 -> mz_1 -> mz_2 -> ...

misc/python/materialize/checks/scenarios_upgrade.py
```diff
@@ -509,3 +510,69 @@ def actions(self) -> list[Action]:
             ),
             Validate(self),
         ]
+
+
+class SelfManagedv25_2_Upgrade(Scenario):
+    """
+    Upgrade from the oldest v25.2 patch release to the latest v25.2 patch release.
```
def-: We also have to upgrade to the current main state after that. We don't want to have tests which test an old state of Materialize. This test is supposed to got from v25.2.0 -> v25.2.1 -> ... -> v25.2.18 (or whatever the latest is) -> main

misc/python/materialize/mzcompose/services/materialized.py
SangJunBak: Currently this and the `listeners_config_path="...testdrive.json"` are needed just for the platform checks. 

However, thought I'd add it here because:
- No reason why non platform tests couldn't use it
- Might be helpful / lead to less boilerplate for other tests to have a consistent password port
- I wanted to add `testdrive.materialize-password-sql-addr` to the testdrive binary, I thought it made sense to also expose the password auth port as a default in `materialized` given it'd be weird to write a testdrive fragment with `testdrive.materialize-password-sql-addr` but require the test writer to plug in their own custom listener config

misc/python/materialize/checks/scenarios_upgrade.py
SangJunBak: Ah so the test already does that, I just messed up the description 😅

SangJunBak:
@def- Feedback addressed from https://github.com/MaterializeInc/materialize/pull/33736/commits/8ad14535bd05a9d9f2a0b383272e53106f756f11 onward. TFTR!

PR #33745, opened 2025-10-01
[sql-pretty] Add support for more objects

sjwiesman:
### Motivation

Add SQL pretty support for all the objects we display in the console. 

### Checklist


sjwiesman:
Looks like a bunch of tests need to be rewritten. I ~~assume~~ hope there is an easy way to do that?

ggevay:
The slts can be auto-rewritten, e.g.
```
MZ_SOFT_ASSERTIONS=true bin/sqllogictest  -- -v test/sqllogictest/privilege_checks.slt  --rewrite-results
```
But all the others have to be rewritten manually. :sadpanda: In my last PR touching the pretty-printing, rewriting tests took more time than the actual work itself. (We tried adding auto-rewriting to testdrive at some point, but there were various difficulties.)

sjwiesman:
Guess its time to try claude code 😅

ggevay:
One difficulty is that the testdrive-based tests will need version guards, because we run them on several different versions of Materialize. So, each test will need to be duplicated, one of them guarded by the materialize version being old, and the other guarded by the materialize version being new. There is an example here: https://github.com/MaterializeInc/materialize/blob/68cd1a4b52a48d48f48e2c2bd8667799fa143eaf/test/mysql-cdc-old-syntax/alter-source.td#L109

sjwiesman:
Refreshed sqllogic

@ggevay can I ask that you review this before I update testdrive? I'd hate to make all these changes then need to update them again 😄

src/sql-pretty/src/doc.rs
```diff
@@ -94,6 +94,470 @@ impl Pretty {
...
+                        include_doc,
+                        bracket(
+                            "(",
+                            comma_separate(|f| self.doc_display_pass(f), filters),
+                            ")",
+                        ),
+                    );
+                }
+                header_docs.push(include_doc);
+            }
```
ggevay, line 146: This does this:
```
 INCLUDE HEADERS              
     (                        
         NOT 'accept',        
         NOT 'content-length',
         NOT 'host',          
         NOT 'x-api-key',     
         NOT 'aaaaaaaaaaaaaa',
         NOT 'bbbbbbbbbbbb',  
         NOT 'cccccccccccc'   
     )                        
```
In other places, I think we tend to put the opening paren on the same line with the clause that it belongs to. (There are various examples by searching for `WITH (` in `doc.rs`.) So, how about something like this instead:
```
if let Some(filters) = &v.include_headers.column {
    if filters.is_empty() {
        header_docs.push(RcDoc::text("INCLUDE HEADERS"));
    } else {
        header_docs.push(bracket(
            "INCLUDE HEADERS (",
            comma_separate(|f| self.doc_display_pass(f), filters),
            ")",
        ));
    }
}
```
which does
```
INCLUDE HEADERS (
    NOT 'accept',
    NOT 'content-length',
    NOT 'host',
    NOT 'x-api-key',
    NOT 'aaaaaaaaaaaaaa',
    NOT 'bbbbbbbbbbbb',
    NOT 'cccccccccccc'
)
```

src/sql-pretty/src/doc.rs
```diff
@@ -94,6 +94,470 @@ impl Pretty {
...
+            for body in &options.bodies {
+                with_items.push(self.doc_display_pass(body));
+            }
+            for secret in &options.secrets {
+                with_items.push(self.doc_display_pass(secret));
+            }
+
+            if !with_items.is_empty() {
+                inner.push(bracket("WITH (", comma_separated(with_items), ")"));
+                inner.push(RcDoc::space());
```
ggevay: Maybe `RcDoc::line()` instead of `RcDoc::space()`? See https://github.com/MaterializeInc/materialize/blob/0f7b21216266eb57e766810dd718c36802d2e8ec/test/testdrive/webhook.td#L244-L254 as an example.

src/sql-pretty/src/doc.rs
```diff
@@ -94,6 +94,470 @@ impl Pretty {
...
+        if v.if_not_exists {
+            title.push_str(" IF NOT EXISTS");
+        }
+
+        let index_clause = if let Some(name) = &v.name {
+            self.doc_display_pass(name)
+        } else {
+            RcDoc::nil()
+        };
+        docs.push(nest_title(title, index_clause));
```
ggevay: This comes out a bit awkwardly for indexes with no names:
```
materialize=> select pretty_sql('CREATE DEFAULT INDEX ON foo;');
pretty_sql
CREATE DEFAULT INDEX  ON foo;
```
How about
```
        if let Some(name) = &v.name {
            docs.push(nest_title(title, self.doc_display_pass(name)));
        } else {
            docs.push(RcDoc::text(title));
        }
```
i.e., moving the `docs.push` into the `if` branches?

ggevay (commented):
Thank you! Looks good, just minor comments.

src/sql-pretty/src/doc.rs
sjwiesman: good catch.

sjwiesman:
@ggevay @jubrad I've updated all the relevant tests on this PR. The only failing test seems unrelated, its around Avro deserialization.

misc/python/materialize/checks/all_checks/webhook.py
```diff
@@ -111,15 +111,24 @@ def validate(self) -> Testdrive:
                 \\\\x01
                 \\\\x01\\x02\\x03\\x04
 
-                >[version>=14000] SHOW CREATE SOURCE webhook_text
+                >[version>=14000 AND version<2600700] SHOW CREATE SOURCE webhook_text
```
def-:
```suggestion
                >[14000<=version<2600700] SHOW CREATE SOURCE webhook_text
```

misc/python/materialize/checks/all_checks/webhook.py
```diff
@@ -111,15 +111,24 @@ def validate(self) -> Testdrive:
...
                 materialize.public.webhook_text "CREATE SOURCE materialize.public.webhook_text IN CLUSTER webhook_cluster FROM WEBHOOK BODY FORMAT TEXT;"
 
-                >[version>=14000] SHOW CREATE SOURCE webhook_json
+                >[version>=2600700] SHOW CREATE SOURCE webhook_text
+                materialize.public.webhook_text "CREATE SOURCE materialize.public.webhook_text\\nIN CLUSTER webhook_cluster\\nFROM WEBHOOK\\nBODY FORMAT TEXT;"
+
+                >[version>=14000 AND version<2600700] SHOW CREATE SOURCE webhook_json
```
def-:
```suggestion
                >[14000<=version<2600700] SHOW CREATE SOURCE webhook_json
```

misc/python/materialize/checks/all_checks/webhook.py
```diff
@@ -111,15 +111,24 @@ def validate(self) -> Testdrive:
...
                 materialize.public.webhook_json "CREATE SOURCE materialize.public.webhook_json IN CLUSTER webhook_cluster FROM WEBHOOK BODY FORMAT JSON INCLUDE HEADERS;"
 
-                >[version>=14000] SHOW CREATE SOURCE webhook_bytes
+                >[version>=2600700] SHOW CREATE SOURCE webhook_json
+                materialize.public.webhook_json "CREATE SOURCE materialize.public.webhook_json\\nIN CLUSTER webhook_cluster\\nFROM WEBHOOK\\nBODY FORMAT JSON\\nINCLUDE HEADERS;"
+
+                >[version>=14000 AND version<2600700] SHOW CREATE SOURCE webhook_bytes
```
def-:
```suggestion
                >[14000<=version<2600700] SHOW CREATE SOURCE webhook_bytes
```

def- (changes requested):

def-:
The failing test is very much related: https://buildkite.com/materialize/test/builds/114075#019b942b-6423-42a0-affb-0f262f7c4d40

> webhook.py:82 31:1: error: version-constraint: invalid version number 14000 AND version<2600700

See the above recommended changes.

sjwiesman:
Looks like there are a lot more to update!

sjwiesman:
Pushed some additional test fixes

def- (approved):
Thanks, triggered another run: https://buildkite.com/materialize/nightly/builds/14653

sjwiesman:
Thank you all!

PR #33751, opened 2025-10-02
[persist] Simplify compaction chunking logic (and support empty batches)

bkirwi:
Previously, we had two separate steps here: "flattening" turned the set of input batches into a set of runs, and "chunking" grouped those runs into sets of runs that could be compacted together without either violating requirements on our batch boundaries or our memory budget. This separation made sense historically, but recent changes made those requirements less trivial and the separation less useful... so this collapses the two methods together and then does a bunch of straightforward refactoring to simplify the logic.

This is mostly a noop, but it does have one useful effect for empty batches. Previously, those were discarded in the first step, but for incremental compaction it's useful to explicitly include those empty inputs in other chunks so incremental compaction will remove them from the output. Since the new logic works over batches directly instead of flattening and regrouping runs, those empty batches and their metadata are not discarded.

### Motivation

Handles the underlying issue uncovered in https://github.com/MaterializeInc/materialize/pull/33667...

### DIFF
--- a/src/persist-client/src/internal/compact.rs
+++ b/src/persist-client/src/internal/compact.rs
@@ -655,11 +655,8 @@ where
                 .compaction_memory_bound_bytes
                 .saturating_sub(in_progress_part_reserved_memory_bytes);
 
-            let ordered_runs =
-                Self::flatten_runs(&req, cfg.batch.preferred_order)?;
-
             let chunked_runs = Self::chunk_runs(
-                &ordered_runs,
+                &req,
                 &cfg,
                 &*metrics,
                 run_reserved_memory_bytes,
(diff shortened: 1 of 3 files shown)


src/dyncfg/src/lib.rs
```diff
@@ -489,7 +489,10 @@ impl ConfigUpdates {
     pub fn apply(&self, set: &ConfigSet) {
         for (name, val) in self.updates.iter() {
             let Some(config) = set.configs.get(name) else {
-                error!("config update {} {:?} not known set: {:?}", name, val, set);
+                error!(
+                    "config update {} {:?} not in known set: {:?}",
+                    name, val, set
+                );
```
DAlperin: seems fine but meant to land in this PR?

src/persist-client/src/internal/compact.rs
```diff
@@ -803,42 +788,27 @@ where
...
-            .map(|(run_id, desc, meta, parts)| (run_id.0, (run_id, *desc, *meta, &**parts)))
-            .fold(BTreeMap::new(), |mut acc, item| {
-                acc.entry(item.0).or_default().push(item.1);
-                acc
-            });
-
-        let mut grouped = grouped.into_iter().peekable();
+        let mut batch_runs: Vec<_> = req.inputs.iter().map(|x| (x.id, &*x.batch)).collect();
+        batch_runs.sort_by_key(|(id, _)| *id);
+        let mut grouped = batch_runs.into_iter();
```
DAlperin: nit: the names/comments here don't quite line up anymore

src/persist-client/src/internal/compact.rs
```diff
@@ -848,158 +818,123 @@ where
...
+                        !parts.iter().any(|r| matches!(r, RunPart::Many(_))),
+                        "unexpected out-of-order hollow run"
+                    );
+                    Either::Right(
+                        parts
+                            .iter()
+                            .map(move |p| (&batch.desc, meta, std::slice::from_ref(p))),
+                    )
+                }
+            });
```
DAlperin, line 851: We lost a comment from flatten_runs explaining why we do this here rather than let downstream consolidation handle this

src/persist-client/src/internal/datadriven.rs
```diff
@@ -190,98 +190,113 @@ mod tests {
...
-            }
-        })
+                    })
+                    .await;
+                    f
+                }
+            },
+            |f| {
+                // Return true iff we want to _exclude_ the given file.
+                let Ok(filter) = std::env::var("DATADRIVEN_FILTER") else {
```
DAlperin, line 293: 👨🏼‍🍳

src/dyncfg/src/lib.rs
bkirwi: Huh, no idea why it chose to reformat this! I'll undo it unless rustfmt insists for some reason...

bkirwi: Ah, typo fix made it slightly longer! Anyways, removed.

PR #33758, opened 2025-10-03
Rollout strategy enum

alex-hunt-materialize:
Deprecates and ignores the `inPlaceRollout` field of the Materialize CR, instead replacing it with a new `rolloutStrategy` enum field.

The enum has two variants: `WaitForRollout` (the default) and `ImmediatelyPromoteCausingDowntime`. Hopefully these new names will be more clear.

We increment the the generation in both variants, so we avoid the weird behavior we saw with `inPlaceRollout` previously. Both variants follow almost the same code path, with the exceptions of `ImmediatelyPromoteCausingDowntime` doing as its name implies, deleting the old generation first and immediately promoting the new generation.

### Motivation

Several customers have accidentally used `inPlaceRollout`, when they should not have. Even internally, this field is confusing, and it has lots of strange behavior caused by not incrementing the generation.

### Tips for reviewer

It's easier to review each commit individually.

### Checklist

### DIFF
--- a/src/cloud-resources/src/crd/materialize.rs
+++ b/src/cloud-resources/src/crd/materialize.rs
@@ -58,11 +58,12 @@ pub mod v1alpha1 {
         // Additional annotations and labels to include in the Certificate object.
         pub secret_template: Option<CertificateSecretTemplate>,
     }
-    #[derive(Clone, Debug, PartialEq, Deserialize, Serialize, JsonSchema)]
+    #[derive(Clone, Debug, Default, PartialEq, Deserialize, Serialize, JsonSchema)]
     pub enum MaterializeRolloutStrategy {
         // Default. Create a new generation of pods, leaving the old generation around until the
         // new ones are ready to take over.
         // This minimizes downtime, and is what almost everyone should use.
+        #[default]
         WaitUntilReady,
 
         // WARNING!!!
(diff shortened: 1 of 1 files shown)


alex-hunt-materialize:
@def- I've manually tested upgrades for both variants, and I've run the nightly orchestratord tests. Is there anything else we need for this? We probably want to add this to the upgrade tests once those are ready, but I don't think that should block this PR.

def-:
The terraform upgrade test, but that is currently WIP in another PR. So all good from QA side.

src/cloud-resources/src/crd/materialize.rs
```diff
@@ -58,6 +58,30 @@ pub mod v1alpha1 {
...
+        //
+        // Tear down the old generation of pods and promote the new generation of pods immediately,
+        // without waiting for the new generation of pods to be ready.
+        //
+        // This strategy should ONLY be used by customers with physical hardware who do not have
+        // enough hardware for the WaitForReady strategy. If you think you want this, please
+        // consult with Materialize engineering to discuss your situation.
+        ImmediatelyPromoteCausingDowntime,
+    }
+    impl Default for MaterializeRolloutStrategy {
```
doy-materialize: you can use `derive(Default)` and the `#[default]` variant annotation instead of deriving this by hand

alex-hunt-materialize: I could have sworn I tried that. Thanks.

PR #33760, opened 2025-10-03
compute,storage: downgrade config update logs

teskje:
This PR downgrades the "Applying configuration update" logs from `info` to `debug`. They have become extremely noisy over time as we have accumulated more config options, and especially with the introduction of the dyncfgs.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/9660

### Tips for reviewer

I would prefer to either (a) log only actual config changes or (b) send only actual config changes from envd to clusterd. But I didn't find a good way to implement (a) with our current `{Compute,Storage}Parameters` definition and (b) requires a refactor of adapter code I'm not very familiar with.

### Checklist

### DIFF
--- a/src/compute/src/compute_state.rs
+++ b/src/compute/src/compute_state.rs
@@ -65,7 +65,7 @@ use timely::progress::frontier::Antichain;
 use timely::scheduling::Scheduler;
 use timely::worker::Worker as TimelyWorker;
 use tokio::sync::{oneshot, watch};
-use tracing::{Level, debug, error, info, span, warn};
+use tracing::{Level, debug, error, info, span, trace, warn};
 use uuid::Uuid;
 
 use crate::arrangement::manager::{TraceBundle, TraceManager};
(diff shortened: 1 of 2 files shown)


antiguru (commented):
I agree with the sentiment, but just last week this was crucial to me to debug an issue. I think we should do something else here: only print the configuration that has changed.

petrosagg:
> but just last week this was crucial to me to debug an issue

But how often have you had such a debugging task? And, was the problem related to the mechanism of propagating configuration to clusters or you just wanted to see what was the value of a particular config at a particular time? The latter need can be served by our configuration service (LD) which tracks the history of each variable.

In general, I think we need a higher bar for having things be logged at info level, leaning towards things that are:
1. High value
2. Low traffic
3. Can't be deduced after the fact

I don't think the LD configs meet that bar so I'm strongly in favor of merging this PR

teskje:
Filed the issue here: https://github.com/MaterializeInc/database-issues/issues/11201

teskje:
TFTRs!

PR #33761, opened 2025-10-03
Increase default session timeout from 1 hour to 8 hours

ohbadiah:
This is supposed to get automatically renewed upon health check anyway, but some browsers might not allow the health check to run if the tab is backgrounded. 8 hours (one workday) should be ok.

See discussion [here](https://materializeinc.slack.com/archives/CU7ELJ6E9/p1759513058586999?thread_ts=1759510882.253709&cid=CU7ELJ6E9) for more context.


jasonhernandez:
This is reasonable given the context Materialize runs in.

PR #33771, opened 2025-10-06
adapter: gracefully handle missing schema in builtin migration

teskje:
During builtin item migration, it is possible to encounter a builtin collection that doesn't have a latest schema. Since this is an expected scenario, we should handle it gracefully.

### Motivation

Upgrade checks are failing because they use an in-memory persist client that doesn't have access to existing schemas. See [Slack thread](https://materializeinc.slack.com/archives/CTESPM7FU/p1759517500626729).

### Tips for Reviewers

From discussion with @antiguru: We should be more intentional about when it's fine to migrate a builtin collection by truncating it, instead of always falling back to truncating if the schema evolution doesn't succeed. We can introduce a "builtin version", akin to the catalog version, and a list of `(version, CatalogItemId)` that specifies at which versions which builtin collection should be migrated using truncation. Then for collections on that list, we unconditionally truncate those, and unconditionally apply schema evolution to the rest of them. That avoids accidentally truncating a collection we didn't mean to truncate. Though we'll still need to deal with the absence of an existing schema somehow.

Filed an issue: https://github.com/MaterializeInc/database-issues/issues/9756

### Checklist

### DIFF
--- a/src/adapter/src/catalog/open/builtin_item_migration.rs
+++ b/src/adapter/src/catalog/open/builtin_item_migration.rs
@@ -102,11 +102,10 @@ pub(crate) async fn migrate_builtin_items(
 
     // Attempt to perform schema evolution.
     //
-    // If we run into an unexpected error (i.e. any persist error other than "schema incompatible")
-    // while trying to perform schema evolution, we abort the migration process, rather than
-    // automatically falling back to replacing the persist shards. We do this to avoid accidentally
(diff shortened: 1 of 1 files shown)


antiguru (approved):
LGTM, as discussed and captured in the description.

bkirwi (approved):

teskje:
TFTRs!

PR #33785, opened 2025-10-06
Document unicode normalization function

ohbadiah:
With #33507, we introduced a new sql function for unicode normalization, but I forgot to include documentation. This PR introduces those docs.

I built the Hugo site locally, and it looks good to me.

### DIFF
--- a/doc/user/content/sql/functions/normalize.md
+++ b/doc/user/content/sql/functions/normalize.md
@@ -10,15 +10,12 @@ menu:
 
 ## Signatures
 
-```mzsql
-normalize(str)
-normalize(str, form)
-```
+{{% include-example file="examples/normalize" example="syntax" %}}
 
 Parameter | Type | Description
 ----------|------|------------
 _str_ | [`string`](../../types/string) | The string to normalize.
-_form_ | keyword | The Unicode normalization form: `NFC`, `NFD`, `NFKC`, or `NFKD` (unquoted keywords). _Defaults to `NFC`_.
+_form_ | keyword | The Unicode normalization form: `NFC`, `NFD`, `NFKC`, or `NFKD` (unquoted, case-insensitive keywords). _Defaults to `NFC`_.
 
 ### Return value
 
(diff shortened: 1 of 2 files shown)


doc/user/content/sql/functions/normalize.md
```diff
@@ -0,0 +1,101 @@
...
+
+```mzsql
+normalize(str)
+normalize(str, form)
+```
+
+Parameter | Type | Description
+----------|------|------------
+_str_ | [`string`](../../types/string) | The string to normalize.
+_form_ | keyword | The Unicode normalization form: `NFC`, `NFD`, `NFKC`, or `NFKD` (unquoted keywords). _Defaults to `NFC`_.
```
kay-kim: Probably, the *case-insensitive* sentence below actually belongs here.

doc/user/content/sql/functions/normalize.md
```diff
@@ -0,0 +1,101 @@
...
+    parent: 'sql-functions'
+---
+
+`normalize` converts a string to a specified Unicode normalization form.
+
+## Signatures
+
+```mzsql
+normalize(str)
+normalize(str, form)
```
kay-kim: Oh ... we're just starting to slowly (and I mean slowly) move towards taking the 1st step towards being able to test our docs examples by having the syntax and examples be in a yaml file.
See https://github.com/MaterializeInc/materialize/blob/main/doc/user/data/examples/alter_sink.yml

Then to include in the nomalize.md file, we would just do
{{% include-example file="examples/alter_sink"
example="syntax" %}}
... (the .yml is excluded from the file name, and example="<name>")

This way, even before we get the testing all set up, at least when people change the syntax, it improves the chances that they'll update the examples too.

If amenable, could you try incorporating into the new format?

ohbadiah: Sure thing!

doc/user/content/sql/functions/normalize.md
ohbadiah: Makes sense, thanks.

kay-kim (approved):
LGTM!  Thank you!

PR #33786, opened 2025-10-07
Allow null values in multi-dimensional arrays

antiguru:
Allow constructing multi-dimensional arrays with null values instead of panicking. Similarly to PostgreSQL, treat null elements as zero-dimensional arrays.

Fixes MaterializeInc/database-issues#9757

### DIFF
--- a/src/expr/src/scalar/func/variadic.rs
+++ b/src/expr/src/scalar/func/variadic.rs
@@ -23,7 +23,7 @@ use md5::Md5;
 use mz_lowertest::MzReflect;
 use mz_ore::cast::{CastFrom, ReinterpretCast};
 use mz_pgtz::timezone::TimezoneSpec;
-use mz_repr::adt::array::{ArrayDimension, InvalidArrayError};
+use mz_repr::adt::array::{ArrayDimension, ArrayDimensions, InvalidArrayError};
 use mz_repr::adt::mz_acl_item::{AclItem, AclMode, MzAclItem};
 use mz_repr::adt::range::{InvalidRangeError, Range, RangeBound, parse_range_bound_flags};
 use mz_repr::adt::system::Oid;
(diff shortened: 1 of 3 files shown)


src/expr/src/scalar/func/variadic.rs
```diff
@@ -77,21 +77,49 @@ pub fn and<'a>(
...
-    // output.
-    if datums.iter().all(|d| d.unwrap_array().dims().is_empty()) {
-        let dims = &[];
-        let datums = &[];
-        let datum = temp_storage.try_make_datum(|packer| packer.try_push_array(dims, datums))?;
-        return Ok(datum);
+    // `true` if any element null or has a dimension of 0.
+    let mut have_empty = false;
+    let mut dim = None;
+    for datum in datums {
```
petrosagg, line 89: If you update the `dim` variable in all cases you can handle the wrong cardinality error in one place and also don't need to keep around `have_empty`:

```rust
let mut dim = None;
for datum in datums {
    let actual = match datum {
        Datum::Null => 0,
        Datum::Array(arr) => arr.len(),
        d => panic!("unexpected datum {d}"),
    };
    if let Some(expected) = dim && actual != expected {
        // All input arrays must have the same dimensionality.
        return Err(InvalidArrayError::WrongCardinality { actual, expected }.into());
    }
    dim = Some(actual);
}
let dim = dim.unwrap_or(0);

// Per PostgreSQL, if all input arrays are zero dimensional, so is the output.
if dim == 0 {
    return Ok(temp_storage.try_make_datum(|packer| packer.try_push_array(&[], &[]))?)
}
```

test/sqllogictest/arrays.slt
```diff
@@ -1460,3 +1460,13 @@ statement ok
 CREATE MATERIALIZED VIEW json_mv AS (
     SELECT * FROM jsons WHERE random_id = CAST(payload->>'my_field' AS uuid[])[random_index]
 )
+
+# Regression test for issue #9757
```
petrosagg:
```suggestion
# Regression test for issue MaterializeInc/database-issues#9757
```

test/sqllogictest/arrays.slt
```diff
@@ -1460,3 +1460,13 @@ statement ok
...
+
+query error db error: ERROR: number of array elements \(0\) does not match declared cardinality \(1\)
+SELECT ARRAY[NULL::BIGINT[], ARRAY[1::BIGINT, 2::BIGINT]];
+
+query T
+SELECT ARRAY[NULL::BIGINT[], ARRAY[]::BIGINT[], NULL::BIGINT[]];
+----
+{}
```
petrosagg, line 1472: I don't know if this was intentional but from reading the original code I realized that you also fixed an extra bug. Materialize on `main` accepts the following expression:

```
materialize=> select array[array[1, 2], array[3, 4, 5], array[6]];
        array
---------------------
 {{1,2},{3,4},{5,6}}
(1 row)
```

But it should be rejected. This is what pg does and I think your PR also catches now:

```
postgres=# select array[array[1, 2], array[3, 4, 5], array[6]];
ERROR:  multidimensional arrays must have array expressions with matching dimensions
```

Can we add one more regression test that doesn't use null arrays?

antiguru: Nope, unintentional, and means that `try_push_array` might be wrong. I'll check tomorrow.

antiguru: Shouldn't have come past this check: https://github.com/MaterializeInc/materialize/blob/main/src/repr/src/row.rs#L2262-L2275

mgree:
This should likely come with changes to `is_instance_of` and `is_instance_of_sql` (and possibly the `Datum` `Arbitrary` instance). (Assuming I understand this as a new behavior correctly. Worth a cross check no matter what!)

src/expr/src/scalar/func/variadic.rs
antiguru: :+1: Adjusted.

antiguru:
> This should likely come with changes to is_instance_of and is_instance_of_sql (and possibly the Datum Arbitrary instance). (Assuming I understand this as a new behavior correctly. Worth a cross check no matter what!)

I understand but I don't know what needs to be true for an array datum to be instances of ~another~ a specific array type. Do we need to consider the dimensions?

src/expr/src/scalar/func/variadic.rs
```diff
@@ -77,19 +77,44 @@ pub fn and<'a>(
...
+            }
+            if let Some((e, a)) = expected
+                .into_iter()
+                .zip_eq(actual_dims.into_iter())
+                .find(|(e, a)| e != a)
+            {
+                let actual = a.length;
+                let expected = e.length;
+                // All input arrays must have the same dimensionality.
+                return Err(InvalidArrayError::WrongCardinality { actual, expected }.into());
```
antiguru, line 110: We might need another error variant to be precise, but I feel we're good with the current one being used in two different contexts (length vs. dims).

petrosagg: Postgres also checks for the lower bound being the same, even if the lengths are the same and it also produces a more generic error message than us:

```
postgres=# SELECT ARRAY['[0:0]={1}'::int[], '[1:1]={2}'::int[]];
ERROR:  multidimensional arrays must have array expressions with matching dimensions
```

What do you think of changing our error variant to give the same generic error message and then here we just do a simple `if actual_dims != expected_dims { ... } `?

mgree:
> > This should likely come with changes to is_instance_of and is_instance_of_sql (and possibly the Datum Arbitrary instance). (Assuming I understand this as a new behavior correctly. Worth a cross check no matter what!)
> 
> I understand but I don't know what needs to be true for an array datum to be instances of ~another~ a specific array type. Do we need to consider the dimensions?

See `is_instance_of_sql`: 
https://github.com/MaterializeInc/materialize/blob/362c175cf3eb8ef50a18f44cdf1ee89c84c9ee3a/src/repr/src/scalar.rs#L1198-L1211

It seems the dimensions don't matter (unless you're dealing with int2vec weirdness).

antiguru:
Thanks for the review! I left the more specific error message in place.

PR #33805, opened 2025-10-08
Turn normalize into unary function

antiguru:
Normalize is a special function in that its second argument is a keyword instead of an expression. Move from a runtime string dispatch to determining the correct implementation at plan time. A strangeness this introduces is that there are a bunch of functions `normalize_*` that do not directly correspond to a function the user can type.


ohbadiah:
I just merged docs for this (#33785) which will need to be updated now.

ohbadiah:
> I just merged docs for this (#33785) which will need to be updated now.

I realized while sitting and staring out a window yesterday that of course the docs would not need to be updated, since the user-facing interface does not change.

ohbadiah:
> Move from a runtime string dispatch to determining the correct implementation at plan time.

@antiguru can you just say a word about the motivation? Just for clarity/posterity's sake.

PR #34105, opened 2025-11-13
Reduce dependencies of catalog-protos

antiguru:
The mz-catalog-protos crate has a large number of dependencies that aren't strictly needed. Sepcifically, it transitively depends on timely and the likes, which is not ideal for a crate that is supposed to maintain the catalog. This change aims at extracting types that have a representation in the catalog into a separate crate that does not have any odd dependency.


src/catalog-protos/src/serialization.rs
```diff
@@ -566,18 +566,6 @@ impl RustType<crate::objects::EpochMillis> for u64 {
     }
 }
 
-impl RustType<crate::objects::Timestamp> for Timestamp {
```
antiguru, line 569: We're not using this implementation. Should remove the `Timestamp` type from the proto, but that'd require a migration.

src/repr/src/adt/mz_acl_item.rs
```diff
@@ -246,10 +246,6 @@ impl RustType<ProtoAclMode> for AclMode {
     }
 }
 
-impl Columnation for AclMode {
-    type InnerRegion = CopyRegion<AclMode>;
-}
-
```
antiguru, line 252: We're not using the columnation implementations 🎉

teskje:
I don't love that this makes compute and storage crates depend on `mz-catalog-types`! Ideally, storage and compute types should be defined in storage/compute crates and used by the catalog, not the other way round.

I also think we should get to a state where `mz-catalog-protos` has no dependencies except for the ones required to define the format of the durably stored objects (mostly: `serde`). I think currently we're not there because it also contains conversion code to convert these durable types to the ones used in Materialize proper (`mz-repr` types, compute types, storage types, ...).

How about we move this conversion code out of `mz-catalog-protos`, so this crate keeps only the bare type defs? On first glance, a fitting place would be `mz-catalog`, which may also be the only place that actually needs these conversions. A wrinkle is that we won't be able to define them as trait impls (`RustType` currently, `From`/`Into` more appropriately) because of the orphan rules. Seems fine to have free-standing conversion functions though. Or, if it's somehow not fine, we can introduce a new trait for that purpose.

PR #34267, opened 2025-11-24
timely-util: specialize reclock compaction to total orders

petrosagg:
Since the reclock operator stores the reclock bindings in time order and we restrict `IntoTime` to be a total order we can perform compaction in a much cheaper way and only when necessary.

### Checklist

### DIFF
--- a/src/timely-util/src/reclock.rs
+++ b/src/timely-util/src/reclock.rs
@@ -242,7 +242,7 @@ where
         // maintenance implemented in this operator becomes problematic.
         let mut remap_upper = Antichain::from_elem(IntoTime::minimum());
         let mut remap_since = as_of.clone();
-        let mut remap_trace = Vec::new();
+        let mut remap_trace = VecDeque::new();
 
         // A stash of source updates for which we don't know the corresponding binding yet.
         let mut deferred_source_updates: Vec<ChainBatch<_, _, _>> = Vec::new();
@@ -296,7 +296,7 @@ where
             while let Some(update) = pending_remap.peek_mut() {
                 if !remap_upper.less_equal(&update.0.0) {
                     let Reverse((into, from, diff)) = PeekMut::pop(update);
-                    remap_trace.push((from, into, diff));
+                    remap_trace.push_back((from, into, diff));
                 } else {
                     break;
                 }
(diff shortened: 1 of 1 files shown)


src/timely-util/src/reclock.rs
```diff
@@ -426,21 +426,40 @@ where
...
+                    if !advanced.is_empty() {
+                        // If we have updates whose time was advanced, further peel the prefix of
+                        // updates that sit *at* the time of the since frontier.
+                        while !remap_trace.is_empty() && remap_since.contains(&remap_trace[0].1) {
+                            advanced.push(remap_trace.pop_front().unwrap());
+                        }
+                        consolidation::consolidate_updates(&mut advanced);
+                        advanced.sort_unstable_by(|(_, t1, _): &(_, IntoTime, _), (_, t2, _)| {
+                            t1.cmp(t2)
+                        });
```
antiguru, line 452: If sorting becomes expensive, you could have a custom consolidate function here.

src/timely-util/src/reclock.rs
```diff
@@ -426,21 +426,40 @@ where
...
+                    remap_since = reclocked_source_frontier;
+
+                    // The remap trace is stored in time order and T is a total order. The updates
+                    // that can have their time advanced will form a prefix, which we extract here.
+                    let mut advanced = vec![];
+                    while !remap_trace.is_empty() && !remap_since.less_equal(&remap_trace[0].1) {
+                        let (d, mut t, r) = remap_trace.pop_front().unwrap();
+                        t.advance_by(remap_since.borrow());
+                        advanced.push((d, t, r));
+                    }
```
antiguru, line 442: Could you search for `remap_since` and then move a block of updates, rather than one-by-one?

antiguru (approved):
Looks fine! I wonder if some of the while loops could be a search + bulk move instead of element-by-element loops.

src/timely-util/src/reclock.rs
petrosagg: afaik VecDeque does not offer this API, which is why we need to walk the updates and call `pop_front` one by one.

antiguru: We should measure the impact of this vs. using vectors.

PR #34325, opened 2025-12-01
DNM  Debugging: Assert various read hold stuff

ggevay:
This is just debugging the `SinceViolation` and the `dataflow has an as_of not beyond the since of collection` in https://github.com/MaterializeInc/database-issues/issues/9593, adding a lot of DNM asserts.


ggevay:
I merged similar asserts in a different PR in the meantime.

PR #34355, opened 2025-12-04
adapter: remove expensive confirm_leadership call

teskje:
This PR removes a potentially expensive `confirm_leadership` call on the sequencing thread. The call needs to potentially block for a long time, as it waits for the persist catalog to catch up and then applies all the observed changes. During that time, the environment would be unresponsive to commands, so it seems like a good idea to get rid of the call.

[Relevant Slack thread.](https://materializeinc.slack.com/archives/C08A62E0751/p1764849734635229)

### Motivation

We've been seeing slow `group_commit_initiate` messages lately. I hope this removes at least one cause of those.

### Tips for reviewer

I don't have enough context to understand why the `confirm_leadership` call was necessary before, but the comment above it makes it seem fine to delete it now. Let's hope we can trust it 😅 

### Checklist

### DIFF
--- a/src/adapter/src/coord/appends.rs
+++ b/src/adapter/src/coord/appends.rs
@@ -445,19 +445,6 @@ impl Coordinator {
             advance_to,
         } = self.get_local_write_ts().await;
 
-        // While we're flipping on the feature flags for txn-wal tables and
-        // the separated Postgres timestamp oracle, we also need to confirm
-        // leadership on writes _after_ getting the timestamp and _before_
-        // writing anything to table shards.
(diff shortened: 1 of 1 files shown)


teskje:
Well, this won't be that simple. [There is a scary zippy panic:](https://buildkite.com/materialize/test/builds/112793#019ae93c-c846-4be2-ba9b-1b8b94b90191)

```
src/storage-controller/src/persist_handles.rs:365:17: cannot register [...] at 1764850133585 because txns is at 1764850133726
```

PR #34372, opened 2025-12-05
Use superuser attribute for Cloud

SangJunBak:
### Motivation
See commit messages for details. Fixes https://github.com/issues/assigned?issue=MaterializeInc%7Cdatabase-issues%7C9948

Right now the behavior is the attribute / table will update when we next update the external_metadata (i.e. when we next refresh the token). We could write a catalog migration / builtin table migration for this too, but not sure how to extrapolate whether a user is a superuser from this. And practically, I believe this shouldn't make a difference given they'd need to authenticate to query mz_roles

Overall this feels kinda clunky. Talked more about it in this slack thread https://materializeinc.slack.com/archives/C08A62E0751/p1764979212773359

### Checklist

### DIFF
--- a/src/adapter/src/catalog/builtin_table_updates.rs
+++ b/src/adapter/src/catalog/builtin_table_updates.rs
@@ -270,12 +270,10 @@ impl CatalogState {
                     builtin_supers.contains(&role.id) || matches
                 };
 
-                let rolsuper = if self_managed_auth_enabled {
-                    Datum::from(role.attributes.superuser.unwrap_or(false))
-                } else if builtin_supers.contains(&role.id) {
+                let rolsuper = if builtin_supers.contains(&role.id) {
                     Datum::from(true)
                 } else {
-                    Datum::Null
+                    Datum::from(role.attributes.superuser.unwrap_or(false))
                 };
 
                 let role_update = BuiltinTableUpdate::row(
--- a/src/adapter/src/coord/command_handler.rs
+++ b/src/adapter/src/coord/command_handler.rs
@@ -683,7 +683,16 @@ impl Coordinator {
             // This includes preventing any user, except a pre-defined set of system users, from
             // connecting to an internal port. Therefore it's ok to always create a new role for the
             // user.
-            let attributes = RoleAttributesRaw::new();
+
+            let attributes = if user
+                .external_metadata
+                .as_ref()
+                .map_or(false, |meta| meta.admin)
+            {
+                RoleAttributesRaw::new().with_superuser()
+            } else {
+                RoleAttributesRaw::new()
+            };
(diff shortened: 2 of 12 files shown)


ggevay:
I got random-assigned to this PR as a reviewer, but unfortunately I don't really know how authentication/Frontegg works, nor have I ever looked into role stuff. I could dig into this, but maybe there is someone (besides you) who already knows this stuff better than me? If not, then I'll dive into it.

Also, maybe a question for @ohbadiah, does GitHub's random reviewer assignments from @MaterializeInc/adapter make sense nowadays? For example, for my last ~10 PRs it was clear that definitely either Jan or Aljoscha should review them, because they had all the context on the frontend peek sequencing, so when Jun got random-assigned I always just overruled it manually. Maybe this random reviewer assignment system plus the recent team shifts are somehow contributing to the phenomenon mentioned in the last R&D Sync that PR reviews gotten less prompt recently? Maybe when somebody gets random-assigned they feel less responsible for the review than when somebody explicitly, personally asks them (sometimes after a discussion of who would be the best reviewer). (Another example is https://github.com/MaterializeInc/materialize/pull/34379, which either Petros or me should review probably, rather than the randomly-assigned Aljoscha.) Edit: [Started a Slack discussion](https://materializeinc.slack.com/archives/C08A62E0751/p1765116662575119).

SangJunBak:
Makes sense to me!

SangJunBak:
Figured @jubrad has the best context and might be more opinionated on the approach here

Edit: If you don't have the spare cycles, Dov mentioned he could review it too

jubrad:
> We could write a catalog migration / builtin table migration for this too, but not sure how to extrapolate whether a user is a superuser from this. And practically, I believe this shouldn't make a difference given they'd need to authenticate to query mz_roles


Yeah, I'm not sure we can do a catalog migration, unless we've durably recorded the roles metadata somewhere. If we haven't we may still want this to be nullable -> if we don't know and a user hasn't logged in the value is null. 

I think there's a good way to do this and a bad way to do this.

**The good way:** Map the user role in FE to a role in the users IDP then automate sink via SCIM, we can still check at token refresh, or we can poll FE more frequently, or ideally stream user membership changes. The more real time you want this the more complicated it its. 
**The bad way:** We note that this is just a view of the roles based on the last login/token refresh and may be stale. Clearly then we only refresh the role on token refresh.



We may want @jasonhernandez to weigh in here as well.

SangJunBak:
Putting as draft to discuss tradeoffs and solutions in https://materializeinc.slack.com/archives/C08A62E0751/p1764979212773359

SangJunBak:
Closing since the catalog shouldn't be the source of truth for this and the console can always get this information using the frontegg APIs

PR #34544, opened 2025-12-17
Design doc for prometheus metrics from queries

alex-hunt-materialize:
Design doc for prometheus metrics from queries

### Motivation

Design doc.

### Checklist

### DIFF
--- a/doc/developer/design/20251217_prometheus_metrics_from_queries.md
+++ b/doc/developer/design/20251217_prometheus_metrics_from_queries.md
@@ -11,13 +11,21 @@ This also applies to Materialize Cloud, as we run our own external SQL exporter 
 
 - Users can define SQL queries that get turned into prometheus metrics.
 - Users can group these metrics into HTTP endpoints, so they may have separate scrape configs (for different auth requirements and/or scrape frequency).
-- Materialize Cloud can remove the promsql exporter.
 
 ## Out of Scope
 
 - Generic HTTP endpoint creation for formats other than Prometheus.
 
     While the proposed solution could easily be extended for other API types, that is not required for this to work for prometheus.
+
+- Removal of the Materialize Cloud promsql exporter.
+
+    The promsql exporter relies on internal tables which cannot have views made from them, which would complicate this proposal.
+    We should just move all those queries into normal metrics endpoints instead. That way, customers can also get access to these metrics.
+
+    There are two open tickets related to this:
+    - https://github.com/MaterializeInc/database-issues/issues/10028
+    - https://github.com/MaterializeInc/database-issues/issues/10030
 
 ## Solution Proposal
 
@@ -26,20 +34,32 @@ Allow users to create HTTP endpoints in SQL with custom prometheus metrics.
 ```sql
 CREATE API mydatabase.myschema.myprometheus FORMAT PROMETHEUS ON CLUSTER "mycluster" ON LISTENER "external";
 ```
-This will create an HTTP endpoint at `/metrics/custom/mydatabase/myschema/myprometheus` on the "external" HTTP listener (as named in the listeners configmap), and a (currently empty) table with the following schema:
+This will create an HTTP endpoint at `/metrics/custom/mydatabase/myschema/myprometheus` on the "external" HTTP listener (as named in the listeners configmap).
+
+This new api object would be added to a system table `mz_apis` for later reference.
+
+Users can then add metrics to that endpoint using SQL commands:
+```sql
+CREATE METRIC <name>
+IN API <api>
+AS (TYPE <prometheus_type>,
+    HELP <help_text>,
+    VALUES_FROM <reference_to_view>,
+    VALUE_COLUMN <name_of_value_column>);
+```
+
+This will add a new metric object to a system table `mz_metrics` for later reference:
 ```
 metric_name TEXT,
 metric_type TEXT,
 help TEXT,
-database TEXT,
-schema TEXT,
-view TEXT,
+values_from TEXT,
 value_column_name TEXT
 ```
 
 The `metric_name`, `metric_type`, and `help` fields describe the prometheus metric itself.
 
-The `database`, `schema`, and `view` fields are a reference to a view containing the metric data. The `value_column_name` is the name of a column in that view which contains the value of the metric. All other columns in the view will be used as labels.
+The `values_from` field is a reference to a view containing the metric data. The `value_column_name` is the name of a column in that view which contains the value of the metric. All other columns in the view will be used as labels.
 
 An example metric view:
 ```sql
@@ -64,14 +84,20 @@ This might look like:
 
 The user can then add this metric to their registry:
 ```sql
-INSERT INTO mydatabase.myschema.myprometheus
-VALUES      ('leads',
-             'gauge',
-             'Count of leads and whether they have been converted',
-             'mydatabase',
-             'myschema',
-             'converted_leads',
-             'count')
+CREATE METRIC leads
+IN API mydatabase.myschema.myprometheus
+AS (TYPE 'gauge',
+    HELP 'Count of leads and whether they have been converted',
+    VALUES_FROM mydatabase.myschema.converted_leads,
+    VALUE_COLUMN 'count');
+```
+
+When querying the HTTP endpoint at `/metrics/custom/mydatabase/myschema/myprometheus`, they would then get a response like:
+```
+# HELP leads Count of leads and whether they have been converted
+# TYPE leads gauge
+leads{converted="TRUE"} 22
+leads{converted="FALSE"} 67
 ```
 
 ## Minimal Viable Prototype
@@ -92,3 +118,4 @@ We currently use one of our own in Materialize Cloud, which we wrote after we hi
 
 - Exact syntax and SQL object types. We might want to have dedicated SQL syntax for creating metrics, or have some dedicated reference to the views rather than text fields, for example.
 - How to specify which listener a metrics endpoint is on. Users don't define the listeners configmap, so referencing the names we define there seems a bit weird.
+- Should we require indexed or materialized views?
--- a/doc/developer/design/20251217_prometheus_metrics_from_queries.md
+++ b/doc/developer/design/20251217_prometheus_metrics_from_queries.md
@@ -1,0 +1,94 @@
+# Prometheus Metrics From Queries
+
+## The Problem
+
+Users want to be able to monitor their Materialize workloads and data products.
+Setting up external tools to convert SQL queries into prometheus metrics is labor intensive, error prone, and often buggy.
+
+This also applies to Materialize Cloud, as we run our own external SQL exporter which could be removed if this could be hosted by environmentd.
+
+## Success Criteria
+
+- Users can define SQL queries that get turned into prometheus metrics.
+- Users can group these metrics into HTTP endpoints, so they may have separate scrape configs (for different auth requirements and/or scrape frequency).
+- Materialize Cloud can remove the promsql exporter.
+
+## Out of Scope
+
+- Generic HTTP endpoint creation for formats other than Prometheus.
+
+    While the proposed solution could easily be extended for other API types, that is not required for this to work for prometheus.
+
+## Solution Proposal
+
+Allow users to create HTTP endpoints in SQL with custom prometheus metrics.
+
+```sql
+CREATE API mydatabase.myschema.myprometheus FORMAT PROMETHEUS ON CLUSTER "mycluster" ON LISTENER "external";
+```
+This will create an HTTP endpoint at `/metrics/custom/mydatabase/myschema/myprometheus` on the "external" HTTP listener (as named in the listeners configmap), and a (currently empty) table with the following schema:
+```
+metric_name TEXT,
+metric_type TEXT,
+help TEXT,
+database TEXT,
+schema TEXT,
+view TEXT,
+value_column_name TEXT
+```
+
+The `metric_name`, `metric_type`, and `help` fields describe the prometheus metric itself.
+
+The `database`, `schema`, and `view` fields are a reference to a view containing the metric data. The `value_column_name` is the name of a column in that view which contains the value of the metric. All other columns in the view will be used as labels.
+
+An example metric view:
+```sql
+ CREATE VIEW converted_leads
+AS
+  (SELECT Count(*),
+          converted
+   FROM   (SELECT id,
+                  CASE
+                    WHEN converted_at IS NULL THEN 'FALSE'
+                    ELSE 'TRUE'
+                  END AS converted
+           FROM   leads)
+   GROUP  BY converted);
+```
+
+This might look like:
+| count | converted |
+|-------|-----------|
+|22|TRUE|
+|67|FALSE|
+
+The user can then add this metric to their registry:
+```sql
+INSERT INTO mydatabase.myschema.myprometheus
+VALUES      ('leads',
+             'gauge',
+             'Count of leads and whether they have been converted',
+             'mydatabase',
+             'myschema',
+             'converted_leads',
+             'count')
+```
+
+## Minimal Viable Prototype
+
+- [Hackathon presentation from May 2025](https://docs.google.com/presentation/d/1ek0tOlECHfpoBp_-vtcDWhN4YHpaWBfRuQyENGFbWLw/edit?slide=id.g35c518b4039_14_3503#slide=id.g35c518b4039_14_3503)
+- [Hackathon code from May 2025](https://github.com/MaterializeInc/materialize/compare/main...alex-hunt-materialize:materialize:external_api)
+- [Hackathon brainstorming from May 2025](https://www.notion.so/materialize/Hackathon-Alex-Justin-1f913f48d37b805e88b0e25a8ad1a763)
+
+While not the exact same interface, it captures the idea proposed here.
+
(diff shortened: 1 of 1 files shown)


teskje (commented):
This design does a good job of explaining the SQL-level changes, but it's missing details on the implementation, specifically about how the view results are to be collected.

In particular, I'm interested in:
* Is the view run as an IMV, or is it queried periodically?
* I assume the cluster specified in `CREATE API` is the cluster on which we run the IVM/view queries?
* How do we plan to support per-replica metrics, i.e. those collected from compute introspection sources?

I'm also not sure about the concept of inserting into a special table to cause DDL to happen. We don't have precedent for this, afaik. It introduces difficulties around error handling: If the user inserts an invalid config, how do we deal report the error, and do we silently remove the broken config again or do we leave it around?

alex-hunt-materialize:
@teskje Thank you for taking a look!

> Is the view run as an IMV, or is it queried periodically?

The individual metrics views are queried every time the HTTP endpoint receives a GET request. The actual view is user-defined, so they can use views, indexed views, materialized views, tables, whatever. If you think we should restrict this to only indexed views or materialized views, we can probably do that.

> I assume the cluster specified in CREATE API is the cluster on which we run the IVM/view queries?

This is solely for querying the metrics registry to find the views. Perhaps the `ON CLUSTER` is unnecessary?

> How do we plan to support per-replica metrics, i.e. those collected from compute introspection sources?

If a user can define a view for it, it should work here. How are these different from other things? Most likely, they would define a metric view which has labels for the cluster name or id and the replica name or id. The view would just have one row per replica.

>I'm also not sure about the concept of inserting into a special table to cause DDL to happen. We don't have precedent for this, afaik. It introduces difficulties around error handling: If the user inserts an invalid config, how do we deal report the error, and do we silently remove the broken config again or do we leave it around?

Perhaps this should be its own type with its own syntax for inserting new metric views, rather than a table. The table is just a holdover from the hackdays implementation. I'd love recommendations for how to better handle this.

teskje:
Making the view queries every time the API is scraped makes sense to me!

> If you think we should restrict this to only indexed views or materialized views, we can probably do that.

Probably not necessary, but we should document that creating indexes is recommended. Hitting persist a lot isn't great for performance, as we learned from the queries the Console is making. Also it probably risks the Prometheus scraper timing out if queries take too long.

> Perhaps the `ON CLUSTER` is unnecessary?

I think we need a way to specify on which cluster the queries should run. Indexes are per cluster after all. And we can't default to `mz_catalog_server` (or any system cluster) because users can't define their own indexes on that cluster.

> If a user can define a view for it, it should work here. How are these different from other things? Most likely, they would define a metric view which has labels for the cluster name or id and the replica name or id. The view would just have one row per replica.

Yeah, that would be nice but is not how things work today 😅 To query the [compute introspection relations](https://materialize.com/docs/sql/system-catalog/mz_introspection/), you need to target both a cluster and a replica, and the query will return results only for that replica. There is no way to collect results from all replicas with a single query. The promsql exporter has special logic to loop over all current replicas and run the introspection query on each of them.

> Perhaps this should be its own type with its own syntax for inserting new metric views, rather than a table. The table is just a holdover from the hackdays implementation. I'd love recommendations for how to better handle this.

I'm thinking something like `ALTER API <api> ADD METRIC <spec>` and `ALTER API <api> DROP METRIC <spec>`? Or make metrics a first-class object and then `CREATE/DROP METRIC <name> IN API <api> AS <spec>`.

The second thing is how we do things usually, so I think I'd prefer that. It would come with system tables `mz_apis` and `mz_metrics`, for introspection.

alex-hunt-materialize:
> Making the view queries every time the API is scraped makes sense to me!
> 
> > If you think we should restrict this to only indexed views or materialized views, we can probably do that.
> 
> Probably not necessary, but we should document that creating indexes is recommended. Hitting persist a lot isn't great for performance, as we learned from the queries the Console is making. Also it probably risks the Prometheus scraper timing out if queries take too long.

Is there a downside to restricting it to indexed things? Prometheus queries are usually run frequently on a schedule, so any memory used by these views would likely need to be allocated in frequent spikes if not indexed. Is there higher CPU used by keeping an index up to date, rather than querying every few seconds or minutes?

> > Perhaps the `ON CLUSTER` is unnecessary?
> 
> I think we need a way to specify on which cluster the queries should run. Indexes are per cluster after all. And we can't default to `mz_catalog_server` (or any system cluster) because users can't define their own indexes on that cluster.

Do we just want to extend the metric definition to include a cluster to run it on?

> > If a user can define a view for it, it should work here. How are these different from other things? Most likely, they would define a metric view which has labels for the cluster name or id and the replica name or id. The view would just have one row per replica.
> 
> Yeah, that would be nice but is not how things work today 😅 To query the [compute introspection relations](https://materialize.com/docs/sql/system-catalog/mz_introspection/), you need to target both a cluster and a replica, and the query will return results only for that replica. There is no way to collect results from all replicas with a single query. The promsql exporter has special logic to loop over all current replicas and run the introspection query on each of them.

Is this something we think will eventually be supported? We would like this internally, and I'm sure externally they would like some per-replica stuff (mostly, are all replicas healthy and "caught up"). Maybe we'll need to move replacing the promsql-exporter to a stretch goal :( .

> > Perhaps this should be its own type with its own syntax for inserting new metric views, rather than a table. The table is just a holdover from the hackdays implementation. I'd love recommendations for how to better handle this.
> 
> I'm thinking something like `ALTER API <api> ADD METRIC <spec>` and `ALTER API <api> DROP METRIC <spec>`? Or make metrics a first-class object and then `CREATE/DROP METRIC <name> IN API <api> AS <spec>`.
> 
> The second thing is how we do things usually, so I think I'd prefer that. It would come with system tables `mz_apis` and `mz_metrics`, for introspection.

`CREATE/DROP METRIC <name> IN API <api> AS <spec>` seems great to me. Would `<spec>` be a reference to a view, like we have here, or would it be the actual query?

teskje:
> Is there a downside to restricting it to indexed things? Prometheus queries are usually run frequently on a schedule, so any memory used by these views would likely need to be allocated in frequent spikes if not indexed. Is there higher CPU used by keeping an index up to date, rather than querying every few seconds or minutes?

If the query is cheap enough, running one ad-hoc dataflow every minute is probably fine. We also have persist peeks nowadays, which can make reading directly from persist very efficient, potentially even more efficient than reading from an index. But I'd say the main downside of requiring an index is the operational overhead of having to create an index for each of your Prometheus metrics, even if it shouldn't be necessary.

> Do we just want to extend the metric definition to include a cluster to run it on?

This sounds like a good idea to me, but I'd be curious about product/FE takes.

> Is this something we think will eventually be supported?

In the fullness of time, sure :D I also really want this, but it's hard to find a good design, and there hasn't been much of a desire for better compute introspection from product, afaik. It certainly won't happen soon.

> Would `<spec>` be a reference to a view, like we have here, or would it be the actual query?

Good question! Things are easier implementation-wise if it's a relation item (a view, materialized view, table, or source). I don't really like the idea of introducing a new item type that has to do SQL planning and optimization. UX wise it probably doesn't make too much of a difference?

doc/developer/design/20251217_prometheus_metrics_from_queries.md
```diff
@@ -0,0 +1,121 @@
...
+## Alternatives
+
+External SQL exporters.
+
+We currently use one of our own in Materialize Cloud, which we wrote after we hit numerous problems with third-party ones. We currently still recommend third-party solutions to our customers, which is not ideal.
+
+## Open questions
+
+- Exact syntax and SQL object types. We might want to have dedicated SQL syntax for creating metrics, or have some dedicated reference to the views rather than text fields, for example.
+- How to specify which listener a metrics endpoint is on. Users don't define the listeners configmap, so referencing the names we define there seems a bit weird.
```
jubrad: I could imagine this working like cluster sizes... we would need an internal table or non modifiable system var to stuff them into.

doc/developer/design/20251217_prometheus_metrics_from_queries.md
```diff
@@ -0,0 +1,121 @@
+# Prometheus Metrics From Queries
+
+## The Problem
+
+Users want to be able to monitor their Materialize workloads and data products.
+Setting up external tools to convert SQL queries into prometheus metrics is labor intensive, error prone, and often buggy.
+
+This also applies to Materialize Cloud, as we run our own external SQL exporter which could be removed if this could be hosted by environmentd.
```
Alphadelta14: As frustrated as i am with new-promsql-exporter, I think it is a fairly reasonable microservice that could stay where it is.
alloy's database_observability can do the same, but it would still want a pod per environment (or a sidecar container).
I can't see us having to build out so many custom metrics for our catalog by default. (or would these be transparently already-present without much visibility to the customer?)

doc/developer/design/20251217_prometheus_metrics_from_queries.md
```diff
@@ -0,0 +1,121 @@
...
+    HELP 'Count of leads and whether they have been converted',
+    VALUES_FROM mydatabase.myschema.converted_leads,
+    VALUE_COLUMN 'count');
+```
+
+When querying the HTTP endpoint at `/metrics/custom/mydatabase/myschema/myprometheus`, they would then get a response like:
+```
+# HELP leads Count of leads and whether they have been converted
+# TYPE leads gauge
+leads{converted="TRUE"} 22
```
Alphadelta14: strongly recommend a static default prefix. mz_custom_ for example. it just helps human operators know where their metrics are being sourced from and also prevents a certain kind of malicious shadowing

doc/developer/design/20251217_prometheus_metrics_from_queries.md
alex-hunt-materialize: Oops, I thought I removed mentions of our sql exporter. I had meant to remove that from the scope.

alex-hunt-materialize: Removed.

doc/developer/design/20251217_prometheus_metrics_from_queries.md
alex-hunt-materialize: Added. I've also updated the syntax and fields to match the actual implementation.

doc/developer/design/20251217_prometheus_metrics_from_queries.md
```diff
@@ -0,0 +1,136 @@
...
+|22|TRUE|
+|67|FALSE|
+
+The user can then add this metric to their registry:
+```sql
+CREATE METRIC leads
+IN API mydatabase.myschema.myprometheus
+AS (TYPE 'gauge',
+    HELP 'Count of leads and whether they have been converted',
+    VALUES FROM mydatabase.myschema.converted_leads,
```
Alphadelta14: are we bound to the term VALUES here?
it's just weird because in
```
mz_custom_leads{converted="TRUE"} 22
```
22 is a Value, converted="TRUE" is a Label

doc/developer/design/20251217_prometheus_metrics_from_queries.md
```diff
@@ -0,0 +1,136 @@
...
+- [Hackathon code from May 2025](https://github.com/MaterializeInc/materialize/compare/main...alex-hunt-materialize:materialize:external_api)
+- [Hackathon brainstorming from May 2025](https://www.notion.so/materialize/Hackathon-Alex-Justin-1f913f48d37b805e88b0e25a8ad1a763)
+
+While not the exact same interface, it captures the idea proposed here.
+
+## Alternatives
+
+External SQL exporters.
+
+We currently use one of our own in Materialize Cloud, which we wrote after we hit numerous problems with third-party ones. We currently still recommend third-party solutions to our customers, which is not ideal.
```
Alphadelta14, line 148: We do have quite a few internal sql endpoints, `/metrics/mz_compute`, `/metrics/mz_usage` etc which are recommended for self-managed right now.
there's a lot in flight going on there.
no change needed here, just finally aware of the patterns in place here.

doc/developer/design/20251217_prometheus_metrics_from_queries.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Open questions
```
Alphadelta14, line 150: how do we configure authn/authz for this endpoint?
should this be exposed in cloud? what does that look like?

alex-hunt-materialize: >how do we configure authn/authz for this endpoint?

Authn/authz is handled exactly like our other HTTP APIs. It uses whatever authenticator kind is configured for the listener(s) it is enabled on. RBAC is checked as the user who does the HTTP get.

>should this be exposed in cloud?

Initially, it will be behind a feature gate. Eventually, yes, I think we will enable it in SaaS. RBAC and authentication will be required, but we will expose it on the external listener.

doc/developer/design/20251217_prometheus_metrics_from_queries.md
alex-hunt-materialize: Do you have a recommended other term?

To me, this is still clear. We get values from that relation, even if we also get other data from it too.

doc/developer/design/20251217_prometheus_metrics_from_queries.md
```diff
@@ -0,0 +1,136 @@
...
+    - https://github.com/MaterializeInc/database-issues/issues/10030
+
+## Solution Proposal
+
+Allow users to create HTTP endpoints in SQL with custom prometheus metrics.
+
+```sql
+CREATE API mydatabase.myschema.myprometheus FORMAT PROMETHEUS IN CLUSTER "mycluster";
+```
+This will create an HTTP endpoint at `/metrics/custom/mydatabase/myschema/myprometheus` on all HTTP listeners with the `endpoint_api` enabled in the listeners configmap.
```
Alphadelta14: can we put this under the /api hierarchy? i think having it under /metrics does a slight disservice to the trust and source of the metric hierarchy.
especially if /metrics is not available to end users.

doc/developer/design/20251217_prometheus_metrics_from_queries.md
Alphadelta14: SERIES FROM, LABELS FROM, idk. i'll leave it to your best judgment. VALUES FROM isn't too bad, just felt conflated by having a value column as well as other values.

doc/developer/design/20251217_prometheus_metrics_from_queries.md
Alphadelta14: what about things like fine-grained access? I'm not sure what toggles our current rbac supplies

doc/developer/design/20251217_prometheus_metrics_from_queries.md
alex-hunt-materialize: Where exactly would you like it? `/api/metrics/custom`?

I think putting it anywhere other than under `/metrics` would be confusing, since that is where all metrics are. The source should be pretty clear, as it's got `custom` in the path, right?

>especially if /metrics is not available to end users.

I don't understand what you mean by this. It has its own toggle in the listeners configmap.

doc/developer/design/20251217_prometheus_metrics_from_queries.md
alex-hunt-materialize: Can you clarify what you mean by fine-grained access?

doc/developer/design/20251217_prometheus_metrics_from_queries.md
Alphadelta14: Yeah, `/api/metrics/custom` would probably be good. I'm referring to if you have a balancer uri (sorry I don't have a cloud instance up) like https://console.mymz.example.com/, you can't reach https://console.mymz.example.com/metrics but https://console.mymz.example.com/metrics/custom would be available.

doc/developer/design/20251217_prometheus_metrics_from_queries.md
Alphadelta14: if you build a metrics endpoint which exposes lead generation status, who gets access to said route? anyone with a view role? or by table? is it the table where the metrics came from? or just the mz_metrics table?

alex-hunt-materialize: The user making the HTTP request needs the following:
1. USAGE on the API object.
2. USAGE on the CLUSTER the API is served from.
3. SELECT on each metric's VALUES FROM relation.

I'll add a section explaining it.

doc/developer/design/20251217_prometheus_metrics_from_queries.md
alex-hunt-materialize: I'll change to SERIES FROM.

doc/developer/design/20251217_prometheus_metrics_from_queries.md
alex-hunt-materialize: done

doc/developer/design/20251217_prometheus_metrics_from_queries.md
alex-hunt-materialize: done

doc/developer/design/20251217_prometheus_metrics_from_queries.md
alex-hunt-materialize: done

Alphadelta14 (approved):

PR #34711, opened 2026-01-15
rbac: add CREATEDATAFLOW privilege for clusters

antiguru:
## Summary

- Adds a new RBAC privilege `CREATEDATAFLOW` that controls whether a user can execute queries requiring dataflow rendering (slow path) on a cluster
- Users without this privilege can still execute fast-path queries (constants, index lookups with MFP operations)
- System clusters (mz_catalog_server, etc.) and superusers bypass the check

## Motivation

In multi-tenant or cost-sensitive environments, organizations may want to restrict certain users to only execute queries that can be satisfied by existing indexes, preventing them from triggering expensive dataflow rendering operations.

## Changes

- Add `CREATE_DATAFLOW` privilege constant ('D', bit 28) to `AclMode` in `mz_acl_item.rs`
- Add `CREATEDATAFLOW` keyword and privilege parsing in sql-parser
- Add `check_create_dataflow_privilege()` function in `rbac.rs`
- Insert privilege check in `peek_finish()` for slow-path queries
- Add design doc and integration tests

## Test plan

- [x] `cargo check` passes
- [x] `cargo clippy` passes  
- [x] Unit tests pass (`cargo test -p mz-repr --lib -- mz_acl`)
- [x] Integration tests pass (`bin/sqllogictest -- test/sqllogictest/createdataflow_privilege.slt`)

🤖 Generated with [Claude Code](https://claude.ai/code)


antiguru:
I've pondered this a bunch and I don't think we get a nice implementation with our current RBAC model. The issue is that adding a `CREATEDATAFLOW` permission is surprising and intrusive, i.e, it needs explaining and mentioning. At the moment, one needs select privilege on all objects one reads from and usage privilege for the cluster. Adding a create-dataflow privilege would add this as a third required privilege. We could add it to existing privileges (a user with usage permission would be granted create-dataflow) with a migration, but then users would need to grant it explicitly. This feels we're shifting the complexity toward the user instead of building a simple-to-use system.

Looking at other database, there isn't much precedent to a permission like create-dataflow. SQL server seems to be the closest with negative permissions. In this case, create-dataflow would need to be revoked explicitly in order to prevent someone from creating dataflows. In SQL server, revokes always win over grants.

I'm not sure this is the way we should go, but at least it gives us an approach we could evaluate further.

SangJunBak:
@antiguru getting up to speed and I think the privilege makes sense conceptually, but as you said before the UX would be unpleasant. Have we considered the session variable approach described in https://github.com/MaterializeInc/database-issues/issues/5277#issuecomment-2369684851?

PR #34769, opened 2026-01-20
storage: add timely step duration metric

antiguru:
Adds `mz_timely_step_duration_seconds` histogram to the storage server, matching the existing metric in the compute server. Moves the `Metrics` class from the cluster mzcompose test to the python module directory.

### DIFF
--- a/src/storage/src/metrics.rs
+++ b/src/storage/src/metrics.rs
@@ -74,6 +74,9 @@ pub struct StorageMetrics {
     // doesn't do anything to let you pinpoint _which_ operator or worker isn't
     // yielding, but it should hopefully alert us when there is something to
     // look at.
+    //
+    // This mirrors an equivalent metric in the compute server
+    // (`mz_compute::metrics::ComputeMetrics`).
     timely_step_duration_seconds: HistogramVec,
 }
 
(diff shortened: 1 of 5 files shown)


src/storage/src/metrics.rs
```diff
@@ -61,6 +63,18 @@ pub struct StorageMetrics {
...
+    //
+    // Note that this particular metric unfortunately takes some care to
+    // interpret. It measures the duration of step_or_park calls, which
+    // undesirably includes the parking. This is probably fine because we
+    // regularly send progress information through persist sources, which likely
+    // means the parking is capped at a second or two in practice. It also
+    // doesn't do anything to let you pinpoint _which_ operator or worker isn't
+    // yielding, but it should hopefully alert us when there is something to
+    // look at.
+    timely_step_duration_seconds: HistogramVec,
```
antiguru, line 80: Add a comment to say that this mirrors an equivalent metric in the compute server. Also add a comment in the compute server to the same note.

src/storage/src/metrics.rs
```diff
@@ -74,6 +88,23 @@ impl StorageMetrics {
             sink_defs: sink::SinkMetricDefs::register_with(registry),
             source_statistics: SourceStatisticsMetricDefs::register_with(registry),
             sink_statistics: SinkStatisticsMetricDefs::register_with(registry),
+            timely_step_duration_seconds: registry.register(metric!(
```
antiguru, line 98: Add a comment saying that this must match the equivalent metric for compute (name, description, labels).

teskje (approved):
Makes sense to me!

Though this might need signoff from a cloud person. Last time I wanted to introduce new histogram buckets, that was prevented by concerns about the amount of time series ([thread](https://materializeinc.slack.com/archives/C08ACQNGSQK/p1754325709411239)).

PR #34939, opened 2026-02-08
CollectionBundles use Rc'ed collections

antiguru:
Rendering sometimes brings collections into regions, or out of regions where nothing consumes the stream. This causes us to clone the contents even though they'll be dropped immediately, wasting CPU time. This change doesn't address this problem, but mitigates its cost. Cloning an Rc is cheaper than cloning a vector.

### DIFF
--- a/src/compute/src/render.rs
+++ b/src/compute/src/render.rs
@@ -1244,13 +1244,23 @@ where
                         CollectionBundle::from_collections(oks, errs)
                     }
                     mz_compute_types::plan::GetPlan::Collection(mfp) => {
-                        let (oks, errs) = collection.as_collection_core(
-                            mfp,
-                            None,
-                            self.until.clone(),
-                            &self.config_set,
-                        );
-                        CollectionBundle::from_collections(oks, errs)
+                        if mfp.is_identity() {
+                            // Identity MFP: return the raw RcCollection directly,
+                            // avoiding Unshared → SharedRcVec round-trip.
+                            let (oks, errs) = collection
+                                .collection
+                                .clone()
+                                .expect("GetPlan::Collection requires a collection.");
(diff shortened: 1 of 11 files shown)


antiguru:
The PR avoids the problem of cloning the data and replaces it by cloning the reference:
image

Unfortunately, it adds many shared/unshared operators:

image

This one shared for putting data into a collection bundle, and one unshared for taking it out, pretty much two (data, error) for each collection imported into an LIR node. I wonder if the cost of this is too high.

src/compute/src/render/join/delta_join.rs
```diff
@@ -385,10 +385,18 @@ where
...
+                let Some(row) = row else {
+                    output.clear();
+                    return;
+                };
 
                 for (time, diff2) in output.drain(..) {
-                    let row = row.as_ref().map(|row| row.cloned()).map_err(Clone::clone);
+                    // The lint is wrong, `clone` returns `&Row` here.
+                    #[allow(clippy::useless_asref)]
+                    let row = row.as_ref().map(|&row| row.clone()).map_err(Clone::clone);
```
petrosagg, line 399: It's the `&` in the argument position that trips it. You can write it like this:

```rust
let row = row.as_ref().map(|row| (*row).clone()).map_err(Clone::clone);
```

src/compute/src/render.rs
```diff
@@ -1308,8 +1321,11 @@ where
             }
             Negate { input } => {
                 let input = expect_input(input);
-                let (oks, errs) = input.as_specific_collection(None, &self.config_set);
-                CollectionBundle::from_collections(oks.negate(), errs)
```
petrosagg, line 1312: Why were we previously going through `as_specific_collection` is it an idiom that we should keep?

petrosagg (approved):

PR #34959, opened 2026-02-09
Simplified rollout triggers and CRD design doc

alex-hunt-materialize:
Simplified rollout triggers and CRD design doc

### Motivation

Part of https://linear.app/materializeinc/issue/DEP-7/design-simplifying-upgrade-rollouts-node-rolls-converted-to-project

### Checklist

### DIFF
--- a/doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
+++ b/doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
@@ -141,7 +141,7 @@ If we can convince the API server we don't need this, we just won't implement it
 
 ### 4. Helm Chart Changes
 
-Conversion webhooks only support HTTPS, so we need a certificate for orchestratord.
+Conversion webhooks only support HTTPS, so we need a certificate for orchestratord. This may also be useful in the future, for admission webhooks for example.
 
 We plan to support two options:
 - Automatic generation using cert-manager.
--- a/doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
+++ b/doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
@@ -256,7 +256,7 @@ I think this is probably fine, since `forcePromote` only really makes sense if t
 4. **Do we need to optionally disable triggering updates on spec changes in SaaS?**
 We may also need to add a new field to the Materialize CRD spec for whether to trigger updates immediately on spec changes, or if the annotation must trigger them.
 This lets us not trigger updates immediately in SaaS when we update the region-controller to change the Materialize CR.
-We don't change the region-controller often, so maybe this isn't necessary.
+We don't change the region-controller often, and we can always update it to not apply changes to the Materialize resource until we trigger that, so this probably isn't necessary.
 
 5. **Interface changes for helm values**
 What I've got in this proposal probably works, but I'm not sure if it's exactly what we want to lock ourselves to.
@@ -267,3 +267,4 @@ It seemed like a lot to bite off at the same time as these other changes, though
 I'm not sure if it will be backwards compatible to add it later.
 I think it will, but am not certain.
 If we want generation objects, should we consider doing that first?
+This is probably fine either way. We can always do another CRD migration if it turns out to not be compatible with this one.
--- a/doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
+++ b/doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
@@ -1,0 +1,260 @@
+# Simplified upgrade rollout triggers and CRD changes
+
+- Associated: https://linear.app/materializeinc/issue/DEP-7/design-simplifying-upgrade-rollouts-node-rolls-converted-to-project
+
+## The Problem
+
+**Manual rollout triggering**
+
+Users must manually set `requestRollout` to a new UUID value to trigger a rollout, even when they've already made meaningful changes to the spec.
+
+- This does not match the behavior of anything else in Kubernetes.
+- This is confusing for users, making mistakes extremely likely and increasing frustration.
+- This makes using tools such as Terraform more difficult, as we need to also update this UUID, but not if the user has overridden it.
+- This only exists to prevent us from updating Materialize instances outside a maintenance window or when orchestratord changes.
+
+Additionally, the current system is difficult to automate when faced with eviction from nodes.
+
+## Success Criteria
+
+1. **Automatic rollout detection**: The system should automatically detect when a rollout is needed based on spec changes, without requiring users to manually set a UUID.
+
+2. **Seamless version migration**: Existing v1alpha1 resources should continue to work, with automatic conversion to v1alpha2 as needed.
+
+3. **Terraform compatibility**: Configuration must not fight with infrastructure as code tools such as Terraform.
+
+4. **SaaS compatibility**: Updating orchestratord should not trigger a rollout if the Materialize CR is unchanged.
+
+5. **Does not prevent implementation of node eviction handling**: While it is not required to have automation to handle node eviction, the implementation should not prevent adding support for node eviction later, without requiring additional major CRD changes.
+
+## Out of Scope
+
+- Rollback support.
+- Node eviction handlers.
+
+## Solution Proposal
+
+### 1. New CRD Version: v1alpha2
+
+Introduce a new `v1alpha2` version of the Materialize CRD with the following changes:
+
+**Spec changes:**
+- Remove `requestRollout` (`Uuid`) - Rollouts are now triggered automatically when the spec hash changes.
+- Remove `inPlaceRollout` (`bool`) - This has been deprecated/ignored for a while, and is replaced with `MaterializeRolloutStrategy::ImmediatelyPromoteCausingDowntime`.
+- Remove `environmentdIamRoleArn` (`Option<String>`) - This has been deprecated for a while, and is replaced with setting `"eks.amazonaws.com/role-arn"` in `serviceAccountAnnotations` instead. The conversion webhook should move this if it exists, with any conflicting value already present in `serviceAccountAnnotations` taking precedence.
+- Change `forcePromote` from `Uuid` to `Option<String>` - Instead of triggering promotion when matching the UUID of `requestRollout`, it triggers promotion when matching the hash stored in `status.requestedRolloutSpecHash`.
+
+**Status changes:**
+- Replace `lastCompletedRolloutRequest` (`Uuid`) with `lastCompletedRolloutSpecHash` (`Option<String>`) - Stores the spec hash of the last successful rollout. Will be `None` if first deploying or if upgrading from `v1alpha1`.
+- Replace `resourcesHash` (`String`) with `requestedRolloutSpecHash` (`Option<String>`) - Stores the spec hash of the currently requested rollout. Will be `None` when no rollout is ongoing.
+
+### 2. Spec Hash Generation
+
+A new `generate_spec_hash()` method computes a SHA256 hash of the spec fields that affect rollouts:
+
+```rust
+pub fn generate_spec_hash(&self) -> String {
+    let mut hasher = Sha256::new();
+    // Hash only fields that should trigger a rollout.
+    // Excludes: balancerd/console resources, forcePromote, certificates
+    // Exclusions are omitted here for brevity.
+    hasher.update(&serde_json::to_vec(&spec).unwrap());
+    // Include force_rollout annotation for manual triggers.
+    // This is future planning so we can trigger rollouts without conflicting
+    // with terraform-managed fields.
+    if let Some(annotation) = self.metadata.annotations
+        .and_then(|a| a.get(FORCE_ROLLOUT_ANNOTATION)) {
+        hasher.update(annotation);
+    }
+    format!("{:x}", hasher.finalize())
+}
+```
+
+Fields **excluded** from the hash (changes don't trigger rollout):
+- `balancerdExternalCertificateSpec`
+- `balancerdReplicas`
+- `balancerdResourceRequirements`
+- `consoleExternalCertificateSpec`
+- `consoleReplicas`
+- `consoleResourceRequirements`,
+- `forcePromote`
+
+These balancerd and console fields are excluded since they are applied on every reconciliation, without a rollout.
+
+The `forcePromote` field is excluded, since it is used to promote the existing generation, and we don't want to tear that down every time it changes.
+
+Fields **included** in the hash (changes trigger rollout):
+- `environmentdImageRef`
+- `environmentdExtraArgs`
+- `environmentdExtraEnv`
+- `environmentdConnectionRoleArn`
+- `environmentdResourceRequirements`
+- `environmentdScratchVolumeStorageRequirement`
+- `serviceAccountName`
+- `serviceAccountAnnotations`
+- `serviceAccountLabels`
+- `podAnnotations`
+- `podLabels`
+- `forceRollout`
+- `rolloutStrategy`
+- `backendSecretName`
+- `authenticatorKind`
+- `enableRbac`
+- `environmentId`
+- `systemParameterConfigmapName`
+- `internalCertificateSpec`
+- `materialize.cloud/force-rollout` annotation
+
+All other spec fields, plus our force-rollout annotation.
+
+### 3. Conversion Webhook
+
+A new HTTPS webhook server handles CRD version conversion:
+
+**Endpoint:** `POST /convert`
+
+**Supported conversions:**
+- v1alpha1 -> v1alpha2
+- v1alpha2 -> v1alpha1\*
+
+\*The API server seemed to want this, I don't know why. We can't reconcile these, so going back never makes sense.
+
+**Key conversion logic:**
+
+###### v1alpha1 to v1alpha2:
+- Spec fields:
+    - `forcePromote: Uuid` becomes `forcePromote: Option<String>` (nil UUID becomes None)
+- Status fields:
+    - `lastCompletedRolloutRequest` and `resourcesHash` are removed.
+    - If `lastCompletedRolloutRequest` and `resourcesHash` match:
+        - This means we weren't in the middle of a rollout.
+        - `lastCompletedRolloutSpecHash` should be set to the calculated spec hash (after conversion), and `requestedRolloutSpecHash` should be set to `None`. This should avoid triggering a rollout during the migration.
+    - Else:
+        - `lastCompletedRolloutSpecHash` and `requestedRolloutSpecHash` should both be set to `None`. `requestedRolloutSpecHash` will be set when reconciling. In this case, we likely have an in-progress rollout, which we will destroy and replace.
+
+###### v1alpha2 to v1alpha1:
+
+No attempt is made to support v1alpha1 beyond giving a valid v1alpha1 structure. Fields that do not exist in v1alpha2 will have their nil value.
+
+If we can convince the API server we don't need this, we just won't implement it at all.
+
+### 4. Helm Chart Changes
+
+Conversion webhooks only support HTTPS, so we need a certificate for orchestratord.
+
+We plan to support two options:
+- Automatic generation using cert-manager.
+    - Creates a self-signed `Issuer` using cert-manager
+    - Creates a `Certificate` with the service DNS name, using Ed25519 algorithm, with rotation enabled.
+    - TODO implement reloading when it changes.
+- User supplied certificates in a secret, containing fields matching the what we would get from cert-manager.
+
+Which of these is used is determined by a new helm value `operator.certificate.source`, which can be either "cert-manager" (the default) or "secret".
+
+If `operator.certificate.source` is set to "secret", the user must also set `operator.certificate.secretName` to the name of a secret in the operator's namespace.
+
+Regardless of which they choose, the resulting secret will be mounted into the orchestratord pods.
+
+Orchestratord will also get readiness probes so nothing tries to call this webhook before it is up.
+
+
+### 5. CRD Registration
+
+The CRD is registered with:
+- Both v1alpha1 and v1alpha2 versions
+- v1alpha2 as the stored version
+- Webhook conversion configuration pointing to the operator service
+
+```rust
+mz_crd.spec.conversion = Some(CustomResourceConversion {
+    strategy: "Webhook".to_owned(),
+    webhook: Some(WebhookConversion {
+        client_config: Some(WebhookClientConfig {
+            ca_bundle: Some(ByteString(ca_bytes)),
+            service: Some(ServiceReference {
+                name: webhook_service_name,
+                namespace: webhook_service_namespace,
+                path: Some("/convert".to_owned()),
+                port: Some(webhook_service_port.into()),
+            }),
+            url: None,
+        }),
+        conversion_review_versions: vec!["v1".to_owned()],
+    }),
+});
+```
+
+## Known testing required
+
+Our existing nightly orchestratord tests cover a lot, but we'll need to extend them to work with multiple CRD versions.
+
+- Upgrades from existing v1alpha1 environments by applying v1alpha1 CR. (this is basically what we have now, but we need to not break it with the orchestratord changes to reconcile v1alpha2 after conversion)
+- Upgrades from existing v1alpha1 environments by applying v1alpha2 CR.
+- Upgrades from existing v1alpha2 environments by applying v1alpha1 CR.
+- Upgrades from existing v1alpha2 environments by applying v1alpha2 CR.
+- Upgrades with a previous rollout already in progress.
+- Upgrades triggered by annotation.
+- Deploy of latest Materialize image versions using v1alpha2 CR.
+- Deploy of older Materialize image versions using v1alpha2 CR.
+
+## Minimal Viable Prototype
+
+Still a work in progress: https://github.com/MaterializeInc/materialize/pull/34904
+
+## Alternatives
+
+### 1. Keep using our current v1alpha1 CRD as-is
+
+Continue using the hash of generated Kubernetes resources rather than the Materialize spec. Maybe changing this will be more problematic than our current issues with it.
+
+**Reasons not chosen:** The resource hash includes implementation details that may change without user intent (e.g., pod labels, spec fields, additional services, etc..), causing unexpected rollouts. Hashing the spec directly is more predictable. We still have all the problems we have now.
+
+### 2. In-place CRD migration without versioning
+
+Modify v1alpha1 directly rather than creating a new version.
+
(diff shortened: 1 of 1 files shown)


doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+
+**Spec changes:**
+- Remove `requestRollout` (`Uuid`) - Rollouts are now triggered automatically when the spec hash changes.
+- Remove `inPlaceRollout` (`bool`) - This has been deprecated/ignored for a while, and is replaced with `MaterializeRolloutStrategy::ImmediatelyPromoteCausingDowntime`.
+- Remove `environmentdIamRoleArn` (`Option<String>`) - This has been deprecated for a while, and is replaced with setting `"eks.amazonaws.com/role-arn"` in `serviceAccountAnnotations` instead. The conversion webhook should move this if it exists, with any conflicting value already present in `serviceAccountAnnotations` taking precedence.
+- Change `forcePromote` from `Uuid` to `Option<String>` - Instead of triggering promotion when matching the UUID of `requestRollout`, it triggers promotion when matching the hash stored in `status.requestedRolloutSpecHash`.
+
+**Status changes:**
+- Replace `lastCompletedRolloutRequest` (`Uuid`) with `lastCompletedRolloutSpecHash` (`Option<String>`) - Stores the spec hash of the last successful rollout. Will be `None` if first deploying or if upgrading from `v1alpha1`.
```
doy-materialize: once we get ready to move to a real `v1`, we should be able to drop the `Option` here - at that point, the only time we don't have a value is when the cr is first created, but we already have the spec at that point, so we can just fill it in.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+- Replace `resourcesHash` (`String`) with `requestedRolloutSpecHash` (`Option<String>`) - Stores the spec hash of the currently requested rollout. Will be `None` when no rollout is ongoing.
```
doy-materialize: i think it'd probably be simpler to have this just always be set (not an `Option`) - the `UpToDate` condition will be an easier thing for users to check. it feels like otherwise it'd be possible to accidentally get into an invalid state (`lastCompletedrolloutSpecHash == requestedRolloutSpecHash`) which would be hard to figure out how to recover from, easier if we just don't make that kind of invalid state representable.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
doy-materialize: maybe just `lastCompletedRolloutHash`/`requestedRolloutHash`? it's not clear we're only going to be looking at only the spec (the force rollout annotation for instance, potentially other things in the future) or the entire spec, so calling it that feels a bit confusing.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+
+### 2. Spec Hash Generation
+
+A new `generate_spec_hash()` method computes a SHA256 hash of the spec fields that affect rollouts:
```
doy-materialize: `generate_rollout_hash` maybe? (as above)

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+Fields **included** in the hash (changes trigger rollout):
+- `environmentdImageRef`
+- `environmentdExtraArgs`
+- `environmentdExtraEnv`
+- `environmentdConnectionRoleArn`
+- `environmentdResourceRequirements`
+- `environmentdScratchVolumeStorageRequirement`
+- `serviceAccountName`
+- `serviceAccountAnnotations`
+- `serviceAccountLabels`
```
doy-materialize, line 100: the service account annotations and labels are also applied immediately, so they probably shouldn't be here (although a change to `serviceAccountName` will require a rollout since that needs to update the corresponding field on the statefulset)

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+- `podAnnotations`
+- `podLabels`
+- `forceRollout`
+- `rolloutStrategy`
```
doy-materialize, line 104: does including `rolloutStrategy` here make sense? i wouldn't expect a change to the rollout strategy to actually trigger a rollout

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+- `backendSecretName`
+- `authenticatorKind`
+- `enableRbac`
+- `environmentId`
```
doy-materialize, line 108: probably not in scope here, but we should really look into how to make `environmentId` immutable

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+
+A new HTTPS webhook server handles CRD version conversion:
+
+**Endpoint:** `POST /convert`
+
+**Supported conversions:**
+- v1alpha1 -> v1alpha2
+- v1alpha2 -> v1alpha1\*
+
+\*The API server seemed to want this, I don't know why. We can't reconcile these, so going back never makes sense.
```
doy-materialize, line 128: if i understand it correctly, it's because the way the migration typically works is by adding the v1alpha2 crd while keeping the storage version at v1alpha1 (to ensure you don't corrupt any existing resources if something goes wrong), ensuring that it works, then swapping the storage version to v1alpha2, then forcing a storage migration of all existing resources. in that first step, you are allowed to create v1alpha2 resources, but they will be stored as v1alpha1 (because kubernetes is only able to have a single version of a crd active at a time), and so you need to be able to convert in that direction as well. i think at the very least, we need to ensure that roundtripping the conversions results in the same field values.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+    - `lastCompletedRolloutRequest` and `resourcesHash` are removed.
+    - If `lastCompletedRolloutRequest` and `resourcesHash` match:
+        - This means we weren't in the middle of a rollout.
+        - `lastCompletedRolloutSpecHash` should be set to the calculated spec hash (after conversion), and `requestedRolloutSpecHash` should be set to `None`. This should avoid triggering a rollout during the migration.
+    - Else:
+        - `lastCompletedRolloutSpecHash` and `requestedRolloutSpecHash` should both be set to `None`. `requestedRolloutSpecHash` will be set when reconciling. In this case, we likely have an in-progress rollout, which we will destroy and replace.
+
+###### v1alpha2 to v1alpha1:
+
+No attempt is made to support v1alpha1 beyond giving a valid v1alpha1 structure. Fields that do not exist in v1alpha2 will have their nil value.
```
doy-materialize: i think this is likely fine as long as converting v1alpha2 -> v1alpha1 -> v1alpha2 doesn't lose information.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+###### v1alpha1 to v1alpha2:
+- Spec fields:
+    - `forcePromote: Uuid` becomes `forcePromote: Option<String>` (nil UUID becomes None)
+- Status fields:
+    - `lastCompletedRolloutRequest` and `resourcesHash` are removed.
+    - If `lastCompletedRolloutRequest` and `resourcesHash` match:
+        - This means we weren't in the middle of a rollout.
+        - `lastCompletedRolloutSpecHash` should be set to the calculated spec hash (after conversion), and `requestedRolloutSpecHash` should be set to `None`. This should avoid triggering a rollout during the migration.
+    - Else:
+        - `lastCompletedRolloutSpecHash` and `requestedRolloutSpecHash` should both be set to `None`. `requestedRolloutSpecHash` will be set when reconciling. In this case, we likely have an in-progress rollout, which we will destroy and replace.
```
doy-materialize: as mentioned above, i think `requestedRolloutHash` should be consistently always set to the calculated rollout hash unconditionally. (setting `lastCompletedRolloutHash` based on whether the previous things matched still makes sense though)

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+\*The API server seemed to want this, I don't know why. We can't reconcile these, so going back never makes sense.
+
+**Key conversion logic:**
+
+###### v1alpha1 to v1alpha2:
+- Spec fields:
+    - `forcePromote: Uuid` becomes `forcePromote: Option<String>` (nil UUID becomes None)
+- Status fields:
+    - `lastCompletedRolloutRequest` and `resourcesHash` are removed.
+    - If `lastCompletedRolloutRequest` and `resourcesHash` match:
```
doy-materialize: `lastCompletedRolloutRequest` should be compared to `spec.requestRollout`, not `resourcesHash`

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+
+###### v1alpha2 to v1alpha1:
+
+No attempt is made to support v1alpha1 beyond giving a valid v1alpha1 structure. Fields that do not exist in v1alpha2 will have their nil value.
+
+If we can convince the API server we don't need this, we just won't implement it at all.
+
+### 4. Helm Chart Changes
+
+Conversion webhooks only support HTTPS, so we need a certificate for orchestratord.
```
doy-materialize: the details here seem fine, but probably worth mentioning that this is a thing we should really have long term, not just for this migration - it'll leave the door open for also doing things like admission webhooks in the future, which i think might be useful for some things

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+                path: Some("/convert".to_owned()),
+                port: Some(webhook_service_port.into()),
+            }),
+            url: None,
+        }),
+        conversion_review_versions: vec!["v1".to_owned()],
+    }),
+});
+```
+
```
doy-materialize, line 241: we'll need a sixth step here - changing the stored version will cause any new resources to be saved as v1alpha2, but won't touch any existing resources. for those, we'll need to have orchestratord list all existing resources and reapply each of them. i'm not entirely sure how to determine whether a resource has already been migrated, we'll have to figure that out (presumably it's possible since https://github.com/kubernetes-sigs/kube-storage-version-migrator exists, but i don't know the details, and i don't think we want to ship an entirely separate controller like that).

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+The current implementation uses cert-manager's automatic rotation, but we haven't tested long-running deployments through certificate renewal cycles, or implemented reloading in orchestratord.
+
+3. **Is it OK to require setting forcePromote in a separate modification?**
+We can't easily know the rollout hash in advance, so users can't set `forcePromote` in the same apply as other changes.
+I think this is probably fine, since `forcePromote` only really makes sense if the rollout is stuck, which won't be the case when initially updating the spec.
+
+4. **Do we need to optionally disable triggering updates on spec changes in SaaS?**
+We may also need to add a new field to the Materialize CRD spec for whether to trigger updates immediately on spec changes, or if the annotation must trigger them.
+This lets us not trigger updates immediately in SaaS when we update the region-controller to change the Materialize CR.
+We don't change the region-controller often, so maybe this isn't necessary.
```
doy-materialize: i don't think this is necessary here - we control the environmentassignment/deployrequest operator, so we can just do whatever we need to do to make it not apply changes to the materialize resource until necessary.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,260 @@
...
+
+5. **Interface changes for helm values**
+What I've got in this proposal probably works, but I'm not sure if it's exactly what we want to lock ourselves to.
+
+6. **Should we introduce an intermediate generation object?**
+A generation object (placeholder name) similar to ReplicaSets for Deployments could simplify some of our reconciliation logic.
+It seemed like a lot to bite off at the same time as these other changes, though.
+I'm not sure if it will be backwards compatible to add it later.
+I think it will, but am not certain.
+If we want generation objects, should we consider doing that first?
```
doy-materialize, line 330: i don't think it really matters much either way. i think we do want to do this eventually, but doing one thing at a time feels like it makes more sense, we can always continue to rearrange things as necessary and do more migrations in the future. one of the main things i'm hoping for us to get from this project is experience in running these kinds of migrations so that we don't get stuck with backwards compatibility requirements forever.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: I don't think that is true. This is the last completed rollout request. A lot happens between starting a rollout and considering it complete. We save the status multiple times along the way, and we should indicate that it isn't complete in that case.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: I think they might still be load bearing, despite being applied immediately. If we change the annotations on the service account to add an AWS IAM role ARN for example, do the credentials get applied to existing pods? I'm not sure if they do.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: It might matter if they change it while they are currently mid rollout. In that case, having it included would cancel the previous rollout, then perform a new one with the new strategy.

I don't have strong feelings about this one, if you still think it shouldn't be included, I'll exclude it.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: I think it does lose information, unfortunately. I suppose we can add the new fields to the v1alpha1, then it would probably work.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: I think you're right. We can always set it to the calculated hash of the current CR, except if we have already begun promoting.

Note to self, ensure we set any status updates after we've reached promoting state using the value from the status, not the currently calculated value.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: I think we can probably make that change at any point. I don't think it needs a migration, just a schemars macro above the field.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: In this case, I'm setting the storage version to v1alpha2 immediately. All the official docs show it that way.

I do agree that most other migrations I've seen have set the storage version to the older version at first, though. In our case, I don't see any reason to keep the storage version at v1alpha1, as it would require multiple phases of the migration, and I don't think it gives us anything. Customers would be forced to update to a specific version before going further, which is something we've been trying to avoid.

I suspect the API server wanting this is because the old version is still a served version, not because of the storage version. I don't think it would be good to remove it from the served versions, though, since that might break customer's automation that needs to read the old version.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
doy-materialize: my point though is that in all of those in progress states, the last completed hash will be different from the requested rollout hash - we aren't actually gaining any extra information from the option. the way it's described here, the option is None if and only if the hashes are equal, which i think just adds an extra invalid state for no benefit?

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
doy-materialize: not against that idea, but we probably want to write down the requirements to make that happen (at the least, we'll need to save the rollout strategy in use for the current rollout in the status somewhere? otherwise we won't have any way to tell that it changed)

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
doy-materialize: yeah, that sounds easy enough

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
doy-materialize: ah, i guess that's true. may be worth testing to see what the behavior here is, but i think you're probably right. we'll probably want to leave a comment explaining this, since it's not immediately obvious

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: I agree about the requested rollout hash, just not about the last completed hash. Specifically, what would you put there during the first rollout?

doy-materialize: oh, yeah, sorry if i was unclear, i was only referring to the requested rollout hash here - the last completed rollout hash being None before the first rollout makes sense to me.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: I don't think we need to do anything special here. We can tell that _something_ changed just by including it in the hash like everything else. We don't actually need to know that it was the rollout strategy specifically.

If we aren't already in "promoting" status, regardless of what changed, we apply the new configuration without completing the old one.

If we are already in promoting status, we don't update the rollout hash and must promote. After we have successfully promoted, we would then update the rollout hash and start a new rollout.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: I think we only need the lastCompletedRolloutHash. I've added that to the doc.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: I've added a comment.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,279 @@
...
+
+**Important Note!!!**
+We **must not** update `resourcesHash` if a rollout has reached the "promoting" state. At this point we have committed to promoting the current rollout, and do not want to trigger another one until it is complete. After the existing rollout has successfully promoted, another reconciliation will be triggered at which point we will update the `resourcesHash` and trigger a new rollout.
+
+### 2. Spec Hash Generation
+
+A new `generate_rollout_hash()` method computes a SHA256 hash of the spec fields that affect rollouts:
+
+```rust
+pub fn generate_spec_hash(&self) -> String {
```
bobbyiliev: nit: The text above says `generate_rollout_hash()` but the code block defines `generate_spec_hash()`

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,279 @@
...
+    let mut hasher = Sha256::new();
+    // Hash only fields that should trigger a rollout.
+    // Excludes: balancerd/console resources, forcePromote, certificates
+    // Exclusions are omitted here for brevity.
+    hasher.update(&serde_json::to_vec(&spec).unwrap());
```
bobbyiliev: Maybe something to consider when implementing, but is the serialization order guaranteed to be stable across serde versions and struct field reorderings? If someone reorders fields in the struct definition, the hash changes silently and triggers an unexpected rollout.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,279 @@
...
+- Spec fields:
+    - `forcePromote: Uuid` becomes `forcePromote: Option<String>` (nil UUID becomes None)
+    - `requestRollout` is removed.
+- Status fields:
+    - `lastCompletedRolloutRequest` and `resourcesHash` are removed.
+    - If `lastCompletedRolloutRequest` and `spec.requestRollout` match:
+        - This means we weren't in the middle of a rollout.
+        - `lastCompletedRolloutHash` and `requestedRolloutHash` should both be set to the calculated hash (after conversion). This should avoid triggering a rollout during the migration.
+    - Else:
+        - `lastCompletedRolloutHash` should be set to `None` and `requestedRolloutHash` should be set to the calculated hash (after conversion). In this case, we likely have an in-progress rollout, which we will destroy and replace.
```
bobbyiliev, line 143: If I'm reading this right, when `lastCompletedRolloutRequest != spec.requestRollout`, the conversion sets `lastCompletedRolloutHash` to `None`, which would restart the rollout from scratch. What we might want to consider is what happens if the existing rollout is already partway through promoting when the orchestratord upgrade happens?

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,279 @@
...
+
+###### v1alpha2 to v1alpha1:
+
+We need to include the `lastCompletedRolloutHash` from v1alpha2 in v1alpha1 as well. This is required for round tripping from v1alpha2 -> v1alpha1 -> v1alpha2,
+which may happen if a user applies a v1alpha1 change over a v1alpha2 object.
```
bobbyiliev, line 150: This section might benefit from a concise, concrete example of the round-trip scenario, e.g., "user has a v1alpha2 object, applies a v1alpha1 manifest, API server converts to stored v1alpha2 via webhook for example." Would make it easier to follow.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,279 @@
...
+```
+
+### 6. Replace all Materialize resources to update their stored versions
+
+We have set v1alpha2 as the stored version, but that doesn't update existing resources. Those are only updated when they are reapplied.
+
+During orchestratord startup, after waiting for the CRD to be established, we need to loop through all Materialize resources and `replace` them.
+
+If it is possible to determine the stored version of these resources, we should only `replace` the ones at the older version.
+
```
bobbyiliev, line 249: Should we consider what the failure mode is if some `replace` calls fail during the startup migration (like maybe a transient API error, resource locked by a finalizer)?

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: serde_json, by default, keeps things sorted internally. If we enable the `preserve_order` feature flag, that would cause them to be stored in struct order.

If we want to be extra sure, we can use `serde_json::to_value` first, then call [sort_all_objects](https://docs.rs/serde_json/latest/serde_json/enum.Value.html#method.sort_all_objects), then send it to the hasher with `to_vec`.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
alex-hunt-materialize: If we're in `promoting` status, we unconditionally call the promotion for the current rollout. After updating the status after completing promotion, if anything is changed, it should trigger another rollout.

I'm pretty sure this does the right thing as-is, but this is probably something we want to write a test for.

I'll update the doc to call this out.

alex-hunt-materialize: @doy-materialize I've just pushed an update that addresses Bobby's feedback. As I was working my way through the `promoting` status scenario, I realized that maybe we need to have `status.requestedRolloutHash` be an `Option`. In cases where we're converting and the v1alpha1 is mid-rollout and already `promoting`, we don't want to set `status.requestedRolloutHash` to the current calculated hash. We have no way of knowing the hash of the spec that reached `promoting` status, so we should set it to `None`.

bobbyiliev (approved):
This looks good to me! Will leave the final check to @doy-materialize, though.

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,331 @@
...
+- Node eviction handlers.
+
+## Solution Proposal
+
+### 1. New CRD Version: v1alpha2
+
+Introduce a new `v1alpha2` version of the Materialize CRD with the following changes:
+
+**Spec changes:**
+- Remove `requestRollout` (`Uuid`) - Rollouts are now triggered automatically when the spec hash changes.
```
maheshwarip, line 42: Nice! this has always bugged me

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
maheshwarip: I'm ok with us including this!

doc/developer/design/20260209_simplified_rollout_triggers_and_crd.md
```diff
@@ -0,0 +1,331 @@
...
+- `consoleExternalCertificateSpec`
+- `consoleReplicas`
+- `consoleResourceRequirements`,
+- `forcePromote`
+
+These balancerd and console fields are excluded since they are applied on every reconciliation, without a rollout.
+
+The `forcePromote` field is excluded, since it is used to promote the existing generation, and we don't want to tear that down every time it changes.
+
+Fields **included** in the hash (changes trigger rollout):
```
maheshwarip, line 91: We know this, so just commenting for completeness sake: whenever we build this, we should add this info into the docs

maheshwarip (approved):

doy-materialize (approved):

PR #34979, opened 2026-02-10
Initial design for a serving layer

antiguru:
Thought experiment for a serving layer in Materialize.

Rendered: https://github.com/antiguru/materialize/blob/serving_layer_design/doc/developer/design/20260210_serving_layer.md

### DIFF
--- a/doc/developer/design/20260210_serving_layer.md
+++ b/doc/developer/design/20260210_serving_layer.md
@@ -1,0 +1,198 @@
+# Serving layer
+
+- Associated: (Insert list of associated epics, issues, or PRs)
+
+<!--
+The goal of a design document is to thoroughly discover problems and
+examine potential solutions before moving into the delivery phase of
+a project. In order to be ready to share, a design document must address
+the questions in each of the following sections. Any additional content
+is at the discretion of the author.
+
+Note: Feel free to add or remove sections as needed. However, most design
+docs should at least keep the suggested sections.
+-->
+
+## The Problem
+
+Selects are slow.
+In the best case, we can support around 5'000 queries per second.
+The reason is that Materialize hasn't been optimized for fast selects, but rather for efficient incremental view maintenance.
+We optimize each query, select its time stamp, send it to a replica, wait for compute to gather results, merge the results and send it back to the client.
+Several steps along this path incur latency, and this design shows alternatives that have the potential to avoid the latency.
+
+Specifically, the design outlines a system that should be able to handle 100s of 1000s requests per second, but it remains to be seen if this will turn out true.
+
+<!--
+What is the user problem we want to solve?
+
+The answer to this question should link to at least one open GitHub
+issue describing the problem.
+-->
+
+## Success Criteria
+
+Materialize can serve queries at a rate of 100k per second.
+More is better, but not required.
+
+<!--
+What does a solution to this problem need to accomplish in order to
+be successful?
+
+The criteria should help us verify that a proposed solution would solve
+our problem without naming a specific solution. Instead, focus on the
+outcomes we hope result from this work. Feel free to list both qualitative
+and quantitative measurements.
+-->
+
+## Out of Scope
+
+<!--
+What does a solution to this problem not need to address in order to be
+successful?
+
+It's important to be clear about what parts of a problem we won't be solving
+and why. This leads to crisper designs, and it aids in focusing the reviewer.
+-->
+
+## Solution Proposal
+
+To increase the queries-per-seconds, we need to change the architecture of Materialize, as incremental improvements will not allow us to reach orders-of-magnitude better performance.
+The main bottle necks are per-query work (optimization), serializing points in code (communication with clusters), and inefficient data structures for key look ups (arrangements).
+We tackle all individually, although for some we have a menu of options at our disposal.
+
+At the moment, the environmentd process optimizes queries, handles communication with the client and cluster replicas.
+Cluster replicas maintain indexes as arrangements, and handle peeks to read information from arrangements,.
+Clusters scale horizontally by adding more and bigger replicas, at the expense of causing higher tail latency.
+Environmentd currently does not scale horizontally.
+
+We could scale environmentd as a whole (which is a separate project), or extract behavior from environmentd to handle it elsewhere, which is the main idea we follow in this design.
+
+### Serving layer considerations
+
+A serving layer has to handle requests for data, at a system-determined time, and provide responses to clients at low latency and high throughput.
+It must scale horizontally and be isolated from other serving layer replicas.
+
+Choices:
+ * Fast data structure to lookup values by key.
+   We do not support scans or inequality comparisons.
+ * Reading from persist as the only way to get data into a serving layer.
+   We can allow map/filter/projects (without temporal filters).
+ * Isolation levels must be respected, i.e., we need to maintain multiple timestamped values per key, and their diff.
+ * We can offer different interfaces.
+   To handle complexity, we might not want to rely on SQL, but rather offer a REST API.
+   All interfaces need to pre-define the queries the serving layer can handle.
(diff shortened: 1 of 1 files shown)


doc/developer/design/20260210_serving_layer.md
```diff
@@ -0,0 +1,198 @@
...
+We could scale environmentd as a whole (which is a separate project), or extract behavior from environmentd to handle it elsewhere, which is the main idea we follow in this design.
+
+### Serving layer considerations
+
+A serving layer has to handle requests for data, at a system-determined time, and provide responses to clients at low latency and high throughput.
+It must scale horizontally and be isolated from other serving layer replicas.
+
+Choices:
+ * Fast data structure to lookup values by key.
+   We do not support scans or inequality comparisons.
```
djahandarie: We would be very interested in btree-like access patterns; for example, in a huge amount of our use cases, we have some sort of paginated display of index contents, where we want to SELECT ... WHERE X < ... ORDER BY X LIMIT 100 or similar. Maybe it's not so easy to incrementally maintain a data structure which allows for something like this and that's why you made this choice, but just sharing the thought.

doc/developer/design/20260210_serving_layer.md
```diff
@@ -0,0 +1,198 @@
...
+   All interfaces need to pre-define the queries the serving layer can handle.
+   We can offer an implicit API (observe frequent queries), or an explicit API (declare queries).
+
+TODO:
+ * What about a query like `SELECT :param`, which is not a keyed lookup (besides the fact it wouldn't make much sense!)
+ * Is it sufficient to say that each parameter must appear in a equality constrain with other data, but can't be free.
+
+### Proposal 1: Serving layer replicas
+
+We introduce the notion of a serving cluster with serving cluster replicas.
```
djahandarie: Is the idea that this would ~fully replace any clusters that currently function as "serving" clusters ([like recommended in the docs](https://materialize.com/docs/concepts/clusters/#three-tier-architecture-in-production))? If so this seems great to me, as it keeps the flexibility of the user being able to define their own compute isolation in whatever way is relevant to them, and now it's just more scalable/faster due to the explicit specialization.

doc/developer/design/20260210_serving_layer.md
```diff
@@ -0,0 +1,198 @@
...
+Its task is to handle specific queries, and return results at low latency.
+
+Supporting SQL prepared statements (or an equivalent)
+ * Investigate lateral joins
+ * If the placeholders are equality lookups, we can convert the query to a materialized view that arranges the data such that the placeholder columns are keys to the prepared statement.
```
djahandarie, line 178: I suppose I'm a bit confused why something special like this would need to be done. Is this to improve the ergonomics of being able to auto-optimize the performance of any arbitrary prepared statement that might come in (compared to users needing to individually define indexes for their access patterns)? I guess I'd be a little afraid of a new query suddenly causing a big rearrangement which starts getting permanently maintained, and manual definition of indexes via dbt or similar feels safer, but maybe I don't see the vision yet or don't fully understand it.

I think the strawman design would be to just extend [CREATE INDEX](https://materialize.com/docs/sql/create-index/#create-index) to have `USING newdatastructure`, and maybe also support something closer to `INCLUDE (col1, col2)` in Postgres to be able to control what is actually in the index (as opposed to the current pattern of always including everything and requiring users to make views in order to achieve this) to improve ergonomics. Then a prepared statement would simply be responsible for finding a matching index, and from there it could be served very fast, right?

doc/developer/design/20260210_serving_layer.md
```diff
@@ -0,0 +1,198 @@
...
+team members, our customers, our prospects, academic research, prior art, or
+competitive research. One of our company values is to "do the reading" and
+to "write things down." This is your opportunity to demonstrate both!
+-->
+
+## Open questions
+
+* What's the right data structure?
+* How do clients talk to the serving replica?
+  Is environmentd a bump on the wire, is there a separate service, or do we allow direct connections?
```
djahandarie, line 322: (One vote for as close to a direct connection as possible 😄 We really want to isolate serving between tenants as much as possible, and don't want any cross-tenant DoS situations even if one makes a couple million connections or something crazy. )

djahandarie (commented):
Just saw this PR scroll by and wanted to share that we are very excited about this direction! I know it's a draft, but left some random thoughts in case they are useful. (Feel free to ignore them, you don't need to give detailed responses unless it's helpful to your thought process somehow.)

aljoscha:
In that design, is `envd` still "terminating" network connections or clients go directly to a serving-layer process? I can see either way working, but if clients go directly to a serving-layer process we need a way of syncing catalog changes over to there, and they need to run a check to make sure they're up to date to be strictly serializable.

You'd need roughly https://github.com/MaterializeInc/database-issues/issues/9738, which builds on two other issues.

antiguru:
I updated the design to incorporate early feedback. It now offers three alternatives (good sink, built in serving layer, external serving layer), and an analysis of what other systems do.

doc/developer/design/20260210_serving_layer.md
```diff
@@ -0,0 +1,316 @@
...
+The main bottlenecks are per-query work (optimization), serializing points in code (communication with clusters), and inefficient data structures for key lookups (arrangements).
+Rather than making the current query path more efficient---which is unlikely to yield orders-of-magnitude improvement---this design proposes to bypass it entirely with a dedicated serving layer.
+The following sections present three broad options, each with different trade-offs around scope, API commitment, and long-term flexibility.
+
+At the moment, the environmentd process optimizes queries, handles communication with the client and cluster replicas.
+Cluster replicas maintain indexes as arrangements, and handle peeks to read information from arrangements.
+Clusters scale horizontally by adding more and bigger replicas, at the expense of causing higher tail latency.
+Environmentd currently does not scale horizontally, but we could scale it as a whole (which is a separate project), or extract behavior from environmentd to handle it elsewhere, which is the main idea we follow in this design.
+
+### Option 1: Build an excellent sink
```
djahandarie, line 39: One crazy idea that's sort of in-between option 1 and option 2: how about SQLite?

Materialize would still be responsible for setting up the SQLite instances so there would be no operational burden on the part of the user, but it would completely delegate all concerns regarding serving to the SQLite code, and provide all the benefits it has (including potentially more complex queries like JOINs, multi-column BTree indexes, multiple million QPS). It seems like it might be a very fast way to get something workable here which could then be optimized or changed in the future if it has any particular downsides.

The natural concern would be if non-concurrent writes can be made fast enough, but I believe they can with enough tuning. (Also FWIW, in our use case we don't expect super high write rates anyways, and are totally fine with non-strict serializable semantics for the cases we need high QPS.)

Also it doesn't have a wire protocol, but I think that's potentially actually desirable here as Materialize could continue owning the pgwire adapter logic etc. (Or https://github.com/erans/pgsqlite if that is not desirable...)

Maybe https://www.rivet.dev/blog/2025-02-16-sqlite-on-the-server-is-misunderstood/ is an interesting blog post to consider in this context.

djahandarie (commented):
I really like the updates, thank you!

One more (potential) "requirement" to share in case it's useful:
* In some cases, we may have two very big arrangements (say 1TB each), and we want to do a query like `SELECT * FROM a JOIN b.a_id = a.id WHERE a.id = 'X'`. It doesn't make sense to combine both arrangements in advance as they would explode in size to 10TB~100TB. This sort of query currently plans to a simple plan that does `lookup`s on two indexes.
* If we imagine some very restricted engine that doesn't support things like the above and say, only supports K-V lookups, then the ergonomics would be potentially reduced. So I think it's worth keeping in mind this type of "more complicated than a K-V lookup but still not too complicated" type of query. But it becomes quite complex where to draw the line I think (for example, can we also do pagination on top of such a query...?). This is what led me to thinking about SQLite actually.


(As usual just food for thought.)

PR #35063, opened 2026-02-18
crd: add extra volumes and volume mounts to environmentd

antiguru:
Add a dedicated `fdbClusterConfigMapName` field to the Materialize CRD spec. When set, orchestratord mounts the referenced ConfigMap (expecting a `cluster-file` key) and sets the `FDB_CLUSTER_FILE` environment variable automatically, following the same pattern as `systemParameterConfigmapName` and the license key.

### DIFF
--- a/doc/user/data/self_managed/materialize_crd_descriptions.json
+++ b/doc/user/data/self_managed/materialize_crd_descriptions.json
@@ -115,6 +115,14 @@
         "deprecated": false
       },
       {
+        "name": "fdbClusterConfigMapName",
+        "type": "String",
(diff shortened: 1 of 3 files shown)


alex-hunt-materialize:
@antiguru Do you have an example of the contents of this config? Is this not stuff that could go in the backend metadata secret?

I don't think we want to add additional stuff to the CRD if we can avoid it.

Maybe we should have a quick zoom or slack discussion about this? The cloud team generally doesn't understand what this is, and we don't want to evaluate anything like this until we understand it better.

jubrad:
I'm really hesitant to couple our CRD with FDB this tightly. It's possible that we could take a different approach where we allow for more generic mounts of the FDB configmap... If this is only needed for the environment pod this could work, but would envd need to pass the config to all running clusters?


https://github.com/MaterializeInc/materialize/compare/main...jubrad:materialize:fdb_crd

antiguru:
I'm a bit out of my depth here. We need some way to pass the `fdb.cluster` file into both environmentd and clusterd, but I'm less clear about the right mechanism for doing so. I agree that our controller should be as consensus-agnostic as possible.

PR #35087, opened 2026-02-19
emulator: add MZ_METADATA_STORE switch for postgres/foundationdb

antiguru:
Add FoundationDB as an alternative metadata store for the emulator image. Both PostgreSQL and FoundationDB are installed in the base image. The entrypoint reads `MZ_METADATA_STORE` (default: `postgres`) to select which backend to start. Set `MZ_METADATA_STORE=foundationdb` to use FDB.


def- (commented):
How much larger does it make our Docker images to have both Postgres and FoundationDB?
Edit: Just checked: 218 MB -> 322 MB, that's pretty significant for an experiment. I'd propose to add a `materialized-fdb` image instead for experimentation.

PR #35191, opened 2026-02-24
rust: Use SQL formatting

def-:
For context see https://materializeinc.slack.com/archives/C08A62E0751/p1771412204274059

> For our internal SQL we should probably switch to a proper SQL query builder that automatically does the escaping

The interface is inspired by https://www.psycopg.org/psycopg3/docs/api/sql.html, but uses macros for compile-time guarantees. The idea is to enforce building SQL strings properly, without injection. Using the Rust type system and clippy to prevent just passing in strings.


def-:
Ready for review

src/storage/src/source/postgres/replication.rs
```diff
@@ -698,13 +699,13 @@ async fn raw_stream<'a>(
...
     let lsn = PgLsn::from(resume_lsn.offset);
-    let query = format!(
-        r#"START_REPLICATION SLOT "{}" LOGICAL {} ("proto_version" '1', "publication_names" {})"#,
-        Ident::new_unchecked(slot).to_ast_string_simple(),
-        lsn,
-        escaped_string_literal(publication),
+    let query = sql!(
+        "START_REPLICATION SLOT {} LOGICAL {} (\"proto_version\" '1', \"publication_names\" {})",
+        Sql::ident(slot),
+        Sql::raw_unchecked(lsn.to_string()),
```
ggevay, line 705: Could we introduce a `impl From<PgLsn> for Sql` to have one less `raw_unchecked` call?

src/ore/src/sql.rs
```diff
@@ -0,0 +1,396 @@
...
+                out.push('\'');
+            }
+            out.push(ch);
+        }
+        out.push('\'');
+        Self(Cow::Owned(out))
+    }
+
+    /// Creates a SQL fragment for a PostgreSQL positional parameter (e.g. `$1`).
+    pub fn param(index: usize) -> Self {
```
ggevay, line 156: (We could add `assert!(index >= 1 && index <= 65535)`.)

src/environmentd/src/http/mcp.rs
```diff
@@ -797,7 +797,7 @@ async fn execute_sql(
     execute_request(
         client,
         SqlRequest::Simple {
-            query: query.to_string(),
+            query: mz_ore::sql::Sql::raw_unchecked(query.to_string()),
```
ggevay: Wondering if this should be `trusted_external_request` instead of `raw_unchecked`.

src/sqllogictest/src/runner.rs
```diff
@@ -848,9 +849,10 @@ impl<'a> Runner<'a> {
             .await?
         {
             let name: &str = row.get("name");
+            let name = Ident::new_unchecked(name).to_ast_string_simple();
             inner
                 .system_client
-                .batch_execute(&format!("DROP DATABASE \"{name}\""))
+                .batch_execute(&format!("DROP DATABASE {name}"))
```
ggevay: Why not use the new API instead of `to_ast_string_simple`?

(and a few more places in the same file)

src/ore/src/sql.rs
```diff
@@ -0,0 +1,396 @@
...
+        // external `/api/sql` endpoint. Any other path that produces a `Sql`
+        // value should use the safe constructors (`Sql::new`, `Sql::ident`,
+        // `Sql::literal`, the `sql!` macro) or `Sql::raw_unchecked`.
+        Ok(Sql::trusted_external_request(String::deserialize(
+            deserializer,
+        )?))
+    }
+}
+
+impl Add for Sql {
```
ggevay: This seems to be currently unused. Is that intended?

ggevay (commented):
Releasing a first batch of comments.

src/storage/src/source/postgres/replication.rs
def-: Orphan rules force this impl into the crate that defines Sql (mz_ore), but PgLsn lives in tokio-postgres. So mz_ore would have to take a Postgres-driver dependency.

ggevay: Ok, then it's not worth it.

src/mz/src/command/secret.rs
```diff
@@ -72,30 +69,29 @@ pub async fn create(
...
-    // The most common ways to write a secret are the following ways:
+    // The most common ways to write a secret are:
     // 1. Decode function: decode('c2VjcmV0Cg==', 'base64')
     // 2. ASCII: 13de2601-24b4-4d8f-9931-375c0b2b5cd4
     // For case 1) we pass through the expression as-is (it's a SQL expression, not a literal).
     // For case 2) we escape the value as a string literal.
-    let buffer = if buffer.starts_with("decode") {
-        buffer
+    let value = if buffer.starts_with("decode") {
+        Sql::raw_unchecked(buffer)
```
ggevay: The `buffer.starts_with("decode")` seems a bit fragile in itself. For example, I could pass
```
decode('AAAA','base64'); CREATE ROLE pwned WITH SUPERUSER; --
```

How about trying to parse it as an expression with something like
```
let value = if buffer.starts_with("decode") {
    // Validate that `buffer` parses as a single SQL expression.
    let expr = mz_sql_parser::parser::parse_expr(&buffer)
        .map_err(|e| Error::InvalidSecretExpression(e.to_string()))?;
    Sql::raw_unchecked(expr.to_ast_string_stable())
} else {
    Sql::literal(&buffer)
};
```

src/ore/src/sql.rs
```diff
@@ -0,0 +1,385 @@
...
+        }
+
+        if args.next().is_some() {
+            return Err(SqlFormatError::ExtraArgument);
+        }
+        Ok(Sql(Cow::Owned(out)))
+    }
+
+    #[doc(hidden)]
+    pub fn format_unchecked(self, args: impl IntoIterator<Item = Sql>) -> Self {
```
ggevay, line 235: Worth a short `// Safety:` block above this fn listing the invariants the `sql!` macro must uphold — `{`/`}` paired only as `{{`/`}}`/`{}`, and arg count matches placeholder count. Right now this is `#[doc(hidden)] pub` and the safety story rests entirely on the macro being the only caller; an explicit comment makes it greppable as the macro's trusted partner and protects against a future accidental direct call.

src/environmentd/src/http/prometheus.rs
```diff
@@ -26,21 +27,21 @@ impl<'a> PrometheusSqlQuery<'a> {
...
-                    self.query
-                )
-            } else {
-                format!(
-                    "SET auto_route_catalog_queries = true; RESET CLUSTER; RESET CLUSTER_REPLICA; {}",
-                    self.query
-                )
-            },
-        }
+        let query = Sql::raw_unchecked(self.query.to_string());
```
ggevay: Every construction of `PrometheusSqlQuery` in this file uses a `&'static str` literal for `query` (the `'a` looks like a copy/paste from the other fields). If you tighten that field to `&'static str`, this line can become `Sql::new(self.query)` — one fewer `raw_unchecked` callsite, and the trust source becomes compiler-enforced rather than convention.

src/fivetran-destination/src/destination/ddl.rs
```diff
@@ -108,10 +111,8 @@ pub async fn describe_table(
 
 pub async fn handle_create_table(request: CreateTableRequest) -> Result<(), OpError> {
     let table = request.table.ok_or(OpErrorKind::FieldMissing("table"))?;
-
-    let schema = Ident::new(&request.schema_name)?;
-    let qualified_table_name =
-        UnresolvedItemName::qualified(&[schema.clone(), Ident::new(&table.name)?]);
+    let schema_name = Sql::ident(&request.schema_name);
```
ggevay: Pre-PR these went through `Ident::new(...)?`, which validated non-emptiness / length / NUL bytes etc.; `Sql::ident` only quotes-and-escapes. The escape behavior is fine (SQL injection is closed), but malformed Fivetran identifiers now error at the Materialize parser instead of at this handler. Worth a one-line comment here noting that identifier validation is now delegated to the SQL parser, to defuse the inevitable future "why no `Ident::new` here?" question. (Same applies in `dml.rs`.)

clippy.toml
```diff
@@ -53,6 +53,27 @@ disallowed-methods = [
...
+    { path = "tokio_postgres::Transaction::batch_execute", reason = "use `Sql` wrappers from `mz_postgres_util` instead" },
+    { path = "tokio_postgres::Transaction::query", reason = "use `mz_postgres_util::query`/`query_prepared` wrappers instead" },
+    { path = "tokio_postgres::Transaction::query_one", reason = "use `mz_postgres_util::query_one`/`query_one_prepared` wrappers instead" },
+    { path = "tokio_postgres::Transaction::query_opt", reason = "use `mz_postgres_util::query_opt`/`query_opt_prepared` wrappers instead" },
+    { path = "tokio_postgres::Transaction::execute", reason = "use `mz_postgres_util::execute`/`execute_prepared` wrappers instead" },
+    { path = "postgres::Client::simple_query", reason = "avoid direct string SQL; use `Sql` composition and wrappers where possible" },
+    { path = "postgres::Client::batch_execute", reason = "avoid direct string SQL; use `Sql` composition and wrappers where possible" },
+
+    { path = "mysql_async::queryable::Queryable::query", reason = "use `Queryable::exec` (prepared statements) instead, or validate inputs and `#[allow]` locally" },
```
ggevay, line 73: Worth a one-sentence comment above these MySQL entries explaining the asymmetry with the PG block: MySQL gets `query → exec` lints (pushing toward prepared statements) rather than an `Sql`-style typed builder, because `mysql_async`'s prepared-statement API already provides the trust boundary. Otherwise a future reader will wonder why MySQL didn't get the same treatment as PG.

ggevay (approved):
LGTM, thanks for iterating on this!

I posted some (low confidence) minor AI comments.

PR #35220, opened 2026-02-25
Adapter: unify RBAC for Side Effecting Functions

DAlperin:
During an incident I tried to kill some backends with mz_system and I couldn't. This was because the usually RBAC bypass doesn't work for the special impl in the coord for side effect functions. I'd prefer to unify this where possible and feel it's worth paying an extra coord hop is worth it.

Fixes: https://github.com/MaterializeInc/database-issues/issues/11333

### DIFF
--- a/src/adapter/src/command.rs
+++ b/src/adapter/src/command.rs
@@ -332,8 +332,15 @@ pub enum Command {
         plan: SideEffectingFunc,
         conn_id: ConnectionId,
         /// The current role of the session, used for RBAC checks.
-        current_role: RoleId,
         tx: oneshot::Sender<Result<ExecuteResponse, AdapterError>>,
+    },
+
+    /// Look up the authenticated role for a connection by its raw connection ID.
+    /// Returns `Some(role_id)` if the connection exists, `None` otherwise.
+    /// Used by frontend peek sequencing to provide `active_conns` data to `rbac::check_plan`.
+    GetConnectionAuthenticatedRole {
+        connection_id: u32,
+        tx: oneshot::Sender<Option<RoleId>>,
     },
 
     /// Register a pending peek initiated by frontend sequencing. This is needed for:
@@ -409,6 +416,7 @@ impl Command {
             | Command::CopyToPreflight { .. }
             | Command::ExecuteCopyTo { .. }
             | Command::ExecuteSideEffectingFunc { .. }
+            | Command::GetConnectionAuthenticatedRole { .. }
             | Command::RegisterFrontendPeek { .. }
             | Command::UnregisterFrontendPeek { .. }
             | Command::ExplainTimestamp { .. }
@@ -449,6 +457,7 @@ impl Command {
             | Command::CopyToPreflight { .. }
             | Command::ExecuteCopyTo { .. }
             | Command::ExecuteSideEffectingFunc { .. }
+            | Command::GetConnectionAuthenticatedRole { .. }
             | Command::RegisterFrontendPeek { .. }
             | Command::UnregisterFrontendPeek { .. }
             | Command::ExplainTimestamp { .. }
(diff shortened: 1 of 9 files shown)


src/adapter/src/frontend_peek.rs
```diff
@@ -398,14 +398,32 @@ impl PeekClient {
...
+                            connection_id: *connection_id,
+                            tx,
+                        })
+                        .await
+                    }
+                };
+
+                rbac::check_plan(
+                    &conn_catalog,
+                    Some(move |_id: u32| active_conns_role),
```
ggevay: Making the closure not use `_id` (just always returning PgCancelBackend's target connection's role) runs the risk that some code written later will call the closure with some other connection, and silently get a wrong role. How about asserting in the closure that the closure's argument is the same as the target conn, i.e., `_id == connection_id`?

src/adapter/src/command.rs
```diff
@@ -282,10 +282,17 @@ pub enum Command {
         plan: SideEffectingFunc,
         conn_id: ConnectionId,
         /// The current role of the session, used for RBAC checks.
-        current_role: RoleId,
```
ggevay, line 335: Dangling doc comment

src/adapter/src/frontend_peek.rs
ggevay: Or we could maybe even simplify this from a closure to a simple `Option<RoleId>`, because the closure is never called with a variety of connections, but only ever called on PgCancelBackend's target connection. Also, the closure's generality is not really needed even for further side-effecting funcs in Postgres that we might want to implement at some point: The only other such functions where roles of some other connections than the issuer are relevant are `pg_terminate_backend` and `pg_log_backend_memory_contexts`, which also just want to look at the role of a single connection, i.e., their target connection. So, an `Option<RoleId>` would be enough for them as well. What do you think?

ggevay (approved):
Looks good! Thank you very much for figuring this out and fixing it! And sorry for the bug!

src/adapter/src/coord/sequencer/inner.rs
```diff
@@ -2245,36 +2243,11 @@ impl Coordinator {
                     return Err(AdapterError::Canceled);
                 }
 
-                // Perform RBAC check: the current user must be a member of the role
-                // that owns the connection being cancelled.
-                if let Some((_id_handle, conn_meta)) =
+                // check_plan already verified role membership.
```
def-, line 2246: What if role membership changed since we verified it? This isn't atomic anymore with this PR

def- (commented):
Are we still planning to merge this?

src/adapter/src/frontend_peek.rs
ggevay: Done

src/adapter/src/command.rs
ggevay: Fixed

src/adapter/src/coord/sequencer/inner.rs
ggevay: Fixed. There were two possible issues:
1. Connection goes away, and conn id is reused for a new connection, so we end up cancelling stuff on a new connection.
2. Permission is revoked between the check and the cancellation.

Fixed 1. in the "adapter: pin cancel target's connection ID across the RBAC check" commit. 

And I think 2. is not an actual issue, so I left that. The check reflects the state at the time the statement is sequenced, matching how privilege changes generally don't affect in-flight statements (Postgres behaves comparably). This is now documented in the code.

ggevay:
I've addressed the comments (mine and @def-'s), so I think this is ready for merging, but maybe it would be good to have one more @MaterializeInc/adapter review, because I made non-trivial changes.

aljoscha (approved):
looks good! @ggevay

PR #35235, opened 2026-02-26
Add IN CLUSTER ... REPLICA ... syntax for materialized views

antiguru:
Adds `IN CLUSTER <cluster> REPLICA <replica>` syntax to `CREATE MATERIALIZED VIEW`, allowing materialized views to be targeted at a specific replica.
The feature is gated behind `enable_replica_targeted_materialized_views` (default false).

Replica-targeted MVs are only installed on the specified replica, reusing the existing `target_replica` mechanism in the compute controller.
The `subscribe_target_replica` parameter is merged into `target_replica` so subscribes and introspection subscribes also benefit from single-replica placement instead of broadcasting to all replicas and filtering responses.
Dropping the target replica implicitly drops all MVs targeting it.

Tests:
* `test/sqllogictest/materialized_view_replica_targeted.slt` — parsing, error cases, round-trip, data correctness
* `test/testdrive/materialized-view-replica-targeted.td` — introspection (hydration status, per-replica frontiers), replica drop/recreate, cluster restart
* Platform check — validates survival across restarts/upgrades
* Parallel workload — randomly creates replica-targeted MVs under concurrency
* Race-condition test — concurrent DDL stress testing

### DIFF
--- a/src/adapter-types/src/dyncfgs.rs
+++ b/src/adapter-types/src/dyncfgs.rs
@@ -154,6 +154,13 @@ pub const ENABLE_S3_TABLES_REGION_CHECK: Config<bool> = Config::new(
     "Whether to enforce that S3 Tables connections are in the same region as the environment.",
 );
 
+/// Whether to enable creation of replica-targeted materialized views.
+pub const ENABLE_REPLICA_TARGETED_MATERIALIZED_VIEWS: Config<bool> = Config::new(
+    "enable_replica_targeted_materialized_views",
+    false,
+    "Enable creation of replica-targeted materialized views.",
+);
+
 /// Adds the full set of all adapter `Config`s.
 pub fn all_dyncfgs(configs: ConfigSet) -> ConfigSet {
     configs
@@ -177,4 +184,5 @@ pub fn all_dyncfgs(configs: ConfigSet) -> ConfigSet {
         .add(&OIDC_AUTHENTICATION_CLAIM)
         .add(&PERSIST_FAST_PATH_ORDER)
         .add(&ENABLE_S3_TABLES_REGION_CHECK)
+        .add(&ENABLE_REPLICA_TARGETED_MATERIALIZED_VIEWS)
 }
--- a/src/adapter-types/src/dyncfgs.rs
+++ b/src/adapter-types/src/dyncfgs.rs
@@ -154,13 +154,6 @@ pub const ENABLE_S3_TABLES_REGION_CHECK: Config<bool> = Config::new(
     "Whether to enforce that S3 Tables connections are in the same region as the environment.",
 );
 
-/// Whether to enable creation of replica-targeted materialized views.
-pub const ENABLE_REPLICA_TARGETED_MATERIALIZED_VIEWS: Config<bool> = Config::new(
-    "enable_replica_targeted_materialized_views",
-    false,
-    "Enable creation of replica-targeted materialized views.",
-);
-
 /// Adds the full set of all adapter `Config`s.
 pub fn all_dyncfgs(configs: ConfigSet) -> ConfigSet {
     configs
@@ -184,5 +177,4 @@ pub fn all_dyncfgs(configs: ConfigSet) -> ConfigSet {
         .add(&OIDC_AUTHENTICATION_CLAIM)
         .add(&PERSIST_FAST_PATH_ORDER)
         .add(&ENABLE_S3_TABLES_REGION_CHECK)
-        .add(&ENABLE_REPLICA_TARGETED_MATERIALIZED_VIEWS)
 }
--- a/src/adapter/src/catalog/transact.rs
+++ b/src/adapter/src/catalog/transact.rs
@@ -2892,6 +2892,23 @@ impl ObjectsToDrop {
 
                 self.replicas
                     .insert(replica.replica_id, (cluster.id, reason));
+
+                // Implicitly drop materialized views that target this replica.
+                // When the target replica is gone, no replica advances the
+                // persist shard's upper frontier, causing reads to hang.
+                for (_id, entry) in state.get_entries() {
+                    if let CatalogItem::MaterializedView(mv) = entry.item() {
+                        if mv.target_replica == Some(replica_id)
+                            && !self.items.contains(&entry.id())
+                        {
+                            tracing::warn!(
+                                "implicitly dropping materialized view {} because target replica was dropped",
+                                entry.name().item,
+                            );
+                            self.items.push(entry.id());
+                        }
+                    }
+                }
             }
             DropObjectInfo::Item(item_id) => {
                 let entry = state.get_entry(&item_id);
--- a/src/adapter/src/coord/sequencer/inner/create_materialized_view.rs
+++ b/src/adapter/src/coord/sequencer/inner/create_materialized_view.rs
@@ -335,6 +335,17 @@ impl Coordinator {
             ..
         } = &plan;
 
+        // Replica-targeted materialized views require a feature flag.
+        if plan.materialized_view.target_replica.is_some()
+            && !mz_adapter_types::dyncfgs::ENABLE_REPLICA_TARGETED_MATERIALIZED_VIEWS
+                .get(self.catalog().system_config().dyncfgs())
+        {
+            return Err(AdapterError::UnavailableFeature {
+                feature: "replica-targeted materialized views".to_string(),
+                docs: None,
+            });
+        }
+
         // Validate any references in the materialized view's expression. We do
         // this on the unoptimized plan to better reflect what the user typed.
         // We want to reject queries that depend on log sources, for example,
@@ -567,6 +578,7 @@ impl Coordinator {
                             dependencies,
                             replacement_target,
                             cluster_id,
+                            target_replica,
                             non_null_assertions,
                             compaction_window,
                             refresh_schedule,
@@ -674,6 +686,7 @@ impl Coordinator {
                     dependencies,
                     replacement_target,
                     cluster_id,
+                    target_replica,
                     non_null_assertions,
                     custom_logical_compaction_window: compaction_window,
                     refresh_schedule: refresh_schedule.clone(),
@@ -755,6 +768,7 @@ impl Coordinator {
                             df_desc,
                             cluster_id,
                             notice_builtin_updates_fut,
+                            target_replica,
                         )
                         .await;
 
--- a/src/adapter/src/coord/sequencer/inner/create_materialized_view.rs
+++ b/src/adapter/src/coord/sequencer/inner/create_materialized_view.rs
@@ -335,17 +335,6 @@ impl Coordinator {
             ..
         } = &plan;
 
-        // Replica-targeted materialized views require a feature flag.
-        if plan.materialized_view.target_replica.is_some()
-            && !mz_adapter_types::dyncfgs::ENABLE_REPLICA_TARGETED_MATERIALIZED_VIEWS
-                .get(self.catalog().system_config().dyncfgs())
-        {
-            return Err(AdapterError::UnavailableFeature {
-                feature: "replica-targeted materialized views".to_string(),
-                docs: None,
-            });
-        }
-
         // Validate any references in the materialized view's expression. We do
         // this on the unoptimized plan to better reflect what the user typed.
         // We want to reject queries that depend on log sources, for example,
--- a/src/compute-client/src/controller.rs
+++ b/src/compute-client/src/controller.rs
@@ -764,6 +764,13 @@ where
 
         let instance = self.instance(instance_id)?;
 
+        // Validation: target replica
+        if let Some(replica_id) = target_replica {
+            if !instance.replicas.contains(&replica_id) {
+                return Err(ReplicaMissing(replica_id));
+            }
+        }
+
         // Validation: as_of
         let as_of = dataflow.as_of.as_ref().ok_or(MissingAsOf)?;
         if as_of.is_empty() && dataflow.subscribe_ids().next().is_some() {
@@ -777,6 +784,14 @@ where
         let storage_ids = dataflow.imported_source_ids().collect();
         let mut import_read_holds = self.storage_collections.acquire_read_holds(storage_ids)?;
         for id in dataflow.imported_index_ids() {
+            if let Some(target_replica) = target_replica
+                && instance
+                    .collection(id)?
+                    .target_replica
+                    .is_some_and(|r| r != target_replica)
+            {
+                return Err(ReplicaNotHostingIndex(target_replica));
+            }
             let read_hold = instance.acquire_read_hold(id)?;
             import_read_holds.push(read_hold);
         }
@@ -874,10 +889,9 @@ where
                 return Err(ReplicaMissing(replica_id));
             }
             let collection = instance.collection(peek_target.id())?;
-            if let Some(collection_replica) = collection.target_replica {
-                if replica_id != collection_replica {
-                    return Err(ReplicaNotHosting(replica_id));
-                }
+            if !collection.write_only && collection.target_replica.is_some_and(|r| r != replica_id)
+            {
+                return Err(ReplicaNotHostingIndex(replica_id));
             }
         }
 
@@ -1172,6 +1186,7 @@ impl<T: ComputeControllerTimestamp> InstanceState<T> {
 
 #[derive(Debug)]
 struct Collection<T> {
+    /// Whether a collection is write-only, i.e., we cannot read it directly like an index.
     write_only: bool,
     compute_dependencies: BTreeSet<GlobalId>,
     shared: SharedCollectionState<T>,
--- a/src/compute-client/src/controller.rs
+++ b/src/compute-client/src/controller.rs
@@ -754,6 +754,11 @@ where
     }
 
     /// Creates the described dataflow and initializes state for its output.
+    ///
+    /// Only materialized views are allowed to have a `target_replica`.
+    ///
+    /// Panics if called with a dataflow description that has index exports
+    /// when `target_replica` is set.
     pub fn create_dataflow(
         &mut self,
         instance_id: ComputeInstanceId,
@@ -769,6 +774,10 @@ where
             if !instance.replicas.contains(&replica_id) {
                 return Err(ReplicaMissing(replica_id));
             }
+            assert!(
+                dataflow.exported_index_ids().next().is_none(),
+                "Replica-targeted indexes are not supported"
+            );
         }
 
         // Validation: as_of
@@ -784,14 +793,6 @@ where
         let storage_ids = dataflow.imported_source_ids().collect();
         let mut import_read_holds = self.storage_collections.acquire_read_holds(storage_ids)?;
         for id in dataflow.imported_index_ids() {
-            if let Some(target_replica) = target_replica
-                && instance
-                    .collection(id)?
-                    .target_replica
-                    .is_some_and(|r| r != target_replica)
-            {
-                return Err(ReplicaNotHostingIndex(target_replica));
-            }
             let read_hold = instance.acquire_read_hold(id)?;
             import_read_holds.push(read_hold);
         }
@@ -821,7 +822,6 @@ where
                 compute_dependencies: dataflow.imported_index_ids().collect(),
                 shared: shared.clone(),
                 time_dependence: time_dependence.clone(),
-                target_replica,
             };
             instance.collections.insert(id, collection);
             shared_collection_state.insert(id, shared);
@@ -887,11 +887,6 @@ where
         if let Some(replica_id) = target_replica {
             if !instance.replicas.contains(&replica_id) {
                 return Err(ReplicaMissing(replica_id));
-            }
-            let collection = instance.collection(peek_target.id())?;
-            if !collection.write_only && collection.target_replica.is_some_and(|r| r != replica_id)
-            {
-                return Err(ReplicaNotHostingIndex(replica_id));
             }
         }
 
@@ -1193,8 +1188,6 @@ struct Collection<T> {
     /// The computed time dependence for this collection. None indicates no specific information,
     /// a value describes how the collection relates to wall-clock time.
     time_dependence: Option<TimeDependence>,
-    /// The single replica this collection is installed on, if any.
-    target_replica: Option<ReplicaId>,
 }
 
 impl<T: Timestamp> Collection<T> {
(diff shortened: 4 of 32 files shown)


src/compute-client/src/controller/instance.rs
```diff
@@ -1103,7 +1137,23 @@ where
...
+                    skipped_ids.extend(dataflow.export_ids());
+                    continue;
+                }
+                ComputeCommand::Schedule(cid) if skipped_ids.contains(cid) => continue,
+                ComputeCommand::AllowCompaction { id: cid, .. } if skipped_ids.contains(cid) => {
+                    continue;
+                }
+                ComputeCommand::AllowWrites(cid) if skipped_ids.contains(cid) => continue,
+                _ => {}
+            }
```
antiguru: Add a function `pub fn target_replica(&self) -> Option<ReplicaId>` to `ComputeCommand` containing the match block, and call this here instead.

src/compute-client/src/controller/instance.rs
```diff
@@ -1223,6 +1273,13 @@ where
             }
         }
 
+        let target_replica = dataflow.target_replica;
+        if let Some(replica_id) = target_replica {
+            if !self.replica_exists(replica_id) {
+                return Err(ReplicaMissing(replica_id));
+            }
+        }
```
antiguru: This needs a comment.

src/compute-types/Cargo.toml
```diff
@@ -14,6 +14,7 @@ columnar = "0.11.1"
 columnation = "0.1.2"
 differential-dataflow = "0.19.1"
 itertools = "0.14.0"
+mz-cluster-client = { path = "../cluster-client" }
```
antiguru: I'm not sure I like this...

antiguru (commented):
We need to ensure that replica-targeted objects can only be created by privileged users, i.e., `mz_system` for the moment.

src/compute-client/src/controller/instance.rs
```diff
@@ -1103,7 +1137,23 @@ where
...
+                continue;
+            }
+            match &command {
+                ComputeCommand::Schedule(cid) if skipped_ids.contains(cid) => continue,
+                ComputeCommand::AllowCompaction { id: cid, .. } if skipped_ids.contains(cid) => {
+                    continue;
+                }
+                ComputeCommand::AllowWrites(cid) if skipped_ids.contains(cid) => continue,
+                _ => {}
+            }
```
antiguru: Please extract a `collection_id` function here.

src/adapter/src/catalog/transact.rs
```diff
@@ -2892,6 +2892,19 @@ impl ObjectsToDrop {
...
+                // persist shard's upper frontier, causing reads to hang.
+                for (_id, entry) in state.get_entries() {
+                    if let CatalogItem::MaterializedView(mv) = entry.item() {
+                        if mv.target_replica == Some(replica_id)
+                            && !self.items.contains(&entry.id())
+                        {
+                            self.items.push(entry.id());
+                        }
+                    }
+                }
```
teskje, line 2911: We might want to consider not having that in the prototype. The danger is that we promote the prototype to a supported feature and forget about handling this properly. Where "properly" might mean requiring CASCADE, or emitting a warning, or something like that. (If we end up productionizing this feature as "introspection MVs", silently dropping would probably be the correct thing to do though!)

I can imagine having this makes it easier to write tests (like platform checks or parallel workload), if they expect to be able to always drop replicas. In this case, maybe just add a WARN log.

src/adapter/src/coord/sequencer/inner/create_materialized_view.rs
```diff
@@ -335,6 +336,15 @@ impl Coordinator {
             ..
         } = &plan;
 
+        // Replica-targeted materialized views are an internal feature.
+        if plan.materialized_view.target_replica.is_some() && !session.user().is_internal() {
```
teskje: As discussed: Consider using a feature flag instead! Shouldn't be more code but allows us to be more intentional about enabling this features, and makes it easier to enable in tests.

src/compute-types/src/dataflows.rs
```diff
@@ -66,6 +67,8 @@ pub struct DataflowDescription<P, S: 'static = (), T = mz_repr::Timestamp> {
     pub debug_name: String,
     /// Description of how the dataflow's progress relates to wall-clock time. None for unknown.
     pub time_dependence: Option<TimeDependence>,
+    /// If set, only install this dataflow on the specified replica.
+    pub target_replica: Option<ReplicaId>,
```
teskje: I'm thinking rather than putting this into the `DataflowDescription`, we put it next to it into the `CreateDataflow` command. I.e. either `CreateDataflow { dataflow, target_replica }` or inside a new wrapper struct, like we do for the `Peek` command. It's cleaner because a `DataflowDescription` keeps describing a dataflow, not where it runs.

src/compute-client/src/controller/instance.rs
```diff
@@ -300,6 +302,11 @@ impl<T: ComputeControllerTimestamp> Instance<T> {
 
         // Add per-replica collection state.
         for replica in self.replicas.values_mut() {
+            if let Some(target) = target_replica {
+                if replica.id != target {
```
teskje:
```suggestion
            if target_replica.is_some_and(|id| id != replica.id) {
```

src/compute-client/src/controller/instance.rs
```diff
@@ -345,6 +352,12 @@ impl<T: ComputeControllerTimestamp> Instance<T> {
             if collection.log_collection && !log_ids.contains(collection_id) {
                 continue;
             }
+            // Skip collections targeted at a different replica.
+            if let Some(target) = collection.target_replica {
+                if target != id {
+                    continue;
+                }
+            }
```
teskje: Nit: Could just add a `&& collection.target_replica.is_some_and(|id| id != replica.id)` to the `if` above.

src/compute-client/src/controller/instance.rs
```diff
@@ -1070,6 +1083,27 @@ where
...
+            let _ = replica.client.send(cmd);
+        }
+    }
+
+    /// Sends a command only to replicas that maintain the given collection.
+    ///
+    /// For untargeted collections, this sends to all replicas.
+    /// For collections targeted at a specific replica, this sends only to that replica.
+    fn send_to_collection_replicas(&mut self, id: GlobalId, cmd: ComputeCommand<T>) {
+        let target_replica = self.collections.get(&id).and_then(|c| c.target_replica);
```
teskje: Should be safe to call `self.expect_collection(id)` here. We should never send a command for a collection that doesn't exist.

src/compute-client/src/controller/instance.rs
```diff
@@ -1070,6 +1083,27 @@ where
...
+        if let Some(rid) = target_replica {
+            self.send_to_replica(rid, cmd);
+        } else {
+            self.send(cmd);
+        }
+    }
+
```
teskje: I'm thinking that instead of letting the caller worry about this, we should keep with the single `send` method and handle replica targeting transparently inside of that. In `send` you can just match on `&ComputeCommand` and `ComputeCommand::collection_id` and do the correct thing. This way the caller code is simpler and we have one place where we can confirm that we have the correct handling for each command type.

src/compute-client/src/protocol/command.rs
```diff
@@ -443,6 +444,26 @@ pub struct Peek<T = mz_repr::Timestamp> {
...
+    }
+
+    /// Returns the collection ID targeted by this command, if any.
+    pub fn collection_id(&self) -> Option<GlobalId> {
+        match self {
+            ComputeCommand::Schedule(id)
+            | ComputeCommand::AllowWrites(id)
+            | ComputeCommand::AllowCompaction { id, .. } => Some(*id),
+            _ => None,
+        }
```
teskje: `Peek` is missing, no? We don't need it for this feature, but good to include it, to avoid future surprises. Also please make this an exhaustive match!

src/compute-types/src/dataflows.rs
teskje: On second thought, does that need to go into the `CreateDataflow` command at all? The replica side isn't doing anything with this information, right? And we have it stored in the compute controller's `CollectionState`, which should be sufficient.

src/compute-client/src/protocol/command.rs
```diff
@@ -443,6 +444,26 @@ pub struct Peek<T = mz_repr::Timestamp> {
     pub otel_ctx: OpenTelemetryContext,
 }
 
+impl<T> ComputeCommand<T> {
+    /// Returns the `target_replica` if this command is a `CreateDataflow` with one set.
+    pub fn target_replica(&self) -> Option<ReplicaId> {
+        match self {
+            ComputeCommand::CreateDataflow(dataflow) => dataflow.target_replica,
+            _ => None,
+        }
```
teskje: Should also be exhaustive. But below I also advocate for just not storing the `target_replica` in the `CreateDataflow` command at all, which would delete this method.

src/compute-types/Cargo.toml
teskje: This also becomes unnecessary if we don't store the `target_replica` inside `ComputeCommand`!

src/sql-parser/src/parser.rs
```diff
@@ -3913,6 +3913,11 @@ impl<'a> Parser<'a> {
             None
         };
         let in_cluster = self.parse_optional_in_cluster()?;
+        let in_cluster_replica = if in_cluster.is_some() && self.parse_keyword(REPLICA) {
```
teskje: Do we need to require `IN CLUSTER`? Could just use the current cluster as usual.

teskje (commented):
I don't hate what I'm seeing! Looks all pretty straightforward. If we don't find any unexpected weird edge cases this seems good to merge behind a feature flag.

I have some suggestions to clean things up.

Also, to find these weird edge cases, can you make sure we create some of these replica-targeted MVs in our chaos tests? parallel-workload and/or platform-checks come to mind.

src/sql-parser/src/parser.rs
antiguru: Changed to `IN (CLUSTER ... (REPLICA ...)?)|(REPLICA ...)` so that IN is followed by something.

src/adapter/src/catalog/transact.rs
antiguru: Added a warning. The `CASCADE` option would be nice, but maybe a bit surprising? On the other hand, we'd only need it for replica-targeted objects, which aren't enabled by default.

The main issue is that managed clusters try to hide what replicas they're managing, and there wouldn't be a good place to add CASCADE, for example when scaling down.

src/compute-client/src/controller/instance.rs
antiguru: Combined two variants into `send` taking an optional `target_replica`.

src/adapter/src/catalog/transact.rs
teskje: Ah, good point about the cluster scaling. I guess we could always reject the DDL and require the user to manually drop the affected MVs. Depends on whether we think users will install load-bearing replica-targeted MVs that they would be sad about losing accidentally.

src/adapter-types/src/dyncfgs.rs
```diff
@@ -155,6 +155,13 @@ pub const ENABLE_S3_TABLES_REGION_CHECK: Config<bool> = Config::new(
     "Whether to enforce that S3 Tables connections are in the same region as the environment.",
 );
 
+/// Whether to enable creation of replica-targeted materialized views.
+pub const ENABLE_REPLICA_TARGETED_MATERIALIZED_VIEWS: Config<bool> = Config::new(
+    "enable_replica_targeted_materialized_views",
+    false,
+    "Enable creation of replica-targeted materialized views.",
+);
```
teskje: A thing to check: What happens when you create a replica-targeted MV, then disable the flag, then restart envd? Does it crash because we don't enable dyncfgs during item parsing? If so, this might need to be defined using the `feature_flags!` macro instead.

src/compute-client/src/controller/instance.rs
```diff
@@ -1057,16 +1071,30 @@ where
...
-    /// Sends a command to all replicas of this instance.
+    /// Sends a command to replicas of this instance.
+    ///
+    /// If `target_replica` is `Some`, the command is sent only to that replica. Otherwise,
+    /// for commands that target a specific collection (via [`ComputeCommand::collection_id`]),
+    /// the command is only sent to the replica maintaining that collection, if the collection
+    /// is targeted at a specific replica. Otherwise, the command is sent to all replicas.
     #[mz_ore::instrument(level = "debug")]
-    fn send(&mut self, cmd: ComputeCommand<T>) {
+    fn send(&mut self, cmd: ComputeCommand<T>, target_replica: Option<ReplicaId>) {
```
teskje: Do we need that extra argument? I think it's only used for `CreateDataflow` now, but we are initializing the collection state before we are sending the `CreateDataflow` command, so can't we just look up the export IDs in `self.collections` as well?

src/compute-client/src/controller/instance.rs
```diff
@@ -1300,7 +1348,7 @@ where
         // Initialize tracking of subscribes.
         for subscribe_id in dataflow.subscribe_ids() {
             self.subscribes
-                .insert(subscribe_id, ActiveSubscribe::new(subscribe_target_replica));
+                .insert(subscribe_id, ActiveSubscribe::new(target_replica));
```
teskje: Iiuc, we have now changed the behavior of replica-targeted subscribes to only send the subscribe dataflow to the targeted replica. Here we pass the target replica to the `ActiveSubscribe` so we can later use it to filter out subscribe responses. I think we can delete that now!

`ActiveSubscribe::target_replica` is also used in `subscribes_targeting`, but I think there we can just look it up in `self.collections` now.

src/compute-client/src/controller/instance.rs
```diff
@@ -1589,6 +1637,19 @@ where
...
+        let collection_target = self
+            .collections
+            .get(&target_id)
+            .and_then(|c| c.target_replica);
+        if let (Some(peek_replica), Some(collection_replica)) = (target_replica, collection_target)
+        {
+            if peek_replica != collection_replica {
+                return Err(ReplicaNotHosting(peek_replica));
+            }
+        }
```
teskje: I think this has a bug with replica-targeted MVs! Note that MVs consist of a dataflow (compute collection) and a persist shard (storage collection). The dataflow may be maintained by a single replica, but the persist shard can be read from any replica.

The fix is probably to special case this to index peeks.

src/compute-client/src/controller/error.rs
```diff
@@ -153,6 +153,9 @@ pub enum PeekError {
     /// The replica that the peek was issued against does not exist.
     #[error("replica does not exist: {0}")]
     ReplicaMissing(ReplicaId),
+    /// The target replica does not host the peeked collection.
+    #[error("replica {0} does not host the target collection")]
```
teskje: Nit: Maybe change this to "target index" to avoid some confusion like the MV thing I mention above.

src/compute-client/src/protocol/command.rs
```diff
@@ -443,6 +443,24 @@ pub struct Peek<T = mz_repr::Timestamp> {
...
 }
 
+impl<T> ComputeCommand<T> {
+    /// Returns the collection ID targeted by this command, if any.
+    pub fn collection_id(&self) -> Option<GlobalId> {
+        match self {
+            ComputeCommand::Schedule(id)
+            | ComputeCommand::AllowWrites(id)
+            | ComputeCommand::AllowCompaction { id, .. } => Some(*id),
+            ComputeCommand::Peek(peek) => Some(peek.target.id()),
```
teskje: Ah, this now also has the MV problem, if the peek is a persist peek. Maybe call this function `compute_collection_id`? Or, if it's not used in many places, consider just inlining it, to ensure we are always doing the thing we indent.

src/compute-client/src/controller.rs
```diff
@@ -754,27 +754,16 @@ where
...
 
         let instance = self.instance(instance_id)?;
 
-        // Validation: target replica
-        if let Some(replica_id) = subscribe_target_replica {
-            if !instance.replicas.contains(&replica_id) {
-                return Err(ReplicaMissing(replica_id));
-            }
-        }
-
```
teskje, line 777: Why don't we need this anymore?

src/compute-client/src/controller.rs
```diff
@@ -883,6 +873,12 @@ where
             if !instance.replicas.contains(&replica_id) {
                 return Err(ReplicaMissing(replica_id));
             }
+            let collection = instance.collection(peek_target.id())?;
+            if let Some(collection_replica) = collection.target_replica {
+                if replica_id != collection_replica {
+                    return Err(ReplicaNotHosting(replica_id));
+                }
+            }
```
teskje: Same issue with false positives for MVs.

src/compute-client/src/controller/instance.rs
```diff
@@ -1203,21 +1251,20 @@ where
...
-    /// configured to target that replica, i.e., only subscribe responses sent by that replica are
-    /// considered.
     #[mz_ore::instrument(level = "debug")]
     pub fn create_dataflow(
         &mut self,
         dataflow: DataflowDescription<mz_compute_types::plan::Plan<T>, (), T>,
         import_read_holds: Vec<ReadHold<T>>,
-        subscribe_target_replica: Option<ReplicaId>,
         mut shared_collection_state: BTreeMap<GlobalId, SharedCollectionState<T>>,
+        target_replica: Option<ReplicaId>,
```
teskje, line 1224: I _think_ that it is now possible to create an index targeted at one replica, and then a dependent index targeted at another replica, which will probably cause a panic on the cluster. We need to add additional validation to prevent this.

teskje (commented):
Most of my concerns now are around the ability to install replica-targeted indexes. Given that we don't expose that ability in SQL (which means we can't test it) how about we keep things simple and just disallow creating replica-targeted indexes in the controller?

src/compute-client/src/controller/instance.rs
antiguru: I think I put the validation in place, but can't test because we don't have syntax to create replica-targeted indexes (and I don't want to add it because I don't think makes sense.)

src/compute-client/src/controller.rs
antiguru: Sorry, Claude decided we don't need this anymore and I missed it :(

src/compute-client/src/protocol/command.rs
antiguru: Removed and inlined.

src/compute-client/src/controller/instance.rs
antiguru: I think I fixed this. It's a bit gross, but uses the `write_only` to check how we can read from the collection.

src/compute-client/src/controller/instance.rs
antiguru: Removed, and adjusted replica drop code.

src/compute-client/src/controller/instance.rs
antiguru: Nope, removed.

src/adapter-types/src/dyncfgs.rs
antiguru: I converted it to a feature flag, which seems to be the common pattern anyway.

src/compute-client/src/controller.rs
antiguru: Fixed by checking if we have a read policy (which we don't have for MVs.)

src/compute-client/src/controller/instance.rs
```diff
@@ -1058,7 +1096,47 @@ where
...
+                    let all_exports_skip = dataflow.export_ids().all(|eid| {
+                        self.collections
+                            .get(&eid)
+                            .and_then(|c| c.target_replica)
+                            .is_some_and(|rid| rid != id)
+                    });
+                    if all_exports_skip {
+                        skipped_ids.extend(dataflow.export_ids());
+                        continue;
+                    }
```
teskje: Not hugely important, since we're not using multi-export dataflows in practice, but since this logic is different than the one in `send` we might get confused if we ever start using it. Specifically, this allows installing collections even if they are targeted at a different replica. But then we'll never send `AllowCompaction` etc. commands for those collections later, because the `send` logic will skip them.

Easiest fix would be to do what `send` does and assert that all collections exported by a dataflow target the same replica. Then you can also get rid of `skipped_ids`.

src/compute-client/src/controller/instance.rs
```diff
@@ -1058,7 +1096,47 @@ where
...
+                // Should have been validated before.
+                ComputeCommand::Peek(peek) => {
+                    if let Some(peek_state) = self.peeks.get(&peek.uuid) {
+                        if let Some(target_replica) = peek_state.target_replica
+                            && target_replica != id
+                        {
+                            continue;
+                        }
+                    }
+                }
```
teskje: This logic also differs from `send`. I think `send` still relies on the behavior that peek responses from the wrong replicas are ignored? Seems fine to change that, but maybe do it in a different PR to avoid blowing up the scope too much.

teskje: Ah, since we're still allowing replica-targeted indexes, the `send` logic for peeks is incorrect, no? It allows sending an index peek to a replica that doesn't have that index. I still think we should just disallow replica-targeted indexes.

src/compute-client/src/controller/instance.rs
antiguru: Added an assertion to `create_dataflow` so we guard against creating replica-targeted indexes, which simplifies a bunch of the logic and allows us to remove the error variants. Extracted the `target_replica` logic into a function.

src/compute-client/src/controller.rs
```diff
@@ -755,24 +755,29 @@ where
 
     /// Creates the described dataflow and initializes state for its output.
     ///
-    /// If a `subscribe_target_replica` is given, any subscribes exported by the dataflow are
-    /// configured to target that replica, i.e., only subscribe responses sent by that replica are
-    /// considered.
+    /// Only materialized views are allowed to have a `target_replica`.
```
teskje:
```suggestion
    /// Only materialized views and subscribes are allowed to have a `target_replica`.
```

teskje (approved):
Thanks, nice simplification!

antiguru:
Thanks for the review!

PR #35328, opened 2026-03-05
compute: move MV sink persist I/O off Timely thread

antiguru:
Convert the MV sink's mint, write, and append operators from `AsyncOperatorBuilder` to `OperatorBuilderRc` (sync) with dedicated Tokio tasks for persist I/O.
This removes async overhead from the Timely scheduling loop while keeping persist latency isolated from dataflow scheduling.

Key changes:
* mint: Tokio task watches persist upper via `SyncActivator`; operator manages capabilities and batch description minting
* write: Tokio task owns `WriteHandle` and `CorrectionBuffer`; operator sends commands via channel; `ArcActivator` coalesces redundant cross-thread activations
* append: Tokio task performs `compare_and_append`; operator forwards descriptions, batches, and frontier updates
* `CorrectionLogger` tracks net batch/record/size counts and retracts outstanding logging state on drop, fixing stale introspection rows after dataflow cancellation
* `ChannelLogging` bridges `Correction` events from Tokio task to Timely thread for introspection logging
* All tasks use `abort_on_drop` for prompt cleanup on dataflow shutdown

Shares the infra commit (`compute/storage: add Send-safe infrastructure for sync Timely operators`) with #35450.
Rebase before merging to deduplicate.

### DIFF
--- a/src/compute/src/sink/correction.rs
+++ b/src/compute/src/sink/correction.rs
@@ -10,11 +10,10 @@
 //! The `Correction` data structure used by `persist_sink::write_batches` to stash updates before
 //! they are written into batches.
 
-use std::cell::RefCell;
 use std::collections::BTreeMap;
 use std::fmt;
+use std::num::NonZeroIsize;
 use std::ops::{AddAssign, Bound, RangeBounds, SubAssign};
-use std::rc::Rc;
 
 use differential_dataflow::consolidation::{consolidate, consolidate_updates};
 use differential_dataflow::logging::{BatchEvent, DropEvent};
@@ -29,6 +28,7 @@ use mz_persist_client::metrics::{SinkMetrics, SinkWorkerMetrics, UpdateDelta};
 use mz_repr::{Diff, Timestamp};
 use timely::PartialOrder;
 use timely::progress::Antichain;
+use tokio::sync::mpsc;
 
 use crate::logging::compute::{
     ArrangementHeapAllocations, ArrangementHeapCapacity, ArrangementHeapSize,
@@ -53,7 +53,7 @@ impl<D: Data> Correction<D> {
     pub fn new(
         metrics: SinkMetrics,
         worker_metrics: SinkWorkerMetrics,
-        logging: Option<Logging>,
+        logging: Option<ChannelLogging>,
         config: &ConfigSet,
     ) -> Self {
         if ENABLE_CORRECTION_V2.get(config) {
@@ -92,7 +92,7 @@ impl<D: Data> Correction<D> {
     pub fn updates_before(
         &mut self,
         upper: &Antichain<Timestamp>,
-    ) -> Box<dyn Iterator<Item = (D, Timestamp, Diff)> + '_> {
+    ) -> Box<dyn Iterator<Item = (D, Timestamp, Diff)> + Send + '_> {
         match self {
             Self::V1(c) => Box::new(c.updates_before(upper)),
             Self::V2(c) => Box::new(c.updates_before(upper)),
@@ -282,7 +282,7 @@ impl<D: Data> CorrectionV1<D> {
     pub fn updates_before<'a>(
         &'a mut self,
         upper: &Antichain<Timestamp>,
-    ) -> impl Iterator<Item = (D, Timestamp, Diff)> + ExactSizeIterator + use<'a, D> {
+    ) -> impl Iterator<Item = (D, Timestamp, Diff)> + ExactSizeIterator + Send + use<'a, D> {
         let lower = Antichain::from_elem(Timestamp::MIN);
         self.updates_within(&lower, upper)
     }
@@ -535,76 +535,96 @@ impl AddAssign<Self> for SizeMetrics {
     }
 }
 
-/// Helper for correction buffer logging.
+/// A logging event sent from the Tokio task back to the Timely thread.
+#[derive(Debug)]
+pub(super) enum LoggingEvent {
+    ChainCreated(usize),
+    ChainDropped(usize),
+    SizeDiff(NonZeroIsize),
+    CapacityDiff(NonZeroIsize),
+    AllocationsDiff(NonZeroIsize),
+}
+
+/// Channel-based logging for corrections on a Tokio task. `Send`-safe.
 ///
+/// Sends logging events to the Timely thread, where they are applied to the real `Logging`
+/// instance. This allows corrections on the Tokio task to participate in introspection logging
+/// without holding `Rc<RefCell<..>>`.
+#[derive(Clone, Debug)]
+pub(super) struct ChannelLogging(mpsc::UnboundedSender<LoggingEvent>);
+
+impl ChannelLogging {
+    pub fn new(tx: mpsc::UnboundedSender<LoggingEvent>) -> Self {
+        Self(tx)
+    }
+
+    pub fn chain_created(&self, updates: usize) {
+        let _ = self.0.send(LoggingEvent::ChainCreated(updates));
+    }
+
+    pub fn chain_dropped(&self, updates: usize) {
+        let _ = self.0.send(LoggingEvent::ChainDropped(updates));
+    }
+
+    pub fn report_size_diff(&self, diff: isize) {
+        if let Some(diff) = NonZeroIsize::new(diff) {
+            let _ = self.0.send(LoggingEvent::SizeDiff(diff));
+        }
+    }
+
+    pub fn report_capacity_diff(&self, diff: isize) {
+        if let Some(diff) = NonZeroIsize::new(diff) {
+            let _ = self.0.send(LoggingEvent::CapacityDiff(diff));
+        }
+    }
+
+    pub fn report_allocations_diff(&self, diff: isize) {
+        if let Some(diff) = NonZeroIsize::new(diff) {
+            let _ = self.0.send(LoggingEvent::AllocationsDiff(diff));
+        }
+    }
+}
+
+/// State for correction buffer logging on the Timely thread.
+///
+/// Drains [`LoggingEvent`]s sent by [`ChannelLogging`] from the Tokio task and applies them
+/// to the compute and differential loggers. Emits `ArrangementHeapSizeOperator` on construction
+/// and `ArrangementHeapSizeOperatorDrop` on drop.
 // TODO: Correction buffer logging currently reuses the arrangement batch and size logging. This
 // isn't strictly correct as a correction buffer is not an arrangement. Consider refactoring this
 // to be about "operator sizes" instead.
-#[derive(Clone, Debug)]
-pub(super) struct Logging(Rc<RefCell<LoggingInner>>);
-
-impl Logging {
-    pub fn new(
-        compute_logger: ComputeLogger,
-        differential_logger: differential_dataflow::logging::Logger,
-        operator_id: usize,
-        address: Vec<usize>,
-    ) -> Self {
-        let inner = LoggingInner::new(compute_logger, differential_logger, operator_id, address);
-        Self(Rc::new(RefCell::new(inner)))
-    }
-
-    /// A new chain was created.
-    pub fn chain_created(&self, updates: usize) {
-        self.0.borrow_mut().chain_created(updates);
-    }
-
-    /// A chain was dropped.
-    pub fn chain_dropped(&self, updates: usize) {
-        self.0.borrow_mut().chain_dropped(updates);
-    }
-
-    /// Report a heap size diff.
-    pub fn report_size_diff(&self, diff: isize) {
-        self.0.borrow_mut().report_size_diff(diff);
-    }
-
-    /// Report a heap capacity diff.
-    pub fn report_capacity_diff(&self, diff: isize) {
-        self.0.borrow_mut().report_capacity_diff(diff);
-    }
-
-    /// Report a heap allocations diff.
-    pub fn report_allocations_diff(&self, diff: isize) {
-        self.0.borrow_mut().report_allocations_diff(diff);
-    }
-}
-
-/// Inner state for correction buffer logging.
-///
-/// This is separate from `Logging` and shared via an `Rc`, to ensure we only emit
-/// `ArrangementHeapSizeOperator{Drop}` events once, even though we pass the logger to both the Ok
-/// and the Err correction buffer.
-struct LoggingInner {
+pub(super) struct CorrectionLogger {
     compute_logger: ComputeLogger,
     differential_logger: differential_dataflow::logging::Logger,
     operator_id: usize,
+    rx: mpsc::UnboundedReceiver<LoggingEvent>,
+    /// Net number of batches logged (BatchEvent - DropEvent).
+    net_batches: isize,
+    /// Net number of records logged across all batch/drop/merge events.
+    net_records: isize,
+    /// Cumulative heap size delta, for retraction on drop.
+    net_size: isize,
+    /// Cumulative heap capacity delta, for retraction on drop.
+    net_capacity: isize,
+    /// Cumulative heap allocations delta, for retraction on drop.
+    net_allocations: isize,
 }
 
-impl fmt::Debug for LoggingInner {
+impl fmt::Debug for CorrectionLogger {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("LoggingInner")
+        f.debug_struct("CorrectionLogger")
             .field("operator_id", &self.operator_id)
             .finish_non_exhaustive()
     }
 }
 
-impl LoggingInner {
-    fn new(
+impl CorrectionLogger {
+    pub fn new(
         compute_logger: ComputeLogger,
         differential_logger: differential_dataflow::logging::Logger,
         operator_id: usize,
         address: Vec<usize>,
+        rx: mpsc::UnboundedReceiver<LoggingEvent>,
     ) -> Self {
         compute_logger.log(&ComputeEvent::ArrangementHeapSizeOperator(
             ArrangementHeapSizeOperator {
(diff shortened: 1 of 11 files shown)


src/compute/src/sink/materialized_view.rs
```diff
@@ -497,117 +498,150 @@ mod mint {
...
+
+        // Spawn a Tokio task to watch the persist shard's upper frontier (active worker only).
+        //
+        // We collect the persist frontier from a write handle directly, rather than inspecting
+        // the `persist` stream, because the latter has two annoying glitches:
+        //  (a) It starts at the shard's read frontier, not its write frontier.
+        //  (b) It can lag behind if there are spikes in ingested data.
+        if worker_id == active_worker_id {
+            let sync_activator = scope.sync_activator_for(info.address.to_vec());
+            mz_ore::task::spawn(
```
teskje: I assume many of these tasks have a chance of getting stuck indefinitely when the dataflow is dropped. Seems like should keep `AbortOnDropHandle`s for them?

src/compute/src/sink/materialized_view.rs
```diff
@@ -497,117 +498,150 @@ mod mint {
...
-                        else => return,
-                    }
+            let mut state = (worker_id == active_worker_id)
+                .then(|| State::new(sink_id, worker_count, as_of, read_only));
+
+            move |frontiers| {
+                // Check shutdown.
+                if shutdown_handle.all_pressed() {
+                    return false;
                 }
```
teskje: Do we still need the shutdown buttons now that we're using `drop_dataflow` in compute?

src/compute/src/sink/materialized_view.rs
```diff
@@ -783,124 +834,235 @@ mod write {
...
-                            Event::Progress(_frontier) => None,
+                }
+
+                // Accept data from all inputs.
+                desired_ok_input.for_each(|_cap, data| {
+                    if let Some(s) = &mut state {
+                        s.corrections.ok.insert(data);
+                    } else {
+                        pending_desired_ok.push(std::mem::take(data));
+                    }
```
teskje: Could we get rid of the `pending_desired_*` and instead initialize the correction buffers outside of `State` and then move them into `State` when we initialize it?

src/compute/src/sink/materialized_view.rs
```diff
@@ -783,124 +834,235 @@ mod write {
...
+        sync_activator: &timely::scheduling::SyncActivator,
+    ) {
+        while let Some(req) = batch_req_rx.recv().await {
+            let batch = writer
+                .batch(req.updates, req.lower, req.upper)
+                .await
+                .expect("valid usage")
+                .into_transmittable_batch();
+            if batch_resp_tx.send(WriteBatchResponse { batch }).is_err() {
+                return;
```
teskje: I think we should try to `batch.delete()` the batch if we can't send it, to avoid leakage.

teskje (commented):
The intention checks out to me. I'll note that for me this makes it noticeably harder to follow the logic and be sure that it's still correct, so we should only make this change if we have measured performance improvements. Aiui, the thinking is that Timely activations are expensive enough to justify the overhead of the additional tokio tasks and channels. It'd be great to have some numbers!

src/compute/src/sink/materialized_view.rs
antiguru: Fixed.

src/compute/src/sink/materialized_view.rs
antiguru: Nope!

src/compute/src/sink/correction.rs
```diff
@@ -537,76 +537,90 @@ impl AddAssign<Self> for SizeMetrics {
...
 ///
-// TODO: Correction buffer logging currently reuses the arrangement batch and size logging. This
-// isn't strictly correct as a correction buffer is not an arrangement. Consider refactoring this
-// to be about "operator sizes" instead.
+/// Sends logging events to the Timely thread, where they are applied to the real `Logging`
+/// instance. This allows corrections on the Tokio task to participate in introspection logging
+/// without holding `Rc<RefCell<..>>`.
 #[derive(Clone, Debug)]
-pub(super) struct Logging(Rc<RefCell<LoggingInner>>);
+pub(super) struct ChannelLogging(mpsc::UnboundedSender<LoggingEvent>);
```
antiguru, line 554: Rename `ChannelLogging` to `Logging` to reduce diff size.

src/compute/src/sink/materialized_view.rs
```diff
@@ -760,152 +834,285 @@ mod write {
...
         let exchange_err = |(d, _, _): &(DataflowError, Timestamp, Diff)| d.hashed();
 
-        let mut desired_inputs = OkErr::new(
-            op.new_disconnected_input(desired.ok, Exchange::new(exchange_ok)),
-            op.new_disconnected_input(desired.err, Exchange::new(exchange_err)),
-        );
-        let mut persist_inputs = OkErr::new(
-            op.new_disconnected_input(persist.ok, Exchange::new(exchange_ok)),
-            op.new_disconnected_input(persist.err, Exchange::new(exchange_err)),
+        // Disconnected inputs: no connection to any output.
```
antiguru: Move comment to output (where we specify the special connection).

src/compute/src/sink/materialized_view.rs
```diff
@@ -760,152 +834,285 @@ mod write {
...
+
+        // descs -> output 0 (batches)
+        let mut descs_input = builder.new_input(descs, Pipeline);
+
+        let (batches_output, batches_output_stream) =
+            builder.new_output_connection([(4, Antichain::from_elem(Default::default()))]);
+        let mut batches_output = OutputBuilder::from(batches_output);
+
+        // Obtain SinkMetrics synchronously from the persist client cache.
+        let sink_metrics = persist_api.persist_clients.metrics().sink.clone();
```
antiguru: Is this OK from an API perspective? I think this is different from how it was previously. Maybe add a comment.

src/compute/src/sink/materialized_view.rs
```diff
@@ -760,152 +834,285 @@ mod write {
...
+                            state.persist_frontiers.frontier().to_owned(),
+                        ))
+                        .expect("write task unexpectedly gone");
+
+                    // Check if we need to force consolidation.
+                    if state.should_force_consolidation() {
+                        cmd_tx
+                            .send(WriteCommand::ForceConsolidation)
+                            .expect("write task unexpectedly gone");
                     }
```
antiguru: This `if` is repeated just below the block, do we really need it twice?

antiguru:
bugbot run

frankmcsherry:
> Aiui, the thinking is that Timely activations are expensive enough to justify the overhead of the additional tokio tasks and channels.

I think there's an additional issue that is maybe addressed that the async operators would not only schedule themselves often, but would also yield quite happily. I believe we figured this was the most likely candidate for "why Frank's catalog server became permanently wedged". The operators do need to be designed to handle overload gracefully, and not fall permanently behind.

> It'd be great to have some numbers!

+1 on this. It may be that this is a recognized issue in a tracker somewhere, but if not definitely worth socializing what's up and why.

teskje:
Doing a diff between `materialized_view.rs` and `materialized_view_v2.rs` I notice that v2 deletes a lot of the comments that were present in v1. Is that intentional? I think many of them are useful context that we want to keep.

antiguru:
> Doing a diff between materialized_view.rs and materialized_view_v2.rs I notice that v2 deletes a lot of the comments that were present in v1. Is that intentional? I think many of them are useful context that we want to keep.

No, you're right, it's one of the more annoying sides of using Claude code, it tends to forget comments.

antiguru:
### Investigation notes: Kafka benchmark regression at c103d5388f

Benchmark failure: `SELECT * FROM kafka_mv (standalone)` — p99 890ms vs 400ms target, qps 7.37 vs 15 target. Likely related: `test_replicated_compute_cluster_on_failing_node` — "Failed to achieve 'live data without disruption in latency' using SERIALIZABLE within 10s". The 890ms shape matches the 1s Timely tick, suggesting frontier advancement lagging until a tick.

**Confirmed so far:**

1. `FrontierInterest` default is `Always` for all inputs (`builder_raw.rs:115` pushes `Always` on every `new_input_connection`). WRITE/MINT/APPEND get scheduled on every frontier change of every input — same as v1 async.

2. Connectivity in v2 matches v1:
   - MINT: inputs 0,1 (desired_ok/err) connect to outputs 0,1 respectively; `desc_output` (2) governed by `cap_set` only.
   - WRITE: `descs_input` (4) → `batches_output` (0) with identity summary; desired/persist inputs disconnected. Ordering (inputs first, then `new_output_connection([(4, identity)])`) works because the output-creation call explicitly back-fills `summaries[4].add_port(0, identity)`.
   - APPEND: no outputs.

3. Tokio → Timely activations:
   - mint's `persist_watch`: activates mint after every frontier send ✓
   - mint's `read_only_watch`: activates mint on change ✓
   - write task: activates write (via `ArcActivator`) after every command ✓
   - **append task: activates nothing** — but append has no outputs, so direct activation of its own operator doesn't propagate anything.

4. `wait_for_upper_past` uses persist's in-process `watch()` — no polling fallback that would explain 890ms.

**Where I got stuck:**

The specific 890ms-producing path isn't obvious from code. The `persist_watch → mint` activation should fire immediately on shard upper change. Mint's `cap_set.try_downgrade(persist_frontier)` should then propagate through `desc_output → batches_output → …`.

**To localize next, I'd add latency logging:**

- In `mint::persist_watch` task, record the wall-clock time right before `sync_activator.activate()`.
- In `mint::drain_persist_rx`, record the time when the frontier is drained.

If that gap is ~800ms, the activation-to-scheduling is delayed (Timely thread starvation by the sync ops?). If it's small, the lag is further downstream — e.g., the `shared_frontier` update cascade to the coordinator via `sink_write_frontier`, or the peek path on the replica.

**Hypotheses still worth exploring:**

- `shared_frontier` is `Rc<RefCell<>>` and updated only on the Timely thread inside `mint::drain_persist_rx`. In v1 async, the `async_stream!` updated `shared_frontier` *immediately* as the stream yielded (inside the operator's async context — same thread). In v2 sync, the round-trip Tokio→channel→activator→mint-scheduling→drain adds latency. If the compute worker's reporting of `sink_write_frontier` to the controller lags as a result, SERIALIZABLE peeks relying on that reported frontier could wait.
- `mz_storage_operators::persist_source` activation on shard-upper advance. Shared with v1 so not a regression per se, but worth verifying its behavior on the sync-sink path.
- The peek path — how does an indexed MV under SERIALIZABLE pick its read timestamp on a replicated cluster, and does that path wait on the MV's reported storage frontier?

antiguru:
### Cleanups after review

Force-pushed to drop/revert two changes that had been flagged as problematic in the past:

1. **Dropped `storage: restore identity MFP filtering in persist_source_core`.** `git log -S` on that line shows it has been added and reverted multiple times (`d943a9789f` introduced, `55db0f4ab9`/`a6345895c2`/`6b0dc2b389` reverted, etc.). Not bringing it back in this PR.

2. **Reverted `internal_control.rs` changes from the Send-safe infrastructure commit.** The switch from `Rc<RefCell<Activator>>` to `Arc<Mutex<SyncActivator>>` wasn't actually required by this PR — only `persist_source`'s `+ Send` bounds and `StartSignal::into_send_future()` are. `internal_control.rs` keeps its original `Activator` (which had been removed at one point and then backed out).

Branch is now 5 commits (down from 6). Last commit `7ae30ce7e2`.

### Still to validate

- **Run CI with `enable_compute_sync_mv_sink = false`** to see whether the Kafka benchmark regression is inherent to the new sync sink or reproduces against the old async sink as well.
- **Tokio cooperative-scheduling concern:** the sync sink moves persist I/O (batch creation in the write task, `compare_and_append` in the append task) onto Tokio worker threads. If other long-running Tokio tasks are on the same runtime, they can delay these I/O futures — and delayed persist I/O is exactly what the 890ms latency shape looks like. Worth checking whether the async-sink version was inadvertently benefiting from Timely's scheduler keeping I/O off the shared Tokio pool.

antiguru:
The regression turned out to be a fluke. The results in main are about the same, and in my run it just happened to be slightly worse, which triggered the benchmark failure.

antiguru:
## Staging benchmark + profile findings

Ran a bulk MV hydration A/B on staging with `enable_compute_sync_mv_sink` toggled, 5 runs × 5 copies each (n=25 per flag). The workload creates 5 `auctions_i` + 5 `bids_i` MVs on a fresh `200cc` cluster with `hydration_concurrency = 1`, so each MV hydrates serially. Bids MV is the interesting one (~22M rows, heavy fan-out).

### Hydration (per-MV, cumulative queue wait stripped)

| MV | OFF median | ON median | Δ | 95% CI (bootstrap) |
|---|---|---|---|---|
| auctions | 1,011.6 ms | 1,006.1 ms | −0.5% | [−35.9, +84.0] ms |
| bids | 126,945.7 ms | 134,176.0 ms | **+5.7%** | [−5,797, +14,788] ms |

Sign is consistent across three separate experiments (all showed ON slightly slower on bids, 2.6–5.7%). 95% CI spans zero at n=25, so statistically borderline — but the direction is reproducible.

### CPU profile (Parca, ~1h windows per flag, relative %)

| Function | OFF | ON | Δ |
|---|---|---|---|
| `timely::subgraph::schedule` (cum) | 23.0% | 17.3% | **−5.7pp** |
| `tokio::task::harness::poll` | 3.1% | 8.8% | +5.7pp |
| `tokio::runtime::blocking::pool::run` | 1.6% | 4.4% | +2.8pp |
| `tracing::Instrumented::poll` | 4.3% | 10.7% | +6.4pp |
| `CorrectionV2::insert_inner` | 2.5% | 3.5% | +1.0pp |
| `CorrectionV2::merge_{cursors,chains}` | 2.8% | 3.9% | +1.1pp |

PR goal confirmed: Timely scheduling time ↓ ~6pp; that work shifts onto Tokio tasks as designed. The correction path grows ~40% in absolute CPU, which matches the wall-clock regression.

### Memory profile (in-use bytes)

| Function | Δ |
|---|---|
| **Total in-use** | **−10%** |
| `CorrectionV2::Chain::push` | −16% |
| `CorrectionV2::merge_cursors` | −16% |
| `mz_ore::region::Region::new_auto` | −12% |
| `tracing::Instrumented::poll` | +110% |
| `tokio::task::harness::poll` | +178% |

Net memory is ~10% **lower** on ON. Correction buffer drains faster (−15%), more than offsetting Tokio task frame overhead.

### Verdict

- Steady-state goal achieved: CPU redistributed off Timely, memory down.
- Bulk hydration pays a small price (~5%) due to cross-thread coordination cost.

### Next step: coalesce `WriteCommand` sends

Current `write::render` emits **one `mpsc::send` per chunk** across four inputs plus separate sends for frontier/consolidation. During hydration this is thousands of tiny channel messages producing many small `Correction::insert` calls.

Plan: accumulate all four data streams into local `Vec`s during a single activation sweep and send at most **one** `WriteCommand::Batch { desired_ok, desired_err, persist_ok, persist_err, persist_frontier, force_consolidation }` per activation. Larger `Correction::insert` batches → better amortized sort/merge. Expected to close most of the hydration regression without affecting the steady-state wins.

DAlperin:
@antiguru do you have the follow up benchmarks with the batch coalescing?

antiguru:
> @antiguru do you have the follow up benchmarks with the batch coalescing?

It's on my list and stalled on CI and it's late here :)

Claude didn't see it, but one nice side effect is that we don't lose access to metrics during hydration. Previously, we'd not report because Timely wouldn't run frequently enough, but with a decoupled infra, it gets a chance to report. What you see in the following picture is on the left a workload using the current mv sink, and on the right the new mv sink.
image

On the flip side, this is less run-to-completion. Not sure what side I'm on.

antiguru:
## Coalescing follow-up (commit 8b0aa8a009)

Reran the staging A/B after pushing the coalesced `WriteCommand::Batch` variant. Same workload (n=25 per label, `200cc` cluster, `hydration_concurrency=1`, ~22M-row bids MV × 5 copies).

### Hydration (per-MV, cumulative queue wait stripped)

| Label | bids median | vs OFF |
|---|---|---|
| off (old sink) | 126,946 ms | — |
| on (pre-coalescing) | 134,176 ms | +5.70% |
| **on2 (coalesced)** | **109,426 ms** | **−13.80%** |

| Comparison | Δ median | 95% CI (bootstrap) |
|---|---|---|
| on2 vs on | **−24,750 ms (−18.45%)** | [−30,704, −17,995] ms |
| on2 vs off | **−17,520 ms (−13.80%)** | [−29,697, −11,616] ms |

Both CIs are comfortably below zero. The coalesced sink is not just regression-free — it's a meaningful improvement over the old sink. Tail tightened too: bids spread dropped from 58s (off) / 51s (on) to **36s** (on2).

auctions unchanged (noise-dominated at ~1 s median).

### CPU profile (Parca, ~50 min window)

| Function | OFF | ON | ON2 | Δ (on2 − on) |
|---|---|---|---|---|
| `CorrectionV2::insert_inner` | 2.5% | 3.5% | **2.2%** | **−37% relative** |
| `CorrectionV2::merge_cursors` | 1.5% | 2.0% | **1.0%** | **−50% relative** |
| `timely::subgraph::schedule` | 23.0% | 17.3% | 17.5% | +0.2pp |
| `tracing::Instrumented::poll` | 4.3% | 10.7% | 10.5% | −0.2pp |
| `MirScalarExpr::eval` | 3.7% | 4.6% | 5.7% | +1.1pp |

Correction path collapses exactly as coalescing predicts: fewer, larger inserts → fewer merges → fewer cursor allocations. The `MirScalarExpr::eval` rise is just a share shift — wall time shrank, so the non-sink work occupies more of the total.

### Memory profile (jemalloc, in-use bytes)

| | OFF | ON | **ON2** | Δ (on2 − on) |
|---|---|---|---|---|
| **Total in-use** | 4,433 | 4,001 | **3,639** | **−10%** |
| `CorrectionV2::Chain::push` | 1,967 | 1,660 | 1,283 | −23% |
| `CorrectionV2::merge_cursors` (alloc) | 1,967 | 1,660 | 1,250 | −25% |
| `mz_ore::region::Region::new_auto` | 1,708 | 1,509 | 1,244 | −18% |

Net in-use is now **18% below the OFF baseline**. Correction buffer residency shrank because the draining task processes larger grouped inserts, reducing the in-flight window.

### Verdict

The ~5% hydration regression profiled to per-chunk channel volume and small `Correction::insert` batches. Coalescing all four input streams + frontier + consolidation flag into one `WriteCommand::Batch` per activation (commit 8b0aa8a009) addresses that root cause. Result: hydration faster, correction CPU sharply lower, memory lower — across the board.

src/compute/src/sink/correction.rs
```diff
@@ -535,76 +535,96 @@ impl AddAssign<Self> for SizeMetrics {
...
```
DAlperin, line 554: not a big deal for now but if this becomes a pattern we should extract it as an API somehow

src/compute/src/sink/materialized_view_v2.rs
```diff
@@ -0,0 +1,1246 @@
...
+        let mut task_handles = Vec::new();
+        let read_only = *read_only_rx.borrow_and_update();
+        let mut state = None;
+        if worker_id == active_worker_id {
+            // Spawn a Tokio task to watch the persist shard's upper frontier.
+            //
+            // We collect the persist frontier from a write handle directly, rather than
+            // inspecting the `persist` stream, because the latter has two annoying glitches:
+            //  (a) It starts at the shard's read frontier, not its write frontier.
+            //  (b) It can lag behind if there are spikes in ingested data.
```
DAlperin, line 208: Is the goal to still have tick granular batch descriptions to work through when there are spikes?

src/compute/src/sink/materialized_view_v2.rs
```diff
@@ -0,0 +1,1246 @@
...
+            };
+            loop {
+                match rx.try_recv() {
+                    Ok(frontier) => {
+                        shared_frontier.borrow_mut().clone_from(&frontier);
+                        let done = frontier.is_empty();
+                        self.advance_persist_frontier(frontier.borrow());
+                        if done {
+                            return;
+                        }
```
DAlperin, line 439: in the case where there are multiple messages to receive this is purely wasted work and could be coalesced, but in practice the work is cheap so it's probably fine.

src/compute/src/sink/materialized_view_v2.rs
```diff
@@ -0,0 +1,1246 @@
...
+                self.trace("requesting correction consolidation");
+                self.force_consolidation_after = None;
+                true
+            } else {
+                false
+            }
+        }
+
+        fn absorb_batch_description(&mut self, desc: BatchDescription, cap: Capability<Timestamp>) {
+            self.batch_description = Some((desc, cap));
```
DAlperin, line 983: should probably enforce monotonicity here

src/compute/src/sink/materialized_view_v2.rs
```diff
@@ -0,0 +1,1246 @@
...
+                    prev_batches_frontier.extend(new_batches_frontier.iter().cloned());
+                    cmd_tx
+                        .send(AppendCommand::BatchesFrontier(
+                            new_batches_frontier.to_owned(),
+                        ))
+                        .expect("append task unexpectedly gone");
+                }
+            }
+        });
+    }
```
DAlperin, line 1122: It took me a bit to convince myself this was safe. I think the load bearing assumption this relies on is that `BatchesFrontier` must follow `Batches` or else `maybe_append_batches` could be triggered prematurely. Perhaps a comment here explaining this invariant.

DAlperin: Lets actually document the ordering requirements for each of the operators

DAlperin (approved):
Approved with a few notes! Very cool!

src/compute/src/sink/correction.rs
antiguru: Yeah, it's not nice but pretty localized. +1 on extracting an API if we encounter the same pattern again.

src/compute/src/sink/materialized_view_v2.rs
antiguru: Yes; updated the code to include a comment.

src/compute/src/sink/materialized_view_v2.rs
antiguru: Fixed

src/compute/src/sink/materialized_view_v2.rs
antiguru: Fixed.

src/compute/src/sink/materialized_view_v2.rs
antiguru: Added comments.

PR #35407, opened 2026-03-10
storage: refactor upsert source to use simplified control flow

DAlperin:
Instead of doing partial timestamp omission we now rely on a rocksdb
cache to consolidate as we go, only omitting full data. The I/O load
should be about the same since we had to go to rocksb anyway for each
key during partial emission.

We could potentially optimize further by having an in memory stash if
data is small and only spilling to rocks after some threshold but in
absence of evidence I'm avoiding the complication.

Further we could conceivably update our Rocks wrapper to support
multiple CF, but again, I don't think it's super worth it for now.


DAlperin:
bugbot run

PR #35418, opened 2026-03-11
Introduce v1 of the Materialize CRD

alex-hunt-materialize:
Adds a new `v1` version of the Materialize CRD, and a conversion webhook in orchestratord to convert between `v1` and `v1alpha1`.

The `v1` CRD removes the `requestRollout` field in favor of calculating a hash over a subset of spec fields to determine when a rollout is needed.

`v1alpha1` remains the storage version, and orchestratord still reconciles `v1alpha1`, so this is backwards compatible with existing automation. When a `v1` CR is applied, the conversion webhook converts it to `v1alpha1`, injecting a `requestRollout` UUID derived deterministically (UUIDv5) from the spec hash: when the hash changes a rollout is triggered immediately (the intended `v1` behavior), and when it hasn't, the same UUID is produced and no unintended rollout occurs.

The whole thing is opt-in:

- The `v1` CRD and the conversion webhook are only installed when the `operator.args.installV1CRD` helm value is set (`--install-v1-crd`). Without it, only `v1alpha1` is registered, no webhook serving certificate or service is created, and cert-manager is not required.
- Even with `installV1CRD` set, existing `v1alpha1` CRs behave exactly as before; users switch individual instances by applying a `v1` CR.

The webhook serving certificate is issued from a long-lived root CA (via an intermediate cert-manager issuer), so routine serving-certificate rotations don't invalidate the CA bundle registered in the CRD. orchestratord periodically reloads the certificate from disk (`--webhook-cert-reload-interval`) and re-registers the CRDs with the new CA bundle if the CA itself ever changes.

Separate documentation PR is at https://github.com/MaterializeInc/materialize/pull/35508

Cloud companion PR is at https://github.com/MaterializeInc/cloud/pull/12275

### Motivation

Simplifying rollout behavior.
Part of https://linear.app/materializeinc/issue/DEP-7/design-simplifying-upgrade-rollouts-node-rolls-converted-to-project

### Verification

Added orchestratord nightly tests for the `v1` opt-in flow, webhook certificate rotation (including CA rotation and CRD re-registration), manual promotion, and CRD version checks asserting that `v1` and the conversion webhook are installed if and only if `installV1CRD` is set. Verified the existing `v1alpha1` tests still pass.

Deployed https://github.com/MaterializeInc/cloud/pull/12275 in a personal stack to validate that this works for SaaS at `v1alpha1`.

### DIFF
--- a/misc/helm-charts/operator/templates/certificate.yaml
+++ b/misc/helm-charts/operator/templates/certificate.yaml
@@ -1,0 +1,41 @@
+# Copyright Materialize, Inc. and contributors. All rights reserved.
+#
+# Use of this software is governed by the Business Source License
+# included in the LICENSE file at the root of this repository.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0.
+
+{{- if eq .Values.operator.certificate.source "cert-manager" -}}
+---
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: {{ include "materialize-operator.fullname" . }}-self-signed
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "materialize-operator.labels" . | nindent 4 }}
+spec:
+  selfSigned: {}
+
+---
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: {{ include "materialize-operator.fullname" . }}-self-signed
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "materialize-operator.labels" . | nindent 4 }}
+spec:
+  dnsNames:
+    - {{ include "materialize-operator.fullname" . }}.{{ .Release.Namespace }}.svc
+  secretName: {{ include "materialize-operator.fullname" . }}-cert
+  privateKey:
+    algorithm: ECDSA
+    rotationPolicy: Always
+  issuerRef:
+    name: {{ include "materialize-operator.fullname" . }}-self-signed
+    kind: Issuer
+    group: cert-manager.io
+{{- end -}}
--- a/misc/helm-charts/operator/templates/certificate.yaml
+++ b/misc/helm-charts/operator/templates/certificate.yaml
@@ -7,7 +7,7 @@
 # the Business Source License, use of this software will be governed
 # by the Apache License, Version 2.0.
 
-{{- if eq .Values.operator.certificate.source "cert-manager" -}}
+{{- if and .Values.operator.args.installV1CRD (eq .Values.operator.certificate.source "cert-manager") -}}
 # We provision the webhook serving certificate from a stable root CA rather
 # than as a bare self-signed certificate. The serving certificate rotates
 # frequently, but it is always signed by the same long-lived CA, so the
--- a/misc/helm-charts/operator/templates/certificate.yaml
+++ b/misc/helm-charts/operator/templates/certificate.yaml
@@ -8,7 +8,15 @@
 # by the Apache License, Version 2.0.
 
 {{- if eq .Values.operator.certificate.source "cert-manager" -}}
+# We provision the webhook serving certificate from a stable root CA rather
+# than as a bare self-signed certificate. The serving certificate rotates
+# frequently, but it is always signed by the same long-lived CA, so the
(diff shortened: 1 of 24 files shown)


src/cloud-resources/src/crd/materialize.rs
```diff
@@ -653,6 +657,792 @@ pub mod v1alpha1 {
...
+    impl From<v1alpha2::Materialize> for Materialize {
+        fn from(value: v1alpha2::Materialize) -> Self {
+            let rollout_hash = value.generate_rollout_hash();
+            // Derive a deterministic UUID from the rollout hash so that the
+            // same v1alpha2 spec always produces the same requestRollout,
+            // making re-applies of an unchanged spec idempotent.
+            let request_rollout = Uuid::new_v5(&Uuid::NAMESPACE_OID, rollout_hash.as_bytes());
+            Materialize {
+                metadata: value.metadata,
+                spec: MaterializeSpec {
```
Alphadelta14, line 813: this looks very challenging to maintain. is there no way to copy value.spec and just update single fields?

src/cloud-resources/src/crd/materialize.rs
```diff
@@ -653,6 +657,792 @@ pub mod v1alpha1 {
...
+                    environmentd_image_ref: value.spec.environmentd_image_ref,
+                    environmentd_extra_args: value.spec.environmentd_extra_args,
+                    environmentd_extra_env: value.spec.environmentd_extra_env,
+                    environmentd_iam_role_arn: None,
```
Alphadelta14, line 817: why is this field not copied?

Alphadelta14 (approved):
Generally okay with it, but lacking a bit of confidence given the size of the changeset and how direct the migration process is.

src/cloud-resources/src/crd/materialize.rs
alex-hunt-materialize: That field is deprecated in v1alpha1 and doesn't exist in v1alpha2. When converting to v1alpha2, we move it to a value in the service_account_annotations field. When converting to v1alpha1, it is already in the service_account_annotations field, so we have nothing to do.

src/cloud-resources/src/crd/materialize.rs
alex-hunt-materialize: The specs are different types with different fields. AFAIK, there is no way to copy between different structs other than field by field.

Maintenance on this shouldn't be too bad. Only a few have special handling, and probably won't be changing the CRD in backward incompatible ways like this very often.

test/orchestratord/mzcompose.py
```diff
@@ -1605,6 +1606,200 @@ def validate(self, mods: dict[type[Modification], Any]) -> None:
...
+            ), f"Expected certificate secret name '{expected}', got '{secret_name}'"
+
+        retry(check, 120)
+
+
+def operator_supports_v1alpha2(definition: dict[str, Any]):
+    operator_version = Version.parse(
+        definition["operator"]["operator"]["image"]["tag"].removeprefix("v")
+    )
+    if operator_version >= Version.parse("26.17.0-dev.0"):
```
def-: Have to hurry with merging this, otherwise it will become 26.18 ;)
Why don't we just use MzVersion.parse?

def- (approved):
Tests lgtm, just a minor comment

test/orchestratord/mzcompose.py
alex-hunt-materialize: Bumped to 26.18 and now using MzVersion.parse.

doy-materialize (approved):

misc/helm-charts/operator/templates/certificate.yaml
```diff
@@ -0,0 +1,41 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "materialize-operator.labels" . | nindent 4 }}
+spec:
+  dnsNames:
+    - {{ include "materialize-operator.fullname" . }}.{{ .Release.Namespace }}.svc
+  secretName: {{ include "materialize-operator.fullname" . }}-cert
+  privateKey:
+    algorithm: ECDSA
+    rotationPolicy: Always
```
def-, line 36: When we rotate the cert, how do we distribute the new cert? Can we test somehow that this doesn't break everything entirely at that point? Maybe we can try deploying and test it with a very short rotation period?

doy-materialize (approved):

PR #35436, opened 2026-03-11
coord: bypass catalog_transact_inner for storage usage updates

aljoscha:
Rewrite storage_usage_update to bypass catalog_transact_inner entirely.
Instead of creating N Op::WeirdStorageUsageUpdates ops and routing them
through the catalog transaction machinery (where the N-iteration op loop
dominates at 91% of total time), we now:

1. Get a write timestamp from the oracle (~2ms)
2. Bump the durable storage-usage id allocator once for this batch
   (~10ms, one persist round-trip)
3. Pack N BuiltinTableUpdate rows directly via pack_storage_usage_update
4. Submit via builtin_table_update().execute() (group commit, ~18ms)

Every row in one collection cycle now shares the same id. The previous
per-row id was arguably the wrong semantics in the first place: a unique
id per row tells you nothing useful on its own, whereas a per-batch id
lets a consumer reconstruct which rows were collected together — which
is what's actually worth recording if anyone ever cares about the
column.

No consumer depends on the old per-row uniqueness today, so the change
is safe:
- mz_storage_usage joins on (shard_id, collection_timestamp), ignores id
- mz_recent_storage_usage same — joins on shard_id and
  collection_timestamp
- No other view or query references the id column
- VersionedStorageUsage::sortable_id() exists but is never called (it
  was modeled after audit log events, where the ID is actually used for
  ordering/dedup during catalog open)

One bump per batch (rather than per row) gives us monotonic batch ids
while keeping the coord-blocking cost proportional to a single
round-trip, no matter how many shards we collect. The data is
append-only during normal operation; retractions only happen at startup
during pruning, which already bypasses catalog_transact_inner.

In read-only mode (e.g. a standby coordinator during a zero-downtime
deployment) we can't write the allocator bump or append to
mz_storage_usage_by_shard, and there's no point running the slow shard
scan on a process that won't record the results. storage_usage_fetch
now early-returns in read-only mode, logs an info message, and
reschedules — so the cycle stays alive and we resume collection
automatically when the coord transitions out of read-only.

Results at 10k shards (optimized build)*:
- Before: ~499ms per collection cycle
- After:  ~20ms per collection cycle
- Speedup: 25x

* Numbers from a prototype that swapped the allocator for an in-memory
  counter (zero persist writes). The final version adds one durable
  allocator bump per cycle (~6–10ms persist round-trip per the cost
  table below), so expect ~25–30ms instead of ~20ms — still a ~17x
  speedup over baseline.

We remove Op::WeirdStorageUsageUpdates, its transact_op match arm, the
weird_builtin_table_update return path on transact_op, the diagnostic
info! log from storage_usage_update, and the now-unused
Transaction::allocate_storage_usage_ids method. We add
Catalog::allocate_storage_usage_id(commit_ts), a thin wrapper around
DurableCatalogState::allocate_id that bumps the allocator without going
through the high-level catalog transaction machinery.

### What happens today (before this fix)

Every collection interval, the coordinator processes
Message::StorageUsageUpdate on the main thread (message_handler.rs).
This calls storage_usage_update, which:

1. Fetches a write timestamp from the oracle (get_local_write_ts) — one
   persist round-trip
2. Creates one Op::WeirdStorageUsageUpdates per shard (N ops for N shards)
3. Calls catalog_transact_inner (ddl.rs), which:
   - Fetches another write timestamp from the oracle (get_local_write_ts)
     — a second persist round-trip
   - Opens a durable catalog Transaction
   - Loops N times through transact_op, each calling
     tx.allocate_storage_usage_ids() which bumps the in-memory id
     allocator counter in the transaction
   - Each op returns its row as a weird_builtin_table_update. The
     id_allocator mutation is explicitly excluded from get_op_updates()
     (transaction.rs), so get_and_commit_op_updates() returns empty — no
     per-op apply_updates calls happen on the COW catalog state
   - Persist write: tx.commit() consolidates all transaction tables and
     calls commit_transaction (compare_and_append) — this durably writes
     the id allocator bump
   - Persist read: sync_updates after commit
   - builtin_table_update().execute() submits rows to group commit and
     waits

The shard-size fetch itself runs off-thread and is not the problem.

### Why does it go through catalog_transact_inner at all?

Only to increment the durable STORAGE_USAGE_ID_ALLOC_KEY allocator,
which populates the id column of mz_storage_usage_by_shard. But a
unique id per row isn't a useful semantics in the first place — what's
worth recording is which rows came from the same collection cycle, and
one bump per batch gives us exactly that without the per-row catalog
transaction overhead.

In the past, we might have thought that going through a catalog
transaction protects us from concurrent writers, but that is a fallacy
so it's nice to also get rid of that false sense of security.

### Measured cost breakdown (10k shards, optimized build)

| Phase                                    | Time    | % of total |
|------------------------------------------|---------|------------|
| Oracle call #1 (storage_usage_update)    | ~2ms    | 0.4%       |
| Open transaction                         | ~10ms   | 2%         |
| transact_inner (10k op loop)             | ~430ms  | 91%        |
| Persist commit (write + read)            | ~6ms    | 1.3%       |
| Group commit (builtin_table_update)      | ~15ms   | 3%         |
| Total                                    | ~475ms  | 100%       |

The op loop dominates — not persist I/O or oracle calls as initially
suspected. Each of the 10k iterations allocates a storage usage id,
packs a row, and does catalog bookkeeping (Cow clone checks,
get_and_commit_op_updates, etc.). The per-iteration cost is ~43µs, which
is individually cheap but adds up across thousands of shards.

### Scaling behavior

Coordinator stall scales linearly with shard count:

| Shards  | Avg coordinator stall per cycle |
|---------|---------------------------------|
| ~2,400  | ~51ms                           |
| ~5,100  | ~150ms                          |
| ~10,100 | ~499ms                          |

### Fix

Stop routing storage usage updates through catalog_transact_inner. The
new storage_usage_update does:

1. Get a write timestamp from the oracle (unchanged)
2. Bump the durable id allocator once for this batch via the thin
   DurableCatalogState::allocate_id path (no high-level catalog
   transaction)
3. Pack N BuiltinTableUpdate rows directly using pack_storage_usage_update
4. Submit via builtin_table_update().execute() — the same group commit
   path that pruning already uses

This eliminates:
- N-1 of N allocator bumps (now one per batch instead of one per row)
- The redundant oracle timestamp (the old path fetched it twice)
- The N-iteration transact_op loop
- Opening a high-level catalog transaction

The coord-blocking work becomes: get one write timestamp (~2ms), bump
the allocator (~10ms), pack N rows (< 1ms), submit to group commit
(~18ms).


src/adapter/src/catalog/transact.rs
```diff
@@ -231,17 +230,6 @@ pub enum Op {
         name: String,
     },
     ResetAllSystemConfiguration,
-    /// Performs updates to the storage usage table, which probably should be a builtin source.
-    ///
-    /// TODO(jkosh44) In a multi-writer or high availability catalog world, this
-    /// might not work. If a process crashes after collecting storage usage events
-    /// but before updating the builtin table, then another listening catalog
-    /// will never know to update the builtin table.
-    WeirdStorageUsageUpdates {
```
mtabebe, line 254: Yay we can get rid of this!

src/adapter/src/coord/message_handler.rs
```diff
@@ -241,45 +243,49 @@ impl Coordinator {
...
+        let collection_timestamp: EpochMillis = if self.controller.read_only() {
             self.peek_local_write_ts().await.into()
         } else {
             // Getting a write timestamp bumps the write timestamp in the
             // oracle, which we're not allowed in read-only mode.
             self.get_local_write_ts().await.timestamp.into()
         };
 
-        let ops = shards_usage
+        // Pack builtin table rows directly and submit via group commit.
```
mtabebe: 👍

mtabebe (approved):
> This is safe because the id column of mz_storage_usage_by_shard is unused.

I don't love relying on this property. What if someone does use it? Can we rename the id column to be shard_id or something?

Anyhow the actual change looks good to me, good find

aljoscha:
> > This is safe because the id column of mz_storage_usage_by_shard is unused.
> 
> I don't love relying on this property. What if someone does use it? Can we rename the id column to be shard_id or something?
> 
> Anyhow the actual change looks good to me, good find

I updated the code and the PR message here. The new framing is that we now do one id allocation per batch, and that this makes more sense to begin with compared to the original one-id-per-row semantics. What do you think @mtabebe?

mtabebe:
Yes that makes sense to me and the code looked good. My approval stands 👍

PR #35441, opened 2026-03-11
sql: Add design doc for addressing the "optimizer/customer trade-off problem"

mgree:
When we make changes to the optimizer, we hope that everyone will have a good time (more freshness, less memory usage, etc.). But not every customer's workload is the same, and there's the potential for a customer to have a bad time.

It would be good to ensure that people can continue to have the kind of time they're having, even as we make changes to the optimizer.

This design doc ([rendered](https://github.com/mgree/materialize/blob/design-doc-optimizer-versioning/doc/developer/design/20260311_optimizer_customer_tradeoff.md)) explores the space of solutions, and proposes a particular one: plan pinning by way of freezing clusters.

### DIFF
--- a/doc/developer/design/20260311_optimizer_customer_tradeoff.md
+++ b/doc/developer/design/20260311_optimizer_customer_tradeoff.md
@@ -199,6 +199,9 @@ A proactive, Mztrail-like thing would help as well.)
 We would want to make `mz-debug` aware of frozen clusters.
 Some kind of `COPY CLUSTER` comand may help users experiment: copy a frozen cluster, unfreeze it, find a new plan you like, freeze _that_, and then green/blue the new one into place.
 
+When using DBT, users should receive an error when attempting to make changes to a frozen cluster.
+Whether that error is soft---i.e., the DBT run continues---or not is less clear.
+
 ### Why at the cluster level?
 
 We propose freezing at the cluster level.
@@ -217,6 +220,11 @@ Worse still, several of these `*Func` types reference external types, like `rege
 We propose the following shift:
 
  - Generate `Lir` versions of the various `*Func`s. This will require a preparatory PR to improve the associated macros, but should not be too complex (even though it generates a fair bit of code; we can engineer that code to point to common implementations of `eval`, so we won't get major code duplication). We will likely want to fold this in to work that parameterizes type holding `*Func`s to hold _either_ `Sql*Type` or `Repr*Type`.
+   + Some of the `TableFunc`s hold on to types for typechecking, but don't need those types for evaluation. We could leave these and `AggregateFunc`s alone.
+   + Many of the other `*Func`s have functions that use their stored  types in their `eval` method... they _must_ hold `Sql` types.
+     * `UnaryFunc`: `CastArrayToString`, `CastListToString`, `CastRecordToString`, `CastMapToString`, `CastStringToList`, `CastStringToMap`, `MapBuildFromRecordList`
+     * `VariadicFunc`: `ArrayCreate`, `ArrayToString`, `CaseLiteral`
+
  - Write `LirScalarExpr`, which omits `CallUnmaterializable` (because it should be resolved before LIR).
  - Ensure that we only serialize the MZ-controlled bits. Wherever `Lir*Func` would include an external structure, ensure we store the information that lets us regenerate that structure using `#[serde(default=...)]` (e.g., store the original regex string, not the `regex::Regex` value).
  - Add a version number to top-level `LIR`.
--- a/doc/developer/design/20260311_optimizer_customer_tradeoff.md
+++ b/doc/developer/design/20260311_optimizer_customer_tradeoff.md
@@ -37,8 +37,8 @@ Moreover, it's not clear how to use mztrail in a self-managed context.
 There are two closely related but not identical problems:
  1. **`our-bad`** MZ optimizer changed and it broke on redeploy.
  2. **`your-bad`** You changed something and it broke in staging.
-We are addressing the `our-bad` case exclusively.
-It is very important that we solve the "optimizer image" problem (you should be able to write SQL to get the good dataflow) and the "optimizer discontinuity" problem (you should be able to make small changes and not experience discontinuous performance).
+We are addressing the **`our-bad`** case exclusively.
+It is very important that we solve the "optimizer image" problem (you should be able to write SQL to get the good dataflow) and the "optimizer discontinuity" problem (you should be able to make small changes and not experience discontinuous performance, part of **`your-bad`**)---at some point, but not with this.
 
 
 ## Weighing Alternatives
@@ -65,13 +65,10 @@ Cons:
   - Code duplication. (Somewhat mitigated by `git subtree`.)
   - We do not know what kind of support window we will want, and may get backed into things we end up disliking.
   - Coarse-grained offramp: you can change versions, but that's it.
-  - Coarse-grained application: regressions are typically local, even
-    within a customer. So optimizer versions may not cut it fine enough---it may be just one query on the cluster that needs a different optimizer.
+  - Coarse-grained application: regressions are typically local, even within a customer. So optimizer versions may not cut it fine enough---it may be just one query on the cluster that needs a different optimizer.
   - Engineering burden of refactor.
   - Engineering burden of backporting.
-  - Punts on but doesn't resolve issues with release qualification.
-
-NB
+  - Punts on release qualification.
 
 ### `feature-flags`
 
@@ -102,6 +99,7 @@ Cons:
   - LIR is a not currently stored anywhere (but is a stable interface between MIR and rendering). DDIR does not actually exist.
   - Once we are committed, may be hard to back out of. (Mitigation: deploy this is as an unstable feature with a customer partner.)
   - More durable state.
+  - Need to manage migrations for LIR. (Mitigation: best effort.)
 
 ### `query-hints`
 
@@ -123,7 +121,7 @@ Cons:
 ## Solution Proposal
 
 We propose using **`plan-pinning`**.
-The ability to pin plans solves exactly the **`our-bad`** problem.
+The ability to pin plans _exactly_ solves the **`our-bad`** problem.
 It's also superior to the alternatives.
 We see it as superior to **`feature-flags`** because we can work more flexibly (change types!) with less uncertainy (known configs!).
 (Feature flags will of course continue to exist!)
@@ -134,7 +132,7 @@ Why have we changed our minds?
 
 `**optimizer-versions**` overfits to particular engineering challenges (wanting to make certain AST changes).
 But recent work on repr types has shown that we can change the tires while the car is moving---we simply have to be careful.
-It has a high engineering burden up front and promises a high maintenance burden in the future.
+Versioning the optimizer has a high engineering burden up front and promises a high maintenance burden in the future.
 Refactoring to have a clean optimizer crate is a good idea, but versioning is a heavyweight way to achieve what could be a lightweight goal.
 
 The balance tips further in **`plan-pinning`**'s favor when we consider that pinned plans are not merely a useful way for customers to have more confidence in Materialize, they are a way to help us identify clusters that are candidates for autoscaling and immediate incident escalation---production clusters.
@@ -151,18 +149,18 @@ ALTER CLUSTER foo UNFREEZE;
 ```
 
 We will store the LIR for all of the dataflows on `foo`, and automatically use those LIR plans on reboot.
+These plans will be stored in the catalog.
 No changes can be made to `foo`: no new dataflows, no removals.
 It will not be part of this work, but it seems sensible to limit other actions on frozen clusters, e.g., you many only run fast-path `SELECT`s and `SUBSCRIBE`s (with the possible exception of queries that touch introspection sources).
 
-### Why at the cluster level?
+### What is the SLA?
 
-We propose freezing at the cluster level.
-The environment and organization level is far too coarse.
-Going finer, the replica and dataflow levels are too fine---freezing these but not the rest of the cluster seems like a recipe for a mismatch with dependencies.
+Our pinning will start off as "best effort".
+At any point, we may simply throw up our hands and replan.
+Users should be notified if pinned plans are replanned, but it should not necessarily rise to the level of pinging an on-call engineer---say, an escalation rather than an incident.
+A possible success metric for plan pinning (beyond e.g., overall usage/number of pinned plans) is how _few_ replans are forced to occur.
 
-## Open questions
-
-### What _can_ change?
+### What can change?
 
 Suppose we have the following dependency diagram, where `S` means "source", `V` means "view", `MV` means "materialized view", and `C` means cluster:
 
@@ -187,49 +185,108 @@ What about `MV1` (which is read from persist)?
 What about `S1`?
 We don't need to fix opinions permanently on these questions up front, but we will need to _have_ opinions to start.
 
-One sensible possibility (**`frozen-deps`**) is that you can only freeze things whose dependencies are frozen.
-That is, if `C3` is frozen, then `C1` and `C2` must also be frozen.
-Changes to `V2` might be allowed, with a warning that these changes will not affect downstream frozen objects (which we could then name).
-In this world, we would likely want a `COPY CLUSTER foo TO bar` DDL command, that creates a copied cluster that could then be safely altered.
-This approach makes it hard to make changes to upstream clusters.
+As a first cut, it seems safe to say:
+ 1. Frozen clusters block `DROP ... CASCADE` and must be unfrozen first.
+ 2. Anything but `V2` may change in business logic; schema alterations have to be additive (cf. `ALTER MATERIALIZED VIEW`).
+That is, we would treat persist as a barrier: a frozen cluster will block changes to things it depends on that are not persisted.
+If `S1` or `MV1` change by altering a computation (but not its type) or by adding a new column, that should be fine (though we may need to generate an intermediate dataflow to project out the new column, since we will not want to change the pinned LIR plan of `MV2`).
 
-Another sensible possibility (**`mv-split`**) is that `MV` dependencies can be changed in schema-compatible ways---we'll read from persist anyway---but views like `V2` could not.
-This is more flexible than **`frozen-deps`**, but may be harder to explain to customers who don't fully grok the difference between materialized and unmaterialized views.
-Would source dependencies also be changeable in schema-compatible ways?
+It is important for us and for customers to know how things would be replanned if clusters were unfrozen.
+In the long term, we will want to know which plans are pinned and how far those pins have drifted from what we would output.
+It is merely engineering for us to know about drift in cloud---replan, emitting a diff of the new plan and the pinned one to store---but less so in self-managed.
+(These diffs will not address the question of "which MZ changes caused the plan changes", though, and we will only get information at release time.
+A proactive, Mztrail-like thing would help as well.)
+We would want to make `mz-debug` aware of frozen clusters.
+Some kind of `COPY CLUSTER` comand may help users experiment: copy a frozen cluster, unfreeze it, find a new plan you like, freeze _that_, and then green/blue the new one into place.
 
-What else might we do? What is the workflow for making changes? Something blue/green-ish or `ALTER MV`-ish seems appropriate.
+### Why at the cluster level?
 
-And finally: what is the MVP version?
-
-### How will users know what would be different?
-
-With a pinned plan, users might not know what would happen if they made changes.
-As users pin plans, we may not know how our changes would affect definitions on frozen clusters, since we won't be replanning.
-
-One sensible possibility is to always replan at a release (and save plans in, say, Snowflake), but to not deploy when clusters are frozen.
-This will not address the question of "which MZ changes caused the plan changes", though, and it will only give us information at release time.
-(Mztrail-like things would help.)
+We propose freezing at the cluster level.
+The environment and organization level is far too coarse.
+The replica and dataflow levels are too fine---freezing these but not the rest of the cluster seems like a recipe for confusion (two replicas on the same cluster with different plans? multiple versions of dependencies?).
 
 ### How does LIR change?
 
 LIR is the interface between the optimizer and rendering.
 While "stable", it's not persisted and has a purely internal contract.
 
-It has a large surface---`MirScalarExpr`, `AggregateFunc`, `TableFunc`, `Row`, etc.---and any changes across that surface could cause a pinned plan to no longer be runnable.
-What is the best way to write down LIR?
-We _had_ protobuf, but (a) we ripped it out and (b) it's migration story is simple but sometimes wrong.
-Something like `bincode` offers a straightforward serialization format, but no support for migration.
-An option (proposed by Claude during some research): `serde_reflection` with a schema registry for relevant types.
-We would:
+LIR has a large surface---`MirScalarExpr` (and with it, `UnaryFunc`, `BinaryFunc`, `VariadicFunc`, and `UnmaterializableFunc`), `AggregateExpr` and `AggregateFunc`, `TableFunc`, `Row`, etc.---and any changes across that surface could cause a pinned plan to no longer be runnable.
+It would be unwise to freeze things as they are in place: there is no support for migrations, and `MirScalarExpr` and the various `*Func` would be locked in time.
+Worse still, several of these `*Func` types reference external types, like `regex::Regex`---if we froze these things, upgrading the `regex` crate would break pinned plans and/or force a migration.
 
- - Use `serde_reflection` to build a `Registry` at when testing `compute-types` (where `plan.rs`/LIR lives).
- - Serialize the `Registry` to YAML/JSON and commit it as `schema.yaml`.
- - In CI, we regenerate `schema.yaml` from the current types and diff to see what changed.
- - We can identify changes that are permissible (e.g., addition of a field with default).
+We propose the following shift:
 
-When we want to make a breaking change, we will likely have to customize deserialization to support a migration.
-This will be painful the first time.
-It will be less painful if we bake in version numbers up front.
+ - Generate `Lir` versions of the various `*Func`s. This will require a preparatory PR to improve the associated macros, but should not be too complex (even though it generates a fair bit of code; we can engineer that code to point to common implementations of `eval`, so we won't get major code duplication). We will likely want to fold this in to work that parameterizes type holding `*Func`s to hold _either_ `Sql*Type` or `Repr*Type`.
+ - Write `LirScalarExpr`, which omits `CallUnmaterializable` (because it should be resolved before LIR).
+ - Ensure that we only serialize the MZ-controlled bits. Wherever `Lir*Func` would include an external structure, ensure we store the information that lets us regenerate that structure using `#[serde(default=...)]` (e.g., store the original regex string, not the `regex::Regex` value).
+ - Add a version number to top-level `LIR`.
+ - Use `serde_reflection` to generate a schema (using `Registry`) that lives in the repo. Any change in LIR serialization will yield a schema change. We can programmatically require migrations/LIR version bumps for certain kinds of schema changes.
+   + We _must_ eventually tolerate schema changes like new LIR operators, new `*Func`s, new fields on existing structures, or pins will break too often.
 
 LIR will be stored in the catalog.
+
+## Open questions
+
+### What is the concrete syntax?
+
+We should pick ergonomic, sensible DDL syntax for this.
+How will users interact with it in DBT, etc.?
+
+### How do alterations work?
+
+Consider the dependency diagram above (["What can change?"](#what-can-change)).
+What are the pragmatics of making changes to `S1` and `MV1`?
+These changes could be changes to business logic or the addition of a column.
+
+- What does the workflow look like for changes to business logic upstream? _Should be fine._
+- What about additive schema changes? _Should be fine._
+- What about changes to a hypothetical `MV3` what depends on `MV2`? _Should be fine._
+
+Suppose the logic in `MV2` is broken.
+What is the workflow for fixing that?
+_Any_ change to `MV2` will lead to replanning, and there is no possibility of "spot" fixes.
+How do we communicate to customers that pinning is "best effort"?
+
+### What is the serialization format for LIR?
+
+Based on ad hoc, not particularly well informed research:
+
+| Format      | Compactness | Evolution     |
+| :------     | :---------- | :------------ |
+| bincode     | +++         | none          |
+| JSON        | ---         | roll your own |
+| MessagePack | ++          | roll your own |
+| CBOR        | +           | roll your own |
+| protobuf    | ~           | theirs        |
+
+MessagePack seems like a strong possibility, but "roll your own" as a migration/evolution story is not my favorite.
+In addition to our schema registry management, we would need some kind of lint to ensure the serde annotations were correct for MessagePack (i.e., defaults are required on new struct fields, or we'll break at runtime when reading old data). Having some kind of CI test that reads a broad variety of pinned plans through an upgrade would help.
+
+Protobuf has a natural form of evolution, but it is not clear that it aligns with _ours_.
+We will likely end up maintaining protobuf wrappers by hand, which is a bad combination of "their weird defaults" and "roll your own".
+
+As an MVP, bincode is fine: we check the version number, and don't use pinned plans if anything changed. The downside is that improving that level of "best effort" to something better is moderately complex.
+
+Another alternative is explicitly versioned LIR enums, where we use version numbers to dispatch and manually write migrations between formats.
+This would allow us to use `bincode`, but would mean far more migrations: any change at all would force one.
+In this case, we would want good tooling to generate migrations and keep migration code to a minimum.
+
+### Where does LIR live in the catalog?
+
 It could go in `CatalogPlans` (all currently keyed by global ID), in `CatalogState` (alongside per-cluster metadata), or as a sidecar (like the `ExpressionCacheHandle`).
+This part of the catalog seems to be in flux, but `CatalogState` seems right---plans for frozen clusters should be stored alongside those clusters.
+That is, whether or not a cluster is frozen is not a boolean---it's an `Option<FrozenLirPlans>`, with `Some(...)` holding the plans.
+
+### What happens when we move to DDIR?
+
+A nice property of plan pinning is that we're always free to go lower.
+Currently, a pinned LIR plan will render directly.
+If we rearchitect things to convert that LIR plan to DDIR, that will be "transparent" to the pinned plan (if we do a good job).
+If we later decide to save the DDIR rather than LIR, that will be as transparent as the original move was.
+
+### What happens to `EXPLAIN OPTIMIZED PLAN` for pinned LIR plans?
+
+Right now, I believe only Gábor uses `EXPLAIN OPTIMIZED PLAN`.
+But in the event someone would like to see MIR for a pinner LIR plan, we would simply not have it.
+We _could_ store a cached version, either a text (with fixed options) or as some kind of structure---though our aim was to _not_ have to serialize MIR.
+I think the best approach here is to improve the default LIR-based `EXPLAIN PLAN` enough so that Gábor stops using `EXPLAIN OPTIMIZED PLAN`.
--- a/doc/developer/design/20260311_optimizer_customer_tradeoff.md
+++ b/doc/developer/design/20260311_optimizer_customer_tradeoff.md
@@ -1,4 +1,4 @@
-# Resolving customer tradeoffs with multiple optimizer crates
+# Resolving customer tradeoffs optimizer changes
 
 - Associated: [#30233 optimizer release engineering](https://github.com/MaterializeInc/materialize/pull/30233),
 [#8768 optimizer crate](https://github.com/MaterializeInc/database-issues/issues/8768)
@@ -86,7 +86,7 @@ Cons:
   - Difficult scaling granularity: not every feature is easy to flag. `**optimizer-versions**` is essentially a particular approach to `feature-flags`, where the flag granularity is "set of optimizer features and types."
   - Exponentially many configurations---we can't test every combination of flags, and flags interact.
   - Who flips the bits? If it's us: high support burden. If it's someone else: what if they break things?
-  - Unknown support windows.
+  - Unknown support windows, and we have not historically done a good job managing feature flags.
 
 ### `plan-pinning`
 
@@ -98,9 +98,10 @@ Pros:
 
 Cons:
 
-  - Any changes to the plan and you lose your pin. (Mitigation: use MVs to separate the units you care about.)
-  - LIR is a not a stable interface. DDIR does not actually exist.
+  - Any changes to the plan and you lose your pin. (Mitigation: use MVs on different clusters to separate the units you care about.)
+  - LIR is a not currently stored anywhere (but is a stable interface between MIR and rendering). DDIR does not actually exist.
   - Once we are committed, may be hard to back out of. (Mitigation: deploy this is as an unstable feature with a customer partner.)
+  - More durable state.
 
 ### `query-hints`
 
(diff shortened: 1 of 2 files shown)


doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,164 @@
...
+Pros:
+
+  + The status quo (less the tooling).
+  + Fine-grained control: you can offramp from old feature settings flag-by-flag. (In principle, at least.)
+  + Flexible: we can create new feature flags as we plase, and we do not need to fix their support windows in advance.
+
+Cons:
+
+  - Feature flags apply to entire environments, not individual clusters. If different teams need different flags, they are out of luck.
+  - Exponentially many configurations---we can't test every combination of flags, and flags interact.
```
mtabebe, line 84: I see this concern a lot with feature flags, and while it is true in theory, I'd love to know how worried I should be about it in practice. That is: if e.g., there are only 3 active optimizer flags at a time, I don't know that the exponential is so bad. That is in the realm of testable

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,164 @@
...
+  - Unknown support windows.
+
```
mtabebe, line 87: Isn't this missing: not everything is feature flaggable (or if it is, it might be VERY hard) - since the intro motivated this with the repr type work.

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+Pros:
+
+  + Ties in neatly with related ideas of "production clusters".
+  + Ties in neatly with related ideas of "DDIR" or some other stable, low-level interface.
+  + Offers the most reliable possible experience---a fixed LIR plan would be stable even if bugfixes in MIR cause queries to change.
+
+Cons:
+
+  - LIR is a not a stable interface. DDIR does not actually exist.
+  - No offramp.
```
mtabebe: What does this mean?

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+
+The `optimizer-types` crate will have traits and definitions that are global across all three live versions of the optimizer. The AST types may change version to version, but the traits will be more stable.
+
+These crates do _not_ include:
+  + the SQL parser
+  + (\*) SQL -> HIR in `sql`
+  + `sequence_*` methods in adapter
+  + (\*) LIR -> dataflow in `compute`
+
+The two bullets marked (\*) above are a tricky point in the interface. SQL will _always_ lower to the latest HIR; we will have methods to convert the latest HIR to the other two versions. Similarly, LIR lowers to dataflow from the latest LIR; we will have methods to convert the other two versions to the latest.
```
mtabebe: I think this means that feature flags still need to exist. Since not *all* optimizer features are versioned like this. 

So probably worth calling out that the design spectrum is not all or nothing

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+[The prior design doc in #30233](https://github.com/MaterializeInc/materialize/pull/30233) proposed a Debian-like versioning scheme, with an unstable/hot branch, a testing/warm branch, and stable/cold branch.
+This is overplanned: we should begin by having a stable branch and a 'future' branch, where we make more drastic changes.
+We can later adopt the Debian-like scheme, or the [fully versioned approach Spanner takes](https://docs.cloud.google.com/spanner/docs/query-optimizer/versions).
+
+The changes proposed below center around breaking off the optimizer as its own subsystem, with an `optimizer` crate (of which we may have multiple versions).
+Treating the optimizer as its own system comes with its own benefits:
+
+  - Better tracing in the optimizer, which currently hacks things to get the right outputs.
+  - The possibility of process isolation for the optimizer.
+  - Clearer interfaces and a better testing surface.
```
mtabebe: 👍

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,164 @@
...
+## The Problem
+
+Customers run operational workloads on Materialize.
+Changes to Materialize can threaten the stability of those workloads---particularly changes in the optimizer.
+
+To date, we have managed optimizer changes using feature flags (e.g., `enable_cast_elimination`, `enable_eager_delta_joins`).
+Not every feature can be feature flagged (e.g., changing `MirRelationExpr` to hold `Repr*` types), though, and we do not have much in the way of tooling for feature flags.
+
+It is hard for us to make changes in the optimizer that won't cause some customers to have a bad time---even if some customers have a much better time with those changes.
+We need a way to change the optimizer without disrupting customer workloads.
```
mtabebe, line 15: Maybe I am missing something, but this doc doesn't talk about how this will be achieved. Suppose we release and see a problem. Wouldn't that be in the "Stable" version, how do we flip back to the previous "stable" version?

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,164 @@
...
+
+## The Problem
+
+Customers run operational workloads on Materialize.
+Changes to Materialize can threaten the stability of those workloads---particularly changes in the optimizer.
+
+To date, we have managed optimizer changes using feature flags (e.g., `enable_cast_elimination`, `enable_eager_delta_joins`).
+Not every feature can be feature flagged (e.g., changing `MirRelationExpr` to hold `Repr*` types), though, and we do not have much in the way of tooling for feature flags.
+
+It is hard for us to make changes in the optimizer that won't cause some customers to have a bad time---even if some customers have a much better time with those changes.
```
mtabebe, line 14: This is an understandable problem, but to be clear, this doc doesn't describe how it is fixed right? There is no notion of e.g., ALTER CLUSTER SET optimizer_version = X.Y.Z

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+We propose using **`optimizer-versions`**.
+We see it as superior to **`feature-flags`** because we can work more flexibly (change types!) with less uncertainy (known configs!).
+We see it as superior to **`plan-pinning`** because we don't need to stabilize new interfaces.
+We see it as superior to **`query-hints`** because we don't want to add query hints.
+
+[The prior design doc in #30233](https://github.com/MaterializeInc/materialize/pull/30233) proposed a Debian-like versioning scheme, with an unstable/hot branch, a testing/warm branch, and stable/cold branch.
+This is overplanned: we should begin by having a stable branch and a 'future' branch, where we make more drastic changes.
+We can later adopt the Debian-like scheme, or the [fully versioned approach Spanner takes](https://docs.cloud.google.com/spanner/docs/query-optimizer/versions).
+
+The changes proposed below center around breaking off the optimizer as its own subsystem, with an `optimizer` crate (of which we may have multiple versions).
```
mtabebe: Maybe this is obvious to others, but how do you make sure that nothing the optimizer does can change e.g., persistence or is not forward/backward compatible?

How is that enforced?

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+
+Pros:
+
+  + The status quo (less the tooling).
+  + Fine-grained control: you can offramp from old feature settings flag-by-flag. (In principle, at least.)
+  + Flexible: we can create new feature flags as we plase, and we do not need to fix their support windows in advance.
+
+Cons:
+
+  - Feature flags apply to entire environments, not individual clusters. If different teams need different flags, they are out of luck.
```
ggevay: We do have per-cluster feature flags, see `cluster_features.slt`.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,164 @@
...
+  + Ties in neatly with related ideas of "production clusters".
+  + Ties in neatly with related ideas of "DDIR" or some other stable, low-level interface.
+  + Offers the most reliable possible experience---a fixed LIR plan would be stable even if bugfixes in MIR cause queries to change.
+
+Cons:
+
+  - LIR is a not a stable interface. DDIR does not actually exist.
+  - No offramp.
+
+### `query-hints`
```
ggevay, line 104: What does this exactly mean? That optimizer changes would mostly be accessible by giving hints, and thereby old queries' plans wouldn't change?

(Also, this alternative is not listed at the top of the section.)

ggevay, line 104: Adding support for query hints could be independently useful, regardless of whether we go with `optimizer-versions` or `feature-flags`. E.g., it will be very useful for broadcast joins.

doc/developer/design/20260311_optimizer_versioning.md
ggevay: For me, by far the biggest con of `plan-pinning` is that we lose the pinned plan when the user wants to redeploy a (possibly slightly changed) object. If it weren't for not this one big issue, I'd totally vote for `plan-pinning`.

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+  + Fixed, known configurations.
+  + Per-cluster control.
+  + Forces more unified optimizer interface.
+  + Ties in neatly with related ideas of "a separate optimizer process".
+  + Flexible versioning: we can cut new optimizer versions as we please, and do not need to fix a support window in advance.
+
+Cons:
+
+  - Code duplication. (Somewhat mitigated by `git subtree`.)
+  - Unknown support windows.
```
ggevay: Could you elaborate on this? Is this more a problem for us (in that we don't know how long we'll need to support some version) or for users (in that they wouldn't know how much time they have to migrate off of an old version)? Could we fix a versioning and support schedule in advance? Edit: Oh, I see below that it is explicitly part of the proposal to _not_ do that. Is it so to give us more flexibility on optimizer development?

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,164 @@
...
+
+Mztrail---testing on customer workloads---would help us predict when optimizer changes will affect customers.
+(It would also help the most proactive customers, who could run tests themselves.)
+While pushing in this direction is good, important work, it's a bigger bite than what's proposed here.
+Moreover, it's not clear how to use mztrail in a self-managed context.
+
+## Weighing Alternatives
+
+- **`optimizer-versions`** Separate optimizer versions, settable per-cluster using a system-level privilege.
+- **`feature-flags`** Feature flag everything, building tooling to support eng, field eng, and customers.
```
ggevay, line 47: If we have `optimizer-versions`, then would we totally stop creating feature flags for optimizer changes, and the "nightly" optimizer version would have all the half-baked stuff? Probably not, right?

ggevay: Btw., I could imagine a light-weight version of `optimizer-versions`, where different versions wouldn't be separate crates, but just a set of feature flag settings.

doc/developer/design/20260311_optimizer_versioning.md
ggevay: Also, "flexible" is listed as a pro at both `optimizer-versions` and `feature-flags`, but this might be more true for `feature-flags`, in that cutting a new optimizer version seems to be a much more heavy-weight operation that adding a flag, right?

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+  - The possibility of process isolation for the optimizer.
+  - Clearer interfaces and a better testing surface.
+
+## Minimal Viable Prototype
+
+This plan is a revised version of the plan from [the prior design doc in #30233](https://github.com/MaterializeInc/materialize/pull/30233).
+
+### Versioning the `optimizer` crate
+
+To create a new version of the `optimizer` crate, we will use `git subtree`, which should preserve features.
```
ggevay: Would each crate have its own version of all the slts as well? (But hopefully not all the tds, right?)

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+
+Pros:
+
+  + Ties in neatly with related ideas of "production clusters".
+  + Ties in neatly with related ideas of "DDIR" or some other stable, low-level interface.
+  + Offers the most reliable possible experience---a fixed LIR plan would be stable even if bugfixes in MIR cause queries to change.
+
+Cons:
+
+  - LIR is a not a stable interface. DDIR does not actually exist.
```
ggevay: But this is a problem for all other options as well, right? That is, if the behavior of an LIR operator changes, that can mess up things both for pinned plans and old optimizer versions and old feature flag settings.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,164 @@
...
+
+  + Fixed, known configurations.
+  + Per-cluster control.
+  + Forces more unified optimizer interface.
+  + Ties in neatly with related ideas of "a separate optimizer process".
+  + Flexible versioning: we can cut new optimizer versions as we please, and do not need to fix a support window in advance.
+
+Cons:
+
+  - Code duplication. (Somewhat mitigated by `git subtree`.)
```
ggevay, line 65: We might also need a policy for backporting bugfixes.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
ggevay: Well, a concern is that the temptation is too big to keep around old feature flags for a very long time, just for, say, 1 big customer. But I guess this danger exists also for `optimizer-versions`.

Anyhow, the solution might be to just be disciplined enough as an organization to be able to say no when a customer asks for keeping around very old configurations (be it old flag settings or old optimizer versions).

Btw. sunsetting old configurations can also be sometimes facilitated by throwing money at the problem: accepting that a specific user workload uses more resources, but adding a discount for that user.

doc/developer/design/20260311_optimizer_versioning.md
ggevay: ((I'm not sure, but my feeling is that DDIR would be less stable than LIR. I mean, LIR's exact shape doesn't matter that much, it's ugly as hell already, being just a big collection of special fields everywhere that do bespoke things. When evolving it, we can just add a new field that does a new thing.

In contrast, DDIR sounds like it is intended to be an "elegant" thing where people will have strong and potentially changing opinions on how it should look like.

Or is the hope here that DDIR would be much more minimalistic than LIR, and that will be the reason for why it should be easier to stabilize it?

Btw., would DDIR's optimization passes be part of the optimizer crate?))

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+  - Major AST overhaul.
+  - Major transform overhaul.
+  - All known forms of this are brittle.
+  - Hard to specify emergent properties (e.g., what to do with operators that do not syntactically appear in the query plan).
+
+## Solution Proposal
+
+We propose using **`optimizer-versions`**.
+We see it as superior to **`feature-flags`** because we can work more flexibly (change types!) with less uncertainy (known configs!).
+We see it as superior to **`plan-pinning`** because we don't need to stabilize new interfaces.
```
ggevay: I don't really understand this. Even `optimizer-versions` would need to target a stable target IR.

Edit: Or is the conversion between LIR versions is what would allow us to not need a stable LIR?

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+We see it as superior to **`query-hints`** because we don't want to add query hints.
+
+[The prior design doc in #30233](https://github.com/MaterializeInc/materialize/pull/30233) proposed a Debian-like versioning scheme, with an unstable/hot branch, a testing/warm branch, and stable/cold branch.
+This is overplanned: we should begin by having a stable branch and a 'future' branch, where we make more drastic changes.
```
ggevay: Hmm, I'm not sure if just a stable branch and a 'future' branch are enough.

I think we'd need at least 2 stable versions (+ 'future') at all times (I mean, as soon as enough time has passed since the initial introduction of the optimizer versions feature), because production users would want to move from stable version to stable version. That is, prod users might refuse to ever move to 'future', because anything can happen at roughly any moment on 'future'.

And even with 2 stable versions, users would get into a bit of a pickle when there is a catastrophic plan resource usage regression when moving from stable to stable. With 3 stable versions they could skip a stable version. So then the workflow when seeing a catastrophic plan regression would be: tell us about the issue, we fix it in the next stable version, and then the user skips the bad stable version. Or we could make just 2 stable versions work by somehow forcing users to try the current 'future' _before_ cutting a stable version? Am I overthinking this?

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+
+At first, there will be two versions: V0 (the current optimizer) and V1 (the future branch).
+
+### The `optimizer` and `optimizer-types` crates
+
+The `optimizer` crate will contain the definitions of HIR, MIR, and LIR, along with the HIR-to-MIR and MIR-to_LIR lowerings and the MIR-to-MIR transformations. These come from the `sql` (HIR, HIR -> MIR), `expr` (MIR), `transform` (MIR -> MIR), and `compute-types` (LIR, MIR -> LIR) crates.
+
+The `optimizer-types` crate will have traits and definitions that are global across all three live versions of the optimizer. The AST types may change version to version, but the traits will be more stable.
+
+These crates do _not_ include:
```
ggevay: (This would leave out the transformations implemented in `transform_ast.rs`. This might be ok.)

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Versioning the `optimizer` crate
+
+To create a new version of the `optimizer` crate, we will use `git subtree`, which should preserve features.
+
+At first, there will be two versions: V0 (the current optimizer) and V1 (the future branch).
+
+### The `optimizer` and `optimizer-types` crates
+
+The `optimizer` crate will contain the definitions of HIR, MIR, and LIR, along with the HIR-to-MIR and MIR-to_LIR lowerings and the MIR-to-MIR transformations. These come from the `sql` (HIR, HIR -> MIR), `expr` (MIR), `transform` (MIR -> MIR), and `compute-types` (LIR, MIR -> LIR) crates.
```
ggevay: Uh, that is a lot of code to duplicate.

Also, there are a lot of crates that are _in-between_ these crates in the dependency graph. What happens with those? (e.g. `mz-catalog`)

Edit: Hmm, is the trait thing what is meant to enable the refactoring of all those in-between crates to be able to snatch them out from in-between the above crates?

doc/developer/design/20260311_optimizer_versioning.md
```diff
@@ -0,0 +1,164 @@
...
+
+The `optimizer-types` crate will have traits and definitions that are global across all three live versions of the optimizer. The AST types may change version to version, but the traits will be more stable.
```
ggevay: Btw. what about type types? Repr types in `optimizer` and sql types in `optimizer-types`?

ggevay: Actually, I realized I don't really understand what exactly the traits will be doing. Below you wrote that "SQL will _always_ lower to the latest HIR" and "LIR lowers to dataflow from the latest LIR". Than why do we need to abstract away optimizer concepts behind traits if optimizer-external code will just refer to the latest concrete thing?

doc/developer/design/20260311_optimizer_versioning.md
ggevay: (and the HIR-to-HIR transformations)

ggevay (commented):
I have a lot of questions

doc/developer/design/20260311_optimizer_customer_tradeoff.md
mtabebe: Right, I guess my point is that optimizer versions looks like another dimension that we will have. When will we retire optimizer versions? What happesn if a customer is stuck on one forever?

It is basically a static set of flags fixed in a point in time. With the added bonus that we can do things that are not feature flaggable.

So I am really struggling with investing in this as opposed to e.g., plan pinning or retiring flags faster.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,164 @@
...
+It would be good to link this work to the "representative workloads".
+It should be easy (for us) to test the same workload with different versions of the optimizer.
+We should use some kind of representative workloads to qualify making the future branch optimizer public.
+
+### Supporting observability
+
+Observability is how people---us and customers---know that something is working (or not).
+It should be easy (for us and for customers) to see what might change in existing plans if the optimizer were to switch.
+
+## Open questions
```
maheshwarip: How would customers specify the optimzier version? Is it possible for it to be at the query level?

maheshwarip: Catching up from an in person conversation - maybe going at the cluster level is the right move

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Minimal Viable Prototype
+
+We will pin plans at the level of clusters.
+
+Two new DDL commands:
+
+```sql
+ALTER CLUSTER foo FREEZE;
+ALTER CLUSTER foo UNFREEZE;
```
maheshwarip, line 148: So this means to pin a plan, I freeze the cluster? As soon as I unfreeze, the plan immediately goes to the latest LIR plan?

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+We will store the LIR for all of the dataflows on `foo`, and automatically use those LIR plans on reboot.
+No changes can be made to `foo`: no new dataflows, no removals.
+It will not be part of this work, but it seems sensible to limit other actions on frozen clusters, e.g., you many only run fast-path `SELECT`s and `SUBSCRIBE`s (with the possible exception of queries that touch introspection sources).
+
+### Why at the cluster level?
```
mtabebe: 👍

mtabebe (commented):
I think this is a much stronger document and proposal. Nice work

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+
+  + Fixed, known configurations.
+  + Per-cluster control.
+  + Forces more unified optimizer interface.
+  + Ties in neatly with related ideas of "a separate optimizer process".
+  + Moderately flexible versioning: we can cut new optimizer versions as we please, and do not need to fix a support window in advance.
+
+Cons:
+
+  - Code duplication. (Somewhat mitigated by `git subtree`.)
```
antiguru, line 65: Please do not use subtrees, it's too much of a headache.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+  - We do not know what kind of support window we will want, and may get backed into things we end up disliking.
```
antiguru, line 66: That's a product question, not an engineering problem I think!

doc/developer/design/20260311_optimizer_customer_tradeoff.md
antiguru: Experience tells me it's an ever-growing list of feature flags. Cleaning up old feature flags is tedious and doesn't happen right away, so I'd say we should engineer for "large number of feature flags" rather than "low number".

doc/developer/design/20260311_optimizer_versioning.md
antiguru: I think we're treating LIR as the stable boundary between optimizer and rendering. It can change, but requires adjustment in all optimizer versions. Changes to rendering are not part of this design doc (but essentially suffer from the same problems explained here.)

antiguru: Well, DDIR doesn't exist, so hard to say, but I think this:
> When evolving it, we can just add a new field that does a new thing.

is potentially dangerous: LIR should have less special-casing in the future, not more. The fact that it has special-casing doesn't justify piling more onto it, but should rather motivate figuring out what's wrong.

Overall, I'd say this design should stop at LIR, especially given we're not going to invest in a DDIR right now.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+
+### `plan-pinning`
+
+Pros:
+
+  + Ties in neatly with related ideas of "production clusters", guarantees, and auto-scaling.
+  + Ties in neatly with related ideas of "DDIR" or some other stable, low-level interface.
+  + Offers the most reliable possible experience---a fixed LIR plan would be stable even if bugfixes in MIR cause queries to change.
+
+Cons:
```
antiguru, line 96: Another cons: More durable state.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+  - Once we are committed, may be hard to back out of. (Mitigation: deploy this is as an unstable feature with a customer partner.)
+
+### `query-hints`
+
+Pros:
+
+  + The finest-grained control.
+  + Avoids/defers the need to have smart query planning.
+
+Cons:
```
antiguru, line 111: Another cons: It feels like a trap-door, undoing hints is hard once people adopt it.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+    MV2
+  end
+
+  S1 --> MV1 --> V2 --> MV2
+```
+
+If we freeze `C3`, we certainly can't make changes to `C3`.
+What about `V2` (which is inlined into the definition of `MV2`)?
+What about `MV1` (which is read from persist)?
+What about `S1`?
```
ggevay, line 185: I'd say the only optimizer-relevant question here is a potential change in `V2`. This is because changes in `MV1` and `S1` would change the persist-level input of `MV2`, so no direct effect on the plan of `MV2`. (And avoiding downstream surprises is an active topic of discussion in the `ALTER MV` workstream.)

For `V2`, actually, our current `ALTER VIEW` can't change the view's query (can only change the name and owner), so we don't have an issue with this for now.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+  + Flexible: we can create new feature flags as we plase, and we do not need to fix their support windows in advance.
+
+Cons:
+
+  - Difficult scaling granularity: not every feature is easy to flag. `**optimizer-versions**` is essentially a particular approach to `feature-flags`, where the flag granularity is "set of optimizer features and types."
+  - Exponentially many configurations---we can't test every combination of flags, and flags interact.
+  - Who flips the bits? If it's us: high support burden. If it's someone else: what if they break things?
+  - Unknown support windows.
+
+### `plan-pinning`
```
ggevay, line 88: A thing that is unclear to me with `plan-pinning` is how will users know when they'll have to migrate away from an old pinned plan? With `optimizer-versions`, this is more clear: We can warn them some time before the scheduled removal of an old optimizer version, and it should be easy even for self-managed users to see whether they'll be affected or not. But with `plan-pinning`, a breaking change that would make a pinned plan impossible to render anymore would not have an obvious warning before it.

To solve this, we might want to explicitly add some code some time before making a breaking change to just hunt down pinned plans that would be affected, but not break them yet, and warn the user loudly that they should try a migration away from the pinned plan, with also telling them when the actual breaking change is scheduled.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+
+Pros:
+
+  + Ties in neatly with related ideas of "production clusters", guarantees, and auto-scaling.
```
ggevay, line 92: Yes, I feel more and more momentum building lately behind the "production clusters" idea.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+  + Ties in neatly with related ideas of "DDIR" or some other stable, low-level interface.
+  + Offers the most reliable possible experience---a fixed LIR plan would be stable even if bugfixes in MIR cause queries to change.
```
ggevay, line 94: Well, `optimizer-versions` would be even more reliable, in that it would also solve the query-tweaking problem, i.e., when a tiny query tweak does a big plan change _due to the tweaked query being planned with different optimizer code_.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+## Out of Scope
+
+Mztrail---testing on customer workloads---would help us predict when optimizer changes will affect customers.
+(It would also help the most proactive customers, who could run tests themselves.)
+While pushing in this direction is good, important work, it's a bigger bite than what's proposed here.
+Moreover, it's not clear how to use mztrail in a self-managed context.
+
+There are two closely related but not identical problems:
+ 1. **`our-bad`** MZ optimizer changed and it broke on redeploy.
+ 2. **`your-bad`** You changed something and it broke in staging.
```
ggevay, line 39: Well, it's kinda debatable that when a user tweaks a query, and it has a big plan change not because of the change being big at the SQL level, or because of a discontinuity in optimizer code, but _because of different optimizer code running_, then is it our fault (optimizer code change) or their fault (query change).

But the "Mitigation: use MVs to separate the units you care about." might mitigate this problem enough.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
ggevay: > As soon as I unfreeze, the plan immediately goes to the latest LIR plan?

I'm not sure, but I would imagine that the typical procedure for moving away from a pinned plan would be to do either a blue-green deploy, or an `ALTER MV`-like workflow, so that you can see whether the new plan would work well. We should think more about the details of this, e.g., how it would look like when the "`ALTER MV`-like workflow" for moving to a newer plan would need to happen on a cluster level.

ggevay (commented):
I'm liking this new version quite a bit! My biggest concern with plan pinning was

> Any changes to the plan and you lose your pin.

(I'd formulate it as "Any changes to the _query_ and you lose your pin.")

But

> Mitigation: use MVs to separate the units you care about.

might be enough of a mitigation.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+  - Major parser overhaul.
+  - Major AST overhaul.
+  - Major transform overhaul.
+  - All known forms of this are brittle.
+  - Hard to specify emergent properties (e.g., what to do with operators that do not syntactically appear in the query plan).
+  - Devolves to plan pinning.
+
+## Solution Proposal
+
+We propose using **`plan-pinning`**.
```
antiguru, line 123: I think we shouldn't underestimate the burden to commit to a durable LIR format. What's the process of evolving it? How do we detect changes? How do we prevent silent incompatibilities?

LIR depends on large parts of Mz's implementation, relation expressions, scalars, etc. and I'm not sure all parts have committed to a stable representation.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+We will pin plans at the level of clusters.
+
+Two new DDL commands:
+
+```sql
+ALTER CLUSTER foo FREEZE;
+ALTER CLUSTER foo UNFREEZE;
+```
+
+We will store the LIR for all of the dataflows on `foo`, and automatically use those LIR plans on reboot.
```
antiguru, line 151: Where will you store the LIR?

antiguru (commented):
Leaving some comments. I'm not yet convinced that durably recording LIR is a good idea!

doc/developer/design/20260311_optimizer_customer_tradeoff.md
maheshwarip: There's a similar question here about what happens during a blue/green dbt deploy. We spin up & spin down new clusters during a blue green - which means that your plan pinning would disappear! 

Is there a way for us to specify plan version explicitly? Something like `CREATE CLUSTER WITH OPTIMIZER PLAN v1` or something like that?

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,190 @@
...
+No changes can be made to `foo`: no new dataflows, no removals.
+It will not be part of this work, but it seems sensible to limit other actions on frozen clusters, e.g., you many only run fast-path `SELECT`s and `SUBSCRIBE`s (with the possible exception of queries that touch introspection sources).
+
+### Why at the cluster level?
+
+We propose freezing at the cluster level.
+The environment and organization level is far too coarse.
+Going finer, the replica and dataflow levels are too fine---freezing these but not the rest of the cluster seems like a recipe for a mismatch with dependencies.
+
+## Open questions
```
ggevay: One more open question for plan-pinning: What do we do with non-LIR EXPLAINs (e.g., EXPLAIN OPTIMIZED PLAN FOR MATERIALIZED VIEW ...) for objects that only have a pinned plan? If we really only save the LIR plan, then we'll lose the ability to do HIR/MIR EXPLAINs on these objects. Maybe we could additionally best-effort save the MIR/HIR plans as well, where by "best effort" I mean just discarding them when there is a deserialization failure, so that at least most of the time we retain the ability to show HIR/MIR plans?

doc/developer/design/20260311_optimizer_customer_tradeoff.md
mgree: I'm not sure I understand: if a blue/green deploy is to change plans, why would you blue/green deploy something that was frozen? You're not allowed to change anything!

doc/developer/design/20260311_optimizer_customer_tradeoff.md
mgree: Okay, this makes sense to me! It's worth confirming that with how our customers are using blue/green deploys, they'll get a sensible error if they try to modify `V2`.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
mgree: I'm fine with best effort here, but... we're already storing these in the catalog... couldn't we continue to do so? (A breaking change in MIR would force a catalog migration anyway, right?)

doc/developer/design/20260311_optimizer_customer_tradeoff.md
mgree: Maybe so---but it's a real con either way! We have not historically been good at managing this (e.g., current EDJ/VOJ situation).

doc/developer/design/20260311_optimizer_customer_tradeoff.md
ggevay: Ah, maybe you want to change something just in one small dataflow of a cluster, but I think a blue-green operates at the whole cluster level, so it would blow away all the pinned plans. Is this the issue, @maheshwarip? If yes, then this is indeed a problem, because it means the splitting workaround (the "Mitigation: use MVs to separate the units you care about.") for the query change issue would not actually work, right?

mgree: The mitigation would work for all of the upstream and downstream frozen clusters. But a downside to plan pinning is that it is very binary: you either have pinned the exact plan, or you are at the mercy of the optimizer.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
ggevay: We are not storing them in the durable catalog. Instead, we recompute them from the SQL at every system startup. But the recomputation wouldn't give back those plans that would correspond to the pinned plans.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
mgree: Makes sense. Per @antiguru, though, it may be hard to know that we're making a breaking change.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
mgree: If you change a query, all bets are off. I look forward to having a more "continuous" optimizer, but I don't think any of the proposals here directly address that.

We're not going to be able to completely eliminate the possibility of regressions (e.g., a change in TD or DD will affect everyone!).

doc/developer/design/20260311_optimizer_customer_tradeoff.md
mgree: This feels like the thorniest question here. LIR has a relatively large surface (e.g., including `MirScalarExpr`, `Row`).

The last big changes to LIR that (a) had to happen and (b) would have been painful for migrations were in February 2026 and then... November 2024. So we likely have some time to think.

We ripped out the protobuf, which seemed like the simplest way to think about migrations. Wrong sometimes, but simple!

If we're truly "best effort", something like bincode will suffice... but then tiny changes at a distance will lead to losing pinned plans. What's the right compromise?

mgree:
Thanks, all (@mtabebe @maheshwarip @antiguru @ggevay) for the feedback! I left comments and updated the document.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
mgree: Updated the doc. The catalog seems like the right place.

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,235 @@
...
+We would:
+
+ - Use `serde_reflection` to build a `Registry` at when testing `compute-types` (where `plan.rs`/LIR lives).
+ - Serialize the `Registry` to YAML/JSON and commit it as `schema.yaml`.
+ - In CI, we regenerate `schema.yaml` from the current types and diff to see what changed.
+ - We can identify changes that are permissible (e.g., addition of a field with default).
+
+When we want to make a breaking change, we will likely have to customize deserialization to support a migration.
+This will be painful the first time.
+It will be less painful if we bake in version numbers up front.
```
mtabebe: Surely we should do this?

doc/developer/design/20260311_optimizer_customer_tradeoff.md
```diff
@@ -0,0 +1,235 @@
...
+
+LIR will be stored in the catalog.
+It could go in `CatalogPlans` (all currently keyed by global ID), in `CatalogState` (alongside per-cluster metadata), or as a sidecar (like the `ExpressionCacheHandle`).
```
mtabebe, line 284: My weakly held opinion is that it doesn't need to be in the catalog, but that is due to my pre-conceived notion of what overhead on the catalog is. In general, it doesn't need to have the same strong consistency, etc.

mgree: I'm not sure where else makes sense to store things, though... could just be my ignorance about the storage set up! If it's not in the catalog, where is it? A persist shard? Where/how do we keep track of that shard?

antiguru (approved):
I think this resolves my concerns. Looks good to me!

PR #35485, opened 2026-03-13
persist: Add KMS envelope encryption for Blob and Consensus

jasonhernandez:
## Summary

Adds optional KMS envelope encryption to `persist`, wrapping both the `Blob`
(S3) and `Consensus` (CockroachDB/Postgres) backends with transparent
encrypt-on-write / decrypt-on-read. Encryption is enabled by passing
`kms_key_id` on the blob and consensus URLs; without it, the backends are
unchanged.

Data is encrypted with AES-256-GCM under a data encryption key (DEK) that is
itself wrapped by a KMS key (envelope encryption). An optional customer-managed
KMS key double-wraps the DEK (two-party mode, the V2 envelope), so decryption
requires both the Materialize and customer keys.

## What's in this PR

- **Envelope format and crypto core** (`crypto.rs`): AES-256-GCM with AAD
  binding the ciphertext to its blob/consensus key, a versioned envelope header
  (V1 single-key, V2 two-party), and `Zeroize`-on-drop for DEK plaintext. DEKs
  are cached and rotated on a configurable interval to bound KMS calls.
- **`EncryptedBlob` and `EncryptedConsensus`**: `Blob`/`Consensus` wrappers that
  encrypt on write and decrypt on read. Consensus coverage matters because shard
  state carries user data (inline batch parts and column statistics), not only
  metadata.
- **Config plumbing** (`cfg.rs`): `BlobConfig`/`ConsensusConfig` parse the KMS
  URL params and construct the wrappers. `strip_kms_query_params` removes the
  persist-specific params before a URL reaches a consumer that does not
  understand them (e.g. the timestamp oracle derived from a shared metadata
  backend URL).
- **environmentd wiring**: KMS params on `--metadata-backend-url` flow to persist
  consensus and are stripped from the derived timestamp oracle URL.
- **Verification**: Kani proofs for encrypt→parse→decrypt structural soundness,
  proptests over the full pipeline, and unit/operational tests including a
  raw-storage check that persisted bytes are ciphertext.

## Non-goals: toggling and migration

Encryption is a create-time property of a persist location. Enabling it for an
environment with existing plaintext data, or disabling it once encrypted data
exists, makes that data unreadable (environmentd fails at boot when consensus
is affected). There is deliberately no plaintext fallback on the read path:
`parse_envelope` rejects non-envelope bytes and `classify_crypto_error` maps
`InvalidEnvelope`/`AuthenticationFailed` to `Determinate` (permanent, no
retry). Two reasons: consensus payloads are arbitrary bytes, so sniffing
plaintext vs. envelope is ambiguous, and a silent fallback is a downgrade
surface. Because AES-GCM is authenticated, every mismatch is a loud classified
error, never silently wrong data. Migrating an existing environment to
encryption (dual-read transitional mode + background re-encrypt, or
blue/green) is explicitly out of scope for this PR.

## Test plan

- [x] `cargo test -p mz-persist` passes
- [x] `cargo clippy -p mz-persist` clean
- [ ] Kani proofs verify (requires Kani toolchain)
- [x] CI green


jasonhernandez:
Closing: deferring KMS envelope encryption work.

PR #35507, opened 2026-03-16
Upgrade lgalloc from 0.6 to 0.7

antiguru:
Upgrade lgalloc to 0.7, which switches from file-backed mappings to anonymous mappings with THP.

* Bump lgalloc to 0.7 in all crates
* Remove `.with_path()` config (no longer file-backed)
* Rename `file_growth_dampener` → `growth_dampener`
* Remove file stats, map stats, and associated metrics/errors
* Remove `MZ_METRICS_LGALLOC_MAP_REFRESH_INTERVAL` dyncfg
* Remove `thiserror` from mz-metrics (no longer needed)
* lgalloc is enabled by default (`enable_lgalloc` was already `true`)

Note: new dyncfg flags (`lgalloc_growth_dampener`) need to be mirrored in LaunchDarkly.


def-:
My main concern here is that we don't test lgalloc anymore. Previously my understanding was that we're planning to retire it, but now I'm surprised we are still developing lgalloc. Should we look into running some tests with lgalloc enabled?

antiguru:
Lgalloc 0.7 doesn't use files anymore, just anonymous memory. The difference is that it uses transparent huge pages instead of 4KiB pages, which can reduce pressure on the TLB. Since Linux 6... swap supports huge pages!

src/compute/src/compute_state.rs
```diff
@@ -284,6 +278,11 @@ impl ComputeState {
         let enable_columnar_lgalloc = ENABLE_COLUMNAR_LGALLOC.get(config);
         mz_timely_util::containers::set_enable_columnar_lgalloc(enable_columnar_lgalloc);
 
+        mz_ore::lgalloc::ENABLE_HEAP_PROFILING.store(
+            ENABLE_LGALLOC_HEAP_PROFILING.get(config),
+            std::sync::atomic::Ordering::Relaxed,
+        );
+
```
def-, line 285: Do we also have to wire this up outside of ComputeState?

src/ore/src/lgalloc.rs
```diff
@@ -0,0 +1,381 @@
...
+
+/// Total bytes currently allocated via lgalloc.
+static LIVE_BYTES: AtomicUsize = AtomicUsize::new(0);
+/// Number of live lgalloc allocations.
+static LIVE_COUNT: AtomicUsize = AtomicUsize::new(0);
+/// Monotonically increasing allocation ID.
+static NEXT_ID: AtomicU64 = AtomicU64::new(0);
+
+/// All profiling state, protected by a single mutex.
+static STATE: Mutex<ProfileState> = Mutex::new(ProfileState::new());
```
def-, line 48: How expensive will this mutex be on a huge cluster with lots of concurrent access?

PR #35508, opened 2026-03-16
docs: document the v1 Materialize CRD and installV1CRD opt-in

alex-hunt-materialize:
Docs follow up for https://github.com/MaterializeInc/materialize/pull/35418

### Motivation

Separate PR so we don't update the docs until we're ready to release.

### Description

Document the changes introduced in https://github.com/MaterializeInc/materialize/pull/35418.

Documents the new `v1` CRD version (hash-based rollouts, no `requestRollout`), which is opt-in twice over:

- The operator only installs the `v1` CRD and its conversion webhook when the `operator.args.installV1CRD` helm value is set. cert-manager (or a user-provided certificate) and API-server ingress to port 8001 are only required when opting in.
- Existing `v1alpha1` CRs keep working unchanged; instances are switched individually by applying a `v1` CR.

Also regenerates the CRD field description data from the `v1` schema and updates the `crd-writer`/`bump-version` tooling for the version rename.


alex-hunt-materialize:
@kay-kim I'm keeping this as a draft so we don't accidentally merge it, but it is ready for review.

I may need to rebase if something comes up in the original PR, but I think the docs commit shouldn't change.

kay-kim:
Thank you!!!

alex-hunt-materialize:
Note to self, we need to document how to run the new orchestratord with webhooks if they aren't using our terraform.

doc/user/shared-content/self-managed/upgrade-notes/v26.19.md
```diff
@@ -0,0 +1,70 @@
+- v26.19.0 introduces a new `v1alpha2` version of the Materialize CRD
```
maheshwarip: I'd recommend phrasing this as something like this:

--
v26.19.0 introduces a new version of the Materialize CRD,`v1alpha2`. The new CRD simplifies rollouts by automatically detecting changes. This means you will no longer need to manually rotate a UUID to trigger a rollout.

Before, on `v1alpha1`:
```yaml
yamlapiVersion: materialize.cloud/v1alpha1
kind: Materialize
metadata:
  name: 12345678-1234-1234-1234-123456789012
  namespace: materialize-environment
spec:
  environmentdImageRef: materialize/environmentd:v26.16.0
  requestRollout: 22222222-2222-2222-2222-222222222222  # ← MUST set a new UUID every upgrade
# forceRollout: 33333333-3333-3333-3333-333333333333   # ← for forced rollouts
  rolloutStrategy: WaitUntilReady
  backendSecretName: materialize-backend
```

After, on `v1alpha2`:
```yaml
apiVersion: materialize.cloud/v1alpha2
kind: Materialize
metadata:
  name: 12345678-1234-1234-1234-123456789012
  namespace: materialize-environment
spec:
  environmentdImageRef: materialize/environmentd:v26.16.0  # ← just change this
# forceRollout: 33333333-3333-3333-3333-333333333333       # ← only for forced rollouts
  rolloutStrategy: WaitUntilReady
  backendSecretName: materialize-backend
```

With the new change, the `requestRollout` field will be removed, along with all previously deprecated fields.

Switching to `v1alpha2` is **opt-in**. You may continue to apply `v1alpha1` CRs and your existing instances will behave exactly as before. We recommend you opt-in to `v1alpha2` at your convenience, as we will migrate to the `v1alpha2` behavior in the next major release. For step-by-step migration instructions, see [Switching from v1alpha1 to v1alpha2](/self-managed-deployments/upgrading/materialize-instances/#switching-from-v1alpha1-to-v1alpha2).

doc/user/shared-content/self-managed/upgrade-notes/v26.19.md
```diff
@@ -0,0 +1,70 @@
...
+  - This new CRD version removes the `requestRollout` field and all previously deprecated fields.
+  - When the Materialize CR is changed using `v1alpha2`, a rollout will begin automatically.
+  - Switching to `v1alpha2` is **opt-in**. You may continue to apply `v1alpha1` CRs and your existing instances will behave exactly as before. We recommend you opt-in to `v1alpha2` at your convenience, as we will migrate to the `v1alpha2` behavior in the next major release. For step-by-step migration instructions, see [Switching from v1alpha1 to v1alpha2](/self-managed-deployments/upgrading/materialize-instances/#switching-from-v1alpha1-to-v1alpha2).
+
+
+{{< important >}}
+
+v26.19 requires infrastructure changes to support migration to v1alpha2 resources.
```
maheshwarip: Do these infrastructure changes need to be made for v26.19 @alex-hunt-materialize ? Or is it only for migrating to v1alpha2?

If it is only to migrate to v1alpha2, I'd actually recommend that we move this into a separate section. At a high level, I'd imagine that the structure can be like this:

- Self-Managed deployments
  - Upgrading page: Give the user basic information about what has changed in v26.19
  - Materialize CRDs Page: Give the user information about v1alpha1 and v1alpha2. Include information about how to migrate from alpha1 to alpha2.

Does that make sense?

doc/user/shared-content/self-managed/upgrade-notes/v26.19.md
```diff
@@ -0,0 +1,70 @@
...
+
+{{< /important >}}
+
+The Materialize Operator now hosts conversion webhooks to convert between
```
maheshwarip: I'd rephrase this a little:
Materialize uses webhooks to allow you to gracefully migrate from alphav1 to alphav2. To enable these webhooks to be delivered, you need to install cert-manager and allow internal network ingress on port 8001. If you're using our terraform modules, this will be handled for you automatically. If you're not using our terraform, you will need to complete these steps manually before upgrading to v26.19. 

- Tab for how to upgrade using our terraform
- Tab for how to upgrade if customers are doing this manually

doc/user/shared-content/self-managed/upgrade-notes/v26.19.md
```diff
@@ -0,0 +1,70 @@
...
+
+**2. Allow network access to the webhook port**
+
+The conversion webhooks require the Kubernetes API server to reach the
+`orchestratord` pod on port `8001`. If your cluster enforces network policies
+or cloud-level firewall rules, you must allow ingress traffic on TCP port
+`8001` from the API server to pods with the label
+`app.kubernetes.io/name: materialize-operator`.
+
+**Kubernetes NetworkPolicy:** Add a policy that allows ingress on port
```
maheshwarip: Could we explicitly specify that this policy will only allow ingress from within the cluster? And not ingress from public internet?

maheshwarip (dismissed):

doc/user/shared-content/self-managed/upgrade-notes/v26.19.md
alex-hunt-materialize: This applies regardless of which CRD version you use. I'll rephrase this line to say:

```v26.19 requires infrastructure changes to support converting between `v1alpha1` and `v1alpha2` resources.```

File: doc/user/shared-content/self-managed/upgrade-notes/v26.19.md (hunk 4 above)
Comment (alex-hunt-materialize) [reply doc/user/shared-content/self-managed/upgrade-notes/v26.19.md]: This example policy allows ingress from everywhere, but they only need to allow it from the Kubernetes API servers. Kubernetes network policies only allow specifying rules based on CIDR blocks or pod labels. The only way to restrict this for the API servers is by CIDR block, which we don't have available here.

Ingress to this port on this pod is safe, there is nothing bad that anyone can do with these endpoints, beyond making orchestratord generate a response.

Additionally, this policy only affects traffic that has reached Kubernetes. They should be running their clusters in an isolated VPC, and the normal cloud firewall rules still apply (ie: security groups for nodes).

File: doc/user/shared-content/self-managed/upgrade-notes/v26.19.md (hunk 3 above)
Comment (alex-hunt-materialize) [reply doc/user/shared-content/self-managed/upgrade-notes/v26.19.md]: > - Tab for how to upgrade using our terraform

They should already know how to upgrade the terraform code. They should have been doing that for literally every release. Do we really need to tell them how to bump a version number? If they can't handle that, _they should not be running self-managed_.

> - Tab for how to upgrade if customers are doing this manually

That's the stuff directly below this.

Comment (maheshwarip) [reply doc/user/shared-content/self-managed/upgrade-notes/v26.19.md]: >They should already know how to upgrade the terraform code. They should have been doing that for literally every release. Do we really need to tell them how to bump a version number? If they can't handle that, they should not be running self-managed.

Yes. Please be explicit here - customers are going to get tripped up otherwise. If I skim this document, I am going to miss the instructions for how to do the upgrade using the MZ terraform.

Comment (maheshwarip) [reply doc/user/shared-content/self-managed/upgrade-notes/v26.19.md]: What I'd recommend doing here is:
- 1 tab for users on our supported terraform
- 1 tab for folks on our legacy terraform. Give them instructions / details for how to migrate to the new terraform.
- 1 tab for manual instructions

Comment (alex-hunt-materialize) [reply doc/user/shared-content/self-managed/upgrade-notes/v26.19.md]: Done, please take another look.

File: doc/user/content/self-managed-deployments/upgrading/materialize-instances/v1.md [hunk 5]
```diff
@@ -0,0 +1,173 @@
...
+
+If you are managing your Materialize instance with the
+[Materialize Terraform modules](https://github.com/MaterializeInc/materialize-terraform-self-managed),
+update the `environmentd_version` variable:
+
+```hcl
+module "materialize_instance" {
+  source = "../kubernetes/modules/materialize-instance"
+
+  crd_version        = "v1"
```
alex-hunt-materialize: Requires https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/242

doc/user/content/security/self-managed/authentication.md
```diff
@@ -50,6 +50,40 @@ following fields:
...
+  license_key: "..."
+  external_login_password_mz_system: "enter_mz_system_password"
+---
+apiVersion: materialize.cloud/v1
+kind: Materialize
+metadata:
+  name: 12345678-1234-1234-1234-123456789012
+  namespace: materialize-environment
+spec:
+  environmentdImageRef: materialize/environmentd:v26.12.1
```
kay-kim: update to v26.30.0

doc/user/content/security/self-managed/sso.md
```diff
@@ -254,7 +254,7 @@ spec:
 ```
 
 Apply the updated manifest to your Kubernetes cluster. See
-[Upgrading](/self-managed-deployments/upgrading/#rollout-configuration) for
+[Upgrading](/self-managed-deployments/upgrading/materialize-instances/v1alpha1/#rollout-configuration) for
```
kay-kim: The above example yaml is for v1alpha and the blurb goes to v1alpha. 
Do we not want to cover v1? If so, we could tab so that we cover both.  And, for this ... not sure if we want to link to the rollout config since we give the exampmle... probably just make sense to link where we apply.

doc/user/content/self-managed-deployments/materialize-crd-field-descriptions/_index.md
```diff
@@ -10,4 +11,7 @@ aliases:
   - /installation/appendix-materialize-crd-field-descriptions/
 ---
 
-{{% self-managed/materialize-crd-descriptions %}}
+Select the CRD API version for your Materialize deployment:
+
+- [v1 (v26.30+)](/self-managed-deployments/materialize-crd-field-descriptions/v1/)
```
kay-kim: Since we're making it an opt-in, would make it (Available starting v26.30+) ...

doc/user/content/security/self-managed/authentication.md
```diff
@@ -97,6 +134,40 @@ To configure Self-Managed Materialize for password authentication, update the fo
...
```
kay-kim: Update version.

Idle curiosity .. what does happen if you try v1 with an older version?

doc/user/shared-content/self-managed/upgrade-notes/v26.30.md
```diff
@@ -0,0 +1,287 @@
...
+
+<div class="code-tabs">
+<ul class="nav-tabs"></ul>
+<div class="tab-content">
+<div class="tab-pane" title="Supported Terraform" id="upgrade-supported-terraform">
+
+If you are using the [supported Terraform
+modules](https://github.com/MaterializeInc/materialize-terraform-self-managed),
+the required infrastructure changes (cert-manager and network ingress) will be
+handled for you automatically.
```
kay-kim: Is this if you have some specific version of the TFs?

doc/user/content/self-managed-deployments/upgrading/materialize-instances/_index.md
```diff
@@ -0,0 +1,86 @@
+---
+title: "Upgrading Materialize Instances"
```
kay-kim: Image

So ... not sure about this section. It cuts right in the middle of the various upgrade tutorials and the upgrade tutorials all have a Upgrade Materialize  instances sections embedded.

- https://preview.materialize.com/materialize/35508/self-managed-deployments/upgrading/upgrade-on-aws/#step-4-upgrading-materialize-instances
- https://preview.materialize.com/materialize/35508/self-managed-deployments/upgrading/upgrade-on-azure/#step-4-upgrading-materialize-instances
- https://preview.materialize.com/materialize/35508/self-managed-deployments/upgrading/upgrade-on-gcp/#step-4-upgrading-materialize-instances

And in this PR, you've already updated those tutorial steps to be v1 ... so, if v1alpha is the default and v1 is "optional", it doesn't quite make sense.  So, instead, could we try incorporating tabbed instructions into the step4s?

doc/user/shared-content/self-managed/upgrade-notes/v26.30.md
```diff
@@ -0,0 +1,287 @@
...
+  environmentdImageRef: materialize/environmentd:v26.16.0  # ← just change this
+# forceRollout: 33333333-3333-3333-3333-333333333333       # ← only for forced rollouts
+  rolloutStrategy: WaitUntilReady
+  backendSecretName: materialize-backend
+```
+
+With the new change, the `requestRollout` field will be removed, along with all previously deprecated fields.
+
+Switching to `v1` is **opt-in**. You may continue to apply `v1alpha1` CRs and your existing instances will behave exactly as before. We recommend you opt-in to `v1` at your convenience, as we will migrate to the `v1` behavior in the next major release. For step-by-step migration instructions, see [Switching from v1alpha1 to v1](/self-managed-deployments/upgrading/materialize-instances/#switching-from-v1alpha1-to-v1).
+
```
kay-kim: Unclear about this content.   We reference  `Switching from v1alpha1 to v1`  content .. so, what's the below?  Is there a sequence to these things? or are they standalone? I guess I'm unclear what user journeys we're supporting.

doc/user/content/self-managed-deployments/_index.md
```diff
@@ -157,6 +157,23 @@ custom resource definitions(CRDs). For a full list of fields available for the
 Materialize CR, see [Materialize CRD Field
 Descriptions](/self-managed-deployments/materialize-crd-field-descriptions/).
 
+{{< tabs >}}
```
kay-kim, line 160: The kind install page seems to have the infra change already, so not in this PR.
Image

- Can we update the Kind install's step 6 to just be a regular "Install cert-manager." instead of "Recommended. Install cert manager"

- Maybe include an Optional step to use v1.

doc/user/content/self-managed-deployments/upgrading/materialize-instances/v1alpha1.md
```diff
@@ -0,0 +1,234 @@
...
+kind: Materialize
+metadata:
+  name: 12345678-1234-1234-1234-123456789012
+  namespace: materialize-environment
+spec:
+  environmentdImageRef: materialize/environmentd:{{< self-managed/versions/get-latest-version >}} # Update version as needed
+  requestRollout: 22222222-2222-2222-2222-222222222222    # Use a new UUID
+  forceRollout: 33333333-3333-3333-3333-333333333333      # Optional: for forced rollouts
+  inPlaceRollout: false                                   # In Place rollout is deprecated and ignored. Please use rolloutStrategy
+  rolloutStrategy: WaitUntilReady                         # The mechanism to use when rolling out the new version. Can be WaitUntilReady or ImmediatelyPromoteCausingDowntime
```
val-materialize: This comment lists only two strategies, but `ManuallyPromote` is documented just below. Suggest including it for completeness (the v1 page has the same comment and could be updated to match):

```suggestion
  rolloutStrategy: WaitUntilReady # The mechanism to use when rolling out the new version. Can be WaitUntilReady, ImmediatelyPromoteCausingDowntime, or ManuallyPromote
```

doc/user/content/self-managed-deployments/upgrading/materialize-instances/v1alpha1.md
```diff
@@ -0,0 +1,234 @@
...
+# Only rolls out if there are changes
+kubectl patch materialize <instance-name> \
+  -n <materialize-instance-namespace> \
+  --type='merge' \
+  -p "{\"spec\": {\"requestRollout\": \"$(uuidgen)\"}}"
+```
+#### `requestRollout` with `forcedRollouts`
+
+Specify a new `UUID` value for `forcedRollout` to roll out even when there are
+no changes to the instance. Use `forcedRollout` with `requestRollout`.
```
val-materialize: Field-name typo: the field is `forceRollout`, not `forcedRollout`/`forcedRollouts` (confirmed against the CRD schema; the `kubectl patch` example on line 144 already uses the correct `forceRollout`). The same typo appears in the note above (lines 122 and 124) and in the section heading on line 135 — worth correcting those too.

```suggestion
Specify a new `UUID` value for `forceRollout` to roll out even when there are
no changes to the instance. Use `forceRollout` with `requestRollout`.
```

doc/user/content/self-managed-deployments/upgrading/materialize-instances/v1.md
```diff
@@ -0,0 +1,173 @@
...
+
+`ImmediatelyPromoteCausingDowntime` tears down the prior generation, and immediately promotes the new generation without waiting for it to hydrate. This causes downtime until the new generation has hydrated. However, it does not require additional resources.
+
+#### *ManuallyPromote*
+
+`ManuallyPromote` allows you to choose when to promote the new generation. This means you can time the promotion for periods when load is low, minimizing the impact of potential downtime for any clients connected to Materialize. This strategy temporarily doubles the required resources to run Materialize.
+
+To minimize downtime, wait until the new generation has fully hydrated and caught up to the prior generation before promoting. To check hydration status, inspect the `UpToDate` condition in the Materialize resource status. When hydration completes, the condition will be `ReadyToPromote`.
+
+To promote, update the `forcePromote` field to match the `requestedRolloutHash` field in the Materialize status. If you need to promote before hydration completes, you can set `forcePromote` immediately, but clients may experience downtime.
```
val-materialize: Quick check on field names here: `forcePromote` checks out in the v1 CRD schema, but I couldn't find `requestedRolloutHash` in the spec — I'm assuming it's a `.status` field. Can you confirm the exact field name and that it lives under `.status` (not `.spec`)? Want to make sure users following this can actually locate the value to match.

doc/user/content/self-managed-deployments/upgrading/materialize-instances/v1.md
```diff
@@ -0,0 +1,173 @@
...
+```yaml
+apiVersion: materialize.cloud/v1
+kind: Materialize
+metadata:
+  name: 12345678-1234-1234-1234-123456789012
+  namespace: materialize-environment
+spec:
+  environmentdImageRef: materialize/environmentd:{{< self-managed/versions/get-latest-version >}} # Update version as needed
+  forceRollout: 33333333-3333-3333-3333-333333333333      # Optional: for forced rollouts
+  rolloutStrategy: WaitUntilReady                         # The mechanism to use when rolling out the new version. Can be WaitUntilReady or ImmediatelyPromoteCausingDowntime
```
val-materialize: Same as on the v1alpha1 page: this comment omits `ManuallyPromote`, which is documented in the Rollout strategies section below. Suggest including it:

```suggestion
  rolloutStrategy: WaitUntilReady # The mechanism to use when rolling out the new version. Can be WaitUntilReady, ImmediatelyPromoteCausingDowntime, or ManuallyPromote
```

doc/user/content/self-managed-deployments/upgrading/materialize-instances/_index.md
```diff
@@ -0,0 +1,86 @@
...
+- [v1 (v26.30+)](/self-managed-deployments/upgrading/materialize-instances/v1/)
+- [v1alpha1 (before v26.30)](/self-managed-deployments/upgrading/materialize-instances/v1alpha1/)
+
+## Switching from v1alpha1 to v1
+
+Switching to v1 is opt-in and does not trigger a rollout on its own. Before switching, ensure you have completed the prerequisites in the [v26.30 upgrade notes](/self-managed-deployments/upgrading/#upgrading-to-v2630-and-later-versions) (cert-manager, network/firewall changes), and have enabled the v1 CRD by setting the Helm value `operator.args.installV1CRD=true` on the operator. Without this value, the operator only installs the v1alpha1 CRD version, and the Kubernetes API server rejects v1 CRs.
+
+### How it works
+
+The v1alpha1 CRD remains the storage version. When you submit a v1 CR, the operator's conversion webhook automatically converts it to v1alpha1 for storage. During conversion, the webhook computes a SHA256 hash of the spec and derives a deterministic `requestRollout` UUID from it. This means:
```
val-materialize: Question on precision: this says the webhook hashes "the spec", and the v1 page says "the spec fields", but #35418 describes the hash as covering a *subset* of spec fields. Could you confirm which it is? If it's a subset, it'd help to say which fields are (or aren't) included, since that's what determines when a rollout fires (e.g. would changing `podAnnotations` trigger one?). Happy to make this page and the v1 page consistent once confirmed.

doc/user/content/self-managed-deployments/upgrading/materialize-instances/_index.md
```diff
@@ -0,0 +1,86 @@
...
+  name: <instance-name>
+  namespace: <materialize-instance-namespace>
+spec:
+  environmentdImageRef: <current-image-ref>
+  backendSecretName: <backend-secret-name>
+  # ... other spec fields (copy from your existing CR, removing requestRollout)
+EOF
+```
+
+Or patch the API version on an existing CR:
```
val-materialize: Does patching only the `apiVersion` actually switch the stored/served version and trigger conversion? My understanding is that `kubectl patch` addresses the object by resource and the stored version stays v1alpha1, so an `apiVersion` in a merge-patch body may be a no-op (it doesn't re-submit the object through the v1 endpoint). If that's the case, this command wouldn't reliably switch an instance — the `kubectl apply` of a full v1 manifest above would. Could you verify this works as written, or drop it in favor of the apply approach?

val-materialize (commented):
Took a pass through the docs against the rendered preview and the generated v1 CRD schema. Overall this is a clear, well-structured set of pages — splitting v1 / v1alpha1 / switching is a real improvement.

I left 6 inline comments: 2 small fixes with suggestions (the `rolloutStrategy` comment missing `ManuallyPromote` on both pages, and a `forceRollout` field-name typo on the v1alpha1 page), and 3 questions to confirm before merge (hash-over-spec wording, whether the `kubectl patch` apiVersion switch actually works, and the `requestedRolloutHash` status field name). Verified separately and looking good: the v1 spec correctly drops `requestRollout`/`inPlaceRollout`, the anchor links resolve, and the legacy-Terraform migration links all resolve.

(Note: this review was prepared with Claude assistance.)

doc/user/content/security/self-managed/authentication.md
```diff
@@ -50,6 +50,40 @@ following fields:
 The following example Kubernetes manifest includes configuration for
 SASL/SCRAM-SHA-256 authentication:
 
+{{< tabs >}}
+{{< tab "v1 (v26.30+)" >}}
```
kay-kim: Actually, since v1alpha is still the default, let's remove from the tabs the `(v26.30+)` and `(before v26.30)`.
Instead:
- Change tab orders so that the v1alpha tab first (this can change in 3-4 releases from now)
- For v1, add a blurb  that it's available starting v26.30+ and link to the specific upgrade note. (We can remove the blurb in 3-4 weeks from now)
- For v1alpha, add a blurb about switching to v1 link to the specific upgrade note. (I don't know if we plan on making it deprecated. In any case, we can remove the blurb in 3-4 weeks from now).

doc/user/content/security/self-managed/authentication.md
```diff
@@ -50,6 +50,40 @@ following fields:
...
+  name: 12345678-1234-1234-1234-123456789012
+  namespace: materialize-environment
+spec:
+  environmentdImageRef: materialize/environmentd:v26.12.1
+  backendSecretName: materialize-backend
+  authenticatorKind: Sasl
+```
+
+{{< /tab >}}
+{{< tab "v1alpha1 (before v26.30)" >}}
```
kay-kim: see comment above about the parenthesis and blurbing.

doc/user/content/security/self-managed/authentication.md
```diff
@@ -97,6 +134,40 @@ To configure Self-Managed Materialize for password authentication, update the fo
 The following example Kubernetes manifest includes configuration for password
 authentication:
 
+{{< tabs >}}
+{{< tab "v1 (v26.30+)" >}}
```
kay-kim: ditto w.r.t. blurb and tab order.

doc/user/content/security/self-managed/authentication.md
```diff
@@ -97,6 +134,40 @@ To configure Self-Managed Materialize for password authentication, update the fo
...
+  name: 12345678-1234-1234-1234-123456789012
+  namespace: materialize-environment
+spec:
+  environmentdImageRef: materialize/environmentd:v26.12.1
+  backendSecretName: materialize-backend
+  authenticatorKind: Password
+```
+
+{{< /tab >}}
+{{< tab "v1alpha1 (before v26.30)" >}}
```
kay-kim: ditto about tabs

doc/user/content/self-managed-deployments/_index.md
```diff
@@ -157,6 +157,23 @@ custom resource definitions(CRDs). For a full list of fields available for the
...
+{{< tab "v1 (v26.30+)" >}}
```
kay-kim: Same comments about tabs as before since it's not the default.

doc/user/content/self-managed-deployments/_index.md
```diff
@@ -168,11 +185,23 @@ spec:
...
+{{< /tab >}}
+{{< /tabs >}}
+
 When you first apply the Materialize custom resource, the operator automatically
 creates all required Kubernetes resources.
 
 #### Modifying the custom resource
 
+{{< tabs >}}
+{{< tab "v1 (v26.30+)" >}}
```
kay-kim: ditto

doc/user/content/self-managed-deployments/configuration-system-parameters.md
```diff
@@ -67,6 +67,24 @@ kubectl apply -f system-params-configmap.yaml
...
+{{< tab "v1 (v26.30+)" >}}
+
+```yaml {hl_lines="9"}
+apiVersion: materialize.cloud/v1
+kind: Materialize
+metadata:
+  name: 12345678-1234-1234-1234-123456789012
+  namespace: materialize-environment
+spec:
+  environmentdImageRef: materialize/environmentd:v26.0.0
```
kay-kim: version.

doc/user/content/self-managed-deployments/configuration-system-parameters.md
```diff
@@ -67,6 +67,24 @@ kubectl apply -f system-params-configmap.yaml
 Reference the ConfigMap in your Materialize custom resource by setting the
 `systemParameterConfigmapName` field to the name of your ConfigMap:
 
+{{< tabs >}}
+{{< tab "v1 (v26.30+)" >}}
```
kay-kim: ditto

doc/user/content/self-managed-deployments/configuration-system-parameters.md
```diff
@@ -129,6 +150,24 @@ Alternatively, you can add the `configmap-reload-trigger` annotation to your
 Materialize custom resource YAML and update it whenever you need to force a
 ConfigMap reload:
 
+{{< tabs >}}
+{{< tab "v1 (v26.30+)" >}}
```
kay-kim: ditto

doc/user/content/self-managed-deployments/upgrading/materialize-instances/v1.md
```diff
@@ -0,0 +1,173 @@
...
+    parent: "upgrading-materialize-instances"
+    identifier: "upgrading-v1"
+    weight: 10
+---
+
+{{< note >}}
+If you are currently using v1alpha1 and want to switch, see [Switching from v1alpha1 to v1](/self-managed-deployments/upgrading/materialize-instances/#switching-from-v1alpha1-to-v1).
+{{</ note >}}
+
+## Upgrading Materialize Instances
```
kay-kim: [See comment above](https://github.com/MaterializeInc/materialize/pull/35508/changes#r3462313738) about incorporating these in the specific upgrade guides so that people aren't on the upgrade guides and in the middle of upgrading, switch over to these pages.

doc/user/content/releases/_index.md
```diff
@@ -1204,7 +1204,7 @@ v26.4.0 introduces several performance improvements and bugfixes.
...
 - **CloudSQL HA failover support** (<red>*Materialize Self-Managed only*</red>): Materialize Self-Managed now offers better support for handling failovers in CloudSQL HA sources, without downtime. [Contact our support team](/support/) to enable this in your environment.
-- **Manual Promotion** (<red>*Materialize Self-Managed only*</red>): [Rollout strategies](/self-managed-deployments/upgrading/#rollout-strategies) allow you control how Materialize transitions from the current generation to a new generation during an upgrade. We've added a new rollout strategy called `ManuallyPromote` which allows you to choose when to promote the new generation. This means that you can minimize the impact of potential downtime.
+- **Manual Promotion** (<red>*Materialize Self-Managed only*</red>): [Rollout strategies](/self-managed-deployments/upgrading/materialize-instances/v1/#rollout-strategies) allow you control how Materialize transitions from the current generation to a new generation during an upgrade. We've added a new rollout strategy called `ManuallyPromote` which allows you to choose when to promote the new generation. This means that you can minimize the impact of potential downtime.
```
kay-kim: I'm just putting the overarching thought process for my review comments on this page since there's not a good place for me to put it (I'll also include this in the slack message to get PMs to weigh in as well ... or if you and the PMs have already decided some of these are out, feel free to pick and choose):

With the v26.30 release: 
:point_right: Starting point: Everyone is on v1alpha (either previously installed or new 26.30+ installs since v1alpha is the default) 
**Goal**: We want to move everyone in the near future to v1 (this could also get people moving off the legacy TFs) ... and no later than before the next major release.
:point_right: We have installation guides and upgrade guides.  To meet the goal, we don't want people popping in and out of the guides.
Can keep the specific upgrade notes and standalone v1alpha to v1 🗒️ 

But, I think for the guides, we should include optional steps to enable v1.
For Kind: 
- Installation guide currently include the v1 infra prep (make this not optional).  We just add an optional step to enable v1 ... so that any new installs starting from now will more than likely use v1.
- Upgrade guide.  Add optional steps to include v1 infra prep (if for some reason, they're upgrading from an older version instead of a teardown and restart)  and enable v1. For the upgrade, making the infra part also optional seems okay (since a really small subset upgrading from an older version for real)

For our non-legacy TFs
- Installation guides: 
  - Add an optional  step to adopt v1 -- basically, v1 enablement  since the latest TF handles the cert-manager and ingress. If we want, we can blurb about Starting in TF version <blah>, the v1 infra prep is automatically handled and  maybe link out to the upgrade notes.
- Upgrade guides.  
  - Add optional steps to include v1 infra prep if on TF versions <earlier than blah> and v1 enablement.
  - Update the upgrading materialize instances to have both v1alpha and v1.

I think that should cover:
- Fresh installs (v1alpha or v1)
- Upgrade from older and remain v1alpha 
- In one upgrade, switch over to v1 as well
- In one upgrade, do just the infra prep. Then, some time later, switch over to v1 either as part of an upgrade or just the crd version switch.

doc/user/data/self_managed/upgrades.yml
```diff
@@ -203,3 +203,39 @@
        ```bash
        kubectl -n materialize-environment describe pod -l app=environmentd
        ```
+
```
kay-kim, line 199: See comment about including both v1alpha and v1 in the upgrade MZ instance step.

doc/user/data/self_managed/upgrades.yml
```diff
@@ -152,7 +152,7 @@
        The **APP VERSION** will be the value that you will use for upgrading
        Materialize instances.
 
-- name: upgrade-materialize-instance
+- name: upgrade-materialize-instance-v1alpha1
```
kay-kim, line 136: Also, add content for enabling v1 in the upgrade guides?

And per the overarching comment up top, to the install, an optional  step to adopt v1 -- basically, v1 enablement  since the latest TF handles the cert-manager and ingress. (We can blurb about Starting in TF version <blah>, the v1 infra prep is automatically handled and maybe link out to the upgrade notes)

doc/user/layouts/shortcodes/self-managed/versions/upgrade/upgrade-steps-local-kind.html
```diff
@@ -63,6 +63,33 @@
 
 1. Create a new `upgrade-materialize.yaml` file, updating the following fields:
 
+   <div class="code-tabs">
```
kay-kim: Add optional steps for both v1 infra prep and v1 enablement (if for some reason, they're upgrading from an older version instead of a teardown and restart)   For the upgrade, making the infra part also optional seems okay (since a really small subset upgrading from an older version for real).  so that it can flow with this step.

(and see comment about including the optional step in the install on kind as well).

doc/user/shared-content/self-managed/upgrade-notes/v26.30.md
```diff
@@ -0,0 +1,287 @@
...
+<div class="tab-content">
+<div class="tab-pane" title="Supported Terraform" id="upgrade-supported-terraform">
+
+If you are using the [supported Terraform
+modules](https://github.com/MaterializeInc/materialize-terraform-self-managed),
+the required infrastructure changes (cert-manager and network ingress) will be
+handled for you automatically.
+
+To upgrade, update each module's `source` to point to the new release tag and
+run `terraform init -upgrade && terraform plan && terraform apply`. To enable
```
kay-kim: There could be other  modules changes if they run the -upgrade? Could they clobber something?

doc/user/data/self_managed/materialize_crd_descriptions_v1.json
```diff
@@ -0,0 +1,558 @@
...
...tion's reason is\n  `ReadyToPromote` the new generation is ready to promote.\n  \n  {{<warning>}}\n  Do not leave new generations unpromoted indefinitely.\n  \n  The new generation keeps open read holds which prevent compaction. Once promoted or\n  cancelled, those read holds are released. If left unpromoted for an extended time, this\n  data can build up, and can cause extreme deletion load on the metadata backend database\n  when finally promoted or cancelled.\n  \n  To guard against this, a rollout that remains in progress longer\n  than `rolloutRequestTimeout` (default 24h) is automatically\n  cancelled.\n  {{</warning>}}\n- `ImmediatelyPromoteCausingDowntime`:<br>  {{<warning>}}\n  THIS WILL CAUSE YOUR MATERIALIZE INSTANCE TO BE UNAVAILABLE FOR SOME TIME!!!\n  \n  This strategy should ONLY be used by customers with physical hardware who do not have\n  enough hardware for the `WaitUntilReady` strategy. If you think you want this, please\n  consult with Materialize engineering to discuss your situation.\n  {{</warning>}}\n  \n  Tear down the old generation of pods and promote the new generation of pods immediately,\n  without waiting for the new generation of pods to be ready.",
```
def-: > When using `ManuallyPromote`, the new generation can be promoted at any time …
> by setting `forcePromote` to the same value as **`requestRollout`** in the
> Materialize spec.

Is this correct for v1? I thought there is no `requestRollout` field.

doc/user/content/security/self-managed/authentication.md
alex-hunt-materialize: It works just fine with older versions. The behavior change is in orchestratord, not in environmentd/clusterd.

Both v1 and v1alpha1 can be used to deploy older or newer versions of Materialize. We just need the updated orchestratord (26.30.0+) and working webhooks for v1.

doc/user/shared-content/self-managed/upgrade-notes/v26.30.md
alex-hunt-materialize: I suppose we should specify the version, but I don't have that yet. I'd like to get https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/242 in, and we also need to bump the helm chart to include v26.30.0 of the operator.

doc/user/content/self-managed-deployments/upgrading/materialize-instances/_index.md
alex-hunt-materialize: Regardless of the method of apply, the stored version will stay v1alpha1. I think you may be correct, though, that `kubectl patch` with a changed `apiVersion` will likely not behave as expected. I'll remove it entirely.

doc/user/data/self_managed/materialize_crd_descriptions_v1.json
alex-hunt-materialize: fixed

doc/user/shared-content/self-managed/upgrade-notes/v26.30.md
alex-hunt-materialize: fixed

alex-hunt-materialize:
@kay-kim and @val-materialize I think I've addressed all your comments. I've done a significant refactor, moving almost everything into tabs or into the root upgrade page.

kay-kim (commented):
Hey Alex -- thanks ever so much for this. ❤️❤️❤️

I don't see any of our terraform install tutorials optionally turning this on.  I can take over from here to make that and any other changes.  But, before I do ... for turning that on ... if people are using the latest TF (assuming that is the one that has v26.30) ... all one has to do is set the `crd_version` variable, yes?  (since the TFs take care of everything else?)

alex-hunt-materialize:
>for turning that on ... if people are using the latest TF (assuming that is the one that has v26.30) ... all one has to do is set the crd_version variable, yes? (since the TFs take care of everything else?)

Yes. I'm still working on getting https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/242 merged, and we'll need to bump the orchestratord helm chart after we release it (Friday?), but otherwise, I think that's all that's required.

I figured we'd also follow up quickly with a major bump of the TF to change the default for that variable. Then all they would need to do is update to the latest TF code.

doc/user/content/self-managed-deployments/upgrading/upgrade-on-kind.md
```diff
@@ -43,7 +43,7 @@ reference](https://kubernetes.io/docs/reference/kubectl/quick-reference/).
 {{< important >}}
 
 The following procedure performs a rolling upgrade, where both the old and new
-Materialize instances are running before the the old instance are removed.
+Materialize instances are running before the old instance are removed.
```
val-materialize: Small grammar fix: `instance are` → `instances are` (agrees with "instances are running" earlier in the sentence).

```suggestion
Materialize instances are running before the old instances are removed.
```

doc/user/data/self_managed/installation.yml
```diff
@@ -86,9 +86,18 @@
     If you run into an error during deployment, refer to the
     [Troubleshooting](/self-managed-deployments/troubleshooting/).
 
+- name: installation-tfvars-crd-version-tip
+  content: |
+    {{< tip >}}
+    CRD API version `v1` is available starting in Materialize v26.30 and
```
val-materialize: This tip (shared across the AWS/Azure/GCP install pages) covers version availability and how to stay on `v1alpha1`, but doesn't mention that choosing `v1` requires cert-manager + API-server ingress to the webhook. A customer reading only this page can't tell whether they have extra setup to do. Could we add a reassuring line that the supported Terraform module provisions this for you? Proposed wording (please confirm module behavior before using verbatim):

> Selecting `v1` requires cert-manager and webhook ingress; the supported Terraform module sets these up for you.

If the module doesn't fully handle it, a link to the v26.30 prerequisites would work instead.

val-materialize (commented):
Reviewed this round (install/upgrade pages + shortcodes) against the rendered preview. Reads well for customers and the FE templates are clean — the variable tables, relabeled "CRD v1alpha1"/"CRD v1" tabs, simplified local-kind flow, and clearer cert-manager step are all nice improvements, and the new anchors (#adopting-the-v1-crd, #install-cert-manager) resolve.

Three small things, none blocking: a version-gating sanity check on the local-kind helm example, a suggestion to reassure readers that the Terraform module handles cert-manager when they pick v1, and a one-click grammar fix. Details inline.

(Note: this review was prepared with Claude assistance.)

doc/user/data/self_managed/installation.yml
kay-kim: So .. I think per https://github.com/MaterializeInc/materialize/pull/35508#issuecomment-4798315810 ... the latest TF does provision them.

doc/user/shared-content/self-managed/upgrade-notes/v26.30.md
kay-kim: So ... to clarify,  we state 
**opt-in** here ->  send people to adopting v1crd , which then points to back to here w.r.t. cert manager and etc. as prereq and back to switching versions.  

So ... given that this section is in the upgrade notes ... it kind of implies needs to do it to upgrade.

Image

But, the [PR's description with  it's "opt-in twice over"](https://github.com/MaterializeInc/materialize/pull/35508/#issue-4083329434) suggests otherwise.  As such, I'm going to move this into the adopting page since we already link there.  Will also ask in slack due to friday availability (and am assuming more than one person should know).

doc/user/content/self-managed-deployments/installation/install-on-azure.md
```diff
@@ -173,20 +173,26 @@ authentication mechanisms.
...
+   {{% include-from-yaml data="self_managed/installation"
+   name="installation-tfvars-crd-version-tip" %}}
 
    ```hcl
    subscription_id     = "your-subscription-id"
    resource_group_name = "mz-demo-rg"
    name_prefix         = "simple-demo"
    location            = "westus2"
    license_key         = "your-materialize-license-key"
+   crd_version = "v1"   # Recommended. Default is v1alpha1; v1 is available starting in Materialize v26.30.
```
kay-kim: > > for turning that on ... if people are using the latest TF (assuming that is the one that has v26.30) ... all one has to do is set the crd_version variable, yes? (since the TFs take care of everything else?)

> Yes. I'm still working on getting https://github.com/MaterializeInc/materialize-terraform-self-managed/pull/242 merged, and we'll need to bump the orchestratord helm chart after we release it (Friday?), but otherwise, I think that's all that's required.

Heh ... so, this doesn't work with the latest TF from github main branch (I had it install v26.30.1)
Is there something else that needs to be flipped?

```
module.materialize_instance.kubectl_manifest.materialize_instance: Creating...
╷
│ Error: materialize-environment/main failed to create kubernetes rest client for update of resource: resource [materialize.cloud/v1/Materialize] isn't valid for cluster, check the APIVersion and Kind fields are valid
│
│   with module.materialize_instance.kubectl_manifest.materialize_instance,
│   on ../../../kubernetes/modules/materialize-instance/main.tf line 29, in resource "kubectl_manifest" "materialize_instance":
│   29: resource "kubectl_manifest" "materialize_instance" {
```

doc/user/shared-content/self-managed/upgrade-notes/v26.30.md
kay-kim: I put a placeholder in the next paragraph as  v3.0.15  😄  so, can change that once things are bumped and such.

doc/user/content/self-managed-deployments/upgrading/adopting-the-v1-crd.md
```diff
@@ -0,0 +1,341 @@
...
+<div class="tab-content">
+<div class="tab-pane" title="Supported Terraform" id="upgrade-supported-terraform">
+
+If you are using the [supported Terraform
+modules](https://github.com/MaterializeInc/materialize-terraform-self-managed),
+the required infrastructure changes (cert-manager and network ingress) will be
+handled for you automatically.
+
+Update each module's `source` to point to the new release tag (v3.0.15 or
+greater), then run `terraform init && terraform plan && terraform apply`. To
```
kay-kim: ... but, pointing the module sources to a new TF version and doing the `terraform init && terraform plan && terraform apply` upgrades the orchestrator version and not the materialize version.
Should we flag this?
- that you're going to end up upgrading your orchestrator version to whatever default is in new version ?

doc/user/content/self-managed-deployments/upgrading/adopting-the-v1-crd.md
```diff
@@ -0,0 +1,341 @@
...
+`v1` CRs.
+
+Choose the tab that matches your deployment method:
+
+<div class="code-tabs">
+<ul class="nav-tabs"></ul>
+<div class="tab-content">
+<div class="tab-pane" title="Supported Terraform" id="upgrade-supported-terraform">
+
+If you are using the [supported Terraform
```
kay-kim, line 83: Q: Do we know how people have been updating their TFs?

That is, are we assuming they do git pull and handle conflicts?

doc/user/content/self-managed-deployments/upgrading/adopting-the-v1-crd.md
```diff
@@ -0,0 +1,341 @@
...
+spec:
+  environmentdImageRef: <current-image-ref>
+  backendSecretName: <backend-secret-name>
+  # ... other spec fields (copy from your existing CR, removing requestRollout)
+EOF
+```
+
+## Using Terraform
+
+If you are managing your Materialize instance with the [Materialize Terraform
```
kay-kim: Note: Need the actual version ; otherwise, they would get the undeclared variable message.

doc/user/content/self-managed-deployments/upgrading/adopting-the-v1-crd.md
```diff
@@ -0,0 +1,341 @@
...
+modules](https://github.com/MaterializeInc/materialize-terraform-self-managed),
+set:
+
+```hcl
+crd_version     = "v1"
+request_rollout = null
```
kay-kim: Same error as a clean install of v1:
```
│ Error: materialize-environment/main failed to create kubernetes rest client for update of resource: resource [materialize.cloud/v1/Materialize] isn't valid for cluster, check the APIVersion and Kind fields are valid
│
│  with module.materialize_instance.kubectl_manifest.materialize_instance,
│  on .terraform/modules/materialize_instance/kubernetes/modules/materialize-instance/main.tf line 29, in resource "kubectl_manifest" "materialize_instance":
│  29: resource "kubectl_manifest" "materialize_instance" {
```

kay-kim (approved):

PR #35529, opened 2026-03-17
Session Scalability tests for console

leedqin:
### Motivation

This PR creates a test suite to understand how different console pages behave in a loaded environment. We can leverage materialize's testing tools like workload capture to create a loaded workload replay environment. 
### What these tests measure

- Query count and latency per page — which queries fire, how often they poll, and their server/queue time breakdown
- Time to content — how long from navigation until meaningful content is visible
- Multi-tab degradation — how latency and queue times increase with 6 concurrent tabs

### What these tests don't measure

- Scale with object count — all tests run against one fixed workload (prod_analytics). They don't answer "what happens with 5000 objects?"
- Server breaking point — tab stress uses 6 browser tabs, not 100+ concurrent users. Finding the actual server limit requires load testing tools (k6)
- Real user experience — these are synthetic lab tests in a controlled environment, not production monitoring

I am keeping track of findings in the markdown file running against production analytics workload capture and will keep updating that as we make improvements to tests.  
###  Future scale testing directions

Object scale | How does page load change with 10 vs 1000 objects? 
- Workload replay with --extra-objects multiplier. I will work on expanding this workload capture to support these tests and update findings 
- User scale:  At what concurrency does environmentd degrade?  Planning to use grafana k6 replaying each page's SQL queries. Will have that in a followup PR


console/e2e-tests/scalability/page-load.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+
+/**
+ * Measures time until one of the given selectors is visible.
+ */
+async function measureUntilVisible(
+  page: Page,
+  selectors: string | string[],
+  timeoutMs = 30_000,
+): Promise<number> {
+  const start = Date.now();
```
SangJunBak: nit: Let's use `performance.now`

console/e2e-tests/scalability/helpers.ts
```diff
@@ -0,0 +1,529 @@
...
+  }
+}
+
+/**
+ * SQL query labels grouped by page.
+ *
+ * Each label maps to a regex that uniquely identifies the query in the
+ * compiled SQL sent to /api/sql. Patterns are derived from the Kysely
+ * query builders in console/src/api/materialize/ and verified against
+ * compiled snapshots in __snapshots__/ directories.
```
SangJunBak: 1. This doesn't seem accurate
2. I know in the future, we want to reduce the number of snapshots and replace them with sql unit tests

console/e2e-tests/scalability/helpers.ts
```diff
@@ -0,0 +1,529 @@
...
+function labelSql(sql: string): string {
+  for (const [label, pattern] of SQL_LABEL_ENTRIES) {
+    if (pattern.test(sql)) return label;
+  }
+  return sql.substring(0, 80);
+}
+
+/** Extracts SQL from a request body and maps each query to a human-readable label. */
+function parseSql(request: Request): string {
+  try {
```
SangJunBak, line 149: The matching logic here feels brittle to me. i.e. if I change the queries or add more queries to each page, there's nothing that tells me to update these tests. If we assume we want to use this e2e tests as a framework for testing scalability and not as a one-off test, I recommend: not trying to match queries but instead, just group them by queryKey, then regex (if a query key doesn't exist if it uses `executeSql` rather than `executeSqlV2`) as a fallback.

Let me know if that's not possible however!

console/e2e-tests/scalability/helpers.ts
```diff
@@ -0,0 +1,529 @@
...
+/** Throws if any known error message is visible on the page. */
+export async function assertNoErrors(page: Page, label: string) {
+  const errors = errorLocator(page);
+  const count = await errors.count();
+  if (count > 0) {
+    const text = await errors.first().textContent();
+    throw new Error(`[${label}] Error visible on page: "${text}"`);
+  }
+}
+
```
SangJunBak: The matching here is brittle but doesn't seem like much we can do about it. I'd suggest the following:

1. Making the patterns more general. i.e. case insensitive, changing from "An error occurred loading cluster data" to "An error occurred.*"
2. Since most of our errors are displayed via ErrorBox, add an aria label / data-test label to it, then match on that selector. As fallback, we can use the regex match too.

console/e2e-tests/scalability/helpers.ts
```diff
@@ -0,0 +1,529 @@
...
+/**
+ * Attaches request/response listeners to record all /api/sql requests on a page.
+ * Uses events instead of route interception so queries are never blocked.
+ */
+export function observeQueries(
+  page: Page,
+  records: QueryRecord[],
```
SangJunBak: I don't think we should accept records as a parameter here. I recommend we create `records` here and return it to keep this function as immutable as we can. However I do notice that in `openTab`, we call it in a loop. In that case, I think we should copy the output `records` into the `records` array outside of the loop.

console/e2e-tests/scalability/helpers.ts
```diff
@@ -0,0 +1,529 @@
...
+    }
+
+    const queuedMs = Math.max(0, timing.requestStart);
+    const serverMs = Math.max(0, timing.responseStart - timing.requestStart);
+    const downloadMs = Math.max(0, timing.responseEnd - timing.responseStart);
+
+    records.push({
+      timestamp: new Date().toISOString(),
+      sql: entry.sql,
+      durationMs: Date.now() - entry.startTime,
```
SangJunBak: nit: we should use `performance.now`

console/e2e-tests/scalability/helpers.ts
```diff
@@ -0,0 +1,529 @@
...
+    entry.maxServerMs = Math.max(entry.maxServerMs, r.serverMs);
+    entry.totalDownloadMs += r.downloadMs;
+    entry.maxDownloadMs = Math.max(entry.maxDownloadMs, r.downloadMs);
+    entry.totalBytes += r.responseBytes;
+    if (r.failed || r.status !== 200) entry.failures++;
+    grouped.set(key, entry);
+  }
+
+  // Main summary table
+  console.log("\n=== Query Summary ===");
```
SangJunBak: Maybe we can use `console.table`?

console/e2e-tests/scalability/helpers.ts
```diff
@@ -0,0 +1,529 @@
...
+  } catch {
+    return "parse-error";
+  }
+}
+
+/**
+ * Attaches request/response listeners to record all /api/sql requests on a page.
+ * Uses events instead of route interception so queries are never blocked.
+ */
+export function observeQueries(
```
SangJunBak: nit: Would maybe rename to `registerQueryListeners` or something with `register` to indicate that we're attaching handlers to the DOM

console/e2e-tests/scalability/query-audit.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+    await page.goto(url);
+    await page
+      .locator('input[placeholder="Search"]')
+      .waitFor({ state: "visible", timeout: 30_000 });
+
+    console.log(`Recording queries for ${AUDIT_DURATION_MS / 1000}s...`);
+    await page.waitForTimeout(AUDIT_DURATION_MS);
+
+    printQuerySummary(records);
+
```
SangJunBak: What happens when `page.waitForTimeout` ends but there are still requests in flight? If I'm understanding it correctly, there wouldn't be a record per inflight HTTP request until it's either finished or errored. Kind of feels like we should mark all inflight queries as "failed"

console/e2e-tests/scalability/helpers.ts
```diff
@@ -0,0 +1,529 @@
...
+      peak = current;
+      peakTime = e.time;
+    }
+  }
+
+  const peakRequests = records.filter((r) => {
+    const start = new Date(r.timestamp).getTime() - r.durationMs;
+    const end = new Date(r.timestamp).getTime();
+    return start <= peakTime && end > peakTime;
+  });
```
SangJunBak, line 553: I think this algorithm is logically sound but could use more comments explaining what's happening 
(e.g. why sort by deltas in `events.sort((a, b) => a.time - b.time || a.delta - b.delta);`,

console/e2e-tests/scalability/helpers.ts
```diff
@@ -0,0 +1,529 @@
...
+  console.log(
+    `Failed requests: ${records.filter((r) => r.failed || r.status !== 200).length}`,
+  );
+}
+
+/**
+ * Compute peak concurrency: the maximum number of requests that were in-flight
+ * simultaneously across all tabs.
+ */
+function printConcurrencySummary(records: QueryRecord[]) {
```
SangJunBak, line 521: nit: I feel like rather than choose a specific intersection point, a better method of profiling might be measuring at a fixed time interval (say, 10 seconds). So every 10 seconds, you can collect how many queries requests were ongoing then you can confidently say "between 4:00:10pm to 4:00:20pm, there were X requests ongoing". That way, you can even create a histogram across time.

console/e2e-tests/scalability/query-audit.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+    const healthChecks = records.filter((r) => r.sql.includes("healthCheck"));
+    const failedHealthChecks = healthChecks.filter(
+      (r) => r.failed || r.status !== 200,
+    );
+    expect(
+      failedHealthChecks.length,
+      `${failedHealthChecks.length} health checks failed`,
+    ).toBe(0);
+
```
SangJunBak: Why do we only do this for the cluster detail test?

console/e2e-tests/scalability/query-audit.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+    const records: QueryRecord[] = [];
+    const page = await context.newPage();
+    observeQueries(page, records, 0);
+
+    const url = objectExplorerUrl(DATABASE_NAME);
+    console.log(`Auditing queries on: ${url}`);
+    await page.goto(url);
+    await page
+      .locator('input[placeholder="Search"]')
+      .waitFor({ state: "visible", timeout: 30_000 });
```
SangJunBak: Why don't we do `openTab` for this test?

SangJunBak (commented):
Overall looking really good! Super happy that you got this working for Staging and workload replay!

console/e2e-tests/scalability/helpers.ts
leedqin: Thanks for the catch! I am now using react queryKeys and only keeping the regex for queries that use snapshots. I will work on migrating them away from `useSqlTyped`

console/e2e-tests/scalability/helpers.ts
leedqin: Added `data-testid="error-box"` to the ErrorBox component and reworked `assertNoErrors` to check that selector first. If ErrorBox isn't visible, it falls back to case-insensitive regexes. Fallback uses` page.getByText() `so it goes through Playwright's text engine.

console/e2e-tests/scalability/query-audit.spec.ts
leedqin: I started out building the tests for this page so overlooked doing it for the rest. I have extracted the logic in a shared auditPage helper — every audit now uses openTab, calls finalize(), and runs expectNoFailedHealthChecks across all five pages.

leedqin:
Addressed comments! Commit message highlights the changes that addresses the suggestions in the review

console/e2e-tests/scalability/helpers.ts
```diff
@@ -0,0 +1,646 @@
...
+    return scopes.join(".");
+  } catch {
+    return undefined;
+  }
+}
+
+function labelFromSqlComment(sql: string): string | undefined {
+  const match = sql.match(/^--([^\n]+)\n/);
+  return match ? match[1].trim() : undefined;
+}
```
SangJunBak: Hmm do we ever comment our sql queries with its name? Kinda feel like the less parsing the better

console/e2e-tests/scalability/helpers.ts
```diff
@@ -511,19 +568,79 @@ function printConcurrencySummary(records: QueryRecord[]) {
...
+  const ends = completed.map((r) => new Date(r.timestamp).getTime());
+  const windowStart = starts.reduce((m, v) => Math.min(m, v), Infinity);
+  const windowEnd = ends.reduce((m, v) => Math.max(m, v), -Infinity);
+  const bucketCount = Math.ceil(
+    (windowEnd - windowStart) / HISTOGRAM_BUCKET_MS,
+  );
+  if (bucketCount <= 1) return;
+
+  const buckets: { bucketStart: number; active: number }[] = [];
+  for (let i = 0; i < bucketCount; i++) {
```
SangJunBak: nit: `i` and j` aren't intuitive here. Maybe `bucketIndex` and sampleIndex`?

console/e2e-tests/scalability/helpers.ts
```diff
@@ -579,33 +580,62 @@ function printConcurrencyHistogram(records: QueryRecord[]) {
...
-  const buckets: { bucketStart: number; active: number }[] = [];
+  // Concurrency only goes up when a request starts, so the peak in a window
+  // is the max active count sampled at the window's start (catches anything
+  // already in flight) plus every request start that falls inside it.
+  const activeIndicesAt = (t: number) =>
+    starts.reduce<number[]>((idxs, s, j) => {
+      if (s <= t && ends[j] > t) idxs.push(j);
+      return idxs;
+    }, []);
+
```
SangJunBak: nit: We should have more verbose/better variable names here. Even `j` feels too arbitrary.

console/e2e-tests/scalability/helpers.ts
```diff
@@ -579,33 +580,62 @@ function printConcurrencyHistogram(records: QueryRecord[]) {
...
-    for (let j = 0; j < starts.length; j++) {
-      if (starts[j] <= midpoint && ends[j] > midpoint) active++;
+    const samplePoints = [
+      bucketStart,
+      ...starts.filter((s) => s >= bucketStart && s < bucketEnd),
+    ];
+    let peak = 0;
+    let peakIdxs: number[] = [];
+    for (const t of samplePoints) {
+      const idxs = activeIndicesAt(t);
```
SangJunBak: Not sure what `peak` does here if we're trying to collect all samples in a bucket

console/e2e-tests/scalability/helpers.ts
```diff
@@ -0,0 +1,744 @@
...
+
+  for (const event of events) {
+    advanceBucketTo(bucketIndexFor(event.time));
+    if (event.type === "start") {
+      currentlyActive.add(event.index);
+    } else {
+      currentlyActive.delete(event.index);
+    }
+    recordPeakAt(currentBucketIndex);
+  }
```
SangJunBak, line 648: Nit: for the most part, functions do make the code more readable/self descriptive. However when functions mutate variables outside its scope, then imo it's actually easier to read the code inline with comments

SangJunBak (approved):

PR #35569, opened 2026-03-20
sources: Speed up MySQL ingestion by parallelizing

def-:
On a cluster with 8 workers, running on my 8 core / 16 thread dev server:
```
$ bin/mzcompose --find feature-benchmark down && bin/mzcompose --find feature-benchmark --release run default --scenario=MySqlInitialLoadMultiWorker --scale=+1
[...]
NAME                                | TYPE            |      THIS       |      OTHER      |  UNIT  | 'THIS' is
---------------------------------------------------------------------------------------------------------------------------------
MySqlInitialLoadMultiWorker         | wallclock       |           2.042 |          11.416 |   s    |    better:  5.6 times faster
MySqlInitialLoadMultiWorker         | memory_mz       |        1213.754 |        1019.012 |   MB   |    worse:  19.1% more
```
Co-written with Claude 🤖

### DIFF
--- a/misc/python/materialize/feature_benchmark/scenarios/benchmark_main.py
+++ b/misc/python/materialize/feature_benchmark/scenarios/benchmark_main.py
@@ -2131,6 +2131,124 @@ CREATE TABLE pk_table (pk BIGINT PRIMARY KEY, f2 BIGINT);
             """)
 
 
+class MySqlInitialLoadMultiWorkerSampled(MySqlCdc):
+    """Measure an 8-worker parallel snapshot of a table whose primary key takes
+    the sampled-boundary split path: a single-column non-integer (char) PK or a
+    composite PK.
+
+    A single-column integer PK is partitioned by a cheap MIN/MAX split. Any
+    other supported PK samples range boundaries from the primary-key index, so
+    its rows also split into disjoint ranges, one per worker. Each worker then
+    reads, decodes, and persists its own range in parallel, so the snapshot of
+    these tables scales with worker count instead of running on a single worker.
+    Compared against the single-worker MySqlInitialLoad this shows the speedup.
+
+    Has subclasses, so the runner treats it as non-leaf and never executes it
+    directly. Pick MySqlInitialLoadMultiWorkerCharPk or
+    MySqlInitialLoadMultiWorkerCompositePk via --root-scenario.
+    """
+
+    # 8 workers naturally use more memory than 1 (more concurrent data in-flight)
+    RELATIVE_THRESHOLD: dict[MeasurementType, float] = {
+        MeasurementType.WALLCLOCK: 0.10,
+        MeasurementType.MEMORY_MZ: 0.60,
+        MeasurementType.MEMORY_CLUSTERD: 0.60,
+    }
+
+    def create_table(self) -> str:
+        """The `CREATE TABLE pk_table (...)` statement whose PK is under test."""
+        raise NotImplementedError
+
+    def row_values(self) -> str:
+        """SELECT list producing one `pk_table` row from the sequence variable
+        `@i`, which must be advanced exactly once via `@i := @i + 1`."""
+        raise NotImplementedError
+
+    def shared(self) -> Action:
+        # Batch the load so scales beyond a single cross-join (mysql.time_zone^2
+        # ≈ 3M rows) still work. @i resumes from the current row count, which is
+        # PK-type agnostic (unlike a MAX(pk) that assumes an integer key).
+        batch = 1_000_000
+        inserts = []
+        remaining = self.n()
+        while remaining > 0:
+            chunk = min(remaining, batch)
+            inserts.append(
+                "SET @i := (SELECT COUNT(*) FROM pk_table);\n"
+                f"INSERT INTO pk_table SELECT {self.row_values()} "
+                f"FROM mysql.time_zone t1, mysql.time_zone t2 LIMIT {chunk};"
+            )
+            remaining -= chunk
+        insert_stmts = "\n".join(inserts)
+        return TdAction(f"""
+$ mysql-connect name=mysql url=mysql://root@mysql password=${{arg.mysql-root-password}}
+
+$ mysql-execute name=mysql
+DROP DATABASE IF EXISTS public;
+CREATE DATABASE public;
+USE public;
+
+{self.create_table()}
+{insert_stmts}
+""")
+
+    def before(self) -> Action:
+        return TdAction("""
+> DROP SOURCE IF EXISTS mz_source_mysqlcdc CASCADE;
+> DROP CLUSTER IF EXISTS source_cluster CASCADE
+            """)
+
+    def benchmark(self) -> MeasurementSource:
+        return Td(f"""
+> CREATE SECRET IF NOT EXISTS mysqlpass AS '${{arg.mysql-root-password}}'
+> CREATE CONNECTION IF NOT EXISTS mysql_conn TO MYSQL (
+    HOST mysql,
+    USER root,
+    PASSWORD SECRET mysqlpass
+  )
+
+> CREATE CLUSTER source_cluster SIZE 'scale=1,workers=8', REPLICATION FACTOR 1;
+
+> CREATE SOURCE mz_source_mysqlcdc
+  IN CLUSTER source_cluster
+  FROM MYSQL CONNECTION mysql_conn;
+> CREATE TABLE pk_table FROM SOURCE mz_source_mysqlcdc (REFERENCE public.pk_table);
+  /* A */
+
+> SELECT count(*) FROM pk_table
+  /* B */
+{self.n()}
+            """)
+
+
+class MySqlInitialLoadMultiWorkerCharPk(MySqlInitialLoadMultiWorkerSampled):
+    """Single-column non-integer (CHAR) PK with ULID-like keys."""
+
+    def create_table(self) -> str:
+        return "CREATE TABLE pk_table (pk CHAR(26) PRIMARY KEY, f2 BIGINT);"
+
+    def row_values(self) -> str:
+        # base-36 of the sequence, left-padded to a fixed width so the keys sort
+        # lexicographically in the same order the boundaries are sampled.
+        return "LPAD(CONV(@i := @i + 1, 10, 36), 26, '0'), @i"
+
+
+class MySqlInitialLoadMultiWorkerCompositePk(MySqlInitialLoadMultiWorkerSampled):
+    """Composite (CHAR, BIGINT) PK, split via row-value range boundaries."""
+
+    def create_table(self) -> str:
+        return (
+            "CREATE TABLE pk_table "
+            "(region CHAR(2) NOT NULL, id BIGINT NOT NULL, f2 BIGINT, "
+            "PRIMARY KEY (region, id));"
+        )
+
+    def row_values(self) -> str:
+        # Spread rows across 5 regions; id stays globally unique so (region, id)
+        # is unique and every region partition is non-empty.
+        return "ELT((@i := @i + 1) % 5 + 1, 'aa', 'bb', 'cc', 'dd', 'ee'), @i, @i"
+
+
 class MySqlStreaming(MySqlCdc):
     """Measure the time it takes to ingest records from MySQL post-snapshot"""
 
--- a/misc/python/materialize/feature_benchmark/scenarios/benchmark_main.py
+++ b/misc/python/materialize/feature_benchmark/scenarios/benchmark_main.py
@@ -2060,6 +2060,77 @@ INSERT INTO pk_table SELECT @i:=@i+1, @i*@i FROM mysql.time_zone t1, mysql.time_
             """)
 
 
+class MySqlInitialLoadMultiWorker(MySqlCdc):
+    """Measure the time it takes to read 1M existing records from MySQL
+    when creating a materialized source with 8 workers.
+
+    With multiple workers and PK-range partitioning, each worker reads
+    a disjoint range of the primary key in parallel. This benchmark
+    measures the effectiveness of that parallelism."""
+
+    # 8 workers naturally use more memory than 1 (more concurrent data in-flight)
+    RELATIVE_THRESHOLD: dict[MeasurementType, float] = {
+        MeasurementType.WALLCLOCK: 0.10,
+        MeasurementType.MEMORY_MZ: 0.60,
+        MeasurementType.MEMORY_CLUSTERD: 0.60,
+    }
+
+    def shared(self) -> Action:
+        # Use batch inserts to support scales beyond what a single
+        # cross-join can produce (mysql.time_zone^2 ≈ 3M rows max).
+        batch = 1_000_000
+        inserts = []
+        remaining = self.n()
+        while remaining > 0:
+            chunk = min(remaining, batch)
+            inserts.append(
+                f"SET @i:=COALESCE((SELECT MAX(pk) FROM pk_table), 0);\n"
+                f"INSERT INTO pk_table SELECT @i:=@i+1, @i*@i "
+                f"FROM mysql.time_zone t1, mysql.time_zone t2 LIMIT {chunk};"
+            )
+            remaining -= chunk
+        insert_stmts = "\n".join(inserts)
+        return TdAction(f"""
+$ mysql-connect name=mysql url=mysql://root@mysql password=${{arg.mysql-root-password}}
+
+$ mysql-execute name=mysql
+DROP DATABASE IF EXISTS public;
+CREATE DATABASE public;
+USE public;
+
+CREATE TABLE pk_table (pk BIGINT PRIMARY KEY, f2 BIGINT);
+{insert_stmts}
+""")
+
+    def before(self) -> Action:
+        return TdAction("""
+> DROP SOURCE IF EXISTS mz_source_mysqlcdc CASCADE;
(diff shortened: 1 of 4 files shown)


def-:
Now also implementation and testing for non-integer and composite keys, numbers from this CI run: https://buildkite.com/materialize/nightly/builds/17115:
```
NAME                                   | TYPE            |      THIS       |      OTHER      |  UNIT  | THRESHOLD  |  Regression?  | 'THIS' is
-----------------------------------------------------------------------------------------------------------------------------------------------------------
MySqlInitialLoadMultiWorker            | wallclock       |           0.619 |           1.854 |   s    |    10%     |      no       | better:  3.0 times faster
MySqlInitialLoadMultiWorker            | memory_mz       |        1017.441 |         952.398 |   MB   |    60%     |      no       | worse:   6.8% more
MySqlInitialLoadMultiWorker            | memory_clusterd |          69.797 |         105.539 |   MB   |    60%     |      no       | better: 33.9% less
MySqlInitialLoadMultiWorkerCharPk      | wallclock       |           0.801 |           2.011 |   s    |    10%     |      no       | better:  2.5 times faster
MySqlInitialLoadMultiWorkerCharPk      | memory_mz       |        1135.566 |         947.895 |   MB   |    60%     |      no       | worse:  19.8% more
MySqlInitialLoadMultiWorkerCharPk      | memory_clusterd |         139.102 |         142.965 |   MB   |    60%     |      no       | better:  2.7% less
MySqlInitialLoadMultiWorkerCompositePk | wallclock       |           0.991 |           2.033 |   s    |    10%     |      no       | better:  2.1 times faster
MySqlInitialLoadMultiWorkerCompositePk | memory_mz       |        1064.559 |         931.352 |   MB   |    60%     |      no       | worse:  14.3% more
MySqlInitialLoadMultiWorkerCompositePk | memory_clusterd |          97.602 |          98.371 |   MB   |    60%     |      no       | better:  0.8% less
```

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -125,6 +132,314 @@ use super::{
...
+            None => (String::new(), chunk),
+        };
+        // Identifiers are quoted via `quote_identifier`, the previous boundary is
+        // itself a value MySQL rendered as a literal, `table` via Display, and the
+        // offset is an integer, so this interpolation is safe; not parameterizable.
+        #[allow(clippy::disallowed_methods)]
+        let row: Option<MySqlRow> = conn
+            .query_first(format!(
+                "SELECT {cols_literal} FROM {table}{predicate} \
+                 ORDER BY {cols_ident} LIMIT 1 OFFSET {offset}"
```
peterdukelarsen, line 367: I think the offset results in an O(offset) scan, so across all partitions you'll have an O(N) scan here. Discussed with some S&S folks and we don't think this is a strict blocker because the decoding is a big part.

peterdukelarsen:
Ran with parallelism 2:

CI: https://buildkite.com/materialize/nightly/builds/17324/list?jid=019f6167-c9db-4e4f-8011-689748bc539b&tab=output

commit tweaking test: https://github.com/MaterializeInc/materialize/pull/37642/changes/5517b8aaf71df1219fa8c596a88ca73daa7881f8

```
NAME                                   | TYPE            |      THIS       |      OTHER      |  UNIT  | THRESHOLD  |  Regression?  | 'THIS' is
---------------------------------------------------------------------------------------------------------------------------------------------------------
MySqlInitialLoadMultiWorkerCharPk      | wallclock       |           1.369 |           2.284 |   s    |    10%     |      no       | better: 40.0% faster
MySqlInitialLoadMultiWorkerCharPk      | memory_mz       |         931.797 |         863.172 |   MB   |    60%     |      no       | worse:   8.0% more
MySqlInitialLoadMultiWorkerCharPk      | memory_clusterd |         157.047 |         138.918 |   MB   |    60%     |      no       | worse:  13.1% more

MySqlInitialLoadMultiWorkerCompositePk | wallclock       |           1.166 |           2.020 |   s    |    10%     |      no       | better: 42.3% faster
MySqlInitialLoadMultiWorkerCompositePk | memory_mz       |         894.367 |         870.566 |   MB   |    60%     |      no       | worse:   2.7% more
MySqlInitialLoadMultiWorkerCompositePk | memory_clusterd |         106.297 |          87.359 |   MB   |    60%     |      no       | worse:  21.7% more

```

misc/python/materialize/feature_benchmark/scenarios/benchmark_main.py
```diff
@@ -2060,6 +2060,195 @@ def benchmark(self) -> MeasurementSource:
...
+
+
+class MySqlInitialLoadMultiWorkerCompositePk(MySqlInitialLoadMultiWorkerSampled):
+    """Composite (CHAR, BIGINT) PK, split via row-value range boundaries."""
+
+    def create_table(self) -> str:
+        return (
+            "CREATE TABLE pk_table "
+            "(region CHAR(2) NOT NULL, id BIGINT NOT NULL, f2 BIGINT, "
+            "PRIMARY KEY (region, id));"
```
peterdukelarsen, line 2243: would be interesting to try:
```
PRIMARY KEY (region ASC, id DESC));
```

I suspect this would cripple performance when running `ORDER BY region, id` (which I think defaults to `ORDER BY region ASC, id ASC`) on a large table because it will trigger a Filesort.

misc/python/materialize/mzcompose/services/mysql.py
```diff
@@ -25,7 +28,7 @@ def create_mysql_server_args(
         "--binlog-row-image=full",
         f"--binlog-row-metadata={binlog_row_metadata}",
         f"--server-id={server_id}",
-        "--max-connections=500",
+        f"--max-connections={max_connections}",
```
peterdukelarsen, line 31: Any sense of how risky this is for customer mysqls? Maybe we want a docs change here. I did a little legwork and most clusters cap out at about 64 workers.

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -29,34 +29,37 @@
...
-//! 'CONSISTENT SNAPSHOT' semantics. Due to linearizability we know that this transaction's view of
-//! the database must some time `t_snapshot` such that `snapshot_upper <= t_snapshot`. We don't
-//! actually know the exact value of `t_snapshot` and it might be strictly greater than
-//! `snapshot_upper`. However, because this transaction will only be used to read the locked tables
-//! and we know that `snapshot_upper` is an upper bound on all the writes that have happened to
-//! them we can safely pretend that the transaction's `t_snapshot` is *equal* to `snapshot_upper`.
-//! We have therefore succeeded in starting a transaction at a known point in time!
+//! For tables with a suitable primary key, the leader computes `worker_count - 1` boundary keys
+//! that split the key domain into disjoint half-open ranges, and broadcasts them. Each worker
+//! reads only its assigned range. A single-column integer PK uses a cheap `MIN(pk)`/`MAX(pk)`
```
peterdukelarsen, line 43: I think we should remove this feature for now and stick with the table-scanning approach for simplicity. The table-scanning approach will be much more resilient to uneven distributions. Let's leave adding this as an optional optimization to consider in the future.

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -125,6 +132,314 @@ use super::{
...
+enum PkColKind {
+    /// Integer type, rendered and compared as a bare numeric literal.
+    Numeric,
+    /// Character type, rendered as a quoted string literal via MySQL `QUOTE()`
+    /// and compared under the column's own collation.
+    Text,
+}
+
+/// Classify a scalar type for range splitting, or `None` if unsupported.
+fn pk_col_kind(scalar_type: &SqlScalarType) -> Option<PkColKind> {
```
peterdukelarsen, line 167: Why stick to such a limited set of types?

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -125,6 +132,314 @@ use super::{
...
+/// the quoted PK columns with their kinds, in key order. Used for the general
+/// (non-integer and/or composite) sampling path in [`compute_sampled_splits`].
+fn formattable_pk(desc: &mz_mysql_util::MySqlTableDesc) -> Option<Vec<(String, PkColKind)>> {
+    let pk = desc.keys.iter().find(|k| k.is_primary)?;
+    let mut cols = Vec::with_capacity(pk.columns.len());
+    for name in &pk.columns {
+        let col = desc.columns.iter().find(|c| &c.name == name)?;
+        let kind = pk_col_kind(&col.column_type.as_ref()?.scalar_type)?;
+        cols.push((quote_identifier(name), kind));
+    }
```
peterdukelarsen, line 192: Did a little research and I'm pretty confident this ordering of respecting the primary key definition is correct:
- there can be at max 1 primary key for a mysql table
- the ordering defined in the primary key definition (pk.columns) is the ordering that's respected for the primary key's underlying b-tree

src/storage/src/source/mysql/snapshot.rs
peterdukelarsen: Did a little claude research and it sounds like float/double conversions to string could be lossy. DECIMAL should work. BINARY and similar types likely won't cast cleanly to a String. ENUM may not compare the same way and could be risky. There were some other types considered too, but generally makes sense that round-tripping to String is not guaranteed to work for all types and that there may be more complex concerns than just round-tripping to String for other types (like ENUMs).

Would be interested if there was a way we could avoid round-tripping to a string to make support easier to extend correctly? i.e. can we keep it in the Value type of the my_sql library?

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -125,6 +132,314 @@ use super::{
...
+}
+
+/// PK-range partition boundaries for a table, computed by the leader.
+///
+/// `pk_cols` are the quoted PK column identifiers in key order. `boundaries`
+/// holds `partition_count - 1` boundary keys; each is one SQL literal per PK
+/// column. Boundary `i` is the lower bound of partition `i + 1`, so the
+/// partitions are the half-open ranges `[boundaries[i-1], boundaries[i])`.
+///
+/// INVARIANT: boundaries are non-decreasing under the same comparison the range
```
peterdukelarsen, line 203: To support multi-column efficiently I think non-decreasing constraint won't be exactly right. They just need to be ordered sequentially in the same sort as applied by mysql for the primary key and we'll need to record what ordering we observed for the primary keys and likely just assume that won't get updated out from under us. Would be good to gut-check this can't change out from under us without a failure occurring.

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -125,6 +132,314 @@ use super::{
...
+    })
+}
+
+/// Split a single-column integer PK into evenly spaced ranges using `MIN`/`MAX`,
+/// without scanning the table. Returns `None` if the key domain is too small to
+/// split across `worker_count` workers, or the table is empty.
+///
+/// Uses `i128` intermediate arithmetic to avoid overflow when the range spans a
+/// large portion of the `i64` domain (e.g. `min = 0, max = 2^62`).
+async fn compute_integer_splits<Q>(
```
peterdukelarsen, line 267: let's simply delete this

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -125,6 +132,314 @@ use super::{
...
+    pk_cols: &[(String, PkColKind)],
+    worker_count: usize,
+) -> Result<Option<PkSplits>, LeaderError>
+where
+    Q: Queryable,
+{
+    // `table` is escaped via `MySqlTableName::Display`; not parameterizable.
+    #[allow(clippy::disallowed_methods)]
+    let total: Option<u64> = conn
+        .query_first(format!("SELECT COUNT(*) FROM {}", table))
```
peterdukelarsen, line 330: We likely want to skip this and use information_schema.TABLES instead. This is an O(table size) operation whereas information_schema.TABLES should be cached and cheap.

By default information_schema.TABLES is cached for 24h. For tests we can force analyze. For customers I think 24h lag should be fine. There's also a chance that it's permanently stale via configuration, in which case I think we should just fall back or expect to partition in a skewed way -- both should be fine. If a customer asks for more accurate partitioning we can ask them to analyze the table more frequently.

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -125,6 +132,314 @@ use super::{
...
+    if partitions < 2 {
+        return Ok(None);
+    }
+    let chunk = total / partitions;
+    let cols_ident = pk_cols.iter().map(|(c, _)| c.as_str()).join(", ");
+    let cols_literal = pk_cols
+        .iter()
+        .map(|(c, kind)| match kind {
+            PkColKind::Numeric => format!("CAST({c} AS CHAR)"),
+            PkColKind::Text => format!("QUOTE({c})"),
```
peterdukelarsen, line 344: Can we avoid to the string casting and round-trip of strings by holding onto the typed value as a parameter? This will add some complexity to the PkSplits broadcast because now we'll need to serde the full type rather than a simple string, but it could make supporting other types simpler and less error prone.

src/storage/src/source/mysql/snapshot.rs
peterdukelarsen: Moving to the `Value` type definitely not required and might lose us some protection of string conversion -- namely enforcing UTF8. Regardless, we should test with a couple of different charsets to make sure this is resilient to that -- or at least isn't a regression.

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -220,73 +545,175 @@ pub(crate) fn render<'scope>(
...
+                        // acquiring the lock. Single-column integer PKs use the
+                        // cheap MIN/MAX split; other supported PKs sample boundaries
+                        // with one PK-index scan, which is O(rows). Doing this work
+                        // before `LOCK TABLES` keeps the lock-hold window O(1) rather
+                        // than scaling with table size. Boundaries need not match the
+                        // exact snapshot point: they only partition the key domain and
+                        // stay correct for any data distribution (see `PkSplits`).
+                        // `None` means single-worker fallback.
+                        let mut pk_bounds_map = BTreeMap::new();
+                        for (table, outputs) in &tables_to_snapshot {
```
peterdukelarsen, line 579: Let's open up a transaction for the computation of all of the partitioned tables here and make sure it's not closed until after the locks are acquired, the other workers have snapshotted, and the locks are released.

The concern driving this suggestion is around collation changes to the char primary key. If collation is altered between computing the partition boundaries and executing our snapshot there's a small risk we could have duplicate or dropped data.

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -220,73 +545,175 @@ pub(crate) fn render<'scope>(
...
+                            let desc = &outputs[0].desc;
+                            // With a single worker there is only one reader, so
+                            // there is nothing to split. Skip the split probes
+                            // (an O(rows) COUNT(*)/index scan for sampled PKs, a
+                            // MIN/MAX probe for integer PKs) and fall back to
+                            // single-worker mode.
+                            let splits = if config.worker_count < 2 {
+                                None
+                            } else if let Some(pk_col) = integer_pk_col(desc) {
+                                compute_integer_splits(
```
peterdukelarsen, line 589: Given the linear cost of reading these tables being imposed on the leader, I'm slightly concerned we could have a regression in snapshot performance, i.e. imagine the case where snapshotting a small table would have been executed successfully within a minute or two but is now blocked behind computing the partitions for the largest table.

One option for improving on this could be some version of concurrency control locally since all the work happens in mysql, i.e. we could submit up to #workers of parallel tasks to compute the partitions which would bring us closer to historical performance.

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -299,127 +726,197 @@ pub(crate) fn render<'scope>(
...
+                    .keys()
+                    .filter_map(|table| {
+                        let plan = match snapshot_info.pk_bounds.get(table) {
+                            Some(Some(splits)) => match worker_pk_range(splits, config.worker_id) {
+                                Some(range) => Some(ReadPlan::Range(range)),
+                                None => config
+                                    .responsible_for(table)
+                                    .then_some(ReadPlan::RewindOnly),
+                            },
+                            _ => config
```
peterdukelarsen, line 747: I think we should error on the `None` case here because we insert something for each table up top.

src/storage/src/source/mysql/snapshot.rs
```diff
@@ -299,127 +726,197 @@ pub(crate) fn render<'scope>(
...
-                                outputs.len(),
-                                export_statistics.get(name).unwrap(),
-                            )
-                        })
-                        .collect(),
-                    metrics,
-                )
-                .await?;
+                // Phase G: The leader publishes the full snapshot size so the summed
+                // worker-local gauges reflect the upstream total without double-counting.
```
peterdukelarsen, line 876: code style nit: I have found these phases somewhat useful, but I wonder if we could get similar benefit with less verbose comments by factoring out some functions to function a little smaller

peterdukelarsen (changes requested):
Let's simplify this a bit by focusing just on a single-column primary key and using just one method for computing partitions (the one that supports the `char(26)` type). Let's flesh out the serde testing for the various types a little more aggressively to try to root out issues with round-tripping the stringified value -- including testing different character sets.

def-:
Closing in favor of https://github.com/MaterializeInc/materialize/pull/37642

PR #35703, opened 2026-03-26
license-keys: Zeroize signing strings, digests, and key bytes

jasonhernandez:
### Motivation

Defense-in-depth: zero sensitive signing material from memory after use.

### Description

- Wrap `signing_string` (JWT header.payload pre-image) in `Zeroizing<String>`
- Extract `compute_digest()` helper returning `Zeroizing<Vec<u8>>` for SHA-256 digests
- Wrap raw public key bytes from AWS KMS in `Zeroizing<Vec<u8>>`
- Document that `jsonwebtoken::DecodingKey` is opaque and cannot be zeroized

No public API changes.

### Depends on

- #35635 (`ore::secure` module)

Part of [SEC-115](https://linear.app/materializeinc/issue/SEC-115).


jasonhernandez:
I stripped some not very useful comments

doy-materialize (commented):
i'm not really sure i understand the purpose here - as far as i can tell, none of the things we're zeroizing here are actually secret? what's the goal we're trying to achieve here?

PR #35753, opened 2026-03-27
Iceberg: support interval and range types

DAlperin:
Fixes: https://github.com/MaterializeInc/database-issues/issues/11305

### DIFF
--- a/src/arrow-util/src/builder.rs
+++ b/src/arrow-util/src/builder.rs
@@ -350,9 +350,10 @@ where
                 ),
             }
         }
-        // arrow::datatypes::IntervalUnit::MonthDayNano is not yet implemented in the arrow parquet writer
-        // https://github.com/apache/arrow-rs/blob/0d031cc8aa81296cb1bdfedea7a7cb4ec6aa54ea/parquet/src/arrow/arrow_writer/mod.rs#L859
-        // SqlScalarType::Interval => DataType::Interval(arrow::datatypes::IntervalUnit::DayTime)
+        SqlScalarType::Interval => (
+            DataType::Interval(arrow::datatypes::IntervalUnit::MonthDayNano),
+            "interval",
+        ),
         SqlScalarType::Array(inner) => {
             // Postgres / MZ Arrays are weird, since they can be multi-dimensional but this is not
             // enforced in the type system, so can change per-value.
@@ -426,6 +427,53 @@ where
                 arrow_fields.push(Arc::new(field));
             }
             (DataType::Struct(arrow_fields.into()), "record")
+        }
+        SqlScalarType::Range { element_type } => {
+            // Ranges are represented as Arrow Structs with 5 fields:
+            //   - lower: nullable element value (the bound datum, null = infinite)
+            //   - upper: nullable element value (the bound datum, null = infinite)
+            //   - lower_inclusive: bool
+            //   - upper_inclusive: bool
+            //   - empty: bool
+            let (inner_type, inner_name) = scalar_to_arrow_datatype_impl(element_type, overrides)?;
+            let lower_field = Arc::new(field_with_typename(
+                "lower",
+                inner_type.clone(),
+                true,
+                &inner_name,
+            ));
+            let upper_field = Arc::new(field_with_typename("upper", inner_type, true, &inner_name));
+            let lower_inclusive_field = Arc::new(field_with_typename(
+                "lower_inclusive",
+                DataType::Boolean,
+                false,
+                "boolean",
+            ));
+            let upper_inclusive_field = Arc::new(field_with_typename(
+                "upper_inclusive",
+                DataType::Boolean,
+                false,
+                "boolean",
+            ));
+            let empty_field = Arc::new(field_with_typename(
+                "empty",
+                DataType::Boolean,
+                false,
+                "boolean",
+            ));
+            (
+                DataType::Struct(
+                    [
+                        lower_field,
+                        upper_field,
+                        lower_inclusive_field,
+                        upper_inclusive_field,
+                        empty_field,
+                    ]
+                    .into(),
+                ),
+                "range",
+            )
         }
         _ => anyhow::bail!("{:?} unimplemented", scalar_type),
     };
@@ -543,6 +591,11 @@ fn builder_for_datatype(
             } else {
                 anyhow::bail!("Expected map entries to be a struct")
             }
+        }
+        DataType::Interval(arrow::datatypes::IntervalUnit::MonthDayNano) => {
+            ColBuilder::IntervalMonthDayNanoBuilder(IntervalMonthDayNanoBuilder::with_capacity(
+                item_capacity,
+            ))
         }
         _ => anyhow::bail!("{:?} unimplemented", data_type),
     };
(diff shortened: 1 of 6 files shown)


bosconi:
@claude what's the word? Use /mz-pr-review please.

bosconi:
@DAlperin I hope it's OK to use this draft PR as a test subject for Claude code review...

bosconi:
@claude can you run on this fork PR now we have patched `claude-code-action`? What branch are you running on? What do you see here?

bosconi:
@claude and now do you start on the PR branch?

src/arrow-util/src/builder.rs
```diff
@@ -890,6 +955,108 @@ impl ArrowColumn {
...
+                                    .unwrap()
+                                    .append_datum(nested.datum())?;
+                            }
+                            None => {
+                                struct_builder
+                                    .field_builder::<ArrowColumn>(0)
+                                    .unwrap()
+                                    .append_datum(Datum::Null)?;
+                            }
+                        }
```
ublubu: _ambivalent nit:_ It might be nice to push the `match` inside the `append_datum`.

Could even do `append_datum(inner.lower.bound.map(DatumNested::datum).unwrap_or(Datum::Null));`

src/arrow-util/src/builder.rs
```diff
@@ -890,6 +955,108 @@ impl ArrowColumn {
...
+                        {
+                            b.append_value(inner.upper.inclusive);
+                        }
+                        if let ColBuilder::BooleanBuilder(b) = &mut struct_builder
+                            .field_builder::<ArrowColumn>(4)
+                            .unwrap()
+                            .inner
+                        {
+                            b.append_value(false);
+                        }
```
ublubu: Are these columns somehow optional in the output?

src/arrow-util/src/builder.rs
```diff
@@ -890,6 +955,108 @@ impl ArrowColumn {
...
+                    }
+                }
+                struct_builder.append(true);
```
ublubu, line 1027: This is some kind of delimiter?

src/arrow-util/src/reader.rs
```diff
@@ -844,6 +899,81 @@ impl ColReader {
...
+                lower,
+                lower_nulls,
+                upper,
+                upper_nulls,
+                lower_inclusive,
+                upper_inclusive,
+                empty,
+                nulls,
+            } => {
+                let is_valid = nulls.as_ref().map(|n| n.is_valid(idx)).unwrap_or(true);
```
ublubu, line 912: It bugs me that this is called "is_valid" (I feel like Nulls are valid too), but I see that's the established pattern.

src/arrow-util/src/reader.rs
```diff
@@ -391,6 +392,49 @@ fn scalar_type_and_array_to_reader(
...
+                .ok_or_else(|| anyhow::anyhow!("range struct missing 'upper' field"))?;
+            let empty_array = struct_array
+                .column_by_name("empty")
+                .ok_or_else(|| anyhow::anyhow!("range struct missing 'empty' field"))?;
+            let lower_inclusive_array = struct_array
+                .column_by_name("lower_inclusive")
+                .ok_or_else(|| anyhow::anyhow!("range struct missing 'lower_inclusive' field"))?;
+            let upper_inclusive_array = struct_array
+                .column_by_name("upper_inclusive")
+                .ok_or_else(|| anyhow::anyhow!("range struct missing 'upper_inclusive' field"))?;
```
ublubu, line 411: These columns look required here.

src/arrow-util/src/reader.rs
```diff
@@ -844,6 +899,81 @@ impl ColReader {
...
+                    lower.read(idx, &mut temp.packer())?;
+                    Some(temp.clone())
+                };
+
+                let upper_row = if upper_is_infinite {
+                    None
+                } else {
+                    let mut temp = SharedRow::get();
+                    upper.read(idx, &mut temp.packer())?;
+                    Some(temp.clone())
```
ublubu: We wrap up the row just to `as_ref().map(...)` back into it a few lines down. What if we build our packing function here instead?

```rust
Some(|packer: &mut RowPacker| -> Result<(), anyhow::Error> {
    packer.push(temp.unpack_first());
    Ok(())
});
```

src/arrow-util/src/builder.rs
ublubu: I see, these are all pattern matches that are expected to succeed.

This feels like an an opportunity to instrument the negative cases with `anyhow_bail!`, since a) we're already yielding errors from this function and b) we're trusting a ColBuilder that was created somewhere else (not in this function).

src/arrow-util/src/builder.rs
DAlperin: Yeah, this is just a weirdness of how the Arrow [StructBuilder](https://arrow.apache.org/rust/arrow/array/struct.StructBuilder.html) works. You call .append(is_valid) to commit the row and associate nullability data with it.

src/arrow-util/src/reader.rs
DAlperin: As with above, stolen from arrows weird convention

src/arrow-util/src/builder.rs
DAlperin: Yeah you are right this is unneeded complexity!

src/arrow-util/src/builder.rs
```diff
@@ -766,6 +820,9 @@ impl ArrowColumn {
                 builder.append_value(val.as_bytes())?
             }
             (ColBuilder::StringBuilder(builder), Datum::String(s)) => builder.append_value(s),
+            (ColBuilder::StringBuilder(builder), _) if self.extension_type_name == "jsonb" => {
+                builder.append_value(JsonbRef::from_datum(datum).to_serde_json().to_string())
+            }
```
ublubu, line 825: This looks reasonable to include. Is it an unrelated bugfix or something that popped up while implementing Interval?

ublubu (approved):
LGTM

src/arrow-util/src/reader.rs
patrickwwbutler: this bothered me too lol

patrickwwbutler (commented):
LGTM as well! This technically should add support for the "range" type to `COPY FROM PARQUET` as well, assuming they use the struct format!

patrickwwbutler (approved):
LGTM as well! This technically should add support for the "range" type to `COPY FROM PARQUET` as well, assuming they use the struct format!

I guess we should probably test it first, but you could even arguably update the COPY FROM docs to say the range type is supported?

test/testdrive/copy-s3-roundtrip-minio.td
```diff
@@ -303,14 +303,14 @@ $ s3-verify-data bucket=copytos3 key=parquet_test/3 sort-rows=true
...
-# Confirm that unimplemented types will early exit before writing to s3
+# Confirm that unsupported types error when writing to s3 parquet
 ! COPY (SELECT '0 day'::interval)  TO 's3://copytos3/parquet_test/5'
   WITH (
     AWS CONNECTION = aws_conn,
     MAX FILE SIZE = "100MB",
     FORMAT = 'parquet'
   );
-contains:Cannot encode the following columns/types: ["interval: Interval"]
+contains:Arrow interval type MonthDayNano to parquet that is not yet implemented
```
patrickwwbutler, line 313: perhaps when I have a sec I'll go here and add a roundtrip COPY for range and update COPY FROM docs

src/arrow-util/src/builder.rs
DAlperin: I think just came up in fuzzing the fix

PR #35776, opened 2026-03-28
compute: serialize DataflowError on dataflow edges

antiguru:
## Summary

- Introduce `DataflowErrorSer`, a newtype over proto-encoded bytes, to replace `DataflowError` on compute-internal dataflow edges (streams, arrangements).
- Make `persist_source` generic over the error type `E`, so compute callers can choose `DataflowErrorSer` while storage callers stay on `DataflowError`.
- Switch all compute rendering, arrangement types (`ErrSpine`, `RowErrSpine`, etc.), `CollectionBundle`, sinks, and logging to use `DataflowErrorSer`.

## Motivation

`DataflowError` is a complex enum with boxed variants (`EvalError` alone has 50+ sub-variants). It requires a 700-line hand-crafted `Columnation` impl. In compute rendering, errors are **never inspected** — they flow opaquely through concat, consolidate, arrange, and are only displayed (`.to_string()`) at sinks.

Replacing it with opaque serialized bytes:
- Eliminates the need for the complex `Columnation` impl in compute (delegates to trivial `Vec<u8>` columnation)
- Removes complex structures from dataflow edges
- Paves the way for eventually using `Row` to represent serialized errors

## Design

**Serialization boundary:** Errors are serialized at the `persist_source` boundary when entering compute, and deserialized only at exit points (MV sink and continual task for persist writes; subscribe and copy-to-s3 via `Display`).

**Canonicality invariant:** The MV sink's self-correction logic computes `desired - persist` and relies on equality for cancellation. Proto3 encoding via prost is deterministic for messages without `map` fields (verified: none in the error proto chain), so byte-equality implies semantic equality.

**Persist boundary unchanged:** `SourceData` already serializes errors as proto bytes in Arrow columnar encoding. This change does not affect the persist format.

## Testing

- Added proptest verifying round-trip canonicality: `ser(deser(ser(e))) == ser(e)` for all `DataflowError` variants.
- Added test verifying `Display` equivalence between `DataflowError` and `DataflowErrorSer`.
- All existing mz-compute and mz-storage-operators tests pass.

## Future work

- Direct `EvalError → proto bytes` (avoids a `Box::new` allocation in the `From<EvalError>` path)
- `SourceData::append_err_bytes` to skip deserialize/reserialize at MV sink
- Error code + optional description format for compute-created errors
- `Columnar` impl for `DataflowErrorSer` (enables newer spine infrastructure)


src/compute/src/render/errors.rs
```diff
@@ -8,8 +8,134 @@
...
+//!
+//! # `DataflowErrorSer`
+//!
+//! [`DataflowErrorSer`] is a serialized byte representation of
+//! [`DataflowError`] used on compute-internal dataflow edges instead of
+//! `DataflowError` directly.
+//!
+//! It is backed by proto-encoded [`ProtoDataflowError`] bytes. Because proto3 +
+//! prost + no map fields = deterministic encoding, byte-equality implies
+//! semantic equality, which lets us use `Ord`, `Hash`, etc. directly on the
```
DAlperin, line 20: This claim is only true within a single prost version https://github.com/tokio-rs/prost/issues/965, but any oddness there is a preexisting issue.

test/testdrive/introspection-sources.td
```diff
@@ -290,7 +290,7 @@ true true true true true
 
 # Cast to numeric to get its rounding behavior
 > SELECT records, (size::numeric/1024/1024)::int FROM mz_introspection.mz_dataflow_arrangement_sizes WHERE name LIKE '%vt5_idx'
-10000 1
```
DAlperin, line 293: hmm why does this change?

src/compute/src/render/errors.rs
```diff
@@ -8,8 +8,134 @@
...
+//! bytes.
+//!
+//! **Invariant**: NEVER add `map` fields to `ProtoDataflowError` or any of its
+//! transitive message types, as map fields have non-deterministic encoding
+//! order in protobuf.
```
DAlperin, line 25: Please add this comment to errors.proto as well. And perhaps add a lint in check-protobuf.sh to make sure we never break this.

test/testdrive/introspection-sources.td
antiguru: I think the representation is just smaller. I don't like this test, so a change in this direction is fine with me.

src/compute/src/render/errors.rs
antiguru: Done in 7584508b0d — copied the invariant to `errors.proto` and added a transitive-closure lint in `check-protobuf.sh` that walks imports from `errors.proto` and fails if any reachable message uses a `map<>` field.

src/compute/src/render/errors.rs
antiguru: Acknowledged — the serialized bytes never cross a prost version boundary (they only live on in-memory dataflow edges between peers of the same binary), so cross-version variance is not a concern here.

DAlperin (approved):

PR #35780, opened 2026-03-30
mcp: validate data product names before interpolating into SQL

bobbyiliev:
Validates that data product names passed to read_data_product are syntactically valid SQL identifiers by parsing a probe query locally, preventing malformed or injected names from reaching the database as opaque SQL errors.

### DIFF
--- a/src/environmentd/src/http/mcp.rs
+++ b/src/environmentd/src/http/mcp.rs
@@ -35,9 +35,10 @@ use mz_adapter_types::dyncfgs::{
 use mz_repr::namespaces::{self, SYSTEM_SCHEMAS};
 use mz_sql::parse::parse;
 use mz_sql::session::metadata::SessionMetadata;
-use mz_sql_parser::ast::display::escaped_string_literal;
+use mz_sql_parser::ast::display::{AstDisplay, escaped_string_literal};
 use mz_sql_parser::ast::visit::{self, Visit};
 use mz_sql_parser::ast::{Raw, RawItemName};
+use mz_sql_parser::parser::parse_item_name;
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 use thiserror::Error;
@@ -768,6 +769,33 @@ async fn get_data_product_details(
     format_rows_response(rows, max_response_size)
 }
 
+/// Parses a data product name and returns it safely quoted for SQL interpolation.
+///
+/// Uses the SQL parser to validate the name as an `UnresolvedItemName`, then
+/// formats it with `FormatMode::Stable` so every identifier part is
+/// double-quoted with proper escaping. This prevents SQL injection regardless
+/// of the input.
+fn safe_data_product_name(name: &str) -> Result<String, McpRequestError> {
+    let name = name.trim();
+    if name.is_empty() {
+        return Err(McpRequestError::QueryValidationFailed(
+            "Data product name cannot be empty".to_string(),
+        ));
+    }
+
+    let parsed = parse_item_name(name).map_err(|_| {
+        McpRequestError::QueryValidationFailed(format!(
+            "Invalid data product name: {}. Expected a valid object name, \
+             e.g. '\"database\".\"schema\".\"name\"' or 'my_view'",
+            name
+        ))
+    })?;
+
+    // Stable formatting forces all identifiers to be double-quoted,
+    // so SQL keywords and special characters cannot escape.
+    Ok(parsed.to_ast_string_stable())
+}
+
 /// Read rows from a data product. Issues a single read-only query.
 ///
 /// When `cluster_override` is provided, sets the cluster explicitly.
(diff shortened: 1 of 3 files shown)


src/environmentd/src/http/mcp.rs
```diff
@@ -683,6 +683,76 @@ async fn get_data_product_details(
...
+    let name = name.trim();
+    if name.is_empty() {
+        return Err(McpRequestError::QueryValidationFailed(
+            "Data product name cannot be empty".to_string(),
+        ));
+    }
+
+    let make_err = || {
+        McpRequestError::QueryValidationFailed(format!(
+            "Invalid data product name: {}. Expected a fully qualified name like '\"database\".\"schema\".\"name\"'",
```
ggevay: Why not allow an unqualified name? Actually, the test
```
assert!(validate_data_product_name("my_view").is_ok());
```
seems to indicate that we do allow unqualified names.

ggevay (commented):
I'm wondering, couldn't we instead use something like `StrExt::quoted`? The quotation should make any SQL keywords not be parsed as keywords, but parsed as part of the name. Also, `StrExt::quoted` makes sure to escape quotation marks, so the user can't break out of the quotation.

src/environmentd/src/http/mcp.rs
bobbyiliev: Ops yes, the error message was misleading! We do accept unqualified names. Updated the message to show both forms as examples.

bobbyiliev:
> I'm wondering, couldn't we instead use something like `StrExt::quoted`? The quotation should make any SQL keywords not be parsed as keywords, but parsed as part of the name. Also, `StrExt::quoted` makes sure to escape quotation marks, so the user can't break out of the quotation.

@ggevay I think `StrExt::quoted` might not work here because it escapes double quotes with backslash (`\"`) which is for display/logging, whereas SQL identifiers use doubled quotes (`""`). So I'm not sure it would be safe to use directly in SQL.

The other thing is the name arrives from `mz_mcp_data_products` as a pre-formatted multi-part identifier like `"materialize"."schema"."view"`, so we can't really wrap the whole thing in one set of quotes since it needs to be parsed as a qualified name.

The AST validation approach makes sure the name is safe to interpolate as-is without transforming it. But I might be missing something, do you have a different approach in mind?

ggevay:
Ah yes, `StrExt::quoted` is not good.

My problem with the PR's approach is that it is a lot of brittle code to maintain. E.g., when someone later adds to the AST, they'll have to not forget to update this code as well. This could be enforced somewhat by relying more on pattern matching, with no wildcards, so that one would get compile errors in `validate_data_product_name` if they forget to attend to `validate_data_product_name` when adding AST stuff.

But another alternative approach would be to call `parse_item_name` (making it `pub`) to get an `UnresolvedItemName` (which takes care of having or not having qualifications), and then when interpolating into the final SQL query (not a probe, but the real query) print it with an `AstFormatter` that has `FormatMode::Stable`, so that everything will be quoted (see `AstDisplay for Ident`), so that nasty things can't escape from it. This would be just a few lines of code, unless I'm missing something. Do you think this would be feasible?

bobbyiliev:
@ggevay ah nice suggestion, just implemented it, much cleaner! Added a public `parse_item_name` to `mz-sql-parser` (following the same pattern as `parse_expr`/`parse_data_type`), and replaced the AST inspection with:

```rust
fn safe_data_product_name(name: &str) -> Result<String, McpRequestError> {
    let parsed = parse_item_name(name).map_err(|_| ...)?;
    Ok(parsed.to_ast_string_stable())
}
```

Thanks for the suggestion and let me know what you think!

src/environmentd/src/http/mcp.rs
```diff
@@ -768,6 +769,38 @@ async fn get_data_product_details(
     format_rows_response(rows, max_response_size)
 }
 
+/// Validates that a data product name is a safe SQL identifier.
+///
+/// The name is interpolated into `SELECT * FROM {name}`, so we parse a
+/// synthetic query and inspect the AST to ensure nothing beyond a bare
```
ggevay: (stale comment)

ggevay (approved):
LGTM, thanks for iterating on it!

src/environmentd/src/http/mcp.rs
```diff
@@ -1498,46 +1529,40 @@ mod tests {
         }
     }
 
-    // ── Response size cap tests ────────────────────────────────────────
```
def-, line 1501: Were these tests removed intentionally or was it a bad rebase? @bobbyiliev

bobbyiliev: Yeah bad rebase on my end, sorry about that. Restored all three tests in https://github.com/MaterializeInc/materialize/pull/36157

PR #35786, opened 2026-03-30
[adapter][persist] PreAllocate Shards to remove from DDL path

mtabebe:
Motivation

A single `CREATE TABLE` currently makes ~27 CRDB round-trips, with
`open_data_handles` being the largest chunk. Two of those operations
`upgrade_version` and `open_critical_since` don't depend on the table
schema and can be done ahead of time. This PR pre-performs them in a
background pool so they can be skipped on the DDL critical path.

Solution:

Commit 1: Pre-opened shard pool

Adds `ShardPool<T>`, a thread-safe pool of pre-opened persist shards, and a
background replenishment task (`shard_pool_replenish_task`) that keeps the
pool filled to a configurable target size.

Commit 2: Catalog tracking for crash recovery

Without catalog tracking, shards pre-opened by the pool but never claimed
(e.g., due to a crash) are leaked permanently. This commit adds a
`pre_allocated_shards` catalog collection (same pattern as
`unfinalized_shards`) to track them.

### Results

Benchmarks show ~18% DDL latency improvement:

| Metric | Before | After | Saved |
|--------|--------|-------|-------|
| Per-DDL | 91.5ms | 76.9ms | 14.6ms |
| 500 tables | 50.7s | 41.2s | 9.4s |
| CRDB round-trips | 27 | 22 | 5 per DDL |

Rollout

The pool is controlled by `storage_shard_pool_enabled` (default: `true`).
It can be disabled at runtime with:

```sql
ALTER SYSTEM SET storage_shard_pool_enabled = false;
```

When disabled, DDL falls back to the existing `ShardId::new()` path with
no behavioral change.


mtabebe:
Nightly is clean except for CDC failures that arise from this new commit: https://github.com/MaterializeInc/materialize/pull/35513

mtabebe:
@bugbot run

bkirwi:
I don't totally follow why this approach is necessary, FWIW, mostly because it's not clear to me why we need to do anything synchronous with Persist in the DDL path.

One of the generated comments suggests that we're waiting on two roundtrips - upgrading the version and initializing the critical since. IIUC we shouldn't need to upgrade the version on a brand-new shard at all. Downgrading the since shouldn't really have to happen synchronously either, since it just allows compaction... though I'd totally believe that the structure of the storage controller makes this painful. In the long run it's probably worth putting some effort into cleaning up that aspect of the storage controller, though!

mtabebe:
> I don't totally follow why this approach is necessary, FWIW, mostly because it's not clear to me why we need to do anything synchronous with Persist in the DDL path.
> 
> One of the generated comments suggests that we're waiting on two roundtrips - upgrading the version and initializing the critical since. IIUC we shouldn't need to upgrade the version on a brand-new shard at all. Downgrading the since shouldn't really have to happen synchronously either, since it just allows compaction... though I'd totally believe that the structure of the storage controller makes this painful. In the long run it's probably worth putting some effort into cleaning up that aspect of the storage controller, though!

@bkirwi: (Maybe I am misunderstanding the code/your comment). But what I observed when profiling and looking at the code is that when we create a table is that we allocate a persist shard for that table. Specifically: `create_collections_for_bootstrap` is called which in turn calls `open_data_handles`. This function then does several subcalls that turn into CRDB operations: `upgrade_version`, `open_critical_since` and `open_writer`. So this PR aims to hide the latency of the first two calls but still keep the logic the same. (We can't remove open_writer, since it needs the actual RelationDesc schema).

I do agree with your points that we shouldn't need to upgrade the version on a brand new shard at all. So maybe we could just strip this out? The since handling is entangled since it is part of the collection state. 

So my intention with this change was to be pragmatic: leave things the way they are, just move it out of the actual critical path by doing things ahead of time.

bkirwi:
> But what I observed when profiling and looking at the code is that when we create a table is that we allocate a persist shard for that table. Specifically: create_collections_for_bootstrap is called [...]

This description of the flow sounds accurate to me! The thing I'm not sure about is why we'd need to call create_collections synchronously in the DDL path. Of course those handles do need to be created sooner or later, but the pattern we're using increasingly in eg. `apply_catalog_implications` lets us push that kind of work downstream of the actual write to the catalog. (So: getting it off the critical path by doing it after returning a successful response instead of pushing it before the request.) In theory the only thing we need to do to record a shard in the catalog is generate a new `ShardId` for it, and that takes no time at all, really...

> So my intention with this change was to be pragmatic: leave things the way they are, just move it out of the actual critical path by doing things ahead of time.

For sure. I think the way things are is not great - there's a general consensus I think that the storage controller is a bit of a mess - so I'm enthusiastic about taking the chance to unpick some of these weird flows instead of working around them. But there's a lot to be said for being pragmatic, and especially if there's some urgency here it makes sense why you'd take this path instead.

mtabebe:
That makes sense, Ben. I will think about what to do here. 

My intuition is the goal for the DDL for March was a spike on how to improve things. This approach does that. And also is evidence that we can further untangle things, since it isn't a specific requirement/goal from the plan. 

So, maybe I'll take a stab at untangling things as a Friday project?

bkirwi:
Whatever makes sense to you and the SQL team, yeah!

src/storage-client/src/storage_collections.rs
```diff
@@ -1823,15 +1917,56 @@ where
...
+                            let current_epoch: PersistEpoch = handle.opaque().decode();
+                            // CAS failure here means another environmentd instance
+                            // advanced the epoch; we'll halt shortly via the normal
+                            // epoch fencing path elsewhere. Safe to ignore.
+                            let _ = handle
+                                .compare_and_downgrade_since(
+                                    &Opaque::encode(&current_epoch),
+                                    (&Opaque::encode(&current_epoch), &joined),
+                                )
+                                .await;
```
def-, line 1946: Why don't we retry like in the slow path?

def- (changes requested):
With this test I'm seeing a panic pretty, by adding a targeted sleep at the failpoint to trigger it reliably:
```diff
diff --git a/src/adapter/src/catalog/open.rs b/src/adapter/src/catalog/open.rs
index 208e1e9d9c..544ca33614 100644
--- a/src/adapter/src/catalog/open.rs
+++ b/src/adapter/src/catalog/open.rs
@@ -669,6 +669,8 @@ impl Catalog {
             Some(id) => assert_eq!(*id, shard_id),
         }

+        fail::fail_point!("before_storage_initialize_state");
+
         storage_collections
             .initialize_state(&mut txn, collections)
             .await
diff --git a/test/persistence/mzcompose.py b/test/persistence/mzcompose.py
index 6f8a283007..2f5ab4ce6e 100644
--- a/test/persistence/mzcompose.py
+++ b/test/persistence/mzcompose.py
@@ -201,6 +201,40 @@ def workflow_shard_pool(c: Composition) -> None:
     c.rm_volumes("mzdata")


+def workflow_shard_pool_gc_race(c: Composition) -> None:
+    """Regression: initialize_state must not GC current-epoch pool shards."""
+
+    c.up("materialized")
+    c.run_testdrive_files("shard-pool-gc-race/before-restart.td")
+
+    c.kill("materialized")
+    with c.override(
+        Materialized(
+            environment_extra=[
+                "FAILPOINTS=before_storage_initialize_state=sleep(2000)"
+            ],
+        )
+    ):
+        c.up("materialized")
+        c.sql("CREATE TABLE gc_race_post_restart (a INT, b TEXT)")
+        c.sql(
+            "INSERT INTO gc_race_post_restart "
+            "VALUES (10, 'alpha'), (20, 'beta'), (30, 'gamma')"
+        )
+        time.sleep(20)
+        result = c.sql_query("SELECT count(*) FROM gc_race_post_restart")
+        assert int(result[0][0]) == 3
+
+    c.kill("materialized")
+    c.up("materialized")
+    result = c.sql_query("SELECT count(*) FROM gc_race_base")
+    assert int(result[0][0]) == 3
+
+    c.kill("materialized")
+    c.rm("materialized", "testdrive", destroy_volumes=True)
+    c.rm_volumes("mzdata")
+
+
 def workflow_inspect_shard(c: Composition) -> None:
     """Regression test for https://github.com/MaterializeInc/materialize/pull/21098"""
     c.up("materialized")
```
Running `bin/mzcompose --find persistence down && bin/mzcompose --find persistence run shard_pool_gc_race` panics:
```
thread 'tokio:work-4' panicked at src/txn-wal/src/txns.rs:229:22:
schema should be registered
   5: core::panicking::panic_fmt
   6: core::option::expect_failed
   7: <mz_txn_wal::txns::TxnsHandle<mz_storage_types::sources::SourceData, (), mz_repr::timestamp::Timestamp, i64, mz_storage_types::controller::TxnsCodecRow>>::register::<core::iter::adapters::map::Map<alloc::vec::into_iter::IntoIter<(mz_repr::global_id::GlobalId, mz_persist_client::write::WriteHandle<mz_storage_types::sources::SourceData, (), mz_repr::timestamp::Timestamp, i64>)>, <mz_storage_controller::persist_handles::TxnsTableWorker<mz_repr::timestamp::Timestamp>>::register::{closure#0}::{closure#2}>>::{closure#0}::{closure#0}
   8: <mz_storage_controller::persist_handles::TxnsTableWorker<mz_repr::timestamp::Timestamp>>::register::{closure#0}
   9: <tracing::instrument::Instrumented<<mz_storage_controller::persist_handles::TxnsTableWorker<mz_repr::timestamp::Timestamp>>::register::{closure#0}> as core::future::future::Future>::poll
  10: <core::pin::Pin<alloc::boxed::Box<<mz_storage_controller::persist_handles::PersistTableWriteWorker<mz_repr::timestamp::Timestamp>>::new_txns::{closure#1}>> as core::future::future::Future>::poll
  11: <tracing::instrument::Instrumented<core::pin::Pin<alloc::boxed::Box<<mz_storage_controller::persist_handles::PersistTableWriteWorker<mz_repr::timestamp::Timestamp>>::new_txns::{closure#1}>>> as core::future::future::Future>::poll
  12: <tokio::runtime::task::core::Core<tracing::instrument::Instrumented<core::pin::Pin<alloc::boxed::Box<<mz_storage_controller::persist_handles::PersistTableWriteWorker<mz_repr::timestamp::Timestamp>>::new_txns::{closure#1}>>>, alloc::sync::Arc<tokio::runtime::scheduler::multi_thread::handle::Handle>>>::poll
  13: <tokio::runtime::task::harness::Harness<tracing::instrument::Instrumented<core::pin::Pin<alloc::boxed::Box<<mz_storage_controller::persist_handles::PersistTableWriteWorker<mz_repr::timestamp::Timestamp>>::new_txns::{closure#1}>>>, alloc::sync::Arc<tokio::runtime::scheduler::multi_thread::handle::Handle>>>::poll
  14: <tokio::runtime::scheduler::multi_thread::worker::Context>::run_task
  15: <tokio::runtime::context::scoped::Scoped<tokio::runtime::scheduler::Context>>::set::<tokio::runtime::scheduler::multi_thread::worker::run::{closure#0}::{closure#0}, ()>
  16: tokio::runtime::context::runtime::enter_runtime::<tokio::runtime::scheduler::multi_thread::worker::run::{closure#0}, ()>
  17: tokio::runtime::scheduler::multi_thread::worker::run
  18: <tracing::instrument::Instrumented<tokio::runtime::blocking::task::BlockingTask<<tokio::runtime::scheduler::multi_thread::worker::Launch>::launch::{closure#0}>> as core::future::future::Future>::poll
  19: <tokio::runtime::task::core::Core<tracing::instrument::Instrumented<tokio::runtime::blocking::task::BlockingTask<<tokio::runtime::scheduler::multi_thread::worker::Launch>::launch::{closure#0}>>, tokio::runtime::blocking::schedule::BlockingSchedule>>::poll
  20: <tokio::runtime::task::harness::Harness<tracing::instrument::Instrumented<tokio::runtime::blocking::task::BlockingTask<<tokio::runtime::scheduler::multi_thread::worker::Launch>::launch::{closure#0}>>, tokio::runtime::blocking::schedule::BlockingSchedule>>::poll
  21: <tokio::runtime::task::UnownedTask<tokio::runtime::blocking::schedule::BlockingSchedule>>::run
  22: <tokio::runtime::blocking::pool::Inner>::run
```

PR #35803, opened 2026-03-31
adapter: Add TLA+ formal verification of timestamp selection

antiguru:
## Summary

- Adds a TLA+ specification modeling the constraint solver in `determine_timestamp_via_constraints` over a bounded timestamp domain (0..4), living next to the Rust source at `src/adapter/src/coord/TimestampSelection.tla`
- Covers all three isolation levels (Serializable, StrictSerializable, StrongSessionSerializable) and all `QueryWhen` variants
- TLC exhaustively verifies 750k input combinations checking: lower/upper bound respect, real-time ordering, completeness, soundness, and preference correctness
- Generic mzcompose test runner at `test/tla/` auto-discovers all `*.tla` files with matching `*.cfg` across the repo, so adding future specs requires no test harness changes
- Adds a cross-reference comment in `timestamp_selection.rs` pointing to the spec
- CI step in the Lints group triggers on any `*.tla` or `*.cfg` change

## Test plan

- [x] `bin/mzcompose --find tla run default` passes locally (750k states, 9s, all 7 properties verified)
- [x] Auto-discovery correctly finds `src/adapter/src/coord/TimestampSelection.tla`

### DIFF
--- a/src/adapter/src/coord/TimestampSelection.tla
+++ b/src/adapter/src/coord/TimestampSelection.tla
@@ -1,0 +1,375 @@
+\* Copyright Materialize, Inc. and contributors. All rights reserved.
+\*
+\* Use of this software is governed by the Business Source License
+\* included in the LICENSE file at the root of this repository.
+\*
+\* As of the Change Date specified in that file, in accordance with
+\* the Business Source License, use of this software will be governed
+\* by the Apache License, Version 2.0.
+
+--------------------------- MODULE TimestampSelection ---------------------------
+\* Formal specification of the timestamp selection constraint solver in
+\* Materialize's adapter layer (src/adapter/src/coord/timestamp_selection.rs).
+\*
+\* This models `determine_timestamp_via_constraints` as a pure function over
+\* a bounded timestamp domain. The spec covers constraint generation for each
+\* isolation level and QueryWhen variant, the solver logic, and the preference
+\* (FreshestAvailable vs StalestValid) that breaks ties.
+\*
+\* Properties verified:
+\*   1. Result respects all lower bounds (>= join of lower bounds)
+\*   2. Result respects all upper bounds (<= meet of upper bounds, if any)
+\*   3. Real-time ordering: result >= oracle_read_ts when oracle is a constraint
+\*   4. Completeness: non-empty valid interval => solver succeeds
+\*   5. Soundness: empty valid interval => solver returns error
+\*   6. FreshestAvailable picks the largest valid timestamp <= upper
+\*   7. StalestValid picks the smallest valid timestamp
+\*
+\* We use a small bounded domain (0..MaxTs) so TLC can exhaustively enumerate
+\* all input combinations.
+
+EXTENDS Integers, FiniteSets, Sequences
+
+CONSTANTS
+    MaxTs       \* Upper bound of the timestamp domain, e.g. 4
+
+\* --------------------------------------------------------------------------
+\* Timestamp domain
+\* --------------------------------------------------------------------------
+
+\* The set of all timestamps in our bounded domain.
+Timestamps == 0..MaxTs
+
+\* We model antichains as single-element sets (totally ordered timestamps).
+\* The empty antichain {} represents the "closed" / unreachable frontier.
+\* A singleton {t} represents the frontier at t.
+\* This is a faithful simplification: Materialize timestamps are u64, totally ordered.
+
+\* --------------------------------------------------------------------------
+\* Enumerations
+\* --------------------------------------------------------------------------
+
+IsolationLevels == {"Serializable", "StrictSerializable", "StrongSessionSerializable"}
+
+\* QueryWhen variants. AtTimestamp and AtLeastTimestamp carry a timestamp parameter,
+\* modeled separately via the `when_ts` variable.
+QueryWhens == {"Immediately", "FreshestTableWrite", "AtTimestamp", "AtLeastTimestamp"}
+
+Preferences == {"FreshestAvailable", "StalestValid"}
+
+\* --------------------------------------------------------------------------
+\* Helper operators
+\* --------------------------------------------------------------------------
+
+\* The maximum of two timestamps.
+Max(a, b) == IF a >= b THEN a ELSE b
+
+\* The minimum of two timestamps.
+Min(a, b) == IF a <= b THEN a ELSE b
+
+\* Step back: the largest timestamp strictly less than t, saturating at 0.
+StepBack(t) == IF t > 0 THEN t - 1 ELSE 0
+
+\* largest_not_in_advance_of_upper: given the write frontier (upper),
+\* compute the largest readable timestamp.
+\* If upper is empty (closed collection), return MaxTs.
+\* If upper is {t}, return StepBack(t).
+LargestNotInAdvanceOfUpper(upper) ==
+    IF upper = {}
+    THEN MaxTs
+    ELSE LET t == CHOOSE x \in upper : TRUE
+         IN  StepBack(t)
+
+\* --------------------------------------------------------------------------
+\* QueryWhen predicates (mirroring the Rust methods)
+\* --------------------------------------------------------------------------
+
+AdvanceToTimestamp(when, when_ts) ==
+    IF when \in {"AtTimestamp", "AtLeastTimestamp"}
+    THEN when_ts
+    ELSE -1  \* sentinel: no advance
+
+ConstrainsUpper(when) ==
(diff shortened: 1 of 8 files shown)


antiguru:
## Prompt for automated spec-implementation sync check

This prompt can be used with the Claude API to verify that the TLA+ spec and Rust implementation remain semantically in sync. It should be run whenever either file changes.

---

You are reviewing whether a TLA+ specification and its corresponding Rust implementation are semantically in sync.

The TLA+ spec models the constraint solver in `determine_timestamp_via_constraints`. It is not expected to be a line-by-line transliteration of the Rust code. Differences in naming, formatting, code structure, comments, helper factoring, error message text, and display/serialization logic are all expected and are NOT issues.

Focus exclusively on **semantic differences** in the core logic:

1. **Constraint generation**: For each isolation level (`Serializable`, `StrictSerializable`, `StrongSessionSerializable`) and each `QueryWhen` variant (`Immediately`, `FreshestTableWrite`, `AtTimestamp`, `AtLeastTimestamp`), do the same lower-bound and upper-bound constraints get added in both the spec and the implementation?

2. **Preference selection**: Is the choice between `FreshestAvailable` and `StalestValid` governed by the same conditions?

3. **Solver logic**: Given the same lower bound, upper bound, and preference, does the solver compute the same candidate timestamp? Does it accept/reject under the same conditions?

4. **Preconditions**: Are the input validity constraints (e.g., `real_time_recency_ts` only present under `StrictSerializable`) consistent?

Ignore everything outside the constraint solver: the `TimestampContext` construction, `TimestampDetermination` struct, display formatting, metrics, error types, and the `determine_timestamp_for` wrapper.

If the spec and implementation are in sync, respond with exactly: `IN SYNC`

If they have diverged, respond with:
```
OUT OF SYNC

<description of each semantic difference>
```

Here are the two files to compare:

```
<tla_spec>
{contents of test/tla/TimestampSelection.tla}
</tla_spec>

<rust_implementation>
{contents of src/adapter/src/coord/timestamp_selection.rs}
</rust_implementation>
```

src/adapter/src/coord/timestamp_selection.rs
```diff
@@ -242,6 +242,10 @@ pub trait TimestampProvider {
     /// Uses constraints and preferences to determine a timestamp for a query.
     /// Returns the determined timestamp, the constraints that were applied, and
     /// session_oracle_read_ts.
+    ///
+    /// A TLA+ specification of this function's logic lives in
```
def-, line 246: How do we ensure the code and specification stay in sync? If we don't have that I don't understand why we need to run it as a test in CI.

antiguru: Yeah, it's the big problem of formally verified systems: If the programming language has no hooks, it's ultimately a social contract. No better answer unfortunately. Putting it in the test pipeline is more temporary to see that it actually works.

The better solution I believe is have a LLM check that the two don't diverge, like the [prompt](https://github.com/MaterializeInc/materialize/pull/35803#issuecomment-4168212322) I added. We could run it regularly and it'd tell us if the implementation and spec diverge with some probability.

def-: One downside is that in the test pipeline we currently run all checks on main, which seems a bit wasteful for this particular one, since it seems unlikely to find flakes. For the console tests we have an exception to that, by adding `ci_glue_exempt: true` in the pipeline.template.yml

def-: I did once [work on something](https://formal.kastel.kit.edu/~klebanov/pubs/ase2014.pdf) to solve issues like that btw, not that it's particularly useful for this use case:
> For regression verification, one does not need to write and maintain complex specifications (which can be a significant bottleneck in the verification process).

src/adapter/src/coord/TimestampSelection.tla
```diff
@@ -0,0 +1,375 @@
...
+    \* Compute the result
+    /\ result = DetermineTimestamp(since, upper, when, when_ts,
+                                    oracle_read_ts, has_oracle,
+                                    real_time_recency_ts, has_rtr,
+                                    isolation_level,
+                                    session_oracle_ts, has_session_oracle)
+
+\* No state transitions - this is a single-state spec for checking properties
+\* over all input combinations.
+Next == UNCHANGED vars
```
teskje, line 297: So what this does is generate all input combinations, run `determine_timestamp` on it, and then confirm the invariants, right? Is this something we could do on the implementation instead? Seems like the only hard part is enumerating all the input combinations, you'd need something like proptest, but exhaustive. Seems plausible to write a bunch of generators by hand and then a nested loop that computes the cross product.

src/adapter/src/coord/timestamp_selection.rs
mtabebe: Is there anything we can add to the claude.md or something that will keep it as almost a rule/lint?

E.g.,
```
## TLA+ Specs
Some critical algorithms have accompanying TLA+ formal specifications that live next to the Rust source. When modifying the logic of a function that has a TLA+ spec, you **must** update the spec to reflect the change. Look for comments in the Rust code referencing a `.tla` file. Known spec-code pairs:
`src/adapter/src/coord/timestamp_selection.rs` (`determine_timestamp_via_constraints`) ↔ `src/adapter/src/coord/TimestampSelection.tla`
```
this would only apply to agents making changes though.

src/adapter/src/coord/TimestampSelection.tla
```diff
@@ -0,0 +1,375 @@
...
+        largest_niao == LargestNotInAdvanceOfUpper(upper)
+
+        lower_bound == ComputeLowerBound(since, when, when_ts, oracle_read_ts, has_oracle,
+                                         real_time_recency_ts, has_rtr,
+                                         isolation_level, session_oracle_ts, has_session_oracle,
+                                         largest_niao)
+
+        upper_bound == ComputeUpperBound(when, when_ts)
+
+        \* We model has_timeline as TRUE for simplicity (the common case).
```
mtabebe, line 229: Why? My reading of the rust code is that there is a case for serializable with no timeline?

src/adapter/src/coord/TimestampSelection.tla
```diff
@@ -0,0 +1,375 @@
...
+    when \in {"Immediately", "FreshestTableWrite"}
+
+MustAdvanceToTimelineTs(when) ==
+    when = "FreshestTableWrite"
+
+\* --------------------------------------------------------------------------
+\* needs_linearized_read_ts (determines whether oracle_read_ts is present)
+\* --------------------------------------------------------------------------
+
+NeedsLinearizedReadTs(isolation_level, when) ==
```
mtabebe, line 108: Is this used?

src/adapter/src/coord/TimestampSelection.tla
```diff
@@ -0,0 +1,375 @@
...
+\*   3. Real-time ordering: result >= oracle_read_ts when oracle is a constraint
+\*   4. Completeness: non-empty valid interval => solver succeeds
+\*   5. Soundness: empty valid interval => solver returns error
+\*   6. FreshestAvailable picks the largest valid timestamp <= upper
+\*   7. StalestValid picks the smallest valid timestamp
+\*
+\* We use a small bounded domain (0..MaxTs) so TLC can exhaustively enumerate
+\* all input combinations.
+
+EXTENDS Integers, FiniteSets, Sequences
```
mtabebe, line 31: I don't know enough TLA but are these used?

mtabebe (commented):
This is super cool, I spent some time looking about both the rust code and TLA code, mostly to learn. I didn't look at the mzcompose stuff or the CI stuff

src/adapter/src/coord/timestamp_selection.rs
mgree: In other projects, I've used comments and a custom "static analysis" (a shell script 🤪 ) to keep things in sync. [Here's the example.](https://github.com/binpash/try/blob/main/scripts/check_trycase.sh)

Another possibility is putting comments or assertions in the Rust that we extract to form the TLA+ model. (But I haven't looked at the code here to see how reasonable that is.)

src/adapter/src/coord/timestamp_selection.rs
```diff
@@ -242,6 +242,10 @@ pub trait TimestampProvider {
...
+    /// `test/tla/TimestampSelection.tla`. If you change the constraint
```
def-, line 247:
```suggestion
    /// `TimestampSelection.tla`. If you change the constraint
```

PR #35825, opened 2026-04-01
Analyze materialized view sink performance issues

antiguru:
https://claude.ai/code/session_016VWvt1mh7GR6SHHFVT79D3

### DIFF
--- a/doc/developer/design/20260325_mv_sink_correction_v3.md
+++ b/doc/developer/design/20260325_mv_sink_correction_v3.md
@@ -1,0 +1,411 @@
+# MV Sink Correction Buffer V3
+
+## The Problem
+
+The MV sink's `CorrectionV2` buffer (in `src/compute/src/sink/correction_v2.rs`) maintains
+chains of chunks backed by columnation (`TimelyStack`). Three interrelated performance
+problems limit its scalability:
+
+**P1: Time advancement during merges.** Consolidation requires advancing all times to the
+`since` frontier so that updates at different original times can cancel. The current
+`advance_by` method handles this by splitting each cursor into one sub-cursor per distinct
+time ≤ `since`. With K chains containing T distinct timestamps each, this produces up to
+K×T cursors, making merges O(N log(K×T)) instead of O(N log K).
+
+**P2: Chunks are too small.** `TimelyStack` targets `BUFFER_SIZE_BYTES = 8192` bytes for
+its index vector (from `timely_container`). For `(Row, Timestamp, Diff)` at ~48 byte index
+entries, each chunk holds ~170 elements. This leads to many small allocations, excessive
+chunk management overhead, and poor lgalloc utilization. Target: ~2 MiB per chunk.
+
+**P3: Many timestamps in flight.** MVs downstream of temporal filters can have corrections
+at thousands of distinct future times. Because `advance_by` splits by time during every
+merge—including merges triggered by `insert_inner` to restore the chain invariant—a steady
+stream of updates at different timestamps creates pathological merge cascades.
+
+## Success Criteria
+
+- Merge operations never perform time-based cursor splitting (eliminates P1/P3).
+- Chunks target ~2 MiB contiguous allocations with lgalloc/spill-to-disk support (fixes P2).
+- Many timestamps in flight are handled in O(log(time_range)) buckets, not O(distinct_timestamps) chains.
+- The existing V2 implementation remains available as a fallback, controlled by a feature flag.
+- Row-typed corrections use the new design; DataflowError corrections reuse V2.
+
+## Out of Scope
+
+- Implementing `Columnar` for `DataflowError` (complex nested enum with hand-written `Columnation`).
+- Changes to the `mint` or `append` operators. Only the `write` operator's correction buffer changes.
+- Multi-dimensional timestamp support. The `BucketChain` requires a total order, which the MV sink uses.
+
+## Solution Proposal
+
+### Architecture Overview
+
+```
+CorrectionV3<Row> {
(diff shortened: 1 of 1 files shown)


doc/developer/design/20260325_mv_sink_correction_v3.md
```diff
@@ -0,0 +1,411 @@
...
+
+**P2: Chunks are too small.** `TimelyStack` targets `BUFFER_SIZE_BYTES = 8192` bytes for
+its index vector (from `timely_container`). For `(Row, Timestamp, Diff)` at ~48 byte index
+entries, each chunk holds ~170 elements. This leads to many small allocations, excessive
+chunk management overhead, and poor lgalloc utilization. Target: ~2 MiB per chunk.
+
+**P3: Many timestamps in flight.** MVs downstream of temporal filters can have corrections
+at thousands of distinct future times. Because `advance_by` splits by time during every
+merge—including merges triggered by `insert_inner` to restore the chain invariant—a steady
+stream of updates at different timestamps creates pathological merge cascades.
```
antiguru, line 23: The reason is not correct, but the effect is. How this happens is when we hydrate catching up after downtime, for example a cluster with zero replicas for a while needs to work through the backlog.

doc/developer/design/20260325_mv_sink_correction_v3.md
```diff
@@ -0,0 +1,411 @@
...
+```rust
+pub trait DataV3: differential_dataflow::Data + Columnar {}
+```
+
+Since `DataflowError` does not implement `Columnar`, V3 is only instantiated for `Row`.
+
+### DataflowError Handling
+
+`DataflowError` corrections continue to use `CorrectionV2`. The write operator maintains
+two separate correction buffers:
```
antiguru, line 259: This can be fixed: We have a working draft for enabling Columnar for `DataflowError`. I didn't have that context when creating this document.

doc/developer/design/20260325_mv_sink_correction_v3.md
teskje: Right, so it's the same as P1, right? The mentioning of temporal filters is misleading, I think, as the correction buffer handles future updates quite well (compared to, e.g., indexes).

doc/developer/design/20260325_mv_sink_correction_v3.md
```diff
@@ -0,0 +1,411 @@
...
+3. **Columnar `Column<C>`** replaces `TimelyStack`-based chunks, giving 2 MiB contiguous
+   allocations with lgalloc/spill-to-disk support.
+
+### Two-Group Architecture
+
+The `since` frontier divides corrections into two fundamentally different groups:
+
+- **Compacted** (times ≤ since): All times collapse to `since_ts` after advancement. The
+  meaningful ordering is **by data only**—time is irrelevant.
+- **Future** (times > since): Times are unaffected by advancement. Standard **(time, data)
```
teskje, line 66: This is not true, right? When the since advances, some of the future times might move from beyond the since to before the since.

doc/developer/design/20260325_mv_sink_correction_v3.md
```diff
@@ -0,0 +1,411 @@
...
+
+**Why data-only ordering works:** All times ≤ since collapse to `since_ts`. Two updates
+`(d, t1, +1)` and `(d, t2, -1)` where both `t1, t2 ≤ since` represent the same logical
+datum at the same effective time. Sorting by data makes them adjacent, so they cancel
+during consolidation.
+
+**Timestamp handling:** Timestamps are stored in chunks but ignored for ordering and
+comparison during merges. When two entries have the same data, their diffs are summed
+regardless of stored timestamp. On output, all entries are yielded as
+`(data, since_ts, diff)`, stamped with the current `since_ts`.
```
teskje, line 86: Why do we need to store timestamps at all? We know that all timestamps are equal to the since, so can we simply omit them and save a bunch of space?

doc/developer/design/20260325_mv_sink_correction_v3.md
```diff
@@ -0,0 +1,411 @@
...
+
+### The Future Group: `BucketChain`
+
+Contains all corrections at times > `since_ts`. Uses `BucketChain<Storage<Row>>` from
+`src/timely-util/src/temporal.rs`.
+
+**Why `BucketChain` instead of chain-of-chains:** V2's chain-of-chains is ordered by
+`(time, data)` globally. When many timestamps are in flight, `advance_by` must split at
+every distinct time during every merge. `BucketChain` avoids this entirely by bucketing
+updates into power-of-two time ranges. Each bucket covers `[offset, offset + 2^bits)`.
```
teskje, line 107: This reasoning doesn't check out for me! `advance_by` doesn't look at times beyond the since, it only splits at times before the since.

doc/developer/design/20260325_mv_sink_correction_v3.md
```diff
@@ -0,0 +1,411 @@
...
+
+Key properties of `BucketChain`:
+- **Insertion:** `find_mut(timestamp)` locates the bucket in O(log B). Push the update into the bucket's internal storage.
+- **Peel:** `peel(frontier)` extracts all buckets strictly before the frontier, splitting boundary buckets as needed.
+- **Restore:** `restore(&mut fuel)` rebalances bucket sizes with fuel-based amortization. The chain tolerates being temporarily un-well-formed.
+- **Scalability:** 1000 distinct timestamps in a range of width 1024 need ~10 buckets (log₂), not 1000 chains.
```
teskje, line 113: Why would it need 1000 chains? A chain can have updates with different times.

PR #35826, opened 2026-04-01
optimizer: Push `COALESCE` into `CASE WHEN`

mgree:
### Motivation

https://materializeinc.slack.com/archives/C0246GEHL8N/p1774867717130349?thread_ts=1774866513.092989&cid=C0246GEHL8N

### Description

Implements the optimization: 
```
COALESCE((CASE WHEN e1 THEN e2 ELSE e3), ...)
=>
CASE WHEN e1 THEN COALESCE(e2, ...) ELSE COALESCE(e3, ...)
```
In order to not pile onto `MSE::reduce`, I implemented this as an optimization that runs in the fixpoint before join implementation.

(Before join implementation is important: https://github.com/MaterializeInc/database-issues/issues/11300 it's not clear what updates are possible once we've settled on arrangements, and others get this wrong [and should be fixed].)

### Verification

SLT updates, spec tests.

### DIFF
--- a/src/sql/src/session/vars/definitions.rs
+++ b/src/sql/src/session/vars/definitions.rs
@@ -2268,6 +2268,12 @@ feature_flags!(
         default: true,
         enable_for_item_parsing: false,
     },
+    {
+        name: enable_coalesce_case_transform,
+        desc: "Allow the optimizer to `COALESCE` of a `NULL`-returning `CASE` into a single `CASE`.",
+        default: false,
+        enable_for_item_parsing: false,
+    },
 );
 
 impl From<&super::SystemVars> for OptimizerFeatures {
@@ -2293,6 +2299,7 @@ impl From<&super::SystemVars> for OptimizerFeatures {
             enable_cast_elimination: vars.enable_cast_elimination(),
             enable_case_literal_transform: vars.enable_case_literal_transform(),
             enable_simplify_quantified_comparisons: vars.enable_simplify_quantified_comparisons(),
+            enable_coalesce_case_transform: vars.enable_coalesce_case_transform(),
         }
     }
 }
@@ -2335,6 +2342,7 @@ mod tests {
             enable_cast_elimination,
             enable_case_literal_transform,
             enable_simplify_quantified_comparisons,
+            enable_coalesce_case_transform,
         } = false_features;
 
         let mut vars = SystemVars::new();
@@ -2365,6 +2373,7 @@ mod tests {
         set_var!(enable_cast_elimination);
         set_var!(enable_case_literal_transform);
         set_var!(enable_simplify_quantified_comparisons);
+        set_var!(enable_coalesce_case_transform);
 
         // Enable for item parsing, then ensure we still get the same optimizer features.
         vars.enable_for_item_parsing();
--- a/src/sql/src/session/vars/definitions.rs
+++ b/src/sql/src/session/vars/definitions.rs
@@ -2270,7 +2270,7 @@ feature_flags!(
     },
     {
         name: enable_coalesce_case_transform,
-        desc: "Allow the optimizer to `COALESCE` of a `NULL`-returning `CASE` into a single `CASE`.",
+        desc: "Allow the optimizer to push `COALESCE` into `CASE WHEN`.",
         default: false,
         enable_for_item_parsing: false,
     },
--- a/src/sql/src/session/vars/definitions.rs
+++ b/src/sql/src/session/vars/definitions.rs
@@ -2271,7 +2271,7 @@ feature_flags!(
     {
         name: enable_coalesce_case_transform,
         desc: "Allow the optimizer to push `COALESCE` into `CASE WHEN`.",
-        default: false,
+        default: true,
         enable_for_item_parsing: false,
     },
 );
--- a/src/transform/src/coalesce_case.rs
+++ b/src/transform/src/coalesce_case.rs
@@ -139,10 +139,7 @@ impl CoalesceCase {
                             .chain(exprs.iter().cloned())
                             .collect(),
                     );
-                    let t = MirScalarExpr::call_variadic(
-                        VariadicFunc::Coalesce(Coalesce),
-                        exprs,
-                    );
+                    let t = MirScalarExpr::call_variadic(VariadicFunc::Coalesce(Coalesce), exprs);
                     *expr = cond.take().if_then_else(t, f);
                 } else if els.is_literal_null() {
                     let MirScalarExpr::If {
--- a/src/transform/src/coalesce_case.rs
+++ b/src/transform/src/coalesce_case.rs
@@ -1,0 +1,163 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! Turn `COALESCE` of `NULL`-returning `CASE` into just a `CASE`
+
+use mz_expr::func::variadic::Coalesce;
+use mz_expr::visit::Visit;
+use mz_expr::{AggregateExpr, JoinImplementation, MirRelationExpr, MirScalarExpr, VariadicFunc};
+use mz_ore::soft_assert_eq_or_log;
+
+use crate::{TransformCtx, TransformError};
+
+/// Replace operators on constant collections with constant collections.
+#[derive(Debug)]
+pub struct CoalesceCase {}
+
+impl Default for CoalesceCase {
+    fn default() -> Self {
+        Self {}
+    }
+}
+
+impl crate::Transform for CoalesceCase {
+    fn name(&self) -> &'static str {
+        "FoldConstants"
+    }
+
+    #[mz_ore::instrument(
+        target = "optimizer",
+        level = "debug",
+        fields(path.segment = "fold_constants")
+    )]
+    fn actually_perform_transform(
+        &self,
+        relation: &mut MirRelationExpr,
+        _: &mut TransformCtx,
+    ) -> Result<(), TransformError> {
+        relation.try_visit_mut_post(&mut |e| self.action(e))?;
+        mz_repr::explain::trace_plan(&*relation);
+        Ok(())
+    }
+}
+
+impl CoalesceCase {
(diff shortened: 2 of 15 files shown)


antiguru:
> MSE::reduce is too complicated, and now I'm part of the problem.

You could write it as an optimizer transform, like the `CaseLiteral` transform I recently added. More code to write, but at least free-standing and not part of a match with unclear/undocumented semantics.

src/transform/src/coalesce_case.rs
```diff
@@ -0,0 +1,163 @@
...
+//! Turn `COALESCE` of `NULL`-returning `CASE` into just a `CASE`
+
+use mz_expr::func::variadic::Coalesce;
+use mz_expr::visit::Visit;
+use mz_expr::{AggregateExpr, JoinImplementation, MirRelationExpr, MirScalarExpr, VariadicFunc};
+use mz_ore::soft_assert_eq_or_log;
+
+use crate::{TransformCtx, TransformError};
+
+/// Replace operators on constant collections with constant collections.
```
antiguru: Stale comment.

src/transform/src/coalesce_case.rs
```diff
@@ -0,0 +1,163 @@
...
+
+impl crate::Transform for CoalesceCase {
+    fn name(&self) -> &'static str {
+        "FoldConstants"
+    }
+
+    #[mz_ore::instrument(
+        target = "optimizer",
+        level = "debug",
+        fields(path.segment = "fold_constants")
```
antiguru: Stale segment.

antiguru (commented):
I think there's a subtle bug:

```
  COALESCE(CASE WHEN x > 0 THEN NULL ELSE y END, 42)
  -- where y can be NULL

  Original behavior (x <= 0, y IS NULL):
  COALESCE(NULL, 42) → 42

  After rewrite:
  CASE WHEN x > 0 THEN COALESCE(42) ELSE y END → NULL
```

mgree:
> I think there's a subtle bug:
> 
> ```
>   COALESCE(CASE WHEN x > 0 THEN NULL ELSE y END, 42)
>   -- where y can be NULL
> 
>   Original behavior (x <= 0, y IS NULL):
>   COALESCE(NULL, 42) → 42
> 
>   After rewrite:
>   CASE WHEN x > 0 THEN COALESCE(42) ELSE y END → NULL
> ```

Wow, good catch, thank you! Fixed now, with spec tests and an SLT test to (a) catch this issue and (b) ensure we're cleaning up `COALESCE` calls we leave behind that can be constant folded.

src/transform/src/coalesce_case.rs
```diff
@@ -0,0 +1,166 @@
...
+            MirRelationExpr::Filter { predicates: exprs, .. } |
+            MirRelationExpr::FlatMap { exprs, .. } => {
+                // NB TableFunc doesn't ever hold an MSE
+                for expr in exprs.iter_mut() {
+                    self.rewrite_scalar(expr)?;
+                }
+            }
+            MirRelationExpr::Join { equivalences, implementation, .. } => {
+                if implementation.is_implemented() {
+                    soft_assert_eq_or_log!(*implementation, JoinImplementation::Unimplemented, "unexpected implemented Join when optimizing coalesce/case, skipping");
```
def-, line 74:
```suggestion
                    mz_ore::soft_panic_or_log!("unexpected implemented Join when optimizing coalesce/case, skipping: {:?}", implementation);
```

src/sql/src/session/vars/definitions.rs
```diff
@@ -2268,6 +2268,12 @@ feature_flags!(
         default: true,
         enable_for_item_parsing: false,
     },
+    {
+        name: enable_coalesce_case_transform,
+        desc: "Allow the optimizer to `COALESCE` of a `NULL`-returning `CASE` into a single `CASE`.",
```
ggevay: A word seems to be missing. Maybe something like:
"Allow the optimizer to turn a `COALESCE` of a `NULL`-returning `CASE` into a single `CASE`."

src/transform/src/coalesce_case.rs
```diff
@@ -0,0 +1,166 @@
...
+    ) -> Result<(), TransformError> {
+        relation.try_visit_mut_post(&mut |e| self.action(e))?;
+        mz_repr::explain::trace_plan(&*relation);
+        Ok(())
+    }
+}
+
+impl CoalesceCase {
+    fn action(&self, relation: &mut MirRelationExpr) -> Result<(), TransformError> {
+        match relation {
```
ggevay, line 52: We'll now have 3 transforms that mainly work with only the scalar expressions of MIR: `ReduceScalars`, `CaseLiteral`, `CoalesceCase`, and all 3 of them re-implement the delicate logic of finding all the scalar expressions in an `MirRelationExpr`. Note that we have `visit_scalars_mut`, which promises to do exactly this, but somehow all 3 of the transforms require various minor bits of extra logic that makes `visit_scalars_mut` insufficient. I'm not really sure what to do about this, but thought I mention it.

ggevay: Btw., what was the problem with centralizing all the scalar expression optimizations in `MSE::reduce`? Is it that it was meant originally only for constant folding, but then various other optimizations crept in? Random idea, but how about having 2 of these functions:
- one that does just constant folding
- and one that does all the other scalar expr optimizations.

Edit: Although, `CaseLiteral` seems like it should be only in the physical optimizer phase, so that seems like a valid reason to not have it together with all the other scalar optimizations.

test/sqllogictest/ldbc_bi.slt
```diff
@@ -874,7 +874,7 @@ Explained Query:
             Get l0 // { arity: 3 }
     Return // { arity: 5 }
       Map (((bigint_to_numeric((1 * #3{count})) + (2 * #1{sum})) + (10 * #2{sum}))) // { arity: 5 }
-        Reduce group_by=[#0{creatorpersonid}] aggregates=[sum(coalesce(case when (#2) IS NULL then null else #1{count} end, 0)), sum(coalesce(case when (#4) IS NULL then null else #3{count} end, 0)), count(*)] // { arity: 4 }
+        Reduce group_by=[#0{creatorpersonid}] aggregates=[sum(case when (#2) IS NULL then 0 else coalesce(#1{count}, 0) end), sum(case when (#4) IS NULL then 0 else coalesce(#3{count}, 0) end), count(*)] // { arity: 4 }
```
ggevay, line 877: Interesting that this also happens in LDBC!

Btw. when the `coalesce` can't be totally eliminated, then the speedup is from executing it less often, i.e., only in one branch?

src/transform/src/lib.rs
```diff
@@ -845,6 +846,7 @@ impl Optimizer {
                 transforms: vec![
                     Box::new(EquivalencePropagation::default()),
                     Box::new(fold_constants_fixpoint(true)),
+                    Box::new(coalesce_case::CoalesceCase::default()),
```
ggevay: I'm wondering why this is considered a physical optimization, and not a logical one.

E.g., for `CaseLiteralTransform`, it clearly seems physical, because it introduces a special, low-level implementation for certain `case` expressions. This would make it problematic to apply earlier, because other things that want to look at `case` expressions wouldn't recognize them.

But `CoalesceCase` seems to do quite a friendly thing to expressions, and wouldn't bother other transforms, as far as I can tell.

(Tbh., I also don't see any specific advantage of moving it earlier into `logical_optimizer`, but still, I guess we should strive for as much clarity as possible on where to put transforms, so thought I bring this up.)

src/sql/src/session/vars/definitions.rs
ggevay: Also, the `coalesce` sometimes doesn't completely disappear, as shown in the LDBC example. So, maybe the doc comment of the `CoalesceCase` struct is more accurate? "Push `COALESCE` into `CASE WHEN` as possible"

test/sqllogictest/ldbc_bi.slt
mgree: Yes, that's right. The goal is to reveal opportunities for further optimization (constant folding, non-nullability detection in certain branches, etc.), but we won't run `COALESCE` any more than we would otherwise.

There _is_ the possibility that this optimization grows the MSE---in the worst case, we could copy the `COALESCE` many times. Gating this optimization on the presence of a literal `NULL` means that we'll copy it at most once, and in so doing we possibly drop the `COALESCE` entirely. I can imagine doing things with term size to try to trigger this only when useful. I can also imagine triggering it unconditionally, or using types to guide things more (e.g., trigger it when one of the branches is not nullable).

src/transform/src/lib.rs
mgree: I put it low down for no particular reason---mostly by analogy to the case_literal work. I'm happy to move it to the logical level.

> Btw., what was the problem with centralizing all the scalar expression optimizations in `MSE::reduce`? Is it that it was meant originally only for constant folding, but then various other optimizations crept in? Random idea, but how about having 2 of these functions:

This is an optimization that should really only fire once, but we invoke `MSE::reduce` many times.
 
> * one that does just constant folding
> * and one that does all the other scalar expr optimizations.

I like the idea of breaking up `MSE::reduce`, but that isn't a small work item---we'd have to start with an inventory of `MSE::reduce`, I figure.

src/transform/src/coalesce_case.rs
mgree: > We'll now have 3 transforms that mainly work with only the scalar expressions of MIR: `ReduceScalars`, `CaseLiteral`, `CoalesceCase`, and all 3 of them re-implement the delicate logic of finding all the scalar expressions in an `MirRelationExpr`. Note that we have `visit_scalars_mut`, which promises to do exactly this, but somehow all 3 of the transforms require various minor bits of extra logic that makes `visit_scalars_mut` insufficient. I'm not really sure what to do about this, but thought I mention it.

Yes. I thought about making a generic "MSE transform" that packages it up, but `CaseLiteral` needs types and `CoalesceCase` doesn't, figured I'd save on some work.

I responded to your latter comment below... in the world where we break up `MSE::reduce`, we'll want to land on an abstraction for transforms that find `MSE`s; we might want a way to run independent such transforms simultaneously (i.e., do a single pass over the MRE and its MSEs, running a bunch of independent rewrites simultaneously, to avoid reprocessing... not dissimilar to the existing `MSE::reduce` work, but a little more structured).

ggevay (approved):
LGTM

PR #35837, opened 2026-04-02
adapter: Further preparations for handling query plans in catalog implications

ggevay:
This is further preparation for handling those DDL commands in the catalog implications that need query plans. The 2 commits move things from the catalog transaction's side effect closure to before the catalog transaction. See more details in the commit messages.

The important thing is that this unlocks a later change that reads plans back from the cache in `parse_item`. (Draft PR: https://github.com/MaterializeInc/materialize/pull/36146)

Nightly: https://buildkite.com/materialize/nightly/builds/16131 (slightly stale)

### DIFF
--- a/src/adapter/src/coord/sequencer/inner/create_index.rs
+++ b/src/adapter/src/coord/sequencer/inner/create_index.rs
@@ -22,6 +22,7 @@ use mz_sql::names::ResolvedIds;
 use mz_sql::plan;
 use tracing::Span;
 
+use crate::catalog::CatalogState;
 use crate::command::ExecuteResponse;
 use crate::coord::sequencer::inner::return_if_err;
 use crate::coord::{
@@ -442,6 +443,9 @@ impl Coordinator {
         } = stage;
         let id_bundle = dataflow_import_id_bundle(global_lir_plan.df_desc(), cluster_id);
 
+        let on_entry = self.catalog().get_entry_by_global_id(&on);
+        let owner_id = *on_entry.owner_id();
+
         let ops = vec![catalog::Op::CreateItem {
             id: item_id,
             name: name.clone(),
@@ -459,7 +463,7 @@ impl Coordinator {
                 physical_plan: None,
                 dataflow_metainfo: None,
             }),
-            owner_id: *self.catalog().get_entry_by_global_id(&on).owner_id(),
+            owner_id,
         }];
 
         // Pre-allocate a vector of transient GlobalIds for each notice.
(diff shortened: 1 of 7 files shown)


src/adapter/src/coord/sequencer/inner/create_index.rs
```diff
@@ -468,11 +472,60 @@ impl Coordinator {
...
+            CatalogState::render_notices_core(
+                &humanizer,
+                (self.catalog().config().now)(),
+                &raw_df_meta,
+                notice_ids,
+                Some(global_id),
+            )
+        };
+
+        // Populate the durable expression cache before the catalog
```
mtabebe, line 511: What happens if the transaction fails? My assumption is that it ok to cache the expression of failed transactions... just confirming

mtabebe: Similarly, how do we handle notices if the transaction doesn't succeed? Will this state just remain around forever?

mtabebe (commented):
Great test coverage. And I *think* that I have reviewed some of these commits before. So I mainly focused on the last one. It looks good. Just one minor question

src/adapter/src/coord/sequencer/inner/create_index.rs
ggevay: Good questions!

**Expression cache on failed transaction:** Yes, harmless. The cache entry is keyed by `(build_version, GlobalId, expr_type)`. If the transaction fails, that `GlobalId` never lands in the catalog, so the entry is never looked up — `get_global_expressions` (in the follow-up PR) only queries GlobalIds from `Op::CreateItem` ops that are actually being committed. The orphaned entry sits inert in the persist shard until the next version bump cleans it up via `remove_prior_versions`.

**Notices on failed transaction:** This is actually the exact bug this PR fixes! Pre-fix, `emit_optimizer_notices` ran *before* the catalog transaction, so the user saw spurious notices (e.g. "identical index already exists") even when `IF NOT EXISTS` hit a name collision and the item wasn't created. Post-fix:
- Raw notices to the user (`emit_raw_optimizer_notices_to_user`) are emitted only in the `Ok(_)` arm after the transaction succeeds.
- Rendered notices persisted to `mz_optimizer_notices` (via `persist_dataflow_metainfo`) only happen inside the side-effect closure, which only runs on success.
- Rendered notices in the expression cache are only consumed by `parse_item` during `apply_updates` for committed transactions, so also inert on failure.

The new `notice.pt` test at the bottom of the PR pins this invariant for both `CREATE INDEX IF NOT EXISTS` and `CREATE MATERIALIZED VIEW IF NOT EXISTS`.

ggevay: Ah lol, Claude just went ahead and posted this when I asked it to draft a reply :D

ggevay: But yeah, its answer is mostly right: The important thing is that this is only a small amount of orphaned data, and it gets cleaned up on the next version upgrade, because the expression cache is scoped to a version.

mtabebe (approved):

PR #4816, opened 2020-11-17
Use enter_region to enter regions

frankmcsherry:
The methods `enter_region` and `leave_region` are specialized to cross scope boundaries without a `map` operator to transform timestamps as they go. This simplifies our dataflows slightly, in cases that we move collection streams across dataflows.

---
This change is [Reviewable](https://reviewable.io/reviews/materializeinc/materialize/4816)

Issue #4759, opened 2020-11-11
Can create a table without any fields. subsequent inserts cause a crash


ruchirK:
I'm on main `materialized v0.5.2-dev (854634eb4)`

Can repro the behavior with

```
materialize=> create table ints;
CREATE TABLE
materialize=> insert into ints values (7, 3);
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
!>
```

Postgres complains with a syntax error for the create table statement.

Backtrace:
```
Nov 10 19:54:19.081 ERROR panic: <unnamed>: index out of bounds: the len is 0 but the index is 0
   0: backtrace::backtrace::libunwind::trace
             at /Users/Test/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.54/src/backtrace/libunwind.rs:90:5
      backtrace::backtrace::trace_unsynchronized
             at /Users/Test/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.54/src/backtrace/mod.rs:66:5
      backtrace::backtrace::trace
             at /Users/Test/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.54/src/backtrace/mod.rs:53:14
   1: backtrace::capture::Backtrace::create
             at /Users/Test/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.54/src/capture.rs:176:9
      backtrace::capture::Backtrace::new
             at /Users/Test/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.54/src/capture.rs:140:22
   2: materialized::handle_panic
             at src/materialized/src/bin/materialized/main.rs:540:65
   3: core::ops::function::Fn::call
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/core/src/ops/function.rs:70:5
   4: std::panicking::rust_panic_with_hook
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:573:17
   5: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:476:9
   6: std::sys_common::backtrace::__rust_end_short_backtrace
[...]

PR #4631, opened 2020-10-26
re-use allocation for quickselect

frankmcsherry:
Our logic for handling `order by .. limit ..` in the fast path does two allocations for each comparison, unpacking the `Row` into a `Vec<Datum>`. This PR allocates two vectors and repeatedly unpacks the rows into the same allocations, avoiding per-comparison allocation.

---
This change is [Reviewable](https://reviewable.io/reviews/materializeinc/materialize/4631)

PR #35858, opened 2026-04-03
ssh-util: add static OpenSSH build and FIPS algorithm enforcement

jasonhernandez:
## Summary

Infrastructure for FIPS 140-3 compliant SSH tunnels, enabling distroless container migration for environmentd and clusterd:

- **Static OpenSSH build Dockerfile** (`misc/images/openssh-static/`): Builds a statically-linked `ssh` client binary (~3-6MB) using AWS-LC-FIPS as the crypto backend. OpenSSH natively supports AWS-LC — no patches needed.
- **FIPS SSH algorithm enforcement** (`src/ssh-util/src/tunnel.rs`): When `MZ_FIPS=1` is set, writes a restrictive SSH config that limits connections to NIST-approved algorithms only (AES-GCM/CTR, ECDH, HMAC-SHA2, ECDSA/RSA).

### Background

environmentd and clusterd can't move to distroless containers because the `openssh` Rust crate shells out to `/usr/bin/ssh`. The `russh` pure-Rust alternative was evaluated but **cannot be made FIPS-compliant** (only AEAD ciphers use aws-lc-rs; key exchange, signing, and HMACs use non-FIPS RustCrypto crates).

Shipping a static OpenSSH binary built against AWS-LC-FIPS is the pragmatic path: zero code changes to tunnel logic, battle-tested SSH implementation, and FIPS-validated crypto.

### Follow-up PRs

- CI builder integration to actually build and cache the static ssh binary
- Migrate environmentd/clusterd Dockerfiles from `prod-base` to `distroless-prod-base`
- Replace bash entrypoint scripts with static/compiled entrypoints

Part of SEC-236.

## Test plan

- [x] `cargo check -p mz-ssh-util` passes
- [x] `cargo fmt` clean
- [ ] Docker build of `misc/images/openssh-static/` produces working static binary
- [ ] SSH tunnel tests pass with `MZ_FIPS=1` against a FIPS-compatible SSH server

---
🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/misc/images/openssh-static/Dockerfile
+++ b/misc/images/openssh-static/Dockerfile
@@ -1,0 +1,88 @@
+# Copyright Materialize, Inc. and contributors. All rights reserved.
+#
+# Use of this software is governed by the Business Source License
+# included in the LICENSE file at the root of this repository.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0.
+
+# Build a statically-linked OpenSSH `ssh` client binary using AWS-LC
+# as the crypto backend. AWS-LC is a faster, smaller alternative to
+# OpenSSL that also supports FIPS 140-3 validation when needed.
+#
+# OpenSSH natively supports AWS-LC as a crypto backend (no patches needed).
+# See: https://github.com/openssh/openssh-portable/blob/master/INSTALL
+#
+# To enable FIPS mode, build with: --build-arg AWS_LC_FIPS=1
+# (requires Go for the FIPS delocator)
+#
+# Usage:
+#   docker build -t openssh-static .
+#   docker create --name extract openssh-static
+#   docker cp extract:/output/ssh ./ssh
+#   docker rm extract
+
+FROM ubuntu:noble-20260210.1 AS builder
+
+ARG AWS_LC_VERSION=v1.54.0
+ARG AWS_LC_FIPS=0
+ARG OPENSSH_VERSION=V_9_9_P2
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+    autoconf \
+    automake \
+    build-essential \
+    ca-certificates \
+    cmake \
+    git \
+    golang \
+    ninja-build \
+    perl \
+    pkg-config \
+    wget \
+    zlib1g-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Build AWS-LC as a static library.
+# When AWS_LC_FIPS=1, enables FIPS mode (requires Go for the delocator).
+WORKDIR /build/aws-lc
+RUN git clone --depth 1 --branch ${AWS_LC_VERSION} https://github.com/aws/aws-lc.git . \
+    && cmake -GNinja -B build \
+        $([ "$AWS_LC_FIPS" = "1" ] && echo "-DFIPS=1") \
+        -DBUILD_SHARED_LIBS=0 \
+        -DBUILD_TESTING=OFF \
+        -DBUILD_TOOL=OFF \
+        -DCMAKE_BUILD_TYPE=Release \
+        -DCMAKE_INSTALL_PREFIX=/opt/aws-lc \
+        -DCMAKE_C_FLAGS="-fPIC" \
+    && ninja -C build \
+    && ninja -C build install
+
+# Build OpenSSH ssh client against AWS-LC.
+WORKDIR /build/openssh
+RUN git clone --depth 1 --branch ${OPENSSH_VERSION} https://github.com/openssh/openssh-portable.git . \
+    && autoreconf \
+    && ./configure \
+        --with-ssl-dir=/opt/aws-lc \
+        --with-zlib \
+        --with-ldflags=-static \
+        --without-pam \
+        --without-libedit \
+        --disable-pkcs11 \
+    # AWS-LC does not define the legacy OpenSSL BN_FLG_CONSTTIME flag.
+    # Setting it to 0 satisfies #ifdef checks in OpenSSH source code.
+    # This is safe: AWS-LC handles constant-time bignum operations
+    # internally and does not rely on this flag.
+    # --disable-pkcs11 avoids link errors from ssh-pkcs11.c calling
+    # RSA_meth_dup/EC_KEY_METHOD_get_sign which AWS-LC does not provide.
+    && make -j"$(nproc)" ssh CFLAGS="-DBN_FLG_CONSTTIME=0" \
+    && strip ssh
+
+# Verify the binary is not dynamically linked and is functional.
+RUN ! ldd ssh 2>/dev/null \
+    && ./ssh -V
+
+# Output stage: just the binary.
+FROM scratch
+COPY --from=builder /build/openssh/ssh /output/ssh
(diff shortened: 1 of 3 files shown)


jasonhernandez:
@def- let me know what you think of this - I would like to move our SSH containers to use Distroless and this unblocks that. The FIPS mode specific behavior can be picked up / refined later, but a staticly built OpenSSH helps get us onto minimal container bases.

def- (approved):
Seems totally fine to me.

misc/images/openssh-static/Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+#   docker build -t openssh-static .
+#   docker create --name extract openssh-static
+#   docker cp extract:/output/ssh ./ssh
+#   docker rm extract
+
+FROM ubuntu:noble-20260210.1 AS builder
+
+ARG AWS_LC_VERSION=v1.54.0
+ARG AWS_LC_FIPS=0
+ARG OPENSSH_VERSION=V_9_9_P2
```
def-, line 30: What kind of version is that? Seems like a pretty old one. I think dependabot won't pick this up.

def-:
The QA LLM review had a bunch of seemingly good remarks:

## Finding 1 — HIGH: `strip ssh` invalidates the FIPS integrity self-test

`misc/images/openssh-static/Dockerfile:79-80`

```dockerfile
    && make -j"$(nproc)" ssh CFLAGS="-DBN_FLG_CONSTTIME=0" \
    && strip ssh
```

The binary is **unconditionally stripped** after linking, including when built with
`--build-arg AWS_LC_FIPS=1` (the binary's entire reason to exist).

A FIPS 140-3 module performs a power-on self-test that computes an HMAC over its own
`.text`/`.rodata` and compares it against a value injected at build time. Both AWS-LC's own
documentation and **this project's own design ticket** explicitly prohibit stripping a FIPS
binary:

- AWS-LC `crypto/fipsmodule/FIPS.md`: the binary must not be stripped, or the integrity
  tooling cannot locate the FIPS module.
- SEC-236, "Constraints": *"FIPS integrity hash computed at link time; **do not strip or UPX
  after linking** if FIPS integrity checks are enabled."*

The `strip` here is exactly the prohibited "strip after linking" operation. Consequence when
`AWS_LC_FIPS=1`:

- Best case: the FIPS power-on self-test fails at process start and `ssh` aborts — i.e. a FIPS
  `ssh` binary **cannot be produced** by this Dockerfile.
- Worse case: `ssh -V` (the only verification, line 84) performs no crypto and so does **not**
  trigger the self-test, the broken binary passes verification and ships, and it fails the
  integrity check at the first real SSH connection — a silent failure that surfaces only in
  production FIPS deployments.

Either way the FIPS deliverable is broken, and the verification step is not strong enough to
catch it.

### Suggested fix

Only strip non-FIPS builds (symbols add ~hundreds of KB but must be retained for FIPS):

```dockerfile
    && make -j"$(nproc)" ssh \
    && if [ "$AWS_LC_FIPS" != "1" ]; then strip ssh; fi
```

(Combine with the `CFLAGS` fix in Finding 3.) The `~3-6MB` size claim in the header comment
applies to the stripped non-FIPS binary; the FIPS binary will be larger, which is expected.

---

## Finding 2 — MEDIUM: AWS-LC is pinned to a non-FIPS-validated release tag

`misc/images/openssh-static/Dockerfile:28,50-52`

```dockerfile
ARG AWS_LC_VERSION=v1.54.0
...
RUN git clone --depth 1 --branch ${AWS_LC_VERSION} https://github.com/aws/aws-lc.git . \
    && cmake -GNinja -B build \
        $([ "$AWS_LC_FIPS" = "1" ] && echo "-DFIPS=1") \
        ...
```

`v1.54.0` is a regular AWS-LC **release** tag. FIPS 140-3 validation applies only to specific
**FIPS-branch snapshots** of the source (each tied to a NIST certificate: AWS-LC-FIPS v1.0 =
cert #4631, v2.0 = #4759, v3.0 static = in process). Passing `-DFIPS=1` against an arbitrary
release tag produces a "FIPS-style" module (self-tests, integrity check) that is **not the
NIST-validated module**, so it provides no actual FIPS 140-3 compliance — defeating the
PR's stated goal of "FIPS-validated crypto."

SEC-236 calls this out directly: *"Must use a FIPS-certified AWS-LC tag (e.g.,
`fips-2024-09-27`, NIST cert #4816)."* The Dockerfile ignores that and uses `v1.54.0` for both
the FIPS and non-FIPS paths. (Secondarily, `-DFIPS=1` on a non-FIPS-branch release tag is also
prone to build failures, e.g. on some release/arch combinations.)

### Suggested fix

Use a FIPS-validated branch/tag when `AWS_LC_FIPS=1` — e.g. a separate
`ARG AWS_LC_FIPS_VERSION=fips-2024-09-27` selected when FIPS is enabled — rather than a generic
release tag.

---

## Finding 3 — MEDIUM: `make CFLAGS=...` discards configure's hardening and optimization flags

`misc/images/openssh-static/Dockerfile:79`

```dockerfile
    && make -j"$(nproc)" ssh CFLAGS="-DBN_FLG_CONSTTIME=0" \
```

The intent (per the comment) is merely to add the define `-DBN_FLG_CONSTTIME=0`. But OpenSSH's
`Makefile.in` (verified for the pinned tag `V_9_9_P2`) uses a plain assignment
`CFLAGS=@CFLAGS@` (no `override`, and there is **no** `CFLAGS_EXTRA` in this version). A
command-line `make CFLAGS=...` therefore **replaces** the entire set of flags that
`./configure` determined.

`@CFLAGS@` is where OpenSSH puts its optimization and exploit-mitigation flags — `-O2`,
`-fstack-protector-strong`, `-fPIE`, `-ftrapv`, warning flags, etc. Include paths live in
`CPPFLAGS` (`-I. -I$(srcdir) … @DEFS@`), which is preserved, so the build still **compiles and
links** — it just silently produces an **unoptimized, unhardened** `ssh` binary (no stack
canaries, no PIE, no FORTIFY). For an internet-facing SSH client whose whole purpose is a
hardened FIPS posture, that's a real regression, and it's invisible to any functional test.

### Suggested fix

Inject the define at `configure` time so it merges with the hardening flags instead of
replacing them, then run `make` with no override:

```dockerfile
    && ./configure CPPFLAGS="-DBN_FLG_CONSTTIME=0" \
        --with-ssl-dir=/opt/aws-lc --with-zlib --with-ldflags=-static \
        --without-pam --without-libedit --disable-pkcs11 \
    && make -j"$(nproc)" ssh
```

---

## Finding 4 — LOW: FIPS `HostKeyAlgorithms` omits `ssh-ed25519`

`src/ssh-util/src/tunnel.rs:250`

```
HostKeyAlgorithms ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,rsa-sha2-256,rsa-sha2-512
```

`HostKeyAlgorithms` controls which **bastion host-key** types the client will accept. With
`ssh-ed25519` omitted (while `PubkeyAcceptedAlgorithms` on the next line *does* include it for
the client's own ed25519 key), a bastion that presents only an ed25519 host key — a common
modern hardening choice — becomes unreachable once `MZ_FIPS=1`, with an opaque
"no matching host key type" error. Most OpenSSH servers also offer RSA/ECDSA host keys, so this
only bites ed25519-only bastions, and restricting host-key algorithms is arguably intended in
FIPS mode. Combined with the fact that the whole path is dormant (`MZ_FIPS` unset everywhere),
this is LOW, but worth confirming it's intentional before FIPS is enabled in production, since
the failure mode is a confusing breakage of a previously-working tunnel.

PR #35863, opened 2026-04-03
[SS-101] Storage introspection mirrored in compute

frankmcsherry:
Events from storage's timely infra are captured and replayed in the compute logging dataflows, with identifiers bumped up by 2^48. They are pretty large at that point, but shouldn't collide with the compute identifiers. The nudged identifiers should include Both the timely operator ids, and also the dataflow identifiers (the lead coordinate of addresses).

Not much testing, other than observing the information flowing after creating an auction leaden source.

### DIFF
--- a/src/clusterd/src/lib.rs
+++ b/src/clusterd/src/lib.rs
@@ -167,6 +167,11 @@ struct Args {
     /// affinity might degrade dataflow performance rather than improving it.
     #[clap(long)]
     worker_core_affinity: bool,
+
+    /// Forward storage's timely logging events to compute so storage operators appear in
+    /// `mz_introspection.mz_dataflow_*` tables.
+    #[clap(long)]
+    enable_storage_introspection_logs: bool,
 }
 
 pub fn main() {
@@ -379,9 +384,14 @@ async fn run(args: Args) -> Result<(), anyhow::Error> {
     compute_timely_config.process = args.process;
 
     // Create per-worker bridges for forwarding storage timely logging events to compute.
-    let num_workers = storage_timely_config.workers;
-    let (storage_log_writers, storage_log_readers): (Vec<_>, Vec<_>) =
-        (0..num_workers).map(|_| arc_event_link()).unzip();
+    let (storage_log_writers, storage_log_readers) =
+        if args.enable_storage_introspection_logs {
+            (0..storage_timely_config.workers)
+                .map(|_| arc_event_link())
+                .unzip()
+        } else {
+            (Vec::new(), Vec::new())
+        };
 
     // Start storage server.
     let storage_client_builder = mz_storage::serve(
--- a/src/clusterd/src/lib.rs
+++ b/src/clusterd/src/lib.rs
@@ -383,15 +383,20 @@ async fn run(args: Args) -> Result<(), anyhow::Error> {
     let mut compute_timely_config = args.compute_timely_config;
     compute_timely_config.process = args.process;
 
+    // We assume each storage worker has a corresponding compute worker that can process its logs.
+    assert_eq!(
+        storage_timely_config.workers, compute_timely_config.workers,
+        "storage and compute must have equal workers-per-process",
+    );
+
     // Create per-worker bridges for forwarding storage timely logging events to compute.
-    let (storage_log_writers, storage_log_readers) =
-        if args.enable_storage_introspection_logs {
-            (0..storage_timely_config.workers)
-                .map(|_| arc_event_link())
-                .unzip()
-        } else {
-            (Vec::new(), Vec::new())
-        };
+    let (storage_log_writers, storage_log_readers) = if args.enable_storage_introspection_logs {
+        (0..storage_timely_config.workers)
+            .map(|_| arc_event_link())
+            .unzip()
+    } else {
+        (Vec::new(), Vec::new())
+    };
 
     // Start storage server.
     let storage_client_builder = mz_storage::serve(
--- a/src/clusterd/src/lib.rs
+++ b/src/clusterd/src/lib.rs
@@ -38,6 +38,7 @@ use mz_service::secrets::SecretsReaderCliArgs;
 use mz_service::transport;
 use mz_storage::storage_state::StorageInstanceContext;
 use mz_storage_types::connections::ConnectionContext;
+use mz_timely_util::capture::arc_event_link;
 use mz_txn_wal::operator::TxnsContext;
 use tokio::runtime::Handle;
 use tower::Service;
@@ -377,6 +378,11 @@ async fn run(args: Args) -> Result<(), anyhow::Error> {
     let mut compute_timely_config = args.compute_timely_config;
     compute_timely_config.process = args.process;
 
+    // Create per-worker bridges for forwarding storage timely logging events to compute.
+    let num_workers = storage_timely_config.workers;
+    let (storage_log_writers, storage_log_readers): (Vec<_>, Vec<_>) =
+        (0..num_workers).map(|_| arc_event_link()).unzip();
+
     // Start storage server.
     let storage_client_builder = mz_storage::serve(
         storage_timely_config,
@@ -387,6 +393,7 @@ async fn run(args: Args) -> Result<(), anyhow::Error> {
         SYSTEM_TIME.clone(),
         connection_context.clone(),
         StorageInstanceContext::new(args.scratch_directory.clone(), args.announce_memory_limit)?,
+        storage_log_writers,
     )
     .await?;
     info!(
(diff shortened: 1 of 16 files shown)


src/storage/src/server.rs
```diff
@@ -104,6 +116,52 @@ impl ClusterSpec for Config {
...
+        // Register a timely logger that forwards events to the compute logging dataflow.
+        let writer = self.timely_log_writers.lock().unwrap().pop();
+        if let Some(writer) = writer {
+            use timely::dataflow::operators::capture::{Event, EventPusher};
+            use timely::logging::TimelyEventBuilder;
+
+            // We use an approach similar to compute's logging: wrap the writer in
+            // a BatchLogger that translates Logger callbacks into Event pushes,
+            // then register the Logger with timely's log_register.
+            let interval_ms = 1000u128; // 1 second batching interval
```
antiguru, line 143: TODO, but low priority: Make the interval configurable.

src/compute/src/server.rs
```diff
@@ -397,21 +417,27 @@ impl<'w, A: Allocate + 'static> Worker<'w, A> {
...
         }
-        self.activate_compute().unwrap().handle_compute_command(cmd);
+        // Take the storage log reader before borrowing self for activate_compute.
+        let storage_log_reader = if is_create_instance {
+            self.storage_log_reader.take()
+        } else {
+            None
+        };
+        let mut active = self.activate_compute().unwrap();
+        active.storage_log_reader = storage_log_reader;
```
antiguru: This is fragile as it only works if the current command is `CrateInstance`, but not in any other case. This is transient state, but it seems to be better to make it part of `ComputeState` to avoid unrelated parts of the code to depend on implementation details.

src/compute/src/server.rs
```diff
@@ -72,6 +78,10 @@ struct Config {
     pub metrics_registry: MetricsRegistry,
     /// The number of timely workers per process.
     pub workers_per_process: usize,
+    /// Per-worker readers for storage timely logging events.
+    // TODO: Consider using the timely config's `process` and `workers` fields to
+    // deterministically assign readers to workers by local index, rather than pop().
+    pub storage_log_readers: Arc<Mutex<Vec<StorageTimelyLogReader>>>,
```
antiguru: The invariant in compute dataflows is that worker id $$x$$ processes introspection data of worker $$x$$. It might be that a more general invariant might be true, too, but we need to audit the implementation. Probably this is true: Each worker ID must send its updates to exactly one processing operator instance, for the lifetime of the worker.

src/clusterd/src/lib.rs
```diff
@@ -368,6 +369,11 @@ async fn run(args: Args) -> Result<(), anyhow::Error> {
     let mut compute_timely_config = args.compute_timely_config;
     compute_timely_config.process = args.process;
 
+    // Create per-worker bridges for forwarding storage timely logging events to compute.
+    let num_workers = storage_timely_config.workers;
+    let (storage_log_writers, storage_log_readers): (Vec<_>, Vec<_>) =
+        (0..num_workers).map(|_| arc_event_link()).unzip();
```
antiguru: This needs to be feature-flagged. It means a new parameter to clusterd that is set by envd depending on a feature flag.

antiguru (commented):
The approach checks out, I left some comments around the implementation.

One though I have is that we should feature-gate the functionality and have it turned off by default to prevent downstream consumers to make assumptions around the operator IDs, but that is contrary to the goal of introspection. So we should at least document it as a temporary behavior until we figure out our plans for cluster unification.

src/compute/src/logging/timely.rs
```diff
@@ -785,3 +827,56 @@ where
...
+
+/// Offset added to storage operator/channel IDs to avoid collisions with compute IDs.
+///
+/// This is large enough that compute IDs (which start from 0 and grow) will never reach it,
+/// but small enough to be representable as a `u64` with room for many storage operators.
+const STORAGE_ID_OFFSET: usize = 1 << 48;
+
+/// Remaps operator, channel, and address IDs in a `TimelyEvent` originating from a storage
+/// timely instance so they don't collide with compute's IDs.
+fn remap_timely_event_ids(event: &mut TimelyEvent) {
```
ublubu, line 843: Question for console folks: How does the UI decide which things show up in the clickable introspection graph thing? If a bunch of storage operators show up in the introspection tables with these gigantic IDs, will that clutter the existing introspection UI for compute?
@leedqin

antiguru: It renders specific dataflows, so I don't think this is a problem.

ublubu:
I've run this on my laptop and in my staging env. I'm able to query introspection tables, filtering for Storage with `id >= 281474976710656`.

In staging, the graphical visualization for Index looks normal (no pollution from Storage introspection logs).

I believe I've addressed all @antiguru's comments except the one about making the logging interval configurable (currently locked to 1s).

src/storage/src/server.rs
```diff
@@ -48,6 +50,19 @@ struct Config {
...
     pub shared_rocksdb_write_buffer_manager: SharedWriteBufferManager,
+    /// Number of timely workers in this process, for local-index computation.
+    pub workers_per_process: usize,
+    /// Per-worker writers for forwarding timely logging events to compute,
+    /// indexed by local worker index.
+    pub timely_log_writers: Arc<
+        Mutex<
+            Vec<
+                Option<
+                    Arc<ArcEventLink<mz_repr::Timestamp, Vec<(std::time::Duration, TimelyEvent)>>>,
```
antiguru: Seems it might be a good idea to introduce a typedef?

src/storage/src/server.rs
```diff
@@ -103,6 +133,54 @@ impl ClusterSpec for Config {
...
+                .duration_since(std::time::SystemTime::UNIX_EPOCH)
+                .expect("Failed to get duration since Unix epoch");
+
+            let logger = timely::logging_core::Logger::<TimelyEventBuilder>::new(
+                now,
+                start_offset,
+                move |time: &std::time::Duration,
+                      data: &mut Option<Vec<(std::time::Duration, TimelyEvent)>>| {
+                    if let Some(data) = data.take() {
+                        event_pusher.push(Event::Messages(time_ms, data));
```
antiguru, line 169: Let's move the data processing (filtering parks, remapping IDs) here. This localizes knowledge about what's specific to storage dataflows in the storage crate, rather than spilling it into compute. Also saves a filter/unary operator on the other end.

src/storage/src/server.rs
```diff
@@ -103,6 +133,54 @@ impl ClusterSpec for Config {
...
+                            time_ms = new_time_ms;
+                        }
+                    }
+                },
+            );
+
+            let mut register = timely_worker
+                .log_register()
+                .expect("Logging must be enabled");
+            register.insert_logger("timely", logger);
```
antiguru:
```suggestion
            if let Some(mut register) = timely_worker
                .log_register() {
                register.insert_logger("timely", logger);
            }
```

src/timely-util/src/capture.rs
```diff
@@ -13,10 +13,33 @@
...
 
+/// A thread-safe linked list for cross-thread event streaming.
+///
+/// This re-exports timely's `link_sync::EventLink`, which uses `Arc`/`Mutex`
+/// (as opposed to the `Rc`/`RefCell`-based `EventLink` in `timely::capture`).
+///
+/// Both writer and reader are `Arc<ArcEventLink<T, C>>`: the writer implements
+/// [`EventPusher`] and the reader implements
+/// [`EventIterator`](timely::dataflow::operators::capture::event::EventIterator).
+pub type ArcEventLink<T, C> = EventLink<T, C>;
```
antiguru: This type def doesn't seem to add much value?

src/environmentd/tests/server.rs
```diff
@@ -1300,11 +1302,13 @@ fn test_cancel_insert_select() {
...
     // Query that returns how many dataflows are currently installed.
-    // Accounts for the presence of introspection subscribe dataflows by ignoring those.
+    // Ignores introspection subscribe dataflows.
+    // Ignores storage operators, whose IDs are offset by STORAGE_ID_OFFSET (1 << 48).
     const DATAFLOW_QUERY: &str = " \
         SELECT count(*) \
         FROM mz_introspection.mz_dataflows \
-        WHERE name NOT LIKE '%introspection-subscribe%'";
+        WHERE name NOT LIKE '%introspection-subscribe%' \
+        AND id < 281474976710656";
```
DAlperin, line 1311: If you want to you can do `AND id < 1::bigint << 48` for ease of reading

antiguru (approved):
Seems good

DAlperin (approved):
Very cool! Approved modulo moritzs comments!

PR #35899, opened 2026-04-07
[SSO] Add design doc for group to role mapping for OIDC

mtabebe:
Proposes JWT-based group-to-role sync for self-managed.

On connection, Materialize reads group claims from the JWT and grants/revokes role memberships accordingly. Track these using a dedicated sentinel grantor (MZ_JWT_SYNC_ROLE_ID) to distinguish sync-managed from manually-managed grants.

### DIFF
--- a/doc/developer/design/20260407_jwt_role_mapping.md
+++ b/doc/developer/design/20260407_jwt_role_mapping.md
@@ -1,0 +1,235 @@
+# Group-to-Role Mapping for Self-Managed
+
+## Associated:
+1. [SSO For Self Managed](https://www.notion.so/materialize/SSO-for-Self-Managed-2a613f48d37b806f9cb2d06914454d15)
+2. [SCIM](https://www.notion.so/materialize/SCIM-33513f48d37b8086be2ad9ec2a9ad554)
+
+## The Problem
+
+Self-managed Materialize customers using OIDC SSO have no way to manage database permissions from their identity provider. Today, an admin must:
+
+1. Configure OIDC so users can log in (already supported)
+2. Manually run `GRANT` statements in Materialize for every user
+
+When a team member joins, leaves, or changes roles, the admin has to update permissions in both the IdP *and* Materialize separately. There's no connection between "Alice is in the analytics team in Okta" and "Alice can read the orders table in Materialize."
+
+## Success Criteria
+
+- Users' database role memberships are dynamically updated based on their IdP group memberships
+- Self-managed OIDC is the primary target
+- When a user's groups change in the IdP, their Materialize permissions update on next connection
+- Permissions manually set for a user are left unchanged (independent of sync)
+- Admins can configure whether sync failures block login (fail-open by default, fail-closed for strict compliance)
+- The design must not lock into any single identity provider (e.g., Frontegg); SCIM for self-managed is a near-term priority
+
+## Out of Scope
+
+- Auto-creating database roles from IdP groups
+- Syncing role *privileges* (GRANTs on objects) from the IdP, only role *membership*
+- Real-time push-based sync (SCIM webhook to Materialize), we sync on connection. This is standard practice (CockroachDB, PostgreSQL LDAP) and aligns with customer expectations. A limitation is that the console and internal tables only reflect state as of the user's last connection.
+- Revoking the user's role itself when removed from IdP (existing token-expiry behavior handles this)
+- Cloud / Frontegg support (future work, but the design intentionally avoids Frontegg-specific coupling to support self-managed SCIM as a near-term follow-up)
+
+### Future Goals
+
+- Admin controls the mapping between external group names and database roles (name remapping via regex transforms, e.g., `materialize_(.*)` -> `$1`)
+- Cloud / Frontegg group sync
+- Mid-session sync on token refresh
+- SCIM-based group provisioning for self-managed (complementary to JWT-based sync)
+
+## Solution Proposal
+
+Follow the CockroachDB pattern: extract groups from JWT claims, match group names directly to database role names (case-insensitive), sync memberships at session startup.
+
+### User Experience
+
+#### Admin Setup
+
+**Step 1: Configure IdP to include groups in the JWT.**
+In their IdP (Okta, Azure AD, etc.), the admin configures the OIDC application to include a `groups` claim in the access token.
+
+Example JWT payload after configuration:
+```json
+{
+  "sub": "alice@example.com",
+  "iss": "https://dev-123456.okta.com/oauth2/default",
+  "groups": ["analytics", "platform_eng"],
+  "exp": 1234567890
+}
+```
+
+**Step 2: Create matching roles in Materialize with privileges.**
+```sql
+CREATE ROLE analytics;
+GRANT SELECT ON TABLE orders, customers TO analytics;
+GRANT USAGE ON SCHEMA production TO analytics;
+
+CREATE ROLE platform_eng;
+GRANT ALL ON SCHEMA infrastructure TO platform_eng;
+```
+
+The role names must match the IdP group names (case-insensitive).
+
+**Step 3: Enable group sync.**
+```sql
+ALTER SYSTEM SET oidc_group_role_sync_enabled = true;
+-- Optional: change the claim name if your IdP uses something other than "groups"
+-- ALTER SYSTEM SET oidc_group_claim = 'groups';
+-- Optional: require group sync to succeed for login (fail-closed mode for strict compliance)
+-- ALTER SYSTEM SET oidc_group_role_sync_strict = true;
+```
+#### End User Experience
+
+Alice logs in via `psql` with her JWT. On connection:
+- Her user role `alice` is auto-provisioned if it doesn't exist (existing behavior)
+- The sync reads `groups: ["analytics", "platform_eng"]` from her JWT
+- Materialize internally runs `GRANT analytics TO alice` and `GRANT platform_eng TO alice`
+- Alice can now query tables she has access to through those roles
+
+Later, Alice moves to a different team. Her IdP admin removes her from `analytics` and adds her to `data_eng`. On her next login:
+- The sync sees her JWT now has `groups: ["data_eng", "platform_eng"]`
+- Materialize revokes `analytics` from Alice and grants `data_eng`
+- Her permissions update automatically
+
+If an admin also manually ran `GRANT reporting TO alice`, that grant is unaffected by sync: manual grants are never touched.
+
+### Where Groups Come From
+
+Groups come from a configurable JWT claim (default: `"groups"`). The claim value can be a JSON array of strings or a single string.
+
+### Configuration
+
+New system variables (set via `ALTER SYSTEM SET`, stored in `SystemVars`):
+
+```sql
+ALTER SYSTEM SET oidc_group_role_sync_enabled = true;
+ALTER SYSTEM SET oidc_group_claim = 'groups';
+ALTER SYSTEM SET oidc_group_role_sync_strict = false;  -- default: fail-open
+```
+
+| Variable | Default | Description |
+|---|---|---|
+| `oidc_group_role_sync_enabled` | `false` | Feature gate for group-to-role sync |
+| `oidc_group_claim` | `'groups'` | JWT claim name containing group memberships |
+| `oidc_group_role_sync_strict` | `false` | When `true`, login is rejected if sync fails (fail-closed). When `false`, sync errors are logged but login proceeds (fail-open). |
+
+Group names from the JWT are matched directly to Materialize role names (case-insensitive). Roles must be pre-created in Materialize; group sync does not auto-create roles (only auto-creates users). Pre-created roles serve as the allowlist for IdP groups. Roles prefixed with `mz_` or `pg_` are always excluded from sync to prevent privilege escalation to system roles.
+
+### Sync Logic
+
+The sync happens in `handle_startup_inner()`, after the user role is auto-provisioned but before the session is fully established.
+
+```
+1. Extract groups from JWT claims (via configured claim name)
+2. Normalize: lowercase, deduplicate, sort
+3. Look up each group name as a role name in the catalog (skip non-existent with warning). Reject any group that maps to a role prefixed with `mz_` or `pg_` (log warning).
+4. Get user's current RoleMembership.map from catalog
+5. Partition current memberships:
+   - sync_granted: entries where grantor == MZ_JWT_SYNC_ROLE_ID (sentinel)
+   - manual_granted: entries where grantor != sentinel (leave untouched)
+6. Diff against sync_granted only:
+   - To grant: target roles NOT in sync_granted AND NOT in manual_granted
+   - To revoke: sync_granted roles NOT in target roles
+7. Execute GRANTs (with sentinel grantor) and REVOKEs via catalog_transact
+```
+
+Note: if a role is already manually granted, we don't also sync-grant it. The manual grant takes precedence and won't be revoked by sync.
+
+### How Manual Grants Are Preserved (Grantor Field)
+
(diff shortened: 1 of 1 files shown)


doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,211 @@
...
+
+When a team member joins, leaves, or changes roles, the admin has to update permissions in both the IdP *and* Materialize separately. There's no connection between "Alice is in the analytics team in Okta" and "Alice can read the orders table in Materialize."
+
+## Success Criteria
+
+- Users' database role memberships are dynamically updated based on their IdP group memberships
+- Self-managed OIDC is the primary target
+- When a user's groups change in the IdP, their Materialize permissions update on next connection
+- Roles must be pre-created in Materialize; group sync does not auto-create roles (only auto-creates users)
+- Permissions manually set for a user are left unchanged (independent of sync)
```
Alphadelta14, line 22: so this line item actually leads to a class of security issues (shadowed permissions). if we are building out a feature around this, it MUST be configurable and should be off by default.

cc @jasonhernandez

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,211 @@
...
+2. Manually run `GRANT` statements in Materialize for every user
+
+When a team member joins, leaves, or changes roles, the admin has to update permissions in both the IdP *and* Materialize separately. There's no connection between "Alice is in the analytics team in Okta" and "Alice can read the orders table in Materialize."
+
+## Success Criteria
+
+- Users' database role memberships are dynamically updated based on their IdP group memberships
+- Self-managed OIDC is the primary target
+- When a user's groups change in the IdP, their Materialize permissions update on next connection
+- Roles must be pre-created in Materialize; group sync does not auto-create roles (only auto-creates users)
```
Alphadelta14, line 21: Probably doesn't need to be specifically mentioned, since it's called out in the Out-Of-Scope section.

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,211 @@
...
+- JWT group sync grants a role → grantor = `MZ_JWT_SYNC_ROLE_ID` (a new dedicated sentinel role)
+- Human runs `GRANT role TO user` → grantor = their own role ID
+
+**Note**: `RoleMembership.map` is `BTreeMap<RoleId, RoleId>` each role membership stores exactly one grantor. A role cannot simultaneously have both a manual grant and a sync grant; only one grantor is recorded. This means "manual wins" semantics must be enforced by the sync logic's ordering, not by the data model. Enforced by: sync checking the current grantor for each role membership before acting; if a role is already granted manually, the sync skips it entirely (to not overwrite the grantor); if a role is granted with the sentinel grantor (sync) and an admin later manually re-grants it, the manual `GRANT` overwrites the grantor to the admin's role ID.
+
+If the admin later `REVOKE`s that manual grant, the role membership is removed entirely. On next login, the sync sees the role is absent and would re-grants it with the sentinel grantor.
+
+### Edge Cases
+
+The principal goal is: log warning and skip on misconfiguration. Never block login due to sync issues.
```
Alphadelta14, line 143: for compliance, this should also be configurable. we can skip by default, but should block if configured more strictly.
"fail-open security" is the weakness here.

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,211 @@
...
+
+**All groups map to non-existent roles**: Login proceeds. User still has manual grants.
+
+**Case sensitivity**: Normalize group names to lowercase for matching against catalog role names.
+
+### Observability
+
+For MVP, sync activity is surfaced through:
+
+- **`mz_audit_log`**: All GRANTs and REVOKEs from sync are already logged (they go through `Op::GrantRole`/`Op::RevokeRole`).
```
Alphadelta14, line 163: if we don't have any, we should indicate the source

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+The role names must match the IdP group names (case-insensitive).
+
+**Step 3: Enable group sync.**
+```sql
+ALTER SYSTEM SET jwt_group_role_sync_enabled = true;
+-- Optional: change the claim name if your IdP uses something other than "groups"
+-- ALTER SYSTEM SET jwt_group_claim = 'groups';
+```
```
jubrad, line 76: We must not sync users to roles prefixed with `mz_` or `pg_`

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,211 @@
...
+- Self-managed OIDC is the primary target
+- When a user's groups change in the IdP, their Materialize permissions update on next connection
+- Roles must be pre-created in Materialize; group sync does not auto-create roles (only auto-creates users)
+- Permissions manually set for a user are left unchanged (independent of sync)
+
+## Out of Scope
+
+- Auto-creating database roles from IdP groups
+- Syncing role *privileges* (GRANTs on objects) from the IdP, only role *membership*
+- Real-time push-based sync (SCIM webhook to Materialize), we sync on connection
```
jubrad, line 28: sync on connect is the standard, but this also make sit very difficult to get an accurate representation of the current state in our web console or to discover access via querying internal tables. 

If this is standard art and aligned with customer expectations this is probably fine.

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,211 @@
...
+```sql
+CREATE ROLE analytics;
+GRANT SELECT ON TABLE orders, customers TO analytics;
+GRANT USAGE ON SCHEMA production TO analytics;
+
+CREATE ROLE platform_eng;
+GRANT ALL ON SCHEMA infrastructure TO platform_eng;
+```
+
+The role names must match the IdP group names (case-insensitive).
```
jubrad, line 69: In the past I've had to create a transformation for upstream IDP to user database roles
IE
`materialize_platform_eng` -> `platform_eng` via something like `materialize_(.*)->$1`

doc/developer/design/20260407_jwt_role_mapping.md
mtabebe: Makes sense to me

doc/developer/design/20260407_jwt_role_mapping.md
mtabebe: That makes sense to me. Who makes the call on customer expectations?

doc/developer/design/20260407_jwt_role_mapping.md
mtabebe: That makes sense, I'll leave it as an open question, we can incorporate it at a later point?

doc/developer/design/20260407_jwt_role_mapping.md
jubrad: This would be @maheshwarip or @val-materialize

Alphadelta14 (approved):
[diff](https://github.com/MaterializeInc/materialize/compare/f4c4dd1c4600751288564332d7ce1face0877020..36bfb1a6a70d2ac40d343ee8b8e0acc2c71bdebf) looks good to me

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Kubernetes Custom Resource Definitions
+
+Alternatively: model group-to-role mappings as Kubernetes CRDs, letting admins declare mappings declaratively in their cluster manifests. A controller would watch these resources and reconcile role memberships in Materialize.
+
+This was rejected because it would require building and maintaining a new API server (or operator) alongside environmentd. The JWT already carries the group information at connection time, so it would simply add latency and complexity without clear benefit over reading the claims directly.
+
+## Open Questions
+
+1. **Sync on the connection hot path**: `handle_startup_inner` runs on every connection. The sync does a `catalog_transact`, which takes a write lock on the catalog. If 50 users reconnect after a deploy, are we serializing these all through the catalog on startup? Or could we skip the operation if the groups haven't changed (which would be the common case).
```
SangJunBak: I think skipping the operation if the groups haven't changed would probably make this fast enough!

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,234 @@
...
+ALTER SYSTEM SET jwt_group_role_sync_strict = false;  -- default: fail-open
+```
+
+| Variable | Default | Description |
+|---|---|---|
+| `jwt_group_role_sync_enabled` | `false` | Feature gate for group-to-role sync |
+| `jwt_group_claim` | `'groups'` | JWT claim name containing group memberships |
+| `jwt_group_role_sync_strict` | `false` | When `true`, login is rejected if sync fails or groups resolve to empty (fail-closed). When `false`, sync errors are logged but login proceeds (fail-open). |
+
+Group names from the JWT are matched directly to Materialize role names (case-insensitive). Roles must be pre-created in Materialize; group sync does not auto-create roles (only auto-creates users). Pre-created roles serve as the allowlist for IdP groups. Roles prefixed with `mz_` or `pg_` are always excluded from sync to prevent privilege escalation to system roles.
```
SangJunBak, line 116: Currently our dyncfgs related to SSO are prefixed with `oidc_` rather than `jwt_`. However it might  Decided to prefix it as OIDC since we were relying on the Open ID connect spec, however there may be plans of.

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,234 @@
...
+
+This was rejected because it would require building and maintaining a new API server (or operator) alongside environmentd. The JWT already carries the group information at connection time, so it would simply add latency and complexity without clear benefit over reading the claims directly.
+
+## Open Questions
+
+1. **Sync on the connection hot path**: `handle_startup_inner` runs on every connection. The sync does a `catalog_transact`, which takes a write lock on the catalog. If 50 users reconnect after a deploy, are we serializing these all through the catalog on startup? Or could we skip the operation if the groups haven't changed (which would be the common case).
+2. **Unmatched group observability**: How should unmatched groups (IdP group with no corresponding Materialize role) be surfaced? The `mz_audit_log` uses a closed `EventType` enum (`Create`, `Drop`, `Alter`, `Grant`, `Revoke`, `Comment`), so adding a new event type requires a new `EventType` variant, a new `EventDetails` struct, and a proto migration. For MVP, server logs are sufficient, with a dedicated system table or audit log variant added later.
```
SangJunBak: I can imagine a user being part of many groups unrelated to Materialize, and if we're doing this per login, we might be storing a lot of redundant data if we decide to persist this. If we're improving on server logs, maybe these can be notices controlled by a new session variable that's off by default?

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,234 @@
...
+3. **Edge case behaviour**: Are these the right choices? (See Security: Shadowed Permissions section for the `strict` mode trade-off.)
+4. **Frontegg group claims in JWT**: Can Frontegg be configured to include group membership as a claim in the JWT access token it issues?
+5. **Two authentication paths for Cloud support**: The Frontegg authenticator uses app passwords. Users authenticate with a client ID and secret key, which are exchanged with Frontegg's API for a JWT (`exchange_app_password()`). This means group sync for Cloud would need its own implementation: extract groups from the JWT returned by the app password exchange, and push updated group memberships through another (or existing channel) on each token refresh. This is a separate code path from OIDC.
```
SangJunBak, line 233: This is unfortunate but not much we can do right now :(

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Open Questions
+
+1. **Sync on the connection hot path**: `handle_startup_inner` runs on every connection. The sync does a `catalog_transact`, which takes a write lock on the catalog. If 50 users reconnect after a deploy, are we serializing these all through the catalog on startup? Or could we skip the operation if the groups haven't changed (which would be the common case).
+2. **Unmatched group observability**: How should unmatched groups (IdP group with no corresponding Materialize role) be surfaced? The `mz_audit_log` uses a closed `EventType` enum (`Create`, `Drop`, `Alter`, `Grant`, `Revoke`, `Comment`), so adding a new event type requires a new `EventType` variant, a new `EventDetails` struct, and a proto migration. For MVP, server logs are sufficient, with a dedicated system table or audit log variant added later.
+3. **Edge case behaviour**: Are these the right choices? (See Security: Shadowed Permissions section for the `strict` mode trade-off.)
+4. **Frontegg group claims in JWT**: Can Frontegg be configured to include group membership as a claim in the JWT access token it issues?
```
SangJunBak: Before implementation, we should definitely should have someone from Cloud verify this, as well as verifying if we get the group membership of not only the user in Frontegg, but the upstream IdP too

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,234 @@
...
+
+Manually-granted permissions that coexist with sync-managed memberships create a risk of **shadowed permissions**: a user may retain access through a manual grant even after their IdP group membership is revoked. This is a known trade-off of the "manual grants are never touched" design.
+
+To mitigate this:
+- The `jwt_group_role_sync_strict` mode (when enabled) rejects login if sync fails, preventing stale permissions from persisting silently.
+- Admins should audit manual grants periodically. The sentinel grantor (`MZ_JWT_SYNC_ROLE_ID`) makes it possible to distinguish sync-managed from manual memberships via `mz_role_members`.
+
+### Edge Cases
+
+The default behavior is fail-open: log warning and skip on misconfiguration, allowing login to proceed. When `jwt_group_role_sync_strict = true`, sync failures and empty group resolution reject the login instead.
```
SangJunBak: By "log", do we mean send a notice?

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,234 @@
...
+
+**Group maps to non-existent role**: Log warning, record in `mz_audit_log`, and skip
```
SangJunBak: What would we record in the audit log?

doc/developer/design/20260407_jwt_role_mapping.md
```diff
@@ -0,0 +1,234 @@
...
+
+**Missing groups claim vs empty groups claim**: These are different.
+- `groups: []` (explicit empty) means revoke all sync-granted roles, keep manual grants. In strict mode, login is rejected if no manual grants remain. In default (fail-open) mode, login proceeds.
```
SangJunBak: > In strict mode, login is rejected if no manual grants remain 

Why would login be rejected here? To me, this should just be the same as if the user had only default privileges (reference: https://materialize.com/docs/security/self-managed/access-control/#initial-privileges)

doc/developer/design/20260407_jwt_role_mapping.md
mtabebe: Good point, nothing to record. I'll remove

doc/developer/design/20260407_jwt_role_mapping.md
mtabebe: Good point

SangJunBak (approved):

doc/developer/design/20260407_jwt_role_mapping.md
jasonhernandez: We should make best efforts to update permissions if we receive SCIM events. This isn't perfect and people in security roles work with a lot of eventually / never consistent systems, but we should try and provide them more real-time visibility into group changes they make. Okta (and probably others) include buttons to re-push SCIM events.

PR #35923, opened 2026-04-09
[SS-96] Setup alternate logic path for upsert_v2

patrickwwbutler:
This sets up logic to potentially use the new prototype upsert implementation written by @DAlperin and loosely described here: https://www.notion.so/materialize/Upsert-V2-Much-Simpler-Boogaloo-31913f48d37b807fa88bdeafc27c02d9?source=copy_link

The plan is to run this alternate implementation in a shadow cluster that copies the work done in production on certain customer's upsert workloads so we can evaluate for correctness and collect performance metrics before fully switching over.

### DIFF
--- a/src/storage/src/upsert_continual_feedback_v2.rs
+++ b/src/storage/src/upsert_continual_feedback_v2.rs
@@ -1,0 +1,912 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! Implementation of the feedback UPSERT operator.
+//!
+//! # Architecture
+//!
+//! The operator converts a stream of upsert commands `(key, Option<value>)` into
+//! a differential collection of `(key, value)` pairs, using a feedback loop
+//! through persist to maintain the "previous value" state needed for computing
+//! retractions.
+//!
+//! ## Dataflow topology
+//!
+//! ```text
+//!   Source input ──► ┌──────────┐ ──► Output ──► Persist
+//!                    │  Upsert  │
+//!   Persist read ──► └──────────┘
+//!       ▲                                           │
+//!       └───────────── feedback ────────────────────┘
+//! ```
+//!
+//! ## Operator loop (each iteration)
+//!
+//! 1. **Ingest source data.** Read upsert commands from the source input,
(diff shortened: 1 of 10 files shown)


DAlperin (approved):
Looks good. I most reviewed the surrounding bits that didn't come from me, but everything should be nicely feature flagged.

The only thing to flag, and this is probably claudeable, is you might want to make this a cluster level config so it's easy to test it in isolation with an unbilled cluster since currently there is no way to target a dyncfg to a specific cluster (something that bit me this week too, which is why I mention it)

patrickwwbutler:
cluster setting controlling upsert v2 will be in a followup PR

src/storage/src/upsert_continual_feedback_v2.rs
```diff
@@ -0,0 +1,1003 @@
...
+            // UpsertDiff (carrying FromTime for dedup), and buffer them.
+            // Events before the resume_upper are dropped (already persisted).
+            while let Some(event) = input.next_sync() {
+                match event {
+                    AsyncEvent::Data(cap, data) => {
+                        let mut pushed_any = false;
+                        for ((key, value, from_time), ts, diff) in data {
+                            assert!(diff.is_positive(), "invalid upsert input");
+                            if PartialOrder::less_equal(&input_upper, &resume_upper)
+                                && !resume_upper.less_equal(&ts)
```
ublubu, line 317: I'm curious about the meaning of these two constraints together.

What does it mean for `input_upper <= resume_upper` and `not (resume_upper <= ts)` to both be true?

To me, `not (resume_upper <= ts)` already looks like "Events before the resume_upper are dropped." What does `input_upper <= resume_upper` do?

ublubu: given: `input_upper <= ts`
does `ts < resume_upper` imply `input_upper <= resume_upper`?

----

the condition in the code:
`input_upper <= resume_upper`
and
`ts < resume_upper`

what if `resume_upper <= ts`? new data, do stuff with it.

ublubu: It looks like an artifact from the previous implementation, where we used the `input_upper <= resume_upper` as an optimization to preemptively skip the per-element `ts < resume_upper` check:

https://github.com/MaterializeInc/materialize/blob/f495cf372ed9a893b0bb2d8f0c913ea0aad6f5e6/src/storage/src/upsert_continual_feedback.rs#L621-L623

optional optimization TODO: use `Vec::extend` instead of a bunch of `Vec::push` (per Marty's suggestion)

PR #35928, opened 2026-04-10
clusterd: remove direct libc dependency

jasonhernandez:
## Summary
- Replace `getaddrinfo`-based FQDN resolution with `/etc/hosts` parsing — in Kubernetes, the kubelet writes the pod's FQDN there for StatefulSet pods, so no FFI needed
- Replace `libc::raise` with `nix::sys::signal::raise` in the termination signal handler
- Remove `libc` from clusterd's direct dependencies

**Depends on:** #35859 (distroless migration)

## Open questions
- Is `/etc/hosts` reliably populated in all deployment scenarios? The kubelet manages it for standard K8s pods, but custom configurations (hostNetwork, non-K8s docker run) may not have the FQDN entry. Falls back to short hostname with a warning in that case.

## Test plan
- [ ] Verify FQDN resolution works in a Kubernetes StatefulSet pod
- [ ] Verify fallback warning when `/etc/hosts` lacks the FQDN entry
- [ ] `cargo check -p mz-clusterd` passes


jasonhernandez:
Closing: clusterd libc changes are included in #36100 (distroless process lifecycle).

PR #35929, opened 2026-04-10
Remove direct libc dependency from orchestrator-process and environmentd

jasonhernandez:
## Summary
- **orchestrator-process**: Replace `libc::{SIGABRT, SIGBUS, SIGILL, SIGSEGV, SIGTRAP}` with `nix::sys::signal::Signal` variants (nix already a dependency)
- **environmentd**: Replace `libc::raise` with `nix::sys::signal::raise` and `libc::c_int` with `std::ffi::c_int`

Both crates already depend on `nix` which re-exports these via safe APIs. No new dependencies added.

## Test plan
- [x] `cargo check -p mz-orchestrator-process -p mz-environmentd` passes


Alphadelta14 (approved):
❤️

PR #35932, opened 2026-04-10
MySQL repro: table drop + create not handled correctly

martykulma:
repro of incorrect behavior with table drop + create


patrickwwbutler:
Batching the `DROP ...; CREATE ...;` into one transaction is a much smarter way of confirming this than my original method which was to... add a sleep statement in the rust handling DDL events

PR #35933, opened 2026-04-10
adapter: bootstrap storage collections in dependency order

teskje:
New builtin MVs that depend on other builtin MVs could fail as-of selection because the bootstrap code registered all derived collections (MVs, CTs) in a single batch. When computing a derived collection's since, the code queried the storage controller for dependency frontiers, but dependencies in the same batch weren't registered yet, so their (persisted) sinces were missed. This led to sinces that were too low, causing unsatisfiable as-of constraints.

Fix by registering collections in topological layers: each iteration drains and registers only the collections whose storage dependencies have all been registered. This ensures dependency frontiers are available when bumping the sinces of derived collections.

### Motivation

Fixes MaterializeInc/database-issues#11273

### Verification

Triggered the 0dt smoke test a couple times, which reproduced the issue before.

### DIFF
--- a/src/adapter/src/coord.rs
+++ b/src/adapter/src/coord.rs
@@ -68,7 +68,6 @@
 
 use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
-use std::fmt;
 use std::net::IpAddr;
 use std::num::NonZeroI64;
 use std::ops::Neg;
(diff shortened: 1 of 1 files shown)


teskje:
All the nightly failures look unrelated, so this should be good to go.

ggevay (approved):

teskje:
TFTR!

PR #21034, opened 2023-08-07
deps: update kube to 0.85.0 and k8s-openapi to 0.19.0+v1_25

rjobanp:
### Motivation

To avoid duplicate package version warnings on `serde_yaml` in the cloud repo caused by the introduction of `utoipa` for generating OpenAPI specs, we need to update the `kube` version to 0.85.0. However doing so in the cloud repo (and in https://github.com/MaterializeInc/k8s-controller) requires doing so in the materialize repo first, hence this PR.

### Tips for reviewer

The one code change is due to a slight type change in the kube crate: https://github.com/kube-rs/kube/blob/main/CHANGELOG.md#listwatch-changes

### Checklist

- [x] This PR has adequate test coverage / QA involvement has been duly considered.
- [x] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [x] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [x] If this PR will require changes to cloud orchestration, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [x] This PR includes the following [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note):
  -

PR #35934, opened 2026-04-10
persist/pubsub: set unlimited decoding message size everywhere

teskje:
This PR sets the `max_decoding_message_size` for all pubsub servers and clients to `usize::MAX`, to avoid pubsub errors and resulting reconnects caused by participants not being able to decode received messages.

There was already code in place to do this for a subset of the pubsub servers created (introduced in https://github.com/MaterializeInc/materialize/pull/21034), but it was (likely due to oversight) not applied to all of them, nor to clients.

Note that we don't need to set `max_encoding_message_size`, as it already defaults to `usize::MAX`.

### Motivation

We observed the decoding size limit to cause pubsub reconnects, which in turn tickles a connection leak ([Slack thread](https://materializeinc.slack.com/archives/C097HDRFJBF/p1775812206787469)). This change doesn't fix the leak yet but makes it less likely that we hit it.

### DIFF
--- a/src/persist-client/src/rpc.rs
+++ b/src/persist-client/src/rpc.rs
@@ -135,6 +135,12 @@ pub(crate) const PUBSUB_RECONNECT_BACKOFF: Config<Duration> = Config::new(
     "Backoff after an established connection to Persist PubSub service fails.",
 );
 
+/// Max message size, used to configure gRPC servers and clients.
+///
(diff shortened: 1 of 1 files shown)


antiguru (approved):
Thank you!

teskje:
TFTR!

PR #35935, opened 2026-04-10
compute: insert temporal bucketing before ArrangeBy nodes

antiguru:
Temporal MFPs (e.g. `WHERE mz_now() > ts`) emit updates with future timestamps. When those land in an arrangement merge batcher, the batcher does work merging data that won't be read for a long time. Bucketing buffers the future updates in a logarithmic `BucketChain` and releases them as the frontier advances.

The previous source-level bucketing applied to all imports regardless of whether the source actually produced future updates. This change targets bucketing precisely at `ArrangeBy` nodes that receive future updates from upstream temporal MFPs.

### Description

* Adds `MapFilterProject::has_temporal_predicates()` to detect MFPs containing `mz_now()`.
* Threads a `has_future_updates` analysis bit through LIR plan lowering. Every node knows whether its output may carry future-stamped updates.
* On `PlanNode::ArrangeBy` / `Expr::ArrangeBy`, stores the rendering decision as `strategy: ArrangementStrategy` (`Direct` or `TemporalBucketing`) — action-named, picked by the lowering and consumed by the renderer.
* In `ensure_collections`, when the strategy is `TemporalBucketing` and `ENABLE_COMPUTE_TEMPORAL_BUCKETING` is on, inserts bucketing in front of the arrangement. Both the collection-from-scratch path (via `as_collection_core`) and the pre-existing-collection path are covered, with a guard against double-bucketing.
* Removes the old source-level bucketing operator.
* Adds a `MaybeBucketByTime` trait sibling to `RenderTimestamp` that dispatches bucketing per timestamp type: real work for `mz_repr::Timestamp`, no-op for `Product<…>` in iterative scopes (follow-up). Bucketing is an arrangement-time concern, so the dispatch lives outside `RenderTimestamp`.
* Adds a feature-benchmark scenario exercising a temporal filter.

Bucketing is gated on `ENABLE_COMPUTE_TEMPORAL_BUCKETING` (default false).

### DIFF
--- a/src/compute-types/src/explain/text.rs
+++ b/src/compute-types/src/explain/text.rs
@@ -488,6 +488,7 @@ impl Plan {
                 input,
                 input_mfp,
                 forms,
+                input_has_future_updates: _,
             } => {
                 ctx.indent.set();
                 if forms.raw && forms.arranged.is_empty() {
@@ -897,6 +898,7 @@ impl Plan {
                 input,
                 input_mfp,
                 forms,
+                input_has_future_updates: _,
             } => {
                 writeln!(f, "{}ArrangeBy{}", ctx.indent, annotations)?;
                 ctx.indented(|ctx| {
--- a/src/compute-types/src/explain/text.rs
+++ b/src/compute-types/src/explain/text.rs
@@ -898,7 +898,7 @@ impl Plan {
                 input,
                 input_mfp,
                 forms,
-                input_has_future_updates: _,
+                input_has_future_updates,
             } => {
                 writeln!(f, "{}ArrangeBy{}", ctx.indent, annotations)?;
                 ctx.indented(|ctx| {
@@ -906,6 +906,9 @@ impl Plan {
                         let key = mode.seq(key, None);
                         let key = CompactScalars(key);
                         writeln!(f, "{}input_key=[{}]", ctx.indent, key)?;
+                    }
+                    if *input_has_future_updates {
+                        writeln!(f, "{}input_has_future_updates=true", ctx.indent)?;
                     }
                     mode.expr(input_mfp, None).fmt_text(f, ctx)?;
                     forms.fmt_text(f, ctx)?;
--- a/src/compute-types/src/explain/text.rs
+++ b/src/compute-types/src/explain/text.rs
@@ -40,7 +40,7 @@ use crate::plan::reduce::{
     AccumulablePlan, BasicPlan, BucketedPlan, HierarchicalPlan, MonotonicPlan, SingleBasicPlan,
 };
 use crate::plan::threshold::ThresholdPlan;
-use crate::plan::{AvailableCollections, LirId, Plan, PlanNode};
+use crate::plan::{ArrangementStrategy, AvailableCollections, LirId, Plan, PlanNode};
 
 impl DisplayText<PlanRenderingContext<'_, Plan>> for Plan {
     fn fmt_text(
@@ -488,7 +488,7 @@ impl Plan {
                 input,
                 input_mfp,
                 forms,
-                input_has_future_updates: _,
+                strategy: _,
             } => {
                 ctx.indent.set();
                 if forms.raw && forms.arranged.is_empty() {
@@ -898,7 +898,7 @@ impl Plan {
                 input,
                 input_mfp,
                 forms,
-                input_has_future_updates,
+                strategy,
             } => {
                 writeln!(f, "{}ArrangeBy{}", ctx.indent, annotations)?;
                 ctx.indented(|ctx| {
@@ -907,8 +907,8 @@ impl Plan {
                         let key = CompactScalars(key);
                         writeln!(f, "{}input_key=[{}]", ctx.indent, key)?;
                     }
-                    if *input_has_future_updates {
-                        writeln!(f, "{}input_has_future_updates=true", ctx.indent)?;
+                    if !matches!(strategy, ArrangementStrategy::Direct) {
+                        writeln!(f, "{}strategy={:?}", ctx.indent, strategy)?;
                     }
                     mode.expr(input_mfp, None).fmt_text(f, ctx)?;
                     forms.fmt_text(f, ctx)?;
--- a/src/compute-types/src/plan/lowering.rs
+++ b/src/compute-types/src/plan/lowering.rs
@@ -346,6 +346,7 @@ impl Context {
                                             input: body,
                                             input_mfp,
                                             forms,
+                                            input_has_future_updates: false,
                                         }
                                         .as_plan(inner_lir_id),
                                     ),
@@ -359,6 +360,7 @@ impl Context {
                                     input: Box::new(lir_value),
                                     input_mfp,
                                     forms,
+                                    input_has_future_updates: false,
                                 }
                                 .as_plan(lir_id)
                             }
@@ -876,6 +878,7 @@ This is not expected to cause incorrect results, but could indicate a performanc
                             input: Box::new(input),
                             input_mfp,
                             forms,
+                            input_has_future_updates: false,
                         }
                         .as_plan(lir_id),
                         input_keys,
@@ -1058,6 +1061,7 @@ This is not expected to cause incorrect results, but could indicate a performanc
                     input,
                     input_mfp,
                     mut forms,
+                    input_has_future_updates,
                 },
             lir_id,
         } = plan
@@ -1071,6 +1075,7 @@ This is not expected to cause incorrect results, but could indicate a performanc
                 input,
                 input_mfp,
                 forms,
+                input_has_future_updates,
             }
             .as_plan(lir_id)
         } else {
@@ -1090,6 +1095,7 @@ This is not expected to cause incorrect results, but could indicate a performanc
                 input: Box::new(plan),
                 input_mfp,
                 forms: collections,
+                input_has_future_updates: false,
             }
             .as_plan(lir_id)
         }
--- a/src/compute-types/src/plan/lowering.rs
+++ b/src/compute-types/src/plan/lowering.rs
@@ -29,6 +29,17 @@ use crate::plan::threshold::ThresholdPlan;
 use crate::plan::top_k::TopKPlan;
 use crate::plan::{AvailableCollections, GetPlan, LirId, Plan, PlanNode};
 
+/// The result of lowering a [`MirRelationExpr`] to a [`Plan`].
+struct LoweredExpr {
+    /// The lowered plan.
+    plan: Plan,
+    /// The arrangement keys that the plan is certain to produce.
+    keys: AvailableCollections,
+    /// Whether the plan's output may contain updates at future timestamps,
+    /// e.g., from a temporal MFP using `mz_now()`.
+    has_future_updates: bool,
+}
+
 pub(super) struct Context {
     /// Known bindings to (possibly arranged) collections.
     arrangements: BTreeMap<Id, AvailableCollections>,
@@ -106,7 +117,11 @@ impl Context {
         let mut objects_to_build = Vec::with_capacity(desc.objects_to_build.len());
         for build in desc.objects_to_build {
             self.debug_info.id = build.id;
-            let (plan, keys, has_future_updates) = self.lower_mir_expr(&build.plan)?;
+            let LoweredExpr {
+                plan,
+                keys,
+                has_future_updates,
+            } = self.lower_mir_expr(&build.plan)?;
 
             self.arrangements.insert(Id::Global(build.id), keys);
             if has_future_updates {
@@ -147,10 +162,7 @@ impl Context {
     ///
     /// An empty list of arrangement keys indicates that only a `Collection` stream can
     /// be assumed to exist.
-    fn lower_mir_expr(
-        &mut self,
-        expr: &MirRelationExpr,
-    ) -> Result<(Plan, AvailableCollections, bool), String> {
+    fn lower_mir_expr(&mut self, expr: &MirRelationExpr) -> Result<LoweredExpr, String> {
         // This function is recursive and can overflow its stack, so grow it if
         // needed. The growth here is unbounded. Our general solution for this problem
         // is to use [`ore::stack::RecursionGuard`] to additionally limit the stack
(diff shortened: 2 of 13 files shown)


src/compute/src/render/context.rs
```diff
@@ -777,6 +780,8 @@ where
         input_mfp: MapFilterProject,
         until: Antichain<mz_repr::Timestamp>,
         config_set: &ConfigSet,
+        input_has_future_updates: bool,
+        as_of: Antichain<mz_repr::Timestamp>,
```
teskje, line 798: Nit: Move `as_of` before `until`.

teskje (commented):
This make sense to me, though somebody who understand the lowering should probably review that part, as it's also the part that has the most potential for logic bugs.

src/compute-types/src/plan/lowering.rs
```diff
@@ -261,14 +268,21 @@ impl Context {
                     AvailableCollections::new_raw()
                 };
 
+                let has_future_updates = match &plan {
+                    GetPlan::Arrangement(_, _, mfp) | GetPlan::Collection(mfp) => {
+                        mfp.has_temporal_predicates()
+                    }
+                    GetPlan::PassArrangements => self.has_future_updates.contains(id),
+                };
```
def-, line 276:
```suggestion
                let has_future_updates = self.has_future_updates.contains(id) || match &plan {
                    GetPlan::Arrangement(_, _, mfp) | GetPlan::Collection(mfp) => {
                        mfp.has_temporal_predicates()
                    }
                    GetPlan::PassArrangements => self.has_future_updates.contains(id),
                };
```
Otherwise `Get` with a non-temporal MFP drops `has_future_updates` from the underlying binding

src/compute-types/src/plan/lowering.rs
```diff
@@ -667,7 +694,7 @@ This is not expected to cause incorrect results, but could indicate a performanc
                             }
                             .as_plan(lir_id),
                         );
-                        *input_plan = self.arrange_by(raw_plan, missing, input_keys, arity);
+                        *input_plan = self.arrange_by(raw_plan, missing, input_keys, arity, false);
```
def-, line 697: Why is it just `false` here? could we propagate `has_future_updates`? (Same in other places calling `arrange_by`)

src/compute/src/render/context.rs
antiguru: Done.

src/compute-types/src/plan/lowering.rs
antiguru: Good catch — fixed. Now `self.has_future_updates.contains(id)` is OR'd in unconditionally, so a Get with a non-temporal MFP still inherits the flag from the underlying binding (PassArrangements branch is now `false` since the top-level check already covers it).

src/compute-types/src/plan/lowering.rs
antiguru: Fixed — `Join`, `TopK`, `Threshold`, and `lower_reduce` now thread the input's `has_future_updates` through to `arrange_by` and into their own return value, matching what `Negate`/`Union`/`ArrangeBy` already did.

src/compute/src/render.rs
```diff
@@ -1502,6 +1500,18 @@ impl RenderTimestamp for mz_repr::Timestamp {
...
         self.saturating_sub(1)
     }
+    fn maybe_apply_temporal_bucketing<'scope>(
+        stream: StreamVec<'scope, Self, (Row, Self, Diff)>,
+        as_of: Antichain<mz_repr::Timestamp>,
+        summary: mz_repr::Timestamp,
+    ) -> VecCollection<'scope, Self, Row, Diff> {
+        use crate::extensions::temporal_bucket::TemporalBucketing;
+        use differential_dataflow::AsCollection;
+        use timely::container::CapacityContainerBuilder;
```
antiguru: Move to top-level imports.

src/compute/src/render.rs
```diff
@@ -1531,6 +1541,15 @@ impl RenderTimestamp for Product<mz_repr::Timestamp, PointStamp<u64>> {
         }
         Product::new(self.outer.saturating_sub(1), PointStamp::new(vec))
     }
+    fn maybe_apply_temporal_bucketing<'scope>(
+        stream: StreamVec<'scope, Self, (Row, Self, Diff)>,
+        _as_of: Antichain<mz_repr::Timestamp>,
+        _summary: mz_repr::Timestamp,
+    ) -> VecCollection<'scope, Self, Row, Diff> {
+        use differential_dataflow::AsCollection;
```
antiguru: Move to top-level import.

src/compute-types/src/plan/lowering.rs
```diff
@@ -143,7 +150,7 @@ impl Context {
     fn lower_mir_expr(
         &mut self,
         expr: &MirRelationExpr,
-    ) -> Result<(Plan, AvailableCollections), String> {
+    ) -> Result<(Plan, AvailableCollections, bool), String> {
```
antiguru: Introduce a type for `(Plan, AvailableCollections, bool)` and use here and elsewhere.

src/compute/src/render.rs
antiguru: Done.

src/compute/src/render.rs
antiguru: Done.

src/compute-types/src/plan/lowering.rs
antiguru: Introduced `LoweredExpr { plan, keys, has_future_updates }` and threaded it through `lower_mir_expr`, `lower_mir_expr_stack_safe`, and `lower_reduce`. The previously-opaque bool now has a name at every destructure site.

antiguru:
## Feature benchmark results

Two scenarios in `misc/python/materialize/feature_benchmark/scenarios/temporal.py`. `enable_compute_temporal_bucketing=true` is set globally in mzcompose, so on both runs the dyncfg is active — main exercises the old source-level bucketing; this PR exercises the ArrangeBy-targeted bucketing.

### `TemporalFilterIndexed` — index hydration

10M rows in a materialized view, then `CREATE INDEX` over a temporal-filtered view on top. Window times `CREATE INDEX` → first peek.

| metric          | PR     | main   | delta              |
|-----------------|--------|--------|--------------------|
| wallclock       | 11.92s | 14.73s | **19.1% faster**   |
| memory_mz       | 516MB  | 535MB  | 3.6% less          |
| memory_clusterd | 961MB  | 862MB  | 11.5% more         |

### `TemporalFilterSustainedInsert` — per-insert latency

Pre-warms the indexed temporal view, then drives 5×100k inserts and a trailing `SELECT COUNT(*)`. The merge batcher's per-insert incremental work lands on the critical path of that SELECT.

| metric          | PR    | main  | delta              |
|-----------------|-------|-------|--------------------|
| wallclock       | 4.64s | 5.33s | **12.9% faster**   |
| memory_mz       | 757MB | 754MB | flat               |
| memory_clusterd | 912MB | 806MB | 13.1% more         |

### Why these scenarios

The original 1M-row `TemporalFilterIndexed` showed only ~3% deltas on every metric — well within noise. Two reasons:

1. **Window doesn't observe the work being saved.** Index becomes ready at `as_of`. Future-stamped retractions are *behind* `as_of`, so the merge-batcher consolidation work on them is amortized into background time the window doesn't cover. Bumping to 10M moved enough work onto the hydration path to surface.
2. **Per-update savings only matter at scale.** The merge batcher's progressive-merge cost is roughly `O(log N)` per element, so halving the spine height (bucketing diverts future retractions to the BucketChain) saves a small constant per element. At 1M rows the cumulative saving is below noise; at 10M it isn't.

`TemporalFilterSustainedInsert` is a more direct test: each post-hydration insert flows through the temporal MFP, lands two updates in the arrangement (or one + bucket chain on this PR), and the trailing SELECT forces sync. Smaller spine → faster per-insert merge → measurably shorter window.

### Memory tradeoff

`memory_clusterd` is consistently ~11–13% higher on the PR — about 100MB on a 10M-row workload, roughly 10 bytes per future-stamped update. The BucketChain stores updates in a columnar merge batcher (`Vec<(D, T, R)>` chunks via `ColumnationChunker`/`ColInternalMerger`) which is slightly less compact than the arrangement's tiered `RowRowSpine` batches. So we trade a small steady-state arrangement-memory increase for the CPU savings.

Both memory deltas sit within feature-benchmark's noise bands (50% for `memory_clusterd`, 20% for `memory_mz`), so no regression gate trips — but the tradeoff is real and worth flagging.

src/compute-types/src/explain/text.rs
```diff
@@ -897,6 +898,7 @@ impl Plan {
                 input,
                 input_mfp,
                 forms,
+                input_has_future_updates: _,
```
ggevay: We should definitely print this at least here, where we are handling `AS VERBOSE TEXT` given to `EXPLAIN`.

We could also consider whether we'd like to print it in the default `EXPLAIN` as well (above in the same file).

src/compute-types/src/plan/lowering.rs
```diff
@@ -527,17 +598,18 @@ impl Context {
...
                             input: Box::new(input),
                             exprs: exprs.clone(),
                             func: func.clone(),
                             mfp_after: mfp,
                         }
                         .as_plan(lir_id),
-                        AvailableCollections::new_raw(),
-                    )
+                        keys: AvailableCollections::new_raw(),
+                        has_future_updates: input_future,
```
ggevay: Would this need ORing with `mfp.has_temporal_predicates()`?

ggevay: (Btw. Join and Reduce also seem at first glance that they might have this bug, but I think they don't actually have it, because they absorb only non-temporal MFPs. Maybe worth a comment on their `LoweredExpr` construction that this is why we don't need to check the absorbed MFPs there.)

src/compute-types/src/plan/lowering.rs
```diff
@@ -346,6 +402,7 @@ impl Context {
                                             input: body,
                                             input_mfp,
                                             forms,
+                                            input_has_future_updates: false,
```
ggevay: Maybe we should put a real thing here rather than hardcoding `false`. Even though we are inside an iterative scope, so there would be no effect, but it still looks a bit weird that this is hardcoded to `false.` (And the same a few lines below.)

src/compute-types/src/plan/lowering.rs
```diff
@@ -273,24 +312,37 @@ impl Context {
...
                         body: Box::new(body),
                     }
                     .as_plan(lir_id),
-                    b_keys,
-                )
+                    keys: b_keys,
+                    has_future_updates: b_future,
+                }
             }
             MirRelationExpr::LetRec {
```
ggevay, line 357: I'm afraid we'd need a fixpoint loop here to correctly compute `has_future_updates` for the final result of a `LetRec`. The current code is just doing one sequential sweep over the let bindings, which misses it when an earlier let binding should get `has_future_updates` only from a later binding (and then the body uses only the earlier let binding).

See `Bundle::analyse_optimistic` for how the MIR Analysis framework does this in general. This would map to the `has_future_updates` computation with the slightly counter-intuitive `true ⊑ false` lattice.

ggevay: Btw., maybe it's worth considering a rewrite of the whole implementation to use the `Analysis` framework. That is, maybe we could create a `HasFutureUpdates` `Analysis`, and rely on that during the lowering. This would solve the above LetRec problem for free. I'm not sure if it would have other benefits.

ggevay: (Or if you'd like to avoid either manually implementing fixpointing here or rewriting with the Analysis framework, it might be ok to just pessimistically set `has_future_updates` to true at the end of a `LetRec`.)

src/compute/src/render/context.rs
```diff
@@ -757,19 +762,32 @@ impl<'scope, T: RenderTimestamp> CollectionBundle<'scope, T> {
...
-                input_key.map(|k| (k, None)),
-                until,
-                config_set,
-            ));
+            let (oks, errs) =
+                self.as_collection_core(input_mfp, input_key.map(|k| (k, None)), until, config_set);
+            // Apply temporal bucketing if the input may contain future updates.
+            // This path fires when the collection must be formed from scratch
+            // (e.g., from an arrangement via as_collection_core).
+            let oks = if input_has_future_updates && ENABLE_TEMPORAL_BUCKETING.get(config_set) {
```
ggevay: Maybe we'd want to also gate this on an arrangement actually being made (i.e., `!collections.arranged.is_empty()`)? Maybe I'm misunderstanding something, but it seems to me that the below bucketing can currently go to waste when we are making just a raw collection, but no arrangement.

ggevay (commented):
It's great that we are solving this! I wrote some comments.

ggevay:
One more thought: It could be good to do a full CI (including Nightly) with the flag turned on by default. That would exercise the code much more.

Edit: Oh, I see it's already on in `get_minimal_system_parameters`. Then it's fine.

Edit 2: Could you do a renaming to make the const's name (`ENABLE_TEMPORAL_BUCKETING`) consistent with the flag name (`enable_compute_temporal_bucketing`)?

DAlperin (approved):
This seems good to me. It kinda spreads some unfortunate complexity around lowering that I agree with @ggevay should maybe live in some sort of analysis pass, but it's not blocking. Probably a bigger convsersation around for whom/when do we want to enable this but it should per perf neutral when disabled!

antiguru:
Thanks for the reviews!

src/compute-types/src/plan/lowering.rs
antiguru: Fixed, thank you.

src/compute-types/src/plan/lowering.rs
antiguru: As you're pointing out the precise value is hard to compute, so we're using an approximation: If the inputs or any operator in the recursive part have temporal filters, assume that we potentially output future updates.

Not a big concern that it's not as accurate as it could be, we're not using WMR in many places.

PR #35937, opened 2026-04-10
compute: diagnostic log when dropping per-replica read holds

teskje:
When a replica is removed, we drop its per-replica input read holds. If the corresponding global read holds have already been released/downgraded (e.g. because the dataflow was dropped by a user), the per-replica holds are the last line of defense against compaction. The dataflow might still be in flight at the replica, and dropping its input read holds can allow the storage inputs to compact past a its as_of, causing the replica to panic when it tries to render the dataflow.

We believe that this is the cause for "cannot serve requested as_of" panics observed in the wild. This commit adds a warning log to help validate that theory.

### Motivation

Part of diagnosing https://github.com/MaterializeInc/incidents-and-escalations/issues/39


ggevay (approved):

teskje:
TFTR!

PR #35938, opened 2026-04-10
persist/pubsub: fix connection leak on gRPC stream error

teskje:
When the client's gRPC response stream errors, the reconnect loop drops the tonic `Channel` but hyper's background task keeps running: It's blocked polling the `broadcast_messages` async stream, which holds a live `BroadcastStream` receiver. This keeps the HTTP2 connection open, leaking one connection per reconnect.

Fix this by giving the stream a cancellation token. When the reconnect loop drops `cancel_tx`, the stream observes it via `select!` and terminates, allowing hyper to close the HTTP2 connection.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/11276

### Verification

There is a repro test in the issue comments and I verified that it passes with this fix, i.e. it doesn't reproduce the leak anymore. It'd be great to include the test in this PR, but it requires patching the `max_decoding_message_size` to allow injecting messages that are too large and produce decoding errors. Is there another way to provoke receive errors? Alternatively, we could make the `max_decoding_message_size` a dyncfg (currently it's hardcoded to `usize::MAX`) but it would be a bit of plumbing that only exists for this one test.

### DIFF
--- a/src/persist-client/src/rpc.rs
+++ b/src/persist-client/src/rpc.rs
@@ -347,10 +347,18 @@ impl GrpcPubSubClient {
                 .broadcast_recv_lagged_count
                 .clone();
 
+            // `client.pub_sub(...)` starts a hyper background task reading from the
+            // `broadcast_messages` stream, to serve the HTTP2 connection. The broadcast stream
+            // doesn't normally terminate, which means the HTTP2 connection doesn't terminate
+            // either. We set up a cancelation token to force termination and avoid a connection
+            // leak.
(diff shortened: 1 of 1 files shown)


antiguru (approved):
Checks out! Maybe wait for someone else to review, too.

DAlperin (approved):
LGTM. Slightly bummed not to have the test in this PR but it would be kind of a pain, I agree. Thanks for fixing!

teskje:
TFTRs!

PR #35948, opened 2026-04-10
Show some notices if pod metrics are not enabled

Alphadelta14:
### Motivation

We currently default observability.podMetrics.enabled to false in our operator helm chart.
This is also the default in our environmentd CLI.

The unfortunate side effect of this is that resource utilization graphs come up entirely empty
unless this is turned on.

### Description

This adds a log message (once at startup) indicating that metrics are disabled and their implication.
In the case of our resource utilization graph, this attempts to show an error message.

### Verification

Need to figure out how to run a console with a port-forwarded cluster...

This is my first console PR... sorry in advance!

<details><summary>Details</summary>
<p>

Working fine with metrics (existing behavior)
image

When there are no metrics
image

</p>
</details>

### DIFF
--- a/console/src/platform/clusters/ClusterOverview.tsx
+++ b/console/src/platform/clusters/ClusterOverview.tsx
@@ -47,7 +47,10 @@ import {
   UtilizationGraph,
 } from "./ClusterOverview/UtilizationGraph";
 import { ClusterParams } from "./ClusterRoutes";
-import { CLUSTERS_FETCH_ERROR_MESSAGE } from "./constants";
+import {
+  CLUSTER_METRICS_UNAVAILABLE_MESSAGE,
+  CLUSTERS_FETCH_ERROR_MESSAGE,
+} from "./constants";
 import LargestMaintainedQueries from "./LargestMaintainedQueries";
 import { useReplicaUtilizationHistory } from "./queries";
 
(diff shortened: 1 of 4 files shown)


leedqin (changes requested):
The` === 0 `checks are dead code. The data mapping in [queries.ts:505-508](https://github.com/MaterializeInc/materialize/blob/main/console/src/platform/clusters/queries.ts#L505-L508) uses truthiness checks:

`cpuPercent: maxCpu?.percent ? maxCpu.percent * 100 : null,`

Since 0 is falsy, it gets mapped to null before it ever reaches this component. 
This simplifies to:

```
const clusterHasNoMetrics =
  graphData?.length &&
  graphData.every((d) =>
    d.data.every(
      (point) => !point.cpuPercent && !point.memoryPercent && !point.diskPercent,
    ),
  );
```

leedqin (changes requested):
Ideally don't want to do an early return for an error or the whole page breaks and it's unideal user experience. I would recommend moving this error in the conditional here in the existing conditional [ClusterOverview.tsx](https://github.com/MaterializeInc/materialize/blob/main/console/src/platform/clusters/ClusterOverview.tsx#L174-L176)

 Something like:

```

{isLoading ? (
  <Spinner ... />
) : clusterHasNoMetrics ? (
  <ErrorBox message={CLUSTER_METRICS_UNAVAILABLE_MESSAGE} />
) : data ? (
  <Grid> ... graphs ... </Grid>
) : null}

```
This would only show the errors in place of the graphs

leedqin:
Thank you for doing this! explains why I wasn't able to see the Memory Utilization graphs in local mz instance for last week.

leedqin (approved):
LGTM! thanks for the changes

PR #35967, opened 2026-04-11
Remove continual tasks (CT) feature

antiguru:
## Summary

- Remove the entire continual tasks (CT) feature from the codebase
- Add catalog migration v81→v82 that deletes any existing CT items and comments
- Preserve audit log `ObjectType::ContinualTask` for backward compatibility with persisted entries

## Details

**Catalog migration (v81→v82):**
- Iterates over all persisted items and deletes any whose `create_sql` starts with `CREATE CONTINUAL TASK`
- Deletes any comments on continual task objects
- Builtin CT removal is handled automatically by the builtin item migration (they're no longer in `BUILTINS_STATIC`)

**What was removed (~4,600 lines deleted across 99 files):**
- SQL parser: `CREATE/DROP/SHOW/ALTER/COMMENT ON CONTINUAL TASK` syntax, `ContinualTaskStmt`, `CreateContinualTaskSugar` (TRANSFORM/RETAIN)
- SQL planning: `CreateContinualTaskPlan`, feature flags (`enable_continual_task_create/transform/retain`), `max_continual_tasks` session var, `CatalogItemType::ContinualTask`, `ObjectType::ContinualTask`, `CommentObjectId::ContinualTask`
- Adapter: `sequence_create_continual_task`, CT bootstrap/dataflow logic, CT quota checks, `ExecuteResponse::CreatedContinualTask`
- Compute: `ContinualTaskCtx`, `ContinualTaskConnection`, CT sink rendering, `render/continual_task.rs`
- Catalog builtins: `BuiltinContinualTask`, 3 system CTs (`mz_cluster_replica_metrics_history_ct`, `mz_cluster_replica_status_history_ct`, `mz_wallclock_lag_history_ct`), `mz_continual_tasks` table/view/indexes
- `BuiltinsConfig` emptied (was only used for `include_continual_tasks`); `BUILTINS::iter()` no longer takes a config parameter
- All CT-specific test files (9 SLT files, 1 testdrive file, 1 parser test file)

**What was preserved:**
- `mz_audit_log::ObjectType::ContinualTask` — audit log entries are immutable
- `audit_log_event_v1::ObjectType::ContinualTask` in catalog protos — same reason
- Historical catalog proto snapshots (v74–v81) — frozen for migration support

https://claude.ai/code/session_01Rx3uSReMsvp3zDTxnXuNUd


def-:
Probably fixes https://github.com/MaterializeInc/database-issues/issues/11284 & https://github.com/MaterializeInc/database-issues/issues/11279

def-:
I'd like to see a platform check that creates a CT if the version is old enough to allow it and then checks that we handle the migration properly, could be done with the existing tests in misc/python/materialize/checks/all_checks/continual_task.py

misc/wasm/Cargo.lock
```diff
@@ -160,9 +160,9 @@ checksum = "1e748733b7cbc798e1434b6ac524f0c1ff2ab456fe201501e6497c8417a4fc33"
 
 [[package]]
 name = "bytesize"
-version = "2.1.0"
+version = "2.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f5c434ae3cf0089ca203e9019ebe529c47ff45cefe8af7c85ecb734ef541822f"
+checksum = "6bd91ee7b2422bcb158d90ef4d14f75ef67f340943fc4149891dcce8f8b972a3"
```
teskje: Seems unrelated?

src/adapter/src/catalog/open.rs
```diff
@@ -164,16 +164,7 @@ impl Catalog {
...
-                    // This will fall back to the default in code (false) if we timeout on the
-                    // initial LD sync. We're just using this to get some additional testing in on
-                    // CTs so a false negative is fine, we're only worried about false positives.
-                    include_continual_tasks: get_dyncfg_val_from_defaults_and_remote(
-                        &config.system_parameter_defaults,
-                        config.remote_system_parameters.as_ref(),
-                        &ENABLE_CONTINUAL_TASK_BUILTINS,
-                    ),
-                },
+                builtins_cfg: BuiltinsConfig {},
```
teskje: Looks like we can delete `BuiltinsConfig {}`?

src/catalog/src/durable/upgrade/v81_to_v82.rs
```diff
@@ -0,0 +1,141 @@
...
+
+use crate::durable::upgrade::MigrationAction;
+use crate::durable::upgrade::objects_v81 as v81;
+use crate::durable::upgrade::objects_v82 as v82;
+
+/// Remove all continual task items and their comments from the catalog.
+///
+/// Continual tasks are being removed from the product. This migration deletes
+/// any user-created continual task items and any comments on continual tasks.
+/// Builtin continual tasks are handled separately by the builtin item migration.
```
teskje: I think this might cause issues when there are things depending on the continual tasks? Rather than trying to remove these too, we could just panic instead, since CTs should not be present in any production environments, and if they are present elsewhere, we can drop them. This precludes Dennis' request for having a test that the migration works, of course.

src/catalog/src/builtin.rs
```diff
@@ -14739,9 +14541,8 @@ pub static BUILTIN_LOG_LOOKUP: LazyLock<HashMap<&'static str, &'static BuiltinLo
 pub static BUILTIN_LOOKUP: LazyLock<
     HashMap<SystemObjectDescription, (usize, &'static Builtin<NameReference>)>,
 > = LazyLock::new(|| {
-    // BUILTIN_LOOKUP is only ever used for lookups by key, it's not iterated,
-    // so it's safe to include all of them, regardless of BuiltinConfig. We
-    // enforce this statically by using the mz_ore HashMap which disallows
+    // BUILTIN_LOOKUP is only ever used for lookups by key, it's not iterated.
+    // We enforce this statically by using the mz_ore HashMap which disallows
     // iteration.
```
teskje: I think the config doesn't make much sense anymore, since there isn't a reason why we need to forbid iteration.

src/compute-client/src/controller.rs
```diff
@@ -1001,8 +995,7 @@ impl ComputeController {
 
         'source: for id in dataflow.imported_source_ids() {
             // We first check whether the id is backed by a compute object, in which case we use
-            // the time dependence we know. This is true for materialized views, continual tasks,
-            // etc.
+            // the time dependence we know. This is true for materialized views, etc.
```
teskje: That was an issue before, but I don't know what "etc" refers to! MVs are the only thing that have both a compute and a storage collection. (Also would prefer to not talk about "MVs" in the context of the compute controller, since it's an adapter-level concept. We call it "storage sink" here.)

teskje (commented):
Got halfway through, but need to get lunch now!

doc/user/content/releases/v26.20.md
teskje, line 1: Seems unintentional!

src/arrow-util/src/reader.rs
teskje, line 1: This reverts parts of https://github.com/materializeInc/materialize/pull/36032. Worth double-checking if other merge skew made it in as well.

src/persist/src/azure.rs
teskje, line 1: Merge merge skew

src/sql-parser/src/parser.rs
```diff
@@ -1985,17 +1985,6 @@ impl<'a> Parser<'a> {
         {
             self.parse_create_materialized_view()
                 .map_parser_err(StatementKind::CreateMaterializedView)
-        } else if self.peek_keywords(&[CONTINUAL, TASK]) {
```
teskje, line 1988: Anything we can delete from `keywords.txt` now?

src/testdrive/src/action/s3.rs
teskje, line 1: Merge skew

src/testdrive/src/action.rs
teskje, line 1: Merge skew

LICENSE
teskje, line 1: Merge skew

teskje (commented):
Looks good, except for the merge skew!

misc/python/materialize/checks/all_checks/continual_task.py
```diff
@@ -10,145 +10,68 @@
...
-                > SELECT * FROM audit_log
-                1
-                6
-                21
-           """))
-
-
-class StreamTableJoinCT(Check):
-    """Continual Task for stream table join"""
+    Note: we cannot create CTs from this check because the new testdrive binary
```
def-, line 27: Oh wow, when did that start?! How can we test compatibility against old versions then?

def-: I had no idea, we always had that. This means the test, and all tests in continual_task.py are now useless. They should instead use `$ postgres-execute` at least, but I guess with the intentional `found continual task "ct_audit_log" in catalog; drop all continual tasks before upgrading` panic they can't do much, so we can remove them after all.

def- (commented):
How do we know that no one is using continual tasks in self-managed btw? I'm wondering if the panic is the best approach here.

antiguru:
This should be good for another look.

ggevay:
> How do we know that no one is using continual tasks in self-managed btw?

We provide no guarantees for such highly experimental features, so I'd say it's fine. But there might be a higher chance of some staging env or prod canary having CTs.

Anyhow, if an env has CTs, this would come out during the Friday upgrade-check (or when bringing up the read-only env during a 0dt upgrade), at which point one can still manually drop the CT in the old env, so we wouldn't get into an unrecoverable situation, right?

ggevay (approved):
Thank you! This will alleviate some complications in the catalog implications work!

One more minor thing: Can we also remove `force_source_non_monotonic`?

src/sql-lexer/src/keywords.txt
```diff
@@ -108,7 +108,6 @@ Confluent
 Connection
 Connections
 Constraint
-Continual
```
DAlperin, line 111: can we not nuke `Task` here too? Did a quick grep through parser.rs and not sure we use it anywhere else.

DAlperin (approved):

PR #35968, opened 2026-04-11
Use vendored `Columnation` types

frankmcsherry:
When we introduced the vendored `Columnation` container types, we didn't fully re-orient to target them vs the ones exported by differential. That will be a break when it happens, so we fix it here.


antiguru (approved):
Thank you!

PR #35969, opened 2026-04-11
Inline `reduce_pair`

frankmcsherry:
The `reduce_pair` method was a helper that allowed you to call `reduce` twice. That's all it did. It is used twice, and much shorter when the two calls are inlined.


antiguru (approved):
Lgtm

PR #35970, opened 2026-04-11
Remove genericity over `Allocate` trait

frankmcsherry:
Some amount of code is generic over `A: Allocate`, which allows swapping the timely channel allocator. We only instantiate this with the most general form, `Generic`. We can remove the false generality and cut out our use of this trait.


antiguru (approved):
Let's do it!

PR #35972, opened 2026-04-11
De-genericize `Context`

frankmcsherry:
Compute's `Context` type has a generic timestamp parameter, and a second one that corresponds to the "root" timestamp type. That type is always `mz_repr::Timestamp`. The genericity existed to signal the relationship, but .. it's arguably not carrying its weight. This PR removes it and simplifies things.


DAlperin (approved):
Yep! Pro not lying to ourselves about timestamps if we have code that only is ever used one way.

antiguru (approved):

PR #35973, opened 2026-04-11
Remove `Timestamp` genericity from `TimelineState`

frankmcsherry:
Removes genericity over timestamp types for `TimelineState` which is only instantiated with `mz_repr::Timestamp`.


teskje (approved):
🎉

antiguru (approved):

PR #35974, opened 2026-04-11
Remove generic timestamp from `ReadHolds`

frankmcsherry:
This instantiates the `T: Timestamp` argument for `ReadHolds<T>` with `mz_repr::Timestamp`.

A next step could be to do the same for `ReadHold<T>`, but it is used by other types (e.g. `ExportState<T>`) which introduce a cascade of specialization. It's probably all good, but attempting to do this piecemeal.


teskje (approved):
🎉

src/adapter/src/coord/read_policy.rs
antiguru: Derive `Default`.

antiguru (approved):

PR #35978, opened 2026-04-11
Convert `Scheduler` to `Activations`

frankmcsherry:
Our compute logging uses `S: Scheduler` when it only needs `Activations`. This PR simplifies it and removes the dependence on the more complex trait (which will be harder to access in future TD).


antiguru (approved):
🚀

PR #36024, opened 2026-04-13
cargo test: Detect incremental build errors & silence clusterd not existing

def-:
Noticed in https://materializeinc.slack.com/archives/C08A62E0751/p1775926643153079

Review with `-w`


frankmcsherry (approved):
I can vouch for the feature, but cannot speak to understanding the implementation.

PR #9698, opened 2021-12-20
expr,sql: support casts from integer types to/from oid

andrioni:
These three casts are supported by Postgres:
- `smallint` -> `oid` (implicit)
- `bigint` -> `oid` (implicit)
- `oid` -> `bigint` (assignment)

The main motivation here is supporting bigint -> oid, as DBeaver required it.

### Motivation

  * This PR fixes a recognized bug: #9695 

### Checklist

- [x] This PR has adequate test coverage / QA involvement has been duly considered.
- [x] This PR adds a release note for any [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/user/content/release-notes.md#what-changes-require-a-release-note).

PR #36030, opened 2026-04-13
sql: Fix casting oid to int64

def-:
Bug introduced in https://github.com/MaterializeInc/materialize/pull/9698


ggevay (approved):
Uhh, thanks!

PR #36032, opened 2026-04-13
[SS-103] Fix COPY FROM PARQUET unsorted map keys

def-:
Fixes https://github.com/MaterializeInc/database-issues/issues/11290 
Updates `ColReader::Map` logic to sort keys when reading in data.

### DIFF
--- a/test/testdrive/copy-from-parquet-unsorted-map-keys.td
+++ b/test/testdrive/copy-from-parquet-unsorted-map-keys.td
@@ -67,7 +67,7 @@ $ s3-set-presigned-url bucket=copyfroms3 key=parquet/unsorted_map.parquet var-na
 
 # Verify equality: maps from COPY FROM should equal maps inserted via SQL.
 # If unsorted, these will NOT match, demonstrating silent data corruption.
-> SELECT p.id, p.map_col = s.map_col AS maps_equal
+> SELECT p.id, p.map_col::text = s.map_col::text AS maps_equal
   FROM from_parquet p
   JOIN from_sql s ON p.id = s.id
   ORDER BY p.id;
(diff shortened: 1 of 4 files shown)


patrickwwbutler:
Had to update the checks in the test you created - it appears MZ doesn't support the equality operator for maps (at least string maps): `ERROR: operator does not exist: map[text=>text] = map[text=>text]`

patrickwwbutler (approved):
feels a bit suspicious reviewing my own changes so I won't merge, but I'm fairly confident this is good to go - also tested without the sort changes and the TD fails as expected

def- (commented):
The fix looks reasonable to me!

test/testdrive/copy-from-parquet-unsorted-map-keys.td
```diff
@@ -0,0 +1,88 @@
...
+# Use of this software is governed by the Business Source License
+# included in the LICENSE file at the root of this repository.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0.
+
+# Regression test for BUG #41: COPY FROM Parquet Map keys not sorted.
+#
+# Materialize's Datum::Map requires keys to be in ascending sorted order.
```
martykulma, line 12: This is a great clarifying comment about the sort - definitely ought to be in the source!

martykulma (approved):
Nice! The reasoning for the sort needs to be documented in the source, but lgtm otherwise.

🦈

test/testdrive/copy-from-parquet-unsorted-map-keys.td
ublubu: :+1: It'd be nice to (also) have it on the doc comment for `Datum::Map` (and/or `push_dict_with`?) if that type is never supposed to be constructed from unsorted values.

ublubu: I see it's already on `push_dict_with`:

https://github.com/MaterializeInc/materialize/blob/15f2490e38ad774a9fa9bf2dadeebb68a60cd147/src/repr/src/row.rs#L2349-L2351

def-: I'd be even happier if we can get rid of that foot gun. Seems way too easy to use wrongly.

PR #36033, opened 2026-04-13
adapter: Clean up secrets instead of leaking them on error

def-:
Bugs introduced in #27545 and #23551, two refactorings


def-:
Ready for review

aljoscha (approved):
I think these look good, but I'd say in the long run we need a periodic process that can clean these up. As of today, I believe it can happen that we create a secret in the controller but then crash, and so we're not cleaning things up.

PR #27641, opened 2024-06-13
storage/sink/kafka: Allow configuring compatibility levels for schema registry subjects

rjobanp:
### Motivation

  * This PR adds a known-desirable feature. Fixes https://github.com/MaterializeInc/materialize/issues/23794

This implements the 2nd part of the linked issue, allowing users to specify new options as part of the `CSR CONNECTION` in a `CREATE SINK` statement:
```
KEY COMPATIBILITY LEVEL
VALUE COMPATIBILITY LEVEL
```
which uses the CSR config api to set these values: https://docs.confluent.io/platform/7.6/schema-registry/develop/api.html#config

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] This PR includes the following [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note):
  -

PR #36034, opened 2026-04-13
ccsr: Fix error code for subject not found

def-:
See https://docs.confluent.io/platform/current/schema-registry/develop/api.html#errors for error codes

Caused by https://github.com/MaterializeInc/materialize/pull/27641

Nikhil noted this correctly in the original review: https://github.com/MaterializeInc/materialize/pull/27641#discussion_r1640358246

### DIFF
--- a/src/ccsr/src/client.rs
+++ b/src/ccsr/src/client.rs
@@ -546,7 +546,7 @@ impl From<UnhandledError> for GetSubjectConfigError {
         match err {
             UnhandledError::Transport(err) => GetSubjectConfigError::Transport(err),
             UnhandledError::Api { code, message } => match code {
-                404 => GetSubjectConfigError::SubjectNotFound,
+                40401 => GetSubjectConfigError::SubjectNotFound,
                 40408 => GetSubjectConfigError::SubjectCompatibilityLevelNotSet,
                 _ => GetSubjectConfigError::Server { code, message },
             },


src/ccsr/src/client.rs
```diff
@@ -546,7 +546,7 @@ impl From<UnhandledError> for GetSubjectConfigError {
         match err {
             UnhandledError::Transport(err) => GetSubjectConfigError::Transport(err),
             UnhandledError::Api { code, message } => match code {
-                404 => GetSubjectConfigError::SubjectNotFound,
+                40401 => GetSubjectConfigError::SubjectNotFound,
```
teskje, line 549: Are you sure "404" can never happen? In the linked docs I see some APIs that document "404 Not Found" but no sub-error codes.

def-: Only if the json body can't be parsed would we return the 404 error code, and would then wrongly consider it as a "SubjectNotFound". All the other error handlers in this file do it the same way.

teskje: For example, the [GET /config/{subject}](https://docs.confluent.io/platform/current/schema-registry/develop/api.html#get-compatibility-level-for-a-subject) API documents these errors:

```
Status Codes:
 * 404 Not Found – Subject not found
 * 500 Internal Server Error –
      Error code 50001 – Error in the backend datastore
```

What error code does the API return for "Subject not found", assuming the response can be parsed? It looks like it's 404. Are you suggesting there is a bug in the documentation or am I missing something?

Btw, for `PublishError` we also match on "409", not a five-digit error code. Is that also a bug?

def-: From what I can tell their documentation is kind of misleading. The code seems to always set the error code to 40401 for subject not found: https://github.com/confluentinc/schema-registry/blob/master/core/src/main/java/io/confluent/kafka/schemaregistry/rest/exceptions/Errors.java#L25

The 409 is also a bug, fixing.

teskje (approved):
Thanks!

PR #30494, opened 2024-11-15
Unified materialized binary

antiguru:
Compile a unified materialized binary for the emulator

The unified binary links both environmentd and clusterd, and uses the executable name to distinguish its behavior.

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #36035, opened 2026-04-13
clusterd: Keep HTTP loop alive on errors

def-:
Introduced in https://github.com/MaterializeInc/materialize/pull/30494

### DIFF
--- a/src/clusterd/src/lib.rs
+++ b/src/clusterd/src/lib.rs
@@ -41,7 +41,7 @@ use mz_storage_types::connections::ConnectionContext;
 use mz_txn_wal::operator::TxnsContext;
 use tokio::runtime::Handle;
 use tower::Service;
-use tracing::{Instrument, error, info, info_span};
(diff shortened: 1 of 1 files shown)


teskje (commented):
I have some questions:

* The commit message says this is to keep the HTTP loop alive during transient errors, but I don't see a check for the error type, so I think it applies to non-transient errors as well. Is this intentional?
* What kinds of transient errors do we expect from `TcpListener::accept` calls?
* The commit message also says it is to prevent replica restarts, but it doesn't look like the replica will restart if the internal_http loop exists. Am I missing something?

I agree that just exiting the loop and getting stuck with a broken internal-http connection isn't great, but I'm not sure retrying in a hot loop if the error is non-transient is much better.

def-:
I reverted us to the behavior we had with hyper before.

teskje (approved):
Makes sense!

PR #22262, opened 2023-10-09
 adapter: add Postgres-backed TimestampOracle

aljoscha:
Part of #22029 
Design document: https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/20230921_distributed_ts_oracle.md

This adds an initial, working version of a postgres-backed timestamp
oracle (really, it's CRDB-backed, of course...).

This adds:

- the new timestamp oracle implementation
- a system var `timestamp_oracle` for switching between the new
  'postgres' oracle and the (retroactively named) 'catalog' oracle
- a new (optional) argument to `environmentd`: `--timestamp-oracle-url`,
  that specifies the postgres connection string (including database and
  schema) where the oracle should be stored.

This is not yet meant to be used in production! Before that, we need to
address at least:

- add internal retries
- add metrics
- add optimizations that are required because oracle operations are now
  more expensive: batching of `read_ts` calls, improved batching of
  group commits (which do `write_ts`-related operations)
- involve QA, figure out any testing/nightly testing/rollout concerns
- teach cloud infra about the newly required postgres url for the
  oracle, including setting up the required database/schema

### Tips for reviewer

Commits have meaningful text, plus there's explanatory comments in the code. I left some WIP questions (largely for @danhhz because he has most context).

I left in the TODOs about retries, and some of the other things mentioned above so that it's easier to review the bare minimum structure of the new oracle code. I will open follow-up PRs for those.

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered.
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] This PR includes the following [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note):
  -

PR #36036, opened 2026-04-13
ts oracle: Only log on info after a few retries

def-:
Follow-up to https://github.com/MaterializeInc/materialize/pull/22262

### DIFF
--- a/src/timestamp-oracle/src/postgres_oracle.rs
+++ b/src/timestamp-oracle/src/postgres_oracle.rs
@@ -941,7 +941,7 @@ where
                         err.display_with_causes()
                     );
                 } else {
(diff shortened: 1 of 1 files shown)


def-:
Ready for review

src/timestamp-oracle/src/postgres_oracle.rs
```diff
@@ -941,7 +941,7 @@ where
                         err.display_with_causes()
                     );
                 } else {
-                    info!(
+                    debug!(
```
aljoscha, line 944: huh, the two branches were identical before, what happened there ... 🙈

aljoscha (approved):

PR #7522, opened 2021-07-25
sql: implement char, varchar

sploiselle:

PR #36037, opened 2026-04-13
types: Fix bpchar mapping

def-:
Broken in https://github.com/MaterializeInc/materialize/pull/7522

Nightly run: https://buildkite.com/materialize/nightly/builds/16149

### DIFF
--- a/src/pgrepr/src/types.rs
+++ b/src/pgrepr/src/types.rs
@@ -560,9 +560,8 @@ impl Type {
             postgres_types::Type::NUMERIC => Type::Numeric { constraints: None },
             postgres_types::Type::OID => Type::Oid,
             postgres_types::Type::TEXT => Type::Text,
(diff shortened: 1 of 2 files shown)


def-:
Ready for review

SangJunBak (approved):
This change makes sense to me. Couple questions:
1. What was the motivation of this inconsistency? Did this cause any errors? How was this found?
2. Noticed we don't handle `postgres_types::Type::CHAR_ARRAY` in `from_oid_and_typmod`. Wonder if we should add that in too?

def-:
It's probably not a major issue, people shouldn't really use this type manually. I haven't verified what happens when you have a Postgres source with a bpchar type before this change.

This was found with Claude Code looking for bugs in each .rs file per session.

> Noticed we don't handle postgres_types::Type::CHAR_ARRAY in from_oid_and_typmod. Wonder if we should add that in too?

Thanks, done!

PR #32555, opened 2025-05-21
`EXPLAIN ANALYZE` for indexes and materialized views

mgree:
Pair programmed with @antiguru.

Adds support for:

```
EXPLAIN ANALYZE 
  (HINTS | 
   ((CPU | MEMORY) [, (CPU | MEMORY)] [WITH SKEW]) 
FOR (INDEX ... | MATERIALIZED VIEW ...) 
[AS SQL]
```

Generates appropriate queries to `mz_introspection` views, effectively blessing certain introspection queries.

### Motivation

  * This PR adds a known-desirable feature.
https://github.com/MaterializeInc/database-issues/issues/6551

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #36040, opened 2026-04-13
sql: escape object names in EXPLAIN ANALYZE generated SQL

antiguru:
Object names containing single quotes (e.g., `"it's"`) cause a parse error when interpolated into the generated SQL query for `EXPLAIN ANALYZE ... FOR MATERIALIZED VIEW`.
Uses `escaped_string_literal` to properly escape the name, preventing a panic in `ShowSelect::new` with "unterminated quoted string".

Reproduces with:
```sql
CREATE MATERIALIZED VIEW "it's" AS SELECT * FROM t;
EXPLAIN ANALYZE MEMORY FOR MATERIALIZED VIEW "it's";
```

Regression from #32555.

Fixes https://github.com/MaterializeInc/database-issues/issues/11293

### DIFF
--- a/src/sql/src/plan/statement/dml.rs
+++ b/src/sql/src/plan/statement/dml.rs
@@ -41,7 +41,7 @@ use mz_storage_types::sinks::{
     MIN_S3_SINK_FILE_SIZE, S3SinkFormat, StorageSinkConnection,
 };
 
-use crate::ast::display::AstDisplay;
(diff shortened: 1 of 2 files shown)


def- (commented):
Would be nice to add the previously crashing query as a regression test

antiguru:
> Would be nice to add the previously crashing query as a regression test

Done!

ggevay (approved):

PR #36041, opened 2026-04-13
mzcompose: Improve SQL Server health check

def-:
Kept annoying me while trying to run SQL Server based tests locally.


martykulma (approved):
🐟

PR #36042, opened 2026-04-13
compute-client: stop being generic over the timestamp type

teskje:
The compute client code is generic over a `T: ComputeControllerTimestamp`, but that trait is so specific that only a single type fulfills it, namely `mz_repr::Timestamp`. There are no plans to support multiple timestamp types, and if we want to make changes to the timestamp type, we can always make them to `mz_repr::Timestamp`. Thus the genericity doesn't give us anything except code noise.

Change the compute client code to not be generic over the timestamp type anymore.


antiguru (approved):
Thanks!

teskje:
TFTR!

PR #36043, opened 2026-04-13
storage-{client,controller}: stop being generic over the timestamp type

teskje:
The storage client code is generic over a `T: Timestamp`, though it's only ever used with `mz_repr::Timestamp`. There are no plans to support multiple timestamp types, and if we want to make changes to the timestamp type, we can always make them to `mz_repr::Timestamp`. Thus the genericity doesn't give us anything except code noise.

Change the storage client and storage controller code to not be generic over the timestamp type anymore.

### Motivation

Part of a push to remove abstraction we don't need, see Slack threads [1](https://materializeinc.slack.com/archives/C08ACQNGSQK/p1775911944776659) and [2](https://materializeinc.slack.com/archives/C08A62E0751/p1775926643153079).

An equivalent change for the compute client landed in https://github.com/MaterializeInc/materialize/pull/36042.

### DIFF
--- a/src/adapter/src/catalog/open.rs
+++ b/src/adapter/src/catalog/open.rs
@@ -638,9 +638,7 @@ impl Catalog {
     /// collections created between versions.
     async fn initialize_storage_state(
         &mut self,
-        storage_collections: &Arc<
-            dyn StorageCollections<Timestamp = mz_repr::Timestamp> + Send + Sync,
-        >,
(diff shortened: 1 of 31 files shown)


antiguru (approved):
🚢

martykulma (approved):
🦈

teskje:
TFTRs!

PR #36044, opened 2026-04-13
Fix fast-path plan creation to be mindful of the finishing's offset

ggevay:
Fixes https://github.com/MaterializeInc/database-issues/issues/11292

### DIFF
--- a/src/adapter/src/coord/peek.rs
+++ b/src/adapter/src/coord/peek.rs
@@ -465,13 +465,16 @@ pub fn create_fast_path_plan<T: Timestamp>(
             {
                 if let Some(finishing) = finishing {
                     if group_key.is_empty() && *order_key == finishing.order_by && *offset == 0 {
-                        // The following is roughly `limit >= finishing.limit`, but with Options.
+                        // The following is roughly `limit >= finishing.limit + finishing.offset`,
+                        // but with Options.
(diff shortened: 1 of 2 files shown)


src/adapter/src/coord/peek.rs
```diff
@@ -465,13 +465,16 @@ pub fn create_fast_path_plan<T: Timestamp>(
...
+                        // but with Options.
                         let finishing_limits_at_least_as_topk = match (limit, finishing.limit) {
                             (None, _) => true,
                             (Some(..), None) => false,
                             (Some(topk_limit), Some(finishing_limit)) => {
                                 if let Some(l) = topk_limit.as_literal_int64() {
-                                    l >= *finishing_limit
+                                    i128::cast_from(l)
+                                        >= i128::cast_from(*finishing_limit)
+                                            + i128::cast_from(finishing.offset)
```
antiguru, line 477: What's the benefit of using u128 over `finishing_limit.saturating_add(finishing.offset)`?

ggevay: If `l` is exactly equal to `i64::max`, then the `saturating_add` capping the right-hand side down to also `i64::max` could result in a false positive detection, no? (I think this bug couldn't actually manifest, because our `Diff` type is also `i64`, but we better not rely on that here.)

frankmcsherry (approved):
This is fine; please merge! :D

PR #36046, opened 2026-04-13
compute/correction_v2: make chunk capacity configurable

teskje:
Previously, `CorrectionV2` used `TimelyStack`'s default capacity (8KiB) for the `Chunk` capacity. We suspect that this size is smaller than ideal and would like a way to experiment with different values in production.

This commit makes the chunk capacity configurable through a dyncfg. Like the chain proportionality, changes to this configuration only take effect when a `CorrectionV2` instance is created.

-------

First commit is pure code movement, to simplify the review of the logic changes in the second commit.

### DIFF
--- a/src/compute/src/sink/correction_v2.rs
+++ b/src/compute/src/sink/correction_v2.rs
@@ -422,6 +422,158 @@ impl<D: Data> CorrectionV2<D> {
         self.prev_size = new_size;
         self.prev_update_count = new_length;
     }
+
+    /// Merge the given chains, advancing times by the given `since` in the process.
+    fn merge_chains<D: Data>(
+        chains: impl IntoIterator<Item = Chain<D>>,
+        since: &Antichain<Timestamp>,
+    ) -> Chain<D> {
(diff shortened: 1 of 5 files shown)


src/compute/src/sink/correction_v2.rs
```diff
@@ -633,7 +639,7 @@ impl<D: Data> Chain<D> {
         match self.chunks.last_mut() {
             Some(c) if !c.at_capacity() => c.push(update),
```
antiguru: We might not rely on `SizableContainer` in this file and rather have `c.len() >= self.chunk_capacity` here.

antiguru (commented):
Looks good, but I'd like to see the use of `SizableContainer` to go away.

src/compute/src/sink/correction_v2.rs
teskje: Done!

antiguru (approved):
Thank you!

teskje:
TFTR!

PR #36047, opened 2026-04-13
storage: introduce StorageScalarExpr for
  postgres source casts

antiguru:
Replace `MirScalarExpr` in `PostgresSourceExportDetails.column_casts` with a new closed-enum `StorageScalarExpr` in `mz-storage-types`.
Storage needs stable cast evaluation that does not change when the expression framework evolves.

* Define `StorageScalarExpr` (4 variants: Column, Literal, CallUnary, ErrorIfNull) and `CastFunc` (24 cast variants) in `mz-storage-types`
* Copy eval implementations from `mz_expr` CastStringTo* functions, delegating to the same `strconv::parse_*` functions
* Rewrite `generate_column_casts` to build `StorageScalarExpr` directly from PG type metadata, bypassing the HIR/MIR planning pipeline
* Update `cast_row` in storage to use `StorageScalarExpr::eval`
* No migration needed — `column_casts` is never durably persisted, always recomputed at plan time
* Parity tests assert structural equality of results and errors against `MirScalarExpr`
* Error snapshot tests hardcode expected `EvalError` variants for each cast function

## Stability contract

Error variants, error messages, and output Datum types produced by `StorageScalarExpr::eval` must not change across releases.
The eval implementations delegate to `mz_repr::strconv::parse_*`, which depend on these external crates whose version bumps may alter parse behavior:

* `chrono` / `chrono-tz` — date, time, timestamp, timestamptz parsing
* `dec` (decnumber-sys) — numeric parsing and rescaling
* `serde_json` — jsonb parsing
* `uuid` — uuid parsing
* `hex` — bytea hex decoding
* `regex` — float special-value detection (inf, NaN)
* `ordered-float` — float Datum representation

Additionally, Materialize-internal crates control behavior for:

* `mz-repr` (strconv, adt types) — core parsing logic
* `mz-pgtz` — timezone resolution
* `mz-ore` — lexing for container types (array, list, map, range)

### DIFF
--- a/src/sql/src/pure/postgres.rs
+++ b/src/sql/src/pure/postgres.rs
@@ -592,12 +592,27 @@ pub(crate) fn generate_column_casts(
             }
         };
 
-        let cast_expr = match pg_type_to_cast_func(scx, &ty)? {
-            // No cast needed (e.g. Text → String identity).
-            None => StorageScalarExpr::Column(i),
-            Some(cast_func) => {
+        let cast_expr = match pg_type_to_cast_func(scx, &ty) {
+            Ok(None) => {
+                // No cast needed (e.g. Text → String identity).
+                StorageScalarExpr::Column(i)
+            }
+            Ok(Some(cast_func)) => {
                 StorageScalarExpr::CallUnary(cast_func, Box::new(StorageScalarExpr::Column(i)))
             }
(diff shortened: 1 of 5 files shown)


martykulma (approved):
lgtm - thanks @antiguru!

PR #36053, opened 2026-04-14
docker: Upgrade Ubuntu images to 26.04 LTS

def-:
Only to be merged after the release (April 23?) and when AWS Inspector supports it: https://docs.aws.amazon.com/inspector/latest/user/supported.html


def-:
Abandoned in favor of https://github.com/MaterializeInc/materialize/pull/37633

PR #36061, opened 2026-04-14
expr: Fix missing aggregate variants in `is_constant` and `propagates_nonnull_constraint`

def-:


ggevay (approved):
Thanks! I also like the elimination of the wildcard!

PR #36063, opened 2026-04-14
Remove PR checklist bot workflow

antiguru:
Nobody reads the automated PR comment — it's noise.
The same guidance (PR title conventions, test coverage, design docs, proto labels, cloud companion PRs) lives in `doc/developer/guide-changes.md` and is enforced through code review.


def-:
I always use the "[trigger-ci for additional test/nightly runs](https://ci.dev.materialize.com/trigger/36063)"

aljoscha:
> I always use the "[trigger-ci for additional test/nightly runs](https://ci.dev.materialize.com/trigger/36063)"

I also use that link a lot, is there maybe a good way we can add a button right in github (I know you can set up a bookmark to do that, but then everyone has to do that themselves...). But other than that I'd be in favor of removing the message!

def- (approved):

def-:
We could use a github status for that, it's slightly more hidden. Added a commit for it. Here's a demo of how it looks for Vercel: 

https://github.com/user-attachments/assets/5f956a7a-2c9b-4347-9ec2-143270776719

antiguru:
> We could use a github status for that

Neat! It seems github sorts the checks alphabetically, so we could pick a name that sorts earlier.

def-:
Yes, I also just realized that, so using "A" now :D

def-:
Screenshot 2026-04-15 at 15 07 32
Works. I didn't realize the "CI Trigger Link" will also show up though, I'll remove that and create the GH status from Buildkite's side instead.

PR #22601, opened 2023-10-23
sql: Add support for `ALTER [CLUSTER | SCHEMA] SWAP ...`

ParkMyCar:
**_work in progress_**

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered.
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] This PR includes the following [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note):
  -

PR #36064, opened 2026-04-14
ddl: Fix temp name for cluster swapping

def-:
Follow-up to #22601


src/sql/src/plan/statement/ddl.rs
```diff
@@ -6953,7 +6953,7 @@ where
     let cluster_b = scx.resolve_cluster(Some(&name_b))?;
 
     let check = |temp_suffix: &str| {
-        let mut temp_name = ident!("mz_schema_swap_");
```
antiguru, line 6956: We should unify the two instances of "mz_lcuster_swap" so this problem won't happen again.

antiguru (approved):
Thank you!

PR #36065, opened 2026-04-14
sql: Simplify quantified comparisons in JOIN ON clauses

def-:
This comes from #9623

### DIFF
--- a/src/sql/src/plan/lowering.rs
+++ b/src/sql/src/plan/lowering.rs
@@ -141,6 +141,7 @@ pub struct Config {
     pub enable_variadic_left_join_lowering: bool,
     pub enable_guard_subquery_tablefunc: bool,
     pub enable_cast_elimination: bool,
+    pub enable_simplify_quantified_comparisons: bool,
 }
 
 impl Default for Config {
@@ -150,6 +151,7 @@ impl Default for Config {
             enable_variadic_left_join_lowering: false,
             enable_guard_subquery_tablefunc: false,
             enable_cast_elimination: false,
+            enable_simplify_quantified_comparisons: false,
         }
     }
 }
@@ -161,6 +163,7 @@ impl From<&SystemVars> for Config {
             enable_variadic_left_join_lowering: vars.enable_variadic_left_join_lowering(),
             enable_guard_subquery_tablefunc: vars.enable_guard_subquery_tablefunc(),
             enable_cast_elimination: vars.enable_cast_elimination(),
+            enable_simplify_quantified_comparisons: vars.enable_simplify_quantified_comparisons(),
         }
     }
 }
@@ -203,7 +206,10 @@ impl HirRelationExpr {
             mut other => {
                 let mut id_gen = mz_ore::id_gen::IdGen::default();
                 transform_hir::split_subquery_predicates(&mut other)?;
-                transform_hir::try_simplify_quantified_comparisons(&mut other)?;
+                transform_hir::try_simplify_quantified_comparisons(
+                    &mut other,
+                    context.config.enable_simplify_quantified_comparisons,
+                )?;
                 transform_hir::fuse_window_functions(&mut other, &context)?;
                 MirRelationExpr::constant(vec![vec![]], ReprRelationType::new(vec![])).let_in(
                     &mut id_gen,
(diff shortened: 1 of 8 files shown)


src/sql/src/plan/transform_hir.rs
```diff
@@ -211,11 +211,23 @@ pub fn try_simplify_quantified_comparisons(
...
+                left, right, on, ..
+            } => {
                 walk_relation(left, outers)?;
+                let left_type = left.typ(outers, &NO_PARAMS);
                 let mut outers = outers.to_vec();
-                outers.insert(0, left.typ(&outers, &NO_PARAMS));
+                outers.insert(0, left_type);
                 walk_relation(right, &outers)?;
+                // Walk the ON clause so that quantified comparisons
+                // (= ANY/= ALL) within it are simplified to EXISTS form.
```
ggevay: Nit: I'd say the first two lines of this comment are not really needed.

> Walk the ON clause

is obvious from the `walk_scalar(on, &outers, true)?;`, and

> quantified comparisons (= ANY/= ALL) within it are simplified to EXISTS form.

is what the whole function is doing, and is thus already explained in the function-level doc comment.

test/sqllogictest/subquery.slt
```diff
@@ -708,3 +708,58 @@ Source materialize.public.y
 Target cluster: quickstart
 
 EOF
+
+# Regression test: quantified comparison simplification (= ANY -> EXISTS) must
+# also apply inside JOIN ON clauses, not just WHERE. The decorrelated plan
```
ggevay, line 713: Could you please explicitly mention the function name `try_simplify_quantified_comparisons`?

ggevay (commented):
Hmm, thanks, this looks good to me!

But I'd say we should create a feature flag for it (on by default), because we'd like an escape hatch in case it causes a catastrophic performance regression at some user. It's extremely hard to fully anticipate the consequences of optimizer changes. Even though the HIR plan change of this PR is generally considered a positive change, there are tons of surprising non-monotonicities in the optimizer, in the sense that when the plan looks better at a certain stage of the pipeline, it often happens that the better-looking plan is handled worse by a later optimizer pipeline stage, and we end up with a regression overall. (See https://github.com/MaterializeInc/materialize/pull/35441 for more discussion on the evolvability of the optimizer.)

I've kicked off the random query parts of Nightly as a further precaution: https://buildkite.com/materialize/nightly/builds/15999

ggevay (approved):
Thank you!

ggevay (changes requested):
Wait, sorry, moment

src/sql/src/plan/lowering.rs
```diff
@@ -203,7 +206,9 @@ impl HirRelationExpr {
             mut other => {
                 let mut id_gen = mz_ore::id_gen::IdGen::default();
                 transform_hir::split_subquery_predicates(&mut other)?;
-                transform_hir::try_simplify_quantified_comparisons(&mut other)?;
+                if context.config.enable_simplify_quantified_comparisons {
+                    transform_hir::try_simplify_quantified_comparisons(&mut other)?;
+                }
```
ggevay: This would completely disable `try_simplify_quantified_comparisons` when the flag is off, so we'd have changed plans from before the PR with the flag both on and off. What I meant is to guard only the new part of `try_simplify_quantified_comparisons` with the feature flag, because the important thing is to be able to get back to the old plans by turning off the flag (if needed).

def-: Oops, thanks a lot for catching!

src/sql/src/session/vars/definitions.rs
```diff
@@ -2262,6 +2262,12 @@ feature_flags!(
         default: false,
         enable_for_item_parsing: false,
     },
+    {
+        name: enable_simplify_quantified_comparisons,
+        desc: "Allow the optimizer to simplify quantified comparisons (= ANY/= ALL subqueries) into semi/anti-join EXISTS form during HIR-to-MIR lowering.",
```
ggevay: `desc` is stale from changing the flag's scope

ggevay (approved):
LGTM, just a minor thing left

def-:
Thanks for the great reviews Gabor!

PR #36066, opened 2026-04-14
lint: Minor fixes with Python 3.10 and on macOS

def-:
I had a weird setup defaulting to Python 3.10 on macOS, which we claim to support. So fixed it up before upgrading to Python 3.14.


antiguru (approved):

bosconi (approved):
Looks good.

PR #36067, opened 2026-04-14
Fix range_of returning half the actual bucket range in temporal bucketing

def-:
range_of used bits.saturating_sub(1) to avoid overflow, but this returned half the bucket range, causing up to 2x more buffer flushes. Fix by using bits directly and introducing BucketRange to handle the overflow case (None = unbounded).


antiguru (approved):
Thank you!

def-:
Will merge once Nightly finishes: https://buildkite.com/materialize/nightly/builds/15997

PR #32220, opened 2025-04-15
[sql_server] support `ALTER SOURCE ... ADD SUBSOURCE`

ParkMyCar:
This PR adds support for adding subsources to a `SQL SERVER` source. All of the changes are in Adapter during purification. We also add a new testcase to the `sql-server-cdc` mzcompose flow to exercise adding and dropping subsources.

### Motivation

Progress towards https://github.com/MaterializeInc/database-issues/issues/8762

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #36070, opened 2026-04-14
sql server: Fix `ALTER SOURCE` silently dropping `TEXT`/`EXCLUDE COLUMNS`

def-:
SQL Server text/exclude column refs truncate to 2-part (schema.table) but external references are 3-part (database.schema.table). The exact equality check always failed, silently dropping all column options on `ALTER SOURCE ADD SUBSOURCE`. Use suffix matching instead.

Introduced by #32220

Test run: https://buildkite.com/materialize/nightly/builds/15998


patrickwwbutler (approved):
good catch!

PR #36071, opened 2026-04-14
`ALTER ROLE`: Fix `CREATEROLE` check

def-:
Fixes: https://github.com/MaterializeInc/database-issues/issues/11277

### DIFF
--- a/src/sql/src/rbac.rs
+++ b/src/sql/src/rbac.rs
@@ -1205,14 +1205,18 @@ fn generate_rbac_requirements(
             }
             // Roles are allowed to change their own password, but only if
             // password is the sole attribute being changed.
-            plan::PlannedAlterRoleOption::Attributes(attributes)
-                if attributes.password.is_some()
-                    && role_id == *id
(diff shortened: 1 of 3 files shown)


src/sql/src/rbac.rs
```diff
@@ -1203,15 +1203,19 @@ fn generate_rbac_requirements(
...
             }
-            // Roles are allowed to change their own password.
+            // Roles are allowed to change their own password, but only if
+            // password is the sole attribute being changed.
             plan::PlannedAlterRoleOption::Attributes(attributes)
-                if attributes.password.is_some() && role_id == *id =>
+                if attributes.password.is_some()
+                    && role_id == *id
+                    && attributes.login.is_none()
+                    && attributes.inherit.is_none() =>
```
ggevay: Doing a (non-wildcard) pattern matching mentioning all fields of `attributes` might be more future-proof.

And maybe it would be worth a code comment to explain why `nopassword` and `superuser` are not being checked in the condition.

ggevay (approved):

PR #36073, opened 2026-04-14
Tighten `Context`

frankmcsherry:
This is a mostly cosmetic PR that tightens up text around `Context`, removing a bunch of blank space. It also pushes `RenderTimestamp` down into bounds, but it is required by `render_plan` so it's not actually optional.


antiguru (approved):
Thank you!

PR #36075, opened 2026-04-14
De-timestamp more parts of the adapter layer

frankmcsherry:
This is a sequence of PRs that remove `T: Timestamp` generality from components that Claude was able to identify as not needing it. They are spelled out in the names of the commits.


ggevay (approved):
Great, thank you! :happypanda:

Issue #28983, opened 2024-08-13
parallel-workload: 0dt: scenario has stuck queries


def-:
### What version of Materialize are you using?

52963756e8

### What is the issue?

Seen in https://buildkite.com/materialize/nightly/builds/9075#01914ccb-9de0-41fa-83c8-4ef2f92ca787:
```
worker_0 still running: FETCH 943 c WITH (timeout='7s');
worker_1 still running: DROP TABLE "db-pw-1723571225-1"."s-3"."t-48";
worker_2 still running: CREATE ROLE role166;
worker_3 still running: DELETE FROM "db-pw-1723571225-0"."s-0"."t-49";
worker_4 still running: rollback;
worker_5 still running: CREATE SINK "db-pw-1723571225-1"."s-3"."sink-264" IN CLUSTER "cluster-16" FROM "db-pw-1723571225-0"."s-0".kafka_table64 INTO KAFKA CONNECTION kafka_conn (TOPIC sink_topic264)  FORMAT AVRO USING CONFLUENT SCHEMA REGISTRY CONNECTION csr_conn ENVELOPE DEBEZIUM;
worker_6 still running: DROP MATERIALIZED VIEW "db-pw-1723571225-2"."s-18"."v-355";
worker_7 still running: COPY (SELECT * FROM "db-pw-1723571225-0"."s-0"."v-342") TO 's3://copytos3/586' WITH (AWS CONNECTION = aws_conn, FORMAT = 'csv');
worker_8 still running: rollback;
worker_9 still running: FETCH 433 c WITH (timeout='9s');
worker_10 still running: FETCH ALL c WITH (timeout='6s');
worker_11 still running: SET TRANSACTION_ISOLATION TO 'SERIALIZABLE';
worker_12 still running: SET REAL_TIME_RECENCY TO FALSE;
worker_13 still running: DECLARE c CURSOR FOR SUBSCRIBE "db-pw-1723571225-0"."s-2"."v-362" ENVELOPE UPSERT (KEY ("key7-table10", "value0-table10", "value2-table10", "value14-table10", "key6-table10", "value3-table10", "value4-table10", "value12-table10", "value10-table10", "key5-table10", "value1-table10", "value9-table10", "value8-table10", "key0-table10", "key2-table10", "key1-table10", "value13-table10", "value7-table10", "value17-table10", "value15-table10", "value16-table10", "value6-table10", "key4-table10"));
worker_14 still running: rollback
[...]

PR #36099, opened 2026-04-15
containers: switch environmentd and clusterd to distroless base image

jasonhernandez:
## Summary

Switch environmentd and clusterd from Debian-based prod images to distroless. Follow-on to #37227: the statically-linked `ssh` from `openssh-static` replaces `/usr/bin/ssh`, so the images no longer need apt, a shell, or OpenSSL.

## Changes

- **`distroless-prod-base`**: new base image off `gcr.io/distroless/cc-debian13:nonroot`. Carries `libeatmydata` (parity with `ubuntu-base`) and a static `tini`.
- **`environmentd` / `clusterd`**: switch to the new base, copy in the static `ssh`, and drop the shell `entrypoint.sh` scripts. Env-var defaults the entrypoints used to set are passed explicitly by the orchestrator.
- **`tini-static`** (new mzimage): static `tini` v0.19.0 as PID 1 in every distroless image, to forward signals and reap the `ssh` tunnel subprocesses. Same init prod images run today, just built static. Kept separate from the expensive `openssh-static` build so their caches stay decoupled.
- **Fenced-out behavior**: the deleted environmentd entrypoint slept forever after a fenced-out `exit!(0)` so Kubernetes would not crash-loop the old generation (#28983). Replaced by `Config::idle_when_fenced_out`: under the Kubernetes orchestrator, a fenced-out environmentd idles in-process until pod deletion terminates it. The all-in-one `materialized` image keeps its shell entrypoint and is unchanged.

## Coordination

Lands **after or with** #36101 (orchestratord UID/GID gate). Its `V26_33_0` gate must equal the release this PR ships in. Details in the coordination comment below.

## Follow-up

Remove the static `ssh` from `prod-base`: its remaining consumers (`mz`, `jobs`, test images) open no SSH tunnels, so OpenSSH drops out of those images and their scanner noise with it.

## Test plan

PR CI already exercises the rebuilt images end to end: container startup and traffic (all mzcompose jobs), SSH tunnels (`ssh-connection-*`), and 0dt upgrades (`0dt-upgrade-smoke-test`).

- [ ] Nightly run on the final revision
- [ ] Kubernetes: fenced-out old generation idles instead of crash-looping and is torn down cleanly (`idle_when_fenced_out` is Kubernetes-only, so mzcompose CI cannot cover it)

## ⚠️ Not independently mergeable — validate via the stacked branch

This PR's own Nightly cannot go fully green in isolation: the `Orchestratord` kind jobs (e.g. `documentation-defaults`) fail with "Never completed" because the nonroot UID/securityContext gate they need lives in **#36101**, not here. Orchestratord built from this branch applies the old root securityContext to the nonroot distroless image, so the pod can't start. This is expected — #36101 must land **before or with** this PR.

The validation of record is the stacked proof branch (#37466 = this + #36101), where the full Orchestratord suite, Cloudtest, and K8s-recovery pass green. Don't chase the lone `orchestratord-plus-documentation-defaults` red here.


def-:
I have tried that too, it's going to be difficult: https://github.com/MaterializeInc/materialize/pull/35631

jasonhernandez:
> [!WARNING]
> **OUTDATED — superseded by the coordination note below.** This PR now ships a static `tini` as PID 1, so the #36100 dependency described here no longer applies. Kept for history.

<del>

**🔗 Distroless migration — coordination note**

This PR is one of three that must ship **together in the same release**:

- **#36099** (this) — switch environmentd/clusterd to distroless + copy in the static `ssh` from `openssh-static`.
- **#36100** — SIGTERM handler for clusterd (distroless runs the binary as PID 1, no tini to forward signals).
- **#36101** — orchestratord pod securityContext UID/GID gating (distroless runs as nonroot 65534, not 999).

**Merge order:** #36100 must merge **before or with** this PR — otherwise distroless clusterd runs as PID 1 with no SIGTERM handler and won't shut down gracefully (k8s waits, then SIGKILLs).

**Release coupling:** #36101 hard-codes `V26_28_0` as the version at which env/clusterd images become distroless. That must equal the release this PR actually lands in. If this slips past the 26.28 cycle, bump that constant in #36101.

Rebased onto current `main`; the `MZFROM openssh-static` / `COPY --from=openssh /output/ssh` wiring was verified by a local `docker build` (statically-linked `OpenSSH_10.3p1`).

</del>

jasonhernandez:

The image change in a set that ships together:

- **#36099** (this) — distroless environmentd/clusterd base, static `tini` as PID 1, static `ssh` copied in.
- **#36101** — orchestratord pod securityContext UID/GID gating (nonroot `65534` vs `999`).
- **#36872** — clusterd derives `CLUSTERD_PROCESS` from the pod hostname (distroless has no shell entrypoint to set it).
- **#36876** — removes the unused CTP server-FQDN check (stacks on #36872).

**Merge order:** #36101 and #36872 land **before or with** this PR.
- #36872: without it, distroless clusterd has no process ordinal and won't start.
- #36101: distroless images run nonroot `65534`; the gate sets the matching securityContext.

Both are version-conditional / unset-guarded, so they are no-ops on current images and safe to land ahead. The reverse is not: shipping distroless images before either lands means clusterd can't start (missing ordinal) or runs with the wrong UID/GID.

#36876 is dead-code cleanup, off the critical path (a missing `CLUSTERD_GRPC_HOST` already makes the check a no-op), and can land any time after #36872.

Signals: `tini` (PID 1) forwards signals and reaps the `ssh` tunnel subprocesses, so no per-binary signal handling is required.

.github/dependabot.yml
```diff
@@ -109,6 +109,36 @@ updates:
...
+      update-types:
+        - minor
+        - patch
+- package-ecosystem: docker
+  directory: /misc/images/distroless-prod-base
+  schedule:
+    interval: weekly
+    day: sunday
+    time: "22:00"
+  # TODO: Add cooldown when https://github.com/dependabot/dependabot-core/issues/14044 is fixed
```
Alphadelta14, line 133: jsyk, this was merged ~month and a half ago in https://github.com/dependabot/dependabot-core/pull/14149

def- (commented):
One thing the QA LLM review found:
## Finding 1 - MEDIUM: distroless clusterd disables CTP FQDN validation in Kubernetes

The PR deletes `src/clusterd/ci/entrypoint.sh`, including the Kubernetes-only default:

```sh
```

The new distroless image runs `clusterd` directly through tini (`src/clusterd/ci/Dockerfile`), but there is no replacement for this value in the generated Kubernetes service args. The controller-generated clusterd args in `src/controller/src/clusters.rs` still pass the listen addresses, pubsub URL, environment id, timely configs, memory options, and secrets flags, but not `--grpc-host`. The Kubernetes orchestrator also only appends `--secrets-reader-name-prefix`, and has no per-pod `CLUSTERD_GRPC_HOST` env.

This makes `args.grpc_host` remain `None` in `src/clusterd/src/lib.rs`. `clusterd::main` then passes that `None` to both storage and compute `transport::serve` calls. In `src/service/src/transport.rs`, the CTP handshake only checks the peer/server FQDN when both sides provide `Some`, so `None` disables the validation entirely. Before this PR, Kubernetes clusterd pods set the value to their pod FQDN, which matched the controller address generated by `orchestrator-kubernetes` (`{statefulset}-{i}.{service}.{namespace}.svc.cluster.local`).

Impact: managed Kubernetes clusterd pods no longer reject CTP connections whose intended server FQDN does not match the pod they reached. This removes the Kubernetes-only guard against misrouted controller connections, stale service/DNS routing, or any internal caller that accidentally connects to the wrong replica process. Since a CTP server accepts one client at a time and cancels the previous connection on a new accepted connection, a wrong connection can disrupt the legitimate controller and can route storage/compute commands to the wrong process instead of failing the handshake with `server FQDN mismatch`.

misc/images/distroless-prod-base/Dockerfile
```diff
@@ -10,7 +10,47 @@
...
+# set as a container environment variable (must be set before process start).
+FROM debian:13-slim AS eatmydata
+RUN apt-get update && apt-get install -y --no-install-recommends eatmydata \
+    && rm -rf /var/lib/apt/lists/*
+
+# Statically-linked `tini`, the PID 1 init for all distroless prod images.
+# Distroless ships no init, so without this each leaf binary would run as PID 1
+# and have to forward signals and reap zombies (e.g. ssh tunnel subprocesses)
+# itself.
+MZFROM tini-static AS tini
```
alex-hunt-materialize, line 24: Any reason to ship tini, rather than just setting `share_process_namespace=True` in our pod template specs? If we set share process namespace to true, kubernetes will set pause as PID 1 to handle signal forwarding and zombie reaping.

In places where we run with docker, we can add the `--init` arg to have it add PID 1.

misc/python/materialize/mzcompose/services/clusterd.py
```diff
@@ -90,6 +90,9 @@ def __init__(
         # Override the materialized entrypoint so that `clusterd` is invoked
         # via the command rather than via the entrypoint. This keeps
         # `c.exec()` working (it prepends the entrypoint to exec commands).
+        # Note: mzcompose uses the Ubuntu-based `materialized` image (with
+        # tini/bash), while production uses the distroless `clusterd` image.
+        # Keep this in mind when debugging CI-vs-prod discrepancies.
```
alex-hunt-materialize, line 95: Do we plan to follow up with moving materialized also? I understand migrating all of CI to avoid running shell stuff might be a lot, but having these be such drastically different images makes me worry that we'll miss differences in behavior.

src/clusterd/ci/entrypoint.sh
```diff
@@ -1,41 +0,0 @@
...
-
-if [[ "${KUBERNETES_SERVICE_HOST:-}" ]]; then
-    # Pass the host's FQDN as the host to be used for GRPC request validation
-    # only when running in Kubernetes. In other contexts (like when running
-    # locally, or in Docker), this is likely not desirable.
-    export CLUSTERD_GRPC_HOST=${CLUSTERD_GRPC_HOST:-$(hostname --fqdn)}
-
-    # When running in Kubernetes, pass the StatefulSet replica's ordinal index
-    # as the process index.
-    export CLUSTERD_PROCESS=${CLUSTERD_PROCESS:-${HOSTNAME##*-}}
```
alex-hunt-materialize, line 32: The LLM review pointed out the CLUSTERD_GRPC_HOST var no longer being set, but I think these others are also missing.

src/environmentd/src/deployment/preflight.rs
```diff
@@ -221,6 +225,14 @@ pub async fn preflight_0dt(
...
             caught_up_trigger: None,
         })
+    } else if idle_when_fenced_out {
+        // Exiting under a Kubernetes StatefulSet (`restartPolicy: Always`) would
+        // restart the container straight back into this fenced-out state, a
+        // crash loop until the orchestrator deletes the StatefulSet. Idle until
+        // pod deletion's SIGTERM terminates us instead.
+        info!("this deployment has been fenced out; idling until terminated");
+        std::future::pending::<()>().await;
+        unreachable!("pending() never resolves");
```
alex-hunt-materialize, line 235: This seems like an unrelated change. It's probably a good idea, but is there some behavior change about fencing that I'm missing?

misc/python/materialize/mzcompose/services/clusterd.py
jasonhernandez: Moving from Ubuntu to a Debian base that matches the distroless root would make sense as a follow on and be a manageable effort. I think we'll still want a shell for CI / dev, and that's not so risky, but the Debian / Ubuntu release difference could bite us.
I'm also working on migrating everything to rustls + aws-lc which will eliminate one of the key libraries that Materialize interacts with on its host containers. That's moving somewhat in parallel because I'm trying to eliminate our exposure to OpenSSL.

misc/images/distroless-prod-base/Dockerfile
jasonhernandez: I was trying to minimize the differences from the existing image where we installed tini via apt-get. I'd support a follow-on PR or look at one pointed at this branch, especially if it makes this set of PRs smaller.

misc/python/materialize/mzcompose/services/clusterd.py
jasonhernandez: Also, half of our containers are already running on distroless... and I don't think we had great consistency in the past. I think this moves us in the right direction, especially if we move materialized to a related Debian base..

Alphadelta14: the recommendation if you wanted a parallel with shell access is to use the :debug tag which does have shells.

misc/images/tini-static/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+    build-essential \
+    ca-certificates \
+    cmake \
+    git \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /build/tini
+RUN git clone --depth 1 --branch ${TINI_VERSION} https://github.com/krallin/tini.git . \
```
Alphadelta14, line 38: fwiw, you don't even need git. you can literally just ADD a git repo. https://docs.docker.com/reference/dockerfile/#source

src/environmentd/src/deployment/preflight.rs
Alphadelta14: There's a ci entrypoint that does a sleep infinity after environmentd exits.
https://github.com/MaterializeInc/materialize/pull/36099/changes#diff-29d710f81d26c40ede0e6ca3409e59656caa397723c5fa430bf15481129579b5

.github/dependabot.yml
jasonhernandez: yeah, I don't trust it yet, but might look at it later

jasonhernandez:
I'm shelving this approach for now in favor of #37526

src/environmentd/src/deployment/preflight.rs
jasonhernandez: I'm focusing in more specifically on slimming down the containers to remove as much Ubuntu userspace as possible. Feel free to pick up any threads / ideas from this closed PR that go beyond the immediate security goal I'm working on (reduce & better identify vulnerability / attack surface).

PR #36100, opened 2026-04-15
clusterd: SIGTERM handler + in-process gRPC FQDN (Variant A)

jasonhernandez:
## Design question: where should clusterd's gRPC FQDN come from?

This is **Variant A** of a two-version pair (see #36872 for Variant B). Both add the SIGTERM handler that distroless clusterd needs as PID 1. They differ **only** in how `CLUSTERD_GRPC_HOST` is set once `entrypoint.sh` is removed by the distroless migration (#36099).

`entrypoint.sh` previously did, in Kubernetes only:
```bash
```

| | **Variant A** (this PR) | **Variant B** (#36872) |
|---|---|---|
| Who resolves the FQDN | clusterd, at startup | orchestrator-kubernetes, at pod-spec build |
| Mechanism | in-process `getaddrinfo(AI_CANONNAME)` (`resolve_fqdn`) | downward API `MZ_POD_NAME` + env interpolation |
| Blocking DNS at startup | **yes** — `getaddrinfo` has no timeout; if CoreDNS is briefly down at pod start, clusterd's main thread blocks | **no** — kubelet substitutes the value |
| clusterd `main()` | SIGTERM handler + `resolve_fqdn` + `CLUSTERD_PROCESS` | SIGTERM handler + `CLUSTERD_PROCESS` only |
| Where pod identity lives | split between binary and orchestrator | with the orchestrator |
| Footprint | self-contained; works under any orchestrator | k8s-specific; +`EnvVar`s in the pod spec |

### Why Variant A

Keeps the change confined to the `clusterd` binary — no orchestrator changes, and it works under any orchestrator (not just k8s) that gives the pod a resolvable hostname. It directly mirrors what `entrypoint.sh` did (`hostname --fqdn`).

### Known downside

`resolve_fqdn` calls `getaddrinfo` with no timeout on the main thread at startup (flagged in the code comment). If DNS is unavailable at pod start, clusterd blocks. This is the same behavior the old shell had, so it's not a regression — but Variant B avoids it entirely by injecting the value from the orchestrator, which already computes the same FQDN.

Part of the distroless migration (#36099 image, this/#36872 lifecycle, #36101 UID/GID).

## Test plan
- [x] `cargo check -p mz-clusterd` (rustc 1.96.0)
- [ ] Verify in k8s that `resolve_fqdn` yields the pod FQDN and gRPC host validation passes
- [ ] Decide A vs B before merge — close the other

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6050,6 +6050,7 @@ dependencies = [
  "futures",
  "hyper 1.9.0",
  "hyper-util",
+ "libc",
  "mz-alloc",
  "mz-alloc-default",
  "mz-build-info",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6022,6 +6022,7 @@ dependencies = [
  "futures",
  "hyper 1.9.0",
  "hyper-util",
+ "libc",
  "mz-alloc",
  "mz-alloc-default",
  "mz-build-info",
--- a/src/clusterd/Cargo.toml
+++ b/src/clusterd/Cargo.toml
@@ -17,6 +17,7 @@ fail.workspace = true
 futures.workspace = true
 hyper.workspace = true
 hyper-util.workspace = true
+libc.workspace = true
 mz-alloc = { path = "../alloc" }
 mz-alloc-default = { path = "../alloc-default", optional = true }
 mz-build-info = { path = "../build-info" }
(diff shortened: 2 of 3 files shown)


jasonhernandez:
### 🔗 Distroless migration — coordination note

One of three PRs that ship together: #36099 (distroless image), #36100 (this — SIGTERM handler), #36101 (UID/GID gating).

Merge this **before or with #36099**: distroless runs the binary as PID 1, which silently ignores SIGTERM without an explicit handler. Safe to merge first (harmless on current images). environmentd already handles termination signals, so this change is correctly clusterd-only despite the commit subject.

jasonhernandez:
Closing in favor of the #36872 → #36876 stack.

The SIGTERM handler from this PR now lives in #36872 (minimal). The FQDN question this PR solved with an in-process `resolve_fqdn` is instead handled by #36876, which removes the CTP server-FQDN validation entirely — it turned out to be an optional, misnamed (it's CTP, not gRPC) check that isn't worth re-plumbing for distroless.

PR #36101, opened 2026-04-15
orchestratord: version-gate pod security context UID/GID for distroless

jasonhernandez:
## Summary

Distroless images run as nonroot (uid/gid 65534) instead of root. This version-gates the pod security context in orchestratord so each Materialize version gets the matching `runAsUser`/`runAsGroup`, avoiding UID mismatches during rolling upgrades.

Gates: balancerd `V26_18_0` (already distroless), environmentd/clusterd `V26_33_0`. The env/clusterd gate must equal the release #36099 ships in; see the coordination comment below.

## Test plan

- [x] `cargo test -p mz-orchestratord` passes
- [ ] Pods for new versions get nonroot uid/gid, old versions keep root

### DIFF
--- a/src/cloud-resources/src/crd/materialize.rs
+++ b/src/cloud-resources/src/crd/materialize.rs
@@ -655,7 +655,7 @@ pub mod v1alpha1 {
     }
 }
 
-fn parse_image_ref(image_ref: &str) -> Option<Version> {
+pub fn parse_image_ref(image_ref: &str) -> Option<Version> {
     image_ref
         .rsplit_once(':')
         .and_then(|(_repo, tag)| tag.strip_prefix('v'))
--- a/src/cloud-resources/src/crd/materialize.rs
+++ b/src/cloud-resources/src/crd/materialize.rs
@@ -1619,7 +1619,14 @@ pub mod v1 {
 }
 
 pub fn parse_image_ref(image_ref: &str) -> Option<Version> {
+    // Strip any `@sha256:...` digest first. Otherwise the digest's `:`
+    // separator is mistaken for the tag separator, so a digest-pinned ref like
+    // `env:v26.32.0@sha256:...` parses the digest hex as the tag and fails,
+    // silently falling into the parse-failure path of every version gate.
     image_ref
+        .split('@')
+        .next()
+        .unwrap_or(image_ref)
         .rsplit_once(':')
         .and_then(|(_repo, tag)| tag.strip_prefix('v'))
         .and_then(|tag| {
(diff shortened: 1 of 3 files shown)


jasonhernandez:

Part of a set that ships together: **#36099** (distroless image), **#36872** (clusterd process ordinal, **merged**), **#36876** (removes the unused CTP FQDN check), and this PR (orchestratord UID/GID gating).

This PR version-gates the pod securityContext so distroless images (uid/gid `65534`) and older Ubuntu-based images (`999`) each get the correct values.

**Merge order:** this PR lands **before or with** #36099. The gate is version-conditional, so it is a no-op until distroless images of the gated version exist and is safe to land ahead. If #36099 ships nonroot images while this gate is absent, orchestratord applies the wrong UID/GID.

**Gates:**
- balancerd: `V26_18_0` (distroless since v26.18.0).
- environmentd/clusterd: `V26_33_0`. Must equal the release #36099 ships in. The 26.32 cycle passed without #36099 landing, so the gate was bumped from `V26_32_0` (2026-07-06). If #36099 lands outside the 26.33 cycle, bump this constant to match.

**Separate, pre-existing:** balancerd runs distroless (uid 65534) as of v26.18.0 but this gating is not yet merged, so balancerd pods can carry a UID mismatch today unless `enable_security_context` is off in practice. Tracked independently.

Alphadelta14:
i was very confused walking into the review when i was like "omg how did i miss distroless being merged" but now I have the context of the comments...

Alphadelta14 (approved):
good luck 🤞🏻 
general approach seems fine (though I am not a huge fan of the version conditionals and how those will need to be updated-- but that's just how this works).

alex-hunt-materialize (approved):

def- (commented):
Two concerns from the QA LLM review:
## Finding 1 - MEDIUM - Distroless pods are forced to run as `nobody`, not the image's `nonroot` user

The PR hard-codes the distroless UID/GID as `65534` in both the balancerd gate and the environmentd/clusterd gate:

- `src/orchestratord/src/controller/balancer.rs:164-185`
- `src/orchestratord/src/controller/materialize/generation.rs:901-912`

That is not the UID/GID of the distroless `nonroot` user used by Materialize's distroless base image. The checked-in base image is `gcr.io/distroless/cc-debian13:nonroot-28078d2e5e77671d2046dcc9e2c75334e31efa4d` with `USER nonroot`, and inspecting that exact image config reports `User = "65532"`. Its `/etc/passwd` maps `nonroot:x:65532:65532` and `nobody:x:65534:65534`. The operator Helm chart also already runs the distroless orchestratord pod as `65532` for `fsGroup`, `runAsGroup`, and `runAsUser`.

Because Kubernetes `runAsUser` overrides the image user, this PR makes distroless balancerd, environmentd, and clusterd run as `nobody` instead of the image's configured `nonroot` user. That defeats the stated goal of matching the pod security context to the image UID/GID. It can also break any file or directory in the image that is readable or writable only by `nonroot`, and it leaves `$HOME=/home/nonroot` paired with the `nobody` identity. For clusterd this value is passed through `--orchestrator-kubernetes-service-fs-group` and consumed as `run_as_user`, `run_as_group`, and `fs_group` in `src/orchestrator-kubernetes/src/lib.rs:1121-1126`, so the blast radius includes replicas launched by the new environmentd.

Use `65532` for the distroless case, or make the images deliberately run as `nobody` and update the base image, chart, comments, and tests consistently.

Tracker check: GitHub searches for `65534 nonroot distroless`, `runAsUser 65534`, and `distroless uid 65532` did not find a duplicate in `MaterializeInc/database-issues` or `MaterializeInc/materialize`. Linear searches found only the broader distroless migration tickets `DB-44` and `CLO-26`, not this UID mismatch.

## Finding 2 - MEDIUM - Digest-pinned old image tags are misclassified as distroless

The newly public `parse_image_ref` still splits the image ref with `rsplit_once(':')` and immediately treats the right-hand side as the tag:

`src/cloud-resources/src/crd/materialize.rs:1621-1632`

That does not handle valid OCI references of the form `image:v26.17.0@sha256:...`. For such refs, the final colon is the digest algorithm separator, so `parse_image_ref` sees the digest hex as the tag and returns `None`. This PR's new gates then treat `None` as distroless:

- balancerd returns `true` from the parse-failure arm and uses the distroless UID/GID at `src/orchestratord/src/controller/balancer.rs:175-185`.
- environmentd/clusterd use `Materialize::meets_minimum_version`, whose parse-failure behavior is also to return `true`, before setting `service_fs_group` at `src/orchestratord/src/controller/materialize/generation.rs:909-917`.

The result is that a valid digest-pinned old image, for example `materialize/balancerd:v26.17.0@sha256:...` or `materialize/environmentd:v26.32.0@sha256:...`, is no longer kept on the old `999` UID/GID path. It gets the distroless path solely because a digest was appended. Digest pinning is a realistic deployment configuration for self-managed operators, and this breaks the PR's rolling-upgrade safety property for those users.

The existing `parse_image_tag` helper in `src/orchestratord/src/lib.rs` already handles `:tag@sha256:...` by stripping the digest before splitting on the tag. `parse_image_ref` should do the same, then parse the stripped tag as semver. Add tests for `v26.17.0@sha256:...` staying below the balancerd gate and `v26.32.0@sha256:...` staying below the environmentd/clusterd gate.

alex-hunt-materialize (approved):
LGTM, assuming there is another PR that changes the actual containers that will be merged in the same release.

jasonhernandez:
closing this in favor of the simpler #37526 approach

PR #36104, opened 2026-04-15
ci: Post trigger-ci link from Buildkite instead

def-:
To reduce noise. Follow-up to https://github.com/MaterializeInc/materialize/pull/36063


antiguru (approved):
Thank you!

def-:
Update: depends on https://linear.app/materializeinc/issue/SEC-308/create-new-github-ci-issue-reference-checker-token because I realized the current token is on my personal account, and I don't want my face to tell people to run Nightly, seems too aggressive!
Screenshot 2026-04-16 at 10 32 48

PR #36105, opened 2026-04-15
Track stashed result size in partitioned compute state

antiguru:
### Motivation

The `tracked.stashed_result_size` field was not being populated when stashed updates were stored, leaving it uninitialized or stale. This metric is important for monitoring and understanding the memory footprint of stashed results in the compute service.

### Description

This change adds a single line to compute the total byte size of stashed results immediately after filtering and keeping collections. The size is calculated by summing the `byte_len()` of each kept collection, providing an accurate snapshot of the stashed result size at the point where stashed updates are finalized.

### Verification

The change is a straightforward metric calculation that mirrors the pattern used elsewhere in the codebase. Existing tests should continue to pass, and the metric will now be properly tracked for observability purposes.

Fixes https://github.com/MaterializeInc/database-issues/issues/11302

https://claude.ai/code/session_01SniGK5xKpYxLcT7nDevt3g

### DIFF
--- a/src/compute-client/src/service.rs
+++ b/src/compute-client/src/service.rs
@@ -316,8 +316,7 @@ impl PartitionedComputeState {
                                     keep.push(keep_coll);
                                 }
                             }
-                            tracked.stashed_result_size =
-                                keep.iter().map(|c| c.byte_len()).sum();
+                            tracked.stashed_result_size = keep.iter().map(|c| c.byte_len()).sum();
                             tracked.stashed_updates = Ok(keep);
                             Ok(ship)
                         }
--- a/src/compute-client/src/service.rs
+++ b/src/compute-client/src/service.rs
@@ -316,6 +316,8 @@ impl PartitionedComputeState {
                                     keep.push(keep_coll);
                                 }
                             }
+                            tracked.stashed_result_size =
+                                keep.iter().map(|c| c.byte_len()).sum();
                             tracked.stashed_updates = Ok(keep);
                             Ok(ship)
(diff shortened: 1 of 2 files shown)


teskje (approved):
🙈

antiguru:
Regression test fails on main:
```
> CREATE TABLE t_subscribe_cumulative (a int4, b text);
rows match; continuing at ts 1776244423.6271753, took 0.023442199s
>> INSERT INTO t_subscribe_cumulative SELECT generate_series, repeat('a', 100) FROM generate_series(1, 3000);
>> BEGIN;
>> DECLARE c CURSOR FOR SUBSCRIBE t_subscribe_cumulative;
>> FETCH ALL c WITH (timeout = '5s');
>> INSERT INTO t_subscribe_cumulative SELECT generate_series + 100000, repeat('a', 100) FROM generate_series(1, 3000);
>> FETCH ALL c WITH (timeout = '5s');
>> INSERT INTO t_subscribe_cumulative SELECT generate_series + 200000, repeat('a', 100) FROM generate_series(1, 3000);
>> FETCH ALL c WITH (timeout = '5s');
>> INSERT INTO t_subscribe_cumulative SELECT generate_series + 300000, repeat('a', 100) FROM generate_series(1, 3000);
>> FETCH ALL c WITH (timeout = '5s');
^^^ +++
+++ !!! Error Report
files involved: oom.td
oom.td:117:1: error: executing postgres query: db error: ERROR: total result exceeds max size of 1.0 MiB
```

PR #36106, opened 2026-04-15
catalog: extract make_privileges_sql helper

teskje:
This commit extracts a `make_priviliges_sql` helper function from the code building `mz_builtin_materialized_views`. We will use this also for the definitions of other `mz_builtin_*` views.

It also adds a system user ownership privileges to `mz_builtin_materialized_views.privileges`, which was missed previously.

### Motivation

Part of SQL-118


mtabebe (approved):

teskje:
TFTR!

PR #36107, opened 2026-04-15
catalog,adapter: convert mz_connections and mz_secrets into view over the catalog

teskje:
This commit converts the `mz_connections` and `mz_secrets` builtin tables into views over the catalog.

Note that there is no need to introduce an `mz_builtin_connections` view to expose the existing `mz_analytics` builtin connection because that connection is inserted into the catalog, as a way to make it runtime alterable.

### Motivation

Part of SQL-118


src/adapter/src/catalog/open/builtin_schema_migration.rs
```diff
@@ -158,6 +158,18 @@ static MIGRATIONS: LazyLock<Vec<MigrationStep>> = LazyLock::new(|| {
             MZ_CATALOG_SCHEMA,
             "mz_materialized_views",
         ),
+        MigrationStep::replacement(
+            "26.21.0-dev.0",
```
ggevay, line 162: (This will need to be updated to 22, I think.)

ggevay (approved):

teskje:
TFTR!

PR #36108, opened 2026-04-15
Select as of at least an object list

frankmcsherry:
Claude wrote this in a voice that sounds like I pretended to write it. I didn't write it. -Frank

  `AS OF AT LEAST FRONTIER OF <objects>` — RFC / eyes-wanted                                
                                                                                          
  Extends the `SELECT ... AS OF AT LEAST <timestamp>` surface to accept a list of object
  names instead of a literal timestamp:                                                   

```sql                                                            
  SELECT * FROM v AS OF AT LEAST FRONTIER OF src_a, src_b, mv_c;                          
  SUBSCRIBE v AS OF AT LEAST FRONTIER OF src_a;                                           
```
  Motivation                                                                              
                                                                                          
  A user (or tool) has observed some set of upstream objects and wants their next read to 
  be at least as fresh as what they already saw. Today they'd have to query mz_frontiers
  themselves, do frontier - 1 arithmetic, and feed the literal back in — racy and awkward.
   This lets the system do it in a single statement.        

  Semantics: take max(upper.step_back()) across the named objects as the AT LEAST lower   
  bound for timestamp selection. "Readable timestamp" = write frontier minus one.
                                                                                          
  Scope                                                     

  Only sources, tables, and materialized views are accepted. Indexes and other compute    
  objects are rejected at plan time — keeps the frontier lookup to the storage controller
  and sidesteps cross-cluster considerations for a first cut. We can revisit if there's   
  demand.                                                   

  Error behavior

  If any named object has no readable timestamp yet (write frontier is [T::minimum()] or  
  empty), the query errors rather than silently ignoring that input. Rationale: the whole
  framing is "be at least as fresh as the user's observation," which is meaningless for an
   object the user couldn't have observed. Happy to soften this to "skip with warning" if
  reviewers prefer.

  Implementation sketch

  - Parser: new AsOf::AtLeastFrontierOf(Vec<ItemName>) variant; FRONTIER added as a       
  non-reserved keyword.
  - Plan: new QueryWhen::AtLeastFrontierOf(Vec<CatalogItemId>) — resolution is deferred to
   sequencing time because frontiers move. Planner rejects non-storage item types with a  
  clear message.
  - Sequencer (Coordinator::resolve_frontier_of in timestamp_selection.rs): at            
  determine_timestamp entry, reads write frontiers via                                    
  controller.storage.collection_frontiers, computes the max readable timestamp, and
  substitutes in an ordinary AtLeastTimestamp before delegating to the existing           
  timestamp-selection path. This means the new variant reuses all existing read-hold,
  linearizability, and timeline-selection machinery unchanged.

  Things I'd like feedback on

  1. Surface syntax: AT LEAST FRONTIER OF <names> vs. something shorter like AT LEAST OF  
  <names> or introducing a function (frontier_of(...)) that could compose with AT LEAST 
  <expr>. The function form is more uniform but requires plumbing catalog lookups through 
  expression evaluation.                                    
  2. Unhydrated → error vs. skip: above.
  3. Tables with multiple GlobalIds: I use latest_global_id() — i.e. "freshest on the     
  current schema." Alternative is to take the max across global_ids(), though older shards
   should be strictly behind.                                                             
  4. Interaction with AS OF AT LEAST <expr>: currently you choose one or the other. Worth 
  supporting both simultaneously (AT LEAST <expr> AND FRONTIER OF <names>)? Probably yes  
  eventually, no for this PR.
  5. Strict-serializable isolation: the resolved timestamp flows through the existing     
  oracle path, so real-time bounds should be respected by construction. But confirmation  
  from someone who owns that invariant would be welcome.
                                                                                          
  Testing                                                   

  Compiles clean and parser tests pass. No e2e tests yet — will add testdrive coverage    
  (happy path, unhydrated error, unsupported object type, subscribe) before merging.
  Hand-testing locally is the immediate next step.  

Edit from @ggevay: Linear: https://linear.app/materializeinc/project/select-as-of-at-least-c7835ea23d60/overview

### DIFF
--- a/src/adapter/src/coord/timestamp_selection.rs
+++ b/src/adapter/src/coord/timestamp_selection.rs
@@ -594,6 +594,44 @@ impl Coordinator {
         oracle_read_ts
     }
 
+    /// Resolves `QueryWhen::AtLeastFrontierOf` by reading the current write
+    /// frontiers of the named storage collections and converting to the
+    /// equivalent `AtLeastTimestamp`. Errors if any named collection has not
+    /// yet produced any output (write frontier is `[T::minimum()]` or empty),
+    /// since there is no readable timestamp to constrain against.
+    fn resolve_frontier_of(&self, when: &QueryWhen) -> Result<Option<QueryWhen>, AdapterError> {
+        let QueryWhen::AtLeastFrontierOf(item_ids) = when else {
+            return Ok(None);
+        };
+        let mut max_ts: Option<Timestamp> = None;
+        for item_id in item_ids {
+            let entry = self.catalog().get_entry(item_id);
+            let gid = entry.latest_global_id();
+            let (_since, upper) =
+                self.controller
+                    .storage
+                    .collection_frontiers(gid)
+                    .map_err(|_| {
+                        AdapterError::Internal(format!(
+                            "AS OF AT LEAST FRONTIER OF: no storage collection for {}",
+                            entry.name().item,
+                        ))
+                    })?;
+            let readable = upper.as_option().and_then(|upper| upper.step_back());
+            let Some(readable) = readable else {
+                return Err(AdapterError::Unstructured(anyhow::anyhow!(
+                    "AS OF AT LEAST FRONTIER OF: {} has no readable timestamp yet \
+                     (collection is unhydrated or closed)",
+                    entry.name().item,
+                )));
+            };
+            max_ts = Some(max_ts.map_or(readable, |cur| cur.max(readable)));
+        }
+        // `item_ids` was non-empty because we matched `AtLeastFrontierOf`, but
+        // the list could in principle be empty; treat that as no constraint.
+        Ok(max_ts.map(QueryWhen::AtLeastTimestamp))
+    }
+
     /// Determines the timestamp for a query, acquires read holds that ensure the
     /// query remains executable at that time, and returns those.
     /// The caller is responsible for eventually dropping those read holds.
@@ -608,6 +646,11 @@ impl Coordinator {
         oracle_read_ts: Option<Timestamp>,
         real_time_recency_ts: Option<mz_repr::Timestamp>,
     ) -> Result<(TimestampDetermination, ReadHolds), AdapterError> {
+        // Resolve `AS OF AT LEAST FRONTIER OF <names>` into a concrete
+        // `AtLeastTimestamp` using current write frontiers. This must happen
+        // here rather than at plan time because the frontiers move.
+        let resolved_when = self.resolve_frontier_of(when)?;
+        let when = resolved_when.as_ref().unwrap_or(when);
         let isolation_level = session.vars().transaction_isolation();
         let (det, read_holds) = self.determine_timestamp_for(
             session,
--- a/src/sql/src/plan.rs
+++ b/src/sql/src/plan.rs
@@ -1936,6 +1936,11 @@ pub enum QueryWhen {
     /// Same as Immediately, but will also advance to at least the specified
     /// expression.
     AtLeastTimestamp(Timestamp),
+    /// Same as Immediately, but will also advance to at least the largest
+    /// readable timestamp (i.e. write frontier minus one) across the named
+    /// objects. The list is resolved to concrete frontiers by the sequencer
+    /// at query time.
+    AtLeastFrontierOf(Vec<CatalogItemId>),
 }
 
 impl QueryWhen {
@@ -1943,7 +1948,9 @@ impl QueryWhen {
     pub fn advance_to_timestamp(&self) -> Option<Timestamp> {
         match self {
             QueryWhen::AtTimestamp(t) | QueryWhen::AtLeastTimestamp(t) => Some(t.clone()),
-            QueryWhen::Immediately | QueryWhen::FreshestTableWrite => None,
+            QueryWhen::Immediately
+            | QueryWhen::FreshestTableWrite
+            | QueryWhen::AtLeastFrontierOf(_) => None,
         }
     }
     /// Returns whether the candidate's upper bound is constrained.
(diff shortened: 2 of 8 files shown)


src/adapter/src/coord/timestamp_selection.rs
```diff
@@ -608,6 +646,11 @@ impl Coordinator {
         oracle_read_ts: Option<Timestamp>,
         real_time_recency_ts: Option<mz_repr::Timestamp>,
     ) -> Result<(TimestampDetermination, ReadHolds), AdapterError> {
+        // Resolve `AS OF AT LEAST FRONTIER OF <names>` into a concrete
+        // `AtLeastTimestamp` using current write frontiers. This must happen
+        // here rather than at plan time because the frontiers move.
```
ggevay, line 651: > This must happen
> here rather than at plan time because the frontiers move.

I'm not sure if this is sound reasoning: I heard somewhere that the point of the feature is to capture data that is in the specified inputs at the moment when the user "presses enter", right? Both planning and `determine_timestamp` are after the user "presses enter", so either could be fine?

But the more important reason for doing it here is that planning generally doesn't look at frontiers, and doesn't really have access to them.

src/adapter/src/coord/timestamp_selection.rs
```diff
@@ -594,6 +594,44 @@ impl Coordinator {
         oracle_read_ts
     }
 
+    /// Resolves `QueryWhen::AtLeastFrontierOf` by reading the current write
+    /// frontiers of the named storage collections and converting to the
+    /// equivalent `AtLeastTimestamp`. Errors if any named collection has not
+    /// yet produced any output (write frontier is `[T::minimum()]` or empty),
```
ggevay, line 600: > or empty

That's the opposite of "not yet produced any output"

src/adapter/src/coord/timestamp_selection.rs
```diff
@@ -594,6 +594,44 @@ impl Coordinator {
         oracle_read_ts
     }
 
+    /// Resolves `QueryWhen::AtLeastFrontierOf` by reading the current write
+    /// frontiers of the named storage collections and converting to the
+    /// equivalent `AtLeastTimestamp`. Errors if any named collection has not
```
ggevay, line 599: > equivalent `AtLeastTimestamp`

Maybe specify it more explicitly that it's the _max_ of the write frontiers of the given objects

src/adapter/src/coord/timestamp_selection.rs
```diff
@@ -608,6 +646,11 @@ impl Coordinator {
         oracle_read_ts: Option<Timestamp>,
         real_time_recency_ts: Option<mz_repr::Timestamp>,
     ) -> Result<(TimestampDetermination, ReadHolds), AdapterError> {
+        // Resolve `AS OF AT LEAST FRONTIER OF <names>` into a concrete
+        // `AtLeastTimestamp` using current write frontiers. This must happen
+        // here rather than at plan time because the frontiers move.
+        let resolved_when = self.resolve_frontier_of(when)?;
+        let when = resolved_when.as_ref().unwrap_or(when);
```
ggevay, line 653: This is called only from the old peek sequencing (which is no longer used in most cases, and I'll remove it soon), and in non-peek code paths. So, we'll need to do the same also in the frontend peek sequencing, i.e., `frontend_determine_timestamp`.

src/adapter/src/coord/timestamp_selection.rs
```diff
@@ -594,6 +594,44 @@ impl Coordinator {
...
+            return Ok(None);
+        };
+        let mut max_ts: Option<Timestamp> = None;
+        for item_id in item_ids {
+            let entry = self.catalog().get_entry(item_id);
+            let gid = entry.latest_global_id();
+            let (_since, upper) =
+                self.controller
+                    .storage
+                    .collection_frontiers(gid)
```
ggevay, line 613: This wouldn't work for indexes, because it's asking only the storage controller, right?

ggevay: See `acquire_read_holds_and_least_valid_write` for how the frontend peek sequencing accesses frontiers. We'll need to integrate `AtLeastFrontierOf` into that in some way. (It's not so easy to access frontiers efficiently in the frontend peek sequencing, because we need to send a message for accessing the Compute controller.)

src/sql/src/plan.rs
```diff
@@ -1936,14 +1936,21 @@ pub enum QueryWhen {
...
 
 impl QueryWhen {
     /// Returns a timestamp to which the candidate must be advanced.
     pub fn advance_to_timestamp(&self) -> Option<Timestamp> {
         match self {
             QueryWhen::AtTimestamp(t) | QueryWhen::AtLeastTimestamp(t) => Some(t.clone()),
-            QueryWhen::Immediately | QueryWhen::FreshestTableWrite => None,
+            QueryWhen::Immediately
+            | QueryWhen::FreshestTableWrite
+            | QueryWhen::AtLeastFrontierOf(_) => None,
```
ggevay, line 1953: It feels a bit awkward to me that the `QueryWhen::AtLeastFrontierOf` match arms of all these methods on `QueryWhen` are basically dead code. For example, I happened to look at `advance_to_timestamp` when starting to review the PR before looking at `resolve_frontier_of`, and I was like, "wait, we do need to advance to some timestamp here, no?", before I realized that the `None` doesn't matter here, because it's just dead code. But I'm not sure what would be a better way to organize this code.

ggevay:
(Linking the accompanying slack thread, which has more context: https://materializeinc.slack.com/archives/C08A62E0751/p1776435415687709?thread_ts=1776285580.267439&cid=C08A62E0751 )

src/adapter/src/coord/timestamp_selection.rs
ggevay: Actually, I see that the PR description calls this out that it only works with storage frontiers. And this might even be fine, because the intended use case ([Linear](https://linear.app/materializeinc/project/select-as-of-at-least-c7835ea23d60/overview)) is to put such collections in the `AS OF AT LEAST` object list that get sinked out, and it's only possible to sink out storage collections (sources, tables, MVs). @maheshwarip, can you confirm?

PR #36114, opened 2026-04-16
sql: Don't panic when details are set explicitly

def-:
Fixes: https://github.com/MaterializeInc/database-issues/issues/10008

I'll scan a bit through the code base to find similar occurrences of where a query causes a panic but should just fail the query instead.

### DIFF
--- a/src/sql/src/pure.rs
+++ b/src/sql/src/pure.rs
@@ -29,8 +29,8 @@ use mz_ore::collections::CollectionExt;
 use mz_ore::error::ErrorExt;
 use mz_ore::future::InTask;
 use mz_ore::iter::IteratorExt;
+use mz_ore::soft_panic_or_log;
 use mz_ore::str::StrExt;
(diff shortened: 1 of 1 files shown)


ggevay (approved):
Thanks!

Btw. there is also the question of why people are typing it in the first place. (It's not in the documentation, obviously.) I think the reason is that `SHOW CREATE TABLE` is showing it, and they copy-paste that. You can see an example if you do
```
show create table qa_canary_environment.public_pg_cdc.pg_people;
```
in the prod sandbox.

We might want to consider making `SHOW CREATE TABLE` hide it. This is what we do with the `AS OF` that we automatically add to `CREATE MATERIALIZED VIEW` statements, i.e., `SHOW CREATE MATERIALIZED VIEW` just doesn't show it.

def-:
Can confirm, that's why I ran into it with workload replay.

def-:
Doing it for more panics is probably more controversial: https://github.com/MaterializeInc/materialize/pull/36120

PR #36115, opened 2026-04-16
compute,storage: make more things not generic over the timestamp type

teskje:
This PR advances the effort of removing unneeded timestamp genericity from the code base. It focuses on occurrences in `mz-compute-client`, `mz-compute-types`, and `mz-storage-types`. See individual commits.


martykulma (approved):
lgtm!

teskje:
TFTR!

PR #36116, opened 2026-04-16
mcp: expose all RBAC-accessible views as data products without requiring indexes or comments

bobbyiliev:
Fixes: https://linear.app/materializeinc/issue/DEX-23/mcp-agent-expose-all-rbac-accessible-views-as-data-products-without

### DIFF
--- a/doc/user/content/reference/system-catalog/mz_internal.md
+++ b/doc/user/content/reference/system-catalog/mz_internal.md
@@ -542,9 +542,10 @@ each materialized view with a refresh strategy other than `on-commit`.
 
 ## `mz_mcp_data_products`
 
-The `mz_mcp_data_products` view lists data products (i.e., indexed materialized
-views) that are available through the Model Context Protocol (MCP) server and
-that the current user can access. This is a lightweight discovery view. Use
+The `mz_mcp_data_products` view lists data products (materialized views)
+that are available through the Model Context Protocol (MCP) server and
+that the current user has SELECT access on. Indexes and comments are optional
+enrichment. This is a lightweight discovery view. Use
 [`mz_mcp_data_product_details`](#mz_mcp_data_product_details) for full column
 schema information.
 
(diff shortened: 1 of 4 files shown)


ggevay:
Hmm, but wouldn't including views run the risk that agents would accidentally submit heavy queries? (Tables and MVs have this issue less than views, since their contents are available from at least Persist. Persist is still slower than accessing an index from memory, but not as slow as recomputing something from scratch.)

bobbyiliev:
Ah yes good call @ggevay! I just removed views and now only exposes MVs

doc/user/content/reference/system-catalog/mz_internal.md
```diff
@@ -564,8 +565,8 @@ with a JSON Schema describing each data product's columns and types.
 | Field         | Type     | Meaning                                                                                  |
 | ------------- | -------- | ---------------------------------------------------------------------------------------- |
 | `object_name` | [`text`] | Fully qualified object name (database.schema.name).                                      |
-| `cluster`     | [`text`] | Cluster where the index is hosted.                                                       |
-| `description` | [`text`] | Index comment (used as data product description).                                        |
+| `cluster`     | [`text`] | Cluster where the object or its index is hosted (NULL if no cluster).                    |
```
ggevay: Maybe explicitly mention that it's fine to access them from any cluster in the NULL case, to make it clear what does it mean when there is "no cluster"? (Btw., this will be NULL for MVs, right? It's tricky, because MVs are computed on clusters, but you can read them from any cluster all the same.)

src/catalog/src/builtin.rs
```diff
@@ -11640,43 +11642,49 @@ pub static MZ_MCP_DATA_PRODUCTS: LazyLock<BuiltinView> = LazyLock::new(|| Builti
...
         ),
-        ("cluster", "Cluster where the index is hosted."),
+        (
+            "cluster",
+            "Cluster where the object or its index is hosted (NULL if no cluster).",
+        ),
         (
             "description",
-            "Index comment (used as data product description).",
+            "Object or index comment (used as data product description).",
```
ggevay: (What if both the index and he underlying object has a comment? Do we / should we include both somehow?)

ggevay (approved):

doc/user/content/reference/system-catalog/mz_internal.md
bobbyiliev: Good catch! Updated the description.

jubrad (approved):

PR #36119, opened 2026-04-16
catalog,adapter: convert mz_indexes table to materialized view

teskje:
This commit converts the `mz_indexes` builtin table into a view over the
catalog. As part of this, it introduces an internal `mz_builtin_indexes`
view exposing the builtin indexes to SQL.

### Motivation

Part of [SQL-118](https://linear.app/materializeinc/issue/SQL-118)


teskje:
Currently misses compute introspection indexes. They are in the catalog as `ClusterIntrospectionSourceIndex` entries, so it should be possible to add them by unioning in these entries.

PR #36120, opened 2026-04-16
adapter: Convert many recoverable panics to query errors

def-:
When a query is bad, even because of a bug from our side, we shouldn't bring down `environmentd` but fail the query.

### DIFF
--- a/src/sql/src/plan/query.rs
+++ b/src/sql/src/plan/query.rs
@@ -2015,7 +2015,7 @@ fn plan_set_expr(
             ) -> Result<(HirRelationExpr, Scope), PlanError> {
                 let rows = vec![plan.row.iter().collect::<Vec<_>>()];
                 let desc = desc.relation_desc.ok_or_else(|| {
-                    sql_err!("internal error: statement description missing relation descriptor")
+                    PlanError::Internal("statement description missing relation descriptor".into())
                 })?;
                 let scope = Scope::from_source(None, desc.iter_names());
                 let expr = HirRelationExpr::constant(rows, desc.into_typ());
@@ -4100,24 +4100,24 @@ fn plan_expr_inner<'a>(
         Expr::MapSubquery(query) => plan_map_subquery(ecx, query),
         Expr::ArraySubquery(query) => plan_array_subquery(ecx, query),
         Expr::Collate { expr, collation } => plan_collate(ecx, expr, collation),
-        Expr::Nested(_) => sql_bail!("internal error: Expr::Nested should have been desugared"),
+        Expr::Nested(_) => bail_internal!("Expr::Nested should have been desugared"),
         Expr::InSubquery { .. } => {
-            sql_bail!("internal error: Expr::InSubquery should have been desugared")
+            bail_internal!("Expr::InSubquery should have been desugared")
         }
         Expr::AnyExpr { .. } => {
-            sql_bail!("internal error: Expr::AnyExpr should have been desugared")
+            bail_internal!("Expr::AnyExpr should have been desugared")
         }
         Expr::AllExpr { .. } => {
-            sql_bail!("internal error: Expr::AllExpr should have been desugared")
+            bail_internal!("Expr::AllExpr should have been desugared")
         }
         Expr::AnySubquery { .. } => {
-            sql_bail!("internal error: Expr::AnySubquery should have been desugared")
+            bail_internal!("Expr::AnySubquery should have been desugared")
         }
         Expr::AllSubquery { .. } => {
-            sql_bail!("internal error: Expr::AllSubquery should have been desugared")
+            bail_internal!("Expr::AllSubquery should have been desugared")
         }
         Expr::Between { .. } => {
-            sql_bail!("internal error: Expr::Between should have been desugared")
+            bail_internal!("Expr::Between should have been desugared")
         }
     }
 }
@@ -5176,7 +5176,7 @@ fn plan_aggregate_common(
 
     let impls = match resolve_func(ecx, name, args)? {
         Func::Aggregate(impls) => impls,
-        _ => sql_bail!("internal error: plan_aggregate_common called on non-aggregate function"),
+        _ => bail_internal!("plan_aggregate_common called on non-aggregate function"),
     };
 
     // We follow PostgreSQL's rule here for mapping `count(*)` into the
@@ -5268,7 +5268,7 @@ fn plan_identifier(ecx: &ExprContext, names: &[Ident]) -> Result<HirScalarExpr, 
     let col_name = normalize::column_name(
         names
             .pop()
-            .ok_or_else(|| sql_err!("internal error: empty identifier"))?,
+            .ok_or_else(|| PlanError::Internal("empty identifier".into()))?,
     );
 
     // If the name is qualified, it must refer to a column in a table.
@@ -5549,7 +5549,7 @@ fn plan_function<'a>(
     };
 
     if over.is_some() {
-        sql_bail!("internal error: OVER clause should have been handled");
+        bail_internal!("OVER clause should have been handled by the window function path above");
     }
 
     if *distinct {
@@ -6037,7 +6037,7 @@ pub fn scalar_type_from_sql(
         ResolvedDataType::Named { id, modifiers, .. } => {
             scalar_type_from_catalog(scx.catalog, *id, modifiers)
         }
-        ResolvedDataType::Error => sql_bail!("internal error: unresolved data type"),
+        ResolvedDataType::Error => bail_internal!("should have been caught in name resolution"),
     }
 }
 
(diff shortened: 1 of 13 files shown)


ggevay:
I'm thinking we could make most of these `AdapterError::Internal`. The important difference is that hopefully at some point we'll implement Sentry reporting for query errors that are marked as internal.

src/sql/src/plan/statement/ddl.rs
```diff
@@ -1608,14 +1608,9 @@ pub fn plan_create_subsource(
...
-    // creating the AST for subsources as a response to CREATE SOURCE
-    // statements, so this would fire in integration testing if we failed to
-    // uphold it.
-    assert!(
-        progress ^ (external_reference.is_some() && of_source.is_some()),
-        "CREATE SUBSOURCE statement must specify either PROGRESS or REFERENCES option"
-    );
+    if !(progress ^ (external_reference.is_some() && of_source.is_some())) {
+        sql_bail!("CREATE SUBSOURCE statement must specify either PROGRESS or REFERENCES option");
+    }
```
ggevay, line 1613: The deleted comment is still useful.

src/sql/src/plan/statement/ddl.rs
```diff
@@ -5635,11 +5641,9 @@ pub fn plan_drop_objects(
         cascade,
     }: DropObjectsStatement,
 ) -> Result<Plan, PlanError> {
-    assert_ne!(
-        object_type,
-        mz_sql_parser::ast::ObjectType::Func,
-        "rejected in parser"
-    );
+    if object_type == mz_sql_parser::ast::ObjectType::Func {
+        sql_bail!("DROP FUNCTION is not supported");
```
ggevay, line 5645: We have the `bail_unsupported!` macro for these things. (It has a standardized error msg, but also hopefully we'll have statistics about these at some point, so that we know what unsupported features are the most requested.)

src/sql/src/plan/statement/ddl.rs
```diff
@@ -7507,11 +7514,11 @@ pub fn plan_alter_sink(
             // First we reconstruct the original CREATE SINK statement
             let create_sql = item.create_sql();
             let stmts = mz_sql_parser::parser::parse_statements(create_sql)?;
-            let [stmt]: [StatementParseResult; 1] = stmts
-                .try_into()
-                .expect("create sql of sink was not exactly one statement");
+            let [stmt]: [StatementParseResult; 1] = stmts.try_into().map_err(|_| {
+                sql_err!("internal error: create SQL of sink was not exactly one statement")
```
ggevay, line 7518: `AdapterError::Internal`

ggevay (commented):
I think it's a good direction to make stuff less panicky!

src/sql/src/plan/statement/ddl.rs
ggevay: We might want to even create a macro for it, and put it just before `sql_bail`, so that it's easy to find. Something like `bail_internal_error`.

djahandarie:
This looks great, thank you! I'd also encourage a lint at some point if you haven't already considered it (clippy::unwrap_used, clippy::expect_used, clippy::panic, clippy::unreachable, etc)

src/sql/src/plan/statement/acl.rs
```diff
@@ -79,7 +79,9 @@ pub fn plan_alter_owner(
         (ObjectType::NetworkPolicy, UnresolvedObjectName::NetworkPolicy(name)) => {
             plan_alter_network_policy_owner(scx, if_exists, name, new_owner.id)
         }
-        (ObjectType::Role, UnresolvedObjectName::Role(_)) => unreachable!("rejected by the parser"),
+        (ObjectType::Role, UnresolvedObjectName::Role(_)) => {
+            sql_bail!("cannot ALTER OWNER of a role")
```
ggevay, line 84: I'd say it's worth preserving that we think the parser should have rejected this

src/sql/src/plan/statement/acl.rs
```diff
@@ -97,7 +99,7 @@ pub fn plan_alter_owner(
             | name @ UnresolvedObjectName::NetworkPolicy(_)
             | name @ UnresolvedObjectName::Role(_),
         ) => {
-            unreachable!("parser set the wrong object type '{object_type:?}' for name {name:?}")
+            sql_bail!("invalid object type '{object_type}' for ALTER OWNER with name {name}")
```
ggevay, line 104: Same here: Would be good to preserve the info that we think the parser messed this up

src/sql/src/plan/statement/acl.rs
```diff
@@ -500,13 +502,16 @@ fn plan_update_privilege(
...
-                        name.try_into()
-                            .expect("name resolution should handle invalid objects")
-                    })
-                    .collect(),
+                GrantTargetSpecificationInner::Objects { names } => {
+                    let mut ids = Vec::with_capacity(names.len());
+                    for name in names {
+                        ids.push(
+                            name.try_into()
+                                .map_err(|e| sql_err!("invalid object name: {}", e))?,
```
ggevay, line 513: Same here: keep referencing name resolution

src/sql/src/plan/statement/ddl.rs
```diff
@@ -1188,7 +1188,7 @@ fn plan_postgres_source_connection(
     Ok(PostgresSourceConnection {
         connection: connection_item.id(),
         connection_id: connection_item.id(),
-        publication: publication.expect("validated exists during purification"),
+        publication: publication.ok_or_else(|| sql_err!("PUBLICATION option is required"))?,
```
ggevay, line 1192: Same: Keep referencing purification

src/sql/src/plan/statement/ddl.rs
```diff
@@ -1214,7 +1214,7 @@ fn plan_kafka_source_connection(
         start_offset,
         seen: _,
     }: KafkaSourceConfigOptionExtracted = options.clone().try_into()?;
-    let topic = topic.expect("validated exists during purification");
+    let topic = topic.ok_or_else(|| sql_err!("TOPIC option is required"))?;
```
ggevay, line 1219: Same: Keep referencing purification

src/sql/src/plan/query.rs
```diff
@@ -6040,7 +6067,7 @@ pub fn scalar_type_from_sql(
         ResolvedDataType::Named { id, modifiers, .. } => {
             scalar_type_from_catalog(scx.catalog, *id, modifiers)
         }
-        ResolvedDataType::Error => unreachable!("should have been caught in name resolution"),
+        ResolvedDataType::Error => bail_internal!("unresolved data type"),
```
ggevay: Same: Keep referencing name resolution

src/sql/src/plan/query.rs
```diff
@@ -6647,7 +6674,7 @@ impl<'a> QueryContext<'a> {
 
                 Ok((expr, scope))
             }
-            ResolvedItemName::Error => unreachable!("should have been caught in name resolution"),
+            ResolvedItemName::Error => bail_internal!("unresolved item name"),
```
ggevay: Same: Keep referencing name resolution

src/sql/src/pure/mysql.rs
```diff
@@ -132,7 +132,9 @@ pub(super) fn generate_source_export_statement_values(
         initial_gtid_set,
     } = purified_export.details
     else {
-        unreachable!("purified export details must be mysql")
+        return Err(sql_err!(
+            "internal error: purified export details must be mysql"
```
ggevay: `PlanError::Internal`?

src/sql/src/pure/postgres.rs
```diff
@@ -230,7 +230,9 @@ pub(super) fn generate_source_export_statement_values(
         exclude_columns,
     } = purified_export.details
     else {
-        unreachable!("purified export details must be postgres")
+        return Err(sql_err!(
+            "internal error: purified export details must be postgres"
```
ggevay: `PlanError::Internal`?

src/sql/src/pure/sql_server.rs
```diff
@@ -410,7 +410,9 @@ pub(super) fn generate_source_export_statement_values(
         initial_lsn,
     } = purified_export.details
     else {
-        unreachable!("purified export details must be SQL Server")
+        return Err(sql_err!(
+            "internal error: purified export details must be SQL Server"
```
ggevay: `PlanError::Internal`?

ggevay (commented):
I wrote some more comments. It would be good to preserve the information of where these invariants are coming from (maybe not in the user-facing error msgs, but at least in code comments).

def-:
> This looks great, thank you! I'd also encourage a lint at some point if you haven't already considered it (clippy::unwrap_used, clippy::expect_used, clippy::panic, clippy::unreachable, etc)

I'd be very happy with that, but there will have to be more discussion about which errors we should *not* recover from because something is so wrong with Materialize that we could be persisting or returning wrong results, which would of course be even worse than panicking.

def-:
Ready for review

ggevay:
Thanks for iterating on this! I started writing some more comments, but then I realized it will be smoother to just do them. So, I've added a bunch of commits to the PR. I hope that's ok! Please review my commits. Ready to merge from my side.

ggevay (approved):

def- (commented):
Had to rebase it, otherwise LGTM

PR #36121, opened 2026-04-16
race-condition test: Extend & reenable

def-:
We currently have some apparent race condition issues we can't reproduce, so would be good to bring this test back. Unfortunately this found nothing new.


bosconi (approved):
Looks good! (Claude agreed :)

PR #36122, opened 2026-04-16
mzcompose: More parallel docker pulls

def-:
Faster to do it in parallel in the beginning, especially if we are still waiting for the build to finish.

Test run: https://buildkite.com/materialize/nightly/builds/16043


bosconi (approved):
Looks good.

PR #36127, opened 2026-04-16
dbt-materialize: Support connection overrides

SangJunBak:
### Motivation
Unblocks OIDC auth because our dbt adapter doesn't allow inputting arbitrary options: https://materializeinc.slack.com/archives/C0A23PK183Z/p1776360806044789

## Verification
1. Had a local materialize instance running via bin/environmentd
2. Saved following profiles.yml in the current directory:
```
dbt_test:
  target: dev
  outputs:
    dev:
      type: materialize
      host: localhost
      port: 6875
      user: materialize
      pass: materialize
      database: materialize
      schema: public
      options:
        oidc_auth_enabled: "true"
        welcome_message: "on"
```

3. Ran 
```
cd misc/dbt-materialize
uv venv dbt-venv
uv pip install --python dbt-venv/bin/python pip
dbt-venv/bin/pip install .
dbt-venv/bin/dbt debug --profiles-dir .
```

4. Verified output had:
```
18:43:49    application_name: dbt-materialize v1.9.8
18:43:49    options: {'oidc_auth_enabled': 'true', 'welcome_message': 'on'}
```

Wasn't sure the best way to automate a unit test for this however.

### DIFF
--- a/misc/dbt-materialize/dbt/adapters/materialize/__version__.py
+++ b/misc/dbt-materialize/dbt/adapters/materialize/__version__.py
@@ -15,4 +15,4 @@
 # limitations under the License.
 
 # If you bump this version, bump it in setup.py too.
-version = "1.9.7"
+version = "1.9.8"
--- a/misc/dbt-materialize/dbt/adapters/materialize/__version__.py
+++ b/misc/dbt-materialize/dbt/adapters/materialize/__version__.py
@@ -15,4 +15,4 @@
 # limitations under the License.
 
 # If you bump this version, bump it in setup.py too.
-version = "1.9.8"
+version = "1.9.7"
--- a/misc/dbt-materialize/dbt/adapters/materialize/connections.py
+++ b/misc/dbt-materialize/dbt/adapters/materialize/connections.py
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 from dataclasses import dataclass
-from typing import Optional
+from typing import Dict, Optional
 
 import dbt_common.exceptions
 import psycopg2
@@ -31,43 +31,27 @@ SUPPORTED_MATERIALIZE_VERSIONS = ">=0.68.0"
 
 logger = AdapterLogger("Materialize")
 
-
-# Override the psycopg2 connect function in order to inject Materialize-specific
-# session parameter defaults.
-#
-# This approach is a bit hacky, but some of these session parameters *must* be
-# set as part of connection initiation, so we can't simply run `SET` commands
-# after the session is established.
-def connect(**kwargs):
-    options = [
-        # Ensure that dbt's catalog queries get routed to the
-        # `mz_catalog_server` cluster, even if the server or role's default is
-        # different.
-        "--auto_route_catalog_queries=on",
-        # dbt prints notices to stdout, which is very distracting because dbt
-        # can establish many new connections during `dbt run`.
-        "--welcome_message=off",
-        # Disable warnings about the session's default database or cluster not
-        # existing, as these get quite spammy, especially with multiple threads.
-        #
-        # Details: it's common for the default cluster for the role dbt is
-        # connecting as (often `quickstart`) to be absent. For many dbt
-        # deployments, clusters are explicitly specified on a model-by-model
-        # basis, and there in fact is no natural "default" cluster. So warning
-        # repeatedly that the default cluster doesn't exist isn't helpful, since
-        # each DDL statement will specify a different, valid cluster. If a DDL
-        # statement ever specifies an invalid cluster, dbt will still produce an
-        # error about the invalid cluster, even with this setting enabled.
-        "--current_object_missing_warnings=off",
-        *(kwargs.get("options") or []),
-    ]
-    kwargs["options"] = " ".join(options)
-
-    return _connect(**kwargs)
-
-
-_connect = psycopg2.connect
-psycopg2.connect = connect
+DEFAULT_SESSION_PARAMETERS = {
+    # Ensure that dbt's catalog queries get routed to the
+    # `mz_catalog_server` cluster, even if the server or role's default is
+    # different.
+    "auto_route_catalog_queries": "on",
+    # dbt prints notices to stdout, which is very distracting because dbt
+    # can establish many new connections during `dbt run`.
+    "welcome_message": "off",
+    # Disable warnings about the session's default database or cluster not
+    # existing, as these get quite spammy, especially with multiple threads.
+    #
+    # Details: it's common for the default cluster for the role dbt is
+    # connecting as (often `quickstart`) to be absent. For many dbt
+    # deployments, clusters are explicitly specified on a model-by-model
+    # basis, and there in fact is no natural "default" cluster. So warning
+    # repeatedly that the default cluster doesn't exist isn't helpful, since
+    # each DDL statement will specify a different, valid cluster. If a DDL
+    # statement ever specifies an invalid cluster, dbt will still produce an
+    # error about the invalid cluster, even with this setting enabled.
+    "current_object_missing_warnings": "off",
+}
 
 
 @dataclass
@@ -84,6 +68,9 @@ class MaterializeCredentials(PostgresCredentials):
     # modified).
     cluster: Optional[str] = None
     application_name: Optional[str] = f"dbt-materialize v{__version__}"
+    # Additional session parameters to pass via the connection options string.
+    # User-provided options override DEFAULT_SESSION_PARAMETERS.
+    options: Optional[Dict[str, str]] = None
 
     @property
     def type(self):
@@ -103,6 +90,7 @@ class MaterializeCredentials(PostgresCredentials):
             "search_path",
             "retries",
             "application_name",
+            "options",
         )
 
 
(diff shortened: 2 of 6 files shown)


SangJunBak:
@bobbyiliev Unsure how we actually deploy an update to our dbt adapter!

Also I decided to go with the approach of allowing an `options` override instead of accepting a `oidc_auth_enabled` boolean as a connection option. This is because:
1. This `oidc_auth_enabled` option doesn't exist for cloud
2. Future flexibility if we need to specify a connection option

bobbyiliev:
> @bobbyiliev Unsure how we actually deploy an update to our dbt adapter!

@SangJunBak This gets auto released after bumping the version bump (__version__.py + setup.py). What we usually do is to split the version bump itself out into a follow-up release PR. We usually merge the feature + CHANGELOG entry first, then do a separate PR to cut the release to PyPI.

Would you mind reverting the verions bump and adding a changelog entry here? Then we can have a follow-up PR for the release itself. That way if there are other pending changes they all get released together.

misc/dbt-materialize/dbt/include/materialize/profile_template.yml
```diff
@@ -41,3 +41,5 @@ prompts:
   cluster:
     hint: 'dev cluster'
     default: 'quickstart'
+  options:
+    hint: 'overrides the PostgreSQL `options` connection parameter'
```
bobbyiliev, line 45: I totally hear you that this is tricky to unit-test, but do you think we could add an integration test in `tests/adapter/` e.g. set a non-default option like `welcome_message: on` via a `dbt_profile_target` override and assert the server honored it? The `options_dict.update(credentials.options)` merge is the whole point of the feature. Happy to help sketch something out if useful!

bobbyiliev:
> 3. Ran
> ```
> cd misc/dbt-materialize
> uv venv dbt-venv
> uv pip install --python dbt-venv/bin/python pip
> dbt-venv/bin/pip install .
> dbt-venv/bin/dbt debug --profiles-dir .
> ```

Did you by any chance test this with a full dbt run / dbt build, not just dbt debug? The verification shows the options making it into the connection string, but I'd feel better knowing a real query round-trip works with an override that actually changes behavior (e.g. setting auto_route_catalog_queries: off). Happy to try this on my end if you hit any blockers setting this all up!

misc/dbt-materialize/dbt/adapters/materialize/connections.py
```diff
@@ -111,7 +99,78 @@ class MaterializeConnectionManager(PostgresConnectionManager):
 
     @classmethod
     def open(cls, connection):
-        connection = super().open(connection)
+        # Much of the `open` method setup is copied from the `PostgresConnectionManager.open` method
+        # https://github.com/dbt-labs/dbt-adapters/blob/v1.17.3/dbt-postgres/src/dbt/adapters/postgres/connections.py#L102,
+        # except we allow users to override options.
```
bobbyiliev, line 124: This new open method copies a chunk of `PostgresConnectionManager.open` from `dbt-postgres`. Did you consider overriding a smaller surface instead (e.g. just the options-string construction and then `super().open())`? My worry is that when `dbt-postgres` adds new kwargs or changes retry logic, we'll silently fall behind. Totally fine if you already looked at this and it wasn't workable, just curious.

misc/dbt-materialize/dbt/adapters/materialize/__version__.py
```diff
@@ -15,4 +15,4 @@
 # limitations under the License.
 
 # If you bump this version, bump it in setup.py too.
-version = "1.9.7"
+version = "1.9.8"
```
bobbyiliev: Mentioned this in the other comment but could we split the version bump (__version__.py + setup.py) out of this PR? Normally I prefer to merge the feature + document it under "Unreleased" in the CHANGELOG first, then do a separate release PR that bumps the version and publishes to PyPI. Keeps the feature history and release history cleanly separate.

bobbyiliev (changes requested):

SangJunBak:
> Did you by any chance test this with a full dbt run / dbt build, not just dbt debug? The verification shows the options making it into the connection string, but I'd feel better knowing a real query round-trip works with an override that actually changes behavior (e.g. setting auto_route_catalog_queries: off). Happy to try this on my end if you hit any blockers setting this all up!

I think the verification also tries to connect, so the way I actually tested it was toggling welcome_message to on and off and seeing the output! I can write a unit test for this however!

misc/dbt-materialize/dbt/adapters/materialize/connections.py
SangJunBak: > My worry is that when dbt-postgres adds new kwargs or changes retry logic, we'll silently fall behind

I think this is a valid concern :( I tried to but because kwargs is defined internally and there's no way to override options through the `credentials` object, we can't use super().open(). The best way I could think of is extending off the previous approach of monkey patching psycopg, then doing some weird python thread blocking such that the monkey patch can accept input from our adapter's `.open`, but that seems worse than copying their implementation. FWIW prior art to this is something like Clickhouse https://github.com/ClickHouse/dbt-clickhouse/blob/main/dbt/adapters/clickhouse/connections.py

SangJunBak: Maybe the correct implementation here however is inheriting off SQLConnectionManager instead of PostgresqlConnectionManager?

misc/dbt-materialize/dbt/adapters/materialize/__version__.py
SangJunBak: For sure!

SangJunBak:
Created an integration test in the latest commit. Can test via `bin/mzcompose --find dbt-materialize run default -k test_options_override`

SangJunBak:
Followup PR to bump the versions https://github.com/MaterializeInc/materialize/pull/36213

bobbyiliev:
@claude review

SangJunBak:
Regarding claude's output, so I did notice that spaces had no negative effect in the connection. I.e. setting the cluster to "weewoo quickstart" would still work, even with other arguments. This might be a postgres specific thing that Materialize already handles?

misc/dbt-materialize/dbt/adapters/materialize/connections.py
```diff
@@ -111,7 +99,78 @@ class MaterializeConnectionManager(PostgresConnectionManager):
...
+
+        options_dict = dict(DEFAULT_SESSION_PARAMETERS)
+        if credentials.options:
+            options_dict.update(credentials.options)
+
+        options_parts = [(k, v) for k, v in options_dict.items()]
+
+        search_path = credentials.search_path
+        if search_path is not None and search_path != "":
+            options_parts.append(("search_path", search_path))
```
bobbyiliev: Just doing a second read and I think I noticed something here, sorry I missed it the first time. The parent has `search_path.replace(" ", "\\ ")` and I had to look up why, looks like Postgres treats spaces in the `options` string as argument separators unless you backslash-escape them ([docs](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-OPTIONS)). I think with the new join we'd lose that, so a `search_path` with a space in it might silently truncate at the first space?

I think the same thing might apply to user-provided values too. If someone wrote something like `options: { welcome_message: "hello world" }`, I think we'd end up sending `--welcome_message=hello world` and the server would parse `world` as a stray positional. Not 100% sure though, would you mind double-checking? Happy to double check this myself if you wish!

If it is an issue, maybe escape spaces in values, or call it out in a docstring that values shouldn't contain spaces? Wdyt?

misc/dbt-materialize/tests/adapter/test_connection_options.py
```diff
@@ -0,0 +1,48 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+
+class TestConnectionOptionsOverride:
```
bobbyiliev, line 20: One more thing, would it be worth adding a second test case (or just another assertion) that checks the defaults still apply when the user doesn't pass an `options` block? Right now I think we'd catch a regression on the override path, but if someone refactored and accidentally dropped the `dict(DEFAULT_SESSION_PARAMETERS)` line we might not notice. Probably just asserting `welcome_message=off`, `auto_route_catalog_queries=on`, and `current_object_missing_warnings=off` in a profile with no `options` would do it. Up to you though, happy either way.

misc/dbt-materialize/dbt/adapters/materialize/connections.py
SangJunBak: > I think we'd end up sending --welcome_message=hello world and the server would parse world as a stray positional. 

Ah so I tested this out by setting the default option for cluster as `weewoo quickstart` and it worked in that case, even with other option overrides. This might be because Materialize is able to handle this without escapes but not postgres? However I think if we get the correct/same behavior **with** the escaping, then it makes sense for us to escape as well to maintain compatibility. I'll check!

bobbyiliev (approved):
I think that this is in a good shape now! Thanks for making all of the changes @SangJunBak 🎉

SangJunBak:
Thanks for the thorough review!

PR #36132, opened 2026-04-17
catalog: Better escaping of names in SQL

def-:
Follow-up to https://github.com/MaterializeInc/materialize/pull/35819


teskje (commented):
Hm, I did consider this but then thought it unnecessary because we control these names and I don't foresee us ever putting special characters in them. Am I missing something?

def-:
> I don't foresee us ever putting special characters in them

Who knows! I'd rather have a layer extra instead of depending on that without having it forbidden technically.

teskje:
Fair enough!

teskje (approved):

PR #36138, opened 2026-04-17
[COPY FROM pq] Dedupe pq map keys, choosing final one

patrickwwbutler:
https://github.com/MaterializeInc/database-issues/issues/11290#issue-4251375732

[Parquet docs state](https://parquet.apache.org/docs/file-format/types/logicaltypes/) that if there are duplicate keys in a map, then the final one is the actual value. This adds logic to dedupe entries with duplicate keys by skipping entries when the next one has an identical key, using the fact that the entries are sorted by this point to guarantee that identical keys will be adjacent. 

`sorted_by_key` is a stable sort function, meaning that entries with identical keys will retain their original ordering, allowing us to dedupe like this.

Updates `test/testdrive/copy-from-parquet-unsorted-map-keys.td` to test this functionality.


def- (approved):
Thanks for adding the tests!

PR #36139, opened 2026-04-17
mcp: fix origin validation to use CORS allowlist instead of Origin/Host self-comparison

bobbyiliev:
Fixes https://github.com/MaterializeInc/database-issues/issues/11311

The previous `validate_origin` function compared the request's Origin against its Host header, which doesn't prevent DNS rebinding.

Replacing the manual check with the existing `allowed_origin` CORS allowlist (built from `localhost`/`127.0.0.1`/`[::1]` defaults or `--cors-allowed-origin` flags), applied via `CorsLayer` on the MCP router. This matches the pattern used by the base router.

### DIFF
--- a/src/environmentd/src/http/mcp.rs
+++ b/src/environmentd/src/http/mcp.rs
@@ -23,12 +23,14 @@
 //!
 //! Data products are discovered via `mz_internal.mz_mcp_data_products` system view.
 
+use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::anyhow;
+use axum::Extension;
 use axum::Json;
 use axum::response::IntoResponse;
-use http::{HeaderMap, StatusCode};
+use http::{HeaderMap, HeaderValue, StatusCode};
 use mz_adapter_types::dyncfgs::{
     ENABLE_MCP_AGENT, ENABLE_MCP_AGENT_QUERY_TOOL, ENABLE_MCP_DEVELOPER, MCP_MAX_RESPONSE_SIZE,
 };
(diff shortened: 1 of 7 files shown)


src/environmentd/src/http/mcp.rs
```diff
@@ -377,66 +377,24 @@ pub async fn handle_mcp_method_not_allowed() -> impl IntoResponse {
...
-    }
     handle_mcp_request(client, body, McpEndpointType::Developer)
         .await
         .into_response()
 }
 
-/// Validates the Origin header to prevent DNS rebinding attacks (MCP 2025-11-25).
-/// Returns Some(403) if the Origin is present but doesn't match the Host.
-/// Returns None if the Origin is absent (non-browser client) or valid.
-fn validate_origin(headers: &HeaderMap) -> Option<axum::response::Response> {
```
def-, line 409: I think we should keep a kind of `validate_origin`:
```rust
/// Validates the Origin header against the CORS allowlist to prevent DNS
/// rebinding attacks (MCP spec 2025-11-25). Returns Some(403) if Origin is
/// present but not on the allowlist. Returns None if absent (non-browser client)
/// or allowed.
fn validate_origin(
    headers: &HeaderMap,
    allowed: &[HeaderValue],
) -> Option<axum::response::Response> {
    let origin = headers.get(http::header::ORIGIN)?;
    if mz_http_util::origin_is_allowed(origin, allowed) {
        return None;
    }
    warn!(
        origin = ?origin,
        "MCP request rejected: origin not in allowlist",
    );
    Some(StatusCode::FORBIDDEN.into_response())
}
```
CorsLayer only controls response headers, but doesn't block non-preflight POST requests server-side. In a DNS rebinding attack the browser considers the request same-origin, so no preflight occurs and CORS headers are irrelevant; without server-side Origin validation, the MCP handler executes and the attacker reads the response.

def- (changes requested):

src/environmentd/src/http/mcp.rs
bobbyiliev: Hm yes good point, you're right that CorsLayer alone isn't enough. Added `validate_origin` back along the lines you suggested, reusing the CORS allowlist.

def- (approved):

PR #36144, opened 2026-04-18
sql: Also eliminate casts in `applied_to`

def-:
Follow-up to https://github.com/MaterializeInc/materialize/pull/35324

Nightly run for verification: https://buildkite.com/materialize/nightly/builds/16133


ggevay (approved):
Wow, thank you! Lots of new eliminations in slts!

mgree:
Wow, good catch/fix! (And on the bug found, as well 😓.)

PR #36151, opened 2026-04-20
sql: int2vector to array cast is not eliminable

def-:
Fixes: https://github.com/MaterializeInc/database-issues/issues/11318

Follow-up to #36144. Unlucky that Nightly didn't catch it when I ran it on that PR, but then on main SQLancer++ failed in 2/2 runs!

Nightly run: https://buildkite.com/materialize/nightly/builds/16137

### DIFF
--- a/src/expr/src/scalar/func/impls/int2vector.rs
+++ b/src/expr/src/scalar/func/impls/int2vector.rs
@@ -18,7 +18,6 @@ use crate::scalar::func::stringify_datum;
     sqlname = "int2vectortoarray",
     is_monotone = true,
     introduces_nulls = false,
-    is_eliminable_cast = true,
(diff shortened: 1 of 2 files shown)


ggevay (approved):
Oops, yes, these are different types even at the repr type level: https://github.com/MaterializeInc/materialize/blob/5f785f23fd696c6e7f6069c0a7c7f7e8a408c2bb/src/repr/src/scalar.rs#L4726

(@mgree, could you check this too? Maybe you remember what was the difference that made it necessary to make these different types at the repr type level?)

mgree:
The difference was in treatment of `NULL`s.

ggevay:
Ok, I've made a small PR to record the difference in code comments: https://github.com/MaterializeInc/materialize/pull/36371

PR #36155, opened 2026-04-20
dbt: retry deploy_promote atomic swap only on concurrent-DDL conflicts

bobbyiliev:
Supersedes #33275 (closing it). Fixes customer issue CS-382 where `ALTER SWAP` in `dbt deploy` fails with DDL interrupt errors if there is concurrent DDL on the catalog.

Fixes: https://linear.app/materializeinc/issue/CS-382/ddl-interrupt-errors-during-alter-swap-in-dbt-deploy

The previous approach tried to add retries in Jinja using `{% try %}/{% except %}`, but Jinja2 does not support that syntax and it would also have retried unrelated errors, potentially hiding real failures.

This PR moves the retry logic to Python by adding `MaterializeAdapter.try_atomic_swap(sql)`, which classifies psycopg2 errors by SQLSTATE. It only retries on `40001` (DDL transaction race) and raises everything else immediately. The `deploy_promote.sql` macro now loops over this with a configurable `max_retries` (default 3), while keeping the existing log messages. Integration tests cover immediate failures for syntax and missing objects, as well as successful deploys under concurrent DDL. A CHANGELOG entry is included.

### DIFF
--- a/misc/dbt-materialize/CHANGELOG.md
+++ b/misc/dbt-materialize/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+* Automatically retry the atomic swap transaction in `deploy_promote` when
+  it is aborted by a concurrent-DDL conflict (Materialize SQLSTATE 40001).
+  Retries are capped by the new `max_retries` argument (default: 3) and
+  each retry waits `retry_backoff` seconds (default: 1.0). Errors that are
+  not concurrent-DDL conflicts, for example, permission, syntax, or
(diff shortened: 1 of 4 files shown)


def- (approved):
Not an expert in all of this, but looks reasonable!

jubrad:
@claude review

jubrad (approved):
I ran a claude review.. it caught some things, but nothing I'm really worried about. Take them or leave, I think the PR is 👍

PR #36160, opened 2026-04-20
clippy: Remove disallowed TimelyStack

def-:
Doesn't exist anymore, causes a warning in Clippy, which can't be promoted to an error unfortunately.


ggevay (approved):
Thank you!

antiguru (approved):
Thank you!

PR #36161, opened 2026-04-20
[CS-16, SS-66] kafka: add MATCHING broker rules for PrivateLink AZ mapping

jubrad:
## Summary

- Adds `BOOTSTRAP BROKER 'addr' USING AWS PRIVATELINK conn (...)` — new top-level option that provides the initial bootstrap address with an explicit PrivateLink tunnel, preserving the real hostname for correct TLS SNI
- Adds `MATCHING 'pattern' USING AWS PRIVATELINK conn (...)` inside `BROKERS (...)` — pattern-based routing rules for dynamically discovered brokers returned in Kafka metadata
- Replaces the `AWS PRIVATELINKS` syntax which used a `TO` keyword inconsistent with the existing `USING AWS PRIVATELINK` syntax and overloaded exact-match patterns as implicit bootstrap addresses
- Updates parser, planner, and storage layer to support the new constructs

### Example
```sql
CREATE CONNECTION kafka TO KAFKA (
    BROKERS (
        'lkc-825730.endpoint.cloud:9092' USING AWS PRIVATELINK pl_conn,
        MATCHING '*use1-az1*' USING AWS PRIVATELINK pl_conn (AVAILABILITY ZONE 'use1-az1'),
        MATCHING '*use1-az4*' USING AWS PRIVATELINK pl_conn (AVAILABILITY ZONE 'use1-az4'),
        MATCHING '*use1-az6*' USING AWS PRIVATELINK pl_conn (AVAILABILITY ZONE 'use1-az6')
    ),
    SASL MECHANISMS 'PLAIN',
    SASL USERNAME 'key',
    SASL PASSWORD SECRET secret,
    SECURITY PROTOCOL 'SASL_SSL'
);
```

## Test plan
- [x] Parser tests updated (roundtrip, error cases)
- [x] Testdrive updated (connection-create-drop, connection-alter)
- [x] `cargo check` passes
- [x] Cloudtest validation ([see follow-up PR](https://github.com/MaterializeInc/materialize/pull/36162))
- [ ] Manual testing against Confluent Cloud PrivateLink

## Docs
- [ ] https://github.com/MaterializeInc/materialize/pull/36163

### DIFF
--- a/src/kafka-util/src/client.rs
+++ b/src/kafka-util/src/client.rs
@@ -288,6 +288,13 @@ pub struct BrokerAddr {
     pub port: u16,
 }
 
+impl BrokerAddr {
+    /// Attempt to resolve this broker address into a list of socket addresses.
+    pub fn to_socket_addrs(&self) -> Result<Vec<SocketAddr>, io::Error> {
+        Ok((self.host.as_str(), self.port).to_socket_addrs()?.collect())
+    }
+}
+
 /// Rewrites a broker address.
 ///
 /// For use with [`TunnelingClientContext`].
@@ -299,6 +306,16 @@ pub struct BrokerRewrite {
     ///
     /// If unspecified, the broker's original port is left unchanged.
     pub port: Option<u16>,
+}
+
+impl BrokerRewrite {
+    /// Apply the rewrite to this broker address.
+    pub fn rewrite(&self, address: &BrokerAddr) -> BrokerAddr {
+        BrokerAddr {
+            host: self.host.clone(),
+            port: self.port.unwrap_or(address.port),
+        }
+    }
 }
 
 #[derive(Clone)]
@@ -313,6 +330,78 @@ enum BrokerRewriteHandle {
     FailedDefaultSshTunnel(String),
 }
 
+#[derive(Clone)]
+/// Parsed from a string, with optional leading and trailing '*' wildcards.
+pub struct ConnectionRulePattern {
+    /// If true, allow any combination of characters before the literal match.
+    pub prefix_wildcard: bool,
+    /// We expect the broker's host:port to match these characters in their entirety.
+    pub literal_match: String,
+    /// If true, allow any combination of characters after the literal match.
+    pub suffix_wildcard: bool,
+}
+
+impl std::fmt::Display for ConnectionRulePattern {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.prefix_wildcard {
+            f.write_str("*")?;
+        }
+        f.write_str(&self.literal_match)?;
+        if self.suffix_wildcard {
+            f.write_str("*")?;
+        }
+        Ok(())
+    }
+}
+
+impl ConnectionRulePattern {
+    /// Does this "{host}:{port}" address fit the pattern?
+    pub fn matches(&self, address: &str) -> bool {
+        if self.prefix_wildcard {
+            if self.suffix_wildcard {
+                address.contains(&self.literal_match)
+            } else {
+                address.ends_with(&self.literal_match)
+            }
+        } else if self.suffix_wildcard {
+            address.starts_with(&self.literal_match)
+        } else {
+            address == self.literal_match
+        }
+    }
+}
+
+#[derive(Clone)]
+/// Given a host address, map it to a different host.
+pub struct HostMappingRules {
+    /// Map matching hosts to a different host. First applicable rule wins.
+    pub rules: Vec<(ConnectionRulePattern, BrokerRewrite)>,
+}
+
+impl HostMappingRules {
+    /// Rewrite this broker address according to the rules. Returns `None` when
+    /// no rule matches.
+    pub fn rewrite(&self, src: &BrokerAddr) -> Option<BrokerAddr> {
+        let address = format!("{}:{}", src.host, src.port);
+        for (pattern, dst) in &self.rules {
+            if pattern.matches(&address) {
+                let result = dst.rewrite(src);
+                info!(
+                    "HostMappingRules: broker {}:{} matched pattern '{}' -> rewriting to {}:{}",
+                    src.host, src.port, pattern, result.host, result.port,
+                );
+                return Some(result);
+            }
+        }
+
+        warn!(
+            "HostMappingRules: broker {}:{} matched no rules, using original address",
+            src.host, src.port,
+        );
+        None
+    }
+}
+
 /// Tunneling clients
 /// used for re-writing ports / hosts
 #[derive(Clone)]
@@ -321,6 +410,8 @@ pub enum TunnelConfig {
     Ssh(SshTunnelConfig),
     /// Re-writes internal hosts using the value, used for privatelink
     StaticHost(String),
+    /// Re-writes internal hosts according to an ordered list of rules, also used for privatelink
+    Rules(HostMappingRules),
     /// Performs no re-writes
     None,
 }
@@ -489,7 +580,12 @@ where
         }
     }
 
+    /// Look up the broker's address in our book of rewrites.
+    /// If we've already rewritten it before, reuse the existing rewrite.
+    /// Otherwise, use our "default tunnel" rewriting strategy to attempt to rewrite this broker's address
+    /// and record it in the book of rewrites.
     fn resolve_broker_addr(&self, host: &str, port: u16) -> Result<Vec<SocketAddr>, io::Error> {
+        info!("kafka: resolve_broker_addr called for {}:{}", host, port);
         let return_rewrite = |rewrite: &BrokerRewriteHandle| -> Result<Vec<SocketAddr>, io::Error> {
             let rewrite = match rewrite {
                 BrokerRewriteHandle::Simple(rewrite) => rewrite.clone(),
@@ -525,8 +621,12 @@ where
         let rewrite = self.rewrites.lock().expect("poisoned").get(&addr).cloned();
 
         match rewrite {
+            // No (successful) broker address rewrite exists yet.
             None | Some(BrokerRewriteHandle::FailedDefaultSshTunnel(_)) => {
+                // "Default tunnel" is actually the configured rewriting strategy used for brokers we haven't already rewritten.
                 match &self.default_tunnel {
+                    // This "default tunnel" is actually a default tunnel.
+                    // Try connecting so we have a valid rewrite for thsi broker address.
                     TunnelConfig::Ssh(default_tunnel) => {
                         // Multiple users could all run `connect` at the same time; only one ssh
                         // tunnel will ever be connected, and only one will be inserted into the
@@ -543,6 +643,7 @@ where
                                 .await
                         });
                         match ssh_tunnel {
+                            // Use the tunnel we just created, but only if nobody beat us in the race.
                             Ok(ssh_tunnel) => {
                                 let mut rewrites = self.rewrites.lock().expect("poisoned");
                                 let rewrite = match rewrites.entry(addr.clone()) {
@@ -565,6 +666,7 @@ where
 
                                 return_rewrite(rewrite)
                             }
+                            // We couldn't connect. Someone else will have to try again.
                             Err(e) => {
                                 warn!(
                                     "failed to create ssh tunnel for {:?}: {}",
@@ -587,15 +689,45 @@ where
                             }
                         }
                     }
+                    // Our rewrite strategy is to use a specific host, e.g. a PrivateLink endpoint.
                     TunnelConfig::StaticHost(host) => (host.as_str(), port)
                         .to_socket_addrs()
                         .map(|addrs| addrs.collect()),
+                    // Rewrite according to the routing rules.
+                    TunnelConfig::Rules(rules) => {
+                        // If no rules match, just use the address as-is.
+                        let resolved = rules.rewrite(&addr).unwrap_or_else(|| addr.clone());
+                        match resolved.to_socket_addrs() {
+                            Ok(addrs) => {
+                                info!(
+                                    "kafka: resolve_broker_addr {}:{} -> {}:{} resolved to {:?}",
+                                    host, port, resolved.host, resolved.port, addrs,
+                                );
+                                Ok(addrs)
+                            }
+                            Err(e) => {
+                                warn!(
+                                    "kafka: resolve_broker_addr {}:{} -> {}:{} DNS resolution FAILED: {e}",
+                                    host, port, resolved.host, resolved.port,
+                                );
+                                Err(e)
+                            }
+                        }
+                    }
+                    // We leave the broker's address as it is.
                     TunnelConfig::None => {
                         (host, port).to_socket_addrs().map(|addrs| addrs.collect())
                     }
                 }
             }
-            Some(rewrite) => return_rewrite(&rewrite),
+            // This broker's address was already rewritten. Reuse the existing rewrite.
+            Some(rewrite) => {
+                info!(
+                    "kafka: resolve_broker_addr {}:{} using cached rewrite",
+                    host, port
+                );
+                return_rewrite(&rewrite)
+            }
         }
     }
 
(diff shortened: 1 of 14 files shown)


jubrad:
**Note on `build_tunnel_definition` removal from `StatementContext`:**

On main, `build_tunnel_definition` is a method on `StatementContext` taking 2 params (`ssh_tunnel`, `aws_privatelink`). All 5 callers (Kafka, CSR, Postgres, MySQL, SQL Server) call it via `scx.build_tunnel_definition(...)`.

This PR moves it to a standalone function in `connection.rs` with a 3rd param (`matching_rules`) to support the Kafka MATCHING rules path. The non-Kafka callers pass `None` for matching rules. Since all callers now use the standalone function, the original method on `StatementContext` has zero callers and is removed.

This is intentional — one function handles all connection types rather than maintaining two parallel implementations.

src/kafka-util/src/client.rs
```diff
@@ -313,6 +330,78 @@ enum BrokerRewriteHandle {
...
 
+#[derive(Clone)]
+/// Parsed from a string, with optional leading and trailing '*' wildcards.
+pub struct ConnectionRulePattern {
+    /// If true, allow any combination of characters before the literal match.
+    pub prefix_wildcard: bool,
+    /// We expect the broker's host:port to match these characters in their entirety.
+    pub literal_match: String,
+    /// If true, allow any combination of characters after the literal match.
+    pub suffix_wildcard: bool,
```
jubrad, line 341: I'm wondering if we actually need a suffix, if this feature is restricted to confluent, confluent provides a suffix, so we only need prefix wildcard matching. I don't think this hurts, but it's not strictly necessary.

test/testdrive/connection-create-drop.td
```diff
@@ -387,14 +387,14 @@ contains:BROKER specified more than once
 ! CREATE CONNECTION no_broker TO KAFKA (
     SECURITY PROTOCOL PLAINTEXT
   );
-contains:must set one of BROKER, BROKERS, or AWS PRIVATELINK
+contains:must set one of BROKER, BROKERS, BOOTSTRAP BROKER, or AWS PRIVATELINK
```
jubrad, line 390: I think this makes the syntax for this feature quite clear, but I'm very glad we got rid of the railroad track diagram because the syntax is getting wild.
-  BROKER implies a single broker, it's exclusive with BROKERS, but both set bootstrap broker
- AWS PRIVATELINK only works when a single top level privatelink is provided with no individual broker matching, although perhaps one could provide BROKER
- BROKERS is the standard way for mapping brokers to aws privatelink for everything but red panda


I really don't know the best way to land things here.
1. keep things as they are in this PR
2. allow broker and brokers where broker is bootstrap and optional, but we error if not static broker is provided
3. we get rid of bootstrap brokers and require at least one static broker if the user specifies BROKERS

ublubu: What's the "railroad track diagram"?

_Re: We need at least one static broker to bootstrap from._

> get rid of BOOTSTRAP BROKER and require at least one static broker in BROKERS

I like this one. It feels clearer than "if you only have MATCHING in BROKERS, you also need a BOOTSTRAP BROKER".

Then BROKER is subsumed by BROKERS and we can just about replace AWS PRIVATELINK with a static broker (static address = PrivateLink address) and a `MATCHING '*'` rule.
Not saying we remove the other ones, but at least we can focus on just BROKERS in docs, etc.

jubrad: > What's the "railroad track diagram"?

We used to generate diagrams for all SQL using BNF that we had to maintain along all SQL changes these diagrams would outline the SQL sytnax which, in this case would be super complicated. 


>> get rid of BOOTSTRAP BROKER and require at least one static broker in BROKERS
> I like this one

Cool, I'll make this change

jubrad:
closes
https://github.com/MaterializeInc/database-issues/issues/8835

def- (commented):
```
mode cockroach

simple conn=mz_system,user=mz_system
ALTER SYSTEM SET max_aws_privatelink_connections = 10;
----
COMPLETE 0

statement ok
CREATE CONNECTION pl TO AWS PRIVATELINK (
    SERVICE NAME 'com.amazonaws.vpce.us-east-1.vpce-svc-0e123abc123198abc',
    AVAILABILITY ZONES ('use1-az1', 'use1-az4')
);

statement error cannot specify BROKER, BOOTSTRAP BROKER, or static BROKERS entries alongside top-level AWS PRIVATELINK
CREATE CONNECTION kafka_bad TO KAFKA (
    BROKER 'kafka:9092',
    AWS PRIVATELINK pl (PORT 9092),
    SECURITY PROTOCOL PLAINTEXT
);
```
Fails with a panic:
```
2026-04-22T02:23:43.120459Z ERROR mz_adapter::coord::sequencer::inner: connection validation panicked
```

Also this one:
```
mode cockroach

simple conn=mz_system,user=mz_system
ALTER SYSTEM SET max_aws_privatelink_connections = 10;
----
COMPLETE 0

simple conn=mz_system,user=mz_system
ALTER SYSTEM SET enable_kafka_broker_matching_rules = true;
----
COMPLETE 0

statement ok
CREATE CONNECTION pl TO AWS PRIVATELINK (
    SERVICE NAME 'com.amazonaws.vpce.us-east-1.vpce-svc-0e123abc123198abc',
    AVAILABILITY ZONES ('use1-az1', 'use1-az4')
);

statement error BOOTSTRAP BROKER is required when using MATCHING broker rules
CREATE CONNECTION kafka_matching_no_bootstrap TO KAFKA (
    BROKERS (
        MATCHING '*.use1-az1.*' USING AWS PRIVATELINK pl (AVAILABILITY ZONE = 'use1-az1'),
        MATCHING '*.use1-az4.*' USING AWS PRIVATELINK pl (AVAILABILITY ZONE = 'use1-az4')
    ),
    SECURITY PROTOCOL PLAINTEXT
) WITH (VALIDATE = FALSE);
```
Fails too, we should reject `CREATE CONNECTION` without a `BOOTSTRAP BROKER`:
```
    CREATE CONNECTION kafka_matching_no_bootstrap TO KAFKA (
        BROKERS (
            MATCHING '*.use1-az1.*' USING AWS PRIVATELINK pl (AVAILABILITY ZONE = 'use1-az1'),
            MATCHING '*.use1-az4.*' USING AWS PRIVATELINK pl (AVAILABILITY ZONE = 'use1-az4')
        ),
        SECURITY PROTOCOL PLAINTEXT
    ) WITH (VALIDATE = FALSE);
    UnexpectedPlanSuccess:test/sqllogictest/kafka_matching_requires_bootstrap_broker.
```

src/kafka-util/src/client.rs
```diff
@@ -525,8 +621,12 @@ where
         let rewrite = self.rewrites.lock().expect("poisoned").get(&addr).cloned();
 
         match rewrite {
+            // No (successful) broker address rewrite exists yet.
             None | Some(BrokerRewriteHandle::FailedDefaultSshTunnel(_)) => {
+                // "Default tunnel" is actually the configured rewriting strategy used for brokers we haven't already rewritten.
                 match &self.default_tunnel {
+                    // This "default tunnel" is actually a default tunnel.
+                    // Try connecting so we have a valid rewrite for thsi broker address.
```
ublubu, line 629:
```suggestion
                    // Try connecting so we have a valid rewrite for this broker address.
```

Fixing my own typo.

src/sql-parser/src/ast/defs/statement.rs
```diff
@@ -603,6 +603,75 @@ impl<T: AstInfo> AstDisplay for ConnectionDefaultAwsPrivatelink<T> {
...
+    /// If true, allow any combination of characters after the literal match.
+    pub suffix_wildcard: bool,
+}
+
+impl ConnectionRulePattern {
+    /// True when the pattern has no wildcards — it matches a single exact broker address.
+    pub fn is_exact(&self) -> bool {
+        !self.prefix_wildcard && !self.suffix_wildcard
+    }
+}
```
ublubu: I don't think this is needed anymore, now that we don't use pattern rules (MATCHING) for the bootstrap broker.

src/sql/src/plan/statement/ddl/connection.rs
```diff
@@ -315,11 +314,20 @@ impl ConnectionOptionExtracted {
...
                 ConnectionDetails::Kafka(KafkaConnection {
-                    brokers: self.get_brokers(scx)?,
-                    default_tunnel: scx
-                        .build_tunnel_definition(self.ssh_tunnel, self.aws_privatelink)?,
+                    brokers: static_brokers,
+                    default_tunnel: build_tunnel_definition(
+                            scx,
+                            self.ssh_tunnel,
+                            self.aws_privatelink,
+                            if matching_rules.is_empty() { None } else { Some(matching_rules) },
```
ublubu, line 329: Interesting.

src/sql/src/plan/statement/ddl/connection.rs
```diff
@@ -673,76 +701,75 @@ impl ConnectionOptionExtracted {
...
+        }
+        if let Some(brokers_list) = &self.brokers {
+            all_brokers.extend(brokers_list.static_entries.iter().cloned());
+            matching_rules.extend(brokers_list.matching_rules.iter().cloned());
+        }
+
+        // MATCHING rules require at least one static broker for bootstrapping.
+        if !matching_rules.is_empty() && all_brokers.is_empty() {
+            sql_bail!(
+                "invalid CONNECTION: BROKERS must contain at least one static broker address when using MATCHING rules"
```
ublubu: I think this simpler phrasing is more true:
```suggestion
                "invalid CONNECTION: BROKERS must contain at least one static broker address"
```

And the (non-obvious) reason we still need the `!matching_rules.is_empty()` is that `matching_rules` and `all_brokers` will both be empty if `aws_privatelink` is present.

src/sql/src/plan/statement/ddl/connection.rs
```diff
@@ -673,76 +701,75 @@ impl ConnectionOptionExtracted {
...
+        if self.broker.is_some()
+            && self
+                .brokers
+                .as_ref()
+                .map_or(false, |b| !b.static_entries.is_empty())
+        {
+            sql_bail!(
+                "invalid CONNECTION: cannot specify both BROKER and static entries in BROKERS"
+            );
+        }
```
ublubu: Maybe this "one of" check should come first, then the "must have one static address" condition can be less fiddly.

src/sql/src/plan/statement/ddl/connection.rs
```diff
@@ -673,76 +701,75 @@ impl ConnectionOptionExtracted {
...
+
+        // BROKER and BROKERS (static entries) are still mutually exclusive.
+        if self.broker.is_some()
+            && self
+                .brokers
+                .as_ref()
+                .map_or(false, |b| !b.static_entries.is_empty())
+        {
+            sql_bail!(
+                "invalid CONNECTION: cannot specify both BROKER and static entries in BROKERS"
```
ublubu: Ok, now I see why this area is more complex than I expected. It's accounting for:

```
BROKER (static address),
BROKERS (MATCHING pattern USING connection)
```

I think we can call that invalid, and just support:
```
BROKERS (static address, MATCHING pattern USING connection)
```

src/storage-types/src/connections.rs
```diff
@@ -957,15 +959,30 @@ impl KafkaConnection {
...
                 )
             }
+            Tunnel::AwsPrivatelinks(_pl) => {
+                let algo = KAFKA_DEFAULT_AWS_PRIVATELINK_ENDPOINT_IDENTIFICATION_ALGORITHM
+                    .get(storage_configuration.config_set());
+                options.insert("ssl.endpoint.identification.algorithm".into(), algo.into());
+
+                if self.brokers.is_empty() {
+                    return Err(ContextCreationError::Other(anyhow::anyhow!(
+                        "at least one static broker is required when using MATCHING broker rules"
```
ublubu: Maybe it should be "at least one static broker is required when using BROKER or BROKERS"

because the only time we get away without a static broker is Redpanda via PrivateLink **(??)**, where we just use the PrivateLink endpoint directly (because no host verification on Redpanda).

src/storage-types/src/connections.rs
```diff
@@ -1067,10 +1084,17 @@ impl KafkaConnection {
...
-                    None, // Default tunnel does not support availability zones.
-                )));
+                context.set_default_tunnel(TunnelConfig::StaticHost(
+                    // Possible bug: We have been ignoring the configured port.
+                    KafkaConnection::from_default_aws_privatelink(pl).host,
+                ));
+            }
+            Tunnel::AwsPrivatelinks(pl) => {
+                // All rules (both exact-match and wildcard) go through dynamic routing.
+                // Exact-match rules also provide bootstrap broker addresses (set above).
```
ublubu:
```suggestion
```

This comment no longer holds because the bootstrap broker does not come from a MATCHING clause.

def- (commented):
Minor thing, we should reject invalid `MATCHING` strings:
```diff
diff --git a/src/sql-parser/tests/testdata/ddl b/src/sql-parser/tests/testdata/ddl
index ea60378216..265a3608b2 100644
--- a/src/sql-parser/tests/testdata/ddl
+++ b/src/sql-parser/tests/testdata/ddl
@@ -671,6 +671,37 @@ error: Expected USING, found right parenthesis
 CREATE CONNECTION kafka_connection TO KAFKA (BROKERS (MATCHING '*.az.*'))
                                                                        ^

+# Regression: MATCHING patterns must reject internal `*` wildcards. Only a
+# single leading and/or trailing `*` is supported; anything else silently
+# produces a rule that can never match a real broker hostname.
+parse-statement
+CREATE CONNECTION kafka_connection TO KAFKA (BROKERS (MATCHING '*foo*bar*' USING AWS PRIVATELINK privatelink_svc (AVAILABILITY ZONE 'use1-az1')))
+----
+error: pattern may only contain `*` as a leading and/or trailing wildcard
+CREATE CONNECTION kafka_connection TO KAFKA (BROKERS (MATCHING '*foo*bar*' USING AWS PRIVATELINK privatelink_svc (AVAILABILITY ZONE 'use1-az1')))
+                                                                        ^
+
+parse-statement
+CREATE CONNECTION kafka_connection TO KAFKA (BROKERS (MATCHING 'foo*bar' USING AWS PRIVATELINK privatelink_svc (AVAILABILITY ZONE 'use1-az1')))
+----
+error: pattern may only contain `*` as a leading and/or trailing wildcard
+CREATE CONNECTION kafka_connection TO KAFKA (BROKERS (MATCHING 'foo*bar' USING AWS PRIVATELINK privatelink_svc (AVAILABILITY ZONE 'use1-az1')))
+                                                                      ^
+
+parse-statement
+CREATE CONNECTION kafka_connection TO KAFKA (BROKERS (MATCHING '**az1**' USING AWS PRIVATELINK privatelink_svc (AVAILABILITY ZONE 'use1-az1')))
+----
+error: pattern may only contain `*` as a leading and/or trailing wildcard
+CREATE CONNECTION kafka_connection TO KAFKA (BROKERS (MATCHING '**az1**' USING AWS PRIVATELINK privatelink_svc (AVAILABILITY ZONE 'use1-az1')))
+                                                                      ^
+
+parse-statement
+CREATE CONNECTION kafka_connection TO KAFKA (BROKERS (MATCHING 'broker-*-az1-*.host.com' USING AWS PRIVATELINK privatelink_svc (AVAILABILITY ZONE 'use1-az1')))
+----
+error: pattern may only contain `*` as a leading and/or trailing wildcard
+CREATE CONNECTION kafka_connection TO KAFKA (BROKERS (MATCHING 'broker-*-az1-*.host.com' USING AWS PRIVATELINK privatelink_svc (AVAILABILITY ZONE 'use1-az1')))
+                                                                                      ^
+
 parse-statement
 CREATE SOURCE mz_source FROM MYSQL CONNECTION mysqlconn FOR TABLES (foo, bar as qux, baz into zop);
 ----
```
Runs with `cargo test -p mz-sql-parser --test sqlparser_common datadriven`

src/sql/src/plan/statement/ddl/connection.rs
```diff
@@ -673,76 +701,64 @@ impl ConnectionOptionExtracted {
...
+            // BROKERS
+            (None, Some(broker_list), None) => {
+                all_brokers.extend(broker_list.static_entries.iter().cloned());
+                matching_rules.extend(broker_list.matching_rules.iter().cloned());
+            }
+            // AWS PRIVATELINK
+            (None, None, Some(_privatelink)) => {
+                // Noting to do here
+            }
+            // noting - invalid
```
ublubu, line 731:
```suggestion
            // nothing - invalid
```

ublubu (approved):
looking good 🤠

PR #36165, opened 2026-04-20
sinks: walk the input arrangement via cursor per batch

DAlperin:
## Summary

Refactor the sink rendering path so sinks walk their input arrangement's cursors directly, eliminating the `Vec<DiffPair<Row>>`-per-group materialization and the trace-reader overhead that `zip_into_diff_pairs` → `combine_at_timestamp` → flat_map previously incurred. Motivated by profiles of very large snapshot sinks where that pipeline dominated allocation and caused major page faults.

## What this buys

- **No per-`(key, time)` `Vec<DiffPair<Row>>` allocation.** Biggest direct win — this was the dominant allocator in `zip_into_diff_pairs` on the profile.
- **Key owned once per key, not per `(key, time)` group.**
- **Fewer operator boundaries.** Kafka: arrangement → encode (was: arrangement → `combine_at_timestamp` → flat_map → encode). Iceberg: one fewer hop too.
- **Spine compacts aggressively.** Dropping the `TraceAgent` lets spine compaction advance its frontier, releasing historical batch state rather than accumulating it.

## What this does *not* fix

- The arrangement's **pre-spine batcher** still buffers un-sealed updates (unavoidable at this layer).
- For a pure-insertion snapshot with no retractions, spine compaction has limited work to do — savings come primarily from the `Vec`/operator overhead, not from the spine itself shrinking.
- Persist snapshot forwarding semantics and sink commit-on-frontier semantics are unchanged.

Follow-ups (not in this PR) worth considering:
- A batcher-only operator that skips the spine entirely for sinks that don't need a trace.
- An append-mode Iceberg fast-path that skips arrangement altogether when no key is configured.

## Test plan

- [x] `cargo check --workspace --all-targets`
- [x] `cargo clippy -p mz-interchange -p mz-storage --tests`
- [x] `bin/lint` (check-no-diff fails locally due to jj colocation; all substantive checks pass)
- [x] `cargo test -p mz-interchange --lib envelopes::` (7 tests, all pass)
- [ ] Full CI (pending)
- [ ] Kafka sink testdrive (snapshot + steady-state, Upsert and Debezium envelopes)
- [ ] Iceberg sink testdrive (snapshot + steady-state, Upsert and Append envelopes)
- [ ] Re-profile a very large snapshot sink to confirm the \`zip_into_diff_pairs\` hotspot is gone

### DIFF
--- a/src/interchange/src/envelopes.rs
+++ b/src/interchange/src/envelopes.rs
@@ -15,7 +15,7 @@ use differential_dataflow::operators::arrange::Arranged;
 use differential_dataflow::trace::implementations::BatchContainer;
 use differential_dataflow::trace::{BatchReader, Cursor, TraceReader};
 use differential_dataflow::{AsCollection, VecCollection};
-use itertools::{EitherOrBoth, Itertools};
+use itertools::{Either, EitherOrBoth, Itertools};
 use maplit::btreemap;
 use mz_ore::cast::CastFrom;
 use mz_repr::{
@@ -25,6 +25,86 @@ use timely::dataflow::channels::pact::Pipeline;
 use timely::dataflow::operators::Operator;
 
 use crate::avro::DiffPair;
+
+/// Walks `batch` and invokes `on_diff_pair` for each `DiffPair` at each
+/// `(key, timestamp)`.
+///
+/// Within a key, diffs are partitioned by sign into retractions (befores) and
+/// insertions (afters), sorted by timestamp, and zipped into `DiffPair`s via a
+/// merge-join. Pairs are emitted in ascending timestamp order for a given key;
+/// no ordering is guaranteed across keys.
+///
+/// This is the batch-level cursor walk underlying [`combine_at_timestamp`].
+/// Sinks that consume an arrangement directly can call this inside their own
+/// operator to stream pairs without materializing a `Vec<DiffPair>` per group.
+pub fn for_each_diff_pair<B, F>(batch: &B, mut on_diff_pair: F)
+where
+    B: BatchReader<Diff = Diff>,
+    B::Time: Copy,
+    B::ValOwn: 'static,
+    F: FnMut(&<B::KeyContainer as BatchContainer>::Owned, B::Time, DiffPair<B::ValOwn>),
+{
+    let mut befores: Vec<(B::Time, B::ValOwn, usize)> = vec![];
+    let mut afters: Vec<(B::Time, B::ValOwn, usize)> = vec![];
+
+    let mut cursor = batch.cursor();
+    while cursor.key_valid(batch) {
+        let k = cursor.key(batch);
+
+        // Partition updates at this key into retractions (befores) and
+        // insertions (afters).
+        while cursor.val_valid(batch) {
+            let v = cursor.val(batch);
+            cursor.map_times(batch, |t, diff| {
+                let diff = B::owned_diff(diff);
+                let update = (
+                    B::owned_time(t),
+                    B::owned_val(v),
+                    usize::cast_from(diff.unsigned_abs()),
+                );
+                if diff < Diff::ZERO {
+                    befores.push(update);
+                } else {
+                    afters.push(update);
+                }
+            });
+            cursor.step_val(batch);
+        }
+
+        befores.sort_by_key(|(t, _v, _diff)| *t);
+        afters.sort_by_key(|(t, _v, _diff)| *t);
+
+        // Fan `(time, val, count)` out to `count` copies of `(time, val)`.
+        // `iter::repeat(x).take(n)` clones on every `next()` — including the
+        // first — so using it with `count == 1` (the snapshot common case)
+        // does a gratuitous clone. `iter::once` moves the value; we only
+        // fall back to `iter::repeat` when fan-out is actually needed.
+        let fan_out = |(t, v, cnt): (B::Time, B::ValOwn, usize)| {
+            if cnt == 1 {
+                Either::Left(iter::once((t, v)))
+            } else {
+                Either::Right(iter::repeat((t, v)).take(cnt))
+            }
+        };
+        let befores_iter = befores.drain(..).flat_map(fan_out);
+        let afters_iter = afters.drain(..).flat_map(fan_out);
+
+        let key_owned = <B::KeyContainer as BatchContainer>::into_owned(k);
+
+        for pair in
+            itertools::merge_join_by(befores_iter, afters_iter, |(t1, _v1), (t2, _v2)| t1.cmp(t2))
+        {
+            let (t, before, after) = match pair {
+                EitherOrBoth::Both((t, before), (_t, after)) => (t, Some(before), Some(after)),
+                EitherOrBoth::Left((t, before)) => (t, Some(before), None),
+                EitherOrBoth::Right((t, after)) => (t, None, Some(after)),
+            };
+            on_diff_pair(&key_owned, t, DiffPair { before, after });
+        }
+
+        cursor.step_key(batch);
+    }
+}
 
 /// Given a stream of batches, produce a stream of groups of DiffPairs, grouped
 /// by key, at each timestamp.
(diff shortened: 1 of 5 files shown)


def- (commented):
When a Kafka or Iceberg sink is configured with a non-unique key (`KEY (...) NOT ENFORCED`), Materialize is supposed to log a warning ("primary key error") whenever multiple rows share the same key at the same timestamp. With this PR the warning doesn't appear when no new data arrives afterwards. Not sure how bad that is.

Reproducer:
```diff
diff --git a/test/source-sink-errors/mzcompose.py b/test/source-sink-errors/mzcompose.py
index ac2b7c3e92..5a9bbeb7ee 100644
--- a/test/source-sink-errors/mzcompose.py
+++ b/test/source-sink-errors/mzcompose.py
@@ -13,6 +13,7 @@ Disruption and then checking the mz_internal.mz_*_statuses tables
 """

 import random
+import time
 from collections.abc import Callable
 from dataclasses import dataclass
 from textwrap import dedent
@@ -561,3 +562,52 @@ def unsupported_pg_table(c: Composition) -> None:
                  $ postgres-execute connection=postgres://postgres:postgres@postgres
                  INSERT INTO source1 VALUES (3, '[2:3]={2,2}')
                  """))
+
+
+def workflow_test_pk_violation_warning(c: Composition) -> None:
+    seed = random.randint(0, 256**4)
+    c.up("redpanda", "materialized", Service("testdrive", idle=True))
+
+    with c.override(
+        Testdrive(
+            no_reset=True,
+            seed=seed,
+            entrypoint_extra=["--initial-backoff=1s", "--backoff-factor=0"],
+        )
+    ):
+        c.testdrive(dedent("""
+                > CREATE CONNECTION kafka_conn
+                  TO KAFKA (BROKER '${testdrive.kafka-addr}', SECURITY PROTOCOL PLAINTEXT);
+
+                > CREATE CONNECTION IF NOT EXISTS csr_conn TO CONFLUENT SCHEMA REGISTRY (
+                  URL '${testdrive.schema-registry-url}'
+                  );
+
+                > CREATE TABLE t (a INT, b INT);
+
+                > CREATE SINK pk_test_sink FROM t
+                  INTO KAFKA CONNECTION kafka_conn (TOPIC 'testdrive-pk-test-${testdrive.seed}')
+                  KEY (a) NOT ENFORCED
+                  FORMAT AVRO USING CONFLUENT SCHEMA REGISTRY CONNECTION csr_conn
+                  ENVELOPE UPSERT;
+
+                > SELECT status FROM mz_internal.mz_sink_statuses WHERE name = 'pk_test_sink'
+                running
+        """))
+
+        # PkViolationWarner rate-limits to once per 10s from construction.
+        time.sleep(12)
+
+        c.testdrive(dedent("""
+                > INSERT INTO t VALUES (1, 10), (1, 20);
+
+                $ set-sql-timeout duration=60s
+                $ kafka-verify-topic sink=materialize.public.pk_test_sink await-value-schema=true
+        """))
+
+        time.sleep(5)
+
+        logs = c.invoke("logs", "materialized", capture=True)
+        assert "primary key error" in logs.stdout, (
+            "PkViolationWarner::flush() not called after batch"
+        )
```

src/storage/src/render/sinks.rs
```diff
@@ -132,55 +145,78 @@ fn zip_into_diff_pairs<'scope>(
...
-    let mut collection =
-        combine_at_timestamp(collection.arrange_named::<OrdValBatcher<_,_,_,_>, RcOrdValBuilder<_,_,_,_>, OrdValSpine<_, _, _, _>>("Arrange Sink"));
-
-    // If there is no user-specified key, remove the synthetic key.
-    //
-    // We don't want the synthetic key to appear in the sink's actual output; we
-    // just needed a value to use to distribute work.
-    if user_key_indices.is_none() {
-        collection = collection.map(|(_key, value)| (None, value))
+    let arranged = keyed.arrange_named::<OrdValBatcher<_, _, _, _>, RcOrdValBuilder<_, _, _, _>, OrdValSpine<_, _, _, _>>("Arrange Sink");
```
martykulma: nit: more explicit RE: dropping the trace
```
let Arranged{ stream, trace: _ } = ...;
stream
```

src/storage/src/render/sinks.rs
```diff
@@ -132,55 +145,78 @@ fn zip_into_diff_pairs<'scope>(
...
-                let k = if idx == max_idx { k.take() } else { k.clone() };
-                (k, dp)
-            })
+    /// Flush the pending `(key, timestamp)` group count. Emits a
+    /// rate-limited warning if more than one `DiffPair` was observed.
+    pub fn flush(&mut self) {
+        if self.count > 1 {
+            let now = Instant::now();
+            if now.duration_since(self.last_warning) >= Duration::from_secs(10) {
+                self.last_warning = now;
```
martykulma, line 207: if the snapshot is quick (<10s), this won't emit a warning.  It looks like the previous one didn't either, so possibly not a big deal.

src/interchange/src/envelopes.rs
```diff
@@ -11,124 +11,93 @@ use std::collections::BTreeMap;
...
+        // `iter::repeat(x).take(n)` clones on every `next()` — including the
+        // first — so using it with `count == 1` (the snapshot common case)
+        // does a gratuitous clone. `iter::once` moves the value; we only
+        // fall back to `iter::repeat` when fan-out is actually needed.
+        let fan_out = |(t, v, cnt): (B::Time, B::ValOwn, usize)| {
+            if cnt == 1 {
+                Either::Left(iter::once((t, v)))
+            } else {
+                Either::Right(iter::repeat((t, v)).take(cnt))
             }
```
martykulma: should do the same
```
std::iter::repeat_n((t,v), cnt)
```

martykulma (approved):
looks great @DAlperin - minor comments, nothing blocking.

src/storage/src/render/sinks.rs
DAlperin: Yeah, stole from old implementation, not that worried about if for now

src/storage/src/render/sinks.rs
```diff
@@ -77,46 +91,45 @@ pub(crate) fn render_sink<'scope>(
...
+        .get_key_indices()
+        .or_else(|| sink_render.get_relation_key_indices())
         .map(|k| k.to_vec());
-    let key_is_synthetic = key_indices.is_none();
 
-    let collection = match key_indices {
+    let keyed = match key_indices {
         None => collection.map(|row| (Some(Row::pack(Some(Datum::UInt64(row.hashed())))), row)),
         Some(key_indices) => {
             let mut datum_vec = mz_repr::DatumVec::new();
```
patrickwwbutler, line 135: this is obviously unrelated to your changes, but im curious - why is `key` a whole `Row` of its own that needs to be packed? Do the sinks represent the key as a sort of struct column that needs to be built from the `Datum`s in each of the key indices? It feels like it would make more sense to me to merge all the key `Datum`s into one combined key `Datum` rather than its own row? I suppose we can't really precompute a key column because its entirely dependent on the sink, different sinks for the same persist collection could want different, arbitrary keys, but the `TODO[perf]` had me wondering

src/storage/src/render/sinks.rs
patrickwwbutler: curious if the trace would be optimized out by the compiler in this current version, or would this nit possibly provide some perf benefit?

agreed with the nit in terms of readability, though

antiguru: You should make sure to drop the trace, otherwise you risk never releasing arrangement memory. A handle to a trace needs to be downgraded to release resources.

DAlperin: I updated the code to make the drop clearer, but this is whats happening, no? The trace goes out of scope and is dropped. `TraceAgent::drop` then calls downgrade before the RC falls.

antiguru: (My comment was less about the implementation and rather what should be true :) I've had some surprises in the past when things were accidentally moved into closures and whatnot.)

src/storage/src/render/sinks.rs
```diff
@@ -77,46 +91,45 @@ pub(crate) fn render_sink<'scope>(
         );
         tokens.extend(persist_tokens);
 
-        let ok_collection =
-            zip_into_diff_pairs(sink_id, sink, &*sink_render, ok_collection.as_collection());
+        let batches = arrange_sink_input(&*sink_render, ok_collection.as_collection());
+        let key_is_synthetic = sink_render.get_key_indices().is_none()
```
ublubu, line 95: Maybe it's obvious what `key_is_synthetic` means if you already know about the _per-row hash used to distribute work in the absence of a natural key_,
but this line confused me until I read the rest of the comments in this file.

src/interchange/src/envelopes.rs
```diff
@@ -11,124 +11,87 @@ use std::collections::BTreeMap;
...
+
+    let mut cursor = batch.cursor();
+    while cursor.key_valid(batch) {
+        let k = cursor.key(batch);
+
+        // Partition updates at this key into retractions (befores) and
+        // insertions (afters).
+        while cursor.val_valid(batch) {
+            let v = cursor.val(batch);
+            cursor.map_times(batch, |t, diff| {
```
ublubu, line 52: _Irrelevant commentary:_ I dislike that this is called _map_ when it's actually _for each_. I was very confused by this block until I realized that.

src/interchange/src/envelopes.rs
```diff
@@ -11,124 +11,87 @@ use std::collections::BTreeMap;
...
+        // when `cnt = 1`. By contrast, `repeat_n((t, v), cnt)` will return the original `(t, v)` when `cnt = 1`,
+        // and only clone when `cnt > 1`.
+        let fan_out = |(t, v, cnt): (B::Time, B::ValOwn, usize)| iter::repeat_n((t, v), cnt);
+        let befores_iter = befores.drain(..).flat_map(fan_out);
+        let afters_iter = afters.drain(..).flat_map(fan_out);
+
+        let key_owned = <B::KeyContainer as BatchContainer>::into_owned(k);
+
+        for pair in
+            itertools::merge_join_by(befores_iter, afters_iter, |(t1, _v1), (t2, _v2)| t1.cmp(t2))
```
ublubu, line 83: I'm trying to understand why this is good enough. Like why it doesn't matter how the values line up.

So we're happy as long as (for a given `(key, timestamp)`) the DiffPairs sum up to the right set of `(value, diff)`?

DAlperin: > So we're happy as long as (for a given (key, timestamp)) the DiffPairs sum up to the right set of (value, diff)?

Basically yeah. Inside a single (key, t) the zip between befores and afters is technically arbitrary. Assume the following:
- befores (v1, v2)
- afters (v3, v4)

Then either of these are valid emissions:

- (-v1, +v3), (-v2, +v4)
- (-v1,+v4), (-v2, +v3)

Because, like you mentioned, it just has to tie out within a given (key, t), it's all the same downstream. There is no concept of time more granular than the timestamp with which to provide a deterministic ordering. We could impose a relatively arbitrary one (like sort by values within (k, t)) but I'm not really sure it's worth it.

ublubu: As long as everyone downstream promises to treat a sequence of DiffPairs (with the same key, t) as equivalent to their sum, the existing strategy seems ok :+1:

PR #36166, opened 2026-04-20
canary-environment: Also check sinks

def-:
Fixed by https://github.com/MaterializeInc/materialize/pull/34742, but forgot to reenable back then. Noticed in https://buildkite.com/materialize/nightly/builds/16146#019dad3b-086b-43a3-9d91-a1306401a640


bosconi (approved):
Looks good.

PR #36168, opened 2026-04-21
mz-debug: Fix replica query

def-:
Removed coalesce(heap_limit, memory_bytes + disk_bytes): current usage was used as heap limit, which seems wrong.

Using `DISTINCT ON (replica_id, process_id) ORDER BY occurred_at DESC` to deterministically pick the latest metric per process, and sum over them; instead of getting one arbitrary process.

Follow-up to #35200


sjwiesman (approved):
Great catch 👍

PR #36169, opened 2026-04-21
agents: Move to vendor-agnostic locations

def-:
I'm not sure if we want to only use Claude, Codex might also be useful.


bosconi (approved):
This looks good. It makes sense for us to switch to the provider-agnostic version of skills.

I can see the OpenAI and Claude versions of the skill name and description diverging. I will extend the `/mz-skills` skill linting command to check for divergence and optionally fix it.

I tested out the changes and Claude seemed happy.

antiguru (approved):

PR #36172, opened 2026-04-21
adapter,storage: fix replacement collection_metadata leak

teskje:
This commit fixes a leak of storage `collection_metadata` entries when dropping replacement MVs without applying them. The previous code intentionally avoided adding the IDs of dropped replacements to `storage_collections_to_drop`, to avoid finalizing the shard, but causing the leak in doing so. The solution implemented here is to make `prepare_state` smarter and only finalize the shard if the dropped collection is a primary.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/11322

### DIFF
--- a/test/cluster/mzcompose.py
+++ b/test/cluster/mzcompose.py
@@ -6431,6 +6431,53 @@ def workflow_github_10086(c: Composition) -> None:
     assert not upper_empty
 
 
+def workflow_github_11322(c: Composition) -> None:
+    """
+    Regression test for database-issues#11322, in which dropping a
+    replacement MV leaks a `collection_metadata` entry.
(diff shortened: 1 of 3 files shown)


test/cluster/mzcompose.py
```diff
@@ -6431,6 +6431,53 @@ def workflow_github_10086(c: Composition) -> None:
     assert not upper_empty
 
 
+def workflow_github_11322(c: Composition) -> None:
```
mtabebe, line 6434: Thanks for the repro

mtabebe (approved):

teskje:
TFTR!

PR #36183, opened 2026-04-21
`repeat_row`: Fix `SingleTimeMonotonic` to take into account FlatMap's TableFunc's `preserves_monotonicity`

ggevay:
This is the 2. PR in the `repeat_row` productionization work stream. Resolves SQL-162.

Fixes a part of https://github.com/MaterializeInc/database-issues/issues/11310, namely that `RelaxMustConsolidate`/`SingleTimeMonotonic` was not taking into account table functions' `preserves_monotonicity`. (can only happen `repeat_row` currently)

PR #36184, opened 2026-04-21
`repeat_row`: Document `refine_single_time_operator_selection`'s assumptions

ggevay:
This is the 3. PR in the `repeat_row` productionization work stream. This is on top of https://github.com/MaterializeInc/materialize/pull/36183. Resolves [SQL-163](https://linear.app/materializeinc/issue/SQL-163).

Only the **last 2 commits** are new:
1. slightly refactor some relevant tests;
2. slightly refactor `refine_single_time_operator_selection` and `refine_single_time_consolidation`.

This is now a pure refactoring and comment clarification PR, because I dropped its original last commit.

-------------

The original last commit wanted to bail out from `refine_single_time_operator_selection` when it sees `repeat_row`. However, I realized that that would not really solve anything, because the dataflow's inputs might also have negative accumulations (consolidated to a single time), coming from `repeat_row` in other dataflows. So, in this longish [slack thread](https://materializeinc.slack.com/archives/C08AENAP691/p1776884657932799?thread_ts=1776871751.101279&cid=C08AENAP691), we decided to not worry about monotonic operators getting negatively accumulating input, but instead just add the errors from monotonic operators to the list of errors that we now have to promote from "definitely internal error" to "maybe you used `repeat_row` wrong, or internal error" (which I will do in a follow-up PR). Note that we are not totally sure whether all monotonic operators will reliably error out when presented with non-monotonic data. However, in the slack thread we also decided to not aim for always erroring when there is negatively accumulating data, because this seems to be ~impossible to achieve without adding further resource usage to various operators (not just monotonic operators).

(Another possibility would be to bail out from `refine_single_time_operator_selection` when there is `repeat_row` in either the current dataflow or any of its inputs, recursively, but that would lose too much performance.)

### DIFF
--- a/src/compute-types/src/plan.rs
+++ b/src/compute-types/src/plan.rs
@@ -490,36 +490,9 @@ impl Plan {
             Self::refine_union_negate_consolidation(&mut dataflow);
         }
 
-        if dataflow.is_single_time() {
-            Self::refine_single_time_operator_selection(&mut dataflow);
+        Self::refine_single_time_operator_selection(&mut dataflow);
 
-            // The relaxation of the `must_consolidate` flag performs an LIR-based
-            // analysis and transform under checked recursion. By a similar argument
-            // made in `from_mir`, we do not expect the recursion limit to be hit.
-            // However, if that happens, we propagate an error to the caller.
-            // To apply the transform, we first obtain monotonic source and index
-            // global IDs and add them to a `TransformConfig` instance.
-            let monotonic_ids = dataflow
-                .source_imports
-                .iter()
-                .filter_map(|(id, source_import)| source_import.monotonic.then_some(*id))
-                .chain(
-                    dataflow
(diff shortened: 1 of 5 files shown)


ggevay:
Bringing this back to draft, because I realized there is an issue: With `repeat_row`, `Get`s are no longer logically monotonic in a single-time context, because somebody might have written negatively accumulating stuff into an MV or index.

Edit: [Being discussed on slack](https://materializeinc.slack.com/archives/C08AENAP691/p1776873024063549?thread_ts=1776871751.101279&cid=C08AENAP691).

ggevay:
Adjusted the PR, see the new PR description. Ready for review.

ggevay:
Bringing this back to draft, because we'd like to think through the exact invariants around monotonic operators more thoroughly, as discussed on zoom.

PR #36185, opened 2026-04-21
`repeat_row`: Make subquery guard aware of negative diffs

ggevay:
This is the 4. PR in the `repeat_row` productionization work stream. Resolve SQL-164.

This resolves https://github.com/MaterializeInc/materialize/pull/32978#discussion_r2200866973. It was not important originally, but now that we are productionizing `repeat_row`, the original error message would become confusing: it would say that "more than one record produced in subquery", when in fact there was a negative number of rows.

I didn't add a test, because I wasn't able to find a query where this error deterministically happens. If I just naively put a `repeat_row(-1)` inside a subquery, then sometimes this error happens, but sometimes a `Distinct` gives the negative accumulation error message.

### DIFF
--- a/src/expr/src/relation/func.rs
+++ b/src/expr/src/relation/func.rs
@@ -22,7 +22,7 @@ use mz_lowertest::MzReflect;
 use mz_ore::cast::CastFrom;
 
 use mz_ore::str::separated;
-use mz_ore::{soft_assert_eq_no_log, soft_assert_or_log};
+use mz_ore::{soft_assert_eq_no_log, soft_assert_or_log, soft_panic_or_log};
 use mz_repr::adt::array::ArrayDimension;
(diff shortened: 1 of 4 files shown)


frankmcsherry (approved):
Seems fine. Personally, the error seems most accurately "Error: not exactly one row", and the different variants for negative vs more than one may not pay dividends. But .. not opposed other than wrt the structured error bloat.

ggevay:
The reason for making the negative case its own error is that we'd like to handle this case specially: The error message should say something like "this is an internal error, unless you are using `repeat_row`". Also, we might want to create an incident for this one if `repeat_row` is not enabled in the user's env, but never create an incident for the `>1` case. (I'll do the error message tweak in a separate PR, together with all the other negative accumulation error msgs.)

PR #36188, opened 2026-04-21
console: Add user profile hook to show OIDC User details

leedqin:
### Motivation
We need to show user details when the user is logged in using SSO. Currently using the sql to fetch user information but I also added the hook to fetch any information from OIDC profile and use sql as the fallback if necessary if no claims are set.  Fixes: [CNS-58](https://linear.app/materializeinc/issue/CNS-58/display-user-id-access-in-console-for-oidc)

### Description
- Introduced `useCurrentUser` hook to fetch the SQL role of the connected session.
- Added `useOidcProfile` hook to extract and type OIDC user profile claims.
- Created `useUserProfile` hook to unify user identity from OIDC and SQL role.
- Implemented `CurrentUserTag` component to display the current user in the footer.
- Enhanced `PageFooter` to include the `CurrentUserTag` for better user visibility.

### Verification
image
image

### DIFF
--- a/console/src/api/materialize/useCurrentUser.ts
+++ b/console/src/api/materialize/useCurrentUser.ts
@@ -1,0 +1,54 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+import { QueryKey, useQuery } from "@tanstack/react-query";
+import { sql } from "kysely";
+
+import { buildRegionQueryKey } from "~/api/buildQueryKeySchema";
+
+import { queryBuilder } from "./db";
+import { executeSqlV2 } from "./executeSqlV2";
+
+export const currentUserQueryKeys = {
+  all: () => buildRegionQueryKey("currentUser"),
+};
+
+function buildCurrentUserQuery() {
+  return queryBuilder.selectNoFrom(sql<string>`current_user`.as("currentUser"));
+}
+
+async function fetchCurrentUser({
+  queryKey,
+  requestOptions,
+}: {
+  queryKey: QueryKey;
+  requestOptions?: RequestInit;
+}) {
+  return executeSqlV2({
+    queries: buildCurrentUserQuery().compile(),
+    queryKey,
+    requestOptions,
+  });
+}
+
+/** Fetches the SQL role the connected session is acting as. */
+export default function useCurrentUser() {
+  const { data, isLoading, error } = useQuery({
+    queryKey: currentUserQueryKeys.all(),
+    queryFn: ({ queryKey, signal }) =>
+      fetchCurrentUser({ queryKey, requestOptions: { signal } }),
+    staleTime: Infinity,
+  });
+
+  return {
+    results: data?.rows[0]?.currentUser,
+    isLoading,
+    error,
+  };
+}
--- a/console/src/layouts/ProfileDropdown.tsx
+++ b/console/src/layouts/ProfileDropdown.tsx
@@ -17,6 +17,7 @@ import {
   MenuGroup,
   MenuItem,
   MenuList,
+  Spinner,
   Tag,
   Text,
   useBreakpointValue,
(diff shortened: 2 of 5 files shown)


src/materialized/ci/listener_configs/oidc.json
```diff
@@ -5,6 +5,12 @@
       "authenticator_kind": "Oidc",
       "allowed_roles": "NormalAndInternal",
       "enable_tls": false
+    },
+    "internal": {
+      "addr": "0.0.0.0:6877",
+      "authenticator_kind": "Password",
+      "allowed_roles": "NormalAndInternal",
+      "enable_tls": false
```
SangJunBak: Should remove this

console/src/hooks/useUserProfile.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/**
+ * Unified view of who the user is, suitable for cross-auth-mode UI.
+ *
+ * The SQL role (`current_user`) is the ground truth — it is the identity
+ * Materialize itself recognizes. OIDC claims layer richer profile data on top
+ * when the IdP supplies `profile`/`email` scopes, but are purely optional:
+ * this hook falls back to the SQL role whenever a claim is missing.
+ */
```
SangJunBak: nit: I don't like how much fluff this comment has. I think we can describe this more concisely

console/src/hooks/useUserProfile.ts
```diff
@@ -0,0 +1,43 @@
...
+export const useUserProfile = (): UserProfile => {
+  const { name, email, picture } = useOidcProfile();
```
SangJunBak: Given we don't reuse useOidcProfile anywhere else, I think we should combine it with this and and possibly rename to `useSelfManagedProfile`

console/src/api/materialize/useCurrentUser.ts
```diff
@@ -0,0 +1,31 @@
...
+export const CURRENT_USER_QUERY_KEY = "currentUser";
+
+/** Fetches the SQL role the connected session is acting as. */
+export default function useCurrentUser() {
+  const query = React.useMemo(() => {
+    return queryBuilder
+      .selectNoFrom(sql<string>`current_user`.as("currentUser"))
+      .compile();
+  }, []);
+  const response = useSqlTyped(query, {
```
SangJunBak: I know we eventually want to phase out `useSqlTyped` and the other `useSql` hooks. I also think this query runs every the user switches pages because it remounts `BaseLayout` each time. Maybe we just use react query here with a stale time INFINITY then invalidate the react query cache on logout?

SangJunBak: We should probably handle the loading states. For the footer, maybe a `...` and for the menu item, maybe a spinner?

SangJunBak (changes requested):

console/src/hooks/useUserProfile.ts
```diff
@@ -0,0 +1,43 @@
...
+  const { results: sqlRole } = useCurrentUser();
+
+  return {
+    name: name ?? sqlRole,
+    email,
+    picture,
+    sqlRole,
```
SangJunBak: I don't think this is used?

console/src/hooks/useUserProfile.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Materialize itself recognizes. OIDC claims layer richer profile data on top
+ * when the IdP supplies `profile`/`email` scopes, but are purely optional:
+ * this hook falls back to the SQL role whenever a claim is missing.
+ */
+export const useUserProfile = (): UserProfile => {
+  const { name, email, picture } = useOidcProfile();
+  const { results: sqlRole } = useCurrentUser();
+
+  return {
+    name: name ?? sqlRole,
```
SangJunBak: There's a case where the menu item now shows the user's name and email via the ID token, but this could technically not be the SQL role name. Wonder if we should either:
- Show the SQL role instead of the name 
- Show all three

console/src/layouts/ProfileDropdown.tsx
```diff
@@ -139,6 +159,9 @@ const OidcSignOutMenuItem = () => {
         // Cookie may not exist if user logged in via OIDC only.
       }
     }
+    // Drop cached session-scoped data so a fresh sign-in doesn't see the
+    // previous user's results.
+    queryClient.clear();
```
SangJunBak: Actually, looking at the code a bit closer, it seems like all auth methods simply do a browser navigation back to the signin page, meaning the query client is implicitly cleared. We can probably get rid of this line of code

SangJunBak (approved):

PR #36189, opened 2026-04-21
`repeat_row`: Add Kafka sink regression test for negative accumulations

ggevay:
The main point of this PR is to show that we don't panic when `repeat_row` produces a negative accumulation in a collection that is sinked out.

The PR adds a test that shows the current behavior of Kafka sinks with negative accumulations: `combine_at_timestamp` unrolls a -N multiplicity into N before-only `DiffPair`s, so the Debezium envelope publishes N bogus delete records. A proper sink error state on net-negative accumulations is left as future work. (Note that we don't yet have a proper error state even for primary key violations: https://github.com/MaterializeInc/database-issues/issues/5099)

Resolves SQL-169.


martykulma (approved):
lgtm!

PR #36190, opened 2026-04-21
mysql: log row shape on decoding error

martykulma:
If MySQL fails to decode a binlog row, it unfortunately doesn't provide much information. This adds additional logging that provide the row shape information for debugging.

In pushing this further down into `pack_mysql_row`, I did some additional refactoring.

Sample output:
```
WARN timely{name="storage" worker_id=0}: mz_mysql_util::decoding: mysql decode error for `public`.`decode_err` column `f2`: Couldn't convert the value `Time("'029:59:59'")` to a desired type; gtid_set=Some("Partitioned((Interval { lower: 693e00a9-3dce-11f1-9e2c-ca0de7a4a79f, upper: 693e00a9-3dce-11f1-9e2c-ca0de7a4a79f }, Active(11)))"); row_shape=[{name=f1, wire=MYSQL_TYPE_LONG, charset=0, expected=Int32, val=int}, {name=f2, wire=MYSQL_TYPE_TIME2, charset=0, expected=Time, val=time}]
```

### DIFF
--- a/src/storage/src/source/mysql/replication/events.rs
+++ b/src/storage/src/source/mysql/replication/events.rs
@@ -7,12 +7,9 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::fmt::Write;
-
 use maplit::btreemap;
-use mysql_common::Value;
 use mysql_common::binlog::events::{QueryEvent, RowsEventData};
-use mz_mysql_util::{MySqlError, MySqlTableDesc, pack_mysql_row};
+use mz_mysql_util::{MySqlError, pack_mysql_row};
 use mz_ore::iter::IteratorExt;
 use mz_repr::{Diff, Row};
 use mz_storage_types::errors::DataflowError;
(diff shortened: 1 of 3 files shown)


ublubu (approved):
seems legit

src/storage/src/source/mysql/replication/events.rs
```diff
@@ -301,16 +304,25 @@ pub(super) async fn handle_rows_event(
         for (binlog_row, diff) in updates.into_iter().flatten() {
             let row = mysql_async::Row::try_from(binlog_row)?;
             for (output, row_val) in outputs.iter().repeat_clone(row) {
+                let row_shape = describe_row_shape(&row_val, &output.desc);
```
ublubu: I'm guessing we aren't bothered by creating all these strings.

martykulma: good catch - I'll push it down to only get built in `pack_mysql_row`

martykulma: @patrickwwbutler ptal - moving the error handling further down required me to make some additional changes that you were probably also making.

I ran with the changes in [11312](https://github.com/MaterializeInc/database-issues/issues/11312) and it passed successfully.

patrickwwbutler:
closed since this was done in another PR

PR #36191, opened 2026-04-21
`repeat_row`: Add comments around `snapshot_latest`, documenting the set invariant

ggevay:
This is another PR in the `repeat_row` workstream. I just wanted to make sure that `repeat_row` with a negative count can't possibly make the `assert_eq!(diff, 1, "snapshot doesn't accumulate to set");` in `snapshot_latest` fire. This seems to be the case, since this is reading a collection that is not a user-collection, so data from `repeat_row` can't reach it. (In fact, the set invariant is a much stronger invariant.)

This PR just adds code comments documenting the set invariant, and why it holds at the only caller of `snapshot_latest`.

Resolves SQL-171.

### DIFF
--- a/src/storage-client/src/storage_collections.rs
+++ b/src/storage-client/src/storage_collections.rs
@@ -163,8 +163,20 @@ pub trait StorageCollections: Debug + Sync {
         as_of: Timestamp,
     ) -> BoxFuture<'static, Result<Vec<(Row, StorageDiff)>, StorageError>>;
 
-    /// Returns a snapshot of the contents of collection `id` at the largest
-    /// readable `as_of`.
+    /// Returns a snapshot of the contents of collection `id` at the largest readable `as_of`.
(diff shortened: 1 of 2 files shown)


src/storage-client/src/storage_collections.rs
```diff
@@ -1492,10 +1501,12 @@ impl StorageCollections for StorageCollectionsImpl {
             Some(f) if f > &Timestamp::MIN => {
                 let as_of = f.step_back().unwrap();
 
-                let snapshot = self.snapshot(id, as_of, &self.txns_read).await.unwrap();
+                let snapshot = self.snapshot(id, as_of, &self.txns_read).await?;
```
ggevay, line 1507: Not so important for the main point of the PR, but I did it because why `unwrap` when you can just `?`-propagate the same error?

SangJunBak (approved):

PR #36193, opened 2026-04-21
console: surface OIDC sign-in failures on the login page

leedqin:
### Motivation

Addressing SSO Feedback to show better error messages in the console if there are SSO failures and not have a confusing user experience. Fixes [CNS-53](https://linear.app/materializeinc/issue/CNS-53/improve-oidc-console-errors)

### Description

 - Added OIDC error messages to the console showing the errors when sign in failed/ surface environmentd warnings up in the console
### Verification
When the sign in failed and if the user is not part of the organization
image

When OIDC might not be configured properly
image

### DIFF
--- a/console/src/platform/auth/OidcCallback.tsx
+++ b/console/src/platform/auth/OidcCallback.tsx
@@ -1,0 +1,27 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+import React from "react";
+import { Navigate } from "react-router-dom";
+
+import { LOGIN_ERROR_PARAM, LOGIN_PATH } from "~/api/materialize/auth";
+import LoadingScreen from "~/components/LoadingScreen";
+import { useAuth } from "~/external-library-wrappers/oidc";
+
+// Forwards IdP callback errors to the login page so all sign-in errors
+// render in one place.
+export const OidcCallback = () => {
+  const auth = useAuth();
+  if (auth.error) {
+    const message = auth.error.message?.trim() || "Sign-in failed";
+    const params = new URLSearchParams({ [LOGIN_ERROR_PARAM]: message });
+    return <Navigate to={`${LOGIN_PATH}?${params.toString()}`} replace />;
+  }
+  return <LoadingScreen />;
+};
--- a/console/src/utils/oidcAuth.ts
+++ b/console/src/utils/oidcAuth.ts
@@ -1,0 +1,23 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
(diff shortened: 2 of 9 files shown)


src/environmentd/src/http.rs
```diff
@@ -854,26 +854,33 @@ pub(crate) enum AuthError {
     FailedToUpdateSession,
     #[error("invalid credentials")]
     InvalidCredentials,
+    /// Payload is `OidcError`'s sanitized `Display` (no expected-values leaks).
```
SangJunBak, line 886: I think similar to the pgwire tests in src/environmentd/tests/auth.rs, we should have tests asserting the details for the http errors.

console/src/utils/oidcAuth.ts
```diff
@@ -0,0 +1,35 @@
...
+import { type LoginReason } from "~/platform/auth/constants";
+
+// Cap on the forwarded OIDC detail so a malformed 401 body can't bloat the URL.
+const MAX_AUTH_DETAIL_LENGTH = 200;
+
+const hasBearer = (input: Parameters<typeof fetch>[0]): boolean =>
+  input instanceof Request &&
+  (input.headers.get("Authorization")?.startsWith("Bearer ") ?? false);
+
+/** Reads the sanitized `OidcError::Display` environmentd returns for OIDC 401s. */
```
SangJunBak: This feels like a hacky way to differentiate login errors from session expiration. Rather than use local storage, can we encode the error message into the URL (like a search param) then on the login page, just show the error there?

console/src/platform/auth/OidcCallback.tsx
```diff
@@ -0,0 +1,53 @@
...
+import LoadingScreen from "~/components/LoadingScreen";
+import { MaterializeLogo } from "~/components/MaterializeLogo";
+import { useAuth } from "~/external-library-wrappers/oidc";
+import { AuthContentContainer, AuthLayout } from "~/layouts/AuthLayout";
+
+// Surfaces OIDC callback errors (access_denied, state mismatch, etc.)
+export const OidcCallback = () => {
+  const auth = useAuth();
+
+  if (!auth.error) return <LoadingScreen />;
```
SangJunBak: I think we should unify the error behavior with errors from the Materialize side and just redirect to the login page with the error message.

console/src/platform/auth/constants.ts
```diff
@@ -0,0 +1,34 @@
...
+  reason: LoginReason;
+  detail?: string;
+};
+
+const AUTH_REJECTED_SUFFIX =
+  " Please try again, or contact your administrator if this persists.";
+
+export const LOGIN_REASON_MESSAGES: Record<LoginReason, string> = {
+  auth_rejected: `Sign-in was rejected.${AUTH_REJECTED_SUFFIX}`,
+  session_expired: "Your previous session ended. Please sign in again.",
```
SangJunBak: We could just not print an error for session expiration because we don't know definitively that a 401 with no detail is the session expiring.

SangJunBak (changes requested):

SangJunBak (approved):
LGTM! Thanks for addressing my changes!

PR #36201, opened 2026-04-22
CI: improve Rust ICE detection, fix cargo doc tests

def-:
See https://materializeinc.slack.com/archives/C097A6PRLQ2/p1776778558640689?thread_ts=1776777023.701559&cid=C097A6PRLQ2 for context

Turns out we were just ignoring Cargo doc test failing before, ouch!


mtabebe (approved):
Thanks for adding the incremental detection logic and the docs fix

PR #36205, opened 2026-04-22
cli: Support `--foundationdb` for nextest

def-:
Follow-up to https://github.com/MaterializeInc/materialize/pull/35025 to support `bin/cargo-test --foundationdb`


antiguru (approved):
Thank you!

PR #36206, opened 2026-04-22
storage: fix iceberg upsert sink dropping deletes with >100k distinct keys

DAlperin:
## Summary

Fixes https://github.com/MaterializeInc/database-issues/issues/11326

The `DeltaWriter` in iceberg-rust evicts from `seen_rows` once the count exceeds `max_seen_rows` (default 100k), and falls back to an equality delete for any delete whose row is no longer tracked. But equality deletes in an Iceberg snapshot only apply to data files with strictly lower sequence numbers — they do not delete rows written to the same snapshot. When an upsert batch inserts more than 100k distinct keys and then issues deletes for any of the evicted-but-same-session rows, the equality deletes silently no-op and the old values remain alongside the new inserts.

- Disable eviction for incremental upsert batches by setting `max_seen_rows` to `usize::MAX`, so every row inserted in this session stays trackable via a position delete (which works within a single snapshot).
- Snapshot batches keep `max_seen_rows=0` — they only produce inserts, so no tracking is needed.
- Add a regression test (`test/iceberg/large-upsert-batch.td`) that inserts 110k rows and then updates 100 of them with a long `COMMIT INTERVAL`, forcing both operations into a single DeltaWriter session. The fixed sink commits exactly 110,000 rows; before the fix it committed 110,100.

### Memory caveat / follow-up

Disabling eviction makes `seen_rows` memory O(distinct keys per batch). For typical upsert workloads this is small, but sinks with very long `COMMIT INTERVAL`s and large update volumes will hold more state in memory during the batch window. A proper bounded-memory fix would require spilling `seen_rows` to disk or something

## Test plan

- [x] `bin/mzcompose --find iceberg run large_upsert_batch`
- [ ] Confirm the test fails on main and passes with the fix.

### DIFF
--- a/src/storage/src/sink/iceberg.rs
+++ b/src/storage/src/sink/iceberg.rs
@@ -309,9 +309,18 @@ impl EnvelopeHandler for UpsertEnvelopeHandler {
             self.equality_ids.clone(),
         );
 
-        if is_snapshot {
-            builder = builder.with_max_seen_rows(0);
-        }
+        // Snapshot batches only produce inserts, so disable seen_rows tracking
+        // to save memory. For incremental batches, disable eviction: the
+        // DeltaWriter falls back to equality deletes when a seen row is
+        // evicted, but equality deletes only apply to prior snapshots (lower
+        // sequence number). Evicting a row inserted in this same session would
+        // silently drop the delete and leave both old and new values in the
+        // committed snapshot.
+        builder = if is_snapshot {
+            builder.with_max_seen_rows(0)
+        } else {
+            builder.with_max_seen_rows(usize::MAX)
+        };
 
         Ok(Box::new(
             builder
--- a/test/iceberg/large-upsert-batch.td
+++ b/test/iceberg/large-upsert-batch.td
@@ -1,0 +1,74 @@
+# Copyright Materialize, Inc. and contributors. All rights reserved.
+#
+# Use of this software is governed by the Business Source License
+# included in the LICENSE file at the root of this repository.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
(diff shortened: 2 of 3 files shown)


def- (approved):

src/storage/src/sink/iceberg.rs
```diff
@@ -309,9 +309,18 @@ impl EnvelopeHandler for UpsertEnvelopeHandler {
...
-        if is_snapshot {
-            builder = builder.with_max_seen_rows(0);
-        }
+        // Snapshot batches only produce inserts, so disable seen_rows tracking
+        // to save memory. For incremental batches, disable eviction: the
+        // DeltaWriter falls back to equality deletes when a seen row is
+        // evicted, but equality deletes only apply to prior snapshots (lower
+        // sequence number). Evicting a row inserted in this same session would
+        // silently drop the delete and leave both old and new values in the
+        // committed snapshot.
```
ublubu, line 318: This last sentence confused me before I read the PR description thoroughly. And it kinda still confuses me.

IIUC, the issue is:
> deletes for any of the evicted-but-same-session rows, the equality deletes silently no-op

It's not clear to me what "old and new values" means here:
> leave both old and new values in the committed snapshot

I thought the problem was that if we overflow the "seen rows" window, if we delete a row we inserted (into the snapshot) at the beginning of that window, it doesn't actually get deleted (from the snapshot).

DAlperin: Lets say I have upsert key `a`. I write the row `a=foo`, then 100k rows later I write `a=bar`. What I would want to happen is `a=foo` is deleted, and `a=bar` is left. But with this bug the snapshot would include `a=foo` and `a=bar` both the "old and new" values.

PR #36208, opened 2026-04-22
postgres source: Use pg_catalog's server_version_num only

def-:
Otherwise we look in search_path, and a user may have their own `server_version_num`

Follow-up to #34934


martykulma (approved):
thanks @def- !

PR #36210, opened 2026-04-22
sql: Clarify fast path peek storage reads

mgree:
### Motivation

https://materializeinc.slack.com/archives/CU7ELJ6E9/p1776710161607899?thread_ts=1776707127.468209&cid=CU7ELJ6E9

### Description

Clarifies wording for `FastPathPlan::PeekPersist` as `ReadStorage` rather than `Indexed` (which is the correct wording for `FastPathPlan::PeekExisting`, which actually does read from an index).

### Verification

Rewrite SLTs as necessary.

### DIFF
--- a/test/sqllogictest/explain/default.slt
+++ b/test/sqllogictest/explain/default.slt
@@ -2128,3 +2128,76 @@ Explained Query:
 Target cluster: no_replicas
 
 EOF
+
+# fast path operator names
+statement ok
+CREATE TABLE fp_t_raw(x int, y text);
+
+statement ok
+CREATE MATERIALIZED VIEW fp_mv_filtered AS SELECT * FROM fp_t_raw WHERE x IS NOT NULL;
+
+query T multiline
+EXPLAIN
+SELECT * FROM fp_mv_filtered LIMIT 5
+----
+Explained Query (fast path):
+  Finish limit=5 output=[#0, #1]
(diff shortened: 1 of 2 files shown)


frankmcsherry (approved):
LGTM!

test/sqllogictest/explain/default.slt
```diff
@@ -2128,3 +2128,76 @@ Explained Query:
...
+Explained Query (fast path):
+  Finish limit=5 output=[#0, #1]
+    →Map/Filter/Project
+      →ReadStorage materialize.public.fp_mv_filtered
+
+Target cluster: no_replicas
+
+EOF
+
+# this does not seem to be successfully hitting the `ReadStorage`/`Lookup` path, so we don't have 100% coverage :(
```
ggevay, line 2152: Hmm, weirdly, even `persist-fast-path.slt` doesn't have coverage for this, as far as I can tell by a quick scan. This would be the point where ask Ben...

ggevay (approved):
Thanks!

test/sqllogictest/explain/default.slt
mgree: 🤷  I'm okay not getting 100% coverage here. Made an issue, for future reference: https://github.com/MaterializeInc/database-issues/issues/11328

PR #36228, opened 2026-04-23
adapter: don't apply Subscribe transaction ops for EXPLAIN SUBSCRIBE

def-:
`subscribe_validate` is shared between real `SUBSCRIBE` and `EXPLAIN SUBSCRIBE` (introduced in #34912). `TransactionOps::Subscribe` is exclusive, so running `EXPLAIN SUBSCRIBE` inside an explicit transaction poisoned it and caused any following statement to fail with `SubscribeOnlyTransaction`.


antiguru (approved):
Thanks!

PR #36229, opened 2026-04-23
`repeat_row`: Wrangle negative accumulation error messages

ggevay:
Resolves [SQL-165](https://linear.app/materializeinc/issue/SQL-165).

We have tons of places in the code where we emit errors for negative accumulations (or similar conditions), but they were quite inconsistent, and also need updating as part of productionizing `repeat_row`. (But we wanted to do this even before `repeat_row` came up: https://github.com/MaterializeInc/database-issues/issues/9526) This PR achieves several things:
- Adjusts the base message (see `Display for MultiplicityError`):
  - Don't start with "internal error:". (Even before `repeat_row`, this was not always the case: sometimes it was a bug/misconfiguration in an external system.)
  - Clarifies "Invalid data in source": either a bug or misconfiguration in the external system feeding the source, or a bug in Materialize.
  - Adds a mention of `repeat_row` being a possible cause.
- Centralizes these errors into a new structured error: `EvalError::MultiplicityError`, so that it will be easier to
  - have an overview of such errors, search for all places that emit them, etc.
  - emit them in a consistent way
  - adjust the wording of the error message for all sites
- Makes Sentry error titles the same (`invalid record multiplicity`), so that Sentry will group them into one issue.
- Fixes various inconsistencies:
  - Some sites didn't do the `warn!` + `error!` double logging dance (which now `log_multiplicity_error` does).
  - Not all sites were mentioning sources.
  - One site used `EvalError::InvalidParameterValue`, which didn't really make sense.

See commit msg for more details.

@MaterializeInc/sources-sinks, could you please check that what the new error message (see `Display for MultiplicityError`) says about sources is accurate?

```
"{problem} in {code_place}; this usually indicates that a \
Materialize source contains invalid data (for example, \
retractions for rows that were never inserted). This can be \
caused either by a bug or misconfiguration in the external \
system feeding the source, or by a bug in Materialize. It \
can also occur due to incorrect usage of the repeat_row \
table function.",
```

@materializekatrina, could you please check this from an alerting point of view? Log lines will now look like the following. Every occurrence will have the string `invalid record multiplicity`. Is this nicely alertable, or should I adjust something on this?
```
cluster-u1-replica-u1-gen-0: 2026-04-26T17:05:54.254080Z  WARN timely{name="compute" worker_id=0}: mz_expr::scalar: [customer-data] invalid record multiplicity: target=t38, diff=-2, row=[Int64(1), UInt32(1)] kind=Negative code_place=index peek
cluster-u1-replica-u1-gen-0: 2026-04-26T17:05:54.254118Z ERROR timely{name="compute" worker_id=0}: mz_expr::scalar: invalid record multiplicity
```

Nightly: https://buildkite.com/materialize/nightly/builds/16212

### DIFF
--- a/src/compute/src/render/errors.rs
+++ b/src/compute/src/render/errors.rs
@@ -27,6 +27,7 @@
 use columnar::Columnar;
 use columnation::{Columnation, Region};
 use mz_expr::EvalError;
+use mz_expr::{MULTIPLICITY_ERROR_TITLE, MultiplicityError};
 use mz_proto::{ProtoType, RustType};
 use mz_repr::Row;
 use mz_storage_types::errors::{DataflowError, ProtoDataflowError};
@@ -213,6 +214,7 @@ impl ErrorLogger {
     /// the breadcrumbs to their associated error in Sentry.
     ///
     // TODO(database-issues#5362): Rethink or justify our error logging strategy.
+    #[allow(dead_code)] // kept for future non-multiplicity uses
     pub fn log(&self, message: &'static str, details: &str) {
         tracing::warn!(
             dataflow = self.dataflow_name,
@@ -230,6 +232,30 @@ impl ErrorLogger {
             "[customer-data] {message} ({details})"
         );
         mz_ore::soft_panic_or_log!("{}", message);
+    }
+
+    /// Log a [`MultiplicityError`] using the standard "static title + [customer-data]
+    /// tagged details" contract, attaching the dataflow name of this [`ErrorLogger`]
+    /// as a structured field.
+    ///
+    /// Use this from call sites that have an [`ErrorLogger`] in scope (dataflow
+    /// rendering code in reduce/top-k); other call sites should use the free
+    /// [`mz_expr::log_multiplicity_error`] function instead.
+    ///
+    /// `kind` and `code_place` are emitted as structured tracing fields.
+    /// `detail` is emitted in the `[customer-data]`-tagged message body.
+    pub fn log_multiplicity_error(&self, err: &MultiplicityError) {
+        tracing::warn!(
+            dataflow = self.dataflow_name,
+            kind = ?err.kind,
+            code_place = %err.code_place,
+            "[customer-data] {}: {}",
+            MULTIPLICITY_ERROR_TITLE,
+            err.detail,
+        );
+        // Note: simply `tracing::error!(MULTIPLICITY_ERROR_TITLE);` wouldn't work well, because it
+        // would surprisingly print as `MULTIPLICITY_ERROR_TITLE="invalid record multiplicity"`.
+        tracing::error!("{}", MULTIPLICITY_ERROR_TITLE);
     }
 }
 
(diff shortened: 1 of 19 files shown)


materializekatrina:
I’ve added the invalid record multiplicity pattern to the existing Corruption Errors alert in Grafana. Just noting that this continues our current (not ideal) approach of alerting on log patterns, but it keeps us covered for now while we work toward a better long-term solution (discussed [here](https://materializeinc.slack.com/archives/C012BPZCGRG/p1772465820469789?thread_ts=1772218250.978049&cid=C012BPZCGRG) most recently).

ggevay:
(Alerting is being discussed [on slack](https://materializeinc.slack.com/archives/C08ACQNGSQK/p1777301556868459?thread_ts=1777296911.137909&cid=C08ACQNGSQK).)

src/compute/src/render/errors.rs
```diff
@@ -231,6 +233,30 @@ impl ErrorLogger {
...
+    /// as a structured field.
+    ///
+    /// Use this from call sites that have an [`ErrorLogger`] in scope (dataflow
+    /// rendering code in reduce/top-k); other call sites should use the free
+    /// [`mz_expr::log_multiplicity_error`] function instead.
+    ///
+    /// `kind` and `code_place` are emitted as structured tracing fields.
+    /// `detail` is emitted in the `[customer-data]`-tagged message body.
+    pub fn log_multiplicity_error(&self, err: &MultiplicityError) {
+        tracing::warn!(
```
patrickwwbutler, line 248: some of the existing tracing messages are `error` level, are we sure we want this to only be a `warn`?

patrickwwbutler (commented):
LGTM from the sources side. The proposed error message seems accurate to me, and I don't see a reason this error template would cause issues unless we have some mysterious, cursed logic parsing error messages to check for negative accumulations.

patrickwwbutler (approved):

src/compute/src/render/errors.rs
ggevay: There is also an `error!` in the line below. So, the `error` triggers some alert, and then one can still look up the details in the corresponding `warn`. For example, Sentry has the "Breadcrumbs", which shows the last few log msgs from before the error that triggered the Sentry issue.

Making the `error` have a simple static message helps with Sentry grouping.

patrickwwbutler:
one thing I just thought of - not sure if it's relevant here, but @martykulma and I were talking about it on another PR, is the fact that we need to be able to retract some errors, and in those cases, the error message must match, or else (ironically) we get a negative multiplicity on the new error message? 

But I don't think this is relevant here because these errors cannot be retracted (retractable errors must be related to one specific record update?), but maybe if there were to be a record that undoes whatever made the negative accumulation, we would want to retract?

Not something to block the PR on, mostly just curiosity on the topic of error retractions

ggevay:
Bringing back to draft for now, because `repeat_row` has been cancelled.

But I'd still like to move forward with this PR regardless of `repeat_row`, because most of the issues being fixed by this PR are still valid, even for just our "normal" negative accumulation issues that sometimes come up without `repeat_row`. So, I'll change the PR to remove the `repeat_row`-specific parts, and then it will be ready for a Cluster Team review.

PR #36233, opened 2026-04-23
[catalog] fix snapshot doubling check to run per-timestamp in sync_inner

mtabebe:
The doubling consolidation check was outside the per-timestamp `while let` loop, only firing once per Progress event. A single `fetch_next` call can deliver many timestamps in one batch, so all those updates were applied before consolidation could fire, letting the snapshot grow past the 2x bound.

Move the check inside the per-timestamp loop so consolidation fires as soon as the snapshot doubles, regardless of how persist batches deliveries. This preserves the amortized O(N log N) cost while maintaining the 2x memory bound.

### DIFF
--- a/src/catalog/src/durable/persist.rs
+++ b/src/catalog/src/durable/persist.rs
@@ -367,6 +367,10 @@ pub(crate) struct PersistHandle<T: TryIntoStateUpdateKind, U: ApplyUpdate<T>> {
     bootstrap_complete: bool,
     /// Metrics for the persist catalog.
     metrics: Arc<Metrics>,
+    /// Snapshot size at the last amortized consolidation, used by
+    /// [`Self::maybe_consolidate`] to decide when to consolidate. Initialized
+    /// lazily on the first call.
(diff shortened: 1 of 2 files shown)


src/catalog/src/durable/persist.rs
```diff
@@ -599,12 +599,12 @@ impl<T: TryIntoStateUpdateKind, U: ApplyUpdate<T>> PersistHandle<T, U> {
                                 },
                             );
                             self.apply_updates(updates)?;
-                        }
```
aljoscha, line 602: This looks good and I normally don't want to bikeshed too much. But maybe we could move this check into a `fn maybe_consolidate()` (which has the amortization check), and call that from the loop and other places. That way we don't have the size check noise right here in this bigger function.

Please ignore if you don't like, because this is okay as is.

aljoscha (approved):

src/catalog/src/durable/persist.rs
mtabebe: Sure, I'll do the refactor :)

PR #36237, opened 2026-04-24
adapter: fix leaked pending_peeks on concurrent replica drop

def-:
PR #34728 changed implement_peek_plan to return an error instead of terminating when compute.peek() fails due to a concurrent replica or cluster drop. However, the pending_peeks and client_pending_peeks map entries are inserted before the compute.peek() call, so on error they are never cleaned up. This leaks memory for the session lifetime and causes statement logging to record the wrong execution outcome (Canceled instead of Errored).

Fix by moving the map insertions to after compute.peek() succeeds. This is safe because the coordinator holds &mut self, so no peek response can be processed before this method returns.

### DIFF
--- a/src/adapter/src/coord/peek.rs
+++ b/src/adapter/src/coord/peek.rs
@@ -830,22 +830,6 @@ impl crate::coord::Coordinator {
             uuid = Uuid::new_v4();
         }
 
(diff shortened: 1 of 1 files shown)


ggevay (approved):
Indeed, thank you!

PR #36238, opened 2026-04-24
adapter: Fix statement_timeout=0 causing immediate timeout in frontend peek

def-:
PR #34796 wrapped the optimizer call in `tokio::time::timeout(statement_timeout)`, but did not map `Duration::ZERO` to `Duration::MAX`. In PostgreSQL (and the rest of this codebase), `statement_timeout = 0` means "disabled / wait forever." With `Duration::ZERO`, `tokio::time::timeout` fires immediately, so every SELECT and EXPLAIN FILTER PUSHDOWN fails with a spurious StatementTimeout error when `statement_timeout = 0`.

Fix both occurrences (frontend_peek.rs and sequencer.rs) to match the existing pattern in coord/sequencer/inner.rs.

### DIFF
--- a/src/adapter/src/coord/sequencer.rs
+++ b/src/adapter/src/coord/sequencer.rs
@@ -15,6 +15,7 @@
 use std::collections::{BTreeMap, BTreeSet};
 use std::str::FromStr;
 use std::sync::Arc;
+use std::time::Duration;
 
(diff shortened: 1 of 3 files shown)


ggevay (approved):
Oops, amazing special case! :D Thank you!

PR #36244, opened 2026-04-24
storage/postgres: drain replication stream via dedicated task

antiguru:
Move the `LogicalReplicationStream` into a dedicated Tokio task that forwards messages through a bounded mpsc and accepts standby status update commands via a second mpsc.
The outer operator's `select!` now wakes on `recv_many` and drains a full batch per wake-up, instead of paying the 4-way poll assembly cost per replication message.

Keepalive cadence, LSN handling, and replication semantics are unchanged.

### Motivation

Profiles of hot Postgres CDC sources show per-message overhead dominated by `tokio::select!` arm assembly and the outer `async_stream::try_stream!` yield round-trip.
Existing sinks (`iceberg`, `persist_sink`) use the drain-first pattern; this brings the Postgres replication reader in line.

### Design notes

* Task owns the stream and replication client. `biased` select prioritises keepalive commands over inbound messages so the timer never starves.
* Bounded channels (1024 msg, 8 cmd) preserve back-pressure: if the outer operator stalls, `msg_tx.send()` blocks the task — same end-to-end rate as before, just one batch buffered.
* `StreamExt::next` is cancel-safe over `CopyBothStream` (mpsc-backed), so racing it against `cmd_rx.recv()` is safe.
* Stream task exit surfaces through `msg_rx` as an error or as `TransientError::ReplicationEOF` (clean close).

### DIFF
--- a/src/storage/src/source/postgres/replication.rs
+++ b/src/storage/src/source/postgres/replication.rs
@@ -752,17 +752,73 @@ async fn raw_stream<'a>(
         })
         .abort_on_drop();
 
-    let stream = async_stream::try_stream!({
-        // Ensure we don't pre-drop the task
-        let _max_lsn_task_handle = max_lsn_task_handle;
+    // Spawn a dedicated task that owns the replication stream. It delivers received replication
+    // messages via `msg_rx` and accepts standby-status-update commands via `cmd_rx`. Moving the
+    // stream off the operator's select! lets the operator drain many buffered messages per wake-up
+    // instead of paying the 4-way poll assembly cost per message.
+    //
+    // `STREAM_MSG_CAPACITY` bounds how many messages may buffer between the stream task and the
+    // outer operator before the task back-pressures on `msg_tx.send()`. The outer loop drains up
+    // to this many messages per wake-up.
+    const STREAM_MSG_CAPACITY: usize = 1024;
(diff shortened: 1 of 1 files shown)


antiguru:
I think this should be fine, but low-ish confidence as I'm not 100% familiar with async intrinsics.

martykulma:
Nice! Could we use `StreamExt::ready_chunks` instead (e.g. `let mut stream = pin!(LogicalReplicationStream::new(copy_stream).ready_chunks(1024));`)? It seems it should give us the batching without having to maintain the task, et al. That's assuming the standby status isn't load bearing in this (doesn't seem to be).

antiguru:
@martykulma the standby status update is load-bearing, so `ready_chunks` alone doesn't cover it. The stream is duplex: we read messages *and* must call `stream.standby_status_update(.., reply=1)` on the same stream to proactively prompt a `PrimaryKeepAlive`, which is how the frontier advances when there's no data. `ready_chunks` wraps and consumes the stream, so we'd still need a separate path to issue those writes (only reachable via `get_pin_mut`, interleaved with the chunk buffer). That duplex requirement is exactly why the task owns the stream and takes commands over `cmd_rx`. `ready_chunks` would batch the read side but leave the write side needing its own machinery, so no net simplification.

Rebased onto current `main` and re-pushed.

PR #36248, opened 2026-04-24
storage: replace ColumnationStack source edges with Vec

antiguru:
Source dataflows used `ColumnationStack<(D, T, Diff)>` on the output edge of every reader (kafka, generator, mysql, postgres, sql_server) only to drive a byte-based yield threshold in `AsyncOutputHandle::give_fueled`.
Downstream operators immediately unpacked every row via `iter`/`drain`, so the region copy into columnation was pure overhead — every emitted row paid one extra heap copy and then was read back out on the next operator hop.

This replaces `AccountedStackBuilder` with a container-agnostic `AccountedBuilder` that forwards to an inner `ContainerBuilder`, and changes `give_fueled` to take an explicit `size_bytes` argument supplied by the caller.
Data rows account for `SourceMessage::byte_len()`; definite errors (rare) account for `std::mem::size_of_val(&update)`.

Source outputs and the `StackedCollection` alias now use `Vec<(D, T, Diff)>`, and the `SourceMessage` `Columnation` implementation is removed.
The `Partition` route closures and downstream unary consumers update to match the new owned-item container semantics.

### Motivation

The `AccountedStackBuilder` + `ColumnationStack` combo existed purely to compute `heap_size` for yield accounting.
Since readers emit rows one at a time and the next operator always re-materializes them, the region-copy cost was pure overhead.

### DIFF
--- a/src/storage/src/source/generator/key_value.rs
+++ b/src/storage/src/source/generator/key_value.rs
@@ -21,7 +21,7 @@ use mz_storage_types::errors::DataflowError;
 use mz_storage_types::sources::load_generator::{KeyValueLoadGenerator, LoadGeneratorOutput};
 use mz_storage_types::sources::{MzOffset, SourceTimestamp};
 use mz_timely_util::builder_async::{OperatorBuilder as AsyncOperatorBuilder, PressOnDropButton};
-use mz_timely_util::containers::stack::AccountedStackBuilder;
+use mz_timely_util::containers::stack::AccountedBuilder;
 use rand_8::rngs::StdRng;
 use rand_8::{RngCore, SeedableRng};
 use timely::container::CapacityContainerBuilder;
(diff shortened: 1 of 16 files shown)


src/storage/src/source/generator/key_value.rs
```diff
@@ -195,7 +203,11 @@ pub fn render<'scope>(
                     for sp in local_partitions.iter_mut() {
                         let mut emitted_all_exports = 0;
                         for u in sp.produce_batch(&mut value_buffer, &output_indexes) {
-                            data_output.give_fueled(&cap, u).await;
+                            let size = match &u.0.1 {
+                                Ok(msg) => msg.byte_len(),
+                                Err(_) => 0,
```
martykulma: should this be Err(e) => std::mem::size_of_val(e)?  I'm unclear why some of these compute to 0 and some not, so I'm assuming these are mistakes vs. intentional.

martykulma (commented):
Rather than have these bespoke size implementations, would it make sense to declare `((usize, Result<SourceMessage, DataflowError>), Timestamp, Diff)` as a type and have `fuel_size()` method to calculate?

```
let update = ((output, message), offset, diff);
let size = update.fuel_size();
```

src/storage/src/source/generator/key_value.rs
antiguru: Done. Introduced a `FuelSize` trait (in `source/types.rs`) with implementations for `SourceMessage`, `Row`, `Vec<u8>`, `Result<T, E>`, the source coordinate primitives, and tuples. Each call site is now just `update.fuel_size()`, and `Result` charges `size_of_val` for the `Err` arm.

src/storage/src/source/postgres/replication.rs
```diff
@@ -559,7 +574,6 @@ pub(crate) fn render<'scope>(
         .cycle();
     let round_robin = Exchange::new(move |_| next_worker.next().unwrap());
     let replication_updates = data_stream
-        .map::<Vec<_>, _, _>(Clone::clone)
```
antiguru, line 562: All to remove this clone...

martykulma (approved):
well, we do call that a lot 😂 

Thanks @antiguru - lgtm!

antiguru:
Thanks for the review!

PR #36249, opened 2026-04-24
docs: update MCP agent guide for the new data product requirements

bobbyiliev:
Updates the MCP agent guide to match the new data product filter: materialized views are exposed automatically, regular views need an index, and comments are now optional enrichment rather than a discovery requirement.


jubrad (approved):

doc/user/content/integrations/mcp-server/mcp-agent.md
kay-kim: More as an FYI   -- no need to patch as I'm changing locally and will push up a patch when finished with my changes ...
but, we don't support `SET ROLE <...>`.  Understandable if you thought we did because weirdly, for a long time,  the docs had a section with a detailed example where `SET ROLE` played a lead role and a note stating you can't do what this section and example suggested because we don't support it  😂😂😂  I got rid of it last summer.

PR #36251, opened 2026-04-24
adapter: Force migrations when upgrading from dev version

def-:
To prevent issues like https://github.com/MaterializeInc/database-issues/issues/11335 from confusing/blocking us in Staging/locally. See also discussion in https://materializeinc.slack.com/archives/C01LKF361MZ/p1777019597269559

Nightly: https://buildkite.com/materialize/nightly/builds/16206

### DIFF
--- a/src/adapter/src/catalog/open/builtin_schema_migration.rs
+++ b/src/adapter/src/catalog/open/builtin_schema_migration.rs
@@ -37,7 +37,8 @@ use futures::future::BoxFuture;
 use mz_build_info::{BuildInfo, DUMMY_BUILD_INFO};
 use mz_catalog::builtin::{
     BUILTIN_LOOKUP, Builtin, Fingerprint, MZ_CATALOG_RAW, MZ_CATALOG_RAW_DESCRIPTION,
-    MZ_STORAGE_USAGE_BY_SHARD_DESCRIPTION, RUNTIME_ALTERABLE_FINGERPRINT_SENTINEL,
+    MZ_STORAGE_USAGE_BY_SHARD, MZ_STORAGE_USAGE_BY_SHARD_DESCRIPTION,
+    RUNTIME_ALTERABLE_FINGERPRINT_SENTINEL,
 };
 use mz_catalog::config::BuiltinItemMigrationConfig;
 use mz_catalog::durable::objects::SystemObjectUniqueIdentifier;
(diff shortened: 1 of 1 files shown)


antiguru (approved):

src/adapter/src/catalog/open/builtin_schema_migration.rs
```diff
@@ -550,7 +562,8 @@ impl Migration {
             .filter(|(_, info)| {
                 use Builtin::*;
                 match info.builtin {
-                    Table(..) | MaterializedView(..) | ContinualTask(..) => true,
+                    Table(table) => **table != *MZ_STORAGE_USAGE_BY_SHARD,
```
ggevay, line 564: What is special about `MZ_STORAGE_USAGE_BY_SHARD`?

src/adapter/src/catalog/open/builtin_schema_migration.rs
```diff
@@ -419,7 +420,18 @@ impl Migration {
...
         self.validate_migration_steps(steps);
 
-        let (force, plan) = match self.config.force_migration.as_deref() {
+        // Version-based migration filter fails for dev versions, see for example
+        // https://github.com/MaterializeInc/database-issues/issues/11335
+        let force_migration = if self.source_version != self.target_version
+            && self.source_version.pre.as_str().starts_with("dev")
+            && self.config.force_migration.is_none()
+        {
+            Some("evolution".to_string())
```
ggevay, line 429: Why "evolution" and not "replacement"? I imagine dev version chaos to also sometimes cause such changes that can't be handled by schema evolution.

src/adapter/src/catalog/open/builtin_schema_migration.rs
def-: It's used for billing. We also exclude it in `src/adapter/src/coord.rs:2681-2695` so that we don't reset it. Not sure if it matters much for dev versions, but I assume we also have billing stats for us internally?

src/adapter/src/catalog/open/builtin_schema_migration.rs
def-: We'd wipe anything on every dev version update, which seems a bit extreme.

ggevay: Ok, let's hope that "evolution" will work. We can still make it "replacement" later if we run into this again.

src/adapter/src/catalog/open/builtin_schema_migration.rs
ggevay: Ok, makes sense! And it probably has random awkward changes less often than other things, even in dev versions.

But could you add a code comment please?

PR #36253, opened 2026-04-24
[mysql source versioning] add binlog setting to mysql source details at purification

patrickwwbutler:
Now adds a check at sql purification time to see what the binlog setting of the upstream mysql db is, and saves that setting in the mysql source details, so that we can check the original value during replication, and maintain a consistent decoding strategy


src/sql/src/pure.rs
```diff
@@ -1095,6 +1096,14 @@ async fn purify_create_source(
             let initial_gtid_set =
                 mz_mysql_util::query_sys_var(&mut conn, "global.gtid_executed").await?;
 
+            let binlog_full_metadata = version_compare::compare_to(
+                mz_mysql_util::query_sys_var(&mut conn, "version").await?,
+                "8.0.1",
+                version_compare::Cmp::Lt,
+            )
```
def-: This means we only enable `FULL` in versions < 8.0.1?

src/sql/src/pure.rs
```diff
@@ -1095,6 +1096,14 @@ async fn purify_create_source(
...
+            .expect("failed to parse version string from mysql")
```
def-: Should we really panic if we can't parse the MySQL version?

src/sql/src/pure.rs
```diff
@@ -1095,6 +1096,14 @@ async fn purify_create_source(
...
+                && mz_mysql_util::query_sys_var(&mut conn, "binlog_row_metadata").await? == "FULL";
```
def-: this one is not case insensitive

src/sql/src/pure.rs
```diff
@@ -1504,6 +1514,14 @@ async fn purify_alter_source_add_subsources(
...
```
def-: not case insensitive

src/sql/src/pure.rs
```diff
@@ -1858,6 +1877,30 @@ async fn purify_create_table_from_source(
...
+            {
+                Err(
+                    MySqlSourcePurificationError::UnsupportedBinlogMetadataSetting {
+                        setting: "Using MySQL version < 8.0.1, which does not support full binlog row metadata".to_string(),
+                    },
+                )?;
+            }
+            let binlog_metadata_setting =
+                mz_mysql_util::query_sys_var(&mut conn, "binlog_row_metadata").await?;
+            let binlog_full_metadata = binlog_metadata_setting.to_uppercase() == "FULL";
```
def-: this one IS case insensitive. Would be nice to have all the checks be consistent

def- (dismissed):

src/sql/src/pure.rs
patrickwwbutler: Hmm, it feels like a "canary in the coal mine" where if we can't get a valid version string from mysql then something is probably quite wrong, but this is in and of itself not an unrecoverable scenario, so we probably should not panic, and the panic here might obfuscate the underlying cause.

src/sql/src/pure.rs
patrickwwbutler: whoops, fixing that

src/sql/src/pure.rs
```diff
@@ -1858,6 +1883,30 @@ async fn purify_create_table_from_source(
...
+                mz_mysql_util::query_sys_var(&mut conn, "version").await?,
+                "8.0.1",
+                version_compare::Cmp::Lt,
+            )
+            .is_ok_and(|is_lt| is_lt)
+            {
+                Err(
+                    MySqlSourcePurificationError::UnsupportedBinlogMetadataSetting {
+                        setting: "Using MySQL version < 8.0.1, which does not support full binlog row metadata".to_string(),
+                    },
```
martykulma: This should just be a separate error variant - e.g.
`MySqlSourcePurificationError::UnsupportedVersion { version }`, or something to that effect.

src/sql/src/pure.rs
```diff
@@ -1858,6 +1883,30 @@ async fn purify_create_table_from_source(
...
+                )?;
+            }
+            let binlog_metadata_setting =
+                mz_mysql_util::query_sys_var(&mut conn, "binlog_row_metadata").await?;
+            let binlog_full_metadata = binlog_metadata_setting.to_uppercase() == "FULL";
```
martykulma: prefer `.eq_ignore_ascii_case("FULL")` over `to_uppercase`

src/sql/src/pure.rs
```diff
@@ -1504,6 +1517,17 @@ async fn purify_alter_source_add_subsources(
...
+            let binlog_full_metadata = version_compare::compare_to(
+                mz_mysql_util::query_sys_var(&mut conn, "version").await?,
+                "8.0.1",
+                version_compare::Cmp::Ge,
+            )
+            .is_ok_and(|is_ge| is_ge)
+                && mz_mysql_util::query_sys_var(&mut conn, "binlog_row_metadata")
+                    .await?
+                    .to_uppercase()
+                    == "FULL";
```
martykulma: Could combine the code into a single helper function that results `Result<(), MySqlSourcePurificationError>`



```
-- for boolean --
let binlog_full_metadata = ensure_binlog_full_metadata(&mut conn).await.is_ok();


-- for error --
ensure_binlog_full_metadata(&mut conn).await?;
// binlog_full_metadata is true if the above didn't fail
```

martykulma (commented):
Thanks @patrickwwbutler - some comments around tidying up.

src/sql/src/pure/error.rs
```diff
@@ -310,6 +310,14 @@ pub enum MySqlSourcePurificationError {
...
     #[error(transparent)]
     InvalidConnection(#[from] MySqlConnectionValidationError),
+    #[error(
+        "The MySQL system variable 'binlog_row_metadata' is set to an unsupported value: {setting}. Materialize requires this variable to be set to 'FULL' to use the \"CREATE TABLE FROM SOURCE\" syntax for MySQL sources."
+    )]
+    UnsupportedBinlogMetadataSetting { setting: String },
+    #[error(
+        "The MySQL version is unsupported for this operation. Materialize requires MySQL 8.0.1 or later to use the \"CREATE TABLE FROM SOURCE\" syntax for MySQL sources."
+    )]
+    UnsupportedMySqlVersion,
```
ublubu: I think it'd be nice to hang onto the MySQL version we actually saw, like you did with `UnsupportedBinlogMetadataSetting` above.

src/sql/src/pure/mysql.rs
```diff
@@ -519,16 +519,15 @@ pub(super) fn references_system_schemas(requested_references: &Option<ExternalRe
...
-        Err(MySqlSourcePurificationError::UnsupportedMySqlVersion)?;
+    let version = mz_mysql_util::query_sys_var(conn, "version")
+        .await
+        .map_err(|err| MySqlSourcePurificationError::InvalidConnection(err.into()))?;
+    if version_compare::compare_to(&version, "8.0.1", version_compare::Cmp::Lt).map_err(|_| {
+        MySqlSourcePurificationError::UnsupportedMySqlVersion {
+            version: version.clone(),
+        }
+    })? {
+        Err(MySqlSourcePurificationError::UnsupportedMySqlVersion { version })?;
```
martykulma, line 530: nit: an alternative if you want to condense the conditional and map_err
```
    version_compare::compare_to(&version, "8.0.1", version_compare::Cmp::Lt)
        .and_then(|res| if res { Err(()) } else { Ok(res) })
        .map_err(|_| MySqlSourcePurificationError::UnsupportedMySqlVersion { version })?;
```

martykulma (approved):
Looks good @patrickwwbutler !

PR #36256, opened 2026-04-25
sql_server: Fix snapshot_table_lock gauge leak on error

def-:
The `snapshot_table_lock_start` metric was incremented before a sequence of six fallible operations, but `snapshot_table_lock_end` was only called on the success path. If any intermediate `?` returned early, the gauge for that table stayed at 1 for the lifetime of the metric object, incorrectly reporting a held lock.

Wrap the fallible region in an async block so `snapshot_table_lock_end` is always called after it completes, regardless of success or failure.

Caused by https://github.com/MaterializeInc/materialize/pull/33962. Review with `-w`

Test run: https://buildkite.com/materialize/nightly/builds/16205


martykulma (approved):
thanks @def- !

PR #36257, opened 2026-04-25
sql-server: Fix composite key column ordering in constraint discovery

def-:
PR #33991 switched constraint discovery to INFORMATION_SCHEMA.CONSTRAINT_COLUMN_USAGE, which does not expose column ordinal position. For composite keys defined in non-alphabetical order (e.g. PRIMARY KEY (z_col, a_col)), SQL Server returns columns in arbitrary order, causing the persisted constraint SQL to record the wrong column order. This surfaces as non-deterministic constraint definitions across re-purification (ALTER SOURCE, refresh) and incorrect constraint text in mz_tables.create_sql.

Fix: use KEY_COLUMN_USAGE instead, which exposes ORDINAL_POSITION, and ORDER BY it to guarantee definition-order delivery.

### DIFF
--- a/src/sql-server-util/src/inspect.rs
+++ b/src/sql-server-util/src/inspect.rs
@@ -499,27 +499,25 @@ pub async fn get_constraints_for_tables(
         .map(|idx| format!("@P{}", idx))
         .join(", ");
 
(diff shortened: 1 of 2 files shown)


martykulma (approved):
nice, thanks @def-

PR #33784, opened 2025-10-06
Add PG source checks for row level security

martykulma:
Add checks to PG source purification to require BYPASSRLS if any of the tables has a row level security policy that affects the MZ user or public.

### Motivation

fixes https://github.com/MaterializeInc/database-issues/issues/9752

### Checklist

- [x] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #36261, opened 2026-04-25
postgres-util: fix RLS policy validation false positives (#33784)

def-:
Fix three issues in `validate_no_rls_policies` introduced in #33784:

- Filter by `polcmd IN ('r', '*')` so INSERT/UPDATE/DELETE-only policies don't spuriously require BYPASSRLS. Only SELECT and ALL policies affect what the replication user can read.
- Use `pg_has_role(CURRENT_USER, role_oid, 'USAGE')` instead of a direct OID comparison so policies applied via inherited roles are matched. PostgreSQL applies RLS policies transitively through role inheritance (via `has_privs_of_role`), so a policy `TO analyst` silently filters reads for any member of `analyst` — the exact silent-data-loss scenario the original check was written to prevent.
- Join `pg_namespace` for the raw `nspname` instead of casting through `regnamespace` (already quoted) into `%I` (quotes again), which produced double-quoted schema names in error messages.

Regression test added: creates a parent role `analyst` with a SELECT policy, grants it to `mz_user`, and verifies purification catches the inherited-role policy. Verified the test fails without the fix (policy silently missed, CREATE TABLE succeeds) and passes with it.

### DIFF
--- a/src/postgres-util/src/replication.rs
+++ b/src/postgres-util/src/replication.rs
@@ -107,8 +107,8 @@ pub async fn bypass_rls_attribute(client: &Client) -> Result<bool, PostgresError
 /// Returns an error if the tables identified by the oid's have RLS policies which
 /// affect the current user. Two checks are made:
 ///
-/// 1. Identify which tables, from the provided oid's, have RLS policies that affecct the user or
-///    public.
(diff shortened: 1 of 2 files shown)


martykulma (approved):
great find - thanks @def-

PR #36262, opened 2026-04-25
coord: Skip bootstrap credit check for expired DisableClusterCreation licenses

def-:
The bootstrap-time credit consumption validation introduced in #33654 returns ResourceExhaustion when an expired license with DisableClusterCreation behavior has any existing managed replicas, since max_credit_consumption_rate() returns 0.0 in that case. This crashes environmentd on restart, converting a soft-fail mode into a permanent lockout.

Skip the check when the license is expired with DisableClusterCreation so existing replicas survive restarts. New cluster creation is still blocked by the DDL-time validation.

### DIFF
--- a/src/adapter/src/coord.rs
+++ b/src/adapter/src/coord.rs
@@ -118,7 +118,7 @@ use mz_controller::clusters::{
 use mz_controller::{ControllerConfig, Readiness};
 use mz_controller_types::{ClusterId, ReplicaId, WatchSetId};
 use mz_expr::{MapFilterProject, MirRelationExpr, OptimizedMirRelationExpr, RowSetFinishing};
-use mz_license_keys::ValidatedLicenseKey;
+use mz_license_keys::{ExpirationBehavior, ValidatedLicenseKey};
 use mz_orchestrator::OfflineReason;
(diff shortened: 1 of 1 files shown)


src/adapter/src/coord.rs
```diff
@@ -2058,17 +2058,28 @@ impl Coordinator {
...
-                    .map_or_else(|| system_vars.max_credit_consumption_rate(), Numeric::from)
-            },
-            "cluster replica",
-            MAX_CREDIT_CONSUMPTION_RATE.name(),
-        )?;
+        // Skip the credit consumption check at bootstrap when the license is expired
+        // with DisableClusterCreation behavior. That mode is a soft-fail: it warns and
+        // prevents *new* cluster creation, but must not block startup with existing replicas.
+        // The Disable case is already handled by a bail! in main.rs before we reach here.
+        let enforce_credit_limit_at_bootstrap = !(self.license_key.expired
```
doy-materialize: i think we probably don't even need the expired check here - we should just skip this validation entirely when it's set to `DisableClusterCreation` since this codepath isn't related to cluster creation.

doy-materialize (approved):

PR #36264, opened 2026-04-25
docs: tweaks to the mcp developer endpoint docs

kay-kim:
preview: https://preview.materialize.com/materialize/36264/integrations/mcp-server/mcp-developer/

PR #36265, opened 2026-04-26
Update install-materialize-emulator.md

nstewart:
Pointing people to the developer mcp server as soon as they download the emulator


kay-kim (approved):
lgtm ... although, we probably want to hold off merging until https://github.com/MaterializeInc/materialize/pull/36264 (with the correct version +emulator  instructions) merges to avoid stranding people.

PR #36266, opened 2026-04-26
mz-debug: percent-encode credentials in Postgres connection URL

def-:
Credentials were interpolated directly into the URI string, so special characters (@, :, /, ?, #, %) in usernames or passwords could produce a malformed URL. Follow-up to #33129.


def-:
Ready for review

SangJunBak (approved):

PR #36269, opened 2026-04-26
copy-to-s3: reject parquet-incompatible types at planning time

DAlperin:
Iceberg and copy-to-s3 have started to diverge in their handling of the less standard datatypes. Whether the arrow-util crate is _capable_ of describing a datatype is no longer a great standin for _if_ that type should be written in a certain config. Adds an explicit check during planning time that the canonical Arrow types picked by the copy-to-s3 and iceberg implementations are allowed. 

To start, bans `DataType::Interval(IntervalUnit::MonthDayNano)` in response to
https://github.com/MaterializeInc/database-issues/issues/11329.

TODO(@DAlperin): go check the parquet crate and compare them against arrow types to maintain a fuller list of incompatibilities.

Fixes https://github.com/MaterializeInc/database-issues/issues/11329

### DIFF
--- a/src/arrow-util/src/builder.rs
+++ b/src/arrow-util/src/builder.rs
@@ -125,6 +125,50 @@ impl ArrowBuilder {
         Ok(())
     }
 
+    /// Helper to validate that a RelationDesc, after applying `overrides`, can
+    /// be encoded into Arrow AND converted from Arrow into parquet by
+    /// arrow-rs's `ArrowWriter`.
+    ///
+    /// Arrow encodability is a superset of parquet encodability — some Arrow
+    /// types are not (yet) implemented by arrow-rs's parquet writer. Callers
+    /// that write parquet must reject these up-front: copy-to-s3 writes an
+    /// `INCOMPLETE` sentinel during preflight and does not clean it up on a
+    /// failed upload, so a runtime failure inside the parquet writer leaves
+    /// the path in a state that blocks subsequent copies.
+    ///
+    /// Pass the same `overrides` you pass to [`desc_to_schema_with_overrides`]
+    /// — otherwise this check will reject types your sink is going to remap.
+    /// To add a new banned arrow datatype, extend `parquet_incompatible_type`.
+    pub fn validate_desc_for_parquet<F>(
+        desc: &RelationDesc,
+        overrides: F,
+    ) -> Result<(), anyhow::Error>
+    where
+        F: Fn(&SqlScalarType) -> Option<(DataType, String)>,
+    {
+        let mut errs = vec![];
+        for (col_name, col_type) in desc.iter() {
+            let dt =
+                match scalar_to_arrow_datatype_with_overrides(&col_type.scalar_type, &overrides) {
+                    Ok((dt, _)) => dt,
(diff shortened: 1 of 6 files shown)


src/sql/src/plan/statement/ddl.rs
```diff
@@ -3772,15 +3773,22 @@ fn plan_sink(
...
-            options,
-            relation_key_indices,
-            key_desc_and_indices,
-            commit_interval,
-        )?,
+        } => {
+            // Reject types that arrow-rs's parquet writer cannot handle, before
+            // sink creation. Pass the iceberg overrides so types iceberg
+            // remaps (e.g. interval -> string) don't trip the check.
+            ArrowBuilder::validate_desc_for_parquet(&desc, iceberg_type_overrides)
```
martykulma: Should this call live in iceberg_sink_builder instead of here?  It seems odd that there are a number of checks in the builder except for this one (including the enable_iceberg_sink flag - which we should probably remove!)

src/arrow-util/src/builder.rs
```diff
@@ -125,6 +125,47 @@ impl ArrowBuilder {
...
+                match scalar_to_arrow_datatype_with_overrides(&col_type.scalar_type, &overrides) {
+                    Ok((dt, _)) => dt,
+                    Err(_) => {
+                        errs.push(format!("{}: {:?}", col_name, col_type.scalar_type));
+                        continue;
+                    }
+                };
+            if parquet_incompatible_type(&dt).is_some() {
+                errs.push(format!("{}: {:?}", col_name, col_type.scalar_type));
+            }
```
martykulma: doesn't looks like we ever use the `str` returned by parquet_incompatible_type - is that intentional or oversight?

src/arrow-util/src/builder.rs
```diff
@@ -236,6 +277,34 @@ fn scalar_to_arrow_datatype(
...
+    }
+    match dt {
+        DataType::List(f) | DataType::LargeList(f) | DataType::FixedSizeList(f, _) => {
+            parquet_incompatible_type(f.data_type())
+        }
+        DataType::Map(f, _) => parquet_incompatible_type(f.data_type()),
+        DataType::Struct(fields) => fields
+            .iter()
+            .find_map(|f| parquet_incompatible_type(f.data_type())),
+        _ => None,
```
martykulma: Should this fail by default? I assume that currently we aren't using types like Dictionary or Union, so if they showed up, it should fail.

Did you opt for 2 match statements just to keep leaf_reason separated out?

martykulma (commented):
Thanks @DAlperin! A few minor comments

src/arrow-util/src/builder.rs
DAlperin: oversight

src/arrow-util/src/builder.rs
DAlperin: yeah, just for the visual seperation, I replaced it with an allowlist since you are right that feels better

martykulma (approved):
looks great, thanks @DAlperin !

PR #36289, opened 2026-04-27
sql_server: fix unbounded memory growth during snapshot

def-:
PR #33341 moved the RowArena allocation outside the snapshot loop, causing it to accumulate decoded data across all rows of all tables. Since RowArena is append-only and was never cleared, this leaked memory proportional to total snapshot size, risking OOM on large tables.

### DIFF
--- a/src/storage/src/source/sql_server/replication.rs
+++ b/src/storage/src/source/sql_server/replication.rs
@@ -250,7 +250,6 @@ pub(crate) fn render<'scope>(
                     SNAPSHOT_PROGRESS_REPORT_INTERVAL.handle(config.config.config_set());
                 let mut last_report = Instant::now();
                 let mut snapshot_lsns = BTreeMap::new();
-                let arena = RowArena::default();
 
                 for table in snapshot_tables {
                     // TODO(sql_server3): filter columns to only select columns required for Source.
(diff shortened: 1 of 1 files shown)


src/storage/src/source/sql_server/replication.rs
```diff
@@ -307,6 +306,7 @@ pub(crate) fn render<'scope>(
                         for (partition_idx, _) in partition_indexes {
                             // Decode the SQL Server row into an MZ one.
                             let mut mz_row = Row::default();
+                            let arena = RowArena::default();
```
antiguru: You can leave the arena outside, but call `clear` here instead.

src/repr/src/row.rs
```diff
@@ -3033,6 +3033,12 @@ impl RowArena {
         }
     }
 
+    /// Clears the arena, dropping all owned byte buffers but retaining
+    /// the allocated capacity of the outer vector.
+    pub fn clear(&self) {
```
antiguru: Did we remove the clear function? It needs to take &mut.

antiguru: I've changed my mind: let's just create a new arena every iteration.

def-: Yeah, there didn't seem to be a clear function. Will go back to the old state.

antiguru (approved):
Thank you, and sorry for the back and forth.

PR #36290, opened 2026-04-27
ore: fix tautological backoff assertions in retry tests

def-:
PR #33097 introduced `||` instead of `&&`, making the bounds check always pass. Restore the intended conjunction so the assertion actually validates that backoff is in (0, 10ms).


def-:
Ready for review

SangJunBak (approved):

PR #36292, opened 2026-04-27
[SSO] Add foundation for JWT group-to-role sync

mtabebe:
Add the building blocks for OIDC JWT group-to-role synchronization:

mz_jwt_sync builtin sentinel role that marks role memberships managed by JWT sync, distinguishing them from manual grants.

Three dyncfg system variables:
- oidc_group_role_sync_enabled (bool, default false): feature gate
- oidc_group_claim (string, default "groups"): JWT claim name
- oidc_group_role_sync_strict (bool, default false): fail-closed mode

- OidcClaims::groups() extraction method that normalizes group names (lowercase, deduplicate, sort) from JWT claims. Handles array-of-strings, single-string, and filters non-string values. Returns None for missing claims (skip sync) vs Some(vec![]) for empty claims (revoke all).

Addresses: SQL-176, SQL-177, SQL-178


test/sqllogictest/builtin_roles.slt
```diff
@@ -131,3 +131,86 @@ SELECT 1 FROM mz_internal.mz_statement_execution_history_redacted WHERE 1 = 0
...
+
+simple conn=mz_system,user=mz_system
+ALTER SYSTEM RESET oidc_group_claim
+----
+COMPLETE 0
+
+simple conn=mz_system,user=mz_system
+ALTER SYSTEM RESET oidc_group_role_sync_strict
+----
+COMPLETE 0
```
SangJunBak: I wonder how valuable these tests are and whether they belong in this test file?

SangJunBak: I'd imagine us extending these tests in `src/environmentd/tests/auth.rs` in which we're going to be testing this code path anyways

src/authenticator/src/oidc.rs
```diff
@@ -244,6 +244,59 @@ impl OidcClaims {
...
+    /// `Some(vec![])` if the claim is present but empty (revoke all sync-granted
+    /// roles), or `Some(vec![...])` with normalized (lowercased, deduplicated,
+    /// sorted) group names.
+    ///
+    /// Accepts arrays of strings, single strings, or mixed arrays (non-string
+    /// elements are filtered out). Other JSON types are treated as absent.
+    pub fn groups(&self, claim_name: &str) -> Option<Vec<String>> {
+        let value = self.unknown_claims.get(claim_name)?;
+
+        let raw_groups: Vec<String> = match value {
```
SangJunBak, line 260: Nit: I wonder if we're being too verbose with the comments? We say the same thing more concisely in the top level comment for each of these comments and the code already reads well

src/authenticator/src/oidc.rs
```diff
@@ -244,6 +244,59 @@ impl OidcClaims {
...
+            _ => return None,
+        };
+
+        // Normalize: lowercase for case-insensitive role matching,
+        // filter out empty strings (not valid role names), deduplicate
+        // via BTreeSet (which also sorts), then collect into a Vec for
+        // the caller. Sorted order makes downstream diff computation
+        // deterministic.
+        let normalized: Vec<String> = raw_groups
+            .into_iter()
```
SangJunBak: If we're already lowercasing, I wonder if it's worth trimming the whitespace from each end too?

src/authenticator/src/oidc.rs
```diff
@@ -244,6 +244,59 @@ impl OidcClaims {
...
+                if s.is_empty() {
+                    vec![]
+                } else {
+                    vec![s.clone()]
+                }
+            }
+            // Any other JSON type can't represent group names — treat as
+            // if the claim were absent so we don't accidentally revoke all
+            // roles based on garbage data.
+            _ => return None,
```
SangJunBak: Wonder if we should log `warn`s each time we expect a group claim to exist but it doesn't?

src/authenticator/src/oidc.rs
```diff
@@ -662,4 +715,266 @@ mod tests {
...
+        // consistent with the single-string case where "" → Some(vec![]).
+        let json = r#"{"sub":"user","iss":"issuer","exp":1234,"aud":"app","groups":["","eng",""]}"#;
+        let claims: OidcClaims = serde_json::from_str(json).unwrap();
+        assert_eq!(claims.groups("groups"), Some(vec!["eng".to_string()]));
+    }
+
+    #[mz_ore::test]
+    fn test_groups_whitespace_only_single_string() {
+        // Single-string claim that is whitespace-only: not lowercased into
+        // anything meaningful, not filtered (only empty strings are dropped).
```
SangJunBak: Re: the trim comment, maybe it makes sense to trim then filter for empty strings?

src/authenticator/src/oidc.rs
SangJunBak: Nevermind, I just reailzed SQL accepts whitespace in role names

src/authenticator/src/oidc.rs
SangJunBak: Nvm

src/authenticator/src/oidc.rs
```diff
@@ -662,4 +715,266 @@ mod tests {
...
+            Some(vec![
+                "org_data-science".to_string(),
+                "role/admin".to_string(),
+                "team-platform.eng".to_string(),
+            ])
+        );
+    }
+
+    #[mz_ore::test]
+    fn test_groups_case_insensitive_dedup() {
```
SangJunBak, line 888: nit: Feels like `test_groups_mixed_case` already handles this

SangJunBak (approved):
All just nits but otw looks good!

test/sqllogictest/builtin_roles.slt
mtabebe: Yeah I can just remove them

src/authenticator/src/oidc.rs
mtabebe: I think we can trim... it is probably sane to do that.

mtabebe:
Thanks for the comments!

mtabebe (commented):
thanks for the comments!

PR #36294, opened 2026-04-27
sql-server-util: Transaction guard includes BEGIN TRANSACTION

martykulma:
During creation of Transaction, possible race can occur where `BEGIN TRANSACTION` completes on upstream, transaction is cancelled, and a rollback doesn't occur on drop.

This should be extremely rare in practice. Noticed when reviewing @def- bugfixes.

### DIFF
--- a/src/sql-server-util/src/lib.rs
+++ b/src/sql-server-util/src/lib.rs
@@ -366,7 +366,14 @@ pub struct Transaction<'a> {
 
 impl<'a> Transaction<'a> {
     async fn new(client: &'a mut Client) -> Result<Self, SqlServerError> {
-        let results = client
+        // Construct the guard *before* awaiting BEGIN to avoid a potential race where
+        // transaction is started on remote, but the transaction is cancelled before returning.
+        let tx = Transaction {
+            client,
+            closed: false,
+        };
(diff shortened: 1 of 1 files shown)


def- (approved):

patrickwwbutler (approved):
Seems good to me. Is the fear that `Transaction::new()` would hang forever on `client.simple_query(...).await` because the upstream already had the transaction cancelled? 

the whole transaction concept with sql server feels a bit opaque in general, but reading through our usage of it in snapshotting it makes some sense. Interestingly we have zero usages of `commit`, so it seems this class is only ever rolled back, even in its usage in testdrive?

martykulma:
> Seems good to me. Is the fear that `Transaction::new()` would hang forever on `client.simple_query(...).await` because the upstream already had the transaction cancelled?
> 
> the whole transaction concept with sql server feels a bit opaque in general, but reading through our usage of it in snapshotting it makes some sense. Interestingly we have zero usages of `commit`, so it seems this class is only ever rolled back, even in its usage in testdrive?

The connection is borrowed, so we would have left it in a weird state (either the transaction started or it failed to start, but because we cancelled, we never read the response).

PR #36295, opened 2026-04-27
clarify comment about iceberg seen_rows tracking and how equality deletes only affect prior snapshots

ublubu:
from comment thread here: https://github.com/MaterializeInc/materialize/pull/36206/changes/BASE..50d64b04a32298442fcc97f0d7cc1cbb3a908f24#r3148322974

### DIFF
--- a/src/storage/src/sink/iceberg.rs
+++ b/src/storage/src/sink/iceberg.rs
@@ -332,7 +332,7 @@ impl EnvelopeHandler for UpsertEnvelopeHandler {
             // i.e. The DeltaWriter assumes that rows outside the "seen" cache come from prior snapshots.
             //
             // If we insert a row a=foo during this snapshot and then evict it from the "seen" cache,
-            // a subsequent update a=bar will lead to:
+            // a subsequent update a=bar (also during this snapshot) will lead to:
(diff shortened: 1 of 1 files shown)


src/storage/src/sink/iceberg.rs
```diff
@@ -321,16 +321,21 @@ impl EnvelopeHandler for UpsertEnvelopeHandler {
...
-        // committed snapshot.
         builder = if is_snapshot {
+            // Snapshot batches only produce inserts, so disable seen_rows tracking to save memory.
             builder.with_max_seen_rows(0)
         } else {
+            // For incremental batches, keep all "seen" rows. Do not evict any rows.
+            // The DeltaWriter issues an equality delete if we update (or delete) a row outside the "seen" cache.
+            // But equality deletes only apply to prior snapshots (lower sequence number).
+            //
+            // i.e. The DeltaWriter assumes that rows outside the "seen" cache come from prior snapshots.
```
ublubu, line 332: Should `builder.with_max_seen_rows` take a boolean arg instead?
It feels like there isn't a place for numbers between 0 and MAX.

DAlperin: Probably I wrote the api while laboring under the missaphrension that equality deletes could apply to the same snaphsot.

DAlperin (approved):

PR #33150, opened 2025-07-25
Short circuit at LIMIT+OFFSET in peek stash

ggevay:
Fixes https://github.com/MaterializeInc/database-issues/issues/9511 as discussed with @aljoscha here: https://materializeinc.slack.com/archives/C08JWHM837U/p1753437243688499?thread_ts=1753369592.747059&cid=C08JWHM837U

I've tested this by setting a low `PEEK_RESPONSE_STASH_THRESHOLD_BYTES` and running all slts like that. [Plus I think Nightly also tests with setting `PEEK_RESPONSE_STASH_THRESHOLD_BYTES` to 0.](https://github.com/MaterializeInc/materialize/blob/a048f88ed6cbd5601febe1a4ad23b40b8eef53e9/misc/python/materialize/mzcompose/__init__.py#L183) Nightly: https://buildkite.com/materialize/nightly/builds/12684

### Motivation

  * This PR fixes a recognized bug: https://github.com/MaterializeInc/database-issues/issues/9511

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #36299, opened 2026-04-28
peek stash: fix `LIMIT`+`OFFSET` short-circuit

def-:
The `break` introduced in #33150 only exits the inner `for` loop, so the outer loop keeps calling `recv()` and processing all remaining rows. Use a labeled `break 'outer` and drop the receiver to also stop the producer.


ggevay (approved):
Nice catch! Thank you!

PR #36306, opened 2026-04-28
Fix session based http auth allowed roles

alex-hunt-materialize:
### Motivation
Fixes https://github.com/MaterializeInc/database-issues/issues/11340

### Description

Fix session based http auth allowed roles.

Prevents obtaining a session token for roles not allowed, and prevents using a previously obtained (not sure how this would be possible) session token for a role that is disallowed.

### Verification

New http-auth test.

### DIFF
--- a/ci/test/pipeline.template.yml
+++ b/ci/test/pipeline.template.yml
@@ -645,6 +645,17 @@ steps:
     agents:
       queue: hetzner-aarch64-4cpu-8gb
 
+  - id: http-auth
+    label: HTTP auth
+    depends_on: build-aarch64
+    timeout_in_minutes: 30
+    inputs: [test/http-auth]
+    plugins:
+      - ./ci/plugins/mzcompose:
+          composition: http-auth
+    agents:
+      queue: hetzner-aarch64-4cpu-8gb
+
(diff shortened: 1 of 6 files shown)


alex-hunt-materialize:
@def- I adapted your https://github.com/MaterializeInc/materialize/pull/36293 change here. Your second test wasn't testing that they couldn't use the session token, since they shouldn't be able to get one anyway. I'm not sure of any way to test that, since it shouldn't be possible to get a session token now.

mtabebe (approved):
LGTM thanks for the test

ci/test/pipeline.template.yml
```diff
@@ -645,6 +645,17 @@ steps:
     agents:
       queue: hetzner-aarch64-4cpu-8gb
 
+  - id: http-auth
```
def-: This can probably live in Nightly, I don't expect breaking to be high risk.

def- (commented):
Thanks! Otherwise lgtm

ci/test/pipeline.template.yml
alex-hunt-materialize: moved

PR #36308, opened 2026-04-29
[SSO] Thread JWT groups and add sync diff logic

mtabebe:
Thread JWT group claims through the auth pipeline and add the core sync diff function:

- Add groups: Option<Vec<String>> to User, SessionConfig, ValidatedClaims, and AuthedUser so JWT group claims extracted during OIDC auth are available in handle_startup_inner. OIDC paths extract groups from ValidatedClaims; all non-OIDC paths pass groups: None.

- Add sync_jwt_groups_diff function that computes Op::GrantRole and Op::RevokeRole operations by diffing current role memberships against JWT group claims. Only manages memberships granted by MZ_JWT_SYNC_ROLE_ID — manually-granted roles are never touched.

No behavioral change yet as nothing calls the sync function.

SQL-179 and SQL-180

### DIFF
--- a/src/adapter/src/coord/group_sync.rs
+++ b/src/adapter/src/coord/group_sync.rs
@@ -1,0 +1,298 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file at the root of this repository.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! JWT group-to-role membership sync logic.
+//!
+//! This module computes the diff between a user's current role memberships
+//! and their JWT group claims, producing `Op::GrantRole` and `Op::RevokeRole`
+//! operations. Only memberships granted by the `MZ_JWT_SYNC_ROLE_ID` sentinel
+//! are managed; manually-granted memberships are never touched.
+
+use std::collections::{BTreeMap, BTreeSet};
+
+use mz_repr::role_id::RoleId;
+use mz_sql::session::user::MZ_JWT_SYNC_ROLE_ID;
+
+use crate::catalog::Op;
+
+/// Result of computing the group-to-role membership sync diff.
+#[derive(Debug, Clone)]
+#[allow(dead_code)] // used in tests; will be consumed by login path in a follow-up
+pub struct GroupSyncDiff {
+    /// Roles to grant to the user (with sentinel grantor).
+    pub grants: Vec<Op>,
+    /// Roles to revoke from the user (with sentinel grantor).
+    pub revokes: Vec<Op>,
+}
+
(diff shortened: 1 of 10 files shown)


src/adapter/src/coord/group_sync.rs
```diff
@@ -0,0 +1,296 @@
...
+}
+
+/// Computes the grant/revoke operations needed to sync a user's role
+/// memberships with their JWT group claims.
+///
+/// # Arguments
+/// - `member_id`: The user's role ID.
+/// - `current_membership`: The user's current `RoleMembership.map`
+///   (role_id → grantor_id).
+/// - `target_role_ids`: Role IDs resolved from the JWT group names.
```
SangJunBak: Might be worth adding `target_role_ids` will be lowercase

src/adapter/src/coord/group_sync.rs
```diff
@@ -0,0 +1,296 @@
...
+/// - Only roles granted by the JWT sync sentinel (`MZ_JWT_SYNC_ROLE_ID`)
+///   are managed by this function.
+/// - Manually-granted roles (grantor != sentinel) are never revoked.
+/// - If a target role is already manually granted, it is skipped — the
+///   manual grant takes precedence and we don't overwrite the grantor.
+#[allow(dead_code)] // used in tests; will be called from login path in a follow-up
+pub fn compute_group_sync_diff(
+    member_id: RoleId,
+    current_membership: &BTreeMap<RoleId, RoleId>,
+    target_role_ids: &BTreeSet<RoleId>,
```
SangJunBak, line 55: One part that's confusing to me is how we'll get from `String` to `RoleId`.

SangJunBak: Like will we just have to search the current membership for roles that match the strings in `groups` outside of this function?

SangJunBak (approved):
I did have a comment but I think the function and code makes sense!

src/adapter/src/coord/group_sync.rs
mtabebe: Yes, the caller (future PR) will have to look it up to get the role ID. Does that make sense to you?

SangJunBak: yup!

src/authenticator/src/oidc.rs
```diff
@@ -293,6 +295,9 @@ impl OidcClaims {
 #[derive(Zeroize, ZeroizeOnDrop)]
 pub struct ValidatedClaims {
     pub user: String,
+    /// Groups extracted from the JWT group claim. None if claim absent.
+    #[zeroize(skip)]
```
def-, line 299: Why do we not want to zeroize groups? CC @jasonhernandez

mtabebe: I was thinking group names are just identifiers so there was no need to wipe them.

But you are right, I'll zeroize them. (I'll just fix in my already open PR)

PR #36310, opened 2026-04-29
orchestrator-kubernetes: suppress minDomains when topology spread is soft

def-:
Kubernetes rejects `minDomains` on a `TopologySpreadConstraint` when `whenUnsatisfiable=ScheduleAnyway` [1]. Since #32695 unconditionally forwarded `cluster_topology_spread_min_domains` into the constraint, setting both `cluster_topology_spread_soft=on` and `cluster_topology_spread_min_domains` caused all replica pod creation to fail with a kube-apiserver admission error.

[1]: https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#spread-constraint-definition

### DIFF
--- a/src/orchestrator-kubernetes/src/lib.rs
+++ b/src/orchestrator-kubernetes/src/lib.rs
@@ -784,9 +784,21 @@ impl NamespacedOrchestrator for NamespacedKubernetesOrchestrator {
                     ..Default::default()
                 };
 
+                if config.soft && config.min_domains.is_some() {
+                    warn!(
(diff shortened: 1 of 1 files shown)


jubrad (approved):
Setting `minDomains` does provide guarantees about our ability to failover to different AZs multi-replica clusters. These guarantees are documented and advertised behavior. 

This seems like a fine way to protect against running to kube server errors, but it may be reasonable to throw a warning message here since this would be an invalid configuration to have our LD flags in.

PR #32668, opened 2025-06-05
adapter: Add `result_rows_first_to_last_byte_seconds` metric

ggevay:
Add histogram metric for time between first result row returned and last result row returned. (This is part of https://github.com/MaterializeInc/materialize/pull/32504, although it's not yet mentioned in the design doc.)

This is more involved than my other similar PRs from yesterday, because the control flow for returning results is a bit complex, with several variations in various situations, see tests. (I wanted to also add a test for a FETCH that has a timeout, but it seems FETCH's timeout is currently not working: https://github.com/MaterializeInc/database-issues/issues/9354 )

Nightly: https://buildkite.com/materialize/nightly/builds/12231

### Motivation

  * This PR adds a feature that has not yet been specified.

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #36311, opened 2026-04-29
pgwire: Fix first-to-last-byte metric observed multiple times on exhaused cursors

def-:
FETCHing past an exhausted cursor re-entered send_rows with no_more_rows already true, recording the histogram again with an ever-growing elapsed duration. Track that the metric has been recorded so it fires at most once per RecordFirstRowStream.

Introduced in [#32668](https://github.com/MaterializeInc/materialize/pull/32668).


ggevay (approved):
Ah yes, thank you!

PR #36312, opened 2026-04-29
controller: Fix replica-name annotation inconsistency across restarts

def-:
The bootstrap path in coord.rs passed the bare replica name while the runtime path in catalog_implications.rs passed the FQN. After an envd restart the annotation would flip from "cluster.replica" to "replica".

Move FQN construction into provision_replica so all callers get consistent "cluster.replica" annotations.

#33327 introduced this


def-:
Ready for review

jubrad (approved):

PR #36313, opened 2026-04-29
expr: Fix mutation-before-validation in `as_mut_temporal_filter`

def-:
`as_mut_temporal_filter` swapped operands via `std::mem::swap` before validating the operator, so an unsupported `BinaryFunc` would leave the expression with swapped operands but the original (un-flipped) operator, violating the documented "unchanged on Err" contract.

Move the swap after operator validation so the expression is only mutated when the full transformation succeeds.

Introduced in PR #33500.


def-:
Ready for review

ggevay (approved):
Thank you!

PR #36314, opened 2026-04-29
catalog: Fix pg_index.indnatts and pg_class.relnatts for index rows

def-:
indnatts was counting columns of the indexed table instead of the index itself. relnatts was always 0 for index rows because mz_columns has no entries keyed by index IDs. Follow-up to https://github.com/MaterializeInc/materialize/pull/33204


def-:
Ready for review

SangJunBak (approved):
Thanks for the change!

PR #36315, opened 2026-04-29
Projection pushdown to parquet

DAlperin:
When a source/MV reader's `RelationDesc` is narrower than the schema
that produced the part (typically because `compute_apply_column_demands`
pre-narrowed it), the persist source still decompresses and arrow-decodes
every column chunk from parquet before schema migration drops the
un-demanded columns. The bytes themselves arrive as part of the persist
blob get either way (persist does not range-read column chunks); what's
wasted is the decompression and decoding work.

Push column projection down into the parquet reader so un-demanded
sub-fields of `k_s` (the row struct on the key side of source/MV
shards) are never decompressed or arrow-decoded. The demand is the
set of stable indices in the read `RelationDesc`, so the mask matches
exactly what the downstream decoder expects. The mask handles the
`Result<Row, _>` envelope, where row sub-fields appear at
`k_s/ok/<idx>` rather than directly under `k_s`.

When the part's schema migration consists purely of dropping columns
under `k_s/ok` (checked via `Migration::pure_drops_under_source_data_ok`),
`parse_internal` overrides it to `SameSchema` because the
post-projection arrays already match the read shape. Migrations that
add fields, alter nullability, recurse into nested types, or do
Map -> List conversions cannot be replaced by projection alone; on
those parts we fall back to no projection and run the migration
normally.

The wins are decompression CPU and peak transient memory: un-demanded
column chunks are never materialized as arrow arrays, so per-part peak
allocation drops by the size of the dropped columns. There is no
network savings -- column chunks are read as part of the persist blob
get regardless; only their decompression and decoding is skipped.

Metrics: mz_persist_parquet_projection_applied_count,
mz_persist_parquet_projection_no_op_count, and the
projection_skipped_bytes_{compressed,uncompressed} counters track
projection outcomes.


src/persist-client/src/lib.rs
```diff
@@ -90,6 +90,20 @@ pub mod operators {
         The maximum amount of work to do in the persist_source mfp_and_decode \
         operator before yielding.",
     );
+
+    /// When enabled, the persist source pushes the MFP's column demand into
+    /// the value decoder so un-demanded columns are not materialized. This
+    /// is a CPU optimization; it does not affect query results, since
+    /// un-demanded columns surface as `Datum::Dummy` (which the MFP cannot
+    /// observe by definition of `demand`).
```
antiguru, line 98: At one point we had the goal of removing `Datum::Dummy` from Materialize because it causes all kinds of trouble. There's an alternative: Rewrite the MFP to project the columns it doesn't require.

src/persist-types/src/columnar.rs
```diff
@@ -135,10 +135,34 @@ pub trait Schema<T>: Debug + Send + Sync {
...
 }
 
+/// A shared, immutable set of input column indices that a decoder must materialize.
+///
+/// Carried alongside read-path operations to enable column projection pushdown:
+/// schemas that honor [`Schema::decoder_any_with_demand`] use this to skip decoding
+/// columns whose values won't be read downstream. The contained indices refer to
+/// the schema's input column ordering and are stored in ascending order with no
+/// duplicates so callers can use `binary_search` for membership tests.
+pub type ColumnDemand = Arc<Vec<usize>>;
```
antiguru, line 149: If you want a more compact representation:
```suggestion
pub type ColumnDemand = Arc<[usize]>;
```

DAlperin:
@claude review

src/persist-types/src/columnar.rs
```diff
@@ -139,6 +139,15 @@ pub trait Schema<T>: Debug + Send + Sync {
     fn encoder(&self) -> Result<Self::Encoder, anyhow::Error>;
 }
 
+/// A shared, immutable set of input column indices that a decoder must materialize.
+///
+/// Carried alongside read-path operations to enable parquet column projection
+/// pushdown: leaves under un-demanded stable indices are skipped before
+/// decompression and arrow decode. The contained indices refer to the schema's
+/// input column ordering and are stored in ascending order with no duplicates
```
mtabebe, line 147: If sorting is a key property this isn't enforced or communicated in the type.

I know there is a single place that is constructing this now, but I think we are losing that information, since callers explicitly call the binary_search.

Maybe it is overkill, but I would strengthen this type and make the constructor take the sorted input, and move the contains to be on the type so that it is hidden.

src/storage-operators/src/persist_source.rs
```diff
@@ -484,6 +488,33 @@ where
...
+    // narrowed (see `compute_apply_column_demands` in compute/render).
+    //
+    // When the read schema already covers every k_s sub-field present in the
+    // blob, the resulting mask is a no-op and parquet decode behaves as
+    // before. Gated by a dyncfg for safe rollout.
+    //
+    // `RelationDesc::iter_all` walks `metadata` (a `BTreeMap<ColumnIndex, _>`)
+    // in stable-index order, so collecting straight into a `Vec` preserves the
+    // sorted-ascending invariant `ColumnDemand` requires for `binary_search`.
+    let column_demand: Option<ColumnDemand> = if cfg.storage_source_enable_column_projection() {
```
mtabebe, line 507: Clean and helpful comment, I appreciate this 👍

src/persist/src/indexed/columnar/parquet.rs
```diff
@@ -141,11 +158,93 @@ pub fn encode_parquet_kvtd<W: Write + Send>(
...
+}
+
 /// Decodes [`BlobTraceUpdates`] from a reader, using [`arrow`].
+///
+/// `row_demand`, when present, must be sorted ascending with no duplicates.
 pub fn decode_parquet_file_kvtd(
     r: impl parquet::file::reader::ChunkReader + 'static,
     format_metadata: Option<&ProtoFormatMetadata>,
     metrics: &ColumnarMetrics,
+    row_demand: Option<&[usize]>,
```
mtabebe, line 247: Why did you lose the type here?

src/persist/src/indexed/columnar/parquet.rs
```diff
@@ -158,6 +257,28 @@ pub fn decode_parquet_file_kvtd(
...
+    // sub-fields of the row struct (`k_s` for source/MV shards) are not
+    // decompressed or materialized.
+    let builder = match row_demand {
+        Some(demand) => match build_row_projection_mask(builder.parquet_schema(), demand) {
+            Some(projection) => {
+                report_projection_metrics(metrics, builder.metadata(), &projection.excluded_leaves);
+                builder.with_projection(projection.mask)
+            }
+            None => {
+                // Demand was supplied but every k_s sub-field present in the
```
mtabebe, line 270: clever

src/persist/src/indexed/columnar/parquet.rs
```diff
@@ -141,11 +158,93 @@ pub fn encode_parquet_kvtd<W: Write + Send>(
...
+///
+/// `k_s` carries `SourceData = Result<Row, DataflowError>`, which is encoded
+/// as a struct with an `ok` branch (the row) and an `err` branch (the error).
+/// So Row sub-fields appear at `k_s/ok/<stable_idx>/...` rather than directly
+/// under `k_s/<stable_idx>/...`. The `err` branch is always retained because
+/// projection demand only describes row columns. Sub-field names under `ok`
+/// are the user column index as a string (e.g. `"0"`, `"1"`, ...), matching
+/// the layout produced by [`mz_persist_types::columnar::Schema`]
+/// implementations on `RelationDesc`.
+fn build_row_projection_mask(schema: &SchemaDescriptor, demand: &[usize]) -> Option<RowProjection> {
```
mtabebe, line 185: This function is the bulk of the logic for this change, deciding what to be masked and what not. I'd love to see some lower level unit tests of this

mtabebe: I guess there are some tests of the decode_trace_parquet_with_demand which calls this 2 layers deep, but this is sufficiently complex that I think it should be tested.

src/persist-client/src/fetch.rs
```diff
@@ -726,11 +736,94 @@ where
...
+        cfg: &FetchConfig,
+        row_demand: Option<&ColumnDemand>,
+    ) -> ShardSourcePart<K, V, T, D> {
+        // Parquet column projection drops un-demanded sub-fields of the row
+        // struct (`k_s`) before the parquet reader decompresses or
+        // arrow-decodes them. The demand is expressed in stable column
+        // indices that name the leaves under `k_s/ok` in the blob, which
+        // `apply_demand` preserves on the read schema, so the post-projection
+        // arrays always match the read shape regardless of whether the
+        // underlying part needs schema migration. The codec-only path (no
```
mtabebe, line 768: Do you have tests for this case? It wasn't obvious to me.

src/persist-client/src/fetch.rs
```diff
@@ -726,11 +736,94 @@ where
...
+        // `k_s`) and the row-validating decode format (which requires every
+        // sub-field present) opt out by passing `None` below.
+        let push_parquet_demand = matches!(self.structured_part_audit, PartDecodeFormat::Arrow);
+        let row_demand_for_arrow = if push_parquet_demand {
+            row_demand
+        } else {
+            None
+        };
+
+        // Decide whether projection pushdown can substitute for the part's
```
mtabebe, line 778: Just restating so I understand:

when schema migration occurs (adding/removing columns), we can't assume the masked columns refer to the same thing. It is only valid to keep the masking if the projection remains the same.

That makes sense. Would it be useful to add a metric to when we get rid of masking in these cases?

It feels like it should spike and then go back to 0 later, right?

mtabebe (commented):
Comments from me. Nothing too major though

PR #36319, opened 2026-04-29
fix x-materialize-user header bypassing allowed_roles

alex-hunt-materialize:
x_materialize_user_header_auth injected an AuthedUser whenever x-materialize-user was set to mz_system or mz_support, without checking the listener's allowed_roles. On an external HTTP listener with authenticator_kind=None and allowed_roles=Normal (the self-managed default), any client could assert x-materialize-user: mz_system and gain mz_system privileges. Run check_role_allowed inside the middleware so both the HTTP and WebSocket routers reject disallowed roles before the user is added to the request extensions.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/11346

### Verification
Additional nightly tests.

### DIFF
--- a/src/environmentd/src/http.rs
+++ b/src/environmentd/src/http.rs
@@ -282,9 +282,10 @@ impl HttpServer {
                     allowed_roles,
                 });
             if let listeners::AuthenticatorKind::None = authenticator_kind {
-                ws_router = ws_router.layer(middleware::from_fn(move |req, next| async move {
-                    x_materialize_user_header_auth(req, next, allowed_roles).await
-                }));
+                ws_router = ws_router.layer(middleware::from_fn_with_state(
+                    allowed_roles,
+                    x_materialize_user_header_auth,
+                ));
             }
             router = router.merge(ws_router);
         }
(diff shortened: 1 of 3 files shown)


src/environmentd/src/http.rs
```diff
@@ -282,7 +282,9 @@ impl HttpServer {
                     allowed_roles,
                 });
             if let listeners::AuthenticatorKind::None = authenticator_kind {
-                ws_router = ws_router.layer(middleware::from_fn(x_materialize_user_header_auth));
+                ws_router = ws_router.layer(middleware::from_fn(move |req, next| async move {
```
bobbyiliev: Axum has [from_fn_with_state](https://docs.rs/axum/latest/axum/middleware/fn.from_fn_with_state.html), do you think it will be a good fit here?

src/environmentd/src/http.rs
```diff
@@ -552,8 +554,9 @@ impl HttpServer {
                 router = router.merge(login_router).layer(session_layer);
             }
             listeners::AuthenticatorKind::None => {
-                base_router =
-                    base_router.layer(middleware::from_fn(x_materialize_user_header_auth));
+                base_router = base_router.layer(middleware::from_fn(move |req, next| async move {
```
bobbyiliev: Same as above.

bobbyiliev (approved):
I think that this looks good from what I can tell. Just 1 comment bellow.

src/environmentd/src/http.rs
alex-hunt-materialize: Yeah, that will be cleaner.

PR #36323, opened 2026-04-29
sql_server: Fix opening progress connection to SQL Server

def-:
#32221 introduced a missing return in the progress operator's non-responsible-worker guard. As a result, every Timely worker — not just the one designated for progress — opens a long-lived connection to the upstream SQL Server, continuously polls get_max_lsn, and runs sp_cdc_cleanup_change_table on every committed upper advancement. With a workers=N cluster, this multiplies upstream connections and cleanup operations by N instead of 1.


patrickwwbutler (approved):
Good catch, LGTM! failed signal is github failure, just microslop things

PR #36324, opened 2026-04-29
sql_server: Enforce external address check for direct connections

def-:
The Tunnel::Direct arm skipped resolve_address. Mirror the Postgres/MySQL pattern by validating the host before connecting.


martykulma (approved):
nice find - thanks @def- !

PR #36327, opened 2026-04-29
storage: fix get_active_replicas_for_object for sinks and top-level ingestions

def-:
Since #32066 sinks have per-replica scheduling on multi-replica clusters, but get_active_replicas_for_object still fell through to the "all replicas" branch for any non-ingestion-export ID. This caused drop_sinks_unvalidated to expect DroppedId from all replicas when only one was running the sink, leaking entries in dropped_objects indefinitely.

Add the missing branches for top-level ingestion IDs and export (sink) IDs to mirror the scheduling logic in update_scheduling/active_replicas.

### DIFF
--- a/src/storage-controller/src/instance.rs
+++ b/src/storage-controller/src/instance.rs
@@ -739,16 +739,15 @@ impl Instance {
     /// object (ingestion, ingestion export (aka. subsource), or export).
     pub fn get_active_replicas_for_object(&self, id: &GlobalId) -> BTreeSet<ReplicaId> {
         if let Some(ingestion_id) = self.ingestion_exports.get(id) {
-            // Right now, only ingestions can have per-replica scheduling decisions.
             match self.active_ingestions.get(ingestion_id) {
                 Some(ingestion) => ingestion.active_replicas.clone(),
-                None => {
-                    // The ingestion has already been compacted away (aka. stopped).
-                    BTreeSet::new()
-                }
+                None => BTreeSet::new(),
             }
+        } else if let Some(ingestion) = self.active_ingestions.get(id) {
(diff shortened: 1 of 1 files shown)


def-:
ready for review

martykulma (commented):
The `else` branch doesn't seem correct. This function is only called for Ingestion/Export and if the GlobalId hasn't matched the 3 previous branches, it means the id is not in the active maps - so it's either unknown or compacted.  It seems the else branch should return `BTreeSet::new()` in that case, otherwise this would also leak DroppedIds.

def-:
Done!

def-:
I have to investigate if this caused https://linear.app/materializeinc/issue/SS-116/thread-coordinator-panicked-at-srcstorage-controllersrclibrs218325

martykulma:
> I have to investigate if this caused https://linear.app/materializeinc/issue/SS-116/thread-coordinator-panicked-at-srcstorage-controllersrclibrs218325

wow - that caused a whole lot of havoc.  I obviously missed something along the way - I'll read through that again.

martykulma:
> > I have to investigate if this caused https://linear.app/materializeinc/issue/SS-116/thread-coordinator-panicked-at-srcstorage-controllersrclibrs218325
> 
> wow - that caused a whole lot of havoc. I obviously missed something along the way - I'll read through that again.

This might be because we set_hold_policies before capturing the active replicas, so we could race. I'm testing a change on a [separate branch](https://github.com/MaterializeInc/materialize/pull/36409) to see if that is the case.

def-:
Feel free to take it over! This is already beyond the level I feel comfortable fixing myself.

PR #36328, opened 2026-04-29
Add bounded staleness isolation level

antiguru:
Builds on #36906 (IsolationLevel string refactor) and #36907 (COPY FROM stdin sequencing), both merged.

Adds a new SQL isolation level `bounded staleness <duration>` that picks the freshest read timestamp `T` in `[oracle.read_ts(now) - D, largest_not_in_advance_of_upper]`, erroring (`SQLSTATE 40001`) when the window is empty. The error fires in the coordinator before any peek dispatch — the feasibility check is cluster-shape-independent.

Spec: `doc/developer/design/20260429_bounded_staleness_isolation.md`
User docs: `doc/user/content/reference/isolation-level.md`

Motivation: today strict serializable forces a wait up to one tick (~1s) for the freshest timestamp; serializable has no upper bound on staleness when an input lags. Bounded staleness gives users an explicit ceiling — "data no older than `D`" — without blocking on input frontiers, modeled on Spanner's bounded-staleness reads.

Anchor: bounded staleness always anchors against `oracle.read_ts` (one round-trip per query, the same one strict serializable already pays). The oracle is monotone, persisted, and shared, which makes it the only anchor that stays correct across crashes/restarts, clock changes, and a future multi-`environmentd` deployment. Caching the oracle read_ts under bounded staleness's relaxed contract is a possible future optimization.

Gating: feature flag `enable_bounded_staleness_isolation` (default on, public preview). Each rejection (writes, `real_time_recency = on`, non-`EpochMilliseconds` timelines) has a dedicated `AdapterError` variant.

Restrictions: writes (`INSERT`/`UPDATE`/`DELETE`/`COPY FROM`) and `real_time_recency = on` are rejected under bounded staleness; `D = 0` is rejected; `D` has no upper cap; timeline-less (constant) queries are allowed.

### DIFF
--- a/doc/developer/design/20260429_bounded_staleness_isolation.md
+++ b/doc/developer/design/20260429_bounded_staleness_isolation.md
@@ -1,0 +1,213 @@
+# Bounded staleness isolation
+
+* Associated: https://github.com/MaterializeInc/materialize/pull/36328
+
+## The problem
+
+Materialize today offers two practical isolation levels for SELECT traffic: serializable and strict serializable.
+
+* Strict serializable picks the freshest oracle-linearized timestamp across all queried objects.
+  This guarantees real-time recency but forces the query to wait until every input collection has caught up to that timestamp.
+  In practice this commonly costs ~1s of latency, governed by the timestamp tick rate, even when much fresher data already exists at a slightly lower timestamp.
+
+* Serializable picks the freshest timestamp available without waiting.
+  It avoids the wait but offers no upper bound on staleness.
+  When an input collection lags --- for example a slow source or a stalled materialized view --- the chosen timestamp can drift arbitrarily far behind the freshest known state of the system.
+
+Customers consistently want the third point on this curve: a bound on staleness rather than a guarantee of perfect freshness.
+A typical request is "give me data no older than five seconds, but do not block to get newer data."
+This is well known as bounded staleness in the literature and in production systems such as Spanner.
+
+## Success criteria
+
+A solution is successful when:
+
+1. A user can configure a session to read at a freshness ceiling of `D` (e.g. five seconds) and have queries served at any timestamp `T` satisfying `oracle.read_ts(now) - T <= D` without waiting on input frontiers.
+2. When the freshness ceiling cannot be met --- because some input collection lags by more than `D` --- the query returns a clear, actionable error that fires in the coordinator before any peek is dispatched, so the failure mode is independent of the cluster the query would have run on.
+3. The new isolation level is opt-in at two levels: the operator must enable a master feature flag, and individual sessions then opt in by `SET transaction_isolation`.
+4. The design preserves all existing strict-serializable correctness invariants documented in `guide-adapter.md`.
+
+## Out of scope
+
+* Bounded staleness for read-then-write transactions.
+  Writes still require a linearized oracle write timestamp, and mixing bounded staleness with writes raises questions about commit ordering that this design does not address.
+  An attempt to issue a write inside a bounded-staleness transaction errors.
+* A wait-with-timeout variant of the failure mode.
+  We expect to add a sibling behavior later, gated on a separate session variable, that waits up to a deadline before erroring.
+  This document specifies error-only behavior.
+* Bounded staleness on timelines other than `EpochMilliseconds`.
+  The freshness math is currently scoped to a single timeline.
+  Other timelines are out of scope until that property is re-derived for them.
+* Cross-session linearizability.
+  Two sessions running under bounded staleness may observe each other's writes at different timestamps within the bound.
+  This is the explicit trade-off and matches the existing serializable contract.
+* Interaction with `real_time_recency`.
+  The two settings are contradictory; combining them errors at session-variable validation time.
+
+## Solution proposal
+
+### High level
+
+Introduce an isolation level `bounded staleness <duration>` carrying an associated `Duration` payload.
+Under this isolation level, the timestamp selector picks the largest `T` in the closed interval `[oracle.read_ts(now) - D, largest_not_in_advance_of_upper]`.
+Each query makes one oracle round-trip to obtain the anchor (the same one strict serializable already pays); the oracle is the only anchor that stays correct across crashes, restarts, clock changes, and a future multi-`environmentd` deployment.
+
+If the interval is empty, the query errors with `BoundedStalenessExceeded` (`SQLSTATE 40001`).
+The whole feature is gated behind a master feature flag; until an operator enables it, `SET transaction_isolation = 'bounded staleness <D>'` errors at planning time.
+
+### Surface area
+
+In `src/sql/src/session/vars/value.rs`:
+
+```rust
+pub enum IsolationLevel {
+    ReadUncommitted,
+    ReadCommitted,
+    RepeatableRead,
+    Serializable,
+    StrongSessionSerializable,
+    StrictSerializable,
+    BoundedStaleness(Duration),
+}
+```
+
+`Duration` is `Copy`, so the enum stays `Copy`.
+
+The parser accepts `bounded staleness <duration>` where `<duration>` is a humantime-style spec (`5s`, `500ms`, `1m30s`); zero and negative durations are rejected, large bounds are accepted (they degrade to serializable-like behavior when input frontiers stay within `D`).
+For Prometheus labels we expose a unit-cardinality variant identifier separate from the user-facing rendering, so the duration does not blow up label cardinality.
+
+`TRANSACTION_ISOLATION` itself does not change shape; it already stores an `IsolationLevel`.
+
+### Feature flag
+
+`enable_bounded_staleness_isolation` (feature-flag table in `src/sql/src/session/vars/definitions.rs`), default off.
+`plan_set_variable` in `src/sql/src/plan/statement/scl.rs` parses the value being assigned; if it parses to `BoundedStaleness(_)`, the planner calls `scx.require_feature_flag(&vars::ENABLE_BOUNDED_STALENESS_ISOLATION)`, which errors when the flag is off and otherwise lets the SET proceed.
+
+### Timestamp selection
+
+In `src/adapter/src/coord/timestamp_selection.rs`:
+
+* `needs_linearized_read_ts` returns `true` for `BoundedStaleness(_)`, so `oracle_read_ts` is fetched on the query path (one oracle round-trip per query, same as strict serializable).
+* `determine_timestamp_via_constraints`, when the isolation level is `BoundedStaleness(D)`:
+  1. Reads `anchor = oracle_read_ts` (always populated by construction; an `expect` documents the invariant).
+  2. Pushes a lower-bound constraint `Antichain::from_elem(anchor - D)` with reason `IsolationLevel(BoundedStaleness(D))`.
+  3. Pushes a *hard upper-bound constraint* `Antichain::from_elem(largest_not_in_advance_of_upper)` with the same reason.
+
+The hard upper bound is essential.
+Without it, a candidate forced above the inputs' upper by the lower bound would be sent to compute and block waiting for the upper to advance, making the failure mode cluster-shape-dependent.
+With it, the existing feasibility check in the constraint solver fires `BoundedStalenessExceeded` deterministically in the coordinator before any peek dispatch.
+
+The within-session monotonicity machinery (the prior read timestamp acting as a lower bound) is unchanged and continues to apply to bounded-staleness reads.
+
+### Feasibility check and error path
+
+After candidate selection, if no `T` exists in `[lower, upper]`, emit `AdapterError::BoundedStalenessExceeded { bound, gap_ms, slowest_input }`.
+The `Display` impl produces:
+
+```
+ERROR: cannot serve query under bounded staleness {D}; freshest available
+       timestamp is {gap_ms}ms older than the bound[; slowest input: {id}]
+```
+
+`SQLSTATE`: `T_R_SERIALIZATION_FAILURE` (`40001`).
+This matches what application retry loops already handle.
+
+`gap_ms` is computed from the constraint solver's `lower_bound().into_option().map_or(0, u64::from) - upper_bound().into_option().map_or(0, u64::from)`.
+`slowest_input` is currently `None` --- determining the specific lagging input requires more bookkeeping than the constraint API exposes today; left for follow-up.
+
+### Freshness math
+
+Let `R(t)` denote the oracle's `read_ts` value at instant `t`.
+The user-visible contract under `BoundedStaleness(D)` is `R(now) - T <= D`.
+
+The oracle is monotone (`R(t_2) >= R(t_1)` for `t_2 >= t_1`) and persisted, so it survives crashes, restarts, and arbitrary clock changes.
+We fetch `R(now)` once at query start and pick `T >= R(now) - D`, which gives `R(now) - T <= D` exactly.
+This is correct on a single `environmentd` and remains correct when the system grows to multiple `environmentd`s sharing one authoritative oracle: every node sees the same `R`.
+
+We considered a wall-clock anchor (`T >= NowFn() - D`, no oracle round-trip) during initial design.
+It is unsafe under crashes/restarts (a fresh `NowFn()` after a restart can regress past previously-served timestamps if the wall clock has been adjusted), and degrades in a multi-`environmentd` deployment by the inter-node clock skew.
+We rejected it; see [Alternatives](#alternatives).
+
+### Interactions
+
+* `AS OF`: when `AS OF` is explicit, the user has chosen `T`; bounded staleness adds no constraint.
+* `real_time_recency`: rejected at session-variable validation time when both are set.
+* Mixed read/write transactions: `INSERT`, `UPDATE`, `DELETE`, and `COPY FROM` all reject under bounded staleness with a clear error.
+* `StrongSessionSerializable`: orthogonal; no interaction.
+* Within-session monotonicity: maintained the same way as today.
+* Cross-session: no linearizability guarantee. Two sessions can observe data at timestamps separated by up to `2D`.
+* Multi-timeline queries: rejected at planning time when the query touches a non-`EpochMilliseconds` timeline.
+
+### File map
+
+* `src/sql/src/session/vars/value.rs` --- enum variant, parser, formatter, `as_str`, `as_variant_str`, `valid_values`.
+* `src/sql/src/session/vars/definitions.rs` --- `enable_bounded_staleness_isolation` feature flag.
+* `src/sql/src/plan/statement/scl.rs` --- gate the SET on the feature flag.
+* `src/adapter/src/coord/timestamp_selection.rs` --- `needs_linearized_read_ts`, the bounded-staleness lower- and upper-bound constraints, the `BoundedStalenessExceeded` emit path.
+* `src/adapter/src/coord/sequencer/inner.rs` plus `src/adapter/src/coord/sequencer/inner/copy_from.rs` --- write-path rejection.
+* `src/adapter/src/error.rs` --- `BoundedStalenessExceeded` variant.
+* `doc/developer/guide-adapter.md` --- invariant on bounded staleness's interaction with the timestamp oracle.
+
+### Testing
+
+* SLT (`test/sqllogictest/bounded_staleness.slt`): parsing, formatting, round-tripping, rejection of zero/negative/out-of-range/unparseable durations, mutual-exclusion with `real_time_recency`, write rejection. Enables the master feature flag at the top.
+* Testdrive (`test/testdrive/bounded-staleness.td`): happy-path SELECT under `bounded staleness 5s`; deterministic error-path scenario (build an MV on a dedicated cluster, drop replication factor to 0, sleep past the bound, verify the error fires). Enables the master feature flag at the top.
+* The error path now bails in the coordinator before peek dispatch (because of the hard upper-bound constraint), so the test does not need a separate serving cluster.
+
+## Minimal viable prototype
+
+A working prototype implements:
+
+1. The enum variant, parser, and formatter, with no behavioral effect yet.
+2. The constraint logic in `determine_timestamp_via_constraints`, gated behind the new isolation level only.
+3. The error path, exercised against a synthetic stalled MV.
+4. The master feature flag.
+
+We can validate end to end with a customer-representative dashboard query against a continuously updating source, comparing tail latencies under strict serializable, serializable, and `bounded staleness 5s`.
+
+## Alternatives
+
+### Wall-clock anchor
+
+An earlier draft of this design picked `T >= NowFn() - D` to skip the per-query oracle round-trip.
+
+This is unsafe.
+On restart, `NowFn()` may regress (NTP step backward, container migration, etc.); a query running before the regress can have served `T` past the post-regress floor, breaking monotonicity.
+In a future multi-`environmentd` deployment the picture is worse: the shared oracle reflects the max clock across all nodes, so a slow-clock node serves `T` outside the `D` contract by the inter-node skew.
+
+The oracle is monotone, persisted, and shared; anchoring against it sidesteps both problems.
+The cost --- one oracle round-trip per query --- is the same one strict serializable already pays, and is small relative to the latency budget that motivated bounded staleness in the first place.
+A future optimization could cache `read_ts` and refresh it asynchronously, which is sound under bounded staleness's relaxed contract; we leave that for later.
+
+### Two session variables instead of a parameterised isolation level
+
+Add a unit variant `BoundedStaleness` plus a separate `max_staleness_ms` session variable.
+Smaller parser change, but introduces an inconsistency window and splits one concept across two settings.
+Rejected for SQL ergonomics.
+
+### Standalone `max_staleness_ms` overlay on serializable
+
+Keep the existing isolation levels and add a separate session variable that overlays a staleness ceiling.
+This conflates a contract change with a knob.
+Rejected for clarity.
+
+### Wait-with-timeout instead of error
+
+Pick `T` at the staleness floor and block on input frontiers up to `D`, then either return or error.
+This silently degrades to strict-serializable-with-cap behavior and breaks the freshness contract during the wait.
+Error-on-infeasible is the primary semantic; wait-with-timeout is a future, opt-in extension behind a separate session variable.
+
+### Caching `read_ts` for all isolation levels (the rejected `peek_read_ts_fast` path)
+
+Use a cached oracle value for *every* isolation level, not just bounded staleness.
+This was attempted and rejected for strict serializable in `guide-adapter.md`; the strict-serializable contract requires the timestamp to be determined during the query's real-time interval.
+Bounded staleness's relaxed contract makes oracle-read caching legitimate for *its* path; we defer that work but flag it as the natural follow-up if the per-query oracle round-trip becomes a bottleneck.
+
+## Open questions
+
+* `slowest_input` in the `BoundedStalenessExceeded` error is currently `None`.
+  Plumbing the per-input upper antichain into the constraint solver would give us a more actionable message.
+* Should bounded staleness be permitted inside a multi-statement read transaction?
+  The within-session monotonicity argument suggests yes, but the design has not been stress-tested for that case.
+* When (if ever) does the per-query oracle round-trip become a bottleneck worth caching?
+  A `timestamp_difference_for_bounded_staleness_ms` metric is already wired in to help spot the regime where caching would pay off.
--- a/doc/user/content/reference/isolation-level.md
+++ b/doc/user/content/reference/isolation-level.md
@@ -20,11 +20,21 @@ visible to a transaction during its execution.
 
 Materialize accepts the following isolation levels:
 
+{{< if-unreleased "v26.29" >}}
 | Isolation level | Behavior in Materialize |
 | --- | --- |
 | [**Strict Serializable**](#strict-serializable) | **Default.** Provides serializability and linearizability. |
 | [**Serializable**](#serializable) | Provides serializability but not linearizability. |
 | Read Uncommitted, Read Committed, Repeatable Read | Accepted for compatibility; treated as Serializable. |
+{{< /if-unreleased >}}
+{{< if-released "v26.29" >}}
+| Isolation level | Behavior in Materialize |
+| --- | --- |
+| [**Strict Serializable**](#strict-serializable) | **Default.** Provides serializability and linearizability. |
+| [**Serializable**](#serializable) | Provides serializability but not linearizability. |
+| [**Bounded Staleness `<duration>`**](#bounded-staleness) | **Public preview.** Serves reads at a timestamp at most `<duration>` stale; never blocks, errors if the bound cannot be met. |
+| Read Uncommitted, Read Committed, Repeatable Read | Accepted for compatibility; treated as Serializable. |
+{{< /if-released >}}
 
 ## Serializable
 
@@ -142,6 +152,99 @@ made available to us (e.g., querying PostgreSQL for the replication slot's LSN).
     Real-time recency queries return at least all data visible to Materialize
     when our client connection communicates with the external system.
 
+{{< if-released "v26.29" >}}
+## Bounded Staleness
+
+{{< public-preview />}}
+
+The Bounded Staleness isolation level lets you trade exact freshness for
+predictable latency. A query under bounded staleness is served at a timestamp
+that is at most `<duration>` stale—but never blocks waiting for input
+collections to catch up. If no timestamp within `<duration>` of "now" is
+available, the query errors immediately.
+
+This sits between [Serializable](#serializable) (no freshness ceiling) and
+[Strict Serializable](#strict-serializable) (always freshest, may wait for
+inputs to advance) on the freshness/latency spectrum. What distinguishes
+bounded staleness is that it never waits on input frontiers — it errors
+instead.
+
+### Syntax
+
+```mzsql
+SET TRANSACTION_ISOLATION TO 'bounded staleness <duration>';
+```
+
+`<duration>` is a duration string like `5s`, `500ms`, or `1m 30s` (compact
+forms like `1m30s` are also accepted). It must be greater than `0`.
+
+Bounded staleness is set through the `transaction_isolation` session variable,
+as shown above. Because it carries a duration, it cannot be set via the
+`BEGIN ... ISOLATION LEVEL` keyword form, which only accepts the standard
+isolation-level names.
+
+For example:
+
+```mzsql
+-- Read data no more than 5 seconds stale, never block.
+SET TRANSACTION_ISOLATION TO 'bounded staleness 5s';
+
+-- A tighter bound for dashboards.
+SET TRANSACTION_ISOLATION TO 'bounded staleness 250ms';
+```
+
+### Behavior
+
+When a query runs under bounded staleness, Materialize picks the freshest
+timestamp at which every queried collection has data, subject to the constraint
+that the timestamp is no more than `<duration>` stale relative to the current
+logical time. The query never blocks waiting for sources, materialized views,
+or indexes to advance. If even the freshest available timestamp is more than
+`<duration>` stale, the query errors with `SQLSTATE 40001`
+(`serialization_failure`):
+
+```
+ERROR: cannot serve query under bounded staleness 5s; freshest available
+       timestamp is 7000ms older than the bound
+```
+
+`40001` is a serialization failure; how to react is up to the application.
+Common responses include retrying, falling back to a different isolation level,
+or surfacing a "data unavailable" state.
+
+### Restrictions
+
+- **Read-only.** Writes (`INSERT`, `UPDATE`, `DELETE`, `COPY FROM`) are not
+  permitted under bounded staleness. The first write in a session running at
+  this isolation level errors.
+
+- **Mutually exclusive with `real_time_recency`.** Setting both errors at
+  session-variable validation time.
+
+- **Single timeline.** Bounded staleness only applies to queries on the
+  standard wall-clock timeline. Queries that touch other timelines error at
+  planning time.
+
+- **Bound must be greater than `0`.** A bound of zero is rejected; use
+  [Strict Serializable](#strict-serializable) if you need exact freshness.
+
+### When to use Bounded Staleness
+
+Pick bounded staleness when predictable latency matters more than perfect
+freshness, and "data is too stale" is a more useful signal than blocking.
+Common cases: dashboards, metric panels, alert evaluation.
+
+- Prefer [Strict Serializable](#strict-serializable) when you need
+  linearizable, end-to-end fresh reads, or when your application is willing to
+  wait for the freshest data.
+
+- Prefer [Serializable](#serializable) when you do not need a freshness bound
+  at all and want the simplest, lowest-latency reads.
+
+- Bounded staleness is read-only: if your session needs writes, use one of the
+  other isolation levels.
+{{< /if-released >}}
+
 ## Isolation levels and query latency
 
 Strict Serializable provides stronger consistency guarantees but may have slower
@@ -160,6 +263,12 @@ reads than Serializable.
   (but possibly slightly stale) snapshot, which avoids that latency at the cost
   of linearizability. However, if a consistent snapshot is not available, the
   query blocks until one becomes available.
+
+{{< if-released "v26.29" >}}
+- **Bounded Staleness** never waits for writes to propagate. It serves the
+  freshest consistent snapshot within the staleness bound, and errors
+  immediately when no such snapshot exists instead of blocking.
+{{< /if-released >}}
 
 ## Setting isolation level
 
--- a/src/adapter/src/coord/timestamp_selection.rs
+++ b/src/adapter/src/coord/timestamp_selection.rs
@@ -228,14 +228,18 @@ pub trait TimestampProvider {
     fn needs_linearized_read_ts(isolation_level: &IsolationLevel, when: &QueryWhen) -> bool {
         // When we're in the context of a timeline (assumption) and one of these
         // scenarios hold, we need to use a linearized read timestamp:
-        // - The isolation level is Strict Serializable and the `when` allows us to use the
-        //   the timestamp oracle (ex: queries with no AS OF).
-        // - The `when` requires us to use the timestamp oracle (ex: read-then-write queries).
+        // - The isolation level anchors against the oracle (Strict Serializable,
+        //   Strong Session Serializable, Bounded Staleness) and the `when` allows
+        //   us to use the timestamp oracle (ex: queries with no AS OF).
+        // - The `when` requires us to use the timestamp oracle (ex: read-then-write
+        //   queries).
         when.must_advance_to_timeline_ts()
             || (when.can_advance_to_timeline_ts()
                 && matches!(
                     isolation_level,
-                    IsolationLevel::StrictSerializable | IsolationLevel::StrongSessionSerializable
+                    IsolationLevel::StrictSerializable
+                        | IsolationLevel::StrongSessionSerializable
+                        | IsolationLevel::BoundedStaleness(_)
                 ))
     }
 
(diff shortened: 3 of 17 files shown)


antiguru:
### Bounded staleness vs. wall clock vs. timestamp oracle

The user-facing contract is `oracle_read_ts(now) - T <= D`.

We satisfy it via `T >= now_ms - D` plus the EpochMilliseconds invariant `oracle_read_ts(now) <= now_ms` (the oracle on this node is wall-clock-capped against this node's `NowFn`).
In single-`environmentd` deployments this is exact: the oracle and the lower-bound check share one clock.

### Horizontally scalable `environmentd`

If multiple nodes share an authoritative oracle, the argument weakens by the inter-node clock skew.
Concretely: node A advances the oracle to `t=10_000` at A's wall-clock 10_000; node B's wall-clock is 9_800 (200 ms slow); B picks `T = 9_800 - D`; real `R` is `10_000`; contract gap is `D + 200 ms`.
The invariant we documented in `guide-adapter.md` is *local-only* — \"oracle on this node is bounded by this node's wall clock.\"
The oracle's `read_ts` reflects the max clock across all nodes that ticked or applied writes, which can exceed any individual node's local `now_ms`.

### Does this change the assessment

For the current single-node topology, the design is sound.
For a future multi-node deployment, the contract degrades by the inter-node clock skew; NTP-synced clusters keep this in single-digit ms, but it is not zero.

Two clean fixes when we go multi-node:

1. Cache a recent oracle observation alongside `NowFn` and lower-bound by `max(cached_oracle, now_ms) - D`.
   Cache age is bounded, so the additional violation is bounded by cache age — which we control.
2. Loosen the documented contract to `oracle_read_ts(now) - T <= D + ε` and rely on time-sync infrastructure to keep `ε` small.

I lean (1) — explicitly anchoring on the oracle is more defensible than asking operators to keep clocks aligned.
The oracle-anchor cache mechanism we considered during design (and rejected as unnecessary for single-node) becomes load-bearing in the multi-node future.

Recommendation: merge as-is with the single-node assumption documented; revisit when horizontal scaling is on the roadmap.

aljoscha:
Before I dive deeper I have one high-level comment: We ultimately do want our system to work in a multi-adapter (multi-envd) world, and so the only viable options seems to be the timestamp-oracle anchor. Even right now, I'd say it's the only option that is correct when it comes to crashes/restarts/clock changes. So could we simplify the design to only have that anchor? Removes a bunch of complexity from both the design doc and the code and we don't have to worry about re-considering this when we go multi-node at some point in the future.

antiguru:
Good point — taking your suggestion. Dropped the wall-clock anchor and the dyncfg toggle entirely; bounded staleness now always anchors against `oracle.read_ts`. The simplification removes ~106 lines (no static `AtomicBool`, no refresh hook in `update_controller_config`, no `now: EpochMillis` plumbing through `determine_timestamp_for{,_inner}` / `determine_timestamp_via_constraints` / `frontend_determine_timestamp` / the datadriven test fixture, no parallel-workload + mzcompose registrations).

The wall-clock anchor isn't merely "less attractive in multi-node" — it's also unsafe today: a fresh `NowFn()` after a restart can regress (NTP step, container migration), letting a post-restart bounded-staleness query pick `T` past a previously-served floor and break the contract on a single node. The oracle is monotone and persisted, so it sidesteps that.

Cost is the per-query oracle round-trip (single-digit ms, same as strict serializable), which is small relative to the ~1s frontier wait that bounded staleness was designed to avoid. If we ever need to reclaim it, caching `read_ts` becomes legitimate under bounded staleness's relaxed contract — left as a future optimization with the metric `timestamp_difference_for_bounded_staleness_ms` in place to spot when it would pay off.

Pushed as `673c21b630` on top of a rebase against current `upstream/main`.

doc/user/content/get-started/isolation-level.md
```diff
@@ -16,7 +16,8 @@ The SQL standard defines four levels of transaction isolation. In order of least
 
 In Materialize, you can request any of these isolation
 levels, but they all behave the same as the Serializable isolation level. In addition to the four levels defined in the
-SQL Standard, Materialize also defines a [Strict Serializable](#strict-serializable) isolation level.
+SQL Standard, Materialize also defines a [Strict Serializable](#strict-serializable) isolation level and a
+[Bounded Staleness](#bounded-staleness) isolation level.
 
 Isolation level is a configuration parameter that can be set by the user on a session-by-session basis. The default isolation level is
 [Strict Serializable](#strict-serializable).
```
kay-kim:
```suggestion
Materialize accepts the following isolation levels:

- **SQL standard levels**:
  - **Read Uncommitted**
  - **Read Committed**
  - **Repeatable Read**
  - **Serializable**

  In Materialize, all four behave as **Serializable**.

- **Additional levels in Materialize**:
  - [**Strict Serializable**](#strict-serializable) (*Default*)
  - [**Bounded Staleness**](#bounded-staleness) (*Private preview*)

The isolation level is a configuration parameter that can be set on a
per-session basis.
```

doc/user/content/get-started/isolation-level.md
```diff
@@ -28,13 +29,14 @@ SET TRANSACTION_ISOLATION TO|= <isolation_level>
...
-| [Serializable](#serializable)               |
-| [Strict Serializable](#strict-serializable) |
+| Valid Isolation Levels                            |
+| ------------------------------------------------- |
+| [Read Uncommitted](#serializable)                 |
+| [Read Committed](#serializable)                   |
+| [Repeatable Read](#serializable)                  |
+| [Serializable](#serializable)                     |
+| [Strict Serializable](#strict-serializable)       |
+| [Bounded Staleness `<duration>`](#bounded-staleness) |
```
kay-kim:
```suggestion
| Valid Isolation Levels                               | Notes                         |
|------------------------------------------------------|-------------------------------|
| [Read Uncommitted](#serializable)                    | Behaves as **Serializable**   |
| [Read Committed](#serializable)                      | Behaves as **Serializable**   |
| [Repeatable Read](#serializable)                     | Behaves as **Serializable**   |
| [Serializable](#serializable)                        |                               |
| [Strict Serializable](#strict-serializable)          | *(default)*                   |
| [Bounded Staleness `<duration>`](#bounded-staleness) | *(private preview)*           |
```

kay-kim (commented):
Will continue in a bit ... but the intro section to this page has bothered me so long :D

doc/user/content/get-started/isolation-level.md
kay-kim: Oh wait ... I failed at suggestion 😞   ... it was supposed to be for the whole section. ugh.

kay-kim: Let me try again.  I'll delete this if I can successfully do it.

kay-kim: Oh .. apparently, I can't ... maybe because I'm commenting on a part that wasn't change.

antiguru: Applied both suggestions in e5249ba15d — intro restructured as bulleted list (SQL standard / Materialize-specific) and table gained a Notes column with the *behaves as Serializable* / *(default)* / *(private preview)* annotations.

aljoscha:
Here's a review from my clanker, you might want to ignore the nits if you feel like it. That first point though is important. In general, we should try as much as possible to not panic. Even for internal bugs where the instinct is to assert or panic. It's often better to instead return an internal error that bubbles up to the user. Panic'ing brings down the envd (or clusterd) for everyone.

# PR #36328 Review — Add bounded staleness isolation level

The design is solid, the new isolation level is well-bounded (gated, read-only, single-timeline), and the surrounding work (design doc, user doc, freshness math, monotonicity reasoning) is unusually thorough. One real panic, a few coverage gaps, and minor polish items.

## Blocking

**1. `AS OF` + bounded staleness panics the server.** `src/adapter/src/coord/timestamp_selection.rs:545`

```rust
let anchor = oracle_read_ts
    .expect("BoundedStaleness sets needs_linearized_read_ts; oracle_read_ts must be populated");
```

The `expect` is unsound. `needs_linearized_read_ts` returns `false` for `QueryWhen::AtTimestamp`/`AtLeastTimestamp` (both have `can_advance_to_timeline_ts() == false` and `must_advance_to_timeline_ts() == false`). The peek pipeline at `coord/sequencer/inner/peek.rs:409` then skips the oracle fetch and passes `oracle_read_ts = None`. With `BoundedStaleness(_)` set, the code reaches the `expect` and panics.

The design doc explicitly says "AS OF: when AS OF is explicit, the user has chosen T; bounded staleness adds no constraint." — that intent matches the fix:

```rust
if let IsolationLevel::BoundedStaleness(d) = isolation_level {
    if let Some(anchor) = oracle_read_ts {
        // existing lower/upper push
    }
}
```

No test covers this. Add an SLT case along the lines of `SET transaction_isolation = 'bounded staleness 5s'; SELECT * FROM tbl AS OF 0;`.

## Strong suggestions

**2. Sub-millisecond durations silently behave like `D = 0`.** `value.rs` rejects `0s`/`0ms`, but `bounded staleness 999us` parses fine; downstream `u64::try_from(d.as_millis()).unwrap_or(u64::MAX)` truncates to 0, so the lower bound is exactly the anchor — i.e. the very degenerate case the zero-rejection meant to forbid. Either reject `d < 1ms` in `parse`, or change the truncation to `as_millis().max(1)`. Either is fine; pick one and add a test.

**3. Test coverage gaps.** Missing:
- Feature flag gating (`SET transaction_isolation = 'bounded staleness 5s'` without `enable_bounded_staleness_isolation` should error).
- `AS OF` + bs (the panic above).
- `SUBSCRIBE` + bs (`subscribe.rs:339` also goes through the same `oracle_read_ts` path; same panic risk).
- `EXPLAIN TIMESTAMP` + bs (`explain_timestamp.rs:320`; same path).
- `COPY FROM` rejection (the variant exists but no `.slt`/`.td` exercises it).
- Non-`EpochMilliseconds` timeline rejection.
- Multi-component duration round-trip (`1m30s` formats as `1m 30s` per humantime; round-trip needs to survive the space).

**4. `BoundedStalenessExceeded` error message gets the gap wrong when `since` dominates.** `timestamp_selection.rs:457-462`:

```rust
let lower_ms = constraints.lower_bound().into_option().map_or(0, u64::from);
let upper_ms = constraints.upper_bound().into_option().map_or(0, u64::from);
let gap = lower_ms.saturating_sub(upper_ms);
```

`lower_bound()` is the join of *all* lower constraints, including `since` from the read holds. When the read-hold `since` is ahead of `anchor - D` (e.g. a collection's since has already advanced past the bs floor), the gap reflects the since gap, not the bs bound — but the message reads "freshest available timestamp is X ms older than the bound." Compute `gap` directly from `anchor - D` and `largest_not_in_advance_of_upper` so the message accurately describes the bs failure mode and isn't misleading when other constraints dominate.

**5. The `frontend_peek` / `coord` metric path is now duplicated for both strict-serializable and bounded-staleness.** `frontend_peek.rs:1577-1604` is "copy-pasted from Coordinator" by its own admission (line 1511). Adding the bs branch doubles the duplication. Not asking for the refactor here — but consider following up: extract a `record_isolation_difference_metric` helper that both paths call.

**6. Error message style.** `error.rs:1148-1166`. Materialize convention is short primary, optional detail/hint. Two of the new `Display` impls combine clauses with `;`:

```
"cannot serve query under bounded staleness 5s; freshest available timestamp is 7000ms older than the bound"
"bounded staleness isolation requires the EpochMilliseconds timeline; this query touches a different timeline"
```

These read better split into a primary + detail (the existing Materialize error infra supports this via `AdapterError::detail()` for some variants — e.g. `Reason::IsolationLevel`). At minimum, drop the `;` clauses; let the primary stay one factual sentence.

**7. Issue link in the design doc is `TBD`.** `doc/developer/design/20260429_bounded_staleness_isolation.md:3` says `Associated: TBD (open issue link to be added before merge)`. File the tracking issue and fill it in before merging.

## Nits

**8. The two rtr/bs checks in `inner.rs:1958-1976` collapse:**

```rust
if (name.as_str() == TRANSACTION_ISOLATION_VAR_NAME || name.as_str() == vars::REAL_TIME_RECENCY.name())
    && bounded_staleness
    && session.vars().real_time_recency()
{
    return Err(AdapterError::BoundedStalenessRealTimeRecencyConflict);
}
```

Same semantics, less duplication. The current form is correct (the staging/rollback model in `SessionVar::end_transaction` ensures the failed SET reverts properly, so the "post-set" placement isn't a bug).

**9. `Duration` is rendered with `{:?}` in the error.** `error.rs:1153` formats `bound` via `{:?}`, which prints `5s`, `1.5s`, etc. Use `humantime::format_duration` for consistency with the SET form (`bounded staleness 5s`). Minor.

**10. `valid_values()` now mixes literal values with a hint.** The vec contains `"read uncommitted"`, `"serializable"`, ..., and finally `"bounded staleness <duration>"`. The error rendering for `ConstrainedParameter` may say "expected one of: ..., bounded staleness <duration>" — readable, but `<duration>` looks like a literal. Consider whether the rendering needs adjustment, or split the hint into a separate `detail`.

**11. `as_str()` signature change from `&'static str` to `Cow<'static, str>`** is a real ergonomic regression for callers that compared its result. The existing `==` comparisons still work via `Deref`, but anything that wanted `&'static str` (e.g. for storing in static maps) would break. The new `as_variant_str()` covers the metric-label case. Consider keeping `as_str` returning `&'static str` and renaming the duration-bearing rendering to something like `to_display_string() -> String` or a `Display`-only form. Not blocking — current callsites compile — but the API shape is off.

**12. The duplicate `bounded_staleness` local binding is computed even on unrelated SETs.** Trivial overhead, just a code-organization quirk: the post-set block lives in the outer match arm rather than scoped to relevant variable names.

## What works well

- `as_variant_str()` for Prometheus labels — exactly the right move; without it the duration would explode label cardinality.
- The hard upper bound at `largest_not_in_advance_of_upper` is the key insight that lets the failure mode be cluster-shape-independent. The comment at `timestamp_selection.rs:351-356` explains it well.
- `IsolationLevel` stays `Copy` (Duration is `Copy`), preserving the existing usage pattern.
- The design doc's "Rejected alternatives" section walks through wall-clock anchoring, two-variable approaches, etc. with concrete reasons.
- Round-trip via `format` ↔ `parse` is tested as a unit test.
- Feature flag default-off + dedicated error variants for each rejection class is clean.

## Summary

Overall direction is good and the design is well-considered. **Do not merge until #1 (AS OF panic) is fixed and tested.** #2 and the testing gaps in #3 are worth addressing in this PR; the rest can be follow-ups.

antiguru:
Working through the review. Two commits pushed:

**`141e1e0328`** — blocking + most strong suggestions:

* **#1 (blocking)** AS OF + bounded staleness panic. Wrapped the bs constraint push in `if let Some(anchor) = oracle_read_ts`. `AS OF` skips `needs_linearized_read_ts` so the oracle is unset; the user-chosen `T` from `AS OF` is the sole timestamp constraint and bs adds nothing further. Same fix transitively covers SUBSCRIBE / EXPLAIN TIMESTAMP.
* **#2** Sub-millisecond bounds rejected in `IsolationLevel::parse`. `999us` truncated to `0ms` downstream and degenerated to "no staleness allowed".
* **#4** `BoundedStalenessExceeded.gap_ms` now derived directly from `anchor - D` and `largest_not_in_advance_of_upper`, not from `constraints.lower_bound()` / `upper_bound()` (which join *all* reasons).

**`2f602c2d96`** — polish:

* **#6** `BoundedStalenessExceeded` / `BoundedStalenessTimelineUnsupported` split into primary sentence + `AdapterError::detail()`.
* **#8 / #12** Collapsed the duplicate rtr/bs conflict checks in `sequence_set_variable`.
* **#9** `humantime::format_duration` for the bound (no more `{:?}`).
* **#11** Dropped `as_str() -> Cow<'static, str>`; kept `as_variant_str() -> &'static str` for label / equality use, Display for the user-facing form.
* **#7** `Associated:` now points at this PR.

**Test additions** (`bounded_staleness.slt`): AS OF regression, sub-ms parse, multi-component duration round-trip, feature-flag gating, `EXPLAIN TIMESTAMP` smoke, `COPY FROM` write rejection.

**Deferred follow-ups:**

* **#5** Extract `record_isolation_difference_metric` shared between `frontend_peek` and `coord` — you flagged this as not-blocking; the strict-vs-bs sign flip and the `read_holds.clone()` shape want their own pass.
* **#10** `<duration>` placeholder rendering in `valid_values()` — reads OK in context; not blocking.
* **#3 remainder** SUBSCRIBE + bs and non-`EpochMilliseconds` timeline SLT cases. The panic path is now closed at the source for any caller; SUBSCRIBE in SLT is awkward and a non-eml timeline is hard to construct deterministically.

PTAL.

def- (commented):
Do we just post Claude reviews now? ;) (normally I would try to reproduce them or at least manually understand before posting)

Mine finds very different things:

### HIGH — `oracle-anchor` mode is silently broken: the existing `oracle_read_ts` lower-bound completely shadows the BoundedStaleness slack

**Location:** `src/adapter/src/coord/timestamp_selection.rs:310-321` (existing)
combined with the new BoundedStaleness branch added at lines 367-393 of the
patched file (new code; near the previous "if let IsolationLevel::BoundedStaleness(d) = isolation_level" addition).

**What happens:**

When `bounded_staleness_use_oracle_anchor = true`, the new
`needs_linearized_read_ts` branch returns `true` for `BoundedStaleness(_)`, so
the caller fetches `oracle.read_ts()` and threads it into
`determine_timestamp_via_constraints` as `oracle_read_ts = Some(ts)`.

The pre-existing block at lines 310-321 then unconditionally pushes
`oracle_read_ts` as a **hard lower bound** for any isolation level except
`StrongSessionSerializable`:

```rust
if let Some(timestamp) = &oracle_read_ts {
    if isolation_level != &IsolationLevel::StrongSessionSerializable
        || when.must_advance_to_timeline_ts()
    {
        constraints.lower.push((
            Antichain::from_elem(*timestamp),
            Reason::IsolationLevel(*isolation_level),
        ));
    }
}
```

Bounded staleness is not excluded from this branch. Right after, the new
BoundedStaleness code adds a second lower bound:

```rust
let anchor = oracle_read_ts.unwrap_or_else(|| Timestamp::from(now));
let lower = anchor.saturating_sub(bound_ms);
constraints.lower.push((Antichain::from_elem(lower), ...));
```

Since `anchor == oracle_read_ts` here, the two pushed lower bounds are
`oracle_read_ts` and `oracle_read_ts - D`. The constraint solver takes the
**maximum** of all lower bounds, which is `oracle_read_ts`. The `- D` slack —
the entire point of bounded staleness — is shadowed.

**Net effect:** Under oracle-anchor mode, `BoundedStaleness(D)` behaves
exactly like strict-serializable: it forces `T >= oracle_read_ts`, can block
or fail when `upper < oracle_read_ts`, and never serves a stale-but-fresher-
than-D timestamp. The PR description's freshness math (`R(now) - T <= D`) does
not hold for the oracle-anchor branch as implemented; in fact the upper bound
constraint added by the new code makes this *more* likely to error than
strict-serializable, since strict-serializable can wait for `upper` to advance
while this code bails immediately.

The wall-clock-anchor branch (the default) is unaffected because
`needs_linearized_read_ts` returns `false` there, so `oracle_read_ts` is `None`
at the call site and the existing block is skipped.

The testdrive happy-path test only runs with the default (wall-clock) anchor,
so this never trips in CI. The dyncfg flip-flag in parallel-workload could
exercise it, but parallel-workload doesn't validate the freshness contract.

**Suggested fix:**

Either (a) exclude `BoundedStaleness(_)` from the existing
`oracle_read_ts → lower bound` push, e.g.:

```rust
if let Some(timestamp) = &oracle_read_ts {
    let push_as_lower_bound = match isolation_level {
        IsolationLevel::StrongSessionSerializable => when.must_advance_to_timeline_ts(),
        IsolationLevel::BoundedStaleness(_) => false,
        _ => true,
    };
    if push_as_lower_bound {
        constraints.lower.push((Antichain::from_elem(*timestamp), ...));
    }
}
```

…or (b) don't pass `oracle_read_ts` into the constraint function at all under
BoundedStaleness; thread the anchor in via a separate parameter so the two
concerns (linearization point vs. freshness anchor) don't share one channel.

(a) is the smaller change and matches the existing pattern.

### MEDIUM — Session-state inconsistency: `real_time_recency` × `BoundedStaleness` mutual-exclusion check runs *after* the variable is applied

**Location:** `src/adapter/src/coord/sequencer/inner.rs:1958-1981` (the new
checks added inside `sequence_set_variable`).

**What happens:**

The new mutual-exclusion checks are added inside `sequence_set_variable`
*after* the call to `vars.set(...)` at line 1907. The author's own comment
acknowledges this:

```rust
// Reject incompatible combinations of bounded staleness and
// real_time_recency after the variable has been applied.
```

If a user issues `SET transaction_isolation = 'bounded staleness 1s'` while
`real_time_recency = on`, the session variable is mutated to
`BoundedStaleness(1s)`, then the check fires and the SET command returns an
error. The session is now in a state where:

- the SET statement reported failure,
- `SHOW transaction_isolation` returns `"bounded staleness 1s"`,
- subsequent `SELECT`s on this session execute under the supposedly-rejected
  isolation level until the user resets it.

The same pattern applies to the symmetric check on
`SET real_time_recency = on` while bounded staleness is active.

This violates Postgres SET atomicity: a failed SET should not change the
session state.

**Suggested fix:**

Move both checks into `plan_set_variable` (`src/sql/src/plan/statement/scl.rs`)
where the value being assigned is parsed but not yet applied, alongside the
existing feature-flag gate. The planner already parses the candidate
`IsolationLevel`; it can reject incompatible combinations there before the
sequencer runs `vars.set`.

If planner-time rejection is awkward (because the *current* state of the
other variable matters and the planner does not have the session vars), at
least snapshot the previous value before `vars.set` and restore it on the
incompatibility error path.

The SLT in `test/sqllogictest/bounded_staleness.slt` does not check the
post-error session state and so does not surface this bug.

### LOW — `timestamp_difference_for_bounded_staleness_ms` baseline is wrong under oracle-anchor mode

**Location:** `src/adapter/src/coord/timestamp_selection.rs` (the new
`!det.respond_immediately() && BoundedStaleness && rtr.is_none()` block) and
the analogous block in `src/adapter/src/frontend_peek.rs`.

The metric computes the staleness gap as `serializable_ts - bs_ts` by calling
`determine_timestamp_for` with `IsolationLevel::Serializable` while passing
the same `oracle_read_ts` that was fetched for the BoundedStaleness call.
Because of the same line-310-321 behavior identified above, the
"serializable" baseline is actually forced to `oracle_read_ts` (i.e. strict-
serializable semantics), not the freshest-available timestamp Serializable
would normally pick. The metric therefore overstates the BS-vs-serializable
gap whenever oracle-anchor is enabled. Informational only; flagged here
because it falls out of the same root cause and will become obvious once the
HIGH bug is fixed.

### LOW — `IsolationLevel::as_str()` return type change ripples through string comparisons

**Location:** `src/adapter/src/coord/sequencer/inner.rs:1947-1956`,
`src/sql/src/session/vars.rs:872`.

`as_str()` now returns `Cow<'static, str>` instead of `&'static str`. The
existing call sites (`v == IsolationLevel::ReadUncommitted.as_str()` etc.,
where `v: String`) keep compiling because of std's `PartialEq<Cow<str>> for
String`, and `VarInput::Flat(&transaction_isolation.as_str())` works via
deref coercion from `&Cow<str>` to `&str`. No bug, but every future caller
that does `&'static str`-flavored things (slice patterns, `Copy` bounds,
struct fields) will now hit a less obvious type error. Minor latent risk.

antiguru:
Pushed `a45ce13d8b` addressing the HIGH and adding diagnostic coverage for the MEDIUM.

### HIGH — constraint shadowing on `oracle_read_ts`

Real bug, thanks for catching it. `needs_linearized_read_ts` now returns `true` for `BoundedStaleness(_)` after `0e050ee794` dropped the wall-clock anchor, so every BS query hits the existing `oracle_read_ts → hard lower bound` push at `timestamp_selection.rs:316-340`. Combined with the new clamp at `largest_not_in_advance_of_upper`, BS becomes *stricter* than strict serializable — strict can wait for the upper to advance, BS bails immediately as `BoundedStalenessExceeded`.

Took your fix (a) verbatim in shape, with `BoundedStaleness(_)` added to the carve-out branch alongside `StrongSessionSerializable` (which was already an exception for the analogous reason — its anchor is the session-local oracle, not the global one). The `must_advance_to_timeline_ts()` override is preserved for both, but BS rejects writes upstream so `FreshestTableWrite` is unreachable for it in practice.

Regression test added in `test/testdrive/bounded-staleness.td`: stop replication, sleep 1s, set `bounded staleness 1h`, query the MV. Before the fix this errors (`oracle.read_ts > upper`); after the fix it succeeds with a stale-but-fresh-enough timestamp. The existing 500ms-bound error scenario still passes — the gap math derivation in `141e1e0328` already used `anchor - D` directly, not `constraints.lower_bound()`, so the error message stays correct.

### MEDIUM — rtr × bs check post-\`vars.set()\`

Two reads on this. Aljoscha's clanker said the staging/rollback model in \`SessionVar::end_transaction\` handles a failed SET — which is plausible since \`vars.set\` mutates \`staged_value\` (not \`session_value\`), and \`end_transaction(Rollback)\` on the implicit txn clears it. Yours says the leak is observable. Without testing I genuinely don't know which is right.

Added a diagnostic SLT case in \`bounded_staleness.slt\` that asserts \`SHOW transaction_isolation\` and \`SHOW real_time_recency\` both return their un-mutated values after the conflicting SET errors. CI will tell us. If it passes, the post-set placement is fine and aljoscha's clanker was right; if it fails (def- right), I'll move the check into \`plan_set_variable\` per your suggestion — that's cleaner anyway, doesn't depend on a rollback path that's easy to break unrelatedly.

### LOW — metric baseline

Falls out of the HIGH fix. The Serializable baseline in \`record_isolation_difference_metric\` was being forced to \`oracle_read_ts\` by the same shadowing, so the gap was structurally 0. With BS carved out of the lower-bound push, both calls now exercise the path correctly. Worth noting that any data this metric collected pre-fix is meaningless; if anyone was watching it, the sample will only become useful after this lands.

doc/user/content/get-started/isolation-level.md
```diff
@@ -7,19 +7,22 @@ aliases:
...
-In Materialize, you can request any of these isolation
-levels, but they all behave the same as the Serializable isolation level. In addition to the four levels defined in the
-SQL Standard, Materialize also defines a [Strict Serializable](#strict-serializable) isolation level.
+    In Materialize, all four behave as **Serializable**.
 
-Isolation level is a configuration parameter that can be set by the user on a session-by-session basis. The default isolation level is
-[Strict Serializable](#strict-serializable).
+-   **Additional levels in Materialize**:
+    -   [**Strict Serializable**](#strict-serializable) (*Default*)
+    -   [**Bounded Staleness**](#bounded-staleness) (*Private preview*)
```
kay-kim: Oh ... just realized as I work on training and we ask people to install docs skills ... that
we'll probably want to enclose all our Bounded Staleness contents in `{{< if-released "v26.23" >}}` and `{{< /if-released >}}` or whichever version we think this will go.
You can wrap whole sections in those.  

I wasn't too fussed since this page is hidden from the toc (not sure why ... but, definitely in my backlog to surface this content in the Reference section) .. but, with AI and such, figured we should be better about when things surface in our docs.

doc/developer/design/20260429_bounded_staleness_isolation.md
```diff
@@ -0,0 +1,219 @@
...
+  Writes still require a linearized oracle write timestamp, and mixing bounded staleness with writes raises questions about commit ordering that this design does not address.
+  An attempt to issue a write inside a bounded-staleness transaction errors.
+* A wait-with-timeout variant of the failure mode.
+  We expect to add a sibling behavior later, gated on a separate session variable, that waits up to a deadline before erroring.
+  This document specifies error-only behavior.
+* Bounded staleness on timelines other than `EpochMilliseconds`.
+  The freshness math is currently scoped to a single timeline.
+  Other timelines are out of scope until that property is re-derived for them.
+* Cross-session linearizability.
+  Two sessions running under bounded staleness may observe each other's writes at different timestamps within the bound.
```
mtabebe, line 42: Can this be constrained by returning the timestamp that the query ran at and the next query having a bound at least as fresh as that timestamp?

mtabebe: That is, not just a duration but also an "as of at least T"  semantic?

(I think this is purely syntatic sugar for the bounded staleness but maybe it would require plumbing to return the timestamp)

doc/developer/design/20260429_bounded_staleness_isolation.md
```diff
@@ -0,0 +1,219 @@
...
+`Duration` is `Copy`, so the enum stays `Copy`.
+
+* `parse`: accept strings of the form `bounded staleness <duration>`, where `<duration>` is a humantime-style spec (`5s`, `500ms`, `1m30s`).
+  Reject `D = 0` and negative durations; large bounds are accepted (they just behave like serializable when input frontiers stay within `D`, which is fine).
+* `as_str` returns the user-facing rendition (`"bounded staleness 5s"`) as `Cow<'static, str>`; round-trips through `SHOW transaction_isolation` and `Value::parse`.
+* `as_variant_str` returns the unit-cardinality variant identifier (`"bounded staleness"`, no duration) as `&'static str`.
+  This is used for Prometheus labels; using `as_str` there would create one bucket per duration value and blow up label cardinality.
+* `valid_values` lists the literal pattern `"bounded staleness <duration>"` for the error string.
+
+The literal strings (`"bounded staleness"`, `"strict serializable"`, ...) live as associated `const`s on `IsolationLevel` so that `as_str`, `as_variant_str`, `valid_values`, and `parse` all reference one source of truth.
```
mtabebe: This isn't really important ... but is this really necessary in a design doc? feels quite granular, I realize it is probably useful to an LLM

doc/developer/design/20260429_bounded_staleness_isolation.md
mtabebe: Ahh I kept reading, there is an AS OF :)

doc/user/content/get-started/isolation-level.md
```diff
@@ -168,6 +176,63 @@ made available to us (e.g., querying PostgreSQL for the replication slot's LSN).
...
+When a query runs under bounded staleness, Materialize picks the freshest timestamp `T` at which every queried collection has data, subject to the constraint that `T` is no more than `D` stale relative to the current logical time.
+The query never blocks waiting for sources, materialized views, or indexes to advance.
+If even the freshest available timestamp is more than `D` stale, the query errors with `SQLSTATE 40001` (`serialization_failure`):
+
+```
+ERROR: cannot serve query under bounded staleness 5s; freshest available
+       timestamp is 7000ms older than the bound
+```
+
+Treat `40001` as a transient failure: retry once at the same isolation level, and if it persists fall back to `serializable` (no freshness bound) or surface a "data unavailable" state to the user.
```
mtabebe: Why retry once? Isn't this entirely up to the application to decide what to do?

mtabebe (commented):
(Leaving comments now on the docs, will get back to the code on Monday)

doc/developer/design/20260429_bounded_staleness_isolation.md
antiguru: Real follow-up, agreed. Two pieces of plumbing missing:

1. **Return the chosen `T` on the wire.** The coordinator picks `T`, executes, and discards it. There's no SQL-level surface for the client to learn what `T` was used. Adding `mz_now()` to `SELECT` projections gets close, but it's the *current* logical time, not the one the read actually happened at.
2. **`AS OF AT LEAST T` semantics.** The existing `AS OF T` *pins* the read to exactly `T`; we'd need a new variant that acts as a freshness floor (`T_chosen >= T`) instead. The constraint solver already supports lower bounds — it's the SQL surface and the planner that would need work.

Together those two get cross-session "no older than the last query I ran" — basically Spanner-style timestamp tokens. Out of scope for this PR; happy to file an issue when this lands so it doesn't get lost.

doc/developer/design/20260429_bounded_staleness_isolation.md
antiguru: Fair — trimmed in `9d6c286f38`. Kept the user-visible parse semantics (accepted forms, zero rejected, large bounds OK) and the label-cardinality consideration (the one bit of the impl that affects the design); dropped the `&'static str` / `Cow<'static, str>` / `const` plumbing.

doc/user/content/get-started/isolation-level.md
antiguru: You're right — application's call, not ours. Replaced in `9d6c286f38` with a neutral statement that 40001 is a serialization failure and listing common responses (retry, fall back, surface "data unavailable") without prescribing one.

doc/developer/design/20260429_bounded_staleness_isolation.md
mtabebe: I agree it is out of scope. Like you said I think it extends it to be more useful so I'd be in favour of adding it down the line.

src/adapter/src/coord/timestamp_selection.rs
```diff
@@ -228,14 +228,24 @@ pub trait TimestampProvider {
...
-        // - The `when` requires us to use the timestamp oracle (ex: read-then-write queries).
+        // - The isolation level requires anchoring against the oracle:
+        //   * Strict Serializable: timestamp == oracle.read_ts.
+        //   * Strong Session Serializable: timestamp == session-local oracle.
+        //   * Bounded Staleness: timestamp >= oracle.read_ts - D. The oracle is the
+        //     only anchor that stays correct across crashes/restarts and clock
+        //     changes, and is the only one that will work in a multi-`environmentd`
+        //     deployment.
+        //   …and the `when` allows us to use the timestamp oracle (ex: queries with
+        //   no AS OF).
```
aljoscha:
```suggestion
        //   * Bounded Staleness: timestamp >= oracle.read_ts - D …and the `when` allows us to use the timestamp oracle (ex: queries with no AS OF).
```

src/adapter/src/coord/timestamp_selection.rs
```diff
@@ -304,13 +314,24 @@ pub trait TimestampProvider {
...
+                // intended semantics. `must_advance_to_timeline_ts()` (only
+                // `FreshestTableWrite`) overrides the carve-out, but bounded
+                // staleness rejects writes upstream so that path is unreachable
+                // for it in practice.
+                let push_oracle_as_lower_bound = match isolation_level {
+                    IsolationLevel::StrongSessionSerializable
+                    | IsolationLevel::BoundedStaleness(_) => when.must_advance_to_timeline_ts(),
+                    _ => true,
+                };
+                if push_oracle_as_lower_bound {
```
aljoscha: For this whole block, I think this would make it clearer for me and wouldn't require the comments:
```
let hard_lower_bound = match isolation_level {
  StrongSession | BoundedStale => false,
  <explicit arms for all levels, so we are safe for future additions> => true,
}

if hard_lower_bound || when.must_advance_to_timeline_ts {
  contraints.push(...)
}
```

src/adapter/src/coord/timestamp_selection.rs
```diff
@@ -334,6 +355,36 @@ pub trait TimestampProvider {
...
+            // this, a candidate forced above the inputs' upper by the lower bound would
+            // be sent to compute and block waiting for the upper to advance — making the
+            // failure mode cluster-shape-dependent rather than a clean coordinator-side
+            // bail. With the upper bound in place the feasibility check fires
+            // `BoundedStalenessExceeded` deterministically in the coordinator.
+            //
+            // `oracle_read_ts` is `None` for `AS OF` queries (which skip
+            // `needs_linearized_read_ts`); the user has chosen `T` explicitly, so
+            // bounded staleness adds no further constraint beyond what `AS OF` itself
+            // imposes via `Reason::QueryAsOf`.
```
aljoscha: IMO, we can drop this whole comment block, it largely duplicates the code below

src/adapter/src/coord/timestamp_selection.rs
```diff
@@ -500,6 +583,17 @@ pub trait TimestampProvider {
             });
         }
 
+        // BoundedStaleness freshness math assumes the EpochMilliseconds timeline,
+        // where timestamps are wall-clock milliseconds. Reject queries whose
+        // timeline doesn't satisfy this contract. Timeline-less queries
+        // (`TimestampIndependent`, e.g. constant queries) are fine: the
+        // freshness contract is vacuous for them.
```
aljoscha:
```suggestion
        //BoundedStaleness freshness math assumes the EpochMilliseconds timeline,
        // where timestamps are wall-clock milliseconds.
```

src/adapter/src/coord/timestamp_selection.rs
```diff
@@ -412,6 +465,36 @@ pub trait TimestampProvider {
             if !constraints.lower_bound().less_equal(&candidate)
                 || constraints.upper_bound().less_than(&candidate)
             {
+                if let IsolationLevel::BoundedStaleness(d) = isolation_level {
```
aljoscha, line 463: I _think_ the rest of the levels try to only setup up contraints and then error reporting doesn't know about them. And this feels a bit long, but I don't have a better suggestion right now. 🙈

src/sql/src/session/vars/value.rs
```diff
@@ -916,14 +921,23 @@ impl IsolationLevel {
...
     /// equality checks against parsed input, and other contexts where one
     /// bucket per kind of isolation is wanted.
     ///
-    /// For the user-facing rendering (which round-trips through
-    /// [`Value::parse`]) use the [`fmt::Display`] impl — that is what
-    /// `SHOW transaction_isolation` returns.
+    /// For the user-facing rendering (which includes the duration on
+    /// `BoundedStaleness` and round-trips through [`Value::parse`]) use the
+    /// [`fmt::Display`] impl — that is what `SHOW transaction_isolation`
+    /// returns.
```
aljoscha: I think these additions don't add much value

aljoscha (commented):
Looks very good! I had a couple of inline comments. These are largely around clarifying some things and cutting down the comment bloat that Claude likes to add. 😅

doc/user/content/reference/isolation-level.md
```diff
@@ -142,6 +143,90 @@ made available to us (e.g., querying PostgreSQL for the replication slot's LSN).
...
+
+{{< public-preview />}}
+
+The Bounded Staleness isolation level lets you trade exact freshness for
+predictable latency. A query under bounded staleness is served at a timestamp
+that is at most `<duration>` stale—but never blocks waiting for input
+collections to catch up. If no timestamp within `<duration>` of "now" is
+available, the query errors immediately.
+
+This sits between [Serializable](#serializable) (no freshness bound, never
```
kay-kim: So ... we do state in lines 244-247 that Serializable can block if a consistent snapshot isn't available 
- **Serializable** does not wait for writes to propagate. It reads a consistent
  (but possibly slightly stale) snapshot, which avoids that latency at the cost
  of linearizability. However, if a consistent snapshot is not available, the
  query blocks until one becomes available.

doc/user/content/reference/isolation-level.md
```diff
@@ -142,6 +143,90 @@ made available to us (e.g., querying PostgreSQL for the replication slot's LSN).
...
+```mzsql
+SET TRANSACTION_ISOLATION TO 'bounded staleness <duration>';
+```
+
+`<duration>` is a duration string like `5s`, `500ms`, or `1m30s`. It must be
+greater than `0`.
+
+For example:
+
+```mzsql
```
kay-kim, line 188: if only available for reads and errors on writes, I wonder if it makes sense to show examples for setting at the transaction level instead of a session level?

src/adapter/src/coord/timestamp_selection.rs
antiguru: Condensed in `e3b2a34a0b` — collapsed the three per-level bullets into one ("the isolation level anchors against the oracle (Strict Serializable, Strong Session Serializable, Bounded Staleness)"); the multi-`environmentd` rationale already lives in `guide-adapter.md`.

src/adapter/src/coord/timestamp_selection.rs
antiguru: Done in `e3b2a34a0b`. Now a `hard_lower_bound` match with explicit arms for every level (no `_ =>`, so a future variant forces a decision), and `if hard_lower_bound || when.must_advance_to_timeline_ts()`. Dropped the long comment in favor of the self-describing match.

src/adapter/src/coord/timestamp_selection.rs
antiguru: Trimmed in `e3b2a34a0b`. Kept only the non-obvious bit — why the upper clamp at `largest_not_in_advance_of_upper` exists (fail-fast in the coordinator vs. cluster-shape-dependent block) — and the `AS OF`/`None` note. Dropped the rest that restated the code.

src/adapter/src/coord/timestamp_selection.rs
antiguru: Applied in `e3b2a34a0b`, kept the one-line note that timeline-less (constant) queries pass the check since that's the non-obvious case the comment guards.

src/adapter/src/coord/timestamp_selection.rs
antiguru: You're right that it breaks the "levels only set up constraints, the error path is generic" pattern. I kept it but tightened it in `e3b2a34a0b` (merged the two comment blocks, dropped the intermediate locals). The reason it can't be fully generic: `constraints.{lower,upper}_bound()` join *all* reasons, so when a non-bs constraint dominates the gap would misdescribe the failure — we need `anchor - D` vs the inputs' upper specifically. The clean fix is to carry the dominating `Reason` out of the solver so the generic error can name it; that's a bigger change I'd rather do as a follow-up than bolt on here.

src/sql/src/session/vars/value.rs
antiguru: Trimmed the doc comment in `281e06addf` down to two lines (purpose + "duration omitted, use Display"). Kept `is_bounded_staleness` since it's used at several call sites.

doc/user/content/reference/isolation-level.md
```diff
@@ -142,6 +143,90 @@ made available to us (e.g., querying PostgreSQL for the replication slot's LSN).
...
+blocks) and [Strict Serializable](#strict-serializable) (perfectly fresh, may
+block) on the freshness/latency spectrum.
+
+### Syntax
+
+```mzsql
+SET TRANSACTION_ISOLATION TO 'bounded staleness <duration>';
+```
+
+`<duration>` is a duration string like `5s`, `500ms`, or `1m30s`. It must be
```
kay-kim: also, should this be `1m 30s` (space)

doc/user/content/get-started/isolation-level.md
antiguru: Done in `5ff1f9cb7e`, gated on `{{< if-released "v26.29" >}}` (the docs also moved to `reference/isolation-level.md` since upstream relocated the page). All three bounded-staleness touchpoints are gated: the `## Bounded Staleness` section, the "isolation levels and query latency" bullet, and the supported-levels table row.

The table row needed `if-unreleased`/`if-released` twin tables rather than an inline wrap — a block shortcode around a single `| … |` row breaks the table (the shortcode re-renders its inner via `RenderString`, so a lone row isn't parsed as a table). Verified both ways with a local `hugo` build: unreleased → no bounded-staleness content and the table stays aligned; with `v26.29` flipped to `released: true` → section, anchor, bullet, and row all render.

doc/user/content/reference/isolation-level.md
antiguru: Good catch — "never blocks" overstated Serializable, which (per the latency section) can block until a consistent snapshot is available. Reworded in `2ed7c880ad`: now "between Serializable (no freshness ceiling) and Strict Serializable (always freshest, may wait for inputs to advance)", and made the distinguishing property explicit — bounded staleness never waits on input frontiers, it errors instead.

doc/user/content/reference/isolation-level.md
antiguru: It can't be set at the transaction level: `BEGIN ... ISOLATION LEVEL` (and `SET TRANSACTION ISOLATION LEVEL`) parse a fixed set of keyword levels in the grammar (`parser.rs:8629`), none of which carry a duration, so `BEGIN ISOLATION LEVEL bounded staleness 5s` doesn't parse. The only way to select it is the `transaction_isolation` session variable with the string value. Added a note to the Syntax section in `2ed7c880ad` stating exactly that, so users don't reach for the keyword form. (Scoping it to a read-only block is still possible the usual way: `SET transaction_isolation = 'bounded staleness 5s'` then `BEGIN; … SELECT …; COMMIT`.)

doc/user/content/reference/isolation-level.md
antiguru: Both parse — `humantime` accepts `1m30s` and `1m 30s` — but the canonical form (what `SHOW transaction_isolation` returns) is the spaced `1m 30s`. Switched the doc to `1m 30s` in `2ed7c880ad` and noted compact forms are also accepted, so the example matches what users see echoed back.

aljoscha (approved):

antiguru:
Thanks for the reviews!

PR #36329, opened 2026-04-29
expr: regexp_replace STRICT on NULL replacement with bad pattern

antiguru:
Regression from #32053. When `regexp_replace` is called with a literal pattern that fails to compile, the reduce-path guard checked only the source for NULL. PostgreSQL's `regexp_replace` is STRICT, so a NULL replacement column must also yield NULL rather than surface the regex error.

```sql
CREATE TABLE e (replacement text);
INSERT INTO e VALUES (NULL);
SELECT regexp_replace('abc', 'b(..', replacement, 'ig') FROM e;
-- before: ERROR: invalid regular expression
-- after:  NULL
```

The fix OR-s the source and replacement IS NULL checks before the if-then-else; pattern and flags are literals on this branch so source and replacement are the only runtime args to guard.

Fixes https://github.com/MaterializeInc/database-issues/issues/11350

### DIFF
--- a/src/expr/src/scalar.rs
+++ b/src/expr/src/scalar.rs
@@ -1272,8 +1272,8 @@ impl MirScalarExpr {
                             // The behavior of `regexp_replace` is that if the data is `NULL`, the
                             // function returns `NULL`, independently of whether the pattern or
                             // flags are correct. We need to check for this case and introduce an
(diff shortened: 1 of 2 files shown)


def- (approved):
Test lgtm

ggevay (approved):
LGTM

PR #36330, opened 2026-04-29
sql_server: Improve direct connection

def-:
Follow-up to https://github.com/MaterializeInc/materialize/pull/36324 to only use the resolved address. I couldn't find a cleaner way (and only realized this problem after merging the original PR, sorry!)

Test run: https://buildkite.com/materialize/nightly/builds/16270

### DIFF
--- a/src/sql-server-util/src/config.rs
+++ b/src/sql-server-util/src/config.rs
@@ -8,6 +8,8 @@
 // by the Apache License, Version 2.0.
 
 use anyhow::Context;
+use std::net::SocketAddr;
+
 use mz_ore::future::InTask;
 use mz_repr::CatalogItemId;
 use mz_ssh_util::tunnel::{SshTimeoutConfig, SshTunnelConfig};
@@ -48,7 +50,9 @@ impl Config {
         let inner = tiberius::Config::from_ado_string(s).context("tiberius config")?;
         Ok(Config {
             inner,
-            tunnel: TunnelConfig::Direct,
+            tunnel: TunnelConfig::Direct {
+                resolved_addresses: Default::default(),
+            },
             in_task: InTask::No,
         })
     }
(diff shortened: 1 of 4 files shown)


src/sql-server-util/src/config.rs
```diff
@@ -59,8 +61,9 @@ impl Config {
 /// TODO(sql_server2): De-duplicate this with MySQL and Postgres sources.
 #[derive(Debug, Clone)]
 pub enum TunnelConfig {
-    /// No tunnelling.
-    Direct,
+    Direct {
+        resolved_addresses: Option<Vec<SocketAddr>>,
```
martykulma: This should never be updated once resolved, so perhaps this could just be a `Box<[SocketAddr]>`.

`Option` doesn't really add anything - this would be the same as if the vec were empty (resolve_address returns error on empty set).

The comment should indicate that providing addresses will use the addresses provided, while using the default value will fallback to DNS lookup of the host.

src/sql-server-util/src/config.rs
```diff
@@ -84,6 +87,14 @@ pub enum TunnelConfig {
...
 }
 
+impl TunnelConfig {
+    pub fn direct() -> Self {
+        Self::Direct {
+            resolved_addresses: None,
+        }
+    }
+}
+
```
martykulma: this is only necessary in the example code, I'd say drop this and change the example to 
```
let tunnel_cfg = TunnelConfig::Direct { resolved_addresses: Default::default() };
let mz_config = Config::new(config.clone(), tunnel_cfg.clone(), InTask::No);
..
let mz_config = Config::new(config.clone(), tunnel_cfg, InTask::No);
```

martykulma (commented):
All good, I read it so fast I completely missed it!

martykulma (approved):
thanks @def- !

PR #36331, opened 2026-04-29
compute: reset memory limiter `last_check` on config change

antiguru:
### Motivation

Fixes a regression introduced in #32939.
Toggling `memory_limiter_interval` from `0s` to any non-zero value at runtime is currently unsafe: it can kill replicas at the next config update, even when memory usage is well below the limit.

Fixes: https://github.com/MaterializeInc/database-issues/issues/11343

### Description

While disabled, the limiter task sleeps with `interval = Duration::MAX` and never updates `last_check`.
On enable, the first `check` runs with `last_check` still set to task startup, so `elapsed` can be arbitrarily large.
The over-limit calculation `excess_bs = excess * elapsed` then exhausts the burst budget on the first sample and triggers `exit_thread_safe(167)`.

`apply_config` now resets `last_check` whenever the configuration changes, so the next check measures excess against the new interval rather than against a stale checkpoint.
No-op config changes still leave `last_check` untouched, to avoid silently delaying the next limit check.

Added unit tests in `src/compute/src/memory_limiter.rs` covering both behaviors.

### DIFF
--- a/src/compute/src/memory_limiter.rs
+++ b/src/compute/src/memory_limiter.rs
@@ -260,6 +260,13 @@ impl LimiterTask {
         info!(?config, "applying memory limiter config");
         self.config = config;
         self.burst_budget_remaining = config.burst_budget;
+        // Reset `last_check` so the next `check` measures the excess interval
+        // against the new config, not against a stale checkpoint left over from
+        // a period when the limiter was disabled or running with a different
+        // interval. Without this, enabling the limiter at runtime can charge an
+        // arbitrarily large excess byte-second amount against the burst budget
+        // and immediately terminate the process.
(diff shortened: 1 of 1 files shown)


def- (approved):
test lgtm, did you verify that it failed before this change?

antiguru:
Verified now: with the `self.last_check = Instant::now();` line removed from `apply_config`, `apply_config_resets_last_check` fails with "apply_config must refresh last_check to avoid charging stale elapsed time against the burst budget when the limiter is enabled at runtime". `apply_config_noop_keeps_last_check` still passes (config equality short-circuit unaffected). With the fix in place, both pass.

PR #36333, opened 2026-04-29
MySQL Source Versioning V2 - reorganized boogaloo

patrickwwbutler:
This PR effectively re-implements source versioning for mysql after we reverted the initial implementation last week due to decoding problems with exclude columns resulting in an incident. In doing so, it fixes a number of issues that existed with the first implementation, including:
https://github.com/MaterializeInc/database-issues/issues/11312
https://github.com/MaterializeInc/database-issues/issues/11313
https://github.com/MaterializeInc/database-issues/issues/11315

And provides a safer mechanism for handling schema changes and changes to the `binlog_row_metadata` MySQL system variable.

The first commit is roughly the changes in https://github.com/MaterializeInc/materialize/pull/36253 which should be merged first, and is required for the other changes.

Second commit updates the decoding logic based on the binlog metadata setting at source creation

Third commit updates the logic to verify mysql schemas with the schemas in the upstream, allowing for certain types of schema changes when binlog_row_metadata is FULL.

Fourth commit contains docs for how to make schema changes to your mysql source without downtime in materialize.

### DIFF
--- a/src/mysql-util/src/decoding.rs
+++ b/src/mysql-util/src/decoding.rs
@@ -7,9 +7,9 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::fmt::Write;
 use std::str::FromStr;
 
-use itertools::{EitherOrBoth, Itertools};
 use mysql_common::value::convert::from_value_opt;
 use mysql_common::{Row as MySqlRow, Value};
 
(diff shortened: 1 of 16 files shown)


src/mysql-util/src/decoding.rs
```diff
@@ -28,46 +28,177 @@ pub fn pack_mysql_row(
...
+            qualified_table_name: format!("{}.{}", table_desc.schema_name, table_desc.name),
+            error: "Table was created with binlog_row_metadata=FULL but binlog_row_metadata has since been set to a different value, meaning we cannot reliably decode the columns".to_string(),
+        });
+    }
+
+    // For each column in `table_desc` (in descriptor order), resolve its wire
+    // index. Non-fallback rows are matched by name so a reordered upstream
+    // still decodes correctly; fallback rows have no names and are matched
+    // positionally. A `None` here means the upstream row is missing this
+    // column and is only tolerated for ignored columns.
```
ublubu: The reference to "fallback" here confuses me.

The only use of `fallback_names` is in `if binlog_full_metadata && fallback_names {` above.

My best guess is that it's being used as a proxy for `binlog_full_metadata`. But we didn't check for `!binlog_full_metadata && !fallback_names`, so it's not a perfect one-to-one `binlog_full_metadata iff !fallback_names`.

src/mysql-util/src/decoding.rs
```diff
@@ -28,46 +28,177 @@ pub fn pack_mysql_row(
...
+        return Err(MySqlError::ValueDecodeError {
+            column_name: "<unknown>".to_string(),
+            qualified_table_name: format!("{}.{}", table_desc.schema_name, table_desc.name),
+            error: "Table was created with binlog_row_metadata=FULL but binlog_row_metadata has since been set to a different value, meaning we cannot reliably decode the columns".to_string(),
+        });
+    }
+
+    // For each column in `table_desc` (in descriptor order), resolve its wire
+    // index. Non-fallback rows are matched by name so a reordered upstream
+    // still decodes correctly; fallback rows have no names and are matched
```
ublubu: Another point of confusion for me:

This reads like you can have "fallback" and "non-fallback" rows side-by-side in the same table.

But the behavior is actually determined by `binlog_full_metadata`, which is a table attribute.

src/mysql-util/src/decoding.rs
ublubu: > A `None` here ... is only tolerated for ignored columns

It's not clear to me what "here" means. At first I thought it meant `if col_desc.column_type.is_none() {` below, but that doesn't make sense because that's how we define which columns are to be ignored.

src/mysql-util/src/decoding.rs
```diff
@@ -28,46 +28,177 @@ pub fn pack_mysql_row(
...
+    // positionally. A `None` here means the upstream row is missing this
+    // column and is only tolerated for ignored columns.
+    for (i, col_desc) in table_desc.columns.iter().enumerate() {
+        let wire_idx = if !binlog_full_metadata {
+            (i < row.len()).then_some(i)
+        } else {
+            row.columns_ref()
+                .iter()
+                .position(|wc| wc.name_str() == col_desc.name.as_str())
         };
```
ublubu, line 56: It might be clearer to put comments directly inside the branches of this `if-else` block instead.

src/mysql-util/src/decoding.rs
```diff
@@ -28,46 +28,177 @@ pub fn pack_mysql_row(
...
         if col_desc.column_type.is_none() {
             // This column is ignored, so don't decode it.
             continue;
         }
```
ublubu: This might as well happen first in the `for` loop.

src/mysql-util/src/decoding.rs
```diff
@@ -28,46 +28,177 @@ pub fn pack_mysql_row(
...
-        match pack_val_as_datum(value, col_desc, &mut packer) {
-            Err(err) => Err(MySqlError::ValueDecodeError {
-                column_name: col_desc.name.clone(),
-                qualified_table_name: format!("{}.{}", table_desc.schema_name, table_desc.name),
-                error: err.to_string(),
-            })?,
-            Ok(()) => (),
+        let wire_idx = match wire_idx {
+            Some(idx) => idx,
+            None => {
```
ublubu, line 60: Ah, this is "A `None` here". This feels like a good spot for that comment.

Or omit the comment. This error message and the `// This column is ignored` above already tell the story.

ublubu (commented):
I took a first pass at the "decode binlog rows by column name" commit. I'll look at the tests more carefully tomorrow.

doc/user/content/ingest-data/mysql/source-versioning.md
```diff
@@ -0,0 +1,193 @@
+---
+title: "Guide: Handle upstream schema changes with zero downtime"
```
kay-kim: Could we have the user-facing docs be a separate PR that incorporates the reference docs + ingest data pages as well?
Otherwise, this page gets published as soon as this branch is merged ... not when the feature is actually released ... so, that if we skip a release week, etc ... and people have hooked up AI to the docs (or Matty), they'll get the wrong idea.

src/mysql-util/src/decoding.rs
```diff
@@ -28,46 +28,177 @@ pub fn pack_mysql_row(
...
-                    column_name: col_desc.name.clone(),
-                    qualified_table_name: format!("{}.{}", table_desc.schema_name, table_desc.name),
-                    error: "extra column description".to_string(),
-                })?
-            }
-            EitherOrBoth::Right(_) => {
-                // If there are extra columns on the upstream table we can safely ignore them
-                break;
-            }
+    // If a column name begins with '@', then the binlog does not have full row metadata,
```
def-: You can also manually create a table with a column name starting with `@`, which would throw this logic off.

src/mysql-util/src/desc.rs
```diff
@@ -90,12 +94,29 @@ impl MySqlTableDesc {
...
+                    continue;
+                }
+                other.columns.iter().find(|c| c.name == self_column.name)
+            } else {
+                other_columns.next()
+            };
+            if self_column.column_type.is_none() {
+                // This is an excluded column and can be ignored.
+                continue;
+            }
```
def-: Do we also need this logic for the snapshot path?

src/sql/src/pure.rs
```diff
@@ -1526,6 +1531,8 @@ async fn purify_alter_source_add_subsources(
             let initial_gtid_set =
                 mz_mysql_util::query_sys_var(&mut conn, "global.gtid_executed").await?;
 
+            let binlog_full_metadata = ensure_binlog_full_metadata(&mut conn).await.is_ok();
```
martykulma: sorry, looking this over with fresh eyes - I see my suggestion had a flaw: a transient connection failure will set this to false and allow it to continue.  Only the new error variants should cause this to be false.

src/mysql-util/src/decoding.rs
```diff
@@ -28,46 +28,177 @@ pub fn pack_mysql_row(
...
+        .is_some_and(|col| col.name_ref().starts_with(b"@"));
+
+    if binlog_full_metadata && fallback_names {
+        // This should never happen, but if it does, it's a sign that something is very wrong with the MySQL server's binlog configuration. We want to error rather than silently producing incorrect results.
+        return Err(MySqlError::ValueDecodeError {
+            column_name: "<unknown>".to_string(),
+            qualified_table_name: format!("{}.{}", table_desc.schema_name, table_desc.name),
+            error: "Table was created with binlog_row_metadata=FULL but binlog_row_metadata has since been set to a different value, meaning we cannot reliably decode the columns".to_string(),
+        });
+    }
```
martykulma: Can we move this up to compare binlog_full_metadata and the TableMapEvent in `handle_rows_event`?  There's optional metadata that's probably set only if FULL.

src/mysql-util/src/decoding.rs
patrickwwbutler: deeply upsetting news. I think then we check for `@`, if it exists, compare it against the expected column name we have for index 0 (possibly check other indices if 0 is an excluded column? we then run into some issues if they do something like `ALTER TABLE foo ADD COLUMN @1 FIRST`, which is super cursed. I guess worst case scenario we can just check if all the binlog columns follow the pattern `@1, @2, ..., @n`, but that feels excessive to do on every single row

martykulma (commented):
Looking good! There are some additional improvements to make; partly my fault - i didn't think through the strict vs. lenient checks in purification (sorry!).

martykulma:
I just realized that `test/mysql-cdc-resumption/mzcompose.py` has verification disabled too - should one of these be re-enabled?

```
    # TODO: database-issues#7683: one of the two following commands must succeed
    # run_testdrive_files(c, "verify-rows-after-restore-t1.td")
    # run_testdrive_files(c, "verify-source-failed.td")
```

src/mysql-util/src/decoding.rs
martykulma: +1 to rewriting as I also found this hard to follow

src/mysql-util/src/decoding.rs
martykulma: I have a related comment to move this check up a level!  TableMapEvent optional metadata columns may be of use here https://dev.mysql.com/doc/dev/mysql-server/9.6.0/classmysql_1_1binlog_1_1event_1_1Table__map__event.html#Table_table_map_event_optional_metadata

doc/user/content/ingest-data/mysql/source-versioning.md
patrickwwbutler: done!

kay-kim: Thank you!!!

src/sql/src/pure.rs
patrickwwbutler: ah this is part of the first PR in this "stacked" PR, good point there, will go address it on that one: https://github.com/MaterializeInc/materialize/pull/36253

this is immediately after the `initial_gtid_set` query to the upstream, so it feels like it would be rare that we would suddenly get a connection failure, but certainly not impossible.

src/mysql-util/src/desc.rs
patrickwwbutler: The snapshot path does end up going through here as well, via `verify_schemas`, so it should be applied the same there

src/mysql-util/src/decoding.rs
patrickwwbutler: I assume you mean we should use the optional metadata to check if there is full metadata, and create an error there rather than in decoding? And then here we can not bother with `fallback_names`, instead just using the `binlog_full_metadata` to determine behavior?

src/mysql-util/src/decoding.rs
patrickwwbutler: Yeah that's a fair confusion, "non-fallback rows" refers not to the table, but the binlog, and whether or not we have metadata (column names), or we need to _fallback_ to indices, but this could use some refactoring for clarity, and should be clearer particularly with marty's suggestion of moving the check up a level, and then only looking at `binlog_full_metadata` here.

src/mysql-util/src/decoding.rs
patrickwwbutler: Ran into an annoying issue here with snapshotting - the snapshot workflow doesnt have a table map event as it doesn't use the binlog stream, so it seems likely that we'll have to manually check the setting there if we don't want to use the "@" column name strategy, which is pretty sweaty.

I think that specifically for snapshotting it's probably acceptable to run a `query_sys_var` to fetch it, though? We already do that for the snapshot gtid set, anyways

martykulma: sorry, I missed the message from Friday



> I assume you mean we should use the optional metadata to check if there is full metadata, and create an error there rather than in decoding? And then here we can not bother with `fallback_names`, instead just using the `binlog_full_metadata` to determine behavior?

yes - my intention was doing the check in the caller since you already have a TME in `handle_rows_event`

> I think that specifically for snapshotting it's probably acceptable to run a query_sys_var to fetch it, though? We already do that for the snapshot gtid set, anyways

yup - for snapshot, it calls `pack_mysql_row` directly, so it's still checked in the caller.  Once the snapshot query is started, it won't matter if the setting changes (if it changes after this point, replication will see it) - so I expect it only needs to be checked once before we start looping over the stream of snapshot data.

patrickwwbutler: as it turns out, because snapshot data is fetched via a query to the upstream, and not from the binlog, it always comes with column name metadata, and we don't actually need to check there!

patrickwwbutler:
> I just realized that `test/mysql-cdc-resumption/mzcompose.py` has verification disabled too - should one of these be re-enabled?
> 
> ```
>     # TODO: database-issues#7683: one of the two following commands must succeed
>     # run_testdrive_files(c, "verify-rows-after-restore-t1.td")
>     # run_testdrive_files(c, "verify-source-failed.td")
> ```

Ah this is actually the test that is flaky because of the DROP-> re-CREATE race condition, which isn't fixed quite yet, so it should stay disabled for now

src/storage/src/source/mysql/replication/events.rs
```diff
@@ -289,20 +295,38 @@ pub(super) async fn handle_rows_event(
...
-                        value: row,
-                        metadata: Row::default(),
-                    }),
-                    // Produce a DefiniteError in the stream for any rows that fail to decode
-                    Err(err @ MySqlError::ValueDecodeError { .. }) => Err(DataflowError::from(
-                        DefiniteError::ValueDecodeError(err.to_string()),
-                    )),
-                    Err(err) => Err(err)?,
+                let event = if !has_full_metadata && output.binlog_full_metadata {
+                    tracing::warn!(%id, "timely-{worker_id} missing full metadata for {table:?} \
```
martykulma: This has the potential to be logged a substantial number of times - num_rows x num_exports - we should probably just report it once per RowsEvent

src/storage/src/source/mysql/snapshot.rs
martykulma, line 544: comment needs an update

src/storage/src/source/mysql/replication/events.rs
```diff
@@ -289,20 +295,38 @@ pub(super) async fn handle_rows_event(
...
+                        - this can lead to incorrect decoding of some data types. This metadata is only available on MySQL 8.0+ with binlog_version=2, and must be enabled with the binlog_row_metadata configuration option.");
+                    Err(DataflowError::from(DefiniteError::ValueDecodeError(
+                        format!(
+                            "Table {0} was created with binlog_row_metadata=FULL but binlog_row_metadata has since been set to a different value, meaning we cannot reliably decode the columns",
+                            output.table_name
+                        ),
```
martykulma, line 308: We never mark the export(s) as borked (`ctx.errored_outputs`), which means we will continue to emit errors as long as the setting is incorrect.  Once the customer corrects it, there isn't a way to recover for MZ, but we will try!

Consider we have source will full metadata, a row with a value A that sees some updates and is deleted, and interleaved someone accidentally changes row metadata:
```
A -> B    .... A:-1 B:+1
------------------------------------ metadata: FULL -> MINIMAL
B -> C    .... Err(B):-1 Err(C):+1
------------------------------------ metadata: MINIMAL -> FULL
C -> NULL .... C:-1
```

patrickwwbutler: This seems like something that is fundamentally unrecoverable, no? You mean that by adding it to `ctx.errored_outputs` we and ensuring that it cannot be recovered, and also that the source is more visibly broken, I assume?

src/storage/src/source/mysql/replication/events.rs
patrickwwbutler: this one is annoying because technically the binlog setting is on a per-output basis, but honestly I think it can just be deleted. The error will be logged further up and the cause is written out quite explicitly.

src/storage/src/source/mysql/replication/events.rs
martykulma: yes, by adding it - we stop processing events for it. The last thing appended would be the error. I believe we do this for the MySQL DDL errors, and you should also find it in PG.

martykulma (approved):
nice - lgtm!

patrickwwbutler:
triggering a nightly before merging, fingers crossed

PR #36336, opened 2026-04-30
Stable prometheus metrics proposal

SangJunBak:
Rendered: https://github.com/SangJunBak/materialize/blob/3b986137b2bf527ddbcf767ad1359ffbfb93acf3/doc/developer/design/20260430_stable_prometheus_metrics.md

Putting this out for more buy-in from other team members

### Motivation

Closes [SQL-152](https://linear.app/materializeinc/issue/SQL-152)  https://linear.app/materializeinc/issue/SQL-152/audit-proposed-metrics


Alphadelta14 (approved):

doc/developer/design/20260430_stable_prometheus_metrics.md
```diff
@@ -0,0 +1,89 @@
...
+
+This is largely inspired by CockroachDB's separation of Essential metrics (https://www.cockroachlabs.com/docs/v25.3/essential-metrics-advanced and https://www.cockroachlabs.com/docs/v26.2/metrics.html) where they don't promise stability on their larger set of metrics. Also Clickhouse's Prometheus integration https://clickhouse.com/docs/integrations/prometheus and their "filtered_metrics" search parameter.
+
+
+I proposed we create three tiers for our metrics:
+
+```rs
+enum Lifecycle {
+    Internal // The default for all our current metrics.
+    Stable,
```
Alphadelta14, line 99: I almost feel like we need a SemiStable or LimitedStable or Experimental tier to indicate intent behind changes to labels or contents.
Maybe we could even not use stable _until_ they have gone through such LimitedStable period for X releases (like 6 months or so).

doc/developer/design/20260430_stable_prometheus_metrics.md
```diff
@@ -0,0 +1,89 @@
...
+**Dataflow**
+- `mz_compute_collection_count`
+- Prometheus-native equivalents of these dataflow SQL exporter queries:
+  - `mz_arrangement_size_bytes`
+  - `mz_compute_replica_park_duration_seconds_total`
+  - `mz_dataflow_elapsed_seconds_total`
+  - `mz_compute_collection_count`
+  - `mz_compute_hydration_time_seconds`
+
+**Source**
```
Alphadelta14, line 148: I'd encourage us to also emit `_info` properties that are fairly static. I know we don't currently do this

SangJunBak: Hmm so my original plan was to attach names as labels to each metric, but I wonder if it makes sense to expose the id->name mappings as `info` metrics and have users join against it?

SangJunBak: I know there's also an experimental info function https://prometheus.io/blog/2025/12/16/introducing-info-function/

Alphadelta14: joins in promql are a big no-no. so it'd be better to have names in both place. usually your _info will also contain other misc metadata like version information. it's really just a best practice.

Alphadelta14: The info() function is non-standard and likely not implemented everywhere. it's not even in prometheus by default, so I wouldn't rely on it. it'd be nice for it to be available in the future, ofc.

SangJunBak: However I'm not seeing a convenient way to join metrics with id->name mappings in datadog.

doc/developer/design/20260430_stable_prometheus_metrics.md
```diff
@@ -0,0 +1,100 @@
...
+- Any metric removed should be called out in the release notes. We can generate html of our current metrics based on the `metric!` macro reusing our `walkabout` crate. For updating the release notes, we can modify our release notes skill to explictly look at diffs surrounding that file.
+- Any stable metric marked for deprecation will be removed in the next major release (common practice according to @Alphadelta14)
+- We can have a lint to not remove stable metrics based on the generated html file, only deprecated metrics.
+- Inspired by Kubernetes' metrics tier https://kubernetes.io/docs/concepts/cluster-administration/system-metrics/
+
+
+### Proposed Stable Metrics
+
+**Replica utilization**
+- `mz_memory_limiter_memory_usage_bytes` — memory + swap
```
maheshwarip, line 117: why would this be named `mz_memory_limiter`?

maheshwarip: Can we split out memory, swap and heap (memory + swap) into separate metrics?

doc/developer/design/20260430_stable_prometheus_metrics.md
```diff
@@ -0,0 +1,186 @@
...
+
+#### Wire format
+
+| | `/metrics` (default Accept) | `/metrics` + `Accept: protobuf` | env's `/metrics/federated` |
+|---|---|---|---|
+| Caller | external Prometheus, curl | env's federated scraper, Prometheus 2.x | external Prometheus |
+| Body | vanilla Prometheus text | vanilla delimited Prometheus protobuf | text, enriched with names |
+| Headers | standard + `X-Mz-Enrich-Rules` | standard + `X-Mz-Enrich-Rules` | standard |
+
+External clients always get the rules header alongside their data; they ignore unknown headers, so behavior is unchanged for them.
```
aljoscha, line 65: Can't we make it so that we only send these headers when the caller requests them, and only our internal federated endpoint requests those. That would keep the existing /metrics endpoint "clean"

SangJunBak: We could also do that too via request headers!

PR #36340, opened 2026-04-30
timely-util: add ConsolidatingColumnBuilder

antiguru:
### Motivation

Infrastructure for the columnar rendering migration. Today, render operators that produce updates use `differential_dataflow::consolidation::ConsolidatingContainerBuilder<Vec<(D, T, R)>>`, which emits `Vec<(D, T, R)>` containers. Downstream operators want `Column<(D, T, R)>`, so we currently re-convert around every operator. This PR adds a builder that consolidates and emits columnar containers directly. No call sites change here.

### Description

`ConsolidatingColumnBuilder<D, T, R>` lives in `src/timely-util/src/columnar/consolidate.rs`. It vendors DD's staging-buffer logic — lazy `Vec` reserve to `2 * default_capacity::<(D, T, R)>()`, `consolidate_updates` when full, drain the largest multiple-of-`preferred_capacity` prefix — and pushes the survivors into an inner `ColumnBuilder<(D, T, R)>`. The inner builder handles ~2MB-aligned chunking, so we don't manage that here.

`extract` delegates to the inner builder (does not flush partial state, matching DD). `finish` drains everything that survives consolidation, then delegates to the inner `finish`.

Not length-preserving — consolidation may drop tuples with diff that sums to zero — so no `LengthPreservingContainerBuilder` impl. Reusing drained `current` capacity is left for a later pass; behavior matches DD first.

### Verification

Five new unit tests in the module: empty `finish`, single-push round-trip, threshold-triggered consolidation (all `+1`/`-1` cancel), multi-key consolidation, and forced emission of multiple inner containers. `cargo test -p mz-timely-util` passes.

### DIFF
--- a/src/timely-util/src/columnar/consolidate.rs
+++ b/src/timely-util/src/columnar/consolidate.rs
@@ -1,0 +1,243 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License in the LICENSE file at the
+// root of this repository, or online at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! A `ContainerBuilder` that consolidates `(D, T, R)` updates and emits columnar containers.
+//!
+//! Vendored from `differential_dataflow::consolidation::ConsolidatingContainerBuilder` for the
+//! staging path; the inner [`ColumnBuilder`] mints aligned columnar containers from the
+//! consolidated tuples.
+
+use columnar::{Columnar, Push};
+use differential_dataflow::Data;
+use differential_dataflow::consolidation::consolidate_updates;
+use differential_dataflow::difference::Semigroup;
+use timely::container::{ContainerBuilder, PushInto};
+
+use crate::columnar::Column;
+use crate::columnar::builder::ColumnBuilder;
+
+/// A container builder that consolidates `(D, T, R)` updates and emits `Column<(D, T, R)>`.
+///
+/// Buffers updates in an internal `Vec`, calls
+/// [`consolidate_updates`] when the buffer fills, and pushes the surviving updates into an inner
+/// [`ColumnBuilder`]. The inner builder chunks output into ~2MB-aligned columnar containers
+/// (see [`ColumnBuilder`] for details).
+///
+/// Does **not** maintain FIFO ordering (consolidation reorders updates).
+pub struct ConsolidatingColumnBuilder<D, T, R>
+where
+    D: Data,
+    T: Data,
+    R: Semigroup + 'static,
+    (D, T, R): Columnar,
+{
+    /// Pre-consolidation staging buffer.
+    current: Vec<(D, T, R)>,
+    /// Inner columnar builder; receives consolidated tuples and emits `Column<(D, T, R)>`.
+    column: ColumnBuilder<(D, T, R)>,
+}
+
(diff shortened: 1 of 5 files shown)


src/timely-util/src/columnar/consolidate.rs
```diff
@@ -0,0 +1,243 @@
...
+/// [`ColumnBuilder`]. The inner builder chunks output into ~2MB-aligned columnar containers
+/// (see [`ColumnBuilder`] for details).
+///
+/// Does **not** maintain FIFO ordering (consolidation reorders updates).
+pub struct ConsolidatingColumnBuilder<D, T, R>
+where
+    D: Data,
+    T: Data,
+    R: Semigroup + 'static,
+    (D, T, R): Columnar,
```
antiguru: Do we need the type annotations here? Reduce to a minimum if possible.

src/timely-util/src/columnar/consolidate.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// Inner columnar builder; receives consolidated tuples and emits `Column<(D, T, R)>`.
+    column: ColumnBuilder<(D, T, R)>,
+}
+
+impl<D, T, R> Default for ConsolidatingColumnBuilder<D, T, R>
+where
+    D: Data,
+    T: Data,
+    R: Semigroup + 'static,
+    (D, T, R): Columnar,
```
antiguru: Same here, minimal parameters.

src/timely-util/src/columnar/consolidate.rs
```diff
@@ -0,0 +1,243 @@
...
+    <(D, T, R) as Columnar>::Container: Push<(D, T, R)>,
+{
+    /// Push an element.
+    ///
+    /// Precondition: `current` is not allocated or has space for at least one element.
+    #[inline]
+    fn push_into(&mut self, item: (D, T, R)) {
+        let preferred_capacity = timely::container::buffer::default_capacity::<(D, T, R)>();
```
antiguru: A risk: the preferred capacity might be rather small, about 200 elements for `(Row, Timestamp, Diff)`, which doesn't justify the cost of building a columnar structure. Think about either increasing this (what we do in the merge batcher, target 64KiB instead of 8KiB), or find an alternative design.

src/timely-util/src/columnar/consolidate.rs
antiguru: Done — struct bounds reduced to just `(D, T, R): Columnar` (the only one any field actually requires). DD-style bounds (`D: Data`, `T: Data`, `R: Semigroup + 'static`) are kept on the impls that call `consolidate_updates`/`column.push_into`, where they're real.

src/timely-util/src/columnar/consolidate.rs
antiguru: Same change applied here.

src/timely-util/src/columnar/consolidate.rs
antiguru: Bumped to a 64 KiB `BUFFER_SIZE_BYTES` constant matching `ColumnationChunker` in `src/timely-util/src/columnation.rs`. `chunk_capacity::<(D, T, R)>()` is the per-chunk item count; staging buffer is `2 * chunk_capacity` so the consolidate-on-full path triggers at the same fill ratio as before, just at 64 KiB instead of 8 KiB.

DAlperin (approved):
Yeah this tracks nicely with a lot of the patterns from my batcher PR. Very cool

src/timely-util/src/columnar/consolidate.rs
```diff
@@ -0,0 +1,373 @@
...
+use differential_dataflow::Data;
+use differential_dataflow::consolidation::consolidate_updates;
+use differential_dataflow::difference::Semigroup;
+use timely::container::{ContainerBuilder, PushInto};
+
+use crate::columnar::Column;
+
+/// Items per staging buffer. Small enough that O(n log n) sort stays cheap, large enough to
+/// amortize the sort + drain overhead across many pushes.
+const STAGING_CAP_ITEMS: usize = 16 * 1024;
```
def-, line 48: Are we ok with the memory overhead this causes?

PR #36342, opened 2026-04-30
storage: gate upsert v2 batcher seal on eligibility preconditions

patrickwwbutler:
upsert_continual_feedback_v2 called MergeBatcher::seal on every wakeup whenever the source stash was non-empty. seal performs an O(N) merge of all chains regardless of how much it extracts, so during upstream snapshots — where input_upper sits at T0 while data piles up at ts=T0 — the operator burned O(N²) work merging-then-extracting-nothing, making snapshot CPU-bound.

Skip the seal unless an eligible entry is at all possible. For an entry at ts to be eligible we need ts == persist_upper && ts < input_upper, so the necessary preconditions are:
  1. cap.time() <= persist_upper (cap.time() lower-bounds min(ts)).
  2. persist_upper < input_upper.

The same expression covers both the post-hydration source-snapshot case (cap == persist == input → #2 fails) and the rehydration- with-source-ahead case (cap > persist → #1 fails).

### Motivation

Closes <issue>
Fixes <bug>
etc.

### Description

diff *can't* tell a reviewer.

### Verification

### DIFF
--- a/src/storage/src/upsert_continual_feedback_v2.rs
+++ b/src/storage/src/upsert_continual_feedback_v2.rs
@@ -397,9 +397,34 @@ where
             //     lookup on the persist trace.
             //   - Ineligible (persist_upper < ts < input_upper): persist
             //     hasn't caught up yet; pushed back into the batcher.
-            if stash_cap.is_some() {
-                let cap = stash_cap.as_mut().unwrap();
-
+            //
+            // We skip the seal entirely unless an eligible entry is at all
+            // possible. `seal` performs an O(N) merge of all chains
+            // regardless of how much it extracts, so calling it when nothing
+            // can be processed makes the operator quadratic in the number of
+            // wakeups (a real pathology during upstream snapshots and during
+            // rehydration when the source races ahead of persist).
+            //
+            // For an entry at `ts` to be eligible we need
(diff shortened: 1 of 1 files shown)


src/storage/src/upsert_continual_feedback_v2.rs
```diff
@@ -397,7 +397,37 @@ where
...
+            // case (cap == persist == input → condition 2 fails) and the
+            // rehydration-with-source-ahead case (cap > persist → condition
+            // 1 fails). It also no-ops correctly when persist has shut down
+            // (empty persist_upper makes condition 2 vacuously false).
+            let any_eligible_possible = if let Some(cap) = stash_cap.as_ref() {
+                !persist_upper.less_than(cap.time())
+                    && PartialOrder::less_than(&persist_upper, &input_upper)
+            } else {
+                false
+            };
```
martykulma: nit: you can extract cap and drop the unwrap 

```
if let Some(cap) = stash_cap.as_mut()
    && !persist_upper.less_than(cap.time())
    && PartialOrder::less_than(&persist_upper, &input_upper)
{
```

martykulma (approved):
lgtm!  minor nit - we don't need the extra conditional block and we can drop the unwrap.

PR #36344, opened 2026-04-30
console [CNS-61]: fix LoadingScreen flicker on every navigation

leedqin:

### Motivation

Navigating in the console is causing some react suspense behavior that forces a full screen reload. It is disruptive to the user experience. Fixes [CNS- 61](https://linear.app/materializeinc/issue/CNS-61/loading-screen-flickers-on-page-navigation-in-the-console)

### Description
usePollEnvironmentHealth's queryKey included `appConfig` and a useRef wrapper. Since appConfig.consoleUrl is a live `window.location` reference, every URL change minted a new cache key, causing useSuspenseQuery to miss and re-throw — triggering the suspense fallback on every route change.

Drop the unstable values from the queryKey useEffect with updating values with a pure mergeEnvironmentMaps helper
### Verification

https://github.com/user-attachments/assets/2eef1cd2-1928-44c0-baee-a92f4ab41750

### DIFF
--- a/console/src/store/environments.ts
+++ b/console/src/store/environments.ts
@@ -11,11 +11,11 @@ import * as Sentry from "@sentry/react";
 import { useSuspenseQuery } from "@tanstack/react-query";
 import { add, Duration, formatDuration, sub } from "date-fns";
 import deepEqual from "fast-deep-equal";
-import { atom, useAtom, useAtomValue } from "jotai";
+import { atom, useAtom, useAtomValue, useSetAtom } from "jotai";
 import { atomFamily, loadable } from "jotai/utils";
 import { sql } from "kysely";
 import { PostgresError } from "pg-error-enum";
-import React, { useRef } from "react";
+import React from "react";
 import { gte as semverGte, parse as semverParse, SemVer } from "semver";
 
 import { buildGlobalQueryKey } from "~/api/buildQueryKeySchema";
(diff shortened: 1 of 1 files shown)


console/src/store/environments.ts
```diff
@@ -310,48 +310,41 @@ export const environmentQueryKeys = {
...
+  const merged = new Map<string, Environment>();
+  for (const [regionId, fresh] of next) {
+    const prev = previous.get(regionId);
+    const updated = updateEnviromentState(prev, fresh);
+    merged.set(regionId, updated);
+    if (!changed && !deepEqual(prev, updated)) {
+      changed = true;
+    }
+  }
+  return changed ? merged : previous;
```
SangJunBak: I'm not sure if returning the original object will stop re-renders, but even if it does, it's still not clear from Jotai's documentation. Can we create a derived write only atom instead then just do an early return if we don't actually need to set anything?

SangJunBak:
>  Since appConfig.consoleUrl is a live window.location reference, every URL change minted a new cache key,

This makes sense!

SangJunBak (approved):
Small suggestion, but otw looks good!

console/src/store/environments.ts
leedqin: Addressed in the latest commit

SangJunBak (approved):

PR #36354, opened 2026-05-02
persist: Fix txn-wal filter pushdown looking up wrong column name

def-:
`TxnsCodecRow::should_fetch_part` looked up column "key" instead of "ok", silently disabling per-shard filter pushdown since SourceData has no such field. Introduced in #29151.


DAlperin (approved):
We should probably make this a const

PR #31689, opened 2025-02-27
Add IAM authentication support for MySQL connections

martykulma:
This adds the ability for users to configure IAM authentication instead of a password when using RDS MySQL.

This change adds syntax to `CREATE CONNECTION` for mysql that will accept `AWS CONNECTION`.
`AWS CONNECTION` is mutually exclusive with the options
1. `PASSWORD`
2. `SSL MODE 'disabled'` or not specified

example:
```
> create connection aws_conn to AWS (region 'us-east-1', assume role arn 'arn:aws:iam::400121260767:role/marty-rds-assumerole-test', assume role session name 'maz-test');
> create connection mysql_conn to MYSQL (host 'marty-mysql-iam.ctthmav6dsti.us-east-1.rds.amazonaws.com', user 'mzuser', ssl mode 'required', aws connection aws_conn);
> create source mysql_iam in cluster quickstart from mysql connection mysql_conn for tables ( test.bark );

> select * from bark;
| id | k | v        |
| -- | - | -------- |
| 11 | y | yams     |
| 13 | h | hops     |
| 14 | h | hops     |
| 10 | z | zebra    |
| 12 | g | guava    |
| 16 | b | bananna  |
| 15 | v | vegimite |

```

### Motivation

fixes https://github.com/MaterializeInc/database-issues/issues/6946

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
[...]

PR #36357, opened 2026-05-03
mysql: percent-encode username in RDS IAM auth token

def-:
Raw interpolation of the username into the sigv4 URL caused query-string corruption for usernames containing `&` or `#`, breaking IAM authentication. Follow-up to https://github.com/MaterializeInc/materialize/pull/31689


martykulma (approved):
thanks @def- !

PR #30812, opened 2024-12-12
storage/kafka: Shrink buffers in `sink_collection` operator

ParkMyCar:
In the Kafka `sink_collection` operator we re-use two `Vec`s to buffer messages, but never shrink them. So it's possible for a large update to blow up the size of these `Vec`s and then we're stuck holding the heap allocation for a while.

### Motivation

Possibly fix an issue we've been seeing.

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #36359, opened 2026-05-03
storage/kafka: move shrink_to calls after drains

def-:
The shrink_to calls added in #30812 were placed before the drains that empty the Vecs. Since shrink_to reduces capacity to max(len, min_capacity), the calls were no-ops in the oversized-buffer scenario they were meant to handle. Move them to after extract_if/drain when len is small or zero.


martykulma (approved):
lgtm - thanks @def- !

PR #36360, opened 2026-05-03
sql: fix seahash catalog return type (uint4 -> uint8)

def-:
The seahash functions return u64 but were registered with UInt32 in func.rs, causing mz_functions.return_type_id to disagree with the actual runtime type. Introduced in PR #29099.


antiguru (approved):
Thank you!

PR #29354, opened 2024-09-03
controller: make `Controller::process` non-`async`

teskje:
This PR makes converts the `Controller::process` method to a non-`async` one, removing one potential source of blocking from the coordinator main loop.

The most significant change here is making the channel on which updates are sent to the `CollectionManager` unbounded. This does introduce the risk of higher memory usage on envd. This is not the first use of an unbounded channel in envd and in the past it has actually been helpful seeing their memory usage grow to realize that something was wrong with the system. Anecdotically, we seem to be better at debugging increased memory usage than blockage in the coordinator loop.

### Motivation

  * This PR fixes a previously unreported bug.

The coordinator main loop can be blocked at various places if the channels feeding the `CollectionManager` run full.

### Checklist

- [x] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [x] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [x] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [x] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [x] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #36362, opened 2026-05-03
webhook: remove dead ResourceExhausted backpressure path

def-:
PR #29354 replaced bounded channels with unbounded ones in MonotonicAppender, making StorageError::ResourceExhausted unreachable. The webhook 429 response is still produced by the Tower GlobalConcurrencyLimitLayer at the HTTP layer, so remove the dead variant and its match arms. Fix the docs to say "concurrent requests" instead of "concurrent requests/second" (it's a semaphore, not a rate limiter).


patrickwwbutler (approved):
LGTM!

PR #29578, opened 2024-09-16
Feature/add mz egress ips cidr support

jubrad:
### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #36363, opened 2026-05-03
catalog: canonicalize CIDRs in mz_egress_ips

def-:
Use `IpNet::network()` instead of `IpNet::addr()` so that host bits are zeroed before writing to the catalog. Without this, a non-canonical input like `10.0.5.7/24` would surface as-is rather than `10.0.5.0/24`.

Introduced in PR #29578.


mtabebe (approved):
I am curious how you found this?

def-:
After having gone through each file I'm currently going through each open/closed PR asking an LLM to look for potential bugs: https://github.com/MaterializeInc/qa-llm-review Then I try to reproduce them, and if it reproduces, fix if the fix is simple enough, otherwise open an issue and assign.

PR #36365, opened 2026-05-03
dbt-materialize: Add PARTITION BY support for materialized views

djahandarie:
## Summary

- Add `partition_by` config option to `MaterializeConfig`, accepting a list of column names
- Generate `PARTITION BY (...)` in the materialized view `WITH` clause
- **Fix existing bug**: the macro previously emitted separate `WITH (...)` blocks for
  `refresh_interval`, `retain_history`, and contract constraints. Materialize only allows
  a single `WITH` clause per `CREATE MATERIALIZED VIEW`, so combining multiple options
  would produce invalid SQL. Refactored to collect all options and emit one unified
  `WITH (...)` block.
- For contract constraints, call `adapter.render_raw_columns_constraints()` directly
  instead of `get_table_columns_and_constraints()`, which wraps its output in parentheses
  that would need to be stripped before merging into the `WITH` clause

## Usage

```sql
{{ config(
    materialized='materialized_view',
    partition_by=['col1', 'col2']
) }}

SELECT col1, col2, col3 FROM ...
```

Partition columns must be a prefix of the relation's columns, matching Materialize's
`PARTITION BY` validation semantics.

## Test plan

- [ ] Added `test_materialized_view_partition_by` fixture and included it in `TestCustomMaterializations`
- [ ] Updated model count assertions (13 → 14) to account for the new model
- [ ] Existing constraint tests (`TestNullabilityAssertions`) validate the refactored
  constraint rendering path still produces correct `ASSERT NOT NULL` DDL

### DIFF
--- a/doc/user/content/manage/dbt/get-started.md
+++ b/doc/user/content/manage/dbt/get-started.md
@@ -586,6 +586,52 @@ You can specify the retention period using common time units like:
 - `'1d'` for one day
 - `'1w'` for one week
 
+##### Using partition by
+
+{{< tip >}}
+For guidance and best practices on how to use `PARTITION BY` in Materialize,
+see [Partitioning and filter pushdown](/transform-data/patterns/partition-by/).
+{{</ tip >}}
+
+To control the internal storage order that Materialize uses for a
+materialized view, use the `partition_by` configuration. This declares the
+columns that Materialize should sort by when writing the underlying data into
+storage, so that rows with similar values are stored together. The main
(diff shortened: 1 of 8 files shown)


djahandarie:

djahandarie:
recheck

djahandarie:
@bobbyiliev PTAL whenever you have a moment (it's not urgent at all for us)

bobbyiliev:
@djahandarie thank you for this PR! I will be reviewing this a bit later on today!

bobbyiliev (approved):
Thanks again for the contribution @djahandarie! It looks great. I had to make some minor changes and it should be good to go!

PR #36367, opened 2026-05-03
catalog: Fix retiring old shard on cache migration

def-:
Wrong function called in #31206.


SangJunBak (approved):
This makes sense to me!

PR #36368, opened 2026-05-03
adapter: Fix `OFFSET` planning in `INSERT INTO ... SELECT`

ggevay:
Fixes https://github.com/MaterializeInc/database-issues/issues/11347 by moving the `OFFSET` simplification and non-negativity check into `bind_parameters` (and renames `bind_parameters`), so that every caller will automatically get this functionality. Previously, this was missing from the call in planning `INSERT INTO ... SELECT`.

A second commit downgrades an `expect` to a soft-panic and a graceful query error, because now the code that ensures the invariant got a bit far from the `expect`.

(The added TODO comment on `bind_parameters_and_simplify_offset` is a pre-existing bug. A follow-up PR is fixing it: https://github.com/MaterializeInc/materialize/pull/36370)

### DIFF
--- a/src/sql/src/plan/query.rs
+++ b/src/sql/src/plan/query.rs
@@ -1629,17 +1629,18 @@ fn plan_query_inner(qcx: &mut QueryContext, q: &Query<Aug>) -> Result<PlannedQue
 }
 
 /// Converts an OFFSET expression into a value.
-pub fn offset_into_value(offset: HirScalarExpr) -> Result<i64, PlanError> {
+pub(crate) fn offset_into_value(offset: HirScalarExpr) -> Result<i64, PlanError> {
     let offset = offset
         .try_into_literal_int64()
         .map_err(|err| PlanError::InvalidOffset(err.to_string_with_causes()))?;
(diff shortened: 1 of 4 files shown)


mgree (approved):
One tiny nit. LGTM!

ggevay:
@mgree, Sorry, I don't see the nit. Where is it posted? Maybe a github glitch?

src/sql/src/plan/query.rs
```diff
@@ -1634,14 +1634,15 @@ pub fn offset_into_value(offset: HirScalarExpr) -> Result<i64, PlanError> {
...
-        return Err(PlanError::InvalidOffset(format!(
-            "must not be negative, got {}",
-            offset
-        )));
+        return Err(negative_offset_error(offset));
     }
     Ok(offset)
 }
 
+pub fn negative_offset_error(offset: i64) -> PlanError {
```
mgree: Document or make `pub(crate)`?

mgree: @ggevay gh is working now 🙃

PR #36369, opened 2026-05-03
Refactor subquery visitation

ggevay:
This adds `visit_direct_subqueries` helper functions on `HirScalarExpr`, to encapsulate the traversal and then manual match on the `Exists` and `Select` in 6 places.

Can be reviewed commit by commit.
1. adds `impl VisitChildren<HirRelationExpr> for HirScalarExpr`, which is the symmetric counterpart of the existing `impl VisitChildren<HirScalarExpr> for HirRelationExpr`.
2. adds `visit_direct_subqueries` and related helpers.
3. adds a bunch of comments to make exploration of HIR visitation code easier.

PR #36370, opened 2026-05-03
sql: Fix `bind_parameters`' visitation

ggevay:
This is on top of https://github.com/MaterializeInc/materialize/pull/36368 and https://github.com/MaterializeInc/materialize/pull/36369. Only the last commit is new. Edit: Added one more commit, so now the last 2 commits are new. The last commit just adds some more doc comments to HIR visitation.

This resolves the issue pointed out in https://github.com/MaterializeInc/materialize/pull/36368's TODO comment, which is that the HIR visitation in `bind_parameters` was messed up:
1. it was quadratic in subquery nesting depth to due to improper usage of HIR visitation functions;
2. calling the scalar expr `bind_parameters` naked (without an outer relation expr) was not doing the OFFSET simplification / checking inside subqueries.

(This is a bit hard to test: 1. can't really be tested with a real query at the moment, because it seems some other code has exponential slowdown with subquery nesting, which completely trumps the quadraticness issue, 2. all the current naked `bind_parameters` calls are in situations where subqueries are not allowed. But I added a simple test that at least exercises the changed code.)

### DIFF
--- a/src/sql/src/plan/hir.rs
+++ b/src/sql/src/plan/hir.rs
@@ -2286,15 +2286,6 @@ impl HirRelationExpr {
     /// Replaces parameter references in the expression with the corresponding datum from `params`.
     /// Additionally, it simplifies OFFSET clauses to constants after parameter binding, and checks
     /// them for non-negativity.
-    ///
-    /// TODO: This is currently quadratic with subquery nesting depth, because it calls
-    /// `bind_parameters` on each scalar expr at every nesting depth, but `bind_parameters` itself
-    /// also visits subqueries. I think we should
-    /// - make `bind_parameters_and_simplify_offset` call `bind_parameters` only on top-level scalar
-    ///   expressions,
-    /// - and make `bind_parameters` call back to us for subqueries.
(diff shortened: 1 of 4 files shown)


mgree (approved):
LGTM!

PR threads are broken for me, so I'll write the comment here:

> Maybe add a comment at hir.rs:2296 indicating that the `0` means we won't be looking at nested subqueries _here_---we'll visit them in `HirScalarExpr::bind_parameters_and_simplify_offset`.

ggevay:
> Maybe add a comment at hir.rs:2296 indicating that the `0` means we won't be looking at nested subqueries _here_---we'll visit them in `HirScalarExpr::bind_parameters_and_simplify_offset`.

Ah, actually, that's not what the `depth` argument means. `visit_scalar_expressions_mut` doesn't go inside subqueries, regardless of the `depth` argument. The `depth` is just a seed for what depths to pass to the closure. E.g., if you give 10 here, then one level deeper the closure would get 11. I'll add more doc commenting tomorrow before I merge the PR!

PR #36371, opened 2026-05-03
repr: document why Int2Vector is distinct from Array

ggevay:
Spell out on `struct Int2Vector` (and via pointers from `ReprScalarType::Int2Vector` and `parse_legacy_vector_inner`) the two invariants that motivate keeping `Int2Vector` as a separate type from `int2[]`.

See discussion in
https://github.com/MaterializeInc/materialize/pull/36151.

### DIFF
--- a/src/repr/src/scalar.rs
+++ b/src/repr/src/scalar.rs
@@ -2473,10 +2473,22 @@ impl<'a, E> OutputDatumType<'a, E> for Array<'a> {
     }
 }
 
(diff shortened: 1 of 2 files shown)


mgree (approved):
LGTM!

It's a bit of a shame that we can't keep treating the cast as eliminable, because it really is a noop---but in this case, we actually might need to track the type difference at the MIR level. (That said... we could push for a more nuanced repr type that treated int2vec as a subtype of a single-dimension array non-nullable int2s. Feels bad, though---suddenly repr types aren't exact anymore.)

PR #36383, opened 2026-05-04
compute: fix divide-by-zero in `LimitProgress` with zero slack

antiguru:
### Motivation

Fixes https://linear.app/materializeinc/issue/CLU-61.

Setting `compute_logical_backpressure_inflight_slack` to `0s` caused
the timely worker to panic with `attempt to divide by zero` at
`src/compute/src/render.rs:1780`, where `LimitProgress` rounds
observed timestamps via `time / slack_ms`.

### Description

Short-circuit the rounding step when `slack_ms == 0` and use the
observed time directly. When slack is zero, no rounding is meaningful,
so the operator just tracks raw observed timestamps as pending
capabilities — preserving the limit-progress and `upper`-gating
behavior without dividing by zero.

### Verification

New testdrive `test/testdrive/compute-logical-backpressure.td` runs
the exact reproducer from CLU-61 (counter source + table + MV under
`enable_compute_logical_backpressure = true` and
`compute_logical_backpressure_inflight_slack = '0s'`) and passes.

### DIFF
--- a/src/compute/src/render.rs
+++ b/src/compute/src/render.rs
@@ -1776,8 +1776,14 @@ where
                             .iter()
                             .flat_map(|(_, time, _)| u64::from(time).checked_add(slack_ms))
                         {
-                            let rounded_time =
-                                (time / slack_ms).saturating_add(1).saturating_mul(slack_ms);
+                            // `slack_ms == 0` means no rounding; otherwise round up to the next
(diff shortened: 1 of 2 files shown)


def- (approved):
Thanks for adding the test, lgtm

PR #36384, opened 2026-05-04
ore: ref-count DeleteOnDropMetric so cloning is safe

antiguru:
Fixes [CLU-63](https://linear.app/materializeinc/issue/CLU-63/replica-replacements-make-compute-metrics-disappear).

`DeleteOnDropMetric` derived `Clone`, but its `Drop` impl ran `remove_label_values` on the parent vec for every clone. Dropping any clone unregistered the metric from the Prometheus registry while other clones were still live and incrementing. The compute controller hit this on replica replacement: `mz_compute_controller_response_{send,recv}_count` permanently disappeared from `/metrics`, even though the in-memory counters kept going up.

### Description

Split the cleanup state out of `DeleteOnDropMetric` into a private `DeleteOnDropCleanup` struct held behind an `Arc`. The outer wrapper still derefs to the underlying Prometheus metric and still derives `Clone`, but cloning is now a refcount bump. `remove_label_values` runs exactly once, when the last clone drops.

### DIFF
--- a/src/ore/src/metrics/delete_on_drop.rs
+++ b/src/ore/src/metrics/delete_on_drop.rs
@@ -30,6 +30,7 @@
 use std::borrow::Borrow;
 use std::collections::BTreeMap;
 use std::ops::Deref;
+use std::sync::Arc;
 
 use prometheus::core::{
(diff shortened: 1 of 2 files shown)


def- (approved):
Thanks for adding the test, lgtm!

PR #36385, opened 2026-05-04
update-docs github workflow: Don't run on forks

def-:
I get a daily email with failed run on my fork: https://github.com/def-/materialize/actions/runs/25308646858


bosconi (approved):
Looks good. It can't even run from a fork, since it won't be given  `secrets.CLAUDE_CODE_OAUTH_TOKEN`.

def-:
Well, it still tries to run and fails.

PR #36387, opened 2026-05-04
Fix `string_to_array`'s handling of `null_string` when `delim` is empty

ggevay:
As it says on the tin. Fixes https://github.com/MaterializeInc/database-issues/issues/11351

Additionally:
- The 2. commit adds a bunch more tests, suggested by my AI agent.
- The 3. commit adds some more edge cases to our docs.


aljoscha (approved):
Sounds good to me, I assume you double checked with postgres on this one

PR #36389, opened 2026-05-04
Console maintained objects list

leedqin:
### Motivation
Maintained Objects List to show objects (sources, indexes, tables, mvs) in a table view to dive deeper into freshness (pMax), hydration status, clusters. 

First PR/ n for maintained objects feature. Fixes [CNS-50](https://linear.app/materializeinc/issue/CNS-50/objects-list-in-maintained-objects) 

### Description
- Maintained objects list wraps around TanStack table with column filters for pMax, hydration status, object type, clusters
- Added sql queries and tests for freshness lag and hydration status and did client side join with the data from allObjects subscribe
- Filter Chips component for the table to show which filters are active 
- Url params logic to save the filters in url params
- Expanded the tanstack table to have filters on the column headers

**Objects List**
image

**Filters for the objects list**
image

Objects filter:
image

pMax Filter:
image

Cluster Filter
image

### Tips for the reviewer
- Would recommend reviewing it by commit since the data/ query logic is in the first commit. And the rendering/ table filtering logic is in the second one

### DIFF
--- a/console/src/api/materialize/maintained-objects/lagAggregate.ts
+++ b/console/src/api/materialize/maintained-objects/lagAggregate.ts
@@ -1,0 +1,41 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+import { sql } from "kysely";
+
+import { IPostgresInterval, queryBuilder } from "~/api/materialize";
+
+export interface LagAggregateParams {
+  /** Lookback window in minutes for the temporal filter on lag history. */
+  lookbackMinutes: number;
+}
+
+export interface LagAggregateRow {
+  object_id: string;
+  lag: IPostgresInterval | null;
+}
+
+/**
+ * One row per object: the maximum (pMAX) lag observed in
+ * `mz_wallclock_global_lag_recent_history` over the lookback window.
+ */
+export function buildLagAggregateQuery({
+  lookbackMinutes,
+}: LagAggregateParams) {
+  return queryBuilder
+    .selectFrom("mz_wallclock_global_lag_recent_history")
+    .select(["object_id", sql<IPostgresInterval | null>`max(lag)`.as("lag")])
+    .where(
+      (eb) =>
+        sql`${eb.ref("occurred_at")} + INTERVAL '${sql.raw(`${lookbackMinutes}`)} MINUTES'`,
+      ">=",
+      sql<Date>`mz_now()`,
+    )
+    .groupBy("object_id");
+}
--- a/console/src/components/Table/tableTypes.ts
+++ b/console/src/components/Table/tableTypes.ts
@@ -8,7 +8,7 @@
 // by the Apache License, Version 2.0.
 
 import { HTMLChakraProps } from "@chakra-ui/react";
-import { RowData, Table } from "@tanstack/react-table";
+import { Column, RowData, Table } from "@tanstack/react-table";
 import React from "react";
 
 /**
@@ -33,6 +33,8 @@ export interface UniversalTableProps<TData> {
   isLoading?: boolean;
   /** Number of skeleton rows to display when loading. */
   skeletonRowCount?: number;
+  /** Chakra `sx` merged onto every body `<Tr>`. */
+  rowSx?: HTMLChakraProps<"tr">["sx"];
   /** `data-testid` forwarded to the root `<Table>` element. */
   "data-testid"?: string;
 }
@@ -66,7 +68,6 @@ export interface TableSearchProps {
  * even though we don't reference them here.
  */
 declare module "@tanstack/react-table" {
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
   interface ColumnMeta<TData extends RowData, TValue> {
     /** Tooltip shown on the column header. */
     tooltip?: string;
@@ -74,5 +75,16 @@ declare module "@tanstack/react-table" {
     minWidth?: Record<string, string> | string;
     /** Chakra props spread onto the `<Td>` for every cell in this column. */
     cellProps?: HTMLChakraProps<"td">;
+    /**
+     * Renders a per-column filter UI inside a popover anchored on the column
+     * header. When defined, `UniversalTable` shows a small filter trigger next
+     * to the sort caret (with an active-state dot when the column has a
+     * filter value applied). Receives `table` so the panel can pull shared
+     * data (e.g. dropdown options) from `table.options.meta`.
+     */
+    renderFilter?: (
+      column: Column<TData, TValue>,
+      table: Table<TData>,
+    ) => React.ReactNode;
   }
 }
--- a/console/src/config/flexibleDeploymentFlags.ts
+++ b/console/src/config/flexibleDeploymentFlags.ts
@@ -12,7 +12,9 @@
  * is too big, we default all flags to true and specify the ones we want
  * to disable.
  */
-export const disabledFlexibleDeploymentFlags: Record<string, boolean> = {};
+export const disabledFlexibleDeploymentFlags: Record<string, boolean> = {
+  "maintained-objects-ui-50": false,
+};
 
 export const flexibleDeploymentFlags = new Proxy(
   {},
--- a/console/src/platform/maintained-objects/filterPanels.tsx
+++ b/console/src/platform/maintained-objects/filterPanels.tsx
@@ -1,0 +1,264 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+import {
+  Box,
+  Button,
+  Checkbox,
+  HStack,
+  Input,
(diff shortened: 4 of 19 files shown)


console/src/api/materialize/maintained-objects/lagAggregate.ts
```diff
@@ -0,0 +1,41 @@
...
+  lag: IPostgresInterval | null;
+}
+
+/**
+ * One row per object: the maximum (pMAX) lag observed in
+ * `mz_wallclock_global_lag_recent_history` over the lookback window.
+ */
+export function buildLagAggregateQuery({
+  lookbackMinutes,
+}: LagAggregateParams) {
```
SangJunBak, line 30: nit: I kind of feel like we should be making sure we're validating that `lookbackMinutes` is less than a day on the API side.

console/src/platform/maintained-objects/queries.ts
```diff
@@ -0,0 +1,154 @@
...
+  lookbackMinutes: number;
+}): UseMaintainedObjectsListResult => {
+  const allObjects = useAllObjects();
+  const lag = useSubscribeToLagAggregate({ lookbackMinutes });
+  const hydration = useSubscribeToHydrationAggregate();
+
+  const data = React.useMemo<MaintainedObjectListItem[]>(() => {
+    const lagById = new Map(lag.data.map((r) => [r.object_id, r]));
+    const hydrationById = new Map(hydration.data.map((r) => [r.object_id, r]));
+    const num = (v: bigint | null | undefined) => (v ? Number(v) : 0);
```
SangJunBak: Let's just make this a top level function

console/src/platform/maintained-objects/queries.ts
```diff
@@ -0,0 +1,154 @@
...
+
+    return allObjects.data
+      .filter((obj) => {
+        if (!obj.id.startsWith("u")) return false;
```
SangJunBak: Let's reuse `isSystemId`

console/src/platform/maintained-objects/queries.ts
```diff
@@ -0,0 +1,154 @@
...
+        if (!MAINTAINED_OBJECT_TYPE_SET.has(obj.objectType)) return false;
+        // Subsources and progress sources are managed by their parent source.
+        return !(
+          obj.objectType === "source" &&
+          (obj.sourceType === "subsource" || obj.sourceType === "progress")
+        );
+      })
```
SangJunBak: Technically progress sources are now just the source and subsources are tables. Did we account for this?

console/src/platform/maintained-objects/queries.ts
```diff
@@ -0,0 +1,154 @@
...
+} from "~/api/materialize/useSubscribe";
+import { useAllObjects } from "~/store/allObjects";
+import { sumPostgresIntervalMs } from "~/util";
+
+export interface MaintainedObjectListItem {
+  id: string;
+  name: string;
+  schemaName: string | null;
+  databaseName: string | null;
+  objectType: MaintainedObjectType;
```
SangJunBak, line 52: Why are these nullable again?

console/src/platform/maintained-objects/queries.ts
```diff
@@ -0,0 +1,154 @@
...
+  clusterId: string | null;
+  clusterName: string | null;
+  /** 0 until the hydration snapshot arrives. */
```
SangJunBak, line 54: Kinda feels like these should be grouped into a `cluster` object, especially if both of them can be null

console/src/platform/maintained-objects/queries.ts
```diff
@@ -0,0 +1,154 @@
...
+  hydratedReplicas: number;
+  /** 0 until the hydration snapshot arrives. */
+  totalReplicas: number;
+  /** null until the lag snapshot arrives. */
+  lag: LagAggregateRow["lag"];
+  /** null until the lag snapshot arrives. */
+  lagMs: number | null;
+}
```
SangJunBak, line 60: Kinda feels like these should be grouped into a `lag` object, especially if both of them can be null

console/src/platform/maintained-objects/queries.ts
```diff
@@ -0,0 +1,154 @@
...
+export const useMaintainedObjectsList = ({
+  lookbackMinutes,
+}: {
+  lookbackMinutes: number;
+}): UseMaintainedObjectsListResult => {
+  const allObjects = useAllObjects();
+  const lag = useSubscribeToLagAggregate({ lookbackMinutes });
+  const hydration = useSubscribeToHydrationAggregate();
+
+  const data = React.useMemo<MaintainedObjectListItem[]>(() => {
```
SangJunBak, line 128: Nit: Rather than making the data structure a list, I like to make the "main" data structure an object / map keyed by the ID then derive a sorted list from it.So in this case, the ID would be object_id.

console/src/components/Table/tableTypes.ts
```diff
@@ -45,7 +47,7 @@ export interface TablePaginationProps<TData> {
 
 export interface TableSearchProps {
   /**
-   * Initial search value. Read once on mount; the component is uncontrolled
+   * Initial search value. Resad once on mount; the component is uncontrolled
```
SangJunBak: Read*

SangJunBak:
I think when we scroll, the search bar should be anchored to the left, similar to the query history

https://github.com/user-attachments/assets/cc3533de-862e-4cd6-9ff1-846ec10327d3

SangJunBak:
Nit: Margins on this chip seem a bit small. Is this consistent with our tags? 

Screenshot 2026-05-04 at 4 53 29 PM

SangJunBak:
Ux nit: It feels jarring to have this go to a number. Can we change this to like "2 selected"?

https://github.com/user-attachments/assets/80d1a47f-be28-4c37-b98e-5d4dd0d45d75

SangJunBak (commented):
Reviewed the first commit! Will review the second later

console/src/api/materialize/maintained-objects/lagAggregate.ts
leedqin: Will add a sql test for for this

leedqin:
image
would recommend to not review the filter chips component in the second commit and look at the third commit instead since that addresses the UX nits and show multi select filter as it's own chip to make it clear to the user which filters are selected

console/src/config/flexibleDeploymentFlags.ts
```diff
@@ -12,7 +12,9 @@
  * is too big, we default all flags to true and specify the ones we want
  * to disable.
  */
-export const disabledFlexibleDeploymentFlags: Record<string, boolean> = {};
+export const disabledFlexibleDeploymentFlags: Record<string, boolean> = {
+  "maintained-objects-ui-50": false,
+};
```
SangJunBak, line 17: What does the rollout look like for self managed? Wanted to call out that if we're aiming for releasing to self managed by next week, then there's no way a user can toggle htis

console/src/platform/maintained-objects/filterPanels.tsx
```diff
@@ -0,0 +1,227 @@
...
+export const FreshnessFilterPanel = ({ column }: { column: AnyColumn }) => {
+  const { colors } = useTheme<MaterializeTheme>();
+  const selected = column.getFilterValue() as number | undefined;
+
+  const setValue = (value: number | undefined) => {
+    column.setFilterValue(value);
+  };
+
+  return (
+    <RadioGroup
```
SangJunBak: style nit: Noticing the color scheme doesn't match our app's (it's blue instead of purple)
Image

SangJunBak:
Noticing an extra footer here:
Screenshot 2026-05-05 at 1 28 53 PM

console/src/api/materialize/maintained-objects/lagAggregate.ts
SangJunBak: Ah is this the test you're referring to https://github.com/MaterializeInc/materialize/pull/36389/changes/6e2ae07d8d55b5fc2aab93a638d2227f5a9717a2? 

Sorry I should've been more clear. So the underlying relation only has 24 hours of history. I was saying along with the client validation (of just hiding the option), maybe we should also validate closer to the API side (similar to how we validate the start time for the query history https://github.com/leedqin/materialize//blob/console-maintained-objects-list/console/src/api/materialize/query-history/queryHistoryList.ts#L164-L178). 

This is more of a nit because we're only exposing upto 24 hours as a lookback window on the client side. However it's a bit more useful if other hooks/pages try to reuse the `build` function. We don't need to do this however

SangJunBak:
It's weird to me how we change the color of freshness to orange for >=5 seconds but then our filters start at a minute

Screenshot 2026-05-05 at 1 47 46 PM

SangJunBak (commented):
Moreso found some design flaws/bugs, but everything else looks good to me!

SangJunBak:
Noticing we're persisting the lookback to local storage but the URL still blips. Why not just have it in the URL, especially if it doesn't sync to the lookback selects on other pages? 

https://github.com/user-attachments/assets/0da042b8-88ff-4edf-a1a8-080cb37cc834

SangJunBak:
I don't think the unit UI makes sense without an apply button, similar to the query history filters. E.g. if I were to filter for >=30 minutes

https://github.com/user-attachments/assets/b17bac84-da4a-45be-a51c-38a4b2bf407f

SangJunBak (changes requested):

leedqin:
https://github.com/user-attachments/assets/64e9f269-8954-4c3c-8855-4f828886d1b7

Added the Apply Filter for freshness

SangJunBak (approved):

PR #36391, opened 2026-05-04
ore: introduce mz_ore::pager with swap and file backends

antiguru:
### Motivation

CLU-65: introduce an explicit pager so we can mark cold columnar data and pull it back on demand, choosing at runtime between an `MADV_COLD`-based swap backend and a file-backed scratch backend. Today Materialize spills to Linux swap and pays direct-reclaim latency on user threads. Design lives at `doc/developer/design/20260504_pager.md`.

Depends on CLU-64 (`Column::Aligned` becomes `Vec<u64>`), which makes columnar buffers a natural caller.

### Description

`mz_ore::pager` exposes a small handle-oriented API (`pageout`, `read_at_many`, `read_at`, `take`) over a global atomic backend selector. The handle's backend is fixed at `pageout` time so live config flips do not invalidate existing data. Two backends:

* `Swap` keeps the input `Vec<u64>` chains resident and calls `MADV_COLD` on the page-aligned subrange. Reads `extend_from_slice` across chunk boundaries; `take` is zero-copy when the handle holds a single chunk.
* `File` writes one named scratch file per handle in a per-process subdirectory under the configured scratch root. No file descriptor is retained across calls — `pageout` opens, vector-writes via `Write::write_vectored`, and closes; reads reopen and `pread`. A reaper sweeps stale subdirectories from crashed predecessors at config time.

Documented as `doc/developer/design/20260504_pager.md`.

### Verification

Unit + integration tests cover round-trip on both backends, scatter/gather correctness, drop-without-take reclaim, the swap fast path's pointer identity, and backend flip mid-process.

Two micro-benches plus a merge-batcher example (`cargo run --release --features pager --example pager_merge`) for sizing the swap-vs-file trade-off under real memory pressure. Sanity numbers on a 32 GiB working set with 16 GiB cap, ext4 scratch: file finishes the merge pass in ~67 s vs swap's ~210 s, dominated by swap-in TLB shootdowns on the swap path.

### DIFF
--- a/doc/developer/design/20260504_pager.md
+++ b/doc/developer/design/20260504_pager.md
@@ -156,9 +156,11 @@ A future optimization is `preadv2` once a profile shows it matters.
 
 `set_scratch_dir(root)` creates `{root}/mz-pager-{pid}-{boot_nonce}/`.
 `boot_nonce` is a random 64-bit value sampled at config time so two processes that briefly share the same pid cannot collide.
-On `set_scratch_dir`, the pager runs a reaper that walks `{root}` and removes any sibling `mz-pager-*` subdirectory whose owning pid is no longer alive (`/proc/{pid}` missing on Linux); this reclaims storage from crashed predecessors.
-On clean process exit, a `Drop` on the global pager state removes the per-process subdirectory.
-The reaper is best-effort and logs failures rather than panicking.
+Per-handle files are unlinked when the handle is dropped or `take`n; the per-process subdirectory itself is **not** reclaimed by the pager.
+
+Reclaiming subdirectories left behind by crashed predecessors is out of scope for the in-process pager.
+In production, clusterd runs on Kubernetes with a per-pod ephemeral volume that is destroyed with the pod, so a crashed predecessor's scratch directory is reclaimed with the volume.
+For local development, sweeping leftover `mz-pager-*` directories is the responsibility of operator tooling (e.g. `bin/environmentd` cleanup or an explicit `rm -rf`).
 
 ### Configuration
 
--- a/doc/developer/design/20260504_pager.md
+++ b/doc/developer/design/20260504_pager.md
@@ -274,7 +274,6 @@ The file backend issues one `writev` per chunk on pageout and one `pread` per co
 * Pick the file backend whenever the working set may exceed RAM.
   The kernel I/O pipeline scales; swap-in does not.
 * The runtime atomic switch is the right place for an operator-level policy: a controller can flip the global at startup based on cluster size or under a pressure signal.
-* Prefetch hints (`prefetch` / `prefetch_at`) help the file backend by ~5 % at depth 16 on this workload; they do not help the swap backend because under pressure the kernel is reclaim-bound, not stall-bound.
 
 ## Alternatives
 
--- a/doc/developer/design/20260504_pager.md
+++ b/doc/developer/design/20260504_pager.md
@@ -223,11 +223,15 @@ miri runs the swap backend; the file backend skips on miri due to syscalls.
 
 A merge-batcher-style example (`src/ore/examples/pager_merge.rs`) builds two chains of 2 MiB chunks, then merges them while reading every cache line of the input.
 Run under `systemd-run --user --scope -p MemoryMax=...` to constrain memory and force real eviction.
-Numbers below were collected on a Linux box with an encrypted NVMe (~1.4 GB/s sustained R+W ceiling) running the example with `--chain-gib 16` (32 GiB total working set) and `--prefetch-depth 4`.
 
-### Throughput sweep
-
+Pager behavior depends sharply on disk topology.
+Slow-storage boxes hit a kernel reclaim ceiling well below the disk and the swap backend looks catastrophic; fast-storage boxes expose new bottlenecks and the swap-vs-file gap collapses at high thread counts.
+Two benches below: a single encrypted NVMe (~1.4 GB/s ceiling, akin to typical EBS-backed clusters) and an r8gd.16xlarge with two striped local instance NVMes (~7 GB/s ceiling).
 `through` is total bytes pumped through the merge divided by wall time, summed across threads.
+
+### Bench A: encrypted NVMe, 32 GiB working set
+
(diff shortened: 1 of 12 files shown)


src/ore/src/pager.rs
```diff
@@ -0,0 +1,207 @@
...
+        Backend::File => BACKEND_FILE,
+    };
+    BACKEND.store(raw, Ordering::Relaxed);
+}
+
+/// Scatter pageout. Logical layout = chunks concatenated in order.
+/// After return, each `Vec` in `chunks` is empty.
+/// File backend preserves capacity; swap backend moves the alloc into the handle.
+/// Empty input returns a `len == 0` handle and performs no I/O.
+pub fn pageout(chunks: &mut [Vec<u64>]) -> Handle {
```
antiguru, line 140: Instead of a Handle, this could be a `Box<dyn Fetch>`.

antiguru:
Review findings (anchor: `src/compute/src/compute_state.rs::apply_worker_config` — config flip site for pager backend, not in this diff but the relevant integration point):

### Critical

* **`src/ore/src/pager/file.rs` `set_scratch_dir`**: `SCRATCH_DIR.set(root)` runs even when `init_subdir` fails, leaving `SUBDIR` unset. Later `scratch_path` calls panic on `SUBDIR.get().expect(...)`. `call_once` also burns the only retry. Fix: set `SCRATCH_DIR` only on success, or propagate the error.
* **`src/ore/src/pager/file.rs` `reap_stale`**: liveness probe uses `/proc/{pid}`, which is Linux-only. On macOS the path never exists, so the reaper deletes live concurrent processes' subdirs. Use a portable probe (e.g. `kill(pid, 0)` → `ESRCH`) or gate scratch use on `target_os = "linux"`.
* **`reap_stale` PID-reuse race**: recycled PID keeps a dead owner's subdir alive forever (or a live unrelated owner protects an orphan). Embed a creation timestamp + grace window, or use a lockfile.
* **Module is not `cfg(unix)`-gated**: `file.rs` uses `std::os::unix::fs::FileExt::read_at`. Builds fail on Windows targets when the `pager` feature is enabled. Add `#[cfg(unix)]` or mark the feature Unix-only in Cargo.toml.
* **Reaper runs once per process** (inside `SCRATCH_INIT.call_once`). Long-lived processes never reclaim subdirs whose owners die later. Sweep periodically or document the one-shot behavior prominently.

### Important

* **`pageout_file` empty input** returns `Handle::from_swap(...)` even when backend = `File`. Contradicts the documented invariant that the variant is fixed at `pageout` time. Either return a `File`-variant zero-byte handle or document the empty-input exception.
* **`read_at_file` overflow**: `byte_off = u64::cast_from(off * 8)` and `byte_len = len * 8` multiply in `usize` *before* the cast. On 32-bit targets this overflows silently in release. Promote to `u64` before the `* 8`.
* **`read_at_file` panic on `open()`**: panics if the inode was unlinked (background tooling, NFS, reaper misclassification). Either return an error or document this as a hard-panic invariant; consider caching a read-only fd.
* **All `read_at*` / `take_*` panic on I/O error**. For a pager backing batchers under memory pressure, this kills the world on transient FS errors. Add fallible `try_*` variants.
* **`src/ore/Cargo.toml` `pager` feature**: mixes `dep:` prefix styles — `["dep:bytemuck", "libc", "rand", "dep:tracing"]`. `libc` / `rand` without `dep:` create implicit features. Use `dep:` consistently or none.
* **`swap.rs::madvise_cold`** arithmetic: `base_addr + page - 1` and `base_addr + len_bytes` lack checked-add. Wraps if the allocation sits at the top of address space. Use `checked_add` / `saturating_add` and expand the SAFETY comment to justify why aligning down can't push the range outside the Vec.
* **`SwapInner::new` / `total_len` / `Handle::len`**: `*prefix.last().unwrap_or(&0)` masks an invariant (`prefix` always carries at least `[0]`). Replace with direct indexing or `expect`, so future code that violates the invariant surfaces immediately.
* **`set_scratch_dir`** silently ignores a second call with the same path even if the first init failed. Combined with the bug above, a transient failure permanently disables the file backend. Retry on subsequent calls when init failed, or document "first call wins" loudly.
* **`read_at_file` short-read panic** message: `filled` is cumulative *within the current coalesced range*, not across the whole call; the message omits `path` and `offset`. Include both for diagnosis.

### API design (response to author's own note about `Box<dyn Fetch>`)

Keep the current `Handle { inner: HandleInner }` enum.
A `Box<dyn Fetch>` forces a heap allocation per handle, defeats the swap fast-path (`take_swap` `mem::take`s the underlying `Vec<u64>` zero-copy — a vtable can't replicate that without an allocation contract), and gains nothing while there are only two backends.

doc/developer/design/20260504_pager.md
```diff
@@ -0,0 +1,413 @@
...
+  We will revisit if a non-`u64` consumer appears.
+* Compression and encryption.
+  Both can wrap the pager externally.
+* Async or `io_uring`.
+  Sync syscalls only for v1.
+  An async wrapper is a follow-up if profiles motivate it.
+* Cross-handle file pooling (one shared scratch file with a free-list).
+  Each handle owns one named file in the scratch directory; pooling is a follow-up if inode pressure shows up.
+* Non-Linux production support.
+  Both backends compile on macOS and others as in-memory no-ops, but the production target is Linux.
```
DAlperin: file can be real on macos :)

src/ore/src/pager/file.rs
```diff
@@ -0,0 +1,486 @@
...
+        }
+        IoSlice::advance_slices(&mut slices, written);
+    }
+    Ok(())
+}
+
+/// Fallible counterpart to [`read_at_file`]. Returns `Err` instead of
+/// panicking on I/O failure (open/pread errors). Bounds violations and
+/// non-file handles still panic — those are caller bugs, not I/O failures.
+pub(crate) fn try_read_at_file(
```
DAlperin, line 193: reading this, and then reading upwards, it's not clear to me what our intention is if someone were to pass overlapping ranges. I think it probably shouln't be allowed (we don't handle it right here) and we should check for it upstream in the pager api.

DAlperin (approved):
LGTM, thanks!

antiguru:
Thanks for the review!

PR #36392, opened 2026-05-04
[SSO] Wire JWT group-to-role sync into login and add edge case handling

mtabebe:
Problem:
Users authenticating via OIDC had no automatic mechanism to sync their JWT group claims to Materialize role memberships. Operators had to manually assign roles to each user, which didn't scale as group membership changed in the identity provider.

Solution:
- Call sync_jwt_groups in handle_startup_inner after role auto-provisioning when the user has OIDC groups and oidc_group_role_sync_enabled is true. Resolves group names to role IDs (case-insensitive) via try_get_role_by_name_case_insensitive on CatalogState.

- Filter reserved role names (mz_/pg_ prefixes, PUBLIC) with warnings.
- Strict mode: reject login on sync failure when oidc_group_role_sync_strict=true.
- Fail-open mode (default): log warning and continue on sync failure.
- Send client NOTICEs for unmatched groups, reserved roles, and sync errors.
- Pass notice_tx to handle_startup_inner for pre-session notice delivery.
- Missing groups claim (None) skips sync; empty claim (Some([])) revokes all.

Test:
- Unit tests
- Small integration test

Fixes: SQL-181, SQL-182

### DIFF
--- a/src/adapter/src/coord/group_sync.rs
+++ b/src/adapter/src/coord/group_sync.rs
@@ -16,14 +16,19 @@
 
 use std::collections::{BTreeMap, BTreeSet};
 
+use mz_adapter_types::dyncfgs::{OIDC_GROUP_ROLE_SYNC_ENABLED, OIDC_GROUP_ROLE_SYNC_STRICT};
 use mz_repr::role_id::RoleId;
 use mz_sql::session::user::MZ_JWT_SYNC_ROLE_ID;
+use tokio::sync::mpsc;
+use tracing::{info, warn};
 
-use crate::catalog::Op;
+use crate::AdapterError;
+use crate::catalog::{self, Op};
+use crate::coord::Coordinator;
+use crate::notice::AdapterNotice;
 
 /// Result of computing the group-to-role membership sync diff.
 #[derive(Debug, Clone)]
-#[allow(dead_code)] // used in tests; will be consumed by login path in a follow-up
 pub struct GroupSyncDiff {
     /// Roles to grant to the user (with sentinel grantor).
     pub grants: Vec<Op>,
@@ -38,9 +43,8 @@ pub struct GroupSyncDiff {
 /// - `member_id`: The user's role ID.
 /// - `current_membership`: The user's current `RoleMembership.map`
 ///   (role_id → grantor_id).
-/// - `target_role_ids`: Role IDs resolved from the JWT group names. Callers must
-///   lowercase JWT group names before catalog lookup since SQL role names are stored
-///   lowercase.
+/// - `target_role_ids`: Role IDs resolved from the JWT group names via
+///   case-insensitive catalog lookup.
 ///
 /// # Semantics
 /// - Only roles granted by the JWT sync sentinel (`MZ_JWT_SYNC_ROLE_ID`)
@@ -48,7 +52,6 @@ pub struct GroupSyncDiff {
 /// - Manually-granted roles (grantor != sentinel) are never revoked.
 /// - If a target role is already manually granted, it is skipped — the
 ///   manual grant takes precedence and we don't overwrite the grantor.
-#[allow(dead_code)] // used in tests; will be called from login path in a follow-up
 pub fn compute_group_sync_diff(
     member_id: RoleId,
     current_membership: &BTreeMap<RoleId, RoleId>,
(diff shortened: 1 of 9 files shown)


def- (commented):
A JWT groups claim that contains the user's own role name (common when an IdP echoes preferred_username/email into groups, or with per-user groups) makes the resolver map that group back to the user's own role, producing `Op::GrantRole { role_id == member_id }`. The grant handler's circular-membership check (`collect_role_membership(&role_id)` always includes `role_id` itself) then treats this as a cycle and `catalog_transact` rejects the whole batch with `CircularRoleMembership`.

Is that intentional? Here's a small test for it in case it's not:
```diff
diff --git a/src/environmentd/tests/auth.rs b/src/environmentd/tests/auth.rs
index c3f719feac..2c3ab0707c 100644
--- a/src/environmentd/tests/auth.rs
+++ b/src/environmentd/tests/auth.rs
@@ -5519,3 +5519,22 @@ async fn test_oidc_group_sync_first_login() {
         assert_eq!(grantor, "mz_jwt_sync", "grantor should be mz_jwt_sync");
     }
 }
+
+#[mz_ore::test(tokio::test(flavor = "multi_thread", worker_threads = 1))]
+#[cfg_attr(miri, ignore)] // unsupported operation: can't call foreign function `TLS_method`
+async fn test_oidc_group_sync_self_named_group() {
+    let (server, admin_client, oidc_server) = setup_group_sync_test().await;
+    admin_client
+        .batch_execute("ALTER SYSTEM SET oidc_group_role_sync_strict = true")
+        .await
+        .unwrap();
+    let token = jwt_with_groups(
+        &oidc_server,
+        serde_json::json!(["Alice@Example.com", "analytics"]),
+    );
+    oidc_connect(&server, &token).await.unwrap();
+    assert_eq!(
+        fetch_user_role_memberships(&admin_client, GROUP_SYNC_USER).await,
+        vec!["analytics"],
+    );
+}
```

src/oidc-mock/src/lib.rs
```diff
@@ -96,8 +96,11 @@ pub struct GenerateJwtOptions<'a> {
     /// Audience claim. If None, uses empty array.
     /// Use `AudClaim::Single` for a single string or `AudClaim::Multiple` for an array.
     pub aud: Option<AudClaim>,
-    /// Additional claims.
+    /// Additional claims as string values.
```
SangJunBak: Maybe we should combine unknown_claims with extra_claims given `unknown_claims`'s values get mapped to a `serde_json::Value::String)` anyways?

src/environmentd/tests/auth.rs
```diff
@@ -5428,3 +5428,154 @@ async fn test_auth_oidc_non_login_role() {
...
+        GenerateJwtOptions {
+            extra_claims: Some(BTreeMap::from([("groups".to_string(), groups)])),
+            ..Default::default()
+        },
+    )
+}
+
+/// First login grants roles from JWT groups, with correct grantor and audit log entries.
+#[mz_ore::test(tokio::test(flavor = "multi_thread", worker_threads = 1))]
+#[cfg_attr(miri, ignore)] // unsupported operation: can't call foreign function `TLS_method`
```
SangJunBak, line 5545: Solid happy path test! Could we also create tests for the unhappy paths? Specifically thinking
1. Group sync fails and we're in strict mode, we should get an error
2. Group sync fails and we're not in strict mode, we should get a notice. Can see https://github.com/MaterializeInc/materialize/blob/main/src/environmentd/tests/server.rs#L4345 as an example

src/environmentd/tests/auth.rs
```diff
@@ -5428,3 +5428,154 @@ async fn test_auth_oidc_non_login_role() {
...
+const GROUP_SYNC_USER: &str = "alice@example.com";
+
+fn make_insecure_tls() -> postgres_openssl::MakeTlsConnector {
+    make_pg_tls(Box::new(|b: &mut SslConnectorBuilder| {
+        Ok(b.set_verify(SslVerifyMode::NONE))
+    }))
+}
+
+/// Helper: connect as the OIDC user with the given token.
+async fn oidc_connect(
```
SangJunBak, line 5517: So in our above tests, you might notice that we also test against our HTTP/WS endpoints as well! 

I'm not sure if having the sync happen for http/ws connections to the adapter client was part of the initial spec, but I think your change here also works for those too since the HTTP/WS endpoints converge to the same path. Might be worth validating and if it already works, might be worth adding tests! Otherwise we can follow up on it

src/adapter/src/coord/group_sync.rs
```diff
@@ -91,9 +94,132 @@ pub fn compute_group_sync_diff(
...
+                    group = group.as_str(),
+                    "OIDC group maps to reserved role name, skipping"
+                );
+                notices.push(AdapterNotice::OidcGroupSyncReservedRole {
+                    group: group.clone(),
+                });
+                continue;
+            }
+
+            match self.catalog().try_get_role_by_name_case_insensitive(group) {
```
SangJunBak: Where `n` is the number of roles, it seems like we're doing an O(n^2) operation when calling `try_get_role_by_name_case_insensitive` per group. I wonder if it makes sense to create a copy of the data structure outside of this for loop then do a lookup on the copy?

SangJunBak (changes requested):

src/adapter/src/coord/group_sync.rs
mtabebe: Good point!

src/adapter/src/coord/group_sync.rs
```diff
@@ -162,6 +162,9 @@ impl Coordinator {
         groups: &[String],
         notices: &mut Vec<AdapterNotice>,
     ) -> Result<(), AdapterError> {
+        // Build a lowercase name → role map once to avoid O(n) catalog scan per group.
```
SangJunBak: Nit: Feel like this comment is unnecessary

SangJunBak (approved):
LGTM!

PR #36399, opened 2026-05-05
feature-benchmark: Ignore memory regression in Retraction scenario

def-:
See https://linear.app/materializeinc/issue/CLU-66

Test run: https://buildkite.com/materialize/nightly/builds/16307


antiguru (approved):
Thank you!

PR #36401, opened 2026-05-05
Standardize name resolution in COPY FROM and AWS CONNECTION

martykulma:
Updates the HTTP clients used by COPY FROM and AWS CONNECTION to use `mz_ore::netio::resolve_address` by building the clients with custom resolvers. 

This also updates the implementation of `addr.is_global()` to cover additional cases and adds unit tests.

I've split the changes into separate commits to make it easier to review.

Addresses [SQL-193](https://linear.app/materializeinc/issue/SQL-193)

### DIFF
--- a/src/ore/src/netio/dns.rs
+++ b/src/ore/src/netio/dns.rs
@@ -68,13 +68,40 @@ pub async fn resolve_address(
 }
 
 fn is_global(addr: IpAddr) -> bool {
-    // TODO: Switch to `addr.is_global()` once stable: https://github.com/rust-lang/rust/issues/27709
+    // TODO: Switch to `addr.is_global()` once stable:
+    // https://github.com/rust-lang/rust/issues/27709
     match addr {
-        IpAddr::V4(ip) => {
-            !(ip.is_unspecified() || ip.is_private() || ip.is_loopback() || ip.is_link_local())
+        IpAddr::V4(ip) => is_global_v4(ip),
+        IpAddr::V6(ip) => {
+            // Treat IPv4-mapped IPv6 addresses (`::ffff:a.b.c.d`) as their
+            // underlying IPv4 address — connecting to `::ffff:127.0.0.1`
+            // reaches loopback on dual-stack sockets.
+            if let Some(v4) = ip.to_ipv4_mapped() {
+                return is_global_v4(v4);
+            }
+            let segments = ip.segments();
+            !(ip.is_loopback()
+                || ip.is_unspecified()
+                // link-local (fe80::/10)
+                || (segments[0] & 0xffc0) == 0xfe80
+                // unique local (fc00::/7)
+                || (segments[0] & 0xfe00) == 0xfc00)
         }
-        IpAddr::V6(ip) => !(ip.is_loopback() || ip.is_unspecified()),
     }
+}
+
+fn is_global_v4(ip: std::net::Ipv4Addr) -> bool {
+    let octets = ip.octets();
+    !(ip.is_unspecified()
+        || ip.is_private()
+        || ip.is_loopback()
+        || ip.is_link_local()
+        || ip.is_broadcast()
+        || ip.is_documentation()
+        // shared address space / carrier-grade NAT (100.64.0.0/10)
+        || (octets[0] == 100 && (octets[1] & 0xc0) == 64)
+        // benchmarking (198.18.0.0/15)
+        || (octets[0] == 198 && (octets[1] & 0xfe) == 18))
 }
 
 #[cfg(test)]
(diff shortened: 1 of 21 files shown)


src/ore/src/netio/dns.rs
```diff
@@ -68,11 +68,149 @@ pub async fn resolve_address(
...
+            // underlying IPv4 address — connecting to `::ffff:127.0.0.1`
+            // reaches loopback on dual-stack sockets.
+            if let Some(v4) = ip.to_ipv4_mapped() {
+                return is_global_v4(v4);
+            }
+            let segments = ip.segments();
+            !(ip.is_loopback()
+                || ip.is_unspecified()
+                // link-local (fe80::/10)
+                || (segments[0] & 0xffc0) == 0xfe80
```
patrickwwbutler: nit: possibly worth encoding these literal values as constants, IMO it's a bit neater than having the comments explaining and also allows re-use

src/ore/src/netio/dns.rs
```diff
@@ -68,11 +68,149 @@ pub async fn resolve_address(
...
+fn is_global_v4(ip: std::net::Ipv4Addr) -> bool {
+    let octets = ip.octets();
+    !(ip.is_unspecified()
+        || ip.is_private()
+        || ip.is_loopback()
+        || ip.is_link_local()
+        || ip.is_broadcast()
+        || ip.is_documentation()
+        // shared address space / carrier-grade NAT (100.64.0.0/10)
+        || (octets[0] == 100 && (octets[1] & 0xc0) == 64)
```
patrickwwbutler: more of the slightly spicy "magic number" code smell, unfortunately looks like IP handing is loaded with these, I won't block on it though, plus bitwise operations are always a bit scary

patrickwwbutler (approved):
Seems reasonable to me, with a nit about magic numbers, but nbd. 

It does seem like it's probably worth testing this maybe locally with real HTTP/AWS (and possibly GCS?) connections, do make sure it plays well with those, and not just the minio environment/links

def- (commented):
With this new test:
```diff
diff --git a/test/testdrive/connection-create-external.td b/test/testdrive/connection-create-external.td
index 56dc994c62..37d913f49a 100644
--- a/test/testdrive/connection-create-external.td
+++ b/test/testdrive/connection-create-external.td
@@ -70,5 +70,26 @@ $ kafka-ingest format=avro topic=csr_test key-format=avro key-schema=${keyschema
   ) WITH (VALIDATE = true);
 contains:Address resolved to a private IP

+$ postgres-execute connection=postgres://mz_system:materialize@${testdrive.materialize-internal-sql-addr}
+ALTER SYSTEM SET enable_copy_from_remote = true
+
+> CREATE TABLE copy_ssrf_target (a text, b text)
+
+! COPY INTO copy_ssrf_target FROM 'http://127.0.0.1:1/file.csv' (FORMAT CSV);
+contains:Address resolved to a private IP
+
+! COPY INTO copy_ssrf_target FROM 'http://[::1]:1/file.csv' (FORMAT CSV);
+contains:Address resolved to a private IP
+
+> CREATE SECRET aws_ssrf_secret AS 'unused'
+
+! CREATE CONNECTION aws_ssrf_endpoint TO AWS (
+    ACCESS KEY ID = 'unused',
+    SECRET ACCESS KEY = SECRET aws_ssrf_secret,
+    ENDPOINT = 'http://127.0.0.1:1/',
+    REGION = 'us-east-1'
+  ) WITH (VALIDATE = true);
+contains:Address resolved to a private IP
+
 $ postgres-execute connection=postgres://mz_system:materialize@${testdrive.materialize-internal-sql-addr}
 ALTER SYSTEM SET storage_enforce_external_addresses = false
```
By running `bin/mzcompose --find testdrive down && bin/mzcompose --find testdrive run default connection-create-external.td` I'm getting:
```
> COPY INTO copy_ssrf_target FROM 'http://127.0.0.1:1/file.csv' (FORMAT CSV);
expected error containing "Address resolved to a private IP", got "COPY FROM: error: reqwest error: error sending request for url (http://127.0.0.1:1/file.csv)
causes: [\"HEAD request\", \"http\", \"list\", \"split\", \"fetch work\", \"decode chunk\", \"stage batches\"]
"
query error didn't match; sleeping to see if dataflow produces error shortly 🕑 50ms
^^^ +++
connection-create-external.td:78:1: error: expected error containing "Address resolved to a private IP", got "COPY FROM: error: reqwest error: error sending request for url (http://127.0.0.1:1/file.csv)
causes: [\"HEAD request\", \"http\", \"list\", \"split\", \"fetch work\", \"decode chunk\", \"stage batches\"]
"
     |
  16 | $ postgres-execute c ... [rest of line truncated for security]
  69 |     URL '${testdrive ... [rest of line truncated for security]
  73 | $ postgres-execute c ... [rest of line truncated for security]
  77 |
  78 | ! COPY INTO copy_ssrf_target FROM 'http://127.0.0.1:1/file.csv' (FORMAT CSV);
     | ^
+++ !!! Error Report
files involved: connection-create-external.td
```
So it seems like you can just use `127.0.0.1` instead of localhost now to get around `enforce_external_addresses=true`

martykulma:
> With this new test:

Great catch @def-, I'll revise!

src/ore/src/netio/dns.rs
martykulma: Constants are probably workable in the IPv6 cases because we're dealing with a u16.  Becomes a bit of a mess in IPv4, where we have to check 2 octets per address.

These functions are an exception, so I'm ok leaving them as hex.  We should not be rolling our own IP masking. This is only being done here because rust hasn't shipped a stable global check yet (it is in nightly).

def- (approved):
Thanks for adding the test. No other concerns from QA side

src/ore/src/netio/dns.rs
martykulma: @jubrad clued me into ipnet, which allows deny lists - so this should now be clearer.

jubrad (approved):

PR #36403, opened 2026-05-05
adapter: thread bare ReadHolds through ComputeController::peek

aljoscha:
Before, slow-path peek dataflows piggy-backed on the adapter machinery
meant for long-lived objects. We'd call
`initialize_compute_read_policies(DisableCompaction)` on the transient
index, which set `ReadPolicy::ValidFrom(as_of)` on the compute collection
and parked an entry in `TimelineState.read_holds` (the structure that
tracks holds for catalog objects). Then, inside
`ComputeController::peek`, the compute controller would acquire its own
hold for the duration of the peek. After the peek finished the adapter
had to remember to clean the timeline entry back up.

That dance was redundant -- `Instance::add_collection` already installs
an implied hold pinning the new collection's `since` at `as_of`, so the
`ValidFrom` policy and the timeline entry weren't actually buying us
anything -- and it had a latent leak: on the peek error path the
timeline entry was never released.

Now the adapter acquires a single bare `ReadHold` for the peek target
and threads it through `ComputeController::peek`. No `TimelineState`
bookkeeping, no second acquire inside the controller, and the leak is
gone by construction. The fast-path branches (`PeekExisting`,
`PeekPersist`) acquire and thread their own holds the same way for
uniformity.

`ComputeController::peek` now requires a caller-supplied `ReadHold`
instead of acquiring one itself. The id-mismatch and
`since <= timestamp` checks both surface as graceful `PeekError`
variants (`ReadHoldIdMismatch`, `SinceViolation`) rather than panics.
On the slow path the adapter additionally `soft_assert`s -- with a
runtime fallback that returns `AdapterError::internal` -- that the
dataflow exports exactly the one transient index id we acquire a hold
for; this documents the single-output assumption baked into the new
hold strategy. The dead `remove_compute_ids_from_timeline` helper --
slow-path peek was its only caller -- goes away.

### DIFF
--- a/src/adapter/src/coord/peek.rs
+++ b/src/adapter/src/coord/peek.rs
@@ -20,7 +20,6 @@ use std::sync::Arc;
 
 use differential_dataflow::consolidation::consolidate;
 use itertools::Itertools;
-use mz_adapter_types::compaction::CompactionWindow;
 use mz_adapter_types::connection::ConnectionId;
 use mz_cluster_client::ReplicaId;
 use mz_compute_client::controller::PeekNotification;
@@ -36,6 +35,8 @@ use mz_expr::{
     RowSetFinishingIncremental, permutation_for_arrangement,
 };
 use mz_ore::cast::CastFrom;
+use mz_ore::collections::CollectionExt;
+use mz_ore::soft_assert_eq_or_log;
 use mz_ore::str::{StrExt, separated};
 use mz_ore::task;
 use mz_ore::tracing::OpenTelemetryContext;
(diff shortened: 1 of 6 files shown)


antiguru:
Looked cursery, and seems fine on a high level. We certainly need to stress-test this code path as we've had surprising errors in the neighborhood.

src/adapter/src/coord/peek.rs
```diff
@@ -731,94 +731,124 @@ impl crate::coord::Coordinator {
...
-                desc: dataflow,
-                // n.b. this index_id identifies a transient index the
-                // caller created, so it is guaranteed to be on
-                // `compute_instance`.
-                id: index_id,
-                key: index_key,
-                permutation: index_permutation,
-                thinned_arity: index_thinned_arity,
-            }) => {
-                let output_ids = dataflow.export_ids().collect();
```
ggevay, line 834: The new code is not looking at `export_ids`. This is kinda ok, because `export_ids` should be just the one `index_id` currently, but would be a problem if we ever wanted to make multi-output dataflows work. We could maybe add at least a soft assert here that `export_ids` has exactly one element, and is the same as `index_id`.

src/compute-client/src/controller.rs
```diff
@@ -878,14 +883,13 @@ impl ComputeController {
...
-                .acquire_read_holds(vec![*id])?
-                .into_element(),
-        };
+        // Validation: the caller-supplied read hold must be for the peek target and must hold
+        // back compaction past the peek timestamp. The id check is a caller-contract assertion.
+        assert_eq!(
+            read_hold.id(),
+            peek_target.id(),
+            "read hold id must match peek target id",
+        );
```
ggevay: `Instance::peek` handles the same problem by gracefully returning a `PeekError::ReadHoldIdMismatch` (that's a different `PeekError`). Maybe we could do the same here as well. (Nowadays, we are moving in the direction of being less panicky.)

ggevay (approved):
LGTM, just minor comments.

> `ComputeController::peek` now requires a caller-supplied `ReadHold` instead of acquiring one internally. 

I like it that we are moving in the direction of passing in read holds rather than reacquiring them at lower levels. The frontend peek also did this at other places. (Btw. the [doc comment](https://github.com/MaterializeInc/materialize/blob/fcc110b5fe019c992bb11ac07eb10c22e28813ce/src/adapter/src/peek_client.rs#L221-L224) of the frontend peek's `implement_fast_path_peek_plan` kind of foreshadowed this change to acquire a read hold also on the target ID of slow-path peek dataflows higher up the call chain.)

One thing to note is that `implement_peek_plan`'s fast path code path is nowadays has very little testing, because we have the frontend peek sequencing on by default, and fast path peeks are handled in `implement_fast_path_peek_plan`. But it's not totally unused, because of those corner cases where the old peek sequencing is still called even when the frontend peek sequencing is turned on. So, if you want to be super-thorough, maybe you could run CI also with `enable_frontend_peek_sequencing` disabled to test this part of the code.

PR #36405, opened 2026-05-05
ci: Disable Kafka Auth test entirely

def-:
Flaked again! https://buildkite.com/materialize/test/builds/122316

I'll try to bisect in parallel. Edit: not much success, doesn't fail for me at all, even in an endless loop :/


martykulma (approved):
lgtm

PR #36406, opened 2026-05-05
[ontology] extend ontology views to cover BuiltinLog / mz_introspection

mtabebe:
Problem:

The ontology views (mz_ontology_entity_types etc.) were generated by iterating builtins with the ontology type, but BuiltinLog entries were skipped.

Therefore, the Compute introspection logs in mz_introspection had no representation in the ontology

Solution:
- Add ontology: Option<Ontology> field to BuiltinLog and annotate all 32 log entries with entity names, descriptions, FK/measures links, and semantic type annotations on GlobalId/MzTimestamp columns.
- Extend generate_views() with a Builtin::Log arm

Testing:
- SLT smoke tests in mz_ontology.slt verify that per-worker entities appear in entity_types with correct keys and relations, that measures links resolve
- Unit tests cover the Log inclusion/exclusion path in generate_views.

### DIFF
--- a/src/catalog/src/builtin.rs
+++ b/src/catalog/src/builtin.rs
@@ -8323,25 +8323,60 @@ UNION ALL
         ontology: Some(Ontology {
             entity_name: "object",
             description: "Union of all object types: relations, indexes, connections, etc. (convenience view)",
-            links: &const { [
-                OntologyLink { name: "union_includes", target: "relation", properties: LinkProperties::Union {
-     discriminator_column: None,
-     discriminator_value: None,
-     note: Some("mz_objects includes all relations plus indexes, connections, secrets, types, functions"),
- } },
-                OntologyLink { name: "union_includes", target: "index", properties: LinkProperties::union_disc("type", "index") },
-                OntologyLink { name: "union_includes", target: "connection", properties: LinkProperties::union_disc("type", "connection") },
-                OntologyLink { name: "union_includes", target: "secret", properties: LinkProperties::union_disc("type", "secret") },
-                OntologyLink { name: "maps_to_global_id", target: "object", properties: LinkProperties::MapsTo {
-     source_column: None,
-     target_column: None,
-     via: Some("mz_internal.mz_object_global_ids"),
-     from_type: Some(mz_repr::SemanticType::CatalogItemId),
-     to_type: Some(mz_repr::SemanticType::GlobalId),
-     note: Some("A CatalogItemId (SQL layer) maps to one or more GlobalIds (runtime layer)."),
- } },
-            ] },
-            column_semantic_types: &const {[("id", SemanticType::CatalogItemId), ("oid", SemanticType::OID), ("schema_id", SemanticType::SchemaId), ("type", SemanticType::ObjectType), ("owner_id", SemanticType::RoleId), ("cluster_id", SemanticType::ClusterId)]},
+            links: &const {
+                [
+                    OntologyLink {
+                        name: "union_includes",
+                        target: "relation",
+                        properties: LinkProperties::Union {
+                            discriminator_column: None,
+                            discriminator_value: None,
+                            note: Some(
+                                "mz_objects includes all relations plus indexes, connections, secrets, types, functions",
+                            ),
+                        },
+                    },
+                    OntologyLink {
+                        name: "union_includes",
+                        target: "index",
+                        properties: LinkProperties::union_disc("type", "index"),
+                    },
+                    OntologyLink {
+                        name: "union_includes",
+                        target: "connection",
+                        properties: LinkProperties::union_disc("type", "connection"),
+                    },
+                    OntologyLink {
+                        name: "union_includes",
+                        target: "secret",
+                        properties: LinkProperties::union_disc("type", "secret"),
+                    },
+                    OntologyLink {
+                        name: "maps_to_global_id",
+                        target: "object",
+                        properties: LinkProperties::MapsTo {
+                            source_column: None,
+                            target_column: None,
+                            via: Some("mz_internal.mz_object_global_ids"),
+                            from_type: Some(mz_repr::SemanticType::CatalogItemId),
+                            to_type: Some(mz_repr::SemanticType::GlobalId),
+                            note: Some(
+                                "A CatalogItemId (SQL layer) maps to one or more GlobalIds (runtime layer).",
+                            ),
+                        },
+                    },
+                ]
+            },
+            column_semantic_types: &const {
+                [
+                    ("id", SemanticType::CatalogItemId),
+                    ("oid", SemanticType::OID),
+                    ("schema_id", SemanticType::SchemaId),
+                    ("type", SemanticType::ObjectType),
+                    ("owner_id", SemanticType::RoleId),
+                    ("cluster_id", SemanticType::ClusterId),
+                ]
+            },
         }),
     }
 });
--- a/src/catalog/src/builtin.rs
+++ b/src/catalog/src/builtin.rs
@@ -2175,7 +2175,12 @@ pub static MZ_DATAFLOW_ADDRESSES_PER_WORKER: LazyLock<BuiltinLog> = LazyLock::ne
             [OntologyLink {
                 name: "address_of",
                 target: "dataflow_operator_per_worker",
-                properties: LinkProperties::fk_composite("id", "id", Cardinality::ManyToOne, &[("worker_id", "worker_id")]),
+                properties: LinkProperties::fk_composite(
+                    "id",
+                    "id",
+                    Cardinality::ManyToOne,
+                    &[("worker_id", "worker_id")],
+                ),
             }]
         },
         column_semantic_types: &[],
@@ -2196,12 +2201,22 @@ pub static MZ_DATAFLOW_CHANNELS_PER_WORKER: LazyLock<BuiltinLog> = LazyLock::new
                 OntologyLink {
                     name: "source_operator",
                     target: "dataflow_operator_per_worker",
-                    properties: LinkProperties::fk_composite("from_index", "id", Cardinality::ManyToOne, &[("worker_id", "worker_id")]),
+                    properties: LinkProperties::fk_composite(
+                        "from_index",
+                        "id",
+                        Cardinality::ManyToOne,
+                        &[("worker_id", "worker_id")],
+                    ),
                 },
                 OntologyLink {
                     name: "target_operator",
                     target: "dataflow_operator_per_worker",
-                    properties: LinkProperties::fk_composite("to_index", "id", Cardinality::ManyToOne, &[("worker_id", "worker_id")]),
+                    properties: LinkProperties::fk_composite(
+                        "to_index",
+                        "id",
+                        Cardinality::ManyToOne,
+                        &[("worker_id", "worker_id")],
+                    ),
                 },
             ]
         },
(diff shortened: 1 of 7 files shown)


ggevay:
I think it would be good if @MaterializeInc/cluster also reviewed this, because I don't know this introspection stuff very well.

src/catalog/src/builtin.rs
```diff
@@ -2291,6 +2559,18 @@ pub static MZ_COMPUTE_LIR_MAPPING_PER_WORKER: LazyLock<BuiltinLog> = LazyLock::n
     oid: oid::LOG_MZ_COMPUTE_LIR_MAPPING_PER_WORKER_OID,
     variant: LogVariant::Compute(ComputeLog::LirMapping),
     access: vec![PUBLIC_SELECT],
+    ontology: Some(Ontology {
+        entity_name: "lir_mapping_per_worker",
+        description: "Mapping from LIR node IDs to dataflow operator address ranges per worker.",
+        links: &const {
+            [OntologyLink {
+                name: "lir_of",
```
ggevay: (This sounds like as if it were the LIR plan, but it's just an id, right?)

ggevay:
I have a few comments for `main` (i.e., the state before the PR):
- `cargo fmt` seems to have given up in some places, see e.g. the `Ontology {` for `mz_objects`. I think the standard fix in these cases is to manually get the code into into a state where `cargo fmt` is willing to operate on it.
- `LinkProperties`:
  - `ForeignKey`'s nullability still has the problem [mentioned](https://github.com/MaterializeInc/materialize/pull/36159#discussion_r3168578967) on the previous PR: I think foreign key traditionally means that you have referential integrity, i.e., when the column is not null then it is guaranteed to find a match. I think this is not true for the [example mentioned there](https://github.com/MaterializeInc/materialize/pull/36159#discussion_r3168696212), `sink_status_history`, because a replica might go away, after which `replica_id` might no longer find a match. I think it needs some careful consideration how exactly to address this issue. E.g. one option would be to just drop these `ForeignKey` links, but that would lose valuable information: It's still a valid thing to do a join between these things, it's just that you need to make it an outer join. So, I guess we still want some kind of link, but I'm not sure what exactly to call this, if it's not exactly a foreign key.
  - `MapsTo`
    - Why are `source_column` and `target_column` optional?
    - `MapsTo` is on both `mz_objects` and `mz_object_global_ids`, but these look like totally different cases: `mz_objects` has a column, `id`, that can be mapped to global ids via another relation (`mz_object_global_ids`), while `mz_object_global_ids` is a relation that itself maps from catalog ids to global ids.
    - Also, I don't understand several things on the `MapsTo` link of `mz_object_global_ids`:
      - Why isn't the `global_id` column mentioned?
      - What is `target_column` here? Does `target: "object"` mean that the target is the thing that has `entity_name: "object"`, which is `mz_objects`? But then `to_type` seems wrong: `mz_object`'s id column is catalog id, not global id.
  - `DependsOn`
    - Is it intended that this is used only for `mz_materialization_dependencies`? The doc comment (not of `DependsOn`, but the doc comment of `OntologyLink`) gives `mz_compute_dependencies` as an example, but it's not actually used there.
    - What is `source entity`? "source_column"/"target_column" makes it look like as if it were the thing itself that has the annotation, but the doc comment is not consistent with this.
    - `name: "materialization_depends_on"` seems off. E.g. one could read it as "the source materialization depends on the target of this link", but this is not the case. This ties back to my [earlier comment](https://github.com/MaterializeInc/materialize/pull/36159#discussion_r3168326195) about `OntologyLink::name` not being defined precisely enough.
- `OntologyLink`'s doc comment has an overview of `LinkProperties` enum variants. Why not put that overview on `LinkProperties` instead?
- `mz_objects` is missing some `Union`s, e.g. `function` and `type`, plus all the things inherited from `mz_relations`
- `mz_arrangement_sizes` has a "Corresponds to" in `column_comments`, but no `ForeignKey` link. Is that intended?
- There seems to be some confusion about the `object_type` for MVs whether it's `materialized-view` or `materialized view`. [dash example](https://github.com/MaterializeInc/materialize/blob/ea77b8b38bed028167adbb39c0c21dd057176d34/src/catalog/src/builtin.rs#L13831), [space example](https://github.com/MaterializeInc/materialize/blob/ea77b8b38bed028167adbb39c0c21dd057176d34/src/catalog/src/builtin.rs#L7817)

src/catalog/src/builtin.rs
```diff
@@ -2123,6 +2143,25 @@ pub static MZ_DATAFLOW_CHANNELS_PER_WORKER: LazyLock<BuiltinLog> = LazyLock::new
...
     access: vec![PUBLIC_SELECT],
+    ontology: Some(Ontology {
+        entity_name: "dataflow_channel_per_worker",
+        description: "Timely dataflow communication channels per worker.",
+        links: &const {
+            [
+                OntologyLink {
+                    name: "source_operator",
+                    target: "dataflow_operator_per_worker",
+                    properties: LinkProperties::fk("from_index", "id", Cardinality::ManyToOne),
```
ggevay: I'm not sure if this is `ManyToOne`: `dataflow_operator_per_worker` has a `worker_id` column, so maybe there is one per worker, and `from_index` finds a match for each worker?

src/catalog/src/builtin.rs
```diff
@@ -2115,6 +2123,18 @@ pub static MZ_DATAFLOW_ADDRESSES_PER_WORKER: LazyLock<BuiltinLog> = LazyLock::ne
...
     variant: LogVariant::Timely(TimelyLog::Addresses),
     access: vec![PUBLIC_SELECT],
+    ontology: Some(Ontology {
+        entity_name: "dataflow_address_per_worker",
+        description: "Scope address of each Timely operator per worker.",
+        links: &const {
+            [OntologyLink {
+                name: "address_of",
+                target: "dataflow_operator_per_worker",
+                properties: LinkProperties::fk("id", "id", Cardinality::ManyToOne),
```
ggevay: Should this be an `id, worker_id` composite key? This seems to be an issue for all the `_per_worker` relations.

ggevay: Btw. there are some complicated join examples in the definition of `mz_message_counts_per_worker`:
```
JOIN received_cte USING (channel_id, from_worker_id, to_worker_id)
JOIN batch_sent_cte USING (channel_id, from_worker_id, to_worker_id)
JOIN batch_received_cte USING (channel_id, from_worker_id, to_worker_id)",
```
This also looks like we need some tricky composite keys, which can't be expressed by `ForeignKey` currently.

src/catalog/src/builtin.rs
```diff
@@ -2349,6 +2695,18 @@ pub static MZ_MESSAGE_COUNTS_RECEIVED_RAW: LazyLock<BuiltinLog> = LazyLock::new(
     oid: oid::LOG_MZ_MESSAGE_COUNTS_RECEIVED_RAW_OID,
```
ggevay, line 2621: I think all (or maybe only some of) the `_raw` relations have this weird thing where the actual payload is stuffed into the diff field, so you need to do weird grouped counts to get useful values. See e.g., the definition of `received_cte`. This can really [mess up your day](https://materializeinc.slack.com/archives/CTESPM7FU/p1757428625358639?thread_ts=1757340038.295369&cid=CTESPM7FU) if you are not careful when querying one of these relations, so the ontology should definitely be made aware of this somehow.

(In that slack thread, it came up that we might even want to gate querying the `raw` stuff behind a session variable that needs to be explicitly turned on, to reduce the chances that people shoot themselves in the foot with queries that want to return 3049347435651 rows, like I did on my staging env.)

ggevay (changes requested):
Unfortunately, the introspection relations have some complications that can't be modeled by the `Ontology` structs' current form, see comments below.

mtabebe:
Thanks for all the comments!

> * `ForeignKey`'s nullability still has the problem [mentioned](https://github.com/MaterializeInc/materialize/pull/36159#discussion_r3168578967) on the previous PR

That does make sense ... what do you think about adding a new UnenforcedForeignKey variant? That just has the slightly different semantic meaning

> Why are source_column and target_column optional?

I have removed the optionality, good feedback.

> MapsTo

mz_object_global_ids now uses ForeignKey so we should be able to track back to the mz_objects.

> What is source entity?

I fixed the docs to make it clear what the from/to relationship is

> DependsOn

Updated the doc comments and switched mz_compute_dependencies.

> mz_arrangement_sizes missing FK

The Measures link already records this information, so I left it out but we could add the foreign key too?

src/catalog/src/builtin.rs
mtabebe: Good point, I'll add extra keys!

src/catalog/src/builtin.rs
mtabebe: My intuition is that for now we just remove the ontology from these ... if we think it is valuable we add more semantics later

src/catalog/src/builtin.rs
```diff
@@ -2224,6 +2339,25 @@ pub static MZ_COMPUTE_DATAFLOW_GLOBAL_IDS_PER_WORKER: LazyLock<BuiltinLog> =
...
         access: vec![PUBLIC_SELECT],
+        ontology: Some(Ontology {
+            entity_name: "dataflow_global_id_per_worker",
+            description: "Mapping from internal dataflow IDs to GlobalIds per worker.",
+            links: &const {
+                [OntologyLink {
+                    name: "global_id_of",
+                    target: "compute_export_per_worker",
+                    properties: LinkProperties::MapsTo {
+                        source_column: "global_id",
```
antiguru, line 2374: Note that the global_id isn't guaranteed to exist on the other end, but it's not a foreign key, so probably ok?

src/catalog/src/builtin.rs
```diff
@@ -17729,11 +18041,16 @@ mod tests {
...
         for builtin in BUILTINS_STATIC.iter() {
-            let (name, desc, ontology) = match builtin {
+            let desc_storage;
+            let (name, desc, ontology): (&str, &RelationDesc, Option<&Ontology>) = match builtin {
                 Builtin::Table(t) => (t.name, &t.desc, t.ontology.as_ref()),
                 Builtin::View(v) => (v.name, &v.desc, v.ontology.as_ref()),
                 Builtin::MaterializedView(mv) => (mv.name, &mv.desc, mv.ontology.as_ref()),
                 Builtin::Source(s) => (s.name, &s.desc, s.ontology.as_ref()),
+                Builtin::Log(l) => {
+                    desc_storage = l.variant.desc();
```
antiguru, line 18102: 💃

src/catalog/src/builtin.rs
```diff
@@ -17729,11 +18041,16 @@ mod tests {
         // we still need to verify the string value matches a real column.
         let mut bad_source_cols = Vec::new();
         for builtin in BUILTINS_STATIC.iter() {
```
antiguru, line 18094: Do you think it'd be possible to move the tests to a separate file? It's getting really hard to view this absolute unit of a file.

antiguru (approved):
Seems fine! Spot-checked the ontology definition of the builtins, and seems to check out. Also, seems low risk if wrong.

src/catalog/src/builtin.rs
mtabebe: Good point, IMO we should split this file up into things like mz_introspection.rs, pg_catalog.rs, mz_internal.rs and mz_catalog.rs.

This is a separate PR. I can do that separately.

ggevay:
> what do you think about adding a new UnenforcedForeignKey variant? That just has the slightly different semantic meaning

Sounds fine to me!

>> DependsOn

> Updated the doc comments and switched mz_compute_dependencies.

Could `mz_object_dependencies` and `mz_object_transitive_dependencies` use this as well?

src/catalog/src/builtin.rs
ggevay: Ok!

ggevay (approved):

src/catalog/src/builtin.rs
mtabebe: https://github.com/MaterializeInc/materialize/pull/36478

PR #36408, opened 2026-05-05
timely-util: columnar-native merge-batcher (chunker + merger + bench)

DAlperin:
## Summary

Adds an all-`Column` merge-batcher path: an input-side chunker
(`ColumnChunker`), an `InternalMerge` impl on `Column<(D, T, R)>`
plugged into DD's standard `InternalMerger` via the `ColumnMerger` type
alias, optimizations for the inner loop and the chunker, and three
microbenchmarks (primitive keys + Row-keyed merger + Row-keyed full
`MergeBatcher`).

The legacy columnation-backed `Chunker` / `ColInternalMerger` path is
unchanged; this PR is purely additive.

## Commits

1. `ColumnChunker` + `InternalMerge for Column<(D,T,R)>` + `ColumnMerger` type alias. Naive 0/1/2-input merge with bulk-swap shortcut, equal-key consolidation via a reusable `R` accumulator. Proptests covering `merge_from` / `extract` against a brute-force reference.
2. Galloping for Less/Greater runs — view-projection so the gallop predicate touches only the D and T leaves; length-1 push fast path that skips `extend_from_self`'s per-leaf 1-element memcpy.
3. Mid-merge yield in `merge_from`. Gallop bypass for length-1 runs (the common case on interleaved data). Microbench in `benches/columnar_merger.rs` comparing `ColumnMerger` against `ColInternalMerger`.
4. Address review: replace the `default_capacity`-driven yield with `ColumnBuilder`'s serialized-words ship heuristic. Drop the `LeafReserve` trait + macro; `SizableContainer for Column<C>` is now minimal (required as a supertrait of `InternalMerge`). Per-leaf inner-loop pushes/extends — destructure `(D, T, R)::Container` once, dispatch to `sd`/`st`/`sr` directly so the gallop bulk-copy resolves to three independent `extend_from_slice` calls the compiler can autovectorize.
5. `mz-timely-util` primitive bench parameterized on per-side heap footprint; sweeps four sizes from L1-resident (32 KiB) to DRAM-resident (128 MiB) so the curves' cache crossovers are visible. Throughput in bytes (cross-record-shape comparable).
6. Row-keyed bench mirroring the same regimes/sizes with `Data = Row` — the variable-length payload regime where parallel-array layout is expected to amortize per-leaf overhead.
7. Both benches emit a side-by-side throughput summary table (GiB/s and latency, with column-vs-columnation ratio) after `group.finish()` — pulls medians from criterion's `estimates.json` so we don't add a `serde_json` dev-dep.
8. Inner-loop optimizations driven by xctrace profiling of the slowest Row regimes:
   - **Drop the `copy_from + push(&owned)` round-trip for D and T in the `Equal` branch.** Refs from the input borrow are valid through the merge step, so push them via each leaf's `Push<Ref<_>>` impl directly. The original shape triggered `Row::copy_from = CompactBytes::clear() + extend_from_slice` per equal-key record, which freed and re-malloc'd every row that overflowed the inline budget.
   - **Pre-size output leaves** via `Container::reserve_for([left_borrow, right_borrow])` in `merge_from`, gated on a record-count threshold to avoid over-reserving on huge inputs where consolidation makes the input bound a 4× over-estimate.
   - **Amortize the in-loop `at_ship_threshold` check** — every 1024 iterations via a power-of-two mask. The closure walks 4–5 leaf slices on variable-length leaves; overshooting the ship boundary by ~1 K records has no practical impact since the framework's outer `at_capacity` ships the chunk regardless.
9. **Apply the same `copy_from` fix to `ColumnChunker`** — the chunker had the same Row-malloc thrash as the merger's Equal branch on every consolidated output record. **Add `extract` ship-threshold yield** required by `InternalMerge::extract`'s contract: the framework only checks `at_capacity` between calls, so without an inner-loop yield a single call could fill an output well past threshold.
10. Move Row benches from `mz-repr` to `mz-compute` and add a full-`MergeBatcher` Row bench that exercises the entire push-rounds + seal cycle (chunker + chain compaction + extract) — companion to the merger-only bench.

## Perf — primitive `((u64, u64), u64, i64)`

`mz-timely-util/benches/columnar_merger.rs`. Three regimes × four per-side heap-footprint targets, measured on Apple Silicon.

```
┌─────────────────┬────────────────────────┬────────────────────────┬───────────────────────┐
│     Config      │      columnation       │         column         │ column vs columnation │
├─────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│ mixed/32K       │ 18.19 GiB/s (2.9 µs)   │ 4.24 GiB/s (12.4 µs)   │          4.29× slower │
├─────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│ collisions/32K  │ 19.13 GiB/s (2.7 µs)   │ 4.01 GiB/s (12.8 µs)   │          4.77× slower │
├─────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│ disjoint/32K    │ 33.68 GiB/s (1.6 µs)   │ 56.73 GiB/s (941 ns)   │          1.68× faster │
├─────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│ mixed/512K      │ 18.84 GiB/s (44.5 µs)  │ 3.56 GiB/s (235.7 µs)  │          5.29× slower │
├─────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│ collisions/512K │ 18.83 GiB/s (44.4 µs)  │ 3.26 GiB/s (256.7 µs)  │          5.78× slower │
├─────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│ disjoint/512K   │ 36.29 GiB/s (23.1 µs)  │ 99.15 GiB/s (8.5 µs)   │          2.73× faster │
├─────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│ mixed/8M        │ 7.33 GiB/s (1.83 ms)   │ 2.95 GiB/s (4.55 ms)   │          2.49× slower │
├─────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│ collisions/8M   │ 7.28 GiB/s (1.84 ms)   │ 2.64 GiB/s (5.07 ms)   │          2.76× slower │
├─────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│ disjoint/8M     │ 34.67 GiB/s (386.3 µs) │ 146.09 GiB/s (91.7 µs) │          4.21× faster │
├─────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│ mixed/128M      │ 6.88 GiB/s (31.17 ms)  │ 2.67 GiB/s (80.21 ms)  │          2.57× slower │
├─────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│ collisions/128M │ 6.87 GiB/s (31.22 ms)  │ 2.48 GiB/s (86.56 ms)  │          2.77× slower │
├─────────────────┼────────────────────────┼────────────────────────┼───────────────────────┤
│ disjoint/128M   │ 27.29 GiB/s (7.86 ms)  │ 18.84 GiB/s (11.37 ms) │          1.45× slower │
└─────────────────┴────────────────────────┴────────────────────────┴───────────────────────┘
```

## Perf — Row-keyed `(Row, u64, i64)`, `Merger::merge` only

`mz-compute/benches/columnar_merger_row.rs`. Each Row is `(Int64, String)` with a 24-character hex-string suffix that overflows `CompactBytes`'s inline budget — so columnation actually exercises its `LgAllocRegion` copy path and column packs into its `Rows` value buffer. 32 KiB / 512 KiB / 8 MiB per side; 128 MiB skipped because Row-heavy setup costs would dominate.

```
┌─────────────────┬───────────────────────┬────────────────────────┬───────────────────────┐
│     Config      │      columnation      │         column         │ column vs columnation │
├─────────────────┼───────────────────────┼────────────────────────┼───────────────────────┤
│ mixed/32K       │ 3.70 GiB/s (19.8 µs)  │ 3.54 GiB/s (20.8 µs)   │          1.05× slower │
├─────────────────┼───────────────────────┼────────────────────────┼───────────────────────┤
│ collisions/32K  │ 7.10 GiB/s (4.7 µs)   │ 6.88 GiB/s (4.9 µs)    │          1.03× slower │
├─────────────────┼───────────────────────┼────────────────────────┼───────────────────────┤
│ disjoint/32K    │ 4.09 GiB/s (16.9 µs)  │ 6.81 GiB/s (10.2 µs)   │          1.67× faster │
├─────────────────┼───────────────────────┼────────────────────────┼───────────────────────┤
│ mixed/512K      │ 3.02 GiB/s (391.0 µs) │ 2.98 GiB/s (395.4 µs)  │          1.01× slower │
├─────────────────┼───────────────────────┼────────────────────────┼───────────────────────┤
│ collisions/512K │ 5.92 GiB/s (93.3 µs)  │ 5.86 GiB/s (94.3 µs)   │          1.01× slower │
├─────────────────┼───────────────────────┼────────────────────────┼───────────────────────┤
│ disjoint/512K   │ 5.55 GiB/s (200.6 µs) │ 20.31 GiB/s (54.8 µs)  │          3.66× faster │
├─────────────────┼───────────────────────┼────────────────────────┼───────────────────────┤
│ mixed/8M        │ 2.54 GiB/s (7.43 ms)  │ 2.78 GiB/s (6.80 ms)   │          1.09× faster │
├─────────────────┼───────────────────────┼────────────────────────┼───────────────────────┤
│ collisions/8M   │ 4.48 GiB/s (1.97 ms)  │ 5.46 GiB/s (1.62 ms)   │          1.22× faster │
├─────────────────┼───────────────────────┼────────────────────────┼───────────────────────┤
│ disjoint/8M     │ 4.09 GiB/s (4.35 ms)  │ 28.94 GiB/s (615.2 µs) │          7.07× faster │
└─────────────────┴───────────────────────┴────────────────────────┴───────────────────────┘
```

## Perf — Row-keyed full `MergeBatcher` push + seal

`mz-compute/benches/columnar_merge_batcher_row.rs`. Drives the entire batcher cycle for both the legacy (`Chunker<ColumnationStack>` + `ColInternalMerger`) and column (`ColumnChunker` + `ColumnMerger`) paths: 64 push-rounds of 4 K records each, then a final `seal` so the chunks flow through `Merger::merge` (chain compaction during push + final merge) and `Merger::extract`. Same regimes/sizes as the merger bench.

```
┌─────────────────┬───────────────────────┬───────────────────────┬───────────────────────┐
│     Config      │      columnation      │        column         │ column vs columnation │
├─────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
│ mixed/32K       │ 824 MiB/s (227.5 µs)  │ 897 MiB/s (209.0 µs)  │          1.09× faster │
├─────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
│ collisions/32K  │ 1.02 GiB/s (180.3 µs) │ 1.02 GiB/s (179.3 µs) │          1.01× faster │
├─────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
│ disjoint/32K    │ 876 MiB/s (214.1 µs)  │ 919 MiB/s (204.1 µs)  │          1.05× faster │
├─────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
│ mixed/512K      │ 547 MiB/s (1.37 ms)   │ 554 MiB/s (1.35 ms)   │          1.01× faster │
├─────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
│ collisions/512K │ 596 MiB/s (1.26 ms)   │ 614 MiB/s (1.22 ms)   │          1.03× faster │
├─────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
│ disjoint/512K   │ 557 MiB/s (1.35 ms)   │ 571 MiB/s (1.31 ms)   │          1.03× faster │
├─────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
│ mixed/8M        │ 337 MiB/s (35.60 ms)  │ 297 MiB/s (40.40 ms)  │          1.14× slower │
├─────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
│ collisions/8M   │ 384 MiB/s (31.23 ms)  │ 369 MiB/s (32.50 ms)  │          1.04× slower │
├─────────────────┼───────────────────────┼───────────────────────┼───────────────────────┤
│ disjoint/8M     │ 338 MiB/s (35.49 ms)  │ 320 MiB/s (37.55 ms)  │          1.06× slower │
└─────────────────┴───────────────────────┴───────────────────────┴───────────────────────┘
```

## Reading the curves

- **Row workloads with galloping room** are where column wins decisively. Headline: `Merger::merge` Row/disjoint/8 MiB tops at **28.94 GiB/s vs columnation's 4.09 GiB/s — 7.07× faster.** That's the merge-batcher's actual hot path for hydration: pre-sorted batches from persist composing into larger sorted runs.

- **Row + variable-length payload closes the per-leaf push tax.** The merger bench shows column tied or winning on every Row config; the only sub-1.0× ratios are within ~5%, well inside thermal noise on Apple Silicon. This is the regime that matters for compute arrangements over real data.

- **Primitives are a structural-floor regime for `Column`.** A 24-byte record × 4 leaf pushes per record vs columnation's single flat-slice push; the per-leaf dispatch cost is fully exposed when the working set fits in cache (worst gap 5.78× at 512 KiB collisions). Galloping (disjoint) still wins decisively at every cache-resident size (peak 4.21× at 8 MiB), where column's per-leaf `extend_from_self` is the right primitive for long runs.

- **Full-`MergeBatcher` bench tightens the 8M gap.** End-to-end the column path is **at-or-better than columnation on every cache-resident size** (32 KiB / 512 KiB) and within 1.04–1.14× at 8 MiB. The 8M residual is the structural per-leaf push tax surfacing in `merge_from` cascades on interleaved keys; cache-resident regimes don't see it because chunker work dominates and the chunker is now competitive.

- **The 128 MiB primitive `disjoint` slowdown** (1.45× slower) is one bandwidth-pathological case: a single ~128 MiB memcpy where columnation's one-slice copy edges out column's three independent leaf copies. Run-to-run noise on Apple Silicon is high here — number swings 15–41 GiB/s across runs without code changes. Not a regime real workloads hit; chunks are ship-threshold-bounded at 2 MiB.

## Test plan

- [x] `cargo nextest run -p mz-timely-util columnar` — 15/15 pass (10 hand-written + 5 proptests).
- [x] `cargo bench -p mz-timely-util --bench columnar_merger` — produces the primitive table above.
- [x] `cargo bench -p mz-compute --bench columnar_merger_row` — produces the Row merger table.
- [x] `cargo bench -p mz-compute --bench columnar_merge_batcher_row` — produces the full-batcher table.
- [ ] CI green.


src/timely-util/src/columnar/batcher.rs
```diff
@@ -114,3 +122,738 @@ where
...
+                // Yield mid-merge when self has reached the target chunk size,
+                // so the driver swaps in a fresh pre-allocated chunk. Without
+                // this, a single `merge_from` call runs to one input's
+                // exhaustion, growing self past the capacity that
+                // `ensure_capacity` reserved and incurring the realloc cost
+                // that pre-allocation is meant to avoid. `borrow().len()`
+                // inlines through the leaf chain; tracking a local counter
+                // instead measured slower, presumably because the extra
+                // updates added register pressure to the hot loop.
+                let target = timely::container::buffer::default_capacity::<(D, T, R)>();
```
antiguru: We shouldn't be using the default capacity here (or ever, really). Instead, I'd copy the heuristic the column builder encodes: If we're within 10% of 2MiB (or a configurable constant, `const`), we ship the batch. Question is whether we should serialize or not, but I lean toward "yes".

src/timely-util/src/columnar/batcher.rs
```diff
@@ -114,3 +122,738 @@ where
...
+
+                while left_pos[0] < upper_l
+                    && right_pos[0] < upper_r
+                    && self_c.borrow().len() < target
+                {
+                    let (d1, t1, r1) = left_borrow.get(left_pos[0]);
+                    let (d2, t2, r2) = right_borrow.get(right_pos[0]);
```
antiguru: What helped in the consolidating column builder: Use the fact that `(D, T, R)::Container` is `(D::Container, T::Container, R::Container)`, which allows us to construct the result using three separate containers, as long as the length matches. Especially with galloping, writing runs of values to a single container enables vectorization.

src/timely-util/src/columnar.rs
```diff
@@ -137,6 +137,76 @@ where
...
+        }
+    };
+}
+
+tuple_leaf_reserve!(A a);
+tuple_leaf_reserve!(A a, B b);
+tuple_leaf_reserve!(A a, B b, C c);
+tuple_leaf_reserve!(A a, B b, C c, D d);
+
+impl<C: Columnar> SizableContainer for Column<C>
```
antiguru: If at all possible, we shouldn't implement `SizableContainer` for `Column`. It's legacy, and each use is a smell.

src/timely-util/src/columnar.rs
```diff
@@ -137,6 +137,76 @@ where
...
+/// reserves a sum of input view lengths, which is awkward to use for
+/// "preallocate N records" without a model view. This trait fills that gap
+/// with a simple element-count reserve, implemented for `Vec<T>` and
+/// tuple-of-leaves.
+///
+/// Used by [`Column::ensure_capacity`] to skip the geometric-grow cost of a
+/// fresh chunk's leaf Vecs filling up. Without it, each chunk's leaves grow
+/// from zero by doubling, and the cumulative realloc traffic shows up as
+/// allocator memmove cost on the hot path.
+pub trait LeafReserve {
```
antiguru: I think if we switch from the default capacity to serialized capacity, we don't need this helper trait anymore. We'd build into one owned copy, and to ship it we'd serialize into a new one and clear. After the first round, we should have appropriate capacity, so we don't have to worry about pre-sizing.

src/timely-util/benches/columnar_merger.rs
```diff
@@ -0,0 +1,157 @@
...
+fn build_column(data: &[Tuple]) -> Column<Tuple> {
+    let mut col: Column<Tuple> = Default::default();
+    for &tup in data {
+        col.push_into(tup);
+    }
+    col
+}
+
+fn bench_merge(c: &mut Criterion) {
+    let n = 10_000usize;
```
antiguru: It might help to define the `n` not in terms of elements but size, and then scale  for specific heap sizes.

antiguru (commented):
Very cool! I left some comments inline, but overall this points in the right direction.

src/timely-util/src/columnar/batcher.rs
DAlperin: Do you mean serialize on the boundaries? It's trivial enough to serialize in `extract` but it will just flow directly into a trace builder which will probably have to decode it to bytes anyway right away

def- (commented):
`Column<(D,T,R)>::extract` (`src/timely-util/src/columnar/batcher.rs:548`) violates the `InternalMerge::extract` contract: it never yields mid-buffer, so the framework can't flush `keep`/`ship` until the input is fully drained. Other impls (`Vec`, `ColumnationStack`) gate their inner loop on `!keep.at_capacity() && !ship.at_capacity()`.

Output chunks overshoot the ship threshold by up to one full input buffer. Not sure how bad that is, but could waste a bunch of memory?
Small test:
```diff
diff --git a/src/timely-util/src/columnar/batcher.rs b/src/timely-util/src/columnar/batcher.rs
index 600b14c4c3..f39b42f846 100644
--- a/src/timely-util/src/columnar/batcher.rs
+++ b/src/timely-util/src/columnar/batcher.rs
@@ -926,4 +926,43 @@ mod proptests {
             prop_assert!(frontier.elements().is_empty());
         }
     }
+
+    proptest! {
+        #![proptest_config(ProptestConfig { cases: 16, ..ProptestConfig::default() })]
+
+        #[mz_ore::test]
+        #[cfg_attr(miri, ignore)]
+        fn extract_yields_when_keep_at_capacity(upper_time in 0u64..=4) {
+            use timely::container::SizableContainer;
+
+            let mut keep: Column<Tuple> = Default::default();
+            let mut k = 0u64;
+            while !keep.at_capacity() {
+                keep.push_into(((k, 0u64), upper_time, 1i64));
+                k += 1;
+            }
+
+            let mut self_col: Column<Tuple> = Default::default();
+            for i in 0..32u64 {
+                self_col.push_into(((i, 1u64), upper_time, 1i64));
+            }
+            let len = self_col.borrow().len();
+
+            let upper = Antichain::from_elem(upper_time);
+            let mut frontier: Antichain<u64> = Antichain::new();
+            let mut ship: Column<Tuple> = Default::default();
+            let mut position = 0;
+
+            InternalMerge::extract(
+                &mut self_col,
+                &mut position,
+                upper.borrow(),
+                &mut frontier,
+                &mut keep,
+                &mut ship,
+            );
+
+            prop_assert!(position < len, "position={}, len={}", position, len);
+        }
+    }
 }
```

DAlperin:
@def- yep good point, will fix tomorrow

src/timely-util/src/columnar/batcher.rs
```diff
@@ -114,3 +121,809 @@ where
...
+                // 4–5 slices and sums lengths — non-trivial, so the
+                // caller checks every `THRESHOLD_PERIOD_MASK + 1`
+                // iterations rather than per-iter. At the ship threshold
+                // a chunk is ~65 K records, so overshooting by ~1 K
+                // records before we notice has no practical impact:
+                // the outer `at_capacity` check sees the oversize chunk
+                // and ships it regardless.
+                let at_ship_threshold =
+                    |sd: &D::Container, st: &T::Container, sr: &R::Container| {
+                        use columnar::AsBytes as _;
```
DAlperin: we already have a function for this

src/timely-util/src/columnar.rs
antiguru: https://github.com/TimelyDataflow/differential-dataflow/pull/738 would allow us to remove the requirement.

src/timely-util/src/columnar/batcher.rs
```diff
@@ -114,3 +121,809 @@ where
...
+    }
+}
+
+/// Counterpart to `ColInternalMerger` (which merges `ColumnationStack` chunks).
+/// Drives the merge batcher with [`Column`]-shaped chunks, no columnation
+/// detour, by way of [`InternalMerge`] below.
+pub type ColumnMerger<D, T, R> = InternalMerger<Column<(D, T, R)>>;
+
+/// `InternalMerge` for [`Column`]-shaped sorted chunks.
+impl<D, T, R> InternalMerge for Column<(D, T, R)>
```
antiguru: Apparently InternalMerge is supposed to go away. Let's implement Merger directly?

src/timely-util/src/columnar/batcher.rs
```diff
@@ -114,3 +120,1205 @@ where
...
+}
+
+impl<U: Columnar> Default for ColumnChunker<U> {
+    fn default() -> Self {
+        Self {
+            target: Column::default(),
+            ready: VecDeque::new(),
+        }
+    }
+}
```
antiguru, line 149: Can `Default` be derived?

src/timely-util/src/columnar/batcher.rs
```diff
@@ -114,3 +120,1205 @@ where
...
+                // is bounded by 2× the actual output and avoids
+                // committing pages we'd never touch.
+                const RESERVE_RECORD_THRESHOLD: usize = 1_000_000;
+                if upper_l + upper_r <= RESERVE_RECORD_THRESHOLD {
+                    use columnar::Container as _;
+                    let inputs = [left_borrow, right_borrow];
+                    sd.reserve_for(inputs.iter().map(|b| b.0));
+                    st.reserve_for(inputs.iter().map(|b| b.1));
+                    sr.reserve_for(inputs.iter().map(|b| b.2));
+                }
```
antiguru, line 424: Once we land this PR, we should look into serializing the chunks instead of passing typed data. This would remove the need for pre-allocating the outputs.

src/timely-util/src/columnar/batcher.rs
```diff
@@ -114,3 +120,1205 @@ where
...
+            // its current position) is sortable-before the other head's
+            // current record, the entire tail can be appended to `output`
+            // without per-record compares or per-leaf byte copies.
+            //
+            // Two probes (one record from each side) settle this. Common
+            // case for hydration paths where chains compose pre-sorted
+            // batches with non-overlapping or tail-disjoint key ranges —
+            // skips an entire `merge_from` invocation, including its gallop
+            // bulk-copies, and replaces the byte-level extend with a
+            // `mem::replace` of the head into `output`.
```
antiguru: Nit: The optimization makes sense, but I'm not sure that we get to benefit it for the reasons listed here. There's no guarantee that we receive sorted data during hydration, as the data passes through too many layers to reason about its properties in such detail.

src/timely-util/src/columnar/batcher.rs
```diff
@@ -114,3 +120,1205 @@ where
...
+            kept.push(keep);
+        }
+        if !ready.is_empty() {
+            ship.push(ready);
+        }
+    }
+
+    fn account(chunk: &Self::Chunk) -> (usize, usize, usize, usize) {
+        let records = usize::try_from(chunk.record_count()).expect("record_count is non-negative");
+        (records, 0, 0, 0)
```
antiguru: Check the other values, it's probably size, capacity, allocations. Pass `.., length_in_words * 8, length_in_words * 8, 1)` here to make sure we can account properly.

src/timely-util/src/columnar/batcher.rs
```diff
@@ -114,3 +120,1205 @@ where
...
+/// here come from the merger and chunker, both of which produce `Typed`.
+#[inline]
+fn recycle_chunk<C: Columnar>(mut chunk: Column<C>, stash: &mut Vec<Column<C>>) {
+    match &mut chunk {
+        Column::Typed(c) => c.clear(),
+        Column::Bytes(_) | Column::Align(_) => {
+            chunk = Column::Typed(Default::default());
+        }
+    }
+    stash.push(chunk);
```
antiguru: No benefit in pushing non-allocated chunks.

src/timely-util/src/columnar.rs
```diff
@@ -138,6 +138,60 @@ where
...
+            Column::Bytes(b) => {
+                let words = b.len() / 8;
+                let round = (words + (SHIP_WORDS - 1)) & !(SHIP_WORDS - 1);
+                round - words < round / 10
+            }
+            Column::Align(a) => {
+                let words = a.len();
+                let round = (words + (SHIP_WORDS - 1)) & !(SHIP_WORDS - 1);
+                round - words < round / 10
+            }
```
antiguru: As discussed: `at_capacity` for bytes/align doesn't make much sense other then `true`. We can't possibly push into serialized forms. (In theory we could ofc, but I don't think it'll happen.)

antiguru (approved):
Seems good. I left some comments, but once addressed I think we're good to merge.

PR #36409, opened 2026-05-05
storage: fix get_active_replicas_for_object for sinks and top-level ingestions

martykulma:
Continuing the work that @def- started in https://github.com/MaterializeInc/materialize/pull/36327

This PR addresses a DroppedIds leak in the storage controller.

# Motivation
From @def- :
> Since https://github.com/MaterializeInc/materialize/pull/32066 sinks have per-replica scheduling on multi-replica clusters, but get_active_replicas_for_object still fell through to the "all replicas" branch for any non-ingestion-export ID. This caused drop_sinks_unvalidated to expect DroppedId from all replicas when only one was running the sink, leaking entries in dropped_objects indefinitely.
>
> Add the missing branches for top-level ingestion IDs and export (sink) IDs to mirror the scheduling logic in update_scheduling/active_replicas.

Additionally, the default branch in `get_active_replicas_for_object` returned all replicas. This seems incorrect, as the function is only called for Ingestion/Export, and if the GlobalId hasn't matched the 3 previous branches, it means the id is not in the active maps.  So it's either unknown or compacted. The default branch should return BTreeSet::new() in that case, otherwise this would also leak DroppedIds.

To avoid racing `set_hold_policies`, collection ids being dropped must be collected first before adjusting hold policies.

### DIFF
--- a/src/storage-controller/src/instance.rs
+++ b/src/storage-controller/src/instance.rs
@@ -739,16 +739,15 @@ impl Instance {
     /// object (ingestion, ingestion export (aka. subsource), or export).
     pub fn get_active_replicas_for_object(&self, id: &GlobalId) -> BTreeSet<ReplicaId> {
         if let Some(ingestion_id) = self.ingestion_exports.get(id) {
-            // Right now, only ingestions can have per-replica scheduling decisions.
             match self.active_ingestions.get(ingestion_id) {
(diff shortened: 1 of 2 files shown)


patrickwwbutler (approved):
LGTM! took me a little while to actually grok what this code is supposed to do and what the issue was, but it seems like it should address the problem!

PR #36410, opened 2026-05-05
sql: parse EXECUTE UNIT TEST

sjwiesman:
## Summary

Adds parser support for `EXECUTE UNIT TEST`, a SQL-level construct for
declarative unit testing of views and materialized views. The plan side returns
`unsupported`.

## Syntax

```sql
 EXECUTE UNIT TEST <test_name>
      FOR <target_name>
      [AT TIME <expr>]
      [MOCK <mock_name> ( <col_def> [, ...] ) AS ( <query> ) ,]...
      EXPECTED ( <col_def> [, ...] ) AS ( <query> )
```

- `<test_name>` — identifier, reported in test output.
- `FOR <view>` — fully-qualified name of the view or materialized view
  under test.
- `AT TIME <expr>` *(optional)* — time at which to evaluate the view, for
  testing temporal logic.
- `MOCK <input_view> (cols) AS <query>` *(zero or more)* — substitutes
  `<input_view>` with the rows produced by `<query>` for the duration of
  this test. Column list defines the schema the test expects from the
  mocked input.
- `EXPECTED (cols) AS <query>` *(required, exactly one)* — the rows the
  view under test must produce.

## Example

```sql
  EXECUTE UNIT TEST t FOR v
      MOCK a(x int) AS (SELECT 1),
      MOCK b(y int) AS (SELECT 2)
      EXPECTED (z int) AS (SELECT 3);
```

## Scope

- New keywords: `UNIT`, `TEST`, `MOCK`, `EXPECTED` (`src/sql-lexer`).
- New AST nodes: `ExecuteUnitTestStatement`, `MockViewDef`,
  `ExpectedResultDef` (`src/sql-parser/ast/defs/statement.rs`).
- Parser entry from `EXECUTE UNIT TEST ...` (`src/sql-parser/parser.rs`).
- Plan side returns `unsupported` — execution is intentionally out of
  scope for this PR.

### DIFF
--- a/src/sql-parser/src/ast/defs/statement.rs
+++ b/src/sql-parser/src/ast/defs/statement.rs
@@ -5115,10 +5115,9 @@ impl<T: AstInfo> AstDisplay for ExecuteUnitTestStatement<T> {
             f.write_str(" AT TIME ");
             f.write_node(at_time);
         }
-        for mock in &self.mocks {
-            f.write_str(" MOCK ");
+        for (i, mock) in self.mocks.iter().enumerate() {
+            f.write_str(if i == 0 { " MOCK " } else { ", MOCK " });
             f.write_node(mock);
-            f.write_str(",");
         }
         f.write_str(" EXPECTED ");
         f.write_node(&self.expected);
--- a/src/sql-parser/src/ast/defs/statement.rs
+++ b/src/sql-parser/src/ast/defs/statement.rs
@@ -105,6 +105,7 @@ pub enum Statement<T: AstInfo> {
     Close(CloseStatement),
     Prepare(PrepareStatement<T>),
     Execute(ExecuteStatement<T>),
+    ExecuteUnitTest(ExecuteUnitTestStatement<T>),
     Deallocate(DeallocateStatement),
     Raise(RaiseStatement),
     GrantRole(GrantRoleStatement<T>),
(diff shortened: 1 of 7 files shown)


sjwiesman:
Purposefully left this undocumented because you can't actually run it.

src/sql-parser/src/ast/defs/statement.rs
```diff
@@ -5049,6 +5095,72 @@ impl<T: AstInfo> AstDisplay for ExecuteStatement<T> {
...
+        f.write_node(&self.target);
+        if let Some(at_time) = &self.at_time {
+            f.write_str(" AT TIME ");
+            f.write_node(at_time);
+        }
+        for mock in &self.mocks {
+            f.write_str(" MOCK ");
+            f.write_node(mock);
+        }
+        f.write_str(" EXPECTED");
```
def-:
```suggestion
        f.write_str(" EXPECTED ");
```

def- (changes requested):
MOCK and EXPECTED need to be added in `is_reserved_in_column_alias` or `is_reserved_in_table_alias`

Multiple `MOCK ..., MOCK...` don't parse:
```
ERR: Expected left parenthesis, found MOCK
```

sjwiesman:
Thanks for the review. I realized the grammar was not as restrictive as I intended. Made a slightly different change but should still handle your case.

ggevay (approved):
Requiring the parens was a good idea to avoid adding stuff to `is_reserved_in_column_alias`/`is_reserved_in_table_alias`!

Looks good to me now, only some minor things:
- Might be a bit surprising that a comma is needed even after the final `MOCK`, i.e., before the `EXPECTED`.
- You could add a few more tests. (also negative ones)

sjwiesman:
Happy to make both updates before merging!

ggevay:
Thanks, looks great to me!

@def-, it seems my approval is not overriding your "Changes requested" review, so I think you need to drop it (if you agree with the fix).

def- (approved):
The PR author can dismiss a review with a reason I think.

sjwiesman:
@def- TIL!

PR #36413, opened 2026-05-05
console: maintained objects freshness diagnosis

leedqin:
### Motivation
Freshness graphs for maintained objects that show the critical path of lag from upstream dependencies on that object. Critical path is shown as a DAG. Fixes [CNS-52](https://linear.app/materializeinc/issue/CNS-52/objects-freshness-visualizations) 

### Description

A user can pick the a point in time in Freshness graph to see the critical path of objects and their respective lag and a table of immediate upstream dependencies. The lookback window on table corresponds to the lookback window in the freshness graph as well to let the user look at freshness few hours back or few minutes back. In that window, they can look at the critical path at a point in time on freshness graph. Design for the DAG: https://dependency-timeline-mz.lovable.app/

I didn't follow this pixel perfect but used the color schemes for the nodes. Since it's really hard to pinpoint to one particular object that it is causing the delay. 

### Different UI scenarios
* Objects with healthy freshness around 2 seconds

https://github.com/user-attachments/assets/e8669359-fd93-4859-984d-b28f53320c23

* Objects with a delay from source at a particular timestamp
image

* Showing delay being introduced  in the critical chain from sources (hence marked red). If a node is yellow then that means (it's lag is > 2s, it's self delay is less than the max self delay in the chain i.e the lag is inherited from the parent). Showing delay being introduced from the sources and the children inheriting them. 

https://github.com/user-attachments/assets/72119b6b-bf77-4f7a-90a0-cde409227f9e

* Self delay of an object
image

*Cluster metrics for the object at a timestamp
image

* Breadcrumbs
image

###  Tips For Reviewer
- All the changes in this PR are post `Freshness Chart` commit. I have broken down most changes by commit so they can be reviewed on their own.

The most review intensive commits are:

- `CriticalPathQuery`:
This is a recursive sql query to trace the chain for the object's backward to determine "what makes the object slow", returning every edge along the chain plus the immediate off-path siblings of each chain node so the UI can show both the spine and what else feeds into it.

- `Critical path DAG`:
This commit deals with transforming the results from the query into a renderable DAG. I would review the files in this order:
- `criticalPathRenderable.ts` has a helper function `buildGraphView` (at the bottom of the file, would recommend reading from here) that takes all the other helpers to return the nodes and edges needed for the graph.  Here is the flow how buildGraphView works 
```
┌─────────────────────────────────────────────────────────┐
│                                                            │
│  ─ CriticalPathData.rows  ← useCriticalPath() query      │
│  ─ probe (MaintainedObjectListItem)                      │
│  ─ visibleDepth           ← React state                  │
│  ─ expandedBottleneckId   ← React state                  │
└───────────────────────────┬─────────────────────────────┘
                            │
                            ▼
┌─────────────────────────────────────────────────────────┐
│  buildGraphView                             │
│                                                          │
│  ┌────────────────────────────────────────────────────┐ │
│  │ STEP 1 — pick the chain                             │ │
│  │ ────────────────────────────────────────────────── │ │
│  │ helper: buildChainMaps                              │ │
│  │   ↳ nodeById             (id → row metadata)        │ │
│  │   ↳ chainParentsByChild  (child → bottleneck dads)  │ │
│  │   ↳ depthById            (BFS hops from probe)      │ │
│  │                                                     │ │
│  │ filter depthById ≤ visibleDepth                     │ │
│  │   ↳ visibleChainIds                                 │ │
│  │   ↳ hiddenChainCount     (drives "Show N more")    │ │
│  └────────────────────────────────────────────────────┘ │
│                            │                             │
│                            ▼                             │
│  ┌────────────────────────────────────────────────────┐ │
│  │ STEP 2 — classify each node                         │ │
│  │ ────────────────────────────────────────────────── │ │
│  │ helper: findPrimaryBottleneck                       │ │
│  │   ↳ primaryId  (chain node w/ largest self-delay)   │ │
│  │ helper: computeOffPathCounts                        │ │
│  │   ↳ offPathCountById  (drives "+N off-path" chip)   │ │
│  │ helper: classifyNode  (called per node)             │ │
│  │   ↳ kind = healthy | primary | upstream             │ │
│  │                                                     │ │
│  │ assemble: probeNode + chainNodes (RenderableNode[]) │ │
│  │ assemble: chainEdges (filtered CriticalPathRow[])   │ │
│  └────────────────────────────────────────────────────┘ │
│                            │                             │
│                            ▼                             │
│  ┌────────────────────────────────────────────────────┐ │
│  │ STEP 3 — extend the node to show it's offPath siblings    │ │
│  │ ────────────────────────────────────────────────── │ │
│  │ if expandedBottleneckId is set AND visible:         │ │
│  │   ↳ look up its downstream child                    │ │
│  │   ↳ pull rows where childId === that & !bottleneck  │ │
│  │   ↳ splice in as offPath nodes + dashed edges       │ │
│  │ else: skip (chain-only result)                      │ │
│  └────────────────────────────────────────────────────┘ │
└───────────────────────────┬─────────────────────────────┘
                            │
                            ▼
┌─────────────────────────────────────────────────────────┐
│  OUTPUT: GraphView                                       │
│  ─ nodes: RenderableNode[]   (probe + chain + offPath)   │
│  ─ edges: CriticalPathRow[]  (chain + offPath edges)     │
│  ─ hiddenChainCount                                      │
└───────────────────────────┬─────────────────────────────┘
                            │
                            ▼
┌─────────────────────────────────────────────────────────┐
│  CriticalPathGraph  (component, wires render)            │
│  ─ useDagreGraph(nodes, edges)  → x/y positions          │
│  ─ <CriticalPathGraphNode>      per node                 │
│  ─ <CriticalPathEdgeLine>       per edge (dashed=offPath)│
│  ─ "Show N more upstream" button + "N off-path" link     │
└───────────────────────────┬─────────────────────────────┘
                            │ user clicks
                            ▼
┌─────────────────────────────────────────────────────────┐
│  React state setters                                     │
│  ─ "Show N more"  → setVisibleDepth(Infinity)            │
│  ─ "N off-path"   → setExpandedBottleneckId(node.id)     │
└───────────────────────────┬─────────────────────────────┘
                            │
                            └──► re-render → buildGraphView runs again

```

* `Freshness breakdown chip` :  If there is a self lag of an object then it highlights that on the graph
* Cluster metrics: we needed to show cluster metrics so reused some infrastructure from replicaUtilizationHistory to show cluster metrics for the object as a point in time
* Added a breadcrumbs component: If a user is looking at the DAG, then needs to drilldown and look at their journey breadcrumbs help navigate that. Since it is navigating to a different page, so it was a little hacky to showcase that journey.

### DIFF
--- a/console/src/api/materialize/maintained-objects/criticalPath.test.sql.ts
+++ b/console/src/api/materialize/maintained-objects/criticalPath.test.sql.ts
@@ -1,0 +1,242 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+import {
+  executeSqlHttp,
+  QUICKSTART_CLUSTER,
+} from "~/test/sql/materializeSqlClient";
+import { testdrive } from "~/test/sql/mzcompose";
+
+import { SEARCH_PATH } from "../executeSql";
+import { buildCriticalPathLiveQuery, CriticalPathRow } from "./criticalPath";
+
+/** Creates the test catalog tables that the query reads from. */
+const seedCatalogSql = (testSchema: string) => `
+      > DROP SCHEMA IF EXISTS ${testSchema} CASCADE;
+      > CREATE SCHEMA ${testSchema};
+      > CREATE TABLE ${testSchema}.mz_materialization_dependencies (
+          object_id TEXT NOT NULL,
+          dependency_id TEXT NOT NULL
+        );
+      > CREATE TABLE ${testSchema}.mz_wallclock_global_lag_recent_history (
+          object_id TEXT NOT NULL,
+          lag INTERVAL,
+          occurred_at TIMESTAMP
+        );
+      > CREATE TABLE ${testSchema}.mz_object_dependencies (
+          object_id TEXT NOT NULL,
+          referenced_object_id TEXT NOT NULL
+        );
+      > CREATE TABLE ${testSchema}.mz_objects (
+          id TEXT NOT NULL,
+          type TEXT,
+          cluster_id TEXT
+        );
+      > CREATE TABLE ${testSchema}.mz_object_fully_qualified_names (
+          id TEXT NOT NULL,
+          name TEXT,
+          schema_name TEXT,
+          database_name TEXT
+        );
+      > CREATE TABLE ${testSchema}.mz_clusters (
+          id TEXT NOT NULL,
+          name TEXT
+        );
+      > CREATE TABLE ${testSchema}.mz_sources (
+          id TEXT NOT NULL,
+          type TEXT
+        );`;
+
+// Bucket size large enough that all seed rows in the same `now()` second land
+// in one bucket — the probe-peak snapshot returns every object's lag from
+// that bucket.
+const TEST_BUCKET_MS = 60 * 60 * 1000;
+
+const runQuery = async (
+  testSchema: string,
+  probeId: string,
+  lookbackMinutes = 60,
+) => {
+  const compiled = buildCriticalPathLiveQuery(
+    probeId,
+    lookbackMinutes,
+    TEST_BUCKET_MS,
+  );
+  const result = await executeSqlHttp(compiled, {
+    sessionVariables: {
+      cluster: QUICKSTART_CLUSTER,
+      search_path: `${testSchema}, ${SEARCH_PATH}`,
+    },
+  });
+  return result.rows as CriticalPathRow[];
+};
+
+describe("buildCriticalPathLiveQuery", () => {
+  /**
+   * Linear chain — probe at u3, two upstream hops back to source u1.
+   *
+   *                ┌──────────────────┐
+   *                │   u1  src · 2s   │  source
+   *                └────────┬─────────┘
+   *                         │
+   *                         ▼
+   *                ┌──────────────────┐
+   *                │   u2  mid · 5s   │
+   *                └────────┬─────────┘
+   *                         │
+   *                         ▼
+   *                ┌──────────────────┐
+   *                │  u3  probe · 8s  │  ← queried object
+   *                └──────────────────┘
+   *
+   * Expectation: two chain edges (u1→u2 and u2→u3), both isBottleneck=true,
+   * each row carrying its own lag and full metadata.
+   */
+  it("walks a linear chain back to the source", async () => {
+    const testSchema = "test_critical_path";
+
+    await testdrive(`${seedCatalogSql(testSchema)}
+      > INSERT INTO ${testSchema}.mz_materialization_dependencies VALUES
+          ('u3', 'u2'),
+          ('u2', 'u1');
+      > INSERT INTO ${testSchema}.mz_wallclock_global_lag_recent_history VALUES
+          ('u1', INTERVAL '2 SECONDS', now()),
+          ('u2', INTERVAL '5 SECONDS', now()),
+          ('u3', INTERVAL '8 SECONDS', now());
+      > INSERT INTO ${testSchema}.mz_objects VALUES
+          ('u1', 'source', 'c1'),
+          ('u2', 'materialized-view', 'c1'),
+          ('u3', 'materialized-view', 'c1');
+      > INSERT INTO ${testSchema}.mz_object_fully_qualified_names VALUES
+          ('u1', 'src', 'public', 'mz'),
+          ('u2', 'mid', 'public', 'mz'),
+          ('u3', 'probe', 'public', 'mz');
+      > INSERT INTO ${testSchema}.mz_clusters VALUES ('c1', 'compute');
+    `);
+
+    const rows = (await runQuery(testSchema, "u3")).sort((a, b) =>
+      a.id.localeCompare(b.id),
+    );
+
+    expect(rows).toHaveLength(2);
+    expect(rows[0]).toMatchObject({
+      id: "u1",
+      childId: "u2",
+      name: "src",
+      objectType: "source",
+      clusterName: "compute",
+      isBottleneck: true,
+      parentSourceId: null,
+    });
+    expect(rows[0].lag?.toPostgres()).toBe("2 seconds");
+    expect(rows[1]).toMatchObject({
+      id: "u2",
+      childId: "u3",
+      name: "mid",
+      objectType: "materialized-view",
+      isBottleneck: true,
+    });
+    expect(rows[1].lag?.toPostgres()).toBe("5 seconds");
+  });
+
+  /**
+   * Tied bottlenecks — three sources all at 4s feeding the probe.
+   *
+   *      ┌────────┐  ┌────────┐  ┌────────┐
+   *      │ u1·4s  │  │ u2·4s  │  │ u3·4s  │  all tied
+   *      └────┬───┘  └────┬───┘  └────┬───┘
+   *           │           │           │
+   *           └───────────┼───────────┘
+   *                       ▼
+   *             ┌──────────────────┐
+   *             │  u9  probe · 5s  │
+   *             └──────────────────┘
+   *
+   * Expectation: all three rows have isBottleneck=true. No arbitrary
+   * tie-breaking — every input that ties for max lag stays on the chain.
+   */
+  it("flags every input tied for max lag as a chain bottleneck", async () => {
+    const testSchema = "test_critical_path";
+
+    await testdrive(`${seedCatalogSql(testSchema)}
+      > INSERT INTO ${testSchema}.mz_materialization_dependencies VALUES
+          ('u9', 'u1'),
+          ('u9', 'u2'),
+          ('u9', 'u3');
+      > INSERT INTO ${testSchema}.mz_wallclock_global_lag_recent_history VALUES
+          ('u1', INTERVAL '4 SECONDS', now()),
+          ('u2', INTERVAL '4 SECONDS', now()),
+          ('u3', INTERVAL '4 SECONDS', now()),
+          ('u9', INTERVAL '5 SECONDS', now());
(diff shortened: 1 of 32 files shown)


SangJunBak:
Didn't realize this before but we should change the commit messages to follow this guide 

https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#change-descriptions

i.e. have them imperative and more descriptive than "addressed feedback"

leedqin:
> 

> Didn't realize this before but we should change the commit messages to follow this guide
> 
> https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#change-descriptions
> 
> i.e. have them imperative and more descriptive than "addressed feedback"

I will update the commits to follow these guidelines. also updated the description in the Tips For Reviewers and the PR description with couple other commits that I added

console/src/components/FreshnessGraph/FreshnessGraph.tsx
```diff
@@ -63,6 +65,11 @@ export type FreshnessGraphProps = {
     lineData: GraphLineSeries;
   }) => React.ReactNode;
   maxTooltipLines?: number;
+  /** Fires as the cursor scrubs across the graph. Debounced (150ms). */
+  onPointSelect?: (timestamp: number | null) => void;
+  onPointLock?: (timestamp: number | null) => void;
```
SangJunBak: Not sure how I feel about the words "Select" and "Lock" given "Select" can be conflated with the `select` DOM event. Why not keep parity with the traditional DOM events like `onPointMove` and `onClick`?

console/src/api/materialize/maintained-objects/criticalPath.ts
```diff
@@ -0,0 +1,239 @@
...
+    objectId,
+    perObjectLagCtePMaxInWindow(lookbackMinutes),
+  );
+
+/**
+ * Compiles the critical-path query for the user's selected `timestamp`. Each
+ * upstream's lag is read from the bucket aligned to that timestamp, so the
+ * chain values match what the freshness chart drew at that point.
+ */
+export const buildCriticalPathAtTimeQuery = (
```
jubrad, line 242: we have a test for `buildCriticalPathPMaxInWindowQuery`, it might also be nice to have a test for `buildCriticalPathAtTimeQuery`

console/src/api/materialize/maintained-objects/criticalPath.test.sql.ts
```diff
@@ -0,0 +1,236 @@
...
+   *           │           │           │
+   *           └───────────┼───────────┘
+   *                       ▼
+   *             ┌──────────────────┐
+   *             │  u9  probe · 5s  │
+   *             └──────────────────┘
+   *
+   * Expectation: all three rows have isBottleneck=true. No arbitrary
+   * tie-breaking — every input that ties for max lag stays on the chain.
+   */
```
jubrad, line 163: One thing to think about here in the critical path is the set of nodes with edges that have 0 slack (the time you can give up before impacting the overall duration), I think showing just the critical path is a bit of a weak story that is likely to lead to whack-a-mole rather than a full understanding of bottle necks. Don't worry about this for now, it was a design decision and we're not changing it here.

console/src/components/FreshnessGraph/FreshnessGraph.tsx
```diff
@@ -354,14 +372,59 @@ const FreshnessGraphInner = (
         ))}
         <GraphEventOverlay
           {...graphEventOverlayProps}
+          style={{
+            cursor: props.onPointLock
+              ? props.isLocked
+                ? "pointer"
+                : "crosshair"
+              : "default",
+          }}
```
SangJunBak: i feel like rather than deciding styling on callback methods, maybe we should decide on separate the two and create like a `variant` prop instead?

console/src/platform/maintained-objects/queries.ts
```diff
@@ -204,3 +211,35 @@ export function useObjectSourceStatistics(sourceId: string) {
...
+  [
+    ...buildRegionQueryKey("maintainedObjects"),
+    buildQueryKeyPart("freshnessHistory", { objectId, lookbackMs }),
+  ] as const;
+
+export function useObjectFreshnessHistory({
+  objectId,
+  lookbackMs,
+}: {
+  objectId: string | undefined;
```
SangJunBak: Why is objectId nullable here? Usage of this hook shows me that it's always defined

console/src/utils/graph.ts
```diff
@@ -24,6 +24,25 @@ export function niceTicks(start: number, stop: number, desiredCount: number) {
   return ticks(niceStart, niceStop, desiredCount);
 }
 
+/** Map a pointer event to its cursor timestamp in ms, or null if outside `[startTime, endTime]`. */
+export function eventToTimestamp({
```
SangJunBak, line 28: i wonder if we should return null or simply just clamp it to both bounds?

console/src/platform/maintained-objects/freshnessHistory.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export interface ObjectFreshnessHistory {
+  historicalData: DataPoint[];
+  lines: GraphLineSeries[];
+  bucketSizeMs: number;
+  startTime: number;
+  endTime: number;
+}
+
+export const buildObjectFreshnessHistory = (
```
SangJunBak: nit: We usually reserve the `build` prefix for SQL builders. Wondering if we should rename to `transform` or something?

console/src/api/materialize/maintained-objects/criticalPath.ts
```diff
@@ -0,0 +1,237 @@
...
+const buildCriticalPathQuery = (
+  objectId: string,
+  perObjectLag: RawBuilder<unknown>,
+): CompiledQuery<CriticalPathRow> =>
+  sql<CriticalPathRow>`
+WITH MUTUALLY RECURSIVE
+  input_of(source text, target text) AS (
+    SELECT dependency_id, object_id
+    FROM mz_compute_dependencies
+  ),
```
SangJunBak, line 145: I think this query misses a few cases like sinks. Can refer to `buildWorkflowGraphQuery`

console/src/api/materialize/maintained-objects/criticalPath.ts
```diff
@@ -0,0 +1,237 @@
...
+    UNION
+    SELECT ci.source, ci.target
+    FROM delayed_by db
+    JOIN crit_input ci ON db.source = ci.target
+  ),
+  chain_targets(target text) AS (
+    SELECT DISTINCT target FROM delayed_by
+  ),
+  -- Subsources are filtered out of the maintained-objects list, so the panel
+  -- redirects drawer links to the parent source.
```
SangJunBak, line 181: I forget but do subsources and sources have the same lag?

console/src/api/materialize/maintained-objects/criticalPath.ts
```diff
@@ -0,0 +1,237 @@
...
+ * user selected in the maintained objects page.
+ */
+const perObjectLagCtePMaxInWindow = (lookbackMinutes: number) => {
+  const lookback = sql.raw(`${lookbackMinutes}`);
+  return sql`per_object_lag(object_id text, lag interval) AS (
+    SELECT object_id, max(lag) AS lag
+    FROM mz_wallclock_global_lag_recent_history
+    WHERE occurred_at + INTERVAL '${lookback} MINUTES' >= mz_now()
+    GROUP BY object_id
+  )`;
```
SangJunBak, line 83: More of product feedback, but one thing that concerns me for `perObjectLagCtePMaxInWindow` is we're effectively getting the max lags of all objects in the DAG for that time period. However that means we're showing a single graph across different snapshots of time. This could lead to inconsistent state where if the lookback is 3 hours:

1pm: 20s ->20s -> 20s   (the root and end node is the bottleneck)
2pm: 1s -> 1s -> 40s (the end node is the bottleneck)

The final graph looks like:
20s -> 20s -> 40s

which isn't an accurate view of what happened.

console/src/platform/maintained-objects/freshnessBreakdown.ts
```diff
@@ -0,0 +1,50 @@
...
+  const maxInputMs = Math.max(0, ...lagsMs);
+  const selfDelayMs = targetMs - maxInputMs;
+  if (selfDelayMs <= 0) return null;
+  const dominant = data.filter(
+    (r, i) => maxInputMs > 0 && lagsMs[i] === maxInputMs,
+  );
+  return {
+    selfDelayMs,
+    maxInputMs,
+    isSelfBottleneck: selfDelayMs > maxInputMs,
```
SangJunBak, line 46: I don't quite get this logic. Isn't this impossible? 
```
targetMs <= maxInputMs // Target is always less or equal to the max
selfDelayMs = targetMs - maxInputMs

selfDelayMs > maxInputMs
targetMs - maxInputMs > maxInputMs
targetMs > 2 * maxInputMs // target must be at least 2* bigger than the max, which contradicts the above
```

SangJunBak (commented):
Overall looks good!

SangJunBak:
For views, noticing an empty DAG and smaller than usual breadcrumb text

Screenshot 2026-05-19 at 11 58 54 AM

console/src/api/materialize/maintained-objects/criticalPath.ts
leedqin: I believe they advance together because they share the same ingestion. subsources do write their own lag to the frontiers but in practice I have seen both of them to share the same lag

console/src/platform/maintained-objects/freshnessBreakdown.ts
leedqin: image
`stg_mqis_attr_cat_attr_id_sid_id_cat_id_idx `has a 3s lag while its only upstream `(mqis_avro_attr_cat)` is at 1s, so selfDelay = 2s > maxInput = 1s and isSelfBottleneck flips. The strip highlights the 2s self-delay in red so the user looks at the index itself rather than chasing the source.

SangJunBak (approved):
LGTM!

PR #36414, opened 2026-05-05
cli: rename mz to mzx

sjwiesman:
Renamed the mz cli to mzx to free up the name `mz`. The diff is large but trivial it's just a lot of renames across many files. Also, this seemingly didn't compile so I fixed that in the last commit. 

### Motivation

Closes <issue>
Fixes <bug>
etc.

### Description

diff *can't* tell a reviewer.

### Verification


sjwiesman:
@def- could you take a quick look at this? I suspect I messed up something obvious. All of the tests are filling with the same error.

def-:
You didn't mess anything up! I just need to make the new mzx package name public on Dockerhub and GHCR, then it should work. Doing that now, then retriggering Test and Nightly pipeline to verify. Edit: Done: https://buildkite.com/materialize/test/builds/122376 & https://buildkite.com/materialize/nightly/builds/16323 Edit2: All good from CI side

def- (commented):
Test changes lgtm. I'm not sure if `mzx` is the least confusing name for this. Maybe `mz-cli-deprecated` if we plan to fully replace it and not work on it.

kay-kim (commented):
I'm wondering if we want to add redirects or yolo it.
- If only internal use, probably don't need redirects.
- If we do want to add redirects, I don't think we want to send them to the developer docs
   - Internal people would already know
   - External people might get confused that we take them out of the docs and send them to github dev docs.

   Instead, if we do decide to redirect, I'd probably send them to the /integrations page.

PR #36415, opened 2026-05-06
feat(orchestratord): [CLO-46] add app_name() to ManagedResource trait for consistent app.kubernetes.io/name labeling.

rjimeno-mz:
As a platform engineer monitoring the Materialize Cloud cluster, 
I want the environmentd StatefulSet and its pods to have the standard app.kubernetes.io/name label, 
so that the workload and its running instances are consistently discoverable and selectable alongside the rest of the resources in the cluster that already follow this Kubernetes best practice.

### Motivation

Most of Materialize's Deployments and pods have the label `app.kubernetes.io/name` set.
Environmentd StatefulSet and Pods do not have the label `app.kubernetes.io/name` set, and that makes them harder to find and select; particularly when it comes to observability (logs and metrics).

After this change, Environmentd StatefulSets and Pods will have the same labels as most other Deployments, StatefulSets, and Pods, making them easier to find and select. That best practice is useful and handy for observability.

### Description

This change creates `app.kubernetes.io/name` labels with value `environmentd` and assigns them to Pod templates and StatefulSets.

### Verification

After this change is deployed in the staging environment, the commands `kubectl config use-context mzcloud-staging-us-east-1-0` and `kubectl get deployments,pods -A -l app.kubernetes.io/name=environmentd` will return resources (before this change, they return no resources).

### Obs.:

This change requires another change in the `cloud` repository to "bump up" the submodule pointer.

### DIFF
--- a/src/orchestratord/src/controller/materialize/generation.rs
+++ b/src/orchestratord/src/controller/materialize/generation.rs
@@ -1282,19 +1282,12 @@ fn create_environmentd_statefulset_object(
         ..Default::default()
     };
 
-    let mut statefulset_labels = mz.default_labels();
-    statefulset_labels.insert(
-        "app.kubernetes.io/name".to_owned(),
-        "environmentd".to_string(),
-    );
-
     StatefulSet {
         metadata: ObjectMeta {
             annotations: Some(btreemap! {
                 "materialize.cloud/generation".to_owned() => generation.to_string(),
                 "materialize.cloud/force".to_owned() => mz.spec.force_rollout.to_string(),
             }),
-            labels: Some(statefulset_labels),
             ..mz.managed_resource_meta(mz.environmentd_statefulset_name(generation))
         },
         spec: Some(statefulset_spec),
--- a/src/orchestratord/src/controller/materialize/generation.rs
+++ b/src/orchestratord/src/controller/materialize/generation.rs
@@ -1132,6 +1132,10 @@ fn create_environmentd_statefulset_object(
         mz.environmentd_app_name(),
     );
     pod_template_labels.insert("app".to_owned(), "environmentd".to_string());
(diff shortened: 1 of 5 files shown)


rjimeno-mz:

rjimeno-mz:

rjimeno-mz:

rjimeno-mz:
recheck

src/orchestratord/src/controller/materialize/generation.rs
```diff
@@ -1278,12 +1282,19 @@ fn create_environmentd_statefulset_object(
...
+    );
+
     StatefulSet {
         metadata: ObjectMeta {
             annotations: Some(btreemap! {
                 "materialize.cloud/generation".to_owned() => generation.to_string(),
                 "materialize.cloud/force".to_owned() => mz.spec.force_rollout.to_string(),
             }),
+            labels: Some(statefulset_labels),
             ..mz.managed_resource_meta(mz.environmentd_statefulset_name(generation))
```
Alphadelta14, line 1291: Wonder if we should pass app_name into managed_resource_meta, since it is already setting labels (and then would be a consistent pattern for balancerd, console, orchestratord).

NOTE: name is the name of the resource (environmentd-my-gen-abcd), but app_name would be a constant that handles all of those for all resources.

alex-hunt-materialize (approved):
I also like @Alphadelta14 's suggestion of adding it to the managed_resource_meta.

src/orchestratord/src/controller/materialize/generation.rs
rjimeno-mz: I think that's a good idea, as long as I understood it correctly.

Please re-review carefully, as I might have missed some details or the "big picture".

PR #36416, opened 2026-05-06
Remove `mz-lsp-server` crate

sjwiesman:
## Summary

Deletes the standalone `mz-lsp-server` crate along with its release pipeline (`ci/deploy_mz_lsp_server/`), the `MzLspServerVersion` tagging class, and its workspace registration. 

## Motivation

The `mz-lsp-server` crate has been effectively unmaintained for some time and offers only limited functionality compared to a modern LSP. Rather than continue to invest in it, we are removing it now and replacing it with two parallel efforts:

* **LSP-like editing functionality embedded directly in the Materialize console**, so the most common authoring flows (parsing, formatting, completion) are available where users already write SQL.
* **A new v2 LSP server with substantially expanded functionality**, planned to ship as part of mz 2.0.

Removing the legacy crate now keeps us from having to maintain three SQL-tooling implementations in parallel during the transition.

## Scope of the change

* Deletes `src/lsp-server/`, `ci/deploy_mz_lsp_server/`, and `doc/developer/generated/lsp-server/`.
* Removes `src/lsp-server` from `[workspace] members` and `[workspace] default-members` in the root `Cargo.toml`.
* Removes `"deploy-mz-lsp-server"` from `MZ_PIPELINES` in `misc/python/materialize/buildkite_insights/buildkite_api/buildkite_config.py`.
* Removes the `MzLspServerVersion` class from `misc/python/materialize/mz_version.py` (its only consumers lived inside `ci/deploy_mz_lsp_server/`, which is also being deleted).

No other workspace crate depends on `mz-lsp-server`, and no editor client in this repo launches it as a subprocess, so the removal is self-contained.


def- (approved):
No concerns from QA side

PR #36420, opened 2026-05-06
console: guard useAuth() in OIDC routes when init fails

def-:
#36193 made OidcProviderWrapper render without `<AuthProvider>` when OIDC initialization fails so password sign-in still works, but OidcAuthGuard and OidcCallback still deref `auth` unconditionally and crash with TypeError. Add `!auth` guards matching the pattern already used in SsoLoginLink: OidcAuthGuard falls through to children (so a valid password session cookie still reaches the app), and OidcCallback redirects to LOGIN_PATH since there's nothing to complete without an AuthProvider.


leedqin (approved):
LGTM!

PR #36425, opened 2026-05-06
compute: temporal bucket via merge-batcher seal coarsening

antiguru:
### Motivation

Replace the operator-based temporal bucketing on imported sources with an integration into the merge batcher inside arrangements. The previous design attaches a separate Timely operator to coalesce future-stamped updates; review feedback rejected the side-channel `Plan` field and the `RenderTimestamp` extension that drove that approach.

### Description

`TemporalBucketingBatcher<B>` wraps an inner `Batcher`. `seal(upper)` snaps `upper` to `last_lower + 2^k` for the largest `k` where the result still satisfies `<= upper`, then forwards the coarsened upper to the inner batcher. `push_container` and `frontier` forward unchanged: held updates remain in the inner batcher and `frontier()` reports their min time, holding back the surrounding arrange operator's capability until the next bucket boundary. Empty or multi-element antichains pass through.

`MzArrangeTemporalCore` mirrors `mz_arrange_core` and is implemented for streams whose timestamp implements `BucketTimestamp`. `RenderTimestamp` gains `BucketTimestamp` as a supertrait. Iterative-scope timestamps get a degenerate `BucketTimestamp` impl in `mz_timely_util::temporal` whose `advance_by_power_of_two` always returns `None`, making the wrapper a no-op there.

`ensure_collections` reads `ENABLE_TEMPORAL_BUCKETING` and routes the row-row arrangement through the temporal variant. The previous operator-based source-side bucketing is removed, and `TEMPORAL_BUCKETING_SUMMARY` (which served only the bypass path) is dropped.

Unit tests in `extensions::temporal_bucketing_batcher` cover `bucket_align` selection and the hold-and-release behavior against a `MergeBatcher<Vec<...>, ContainerChunker, VecMerger>`.


antiguru:
Closing — wrong mechanism. The wrapper coarsened `seal(upper)`, which violates the `Batcher` contract (`seal` must return updates strictly less than `upper`). The plan has been rewritten to vendor DD's `MergeBatcher` / `InternalMerge` into `mz_timely_util` and integrate `BucketChain` inside the vendored copy, so the bucket-aware logic operates on chunks below the seal layer. Will reopen as a fresh PR.

PR #36429, opened 2026-05-06
compute: Fix UpdateNotBeyondLower panic in v2 MV sink at startup

antiguru:
### Motivation

The v2 MV sink (`ENABLE_SYNC_MV_SINK`) panics at startup under
`MV_SINK_ADVANCE_PERSIST_FRONTIERS`:

```
thread 'tokio:work-2' panicked at src/compute/src/sink/materialized_view_v2.rs:831:22:
valid usage: UpdateNotBeyondLower {
    ts: 1778067290977,
    lower: Antichain { elements: [1778067594672] }
}
```

v2 splits sink state across a Timely thread (holds `persist_frontiers`,
the active batch description) and a Tokio task (holds the `corrections`
buffer and `WriteHandle`).
With the flag enabled, `write::State::new` advanced the Timely-side
`persist_frontiers` to the sink's `as_of` so the `maybe_start_batch` gate
(`desc.lower <= persist_frontiers.frontier()`) opened immediately for the
first description minted with `lower = as_of`.

The Tokio-side `corrections.since` only moves forward through
`WriteCommand::Batch { persist_frontier, .. }`, which the Timely closure
populates only when a persist input frontier *actually advances*.
At startup the persist input frontiers begin at `Timestamp::MIN`, so no
`Batch` carries `persist_frontier` until the persist input catches up — yet
a `WriteBatch(desc)` with `desc.lower = as_of` could already be sent.
Snapshot-replay updates inserted in the meantime stayed at their original
timestamps (`Correction::insert` rounds to `max(t, since)` and
`since == MIN`) and slipped into the written batch, tripping persist's
`UpdateNotBeyondLower` invariant.

The v1 sink does not have this bug because corrections live inside `State`;
`apply_persist_frontier_advancement` synchronously calls
`corrections.advance_since` for every persist-frontier advancement, so the
invariant `corrections.since == persist_frontiers.frontier()` is upheld
trivially.

### Description

Restore the v1 invariant in v2:

* Read `MV_SINK_ADVANCE_PERSIST_FRONTIERS` exactly once in `write::render`
  and reuse the captured value in both initialization sites.
  Re-reading the dyncfg twice could let the value flip between reads and
  reproduce the same divergence the fix is meant to close
  (`persist_frontiers = as_of` with `corrections.since = MIN`).
* Advance `corrections.{ok, err}.since` to `as_of` on the Timely thread
  before the corrections buffer is moved into the Tokio task, whenever the
  captured flag is enabled.
* Thread the captured bool into `State::new` instead of having it re-read
  the dyncfg.
* Mark `advance_persist_ok_frontier` and `advance_persist_err_frontier`
  `#[must_use]`.
  Their bool return value is the trigger for queuing a
  `corrections.advance_since` send to the Tokio task; silently dropping it
  is exactly the bug class fixed here.

### Verification

`cargo check`, `cargo fmt`, `cargo clippy -p mz-compute --no-deps` clean.
Existing platform-checks coverage exercises the v2 sink hot path that
triggered the panic.

### DIFF
--- a/src/compute/src/sink/materialized_view_v2.rs
+++ b/src/compute/src/sink/materialized_view_v2.rs
@@ -641,6 +641,35 @@ mod write {
             ),
         );
 
+        // Read `MV_SINK_ADVANCE_PERSIST_FRONTIERS` exactly once and reuse the captured value for
+        // both the Tokio-side `corrections.since` initialization below and the Timely-side
+        // `persist_frontiers` initialization in `State::new`. Re-reading the dyncfg per init site
+        // would let the value flip between reads and produce the very inconsistency this fix
+        // addresses: `persist_frontiers = as_of` (gate open) with `corrections.since = MIN`
+        // (snapshot updates not advanced) reproduces the original `UpdateNotBeyondLower` panic.
+        let advance_persist_frontiers_at_startup =
+            MV_SINK_ADVANCE_PERSIST_FRONTIERS.get(&worker_config);
+
+        // Mirror the persist-frontier initialization performed by `State::new` below. With the
+        // flag enabled, `State` advances its Timely-side `persist_frontiers` to `as_of`, opening
+        // the `maybe_start_batch` write gate (`desc.lower <= persist_frontiers.frontier()`)
+        // immediately for the first description minted with `lower = as_of`. The corrections
+        // buffer lives on the Tokio task and only learns of frontier advancements through
+        // `WriteCommand::Batch { persist_frontier, .. }`, which the Timely closure populates only
+        // when an input frontier actually moves. On startup, the input frontiers begin at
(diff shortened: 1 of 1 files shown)


def- (commented):
No reliable way of reproducing it in a regression test? Maybe we could add something in test/race-condition? I can also try

def-:
Also seen in https://buildkite.com/materialize/nightly/builds/16333#019dfdeb-0204-4f50-96df-af10a6f4fa4c, but I assume nothing left to be done since we disabled the feature flag already, so no need to backport this.

def-:
Would be nice to fix since it's pretty flaky in CI, another occurrence: https://buildkite.com/materialize/nightly/builds/16368

DAlperin (approved):
yep good fix

PR #36440, opened 2026-05-07
fix(adapter): preserve workload_class during zero-downtime cluster alter

jubrad:
## Summary

- `ALTER CLUSTER ... SET (SIZE, WORKLOAD CLASS) WITH (WAIT FOR / UNTIL READY)` was silently dropping the workload class change because the zero-downtime finalization code read `workload_class` from the catalog (old value) instead of the planned config.
- Thread the new `workload_class` through `AlterClusterWaitForHydrated` and `AlterClusterFinalize` so `finalize_alter_cluster_stage` uses the planned value.
- Add regression test covering both `WAIT FOR` and `WAIT UNTIL READY` paths.

Fixes SQL-211.

## Test plan

- [x] `bin/sqllogictest test/sqllogictest/managed_cluster.slt` passes with the new regression test
- [x] Existing managed cluster tests continue to pass

### DIFF
--- a/src/adapter/src/coord.rs
+++ b/src/adapter/src/coord.rs
@@ -832,6 +832,7 @@ pub struct AlterClusterWaitForHydrated {
     validity: PlanValidity,
     plan: plan::AlterClusterPlan,
     new_config: ClusterVariantManaged,
+    workload_class: Option<String>,
     timeout_time: Instant,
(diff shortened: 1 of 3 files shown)


ggevay (approved):

def- (approved):
Thanks for adding the test

PR #36442, opened 2026-05-07
Script for GitHub to Linear migration

def-:
https://linear.app/materializeinc/issue/QAR-41/migrate-existing-issues-from-github-database-issues-linear

We could merge this in case someone else wants to use it. I keep using it regularly so might be worth it.


bosconi (approved):
Looks good.

PR #36443, opened 2026-05-07
sql: Remove enable_multi_replica_sources feature flag

jubrad:
## Summary

- Remove the `enable_multi_replica_sources` dyncfg flag entirely, making multi-replica sources always enabled.
- This eliminates the bug where `ALTER CLUSTER ... SET (SIZE, REPLICATION FACTOR 1) WITH (WAIT FOR ...)` bypassed the single-replica guard when `enable_multi_replica_sources=false`.
- Clean up all related dead code: the `contains_single_replica_objects` function, `CreateReplicaFailStorageObjects` error variant, and guard blocks in the sequencer and planner.

Fixes SQL-212.

## Test plan

- [x] Existing `managed_cluster.slt` tests pass (the guards being removed are dead code since the flag defaults to `true`)
- [ ] CI passes with no compilation warnings

### DIFF
--- a/misc/python/materialize/mzcompose/__init__.py
+++ b/misc/python/materialize/mzcompose/__init__.py
@@ -92,6 +92,7 @@ def get_minimal_system_parameters(
         "enable_load_generator_counter": "true",
         "enable_logical_compaction_window": "true",
         "enable_multi_worker_storage_persist_sink": "true",
+        "enable_multi_replica_sources": "true",
         "enable_rbac_checks": "true",
         "enable_reduce_mfp_fusion": "true",
         "enable_refresh_every_mvs": "true",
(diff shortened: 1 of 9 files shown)


Alphadelta14 (approved):

misc/python/materialize/mzcompose/__init__.py
```diff
@@ -92,7 +92,6 @@ def get_minimal_system_parameters(
         "enable_load_generator_counter": "true",
         "enable_logical_compaction_window": "true",
         "enable_multi_worker_storage_persist_sink": "true",
-        "enable_multi_replica_sources": "true",
```
def-, line 95: The only concern I have is that we might need this for tests running on older versions, so I triggered nightly: https://buildkite.com/materialize/nightly/builds/16363

def- (approved):
Ok if nightly is green

misc/python/materialize/mzcompose/__init__.py
def-: Fails indeed. So let's just keep this line for now: https://buildkite.com/materialize/nightly/builds/16363#019e051c-d34f-4514-9e78-a0665d557364

def- (changes requested):

jubrad:
I think we're good if this passes
https://buildkite.com/materialize/nightly/builds/16371

def- (approved):

PR #36444, opened 2026-05-07
persist: add missing variants to BatchColumnarFormat Arbitrary impl

def-:
The proptest Arbitrary impl for BatchColumnarFormat only generated Row, Both(0), and Both(1), missing Both(2) (the production default) and Structured. This meant proptest round-trip tests for HollowBatchPart never exercised the formats actually used in production.


def-:
Ready for review

DAlperin (approved):
LGTM, thanks!

PR #36448, opened 2026-05-07
persist: Accept plain compression names without level suffix

def-:
"brotli", "gzip", and "zstd" (without a dash-level suffix) silently fell through to CompressionFormat::None since #28678. Use the default compression level for these plain names instead.


def-:
Ready for review

DAlperin (approved):

DAlperin (approved):
Thanks!

PR #30470, opened 2024-11-13
alter_table: Support adding columns to tables

ParkMyCar:
This PR implements the SQL feature `ALTER TABLE ... ADD COLUMN ...`.

Note: There are a lot of lines changed but the majority are new tests!

Specifically it:
1. Uses `VersionedRelationDesc` on `Table`s to track new columns
2. Adds a `CatalogCollectionEntry` which adds some typing around getting the current `RelationDesc` for an entry.
3. Updates the `storage-controller` to create new Persist `WriteHandle`s and pass them to the `TxnsTableWorker`
4. When creating a new version of a Table, will add the new version as dependency for the old version. This proved necessary to get proper read handles for objects on top of tables.

Otherwise it also adds several tests:
1. `alter-table.slt` which exercises a number of different scenarios
2. A new `Check` for the `platform-check`s test framework
3. A new `Action` for the `parallel-workload` test framework.
    * This new action is currently disabled because off a race condition in Persist's schema registry
4. A legacy upgrade test to make sure we have coverage on the restart of MZ test case.

### Motivation

Fixes https://github.com/MaterializeInc/database-issues/issues/8233

### Tips for reviewer

I split the PR up into separate commits to ideally make it easier to review, most the changes here are new tests!

1. Changes to the Catalog APIs and name resolution to support versions.
2. Changes to sequencing and the storage controller.
3. Tests
4. Formatting and clippy

### Checklist

- [x] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [x] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [x] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [x] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
[...]

PR #36449, opened 2026-05-07
audit: add missing audit log entry for `ALTER TABLE ADD COLUMN` and `ALTER SOURCE ... SET (TIMESTAMP INTERVAL)`

def-:
The `Op::AlterAddColumn` handler was missing an audit log entry, unlike every other ALTER operation. This adds an `AlterAddColumnV1` event that records the table ID, column name, column type, and nullability.

Introduced in #30470.

Nightly: https://buildkite.com/materialize/nightly/builds/16495


def-:
Ready for review

aljoscha (changes requested):
I think this looks good, but will have to be rebased on top of the latest catalog migration shenanigans

def-:
Done

src/adapter/src/catalog/transact.rs
```diff
@@ -887,16 +887,36 @@ impl Catalog {
...
+                    .update_timestamp_interval(value.clone(), interval)
                     .map_err(|_| {
                         AdapterError::Catalog(Error::new(ErrorKind::Internal(
                             "planner should have rejected invalid alter timestamp interval item type"
                                 .to_string(),
                         )))
                     })?;
 
+                if Self::should_audit_log_item(new_entry.item()) {
+                    let details = EventDetails::AlterSourceTimestampIntervalV1(
```
aljoscha, line 901: This is a nit, so please ignore if you don't feel like it: looks like we have the same shape in two call-sites, and could pull out into a shared helper. Might be that the signature of the helper is super large, though, and then we don't really gain much. So please do as you want 😅

aljoscha (approved):
nice!

PR #36305, opened 2026-04-28
persist: skip lgalloc memcpy in S3 get path

antiguru:
The S3 get path copied each `Bytes` chunk returned by the SDK into a freshly-allocated `MetricsRegion<u8>`.
With lgalloc disabled in practice that region falls back to the heap, making the copy pure overhead.
Push SDK `Bytes` straight into `SegmentedBytes` instead.

Also removes the now-unused `persist_enable_s3_lgalloc_{cc,noncc}_sizes` dyncfgs, the `cfg`/`is_cc_active` fields on `S3BlobConfig`/`S3Blob`, and the dead `LgBytesMetrics::persist_s3` metric.

### Motivation

Reduces CPU and allocator pressure on every S3 read in persist.

### Tips for reviewer

The behavior change is in `Blob::get` in `src/persist/src/s3.rs`; everything else is dead-code removal that follows.

### Checklist

- [ ] This PR has adequate test coverage / QA is not needed.
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #36451, opened 2026-05-07
persist: copy S3 SDK bytes out of hyper pool buffers

antiguru:
### Motivation

Follow-up to #36305. Dropping the `MetricsRegion<u8>` copy left the SDK `Bytes` chunks slicing into hyper's shared pool buffers. Each pool buffer stays alive as long as any chunk borrows it, so a returned blob pins the full backing allocation. Profile shows the RSS regression plus extra time in `SegmentedBytes::advance` / `get_bytes` from walking many small fragments.

### Description

Push each chunk through `Bytes::copy_from_slice` so the resulting `Bytes` owns a private allocation. Hyper's pool buffer can be released as soon as the chunk lands in `SegmentedBytes`. Cheaper than the old `MetricsRegion` round-trip and avoids the lgalloc fallback.

### Verification

`cargo check -p mz-persist`, `cargo clippy -p mz-persist --all-targets -- -D warnings`, `bin/ci-builder run stable bin/lint`.

### DIFF
--- a/src/persist/src/s3.rs
+++ b/src/persist/src/s3.rs
@@ -451,7 +451,11 @@ impl Blob for S3Blob {
                 }
 
                 while let Some(data) = object.body.next().await {
-                    body_parts.push(data.context("s3 get body err")?);
+                    let data = data.context("s3 get body err")?;
+                    // Copy out of the hyper pool buffer so we don't pin the
+                    // entire shared backing allocation for the lifetime of the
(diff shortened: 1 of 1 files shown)


DAlperin (approved):
Looks good to me for the memory usage. Still slightly worried that without coalescing the allocations we are still going to see a bunch of time in SegmentedBytes.

PR #36452, opened 2026-05-07
license_key_entitlements

alex-hunt-materialize:
Add a new `entitlements` field to our license keys, which is a list of strings.

### Motivation

Part of https://linear.app/materializeinc/issue/DEP-130/include-entitlements-in-newly-generated-license-keys
The first use case for these entitlements are for downloading the Ory Enterprise images through our registry proxy. Not all license keys will be allowed to get ory images.

### Verification

New rust tests checking serialization/deserialization both with and without entitlements.

### DIFF
--- a/src/license-keys/examples/sign.rs
+++ b/src/license-keys/examples/sign.rs
@@ -31,6 +31,10 @@ struct Opt {
     max_credit_consumption_rate: f64,
     #[clap(long, default_value_t = 365 * 24 * 60 * 60)]
     validity_secs: u64,
+    /// Repeatable: each `--entitlement <name>` adds one entitlement to the
+    /// signed key (e.g. `--entitlement ory`).
+    #[clap(long = "entitlement")]
+    entitlements: Vec<String>,
 }
 
 #[tokio::main]
@@ -53,6 +57,7 @@ async fn main() {
         opt.max_credit_consumption_rate,
         false,
         ExpirationBehavior::Warn,
+        opt.entitlements,
     )
     .await
     .unwrap();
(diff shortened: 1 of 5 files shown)


src/license-keys/examples/sign.rs
```diff
@@ -31,6 +31,10 @@ struct Opt {
     max_credit_consumption_rate: f64,
     #[clap(long, default_value_t = 365 * 24 * 60 * 60)]
     validity_secs: u64,
+    /// Repeatable: each `--entitlement <name>` adds one entitlement to the
+    /// signed key (e.g. `--entitlement ory`).
+    #[clap(long = "entitlement")]
+    entitlements: Vec<String>,
```
jubrad, line 37: I'm not positive if Vec<String> is the right shape. 
Abrief discussion:

We do have https://github.com/MaterializeInc/materialize/pull/32317/changes#diff-c01e42909e80f54d64977d77b5db88d93959e5e91454a4fcd8f650409848c3d4R144

I think this makes sense if we roughly think of these as "FeatureFlags" which are always bool, and we treat inclusion as true, and omission as take the default. If we wanted to adopt broader systemVar it could be done in a seperate field.

jubrad (commented):
Yeah, I think I've convinced myself that this makes sense.

src/license-keys/examples/sign.rs
alex-hunt-materialize: That seems very complicated. We also don't want to lock to a specific version of MZ, and we don't have access to this code from the oci-registry-proxy repo.

I see the work `entitlement` as "something you are allowed to get". To me, this is a simple boolean, and omission is false (or default I guess, but if you aren't "entitled" to it, the default is probably false).

Do you have a more concrete example of what type you want here? I don't really want to be juggling `serde_json::Value` everywhere.

jubrad (approved):

PR #36453, opened 2026-05-07
adapter: route SQL `EXECUTE` through frontend sequencing

aljoscha:
A prepared statement invoked via SQL `EXECUTE foo` previously
skipped frontend sequencing in the session task and was
re-dispatched via `Command::Execute` from the coordinator's
`Plan::Execute` handler. That made `EXECUTE foo` and the same
statement issued directly take divergent routes through the
adapter.

Unroll `Statement::Execute` in the session client: resolve the
prepared statement, install a fresh portal for the inner
statement, then run `try_frontend_peek` against that portal.
Statement logging still begins on the outer `EXECUTE foo (...)`,
so `mz_statement_execution_history` records the user-visible SQL
text and pgwire parameters rather than the inner statement.

Needed for the upcoming OCC-based read-then-write path: without
this, a read-then-write issued via SQL EXECUTE would bypass
frontend sequencing and reach the coordinator's legacy lock-based
implementation, and only one of the two can safely run at a time.

### DIFF
--- a/src/adapter/src/client.rs
+++ b/src/adapter/src/client.rs
@@ -30,6 +30,7 @@ use mz_ore::collections::CollectionExt;
 use mz_ore::id_gen::{IdAllocator, IdAllocatorInnerBitSet, MAX_ORG_ID, org_id_conn_bits};
 use mz_ore::instrument;
 use mz_ore::now::{EpochMillis, NowFn, to_datetime};
+use mz_ore::str::StrExt;
 use mz_ore::task::AbortOnDropHandle;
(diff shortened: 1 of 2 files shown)


ggevay:
I've triggered a Nightly run, because it has extra tests both for EXECUTE (through at least Parallel Workload) and statement logging. https://buildkite.com/materialize/nightly/builds/16351

ggevay (approved):
LGTM, provided that Nightly passes. Thanks for solving this!

aljoscha:
thanks for the review!

PR #36454, opened 2026-05-07
agents: compress skill and CLAUDE.md descriptions

antiguru:
### Motivation

Skill descriptions in `.agents/skills/*/SKILL.md` and the project CLAUDE.md were verbose, costing tokens in every Claude Code session that loads the skill catalog.

### Description

Compress the `description:` frontmatter on all 12 `mz-*` skills and the prose in CLAUDE.md (AGENTS.md, via symlink) to caveman-terse form. Trigger keywords, technical terms, and file paths are preserved verbatim so skill matching is unaffected. Only fluff (articles, hedging, "this skill should be used when") is dropped. Skill body content is untouched.

### DIFF
--- a/.agents/skills/mz-adapter-guide/SKILL.md
+++ b/.agents/skills/mz-adapter-guide/SKILL.md
@@ -1,12 +1,11 @@
 ---
 name: mz-adapter-guide
 description: >
-  Correctness invariants and architectural guidance for the adapter layer,
-  coordinator, pgwire, peek paths, and timestamp oracle. Trigger when the user
-  works on or asks questions about these subsystems — including "how does the
-  coordinator work", "what are read holds", "explain the peek path", "how does
-  timestamp selection work", "why does this query block". Also trigger when
-  editing files in src/adapter/, src/pgwire/, or src/timestamp-oracle/.
+  Correctness invariants + architecture: adapter, coordinator, pgwire, peek
+  paths, timestamp oracle. Trigger: questions about these subsystems —
+  "how does coordinator work", "what are read holds", "explain peek path",
+  "how does timestamp selection work", "why does this query block". Also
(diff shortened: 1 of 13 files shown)


def- (commented):
I'm lacking a lot of LLM insights to judge if this is good or not. I've heard of LLMs using the "fluff" as space for thinking, but don't know how true that is nowadays.

mtabebe (commented):
I understand the rename/relocation but I question if we really want to do this caveman thing...

I just looked and this is what the token size of what we have:
```
     Skills · /skills

     Project
     └ mz-test: 190 tokens
     └ mz-benchmark: 186 tokens
     └ mz-parallel-workload: 136 tokens
     └ mz-adapter-guide: 116 tokens
     └ mz-run: 115 tokens
     └ mz-profile: 111 tokens
     └ mz-debug-ci: 108 tokens
     └ mz-pr-review: 103 tokens
     └ mz-query-tracing: 96 tokens
     └ mz-platform-checks: 96 tokens
     └ mz-commit: 93 tokens
     └ mz-limits-test: 92 tokens
     └ update-docs: 12 tokens
```

It looks like this saves about 1700 chars total. Apparently the front matter is what gets loaded into the context, so taking mz-test as an example, this saves about 220 characters, or ~55 tokens...

So I guess my question is: these skills in my claude are <0.5% of my context. Do I really care about compressing them to 0.3%?

mtabebe (approved):
That said, I am happy to approve ... I don't think the compaction is necessary but it isn't wrong

antiguru:
I agree, but one thing I noticed is that resuming sessions after their timeout means one needs to restore the whole context, so the .3% accumulates over time. (if you're like me who has too many parallel sessions :laughing:)

PR #36435, opened 2026-05-07
ci: Nightly/test fixes (2026-05-07)

def-:
Mostly follow-ups to #36214, except the pg-cdc flake which is old
https://github.com/MaterializeInc/materialize/pull/36340 caused Miri to time out
Based on this run: https://buildkite.com/materialize/nightly/builds/16340

Test run: https://buildkite.com/materialize/nightly/builds/16341 & https://buildkite.com/materialize/nightly/builds/16343 Edit: all green

PR #36455, opened 2026-05-07
cargo test: Fix non-deterministic Storage inputs

def-:
Fix from https://github.com/MaterializeInc/materialize/pull/36435 didn't work. Seems to be caused by https://github.com/MaterializeInc/materialize/pull/36214


bosconi (approved):
Looks good.

PR #28675, opened 2024-07-31
balancerd: use dycnfgs and launch darkly

maddyblue:
Add a new crate to sync launchdarkly to a configset. Hook it up to the balancer.

### Motivation

  * This PR adds a feature that has not yet been specified.

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [x] This PR includes the following [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note):
  - n/a

PR #36456, opened 2026-05-07
dyncfg-launchdarkly: skip bad entries instead of aborting sync

def-:
A single malformed Duration flag (e.g. a typo like "10 munites") in LaunchDarkly would abort the entire config sync pass via the `?` operator, preventing ALL other config updates from being applied. This includes unrelated configs like booleans and tracing filters.

Only balancerd is affected — environmentd uses a separate code path (SystemParameterFrontend) that converts all flag values to strings and doesn't call humantime::parse_duration during sync.

Handle per-entry errors gracefully by warning and skipping the bad entry, matching the behavior of dyncfg-file. Valid configs are still applied even when one entry is misconfigured.

Fixes a bug introduced in #28675.


SangJunBak (approved):
This change makes sense to me

PR #36457, opened 2026-05-07
COPY FROM: disable http client redirect

martykulma:
Disable redirects in HTTP client used by COPY FROM

- Add meaningful user error on redirect detection
- Add @def-'s mzcompose test
- Add note to COPY FROM doc that redirects are not followed

Fixes: [SS-121](https://linear.app/materializeinc/issue/SS-121)

### DIFF
--- a/test/copy/redirect_server.py
+++ b/test/copy/redirect_server.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+
+# Copyright Materialize, Inc. and contributors. All rights reserved.
+#
+# Use of this software is governed by the Business Source License
+# included in the LICENSE file at the root of this repository.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0.
+
+import socket
+import sys
+from http.server import BaseHTTPRequestHandler, HTTPServer
+
+CONTAINER_IP = socket.gethostbyname(socket.gethostname())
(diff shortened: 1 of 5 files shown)


test/copy/redirect_server.py
```diff
@@ -0,0 +1,67 @@
...
+
+        elif self.path == "/redirect-to-loopback":
+            self.send_response(302)
+            self.send_header("Location", "http://127.0.0.1:8080/data.csv")
+            self.end_headers()
+
+        elif self.path == "/redirect-to-metadata":
+            self.send_response(302)
+            self.send_header("Location", "http://169.254.169.254/latest/meta-data/")
+            self.end_headers()
```
ublubu, line 53: Are these not used?

ublubu (approved):
:+1: It looks like all the `response =` are followed by a `check_not_redirect` in `http_source.rs`.
And all the other `.send()`s are in unit tests.

I'm guessing the test case (successfully) failed before the fix was added. One comment inline about the implied test cases that don't seem to exist any more.

test/copy/redirect_server.py
martykulma: I copied this from @def-'s bug report and left these in as I wasn't sure if he wanted to use it for other things.

Since these all return the same status code, the `Location` header doesn't matter - MZ shouldn't look at it. So I didn't add additional tests as they don't add any additional coverage.

@def- - let me know if we need these extra paths, if not, I can remove them in a subsequent PR.

PR #36460, opened 2026-05-07
[MySQL] add additional test cases for mysql cdc SS-122

patrickwwbutler:
Adds a few more cases for good measure to mysql cdc testing, particularly a different arrangement of excluded columns and a way to overify that `binlog_row_metadata` is correctly configured for all tests.


martykulma (approved):
🦈

PR #36461, opened 2026-05-07
SS-112 Kafka Sink: add check at planning time for non-positive `TOPIC METADATA REFRESH INTERVAL`

patrickwwbutler:
A non positive duration causes the `tokio::time::interval` class to panic on creation. Panicking is bad and we should just reject the statement at planning time

### DIFF
--- a/src/sql/src/plan/statement/ddl.rs
+++ b/src/sql/src/plan/statement/ddl.rs
@@ -1228,7 +1228,7 @@ fn plan_kafka_source_connection(
     if topic_metadata_refresh_interval < Duration::from_secs(1) {
         // This is a librdkafka-enforced restriction that, if violated,
         // would result in a runtime error for the source.
-        sql_bail!("TOPIC METADATA REFRESH INTERVAL cannot be less than 1 second");
+        sql_bail!("TOPIC METADATA REFRESH INTERVAL must be at least 1 second");
     }
     let metadata_columns = include_metadata
         .into_iter()
@@ -3810,7 +3810,7 @@ fn kafka_sink_builder(
     } else if topic_metadata_refresh_interval < MIN_KAFKA_TOPIC_METADATA_REFRESH_INTERVAL {
         // We enforce a minimum of 1 second here to prevent excessive refreshes, and ensure that
         // tokio::time::interval receives a valid (positive) duration.
-        sql_bail!("TOPIC METADATA REFRESH INTERVAL must be at least one second");
+        sql_bail!("TOPIC METADATA REFRESH INTERVAL must be at least 1 second");
     }
 
     let assert_positive = |val: Option<i32>, name: &str| {
--- a/src/sql/src/plan/statement/ddl.rs
+++ b/src/sql/src/plan/statement/ddl.rs
@@ -1225,6 +1225,11 @@ fn plan_kafka_source_connection(
         // would result in a runtime error for the source.
         sql_bail!("TOPIC METADATA REFRESH INTERVAL cannot be greater than 1 hour");
     }
+    if topic_metadata_refresh_interval < Duration::from_secs(1) {
+        // This is a librdkafka-enforced restriction that, if violated,
+        // would result in a runtime error for the source.
+        sql_bail!("TOPIC METADATA REFRESH INTERVAL cannot be less than 1 second");
+    }
     let metadata_columns = include_metadata
         .into_iter()
         .flat_map(|item| match item {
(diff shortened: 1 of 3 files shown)


src/sql/src/plan/statement/ddl.rs
```diff
@@ -3785,6 +3785,10 @@ fn kafka_sink_builder(
         // This is a librdkafka-enforced restriction that, if violated,
         // would result in a runtime error for the source.
         sql_bail!("TOPIC METADATA REFRESH INTERVAL cannot be greater than 1 hour");
+    } else if topic_metadata_refresh_interval <= Duration::new(0, 0) {
```
antiguru: Consider excluding very small values. What'd happen if the user specified 1us?

ublubu (approved):
+2

src/sql/src/plan/statement/ddl.rs
patrickwwbutler: Hmm probably bad things. Any thoughts on a reasonable minimum? The default value is 30 seconds

antiguru: 1s? 10s? I can't tell :)

martykulma: `10s` is very aggressive for production environments, but we probably use this in testing for snappier results - so my vote would be `1s`. disclaimer: I haven't looked at the tests.

patrickwwbutler:
also added consts to replace the magic numbers so that @martykulma can't accuse me of hypocrisy

src/sql/src/plan/statement/ddl.rs
```diff
@@ -3781,10 +3784,14 @@ fn kafka_sink_builder(
...
 
-    if topic_metadata_refresh_interval > Duration::from_secs(60 * 60) {
+    if topic_metadata_refresh_interval > MAX_KAFKA_TOPIC_METADATA_REFRESH_INTERVAL {
         // This is a librdkafka-enforced restriction that, if violated,
         // would result in a runtime error for the source.
         sql_bail!("TOPIC METADATA REFRESH INTERVAL cannot be greater than 1 hour");
+    } else if topic_metadata_refresh_interval < MIN_KAFKA_TOPIC_METADATA_REFRESH_INTERVAL {
+        // We enforce a minimum of 1 second here to prevent excessive refreshes, and ensure that
+        // tokio::time::interval receives a valid (positive) duration.
+        sql_bail!("TOPIC METADATA REFRESH INTERVAL must be greater than one second");
```
def-:
```suggestion
        sql_bail!("TOPIC METADATA REFRESH INTERVAL must be at least one second");
```

test/testdrive/kafka-sink-errors.td
```diff
@@ -41,6 +41,17 @@ contains:column referenced in KEY does not exist: f2
...
+! CREATE SINK zero_interval_sink
+  IN CLUSTER ${arg.single-replica-cluster}
+  FROM v1
+  INTO KAFKA CONNECTION kafka_conn (
+    TOPIC 'testdrive-kafka-sink-zero-interval-${testdrive.seed}',
+    TOPIC METADATA REFRESH INTERVAL '0s'
+  )
+  FORMAT AVRO USING CONFLUENT SCHEMA REGISTRY CONNECTION csr_conn
+  ENVELOPE DEBEZIUM
+contains:TOPIC METADATA REFRESH INTERVAL must be greater than one second
```
def-:
```suggestion
contains:TOPIC METADATA REFRESH INTERVAL must be at least one second
```

def- (changes requested):
Kafka sources also don't check this at planning time, worth fixing too?

patrickwwbutler:
> Kafka sources also don't check this at planning time, worth fixing too?

Fixed the issue for sources as well, if you want to take another look! @def-

test/testdrive/kafka-sink-errors.td
```diff
@@ -41,6 +41,17 @@ contains:column referenced in KEY does not exist: f2
...
+! CREATE SINK zero_interval_sink
+  IN CLUSTER ${arg.single-replica-cluster}
+  FROM v1
+  INTO KAFKA CONNECTION kafka_conn (
+    TOPIC 'testdrive-kafka-sink-zero-interval-${testdrive.seed}',
+    TOPIC METADATA REFRESH INTERVAL '0s'
+  )
+  FORMAT AVRO USING CONFLUENT SCHEMA REGISTRY CONNECTION csr_conn
+  ENVELOPE DEBEZIUM
+contains:TOPIC METADATA REFRESH INTERVAL must be at least one second
```
def-: (others use 1 second)

def- (approved):
Thanks for adding tests, only a minor thing

PR #36462, opened 2026-05-07
persist: coalesce S3 chunks into a single Bytes per part

antiguru:
Follow-up to #36305 / #36451.
Those PRs left each hyper chunk pushed into `SegmentedBytes` as its own `Bytes`, so a returned blob contained hundreds of small segments per part.
Parquet `ChunkReader::get_bytes` walks the segment list per read, so decode CPU ended up dominated by `SegmentedBytes::advance` / `get_bytes` — over 85% of on-CPU time on a storage replica in production profiles, with `decode_and_mfp` cumulative ~471 s vs. `decode_kv` only ~3 s.

This change extends every chunk into a pre-sized `BytesMut` and freezes to a single `Bytes`, so each part contributes one segment.
The hyper pool buffer is still copied out of immediately, preserving the RSS fix from #36451.


DAlperin (approved):
lgtm, thanks!

PR #36464, opened 2026-05-08
Fdb emulator 2

jubrad:
Add FoundationDB as an alternative metadata store for the emulator image. Both PostgreSQL and FoundationDB are installed in the base image. The entrypoint reads MZ_METADATA_STORE (default: postgres) to select which backend to start. Set MZ_METADATA_STORE=foundationdb to use FDB.

in this case we'll publish a `-fdb` materialize emulator image for every image ex:
`materialize/materialized:26.25.0-fdb`

revival of https://github.com/MaterializeInc/materialize/pull/35087


src/materialized/ci-fdb/mzbuild.yml
```diff
@@ -0,0 +1,25 @@
...
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0.
+
+name: materialized-fdb
+description: Materialize is a fast, distributed SQL database built on streaming internals (FoundationDB backend).
+publish-as: materialized
+tag-suffix: -fdb
+pre-image:
+  - type: cargo-build
```
antiguru, line 15: Does this work?

antiguru (commented):
Makes sense, but I think you need a review from @def- here!

src/materialized/ci-fdb/mzbuild.yml
jubrad: Everything builds and the image seems to work

def- (commented):
I assume this still makes our docker images for the emulator significantly larger for an experiment. What is the goal? Edit: I see, so we use a -fdb version suffix, that's fine then. But it does make the build a bit slower
Based on https://buildkite.com/materialize/test/builds/122662 we seem to be still pushing materialized-fdb:
```
materialize/materialized-fdb-base:mzbuild-CEVVAAJFWG4OCU3BJ7F6PZALKAB477RU
materialize/materialized-fdb:mzbuild-3OCYPBBDFDNOF2H77HQGG4V7L7TEJ6HX
```

jubrad:
> I assume this still makes our docker images for the emulator significantly larger for an experiment. What is the goal? Edit: I see, so we use a -fdb version suffix, that's fine then. But it does make the build a bit slower

Why would it? we shouldn't be packaging FDB in our postgres based emulator so it shouldn't change the size. 

> What is the goal?

My suspision is that this will actually make the emulator more resource efficient, especially for larger local testing with a lot of sources or low tic intervals. It seems like it lowers CPU a decent amount on some local demos where I'm setting a very low tic interval. 

> Based on https://buildkite.com/materialize/test/builds/122662 we seem to be still pushing materialized-fdb

@def- wait... I don't see us pushing this to dockerhub? 
https://hub.docker.com/u/materialize?page=1&search=materialized

I think we're just using materialize-fdb as an internal name. 
```

$ docker manifest create --amend materialize/materialized:v26.24.0-dev.0--pr.g9a5b17bdaa0f69293815c6b6d2531b38bef6ba80-fdb materialize/materialized-fdb:mzbuild-3OCYPBBDFDNOF2H77HQGG4V7L7TEJ6HX materialize/materialized-fdb:mzbuild-X7VXA2N6HLSO5W5NBP5UZD45WEQ67L3T
--
Created manifest list docker.io/materialize/materialized:v26.24.0-dev.0--pr.g9a5b17bdaa0f69293815c6b6d2531b38bef6ba80-fdb
$ docker manifest push materialize/materialized:v26.24.0-dev.0--pr.g9a5b17bdaa0f69293815c6b6d2531b38bef6ba80-fdb
Pushed ref docker.io/materialize/materialized@sha256:dd79ad5d44adf6d5ce6d45039f4284f4a47da9e8583aa7225e70f4c85b43f820 with digest: sha256:dd79ad5d44adf6d5ce6d45039f4284f4a47da9e8583aa7225e70f4c85b43f820
Pushed ref docker.io/materialize/materialized@sha256:2f2e58db5de1984af3b89269cf7349c9c9309c1b7af7628bfd0924c0b29b6a50 with digest: sha256:2f2e58db5de1984af3b89269cf7349c9c9309c1b7af7628bfd0924c0b29b6a50
sha256:b9a8064ab878d42e299d65002e90be143295c2482c2a6c9d97e56041f6dfdc1f


```

misc/python/materialize/mzbuild.py
```diff
@@ -852,6 +859,20 @@ def docker_name(self, tag: str | None = None) -> str:
             name += f":{tag}"
         return name
 
+    def publish_docker_name(self, tag: str | None = None) -> str:
```
def-, line 862: I'm not super happy with this whole logic, and translating between materialized-fdb and materialized:...-fdb. Why don't we just build it as materialized-fdb directly instead of having this translation layer?

def- (approved):
Thanks, lgtm for some experiments. I think long term we should decide what to use for emulator and stick with it. When I ran my experiments I saw lower CPU usage with postgres compared to fdb fwiw. I'm mostly afraid that we stay in this state and have to push two versions out for months or even years.

PR #36466, opened 2026-05-08
cluster & restart test: Disable two flaky tests

def-:
Seen flaking in recent test runs. I'm not sure if [workflow_arrangement_sizes_hydration_gate](https://github.com/MaterializeInc/materialize/commit/faada045b3bc2aa190e1ea89a624437ffa041f58) is useful enough to fix, I couldn't find a good way.


antiguru (approved):
Thank you!

PR #36467, opened 2026-05-08
adapter: restore RTR error humanization

def-:
The staged sequencer refactor propagated RtrTimeout and RtrDropFailure as generic StorageError, losing the humanized display, detail text, and correct SQL state (QUERY_CANCELED / UNDEFINED_OBJECT). Convert them to the dedicated AdapterError variants before propagating.

lost in #27693

### DIFF
--- a/src/adapter/src/coord/sequencer/inner/explain_timestamp.rs
+++ b/src/adapter/src/coord/sequencer/inner/explain_timestamp.rs
@@ -16,7 +16,7 @@ use mz_controller_types::ClusterId;
 use mz_expr::CollectionPlan;
 use mz_ore::instrument;
 use mz_repr::explain::ExplainFormat;
-use mz_repr::{Datum, GlobalId, Row};
+use mz_repr::{Datum, Row};
 use mz_sql::plan::{self};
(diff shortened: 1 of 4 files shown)


def-:
Ready for review

src/adapter/src/coord/sequencer/inner/explain_timestamp.rs
```diff
@@ -181,11 +183,27 @@ impl Coordinator {
 
         match fut {
             Some(fut) => {
+                let catalog = Arc::clone(&self.catalog);
                 let span = Span::current();
                 Ok(StageResult::Handle(mz_ore::task::spawn(
                     || "explain timestamp real time recency",
                     async move {
-                        let real_time_recency_ts = fut.await?;
+                        let rtr_name = |id: &GlobalId| {
```
aljoscha: we have the same (or almost the same) code duplicated in three places, might make sense to factor out into a helper?

def-: Done!

aljoscha (approved):

PR #27593, opened 2024-06-12
compute: reconciliation of logging collections (preserving `SpecializedTraceHandle`)

teskje:
Upon reconnecting to a replica, the compute controller expects the logging collections to be uncompacted and readable from the minimum time. To ensure that we introduce the ability to "pad" traces in the `TraceManager` by filling up their compacted times with empty contents. This is implemented as a thin `TraceReader` wrapper that intercepts accesses to the trace's logical compaction frontier and otherwise passes requests through to the inner trace.

This is an alternative proposal to https://github.com/MaterializeInc/materialize/pull/27532 that does not remove the `SpecializedTraceHandle` type but works around it instead.

### Motivation

  * This PR fixes a previously unreported bug.

Compute reconciliation introduces a mismatch between the controller's expectation of the read frontiers of logging collections (`[0]`) and the actual read frontiers the respective arrangements have on the replica. This mismatch can let replicas crash when the controller tries to read the logging collections at a time before their physical read frontier.

### Tips for reviewer

There is one TODO comment left about how a padded `TraceHandle` should implement `TraceReader::map_batches`. The current implementation simply defers to the wrapped trace, but I'm not sure that is correct. Would callers of `map_batches` expect to see an empty batch containing the times from `padded_since` to `trace.get_compaction_frontier()`?

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] This PR includes the following [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note):
  -

PR #36468, opened 2026-05-08
compute: fix PaddedTrace returning logical instead of physical compaction

def-:
`get_physical_compaction` incorrectly delegated to `get_logical_compaction` on the inner trace, introduced in PR #27593.


antiguru (approved):
Good find!

PR #36471, opened 2026-05-08
adapter: cascade dependents when dropping replica-targeted MVs

antiguru:
### Motivation

Fixes database-issues#11288.

When a target replica is dropped, the targeted materialized view is implicitly dropped along with it.
The cascade to dependents of that MV was missing, so any view, materialized view, or index that depended on the implicitly-dropped MV was left referencing a non-existent catalog entry.
In debug builds the post-transaction consistency check tripped on `MissingUses` and panicked; in release builds subsequent reads against the orphaned dependent panicked or errored.

### Description

Fixed in two places:

* `cluster_replica_dependents` in `src/adapter/src/catalog/state.rs` now walks targeted MVs and includes their `item_dependents` in the returned set.
  This makes the plan-driven drop path produce a correct `DropObjectsPlan` and surface a `CascadeDroppedObject` notice for the dependents.
* `ObjectsToDrop::add_item` in `src/adapter/src/catalog/transact.rs` extends the implicit-MV addition with `item_dependents` of the targeted MV.
  The plan path already includes them via the change above; this is a backstop for internal callers (e.g. `drop_reconfiguration_replicas`) that build `Op::DropObjects` directly without going through the plan stage.

The bespoke `*_dependents` family in `state.rs` is fragile.
A fix-point worklist with per-object-kind direct-edge enumeration plus a single topo-sort would localize this knowledge and is tracked separately as a follow-up.

### Verification

Added a regression case to `test/sqllogictest/materialized_view_replica_targeted.slt` covering view, materialized-view, and index dependents of a replica-targeted MV when the target replica is dropped.

### DIFF
--- a/src/adapter/src/catalog/transact.rs
+++ b/src/adapter/src/catalog/transact.rs
@@ -3015,16 +3015,17 @@ impl ObjectsToDrop {
                 // through the plan stage.
                 let mut seen: BTreeSet<ObjectId> =
                     self.items.iter().copied().map(ObjectId::Item).collect();
-                for (_id, entry) in state.get_entries() {
+                for item_id in &cluster.bound_objects {
+                    let entry = state.get_entry(item_id);
                     if let CatalogItem::MaterializedView(mv) = entry.item()
                         && mv.target_replica == Some(replica_id)
-                        && !seen.contains(&ObjectId::Item(entry.id()))
+                        && !seen.contains(&ObjectId::Item(*item_id))
                     {
-                        tracing::warn!(
+                        tracing::info!(
                             "implicitly dropping materialized view {} because target replica was dropped",
                             entry.name().item,
                         );
-                        for dep in state.item_dependents(entry.id(), &mut seen) {
+                        for dep in state.item_dependents(*item_id, &mut seen) {
                             if let ObjectId::Item(dep_id) = dep {
                                 self.items.push(dep_id);
                             }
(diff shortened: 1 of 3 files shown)


src/adapter/src/catalog/transact.rs
```diff
@@ -3006,16 +3006,28 @@ impl ObjectsToDrop {
...
-                                "implicitly dropping materialized view {} because target replica was dropped",
-                                entry.name().item,
-                            );
-                            self.items.push(entry.id());
+                    if let CatalogItem::MaterializedView(mv) = entry.item()
+                        && mv.target_replica == Some(replica_id)
+                        && !seen.contains(&ObjectId::Item(entry.id()))
+                    {
+                        tracing::warn!(
+                            "implicitly dropping materialized view {} because target replica was dropped",
```
ggevay, line 3025: Why `warn`? This seems like a normal thing to happen.

src/adapter/src/catalog/transact.rs
```diff
@@ -3006,16 +3006,28 @@ impl ObjectsToDrop {
...
+                //
+                // Also cascade to anything depending on the implicitly-dropped
+                // MV so we don't leave dangling references in the catalog.
+                // Plan-driven drops already include these via
+                // `cluster_replica_dependents`; this branch handles internal
+                // callers that build `Op::DropObjects` directly without going
+                // through the plan stage.
+                let mut seen: BTreeSet<ObjectId> =
+                    self.items.iter().copied().map(ObjectId::Item).collect();
                 for (_id, entry) in state.get_entries() {
```
ggevay: Could this be just `cluster.bound_objects()`? (Like in `state.rs` above.)

ggevay (approved):
Thanks! LGTM, just minor comments.

src/adapter/src/catalog/transact.rs
antiguru: Moved to info.

antiguru:
Thanks for the review!

PR #36474, opened 2026-05-08
coord: remove O(n) table advancement loop from group_commit

aljoscha:
Remove the loop in `group_commit()` that iterated ALL catalog entries on every group commit to add empty advancement entries for every table. With ~20k+ tables, this was the dominant DDL bottleneck at ~23% of DDL time.

The txn-wal protocol makes explicit per-table advancement unnecessary: when any transaction commits to the txns shard, the logical upper of ALL registered data shards advances automatically, including those not involved in the transaction. The empty advancement entries were performing no useful work on the storage side.

Also removed the early-return optimization in PersistTableWriteWorker::append that skipped the txn-wal commit for empty updates. This ensures periodic group commits (with no actual data writes) still advance the txns shard upper, maintaining the property that table logical uppers advance even when no writes are happening.

Results at ~28k objects (optimized build):
- CREATE TABLE: 374ms → 131ms (65% faster)
- CREATE VIEW: ~96ms
- DROP TABLE: ~97ms
- DROP VIEW: ~86ms

### DIFF
--- a/src/adapter/src/coord/appends.rs
+++ b/src/adapter/src/coord/appends.rs
@@ -504,15 +504,6 @@ impl Coordinator {
             }
         }
 
-        // Add table advancements for all tables.
-        let table_advancement_start = Instant::now();
-        for table in self.catalog().entries().filter(|entry| entry.is_table()) {
-            appends.entry(table.id()).or_default();
-        }
-        self.metrics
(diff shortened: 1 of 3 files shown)


ggevay (approved):
Nice!

PR #36479, opened 2026-05-08
console: clean up connect modal MCP instructions CNS-63

jubrad:
## Summary
- Move MCP token generation into the MCP tab with a "Generate MCP token" button instead of showing it stacked below with the app password
- Restructure MCP instructions into clear steps: (1) get your token (generate or base64-encode existing), (2) connect your client
- Simplify client instructions to show `claude mcp add` command and link to docs for other clients (Claude Desktop, Cursor, Windsurf, etc.) instead of separate tabs for each
- Add `wrap` prop to CopyableBox for multiline command display
- Hide the "Create app password" section when on the MCP tab since token generation is now inline

## Test plan
- [ ] Open connect modal, switch to MCP Server tab
- [ ] Verify "Generate MCP token" button appears and generates a token inline
- [ ] Verify base64-encode option shows below with the printf command
- [ ] Verify `claude mcp add` command displays with line breaks
- [ ] Verify docs link points to `/docs/integrations/mcp-server/`
- [ ] Switch to External tools / Terminal tabs and verify "Create app password" section still shows
- [ ] Test in self-managed mode (no generate button, just base64 instructions)

### DIFF
--- a/console/src/components/McpConnectInstructions.tsx
+++ b/console/src/components/McpConnectInstructions.tsx
@@ -7,46 +7,42 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-import { BoxProps, Text, useTheme, VStack } from "@chakra-ui/react";
+import {
+  BoxProps,
+  Button,
+  Flex,
+  Spinner,
+  Text,
+  useTheme,
+  VStack,
+} from "@chakra-ui/react";
 import { useAtom } from "jotai";
 import React from "react";
 
 import { useAppConfig } from "~/config/useAppConfig";
+import docUrls from "~/mz-doc-urls.json";
 import { currentEnvironmentState } from "~/store/environments";
 import { MaterializeTheme } from "~/theme";
+import { obfuscateSecret } from "~/utils/format";
 
-import { CopyableBox, TabbedCodeBlock } from "./copyableComponents";
+import { CopyableBox, SecretCopyableBox } from "./copyableComponents";
+import TextLink from "./TextLink";
 
 interface McpConnectInstructionsProps extends BoxProps {
   userStr: string;
   /** Pre-computed Base64 token for MCP configuration (cloud only). */
   mcpBase64Token?: string;
+  /** Callback to generate a new MCP token (creates an app password). */
+  onGenerateToken?: () => void;
+  /** Whether token generation is in progress. */
+  isGeneratingToken?: boolean;
 }
-
-const mcpConfigJson = (
-  baseUrl: string,
-  endpoint: "agents" | "developer",
-  opts?: { includeType?: boolean },
-) =>
-  JSON.stringify(
-    {
-      mcpServers: {
-        [`materialize-${endpoint}`]: {
-          ...(opts?.includeType && { type: "http" }),
-          url: `${baseUrl}/api/mcp/${endpoint}`,
-          headers: {
-            Authorization: "Basic <base64-token>",
-          },
-        },
-      },
-    },
-    null,
-    2,
-  );
 
 const McpConnectInstructions = ({
   userStr,
   mcpBase64Token,
+  onGenerateToken,
+  isGeneratingToken,
   ...props
 }: McpConnectInstructionsProps) => {
   const { colors } = useTheme<MaterializeTheme>();
@@ -73,7 +69,7 @@ const McpConnectInstructions = ({
   const base64Command = `printf '${user}:<password>' | base64 -w0`;
 
   const endpointUrl = `${baseUrl}/api/mcp/${endpoint}`;
-  const claudeCodeCliCommand = `claude mcp add --transport http materialize-${endpoint} ${endpointUrl} --header "Authorization: Basic <base64-token>"`;
+  const claudeCodeCliCommand = `claude mcp add --transport http materialize-${endpoint} \\\n  ${endpointUrl} \\\n  --header "Authorization: Basic <mcp-token>"`;
 
   return (
     <VStack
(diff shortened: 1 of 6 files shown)


jubrad:
image
image

Alphadelta14 (approved):

console/src/components/McpConnectInstructions.tsx
```diff
@@ -89,82 +85,81 @@ const McpConnectInstructions = ({
...
-          {isCloud
-            ? "Create a new app password and copy the MCP Token, or encode an existing app password by running the following in your terminal:"
-            : "Use a role with login and password attributes. Run the following in your terminal:"}
-        </Text>
-        <CopyableBox variant="default" contents={base64Command} />
-        {isCloud && mcpBase64Token && (
-          <Text fontSize="sm" color={colors.foreground.secondary}>
-            Your MCP token is available above, under the app password.
-          </Text>
+      <VStack alignItems="stretch" spacing="3">
```
leedqin:
```suggestion
    <VStack
      alignItems="stretch"
      spacing="6"
      p="6"
      overflowY="auto"
      maxHeight="60vh"
      {...props}
    >
```

console/src/components/McpConnectInstructions.tsx
```diff
@@ -89,82 +85,81 @@ const McpConnectInstructions = ({
...
+        <Text textStyle="heading-xs">1. Get your MCP token</Text>
+
+        {isCloud && onGenerateToken && (
+          <VStack alignItems="stretch" spacing="2">
```
leedqin:
```suggestion
   <VStack alignItems="stretch" spacing="4">
```

console/src/components/McpConnectInstructions.tsx
```diff
@@ -89,82 +85,81 @@ const McpConnectInstructions = ({
...
+              Base64-encode your username and password:
+            </Text>
+          )}
+          <CopyableBox variant="default" contents={base64Command} />
+        </VStack>
       </VStack>
 
-      <VStack alignItems="stretch" spacing="2">
-        <Text textStyle="heading-xs">Step 2. Connect your client</Text>
+      <VStack alignItems="stretch" spacing="3">
```
leedqin:
```suggestion
      <VStack alignItems="stretch" spacing="4">
```

console/src/components/copyableComponents.tsx
```diff
@@ -127,9 +129,9 @@ export const CopyableBox: React.FC<CopyableBoxProps> = ({
         py={2}
         flex={1}
```
leedqin, line 131:
```suggestion
        <Box
        as="pre"
        fontFamily="mono"
        pl={4}
        pr={wrap ? 2 : 0}
        py={wrap ? 3 : 2}
        flex={1}
        lineHeight={wrap ? "tall" : "4"}
        whiteSpace={wrap ? "pre-wrap" : "nowrap"}
        overflow={wrap ? "auto" : "hidden"}
        overflowWrap={wrap ? "anywhere" : undefined}
        textOverflow={wrap ? undefined : "ellipsis"}
        minWidth={0}
        {...(maxHeight && {
          maxHeight,
          overflow: "auto",
        })}
      >
        {props.children ?? contents}
      </Box>
```

leedqin:
Added some stylistic suggestions to make the claude code instructions wrap a bit better
image

leedqin:
There is a little confusing behaviour that clicking on "Create MCP token" will also create the App Password but it will not show up until you are in other non-MCP tabs which I think is okay since it won't confuse the users if they are not in the MCP tab. I also ran into this when I tried adding the MCP token inline. I can revisit this design choice when we do Create App Password rehaul that already confuses some people

leedqin (changes requested):
Added some inline suggestions for styles but looks good otherwise

console/src/components/McpConnectInstructions.tsx
```diff
@@ -89,82 +85,81 @@ const McpConnectInstructions = ({
...
+            <Text fontSize="sm" color={colors.foreground.secondary}>
+              Or Base64-encode an existing app password:
+            </Text>
+          )}
+          {!isCloud && (
+            <Text fontSize="sm" color={colors.foreground.secondary}>
+              Base64-encode your username and password:
+            </Text>
+          )}
+          <CopyableBox variant="default" contents={base64Command} />
```
kay-kim, line 145: So ... 
if Cloud, we state  "Or Base64-encode an existing app password:"

double checking ... when using a **service** account/app password (and not a personal app password), don't we usually use the role associated with the app password? (at least, I think with other clients we do).

But since `base64Command` is `printf '${user}:<password>' | base64 -w0`;

But not sure if we do want people to use the personal app.

console/src/components/McpConnectInstructions.tsx
```diff
@@ -89,82 +85,81 @@ const McpConnectInstructions = ({
...
-                  <CopyableBox
-                    variant="default"
-                    contents={mcpConfigJson(baseUrl, endpoint)}
-                  />
-                </VStack>
-              ),
-            },
-          ]}
-        />
+        <CopyableBox variant="default" wrap contents={claudeCodeCliCommand} />
```
kay-kim, line 164: A huge wishlist (not sure if possible) but maybe include the redacted token here but allow people to copy unredacted?

console/src/components/McpConnectInstructions.tsx
```diff
@@ -89,82 +85,81 @@ const McpConnectInstructions = ({
...
+              </VStack>
+            ) : (
+              <Button
+                onClick={onGenerateToken}
+                variant="primary"
+                size="sm"
+                alignSelf="flex-start"
+              >
+                Generate MCP token
+              </Button>
```
kay-kim: Do people need to connect first to create the associated db role in MZ?

kay-kim: Oh ... I just remembered .. we generate personal app password ... in which case they don't need to ..
But I think this blurb might apply below if using existing one (like if someone generated a service account specifically for this). .. will pop it down.

console/src/components/McpConnectInstructions.tsx
kay-kim: Also, if they had separately just created a service account for this(and  not a personal account) and if they were following the docs, they would know they need to connect first to create the associated db role ... 
but, if they're in the console ... and just thought to create a service account  and try connecting via MCP ... would there be a problem? If so ... we might want to blurb for both the base64 encoding line and this case.

Something like (spitballing):

If using a service account, 
- Use the role name associated with the service account.
- The service account must be activated; i.e., you must have logged in with the service account at least once.

console/src/components/McpConnectInstructions.tsx
jubrad: The connect modal creates an account for the current user so I don't think any of this is relevant. The current user will have logged in. Service accounts don't really come into play here. I could change the button to "Create Personal MCP Token". I can make a note about service accounts though.

leedqin (approved):
LGTM!! Thank you for addressing the changes

PR #36480, opened 2026-05-08
sql: Add restrict_to_user_objects for MCP agent isolation

jubrad:
## Summary

- Adds a `restrict_to_user_objects` session variable that blocks access to system catalog objects, designed for MCP agent isolation
- Variable is read-only; can only be set as a role default by superusers via `ALTER ROLE ... SET`
- Blocks system tables/views across `mz_catalog`, `mz_internal`, `pg_catalog`, and `information_schema`
- Blocks SHOW commands (rewritten to `mz_show_*` system views)
- Blocks SQL-implemented functions that query system tables (`has_table_privilege`, `pg_get_viewdef`, etc.) by propagating resolved IDs from `sql_impl` function bodies
- Blocks unmaterializable functions that leak system metadata (`mz_version`, `mz_role_oid_memberships`, etc.) via explicit safety classification
- Hardened against bypass via `SET`, `RESET`, `DISCARD ALL`, `ALTER ROLE` (non-superuser), and case-variant quoted identifiers

Design doc: `doc/developer/design/20260508_restrict_to_user_objects.md`


doc/developer/design/20260508_restrict_to_user_objects.md
```diff
@@ -0,0 +1,108 @@
...
+```sql
+-- Superuser sets the restriction as a role default
+ALTER ROLE agent SET restrict_to_user_objects = true;
+
+-- Agent reconnects; restriction is active for the session
+-- All of these fail:
+SET restrict_to_user_objects = false;                   -- ERROR: parameter cannot be changed
+ALTER ROLE agent SET restrict_to_user_objects = false;   -- ERROR: requires superuser
+ALTER ROLE agent RESET restrict_to_user_objects;         -- ERROR: requires superuser
+DISCARD ALL;  -- Resets to role default (true), not the definition default (false)
```
mtabebe, line 42: Nit, but it would complete the story to show how you can change the variable

doc/developer/design/20260508_restrict_to_user_objects.md
```diff
@@ -0,0 +1,108 @@
...
+
+Fix: `StatementContext` carries a shared `Arc<Mutex<ResolvedIds>>` accumulator. When `sql_impl` or `sql_impl_table_func_inner` resolves a function body, the resulting IDs are merged into this accumulator. At the end of `plan()`, accumulated IDs are merged into the caller's `resolved_ids`, which flows to `check_plan` -> `check_usage` where the restriction catches them. This follows the same pattern as `ShowSelect::new_from_bare_query`, which already captures and propagates resolved IDs from re-parsed SQL.
+
+**Unmaterializable functions** (`mz_version()`, `mz_role_oid_memberships()`, `mz_environment_id()`, `pg_backend_pid()`, etc.) are evaluated at runtime directly from coordinator/session state. They never reference catalog items by ID, so resolved ID propagation cannot catch them.
```
mtabebe, line 76: I'm not sure I understand, are these allowed to be returned? I can't imagine it is harmful to know the version/roles/environment id, etc?

doc/developer/design/20260508_restrict_to_user_objects.md
```diff
@@ -0,0 +1,108 @@
...
+
+Fix: `UnmaterializableFunc` has an `is_safe_for_restricted_session()` method with an explicit match — no wildcard arm, so adding a new variant is a compile error until the developer classifies it. The check runs in `eval_unmaterializable_func()` during optimization.
+
+| Classification | Functions |
+|---|---|
+| **Allowed** (session identity/time) | `current_database`, `current_schema`, `current_schemas`, `current_timestamp`, `current_user`, `session_user`, `mz_now`, `mz_session_id`, `is_rbac_enabled`, `viewable_variables` |
+| **Blocked** (system metadata) | `mz_version`, `mz_version_num`, `version`, `mz_environment_id`, `mz_is_superuser`, `mz_role_oid_memberships`, `mz_uptime`, `pg_postmaster_start_time`, `pg_backend_pid` |
```
mtabebe: I guess they are blocked. But I don't understand why

jubrad: To me the focus on this is about data products and joining them, I think these are somewhat out of scope for what we want agents to be able to query.. they don't really represent a security issue though.

mtabebe: Yes, that is why I was confused why they were called out

doc/developer/design/20260508_restrict_to_user_objects.md
jubrad: Good point done!

doc/developer/design/20260508_restrict_to_user_objects.md
jubrad: Yeah, I think we should still not allow them, but perhaps worth noting it's not because they're explicitly unsafe

src/sql/src/rbac.rs
```diff
@@ -301,6 +309,29 @@ pub fn check_usage(
...
+    // This is used by MCP tool queries that should only access user-created data products.
+    if session.restrict_to_user_objects() {
+        for item_id in resolved_ids.items() {
+            if item_id.is_system() {
+                if let Some(item) = catalog.try_get_item(item_id) {
+                    // Only block data-bearing system objects, not functions or types
+                    // which are needed for query execution.
+                    match item.item_type() {
+                        // Allow functions and types - they're needed for query execution
+                        CatalogItemType::Func | CatalogItemType::Type => {}
```
mtabebe: Maybe it is a NIT ... but it feels weird for me to just see this here. Any way to define this?

Also, we should probably refer to the design doc for more context

src/sql/src/session/vars.rs
```diff
@@ -596,6 +603,17 @@ impl SessionVars {
             .ok_or_else(|| VarError::UnknownParameter(name.to_string()))
     }
 
+    /// Returns true if the variable can be set as a role default even if it's
+    /// otherwise read-only from direct SET commands.
+    ///
+    /// SECURITY: Any variable listed here must also have a corresponding
```
mtabebe, line 609: Is there any way to add a test that ensures these are in sync?

test/sqllogictest/rbac_mcp_agent.slt
```diff
@@ -0,0 +1,664 @@
...
+DETAIL: Access to system catalog objects is restricted for this role. Contact your administrator if you need access.
+
+# =============================================================================
+# Test: SHOW commands are NOT YET blocked by restrict_to_user_objects
+# SHOW commands are rewritten to SELECT on mz_internal.mz_show_* views during
+# planning, but their resolved IDs go through ShowSelect's own resolution path
+# which does not flow through the restrict check. This is a known limitation.
+# =============================================================================
+
+# SHOW DATABASES still works (known limitation — SHOW uses separate resolution)
```
mtabebe, line 575: Thanks for this test, it is good to have for demonstrating we fix it.

mtabebe (commented):
LGTM in general!

mtabebe (approved):
Justin's commits LGTM (I reviewed) he needs to review mine now. Please don't merge until Justin reviews.

jubrad (commented):
Approval of @mtabebe's changes!

There is a known issue that remains with SHOW commands still working. We're opting to get this in and fix that as a follow up.

PR #36486, opened 2026-05-10
cluster test: Reenable test-compute-controller-metrics

def-:
Follow-up to https://github.com/MaterializeInc/materialize/pull/36473


antiguru (approved):
Thank you! Missed your comment in my PR.

PR #36487, opened 2026-05-10
adapter: fix fast_path_limit doc to match strict-less-than check and offset

def-:
The eligibility check is `LIMIT + OFFSET < persist_fast_path_limit`, but the doc claimed `LIMIT <= this`. Document the actual bound so consumers (e.g. the console) don't suggest LIMIT values that fail to fast-path.


ggevay (approved):
Thank you!

PR #36488, opened 2026-05-10
testdrive: escape quotes and backslashes in rewrite_result

def-:
Values containing whitespace combined with `"` or `\` were wrapped in quotes without escaping, causing the parser to either fail with "unterminated quote" or silently corrupt expected values (e.g. `\n` inside a path becomes a newline). Use Debug formatting like TestdriveRow::fmt does, which escapes exactly what split_line expects.


bosconi (approved):
Looks good.

PR #36489, opened 2026-05-10
testdrive: fix --rewrite-results dropping bytes between commands

def-:
`parse_sql` and `parse_explain_sql` captured `expected_end` from `LineReader::raw_pos`, but `peek()` had already advanced `raw_pos` past the blank separator and the next command's query line. Rewrites of any non-last command silently dropped those bytes. Track a separate `consumed_raw_pos` that only advances when a line is actually returned to an external caller.


bosconi (approved):
Looks good.

PR #36498, opened 2026-05-11
mzbuild: Fix parsing --arch parameters, bump bootstrap version for xcompile

def-:
Noticed with `bin/mzcompose --arch x86_64 --find testdrive run default`

Makes our target-xcompile pick up the latest https://github.com/MaterializeInc/homebrew-crosstools/pull/18


DAlperin (approved):
lgtm thanks!

PR #36499, opened 2026-05-11
arrow-util: canonicalize ranges decoded from Parquet

DAlperin:
The reader for range columns called `push_range_with`, which writes the range to the row verbatim. For discrete range types (int4/int8/date), Parquet files authored by external engines may encode ranges with non-canonical bounds — e.g. `[1,10]` for `int4range`, which MZ stores internally as `[1,11)`. Without canonicalization those rows do not compare or hash equal to logically-identical values constructed inside MZ, so `COPY FROM PARQUET` rows mismatch their pure-SQL counterparts.

Switch the decode path to `push_range`, which canonicalizes the range before packing. Add a unit test that decodes a hand-built non-canonical arrow StructArray and a testdrive regression test that round-trips a DuckDB-authored Parquet file through `COPY FROM ... (FORMAT PARQUET)`.

Fixes DB-55.

https://claude.ai/code/session_01KXioUsT2r5o2tRwrsHm4iR

### DIFF
--- a/test/iceberg/mzcompose.py
+++ b/test/iceberg/mzcompose.py
@@ -275,3 +275,19 @@ def workflow_large_upsert_batch(c: Composition) -> None:
         "--var=aws-endpoint=minio:9000",
         "large-upsert-batch.td",
     )
+
+
+def workflow_range_noncanonical(c: Composition) -> None:
+    """Regression test for database-issues#11330: COPY FROM PARQUET must
+    canonicalize range values reconstructed from external Parquet, otherwise
+    non-canonical encodings written by other engines land verbatim in MZ rows
+    and break equality against values constructed inside MZ. DuckDB authors the
+    Parquet file directly in minio so the bytes are not something MZ's sink
+    produced."""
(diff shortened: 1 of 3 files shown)


test/iceberg/mzcompose.py
```diff
@@ -275,3 +275,19 @@ def workflow_large_upsert_batch(c: Composition) -> None:
...
+    and break equality against values constructed inside MZ. DuckDB authors the
+    Parquet file directly in minio so the bytes are not something MZ's sink
+    produced."""
+    key = _setup(c)
+
+    c.run_testdrive_files(
+        f"--var=s3-access-key={key}",
+        "--var=aws-endpoint=minio:9000",
+        "range-noncanonical.td",
+    )
```
def-, line 293: Many workflows in test/iceberg do just this, we could put them all in a workflow_cdc instead that iterates over the files? But more of a test cleanup, not that important for this PR itself.

def- (approved):
Thanks for adding a test, lgtm!

PR #36500, opened 2026-05-11
Fix row to recordbatch conversion errors

DAlperin:
Follow up to some of the complicated type work. 

Three runtime errors surface when a parallel-workload run drives an Iceberg sink with a wider mix of column types:

1. "Datum Int16 does not match builder Int32Builder" — Iceberg has no smallint, so the writer context arrow schema (derived from the iceberg schema) uses Int32 while Materialize rows still carry Datum::Int16. Add a lossless Int16 -> Int32 promotion in ArrowColumn::append_datum, mirroring the existing UInt16 -> Int32 case.

2. "Field 'value' missing extension metadata" — Materialize names map fields entries/keys/values, but iceberg-rust's arrow conversion names them key_value/key/value. merge_field_metadata_recursive matched by name and silently dropped the value field's extension metadata, so ArrowBuilder later failed when constructing the inner builder. Match the map entries struct positionally instead.

3. "Failed to create EqualityDeleteWriterConfig: field_id N not found" — the planner accepted Range types as Iceberg equality delete keys, but ranges lower into Iceberg structs and iceberg-rust's RecordBatchProjector skips nested fields, so the equality field id is unreachable at runtime. Drop Range from the allow-list so the failure is caught at sink creation instead.

Fixes SS-144, SS-143, SS-142

### DIFF
--- a/src/storage/src/sink/iceberg.rs
+++ b/src/storage/src/sink/iceberg.rs
@@ -2069,6 +2069,71 @@ mod tests {
         assert_eq!(equality_ids, vec![expected_id]);
         assert_ne!(expected_id, 2);
     }
+
+    /// Regression test: iceberg-rust names map fields `key_value`/`key`/`value`
+    /// while Materialize uses `entries`/`keys`/`values`. The schema merge must
+    /// still copy the value field's extension metadata across so ArrowBuilder
+    /// can build the inner builder.
+    #[mz_ore::test]
+    #[allow(clippy::disallowed_types)]
+    fn merge_map_entries_preserves_value_extension_metadata() {
+        use std::collections::HashMap;
+
+        let mz_value_metadata = HashMap::from([(
+            ARROW_EXTENSION_NAME_KEY.to_string(),
+            "materialize.v1.string".to_string(),
+        )]);
+        let mz_entries = Field::new(
+            "entries",
+            DataType::Struct(
+                vec![
+                    Field::new("keys", DataType::Utf8, false),
+                    Field::new("values", DataType::Utf8, true).with_metadata(mz_value_metadata),
+                ]
+                .into(),
+            ),
+            false,
+        );
+        let mz_map = Field::new("m", DataType::Map(Arc::new(mz_entries), false), true)
+            .with_metadata(HashMap::from([(
+                ARROW_EXTENSION_NAME_KEY.to_string(),
+                "materialize.v1.map".to_string(),
+            )]));
+
+        let iceberg_entries = Field::new(
+            "key_value",
+            DataType::Struct(
+                vec![
+                    Field::new("key", DataType::Utf8, false),
+                    Field::new("value", DataType::Utf8, true),
+                ]
+                .into(),
+            ),
+            false,
+        );
+        let iceberg_map = Field::new("m", DataType::Map(Arc::new(iceberg_entries), false), true);
+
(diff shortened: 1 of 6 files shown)


def- (commented):
Thanks for adding tests. I will rebase my iceberg sink test in parallel-workload on top of this and https://github.com/MaterializeInc/materialize/pull/36499 and report if anything else falls over. Edit: Just noticed you made it draft again, so I'll wait it's un-drafted before doing that.

def-:
New failure in https://buildkite.com/materialize/nightly/builds/16384:
```
Sinks in stalled state: icesink-18: iceberg: failed to convert row to recordbatch: Failed to add insert row to builder: cannot represent decimal value 494699.19575454644 in column with scale 10
```

src/storage/src/sink/iceberg.rs
```diff
@@ -665,6 +669,73 @@ fn merge_field_metadata_recursive(
...
+/// Future cleanup: we could instead align Materialize's arrow map field names
+/// with the Parquet/Iceberg convention (`key_value`/`key`/`value`) in
+/// `mz_arrow_util::builder::scalar_to_arrow_datatype_impl` and drop this
+/// positional helper. That would also affect `COPY TO S3 ... FORMAT = 'parquet'`
+/// output schemas, so we'd need to confirm no downstream consumers depend on
+/// the current `entries`/`keys`/`values` names before flipping.
+fn merge_map_entries_metadata(
+    iceberg_entries: &Field,
+    mz_entries: Option<&Field>,
+) -> anyhow::Result<Field> {
```
ublubu, line 699: It looks like this function does nothing if `mz_entries` is `None`. Might as well do that check at the call site.

```suggestion
fn merge_map_entries_metadata(
    iceberg_entries: &Field,
    mz_entries:&Field,
) -> anyhow::Result<Field> {
```

test/iceberg/catalog.td
```diff
@@ -270,3 +270,63 @@ SELECT id, r.lower, r.upper, r.lower_inclusive, r.upper_inclusive, r.empty FROM
...
+$ duckdb-query name=iceberg
+SELECT id, val FROM iceberg_scan('s3://test-bucket/default_namespace/smallint_table') ORDER BY id
+1 -32768
+2 0
+3 32767
+
+# Test map columns
+# Materialize's arrow schema for maps uses field names entries/keys/values,
+# while iceberg-rust uses key_value/key/value. The metadata merge must align
+# them positionally so the value field's extension metadata survives.
```
ublubu, line 304: How does this test check the extension metadata?

ublubu (commented):
Looks good. :+1:

This should have been three separate PRs. (Just a note. No action required this time.)

Anyway, a couple comments inline.

src/storage/src/sink/iceberg.rs
DAlperin: Good point, changing

test/iceberg/catalog.td
DAlperin: It doesn't directly, we don't currently have a way in testdrive to inspect the raw parquet. This is a regression test to make sure it doesn't stall the sink.

ublubu: F

Alright, it'll have to do, then.

ublubu (approved):

PR #36502, opened 2026-05-11
cluster-spec-sheet/cloud-canary tests: Make sure BUILDKITE_COMMIT is set

def-:
Noticed in https://materializeinc.slack.com/archives/CTESPM7FU/p1778478849012359


antiguru (approved):
Thank you

PR #36504, opened 2026-05-11
timely-util: revert ConsolidatingColumnBuilder staging cap to 2*default_capacity

antiguru:
### Motivation

PR #36340 introduced `ConsolidatingColumnBuilder` with a fixed 16k-item staging buffer.
DD's `ConsolidatingContainerBuilder` uses `2 * default_capacity::<(D, T, R)>()`, which scales with row size and lands around 400–700 items for typical Mz tuples.

### Description

Switch the staging buffer cap from the hardcoded 16k to `2 * default_capacity::<(D, T, R)>()`.
Drain grain is `cap / 2`.
A benchmark sweep over 128–32768 items on `(u64, u64, i64)` shows the 512–4096 range sits within ~3% of the best (~190 ms total across four workloads), while 32k regresses ~6% on `single-key`.
The previous 16k default was mid-pack; the new type-dependent default lands inside the best band and matches the upstream formula.

### DIFF
--- a/src/timely-util/src/columnar/consolidate.rs
+++ b/src/timely-util/src/columnar/consolidate.rs
@@ -43,12 +43,18 @@ use timely::container::{ContainerBuilder, PushInto};
 
 use crate::columnar::Column;
 
-/// Items per staging buffer. Small enough that O(n log n) sort stays cheap, large enough to
-/// amortize the sort + drain overhead across many pushes.
-const STAGING_CAP_ITEMS: usize = 16 * 1024;
-/// Drain a multiple of this on consolidate. Remainder stays in staging so cross-batch keys
-/// keep consolidating on the next sort.
-const DRAIN_GRAIN: usize = STAGING_CAP_ITEMS / 2;
+/// Per-buffer byte budget for the staging cap. Matches the 8 KiB basis DD's
+/// `ConsolidatingContainerBuilder` uses via `timely::container::buffer::default_capacity`.
+const STAGING_BUFFER_BYTES: usize = 8 * 1024;
+
+/// Default items per staging buffer: `2 * STAGING_BUFFER_BYTES / size_of::<(D, T, R)>()`,
+/// matching DD's `ConsolidatingContainerBuilder`. Small enough that O(n log n) sort stays
+/// cheap, large enough to amortize the sort + drain overhead across many pushes.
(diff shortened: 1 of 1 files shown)


def- (approved):
Thanks!

def-:
QA LLM review found something (after merging, sorry, was out of tokens before). I'm not sure if it's actually as serious as it claims, and we can probably clean that up when we wire things up further?

## Finding 1: MEDIUM - small type-dependent caps break cross-batch consolidation for valid large `Columnar` rows

The new `default_staging_cap`/`staging_cap / 2` pairing in `src/timely-util/src/columnar/consolidate.rs:53-56` and `:213-216` makes the remainder-preserving staging invariant size-dependent. Once `size_of::<(D, T, R)>()` is large enough that `default_staging_cap::<D, T, R>()` becomes `2` or `3`, the drain grain becomes `1`, so every threshold-triggered `consolidate_and_drain` drains *all* consolidated rows out of `staging` instead of leaving a remainder behind for the next refill.

That violates the behavior documented a few lines below in `cross_batch_consolidation` (`src/timely-util/src/columnar/consolidate.rs:327-337`): "The drain-multiple-of-grain trick keeps the in-progress consolidated row in staging so subsequent pushes merge into it." With this commit, that is no longer true for perfectly valid `Columnar` types like `[u8; 5000]`, where `(D, T, R)` is large enough to compute a cap of `3`.

I reproduced the regression against the checked-out crate with:

```rust
let mut builder: ConsolidatingColumnBuilder<[u8; 5000], u64, i64> = Default::default();
for _ in 0..10 {
    builder.push_into(([7u8; 5000], 0u64, 1i64));
}
```

Draining that builder yields `rows=4 diffs=[3, 3, 3, 1]`, not one consolidated row with diff `10`.

This is not just "less optimal than before": the builder is public and explicitly documented as consolidating updates across staging refills, but after this change that guarantee silently stops holding for large fixed-size row types. There are no in-tree production callers today, so the current blast radius is limited, but this is exactly the kind of latent generic bug that will surface as soon as someone uses `ConsolidatingColumnBuilder` with a larger `Columnar` tuple shape.

antiguru:
Thanks for looking into this. I hope we never ship rows that have an inline size of 5016 bytes :) I think the behavior is acceptable.

PR #36507, opened 2026-05-11
compute: scaffold CollectionEdge for columnar dataflow edges

antiguru:
### Motivation

Today's row-based dataflow edges in compute force per-record consumer handling.
Switching producers to columnar first imposes an unpack tax on every consumer that is not yet columnar-aware; this PR begins a consumer-first migration that avoids the tax by making every consumer ready to absorb columnar batches before any producer emits them.

### Description

Adds `CollectionEdge<'scope, T>` with `Vec` and `Columnar` arms in `src/compute/src/render/columnar.rs`, plus a `columnar_negate` stub.
Scaffolding only: no producer emits the columnar variant, `CollectionBundle` is unchanged, and columnar arms carry `todo!()` bodies that land alongside the producer switch.
The design rule encoded in the API is that a columnar-to-row decode at a consumer's input is only acceptable when the consumer would have decoded `Row` to `Datum` anyway; passthrough consumers must round-trip columnar without decoding.

### DIFF
--- a/src/compute/src/render.rs
+++ b/src/compute/src/render.rs
@@ -169,6 +169,7 @@ use crate::render::errors::DataflowErrorSer;
 use crate::typedefs::{ErrBatcher, ErrBuilder, ErrSpine, KeyBatcher, MzTimestamp};
 use mz_row_spine::{DatumSeq, RowRowBatcher, RowRowBuilder};
 
+mod columnar;
 pub mod context;
 pub(crate) mod errors;
 mod flat_map;
--- a/src/compute/src/render.rs
+++ b/src/compute/src/render.rs
@@ -169,7 +169,7 @@ use crate::render::errors::DataflowErrorSer;
 use crate::typedefs::{ErrBatcher, ErrBuilder, ErrSpine, KeyBatcher, MzTimestamp};
 use mz_row_spine::{DatumSeq, RowRowBatcher, RowRowBuilder};
 
-mod columnar;
+pub(crate) mod columnar;
 pub mod context;
 pub(crate) mod errors;
 mod flat_map;
@@ -937,6 +937,7 @@ impl<'scope> Context<'scope, Product<mz_repr::Timestamp, PointStamp<u64>>> {
                 // We need to ensure that the raw collection exists, but do not have enough information
                 // here to cause that to happen.
                 let (oks, mut err) = bundle.collection.clone().unwrap();
+                let oks = oks.expect_vec();
                 self.insert_id(Id::Local(id), bundle);
                 let (oks_v, err_v) = variables.remove(&Id::Local(id)).unwrap();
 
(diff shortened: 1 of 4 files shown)


src/compute/src/render.rs
```diff
@@ -1316,8 +1320,11 @@ impl<'scope, T: RenderTimestamp + MaybeBucketByTime> Context<'scope, T> {
             }
             Negate { input } => {
                 let input = expect_input(input);
-                let (oks, errs) = input.as_specific_collection(None, &self.config_set);
-                CollectionBundle::from_collections(oks.negate(), errs)
+                let (oks, errs) = input
+                    .collection
+                    .clone()
+                    .expect("Negate input must be an unarranged collection");
+                CollectionBundle::from_edge(oks.negate(), errs)
```
frankmcsherry, line 1327: Should this change still use `as_specific_collections` instead?

src/compute/src/render.rs
```diff
@@ -1334,8 +1341,10 @@ impl<'scope, T: RenderTimestamp + MaybeBucketByTime> Context<'scope, T> {
                 let mut oks = Vec::new();
                 let mut errs = Vec::new();
                 for (input, strategy) in inputs.into_iter().zip_eq(temporal_bucketing_strategies) {
-                    let (os, es) =
-                        expect_input(input).as_specific_collection(None, &self.config_set);
+                    let (os, es) = expect_input(input)
+                        .collection
+                        .clone()
+                        .expect("Union input must be an unarranged collection");
```
frankmcsherry, line 1347: Same as above.

src/compute/src/render/columnar.rs
```diff
@@ -0,0 +1,556 @@
...
+        match self {
+            CollectionEdge::Vec(c) => CollectionEdge::Vec(CollectionExt::consolidate_named::<
+                KeyBatcher<_, _, _>,
+            >(c, name)),
+            CollectionEdge::Columnar(c) => {
+                // TODO: Consolidate natively over columns. The pieces exist
+                // (`columnar_exchange`, the columnar merge batchers), which
+                // would avoid the row round-trip below.
+                let c = columnar_to_vec(c);
+                let c = CollectionExt::consolidate_named::<KeyBatcher<_, _, _>>(c, name);
```
frankmcsherry, line 262: Worth exploring whether `consolidate_named`'s opinion about `Vec` and whether we should generalize this as a next step.

src/compute/src/render/columnar.rs
```diff
@@ -0,0 +1,556 @@
...
+            "ColumnarNegate",
+            |_cap, _info| {
+                move |input, output| {
+                    input.for_each(|time, data| {
+                        let mut session = output.session_with_builder(&time);
+                        for (v, t, d) in data.borrow().into_index_iter() {
+                            let d = -Diff::into_owned(d);
+                            session.give((v, t, &d));
+                        }
+                    });
```
frankmcsherry, line 298: Let's look into only working with the `Diff` container, leaving the other containers intact. We should be able to only negate the elements of that container and churn much less.

frankmcsherry (approved):
Looks good to me. Currently non-impacting, but a good starting point for migrations.

PR #36508, opened 2026-05-11
Orchestratord OIDC testing and dyncfg-file Option<String> support

alex-hunt-materialize:
- Add Option<String> support to dyncfg-file, so that we can set the oidc_issuer in a system param configmap.
- Add end to end OIDC testing through orchestratord nightly mzcompose tests using Ory Hydra.

### Motivation

Resolves https://linear.app/materializeinc/issue/DEP-39/orchestratord-happy-path-test

### Verification

Run the tests.


def- (approved):
Thanks, lgtm!

PR #36509, opened 2026-05-11
[SSO] Fix OIDC group sync privilege escalation via case folding

mtabebe:
Problem:
JWT group names were lowercased before catalog lookup. When two roles differ only by case (e.g. `admin` and `"Admin"`), the lowercase map built by `roles_by_lowercase_name` silently picked the wrong role due to BTreeMap last-writer-wins semantics.

Solution:
- remove `to_lowercase()` from group extraction, replace the lowercase-map lookup in `sync_jwt_groups` with exact `try_get_role_by_name`,
- delete the now-unused `roles_by_lowercase_name` and `try_get_role_by_name_case_insensitive` methods

Testing:
- updated three oidc.rs unit tests
- added mzcompose workflow `oidc_group_sync_case_sensitive` with an OIDC mock server to verify "Admin" cannot escalate to the `admin` role

Fixes: SQL-276


src/adapter/src/coord/group_sync.rs
```diff
@@ -162,13 +162,12 @@ impl Coordinator {
...
-        let role_map = self.catalog().roles_by_lowercase_name();
-
-        // Resolve group names to role IDs (case-insensitive).
+        // Resolve group names to role IDs (exact, case-sensitive match).
         let mut target_role_ids = BTreeSet::new();
         for group in groups {
             // Filter out reserved role names (mz_/pg_ prefixes, PUBLIC).
-            if catalog::is_reserved_role_name(group) {
+            // Check case-insensitively so "MZ_SYSTEM" is also blocked.
+            if catalog::is_reserved_role_name(&group.to_lowercase()) {
```
SangJunBak, line 170: I wonder if we need to check this case? Thinking the "No matching Materialize Role" warning is fine

src/authenticator/src/oidc.rs
```diff
@@ -280,15 +281,15 @@ impl OidcClaims {
             }
         };
 
-        let normalized: Vec<String> = raw_groups
+        let groups: Vec<String> = raw_groups
             .into_iter()
-            .map(|g| g.trim().to_lowercase())
+            .map(|g| g.trim().to_string())
```
SangJunBak: Should we still trim the groups? Given we're not normalizing anymore. E.g. there could exist a role "  wee wwoo "

src/adapter/src/coord/group_sync.rs
mtabebe: I think we should still warn as early as possible? Yes there is a false positive situation, but I feel like it is better to warn in these cases

src/authenticator/src/oidc.rs
mtabebe: Good catch

src/adapter/src/coord/group_sync.rs
SangJunBak: Fair enough!

test/http-auth/mzcompose.py
```diff
@@ -7,13 +7,112 @@
...
 import requests
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
 
 from materialize import MZ_ROOT
 from materialize.mzcompose.composition import Composition
+from materialize.mzcompose.service import Service
 from materialize.mzcompose.services.materialized import Materialized
 
+OIDC_PORT = 8443
```
SangJunBak: Couple things I'm wondering:
- What are the benefits of doing this in a bespoke mzcompose test vs. reusing oidc-mock in src/environmentd/tests/auth.rs
- If we need to do this via mzcompose, similar to frontegg-mock, I wonder if it makes sense to make an mzbuild wrapper around oidc-mock and just use that?

mtabebe: Honestly, I just took Dennis' test ... I'll move it over!

SangJunBak (approved):
LGTM!

PR #36514, opened 2026-05-11
Container builders for the persist source

antiguru:
### Motivation

The persist source hardcodes its row and error outputs to `Vec<(_, Timestamp, Diff)>` containers, which forces every downstream operator to consume that exact shape and prevents callers from selecting a more suitable container builder (e.g. for columnar batches).

### Description

Generalize `persist_source` over caller-supplied container builders `DCB`/`ECB` (each `ContainerBuilder + PushInto<(_, Timestamp, Diff)>`) and switch the ok/err demux from `ok_err` to a `unary_fallible` operator parameterized over them. All current callers pass `CapacityContainerBuilder<_>` to preserve existing behavior.

Supporting changes propagate the same generalization through `txn-wal::operator::txns_progress` and `txns_progress_frontiers` (now over `Stream<_, P: Container>`), relax `builder_async`'s output/input handles to accept any `ContainerBuilder`/`Container`, and add a `NoopContainerBuilder` in `mz-timely-util::containers` for the pass-through case.

See the commit message for the full breakdown.

### DIFF
--- a/src/compute/src/render.rs
+++ b/src/compute/src/render.rs
@@ -268,22 +268,25 @@ pub fn build_compute_dataflow(
 
                     // Note: For correctness, we require that sources only emit times advanced by
                     // `dataflow.as_of`. `persist_source` is documented to provide this guarantee.
-                    let (mut ok_stream, err_stream, token) =
-                        persist_source::persist_source::<DataflowErrorSer>(
-                            inner,
-                            *source_id,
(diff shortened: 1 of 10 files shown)


DAlperin (approved):
LGTM! Thanks

antiguru:
Thanks for the review!

PR #36515, opened 2026-05-11
orchestrator-kubernetes: make availability-zone node label configurable

jubrad:
## Problem

Two related issues with the Kubernetes orchestrator's AZ scheduling:

1. The AZ node affinity label was hardcoded as `materialize.cloud/availability-zone`, a proprietary label that only works in Materialize's managed cloud.

## Solution

**Configurable AZ label** — Replace the hardcoded label with a configurable field defaulting to the standard `topology.kubernetes.io/zone`:

- `mz-orchestrator-kubernetes`: new `availability_zone_label: String` on `KubernetesOrchestratorConfig`
- `environmentd`: new `--orchestrator-kubernetes-az-label` flag (env: `ORCHESTRATOR_KUBERNETES_AZ_LABEL`), default `topology.kubernetes.io/zone`
- `orchestratord`: new `--orchestrator-kubernetes-az-label` flag; forwarded to `environmentd`
- Helm chart: new `availabilityZoneLabel` value

Supported values: `topology.kubernetes.io/zone` (default), `topology.k8s.aws/zone-id`, `materialize.cloud/availability-zone` (previous behavior), or any custom label.

### DIFF
--- a/misc/helm-charts/operator/templates/deployment.yaml
+++ b/misc/helm-charts/operator/templates/deployment.yaml
@@ -239,6 +239,8 @@ spec:
         {{- end }}
         {{- if .Values.availabilityZoneLabel }}
         - "--orchestrator-kubernetes-az-label={{ .Values.availabilityZoneLabel }}"
+        {{- else if eq .Values.operator.cloudProvider.type "aws" }}
+        - "--orchestrator-kubernetes-az-label=topology.k8s.aws/zone-id"
         {{- end }}
 
         {{- if .Values.operator.additionalMaterializeCRDColumns }}
--- a/misc/helm-charts/operator/templates/deployment.yaml
+++ b/misc/helm-charts/operator/templates/deployment.yaml
@@ -237,6 +237,9 @@ spec:
         {{- if .Values.schedulerName }}
         - "--scheduler-name={{ .Values.schedulerName }}"
         {{- end }}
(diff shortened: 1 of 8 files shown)


misc/helm-charts/operator/templates/deployment.yaml
```diff
@@ -237,6 +237,11 @@ spec:
         {{- if .Values.schedulerName }}
         - "--scheduler-name={{ .Values.schedulerName }}"
         {{- end }}
+        {{- if .Values.availabilityZoneLabel }}
+        - "--orchestrator-kubernetes-az-label={{ .Values.availabilityZoneLabel }}"
+        {{- else if eq .Values.operator.cloudProvider.type "aws" }}
+        - "--orchestrator-kubernetes-az-label=topology.k8s.aws/zone-id"
```
Alphadelta14, line 243: this behavior [upstream] upsets me. the non-deterministic zone names are deprecated, so you'd think they could just use the only ones they support.

Alphadelta14 (approved):

misc/helm-charts/operator/templates/deployment.yaml
jubrad: yeah, it's certainly not ideal, also not ideal that we were taking it to another level.

PR #36516, opened 2026-05-11
orchestrator-kubernetes: suppress minDomains when availability_zones is set

jubrad:
## Problem

When `availability_zones` pins pods to specific AZs via node affinity, `minDomains` was still being applied to the topology spread constraint. `minDomains` tells the scheduler "there must be at least N eligible topology domains." If `minDomains` exceeds the number of pinned zones, Kubernetes treats the global minimum as 0 — and with `maxSkew=1`, only one pod can be scheduled at a time, leaving additional replicas stuck pending indefinitely.

Fixes CLO-74.

## Solution

Suppress `minDomains` in the `TopologySpreadConstraint` when `availability_zones` is set, mirroring the existing behavior for soft spread constraints. A warning is logged when `min_domains` is configured but will be ignored.

## Testing

See follow-up commit for unit test.

### DIFF
--- a/src/orchestrator-kubernetes/src/lib.rs
+++ b/src/orchestrator-kubernetes/src/lib.rs
@@ -801,11 +801,11 @@ impl NamespacedOrchestrator for NamespacedKubernetesOrchestrator {
 
                 let constraint = TopologySpreadConstraint {
                     label_selector: Some(ls),
-                    min_domains: if config.soft || availability_zones.is_some() {
-                        None
-                    } else {
-                        config.min_domains
-                    },
+                    min_domains: topology_spread_min_domains(
+                        config.soft,
+                        availability_zones.is_some(),
+                        config.min_domains,
+                    ),
                     max_skew: config.max_skew,
                     topology_key: "topology.kubernetes.io/zone".to_string(),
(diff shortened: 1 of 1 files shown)


src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -791,14 +791,21 @@ impl NamespacedOrchestrator for NamespacedKubernetesOrchestrator {
...
                 let constraint = TopologySpreadConstraint {
                     label_selector: Some(ls),
-                    min_domains: if config.soft {
-                        None
-                    } else {
-                        config.min_domains
-                    },
+                    min_domains: topology_spread_min_domains(
+                        config.soft,
+                        availability_zones.is_some(),
```
Alphadelta14, line 806: this is fine for clusters of our size, but if we were to pin to say 4 AZs, then having minDomains as 2 or 3 is totally reasonable.

Alphadelta14 (approved):

src/orchestrator-kubernetes/src/lib.rs
jubrad: Totally! It's possible this should just be a var on environmentd to make it more configurable.

PR #36519, opened 2026-05-11
Fix negative multiplicities caused by v80->v81 catalog migration

SangJunBak:
Rebased off and dependent on https://github.com/MaterializeInc/materialize/pull/36524 

### Motivation

Creates a regression test for fixing negative multiplicities caused by caused by v80->v81 catalog migration seen in incident-1007

### Description

- Adds a v82 to v83 migration that fixes forward and clears out any negative multiplicities.

### Verification

### DIFF
--- a/test/catalog-migration/incident-1007/mzcompose.py
+++ b/test/catalog-migration/incident-1007/mzcompose.py
@@ -1,0 +1,175 @@
+# Copyright Materialize, Inc. and contributors. All rights reserved.
+#
+# Use of this software is governed by the Business Source License
+# included in the LICENSE file at the root of this repository.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0.
+
+"""
+End-to-end repro and repair of the v80-form Role byte-drift bug.
+
+Catalog versions:
+  v26.17.x -> catalog 80 (pre-bug)
+  v26.18.0 -> catalog 81, introduces the buggy `v80_to_v81::upgrade` whose
+             `is_cloud` heuristic silently no-ops on self-managed envs.
+             Role rows stay in their v80 byte form
+             (no `auto_provision_source` key) while the catalog version
+             advances. Any subsequent ALTER reads the v80 form, parses it,
+             and writes a retract+insert pair re-serialized through the
+             current proto — which always includes `auto_provision_source:
+             null` — leaving a dangling `-1`.
+  current  -> catalog 83, this PR's `v82_to_v83::upgrade` scans for the
+             structural signature of the drift and emits compensating
+             updates to retire the dangling `-1` and the stale `+1`.
+
+The workflow chains v26.17 -> v26.18 -> current. At v26.18, we expect querying
+`mz_internal.mz_catalog_raw` to fail — `PersistPeek` enforces a per-row
+non-negativity check that fires on the dangling `-1`. After upgrading to
+current, the repair fires and the same query succeeds.
+"""
+
+from textwrap import dedent
+
+from materialize.docker import image_registry
+from materialize.mz_version import MzVersion
+from materialize.mzcompose.composition import Composition, WorkflowArgumentParser
+from materialize.mzcompose.services.materialized import Materialized
+from materialize.mzcompose.services.minio import Minio
+from materialize.mzcompose.services.postgres import PostgresMetadata
+from materialize.mzcompose.services.testdrive import Testdrive
+from materialize.ui import UIError
+
+# Pre-bug release: catalog 80.
+PRE_BUG_VERSION = MzVersion.parse_mz("v26.17.1")
+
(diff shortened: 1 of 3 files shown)


def- (commented):
Nightly triggered, I'll take a closer look at the code now: https://buildkite.com/materialize/nightly/builds/16399

Edit: Fixed the miri test failures.

ci/test/pipeline.template.yml
```diff
@@ -496,6 +496,16 @@ steps:
     agents:
       queue: hetzner-x86-64-4cpu-8gb
 
+  - id: catalog-migration-incident-1007
```
def-: Upgrade tests can't run in Test pipeline, they are way too slow. Should be moved to Nightly. I would also prefer if this was caught in platform-checks additionally, not just with those specific versions, but with any version in the future exhibiting similar bugs.
I'll try add it to platform-checks.

test/catalog-migration/incident-1007/mzcompose.py
```diff
@@ -0,0 +1,175 @@
...
+             `is_cloud` heuristic silently no-ops on self-managed envs.
+             Role rows stay in their v80 byte form
+             (no `auto_provision_source` key) while the catalog version
+             advances. Any subsequent ALTER reads the v80 form, parses it,
+             and writes a retract+insert pair re-serialized through the
+             current proto — which always includes `auto_provision_source:
+             null` — leaving a dangling `-1`.
+  current  -> catalog 83, this PR's `v82_to_v83::upgrade` scans for the
+             structural signature of the drift and emits compensating
+             updates to retire the dangling `-1` and the stale `+1`.
```
def-, line 25: I'm not sure if we still need this test at all then, we could add a scenario to platform checks that does these versions too one we add the scenario that catches it by altering the role and then selecting from the catalog.

test/catalog-migration/incident-1007/mzcompose.py
```diff
@@ -0,0 +1,175 @@
...
+    # repair). After that the dangling `-1` and stale `+1` are gone and the
+    # SELECT succeeds.
+    _start_at(c, None)
+    _select_roles(c)
+
+    # Ensure the role is still dropped after the repair.
+    c.testdrive(dedent("""
+            > SELECT EXISTS (SELECT 1 FROM mz_roles WHERE name = 'user1');
+            false
+            """))
```
def-, line 175: The entire file smelly pretty LLMy, I think we shouldn't have one mzcompose.py test per bug/incident we run into, but use the existing frameworks when appropriate, so that we find similarly shaped bugs more generally.

def-:
I pushed platform checks extensions that would have caught this:
```
source: /home/deen/git/materialize/misc/python/materialize/checks/all_checks/password_auth.py:114
2:1: error: executing postgres query: db error: ERROR: Invalid data in source, saw retractions (1) for row that does not exist: [Map({"key": Map({"id": Map({"User": Numeric(<#>)})}), "kind": String(<"XXXX">), "value": Map({"attributes": Map({"auto_provision_source": JsonNull, "inherit": True, "login": True, "superuser": JsonNull}), "membership": Map({"map": List([])}), "name": String(<"XXXX#">), "oid": Numeric(<#####>), "vars": Map({"entries": List([])})})})]
     |
   1 |
   2 | $ postgres-execute c ... [rest of line truncated for security]
     | ^
```
Test run on this PR now: https://buildkite.com/materialize/nightly/builds/16406

Edit: Removed it for now, I'll do it in a separate PR so that we can merge this test already.

ci/test/pipeline.template.yml
def-: Upgrade tests can't run in Test pipeline, they are way too slow. Should be moved to Nightly. I would also prefer if this was caught in platform-checks additionally, not just with those specific versions, but with any version in the future exhibiting similar bugs.
I'll try add it to platform-checks.

def- (dismissed):

test/catalog-migration/incident-1007/mzcompose.py
SangJunBak: Fair point!

test/catalog-migration/incident-1007/mzcompose.py
SangJunBak: I agree! I was actually using platform checks when I was first testing 😅 ! But then I wanted more flexibility like testing multiple `DROP`+ after the corruption. But I agree that if this can be caught using platform checks, we should remove this test in favor of the framework.

SangJunBak:
@def- I agree that we should write a platform check test instead of a bespoke mzcompose one, but I think we can simply delete this test when the platform check test merges? The benefit I see of having this in main to have a test for the current release / patch fix.

bosconi (approved):
As discussed, Jun and I are going with merging this now, to have something in place ahead of `v26.23.3`. We can back it out again after.

PR #36526, opened 2026-05-12
compute: dual-output sessions for flat_map

antiguru:
### Motivation

The arrangement flat_map path threaded results through a single stream of `Result` tuples and re-split them with a downstream `map_fallible("OkErr", ...)`.
The demux is a per-callsite operator we can drop.
The session-based shape is also a prerequisite for moving to columnar dataflow edges: the closure has to write *references* into the output builder, and iterators can't return references with the right lifetimes.

### Description

* `PendingWork` retains a capability per output; `do_work` writes to ok+err sessions and returns work directly.
* `flat_map_core_fallible` builds a two-output operator and returns `(ok_stream, err_stream)`.
  Fallible callers (`as_collection_core`, the reduce key/val selector) drop their `OkErr` demux.
* `flat_map_core_ok` is a single-output sibling for callers that statically never emit errors.
  `as_specific_collection` uses it so imported arrangements don't pay for an empty err output and concat.
* The ok output container builder is generic at each call site rather than hardcoded.
  `as_specific_collection` opts into `CapacityContainerBuilder` (1:1 cursor walk, no duplicates); `as_collection_core` and the reduce key/val selector keep `ConsolidatingContainerBuilder` (MFP and key/val fan-out can collide).
  This also unblocks plugging in a columnar ok container without changing the API.
* Cursor walking is factored into `walk_cursor`, shared between the two `do_work` impls.

### Fuel

The closure returns the number of records *produced*, not input tuples consumed.
This is intentional and asymmetric: a selective `filter(false)` MFP runs the batch to completion in one activation so the operator can drop its `Batch` clone and release upstream memory; a wide `map(...)` MFP is bounded by emit count so a single activation can't flood downstream.
`REFUEL = 1_000_000` is a pragmatic compromise — no universal value exists across MFP shapes.

### DIFF
--- a/src/compute/src/render/context.rs
+++ b/src/compute/src/render/context.rs
@@ -330,6 +330,54 @@ impl<'scope, T: RenderTimestamp> ArrangementFlavor<'scope, T> {
             }
         }
     }
+
+    /// Ok-only variant of [`Self::flat_map`]. The `logic` callback receives a single output
+    /// session and cannot produce errors. The returned err collection comes solely from the
+    /// arrangement; no extra operator is built to carry an empty MFP-error stream.
+    pub fn flat_map_ok<D, L>(
+        &self,
+        key: Option<&Row>,
+        max_demand: usize,
+        mut logic: L,
+    ) -> (
+        StreamVec<'scope, T, (D, T, Diff)>,
+        VecCollection<'scope, T, DataflowErrorSer, Diff>,
+    )
+    where
+        D: Data,
+        L: for<'a, 'b> FnMut(
+                &'a mut DatumVecBorrow<'b>,
+                T,
+                Diff,
+                &mut Session<T, (D, T, Diff)>,
+            ) -> usize
+            + 'static,
+    {
+        let refuel = 1000000;
+
+        let mut datums = DatumVec::new();
+        let logic =
+            move |k: DatumSeq, v: DatumSeq, t, d, ok_session: &mut Session<T, (D, T, Diff)>| {
+                let mut datums_borrow = datums.borrow();
+                datums_borrow.extend(k.to_datum_iter().take(max_demand));
+                let max_demand = max_demand.saturating_sub(datums_borrow.len());
+                datums_borrow.extend(v.to_datum_iter().take(max_demand));
+                logic(&mut datums_borrow, t, d, ok_session)
+            };
+
+        match &self {
+            ArrangementFlavor::Local(oks, errs) => {
+                let oks = CollectionBundle::<T>::flat_map_core_ok(oks.clone(), key, logic, refuel);
+                let errs = errs.clone().as_collection(|k, &()| k.clone());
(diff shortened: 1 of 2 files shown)


src/compute/src/render/context.rs
```diff
@@ -520,38 +619,72 @@ impl<'scope, T: RenderTimestamp> CollectionBundle<'scope, T> {
...
+                        let err_cap = time.retain(1);
+                        let mut ok_session = ok_output.session_with_builder(&ok_cap);
+                        let mut err_session = err_output.session_with_builder(&err_cap);
+                        for (v, t, d) in data.iter() {
+                            logic(
+                                &mut datums.borrow_with_limit(v, max_demand),
+                                t.clone(),
+                                d.clone(),
+                                &mut ok_session,
+                                &mut err_session,
```
DAlperin: this is perhaps like marginally worse than befor in that we clone d and t. Could perhaps get a mut out of a data.drain() but relatively minor

src/compute/src/render/context.rs
```diff
@@ -926,11 +1165,30 @@ impl<'scope, T: RenderTimestamp> CollectionBundle<'scope, T> {
...
+/// Type alias for a timely output `Session` whose capability is a `Capability<T>`. The container
+/// builder `CB` is left to the caller; sessions can therefore drive consolidating, capacity, or
+/// (in the future) columnar output builders without changing call sites.
+type Session<'a, 'b, T, CB> =
+    timely::dataflow::operators::generic::Session<'a, 'b, T, CB, Capability<T>>;
+
+/// Container builder used for the err output of every flat_map variant. Errors are low volume
+/// and we don't expect within-batch duplicates, so [`CapacityContainerBuilder`] is the right
+/// default; this matches the pre-refactor behavior of the now-removed `map_fallible` demux.
+type ECB<T> = CapacityContainerBuilder<Vec<(DataflowErrorSer, T, Diff)>>;
```
DAlperin: Didn't we Consolidate upstream which we no longer do https://github.com/MaterializeInc/materialize/pull/36526/changes#diff-3cecdf974fdf8212982749369ded60f390b84b56786af3fad9a5eee34a310c75L149? Errors are sparse so this is probably fine but not sure it's strictly pre-refactor

DAlperin (approved):
One nit and one structural question otherwise lgtm

src/compute/src/render/context.rs
antiguru: Switched to `data.drain(..)` in 531af764ae's parent (f18f4cf8af) — owns the tuple, drops the clones.

src/compute/src/render/context.rs
antiguru: Good catch. Restored consolidation on the err side in 531af764ae by switching `ECB` to `ConsolidatingContainerBuilder`.

antiguru:
Thanks for the review!

PR #36528, opened 2026-05-12
sql: place NULL rows last in EXPLAIN ANALYZE CLUSTER output

leedqin:
### Motivation
Explain analyze cluster and memory was placing null rows first
Fixes [CNS-65](https://linear.app/materializeinc/issue/CNS-65/nit-explain-analyze-cluster-memory-sorts-by-memory-descending-but-null)

### Description

Fixed the sorting order for Explain Analyze output
image


ggevay (approved):
Thank you, LGTM!

PR #36530, opened 2026-05-12
environmentd: tag built-in MCP sessions with application_name

sjwiesman:
Sets a default application name for mcp endpoint queries so they can be attributed in the query and session logs. 

### Motivation

Closes <issue>
Fixes <bug>
etc.

### Description

diff *can't* tell a reviewer.

### Verification


mtabebe (approved):
@sjwiesman this looked good to me, I just added the name to the hint.rs so we can get this in Prometheus metrics as a recognized app name

sjwiesman:
Good catch!

PR #36525, opened 2026-05-12
platform-checks: Add check and scenario for incident #1007 & extend existing incident-1007 test

def-:
Depends on https://github.com/MaterializeInc/materialize/pull/36534, based on top of it.

Test run with platform-checks panicking: https://buildkite.com/materialize/nightly/builds/16420#019e1ef7-f83e-4ba4-a950-cb5e89b31271
Test run with incident-1007: https://buildkite.com/materialize/nightly/builds/16421

New test run rebased on top of https://github.com/MaterializeInc/materialize/pull/36534: https://buildkite.com/materialize/nightly/builds/16425

PR #36534, opened 2026-05-13
catalog v82->v83: normalize live Role rows still in pre-v81 byte form

DAlperin:
From @def- LLM review:

> The new migration only emits repairs for keys that already have a dangling negative diff. compute_repairs first records +1 Role rows, then only examines snapshot entries whose diff is not +1; a Role is repaired only if there is an existing -1 plus a parsed-equal +1 sibling ([src/catalog/src/durable/upgrade/v82_to_v83.rs](https://github.com/MaterializeInc/qa-llm-review/blob/master/home/deen/git/materialize-review/src/catalog/src/durable/upgrade/v82_to_v83.rs:155), especially lines 162-165 and 180-240). A lone v80-form +1 Role row is therefore treated as healthy and left untouched; the new tests also only cover already-corrupted shapes (stale +1 plus dangling -1) and never a snapshot that contains just the stale pre-v81 byte form ([src/catalog/src/durable/upgrade/v82_to_v83.rs](https://github.com/MaterializeInc/qa-llm-review/blob/master/home/deen/git/materialize-review/src/catalog/src/durable/upgrade/v82_to_v83.rs:404)).

This fix recanonicalizes Roles in v83 format if they are different. I added point tests but dennis will also rebase on this.


def- (approved):
lgtm from a testing perspective: https://buildkite.com/materialize/nightly/builds/16425 with https://github.com/MaterializeInc/materialize/pull/36525 rebased on top of this

def-:
@claude /review

src/catalog/src/durable/upgrade/v82_to_v83.rs
```diff
@@ -644,6 +717,137 @@ mod tests {
...
+        let canonical = role_kind(51, "dave@materialize.com", 20101, Some(true), None, None);
+        let snap = snapshot(vec![(canonical, Diff::ONE)]);
+        let (repairs, stats) = compute_repairs(&snap);
+        assert!(repairs.is_empty());
+        assert_eq!(stats.normalized, 0);
+    }
+
+    #[mz_ore::test]
+    #[cfg_attr(miri, ignore)] // can't call foreign function `decContextDefault` on OS `linux`
+    fn multiple_stale_byte_forms_no_dangling_collapse_to_one_canonical() {
```
bosconi: A function name James Joyce would be proud  of!

bosconi (approved):
Looks good!

def-:
One thing popped up based on the [QA LLM review](https://github.com/MaterializeInc/qa-llm-review/blob/master/commit-bugs/done/analysis-pr-36534.md):

`auto_provision_source: null` is now set for everything instead of `Some(Frontegg)` for the original v80->v81 migration. I'm not sure how bad that is, and how much we care. CC @SangJunBak

aljoscha:
I think @def- ia right. To preserve the spirit of the original migration, we need to apply the same logic that decided for frontegg or None. Otherwise this looks very good.

Basically, in the normalization function apply the logic because the normalized form is dynamic, if that makes sense.

ggevay:
Now we need to put this in a new migration before merging, right? Discussed [here](https://materializeinc.slack.com/archives/C0B2XR8SLR4/p1778851325504729?thread_ts=1778827001.541149&cid=C0B2XR8SLR4)

SangJunBak:
Added three new commits which:

- Adds a new catalog migration
- Add the original logic of the v80_to_v81 migration (adding the auto_provision_source property) to this migration
- Makes sure this migration treats environments with an unmanaged mz_system cluster but with at least one replica as cloud environments

SangJunBak:
@claude /review

src/catalog/src/durable/upgrade/v83_to_v84.rs
```diff
@@ -0,0 +1,1357 @@
...
+    }
+
+    let mut updates: Vec<(StateUpdateKindJson, Diff)> = repairs;
+    updates.push((version_update_kind(FROM_VERSION), Diff::MINUS_ONE));
+    updates.push((version_update_kind(TO_VERSION), Diff::ONE));
+
+    // Print the updates for inspection.
+    tracing::info!(
+        "catalog upgrade v{FROM_VERSION}_to_v{TO_VERSION}: about to apply updates: {updates:?}"
+    );
```
DAlperin, line 147: might be a bit verbose in self managed, could be debug, but also 🤷🏼

src/catalog/src/durable/upgrade/v83_to_v84.rs
```diff
@@ -0,0 +1,1357 @@
...
+            continue;
+        }
+        tracing::info!(
+            role_name = %role.value.name,
+            "normalizing Role row to canonical byte form",
+        );
+        repairs.push((kind_json.clone(), Diff::MINUS_ONE));
+        if inserted_canonical.insert(canonical.clone()) {
+            repairs.push((canonical, Diff::ONE));
+        }
```
DAlperin, line 354: nice, we should definitely consider a follow on where we run this kind of check after every upgrade against all object types

src/catalog/src/durable/upgrade/v83_to_v84.rs
```diff
@@ -0,0 +1,1357 @@
...
+///
+/// In both cases, the `enable_password_auth` system parameter must not be
+/// `on` — only self-managed envs enable it, so that's the deciding tie-breaker
+/// for any env whose mz_system shape would otherwise look cloud-ish.
+///
+/// Returns false on any error path (missing `mz_system`, failed parse), which
+/// is the safe default — a false negative means we leave
+/// `auto_provision_source` at `None` for a cloud env's roles, which is no
+/// worse than the v80->v81 bug we're already fixing.
+fn is_cloud_env(snapshot: &[(StateUpdateKindJson, Timestamp, Diff)]) -> bool {
```
DAlperin, line 380: The heuristics here seem fine, but for how often this comes up lets start just injecting a IS_MZ_CLOUD=1 env var we can use in cloud lol

test/catalog-migration/incident-1007/mzcompose.py
```diff
@@ -161,15 +175,38 @@ def workflow_default(c: Composition, parser: WorkflowArgumentParser) -> None:
...
-            > SELECT EXISTS (SELECT 1 FROM mz_roles WHERE name = 'user1');
-            false
+            $ postgres-execute connection=postgres://mz_system:materialize@${testdrive.materialize-internal-sql-addr}
+            ALTER ROLE user3 SUPERUSER;
             """))
+
+    c.kill("materialized")
+    _start_at(c, None)
+    # We expect there to not be a negative multiplicity on user3 like there was for the same test on user2.
+    _select_roles(c)
```
DAlperin, line 212: Nice, good test, thanks

DAlperin (commented):
This is me saying LGTM but I can't actually stamp it since it's my PR

src/catalog/src/durable/upgrade/v83_to_v84.rs
SangJunBak: Ah good catch actually. This was from me debugging

src/catalog/src/durable/upgrade/v83_to_v84.rs
SangJunBak: Following up here https://materializeinc.slack.com/archives/C08A62E0751/p1778880924043329

src/catalog/src/durable/upgrade/v83_to_v84.rs
SangJunBak: Following up here https://materializeinc.slack.com/archives/C08A62E0751/p1778880924043329

SangJunBak (approved):
LGTM

PR #36537, opened 2026-05-13
txn-wal: deasync txns_progress_frontiers + retain remap across close

antiguru:
Three changes to `txns_progress_frontiers`:

1. **Deasync.** Convert from `AsyncOperatorBuilder` to sync `OperatorBuilderRc`. The operator's only awaits were on input handles, so the async state machine collapses into a schedule closure. Inlines the `txns_progress_frontiers_read_remap_input` helper. Shutdown plumbed via `builder_async::button` to preserve the `PressOnDropButton` return.

2. **Fix latent stall bug (PER-4).** The previous implementation dropped the last `DataRemapEntry` as soon as the remap input's frontier reached the empty antichain. From that point the "advance to `logical_upper`" branch was gated off, so the output capability could only advance via the passthrough input's frontier — bounded by the data shard's physical upper. Retain the most recently observed `DataRemapEntry` indefinitely so subsequent activations keep advancing past `physical_upper` to `logical_upper`. Regression test added.

3. **Gate passthrough frontier on remap state.** The deasync rewrite naively consulted the passthrough frontier every activation, which broke `SELECT AS OF MAX`: the passthrough frontier eventually reaches the empty antichain, the operator drops its capability, and the SELECT completes with an empty result instead of blocking until the (impossible) timestamp exists. `test-replica-targeted-select-abort` relied on the blocking behavior. Mirror the async state machine: only consult the passthrough frontier (until check, capability downgrade, empty-antichain shutdown) when the operator is in "WaitingForPassthrough" mode — the remap input has closed, or remap's `physical_upper` has moved ahead of the output capability. In "WaitingForRemap" mode the operator still advances via `logical_upper` bumps but ignores the passthrough frontier, so a stuck remap stream keeps the SELECT blocked indefinitely.

### DIFF
--- a/src/txn-wal/src/operator.rs
+++ b/src/txn-wal/src/operator.rs
@@ -270,11 +270,14 @@ where
         // copying along the passthrough input. `None` indicates that we've
         // dropped the capability to shut down.
         let mut capability: Option<_> = capabilities.into_iter().next();
-        // None is used to indicate that both uppers are the empty antichain.
-        let mut remap = Some(DataRemapEntry {
+        // The most recently observed remap state. Retained even after the
+        // remap input closes so we can still advance the output capability to
+        // the last known `logical_upper` while the passthrough input is
+        // still draining.
+        let mut remap = DataRemapEntry {
             physical_upper: T::minimum(),
             logical_upper: T::minimum(),
-        });
+        };
 
         move |frontiers| {
             if shutdown_handle.local_pressed() {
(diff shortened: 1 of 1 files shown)


src/txn-wal/src/operator.rs
```diff
@@ -244,166 +244,161 @@ where
...
+    let mut remap_input = builder.new_input_connection(remap, Pipeline, []);
+    let mut passthrough_input = builder.new_input_connection(passthrough, Pipeline, []);
+
+    let (shutdown_handle, shutdown_button) = button(scope, info.address);
+
+    builder.build(move |capabilities| {
+        // The output capability's time tracks how far we've progressed in
+        // copying along the passthrough input. `None` indicates that we've
+        // dropped the capability to shut down.
+        let mut capability: Option<_> = capabilities.into_iter().next();
```
aljoscha: nit: the previous version that asserts on the expected number is kinda nice, but oh well ...

src/txn-wal/src/operator.rs
```diff
@@ -312,52 +316,71 @@ where
...
-                // Recall that any reads of the data shard are always correct, so
-                // given that we've passed through any data from the input, that
-                // means we're free to pass through frontier updates too.
-                if let Some(cap) = capability.as_mut() {
-                    if cap.time() < new_progress {
-                        debug!("{} downgrading cap to {:?}", name, new_progress);
-                        cap.downgrade(new_progress);
+            debug!("{} remap {:?} remap_closed={}", name, remap, remap_closed);
+
+            // Mirror the async state machine: only consult the passthrough
```
aljoscha: This comment will not age well and should stand on it's own. There is no async state machine anymore after this change

src/txn-wal/src/operator.rs
antiguru: Fixed!

src/txn-wal/src/operator.rs
antiguru: True, reworded.

aljoscha (approved):
Looks good once we have a green ci run _and_ nightly!

antiguru:
CI is green, I triggered another nightly run (https://buildkite.com/materialize/nightly/builds/16433). Thanks for the review!

PR #36540, opened 2026-05-13
spec-sheet: add new envd-object-scalability and cluster-object-limits scenarios

aljoscha:
Fixes SQL-222


test/cluster-spec-sheet/mzcompose.py
```diff
@@ -200,7 +253,23 @@ def add_result(
         size_bytes: int | None,
         time: float | None = None,
         qps: float | None = None,
+        time_ms: int | None = None,
+        healthy: int | None = None,
     ) -> None:
+        """Write one result row.
```
aljoscha, line 255: Do we need this dance around time/time_ms? Couldn't we just update the existing callsites to pass in time_ms because apparently here we always convert to time_ms

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -445,7 +515,17 @@ def size_of_dataflow(self, object: str) -> int | None:
...
 
 
-class Scenario(ABC):
+class ClusterScalingScenario(ABC):
+    """Base for scenarios driven by the strong/weak/envd_qps families.
+
+    These scenarios share a four-step lifecycle: `drop()` cleans up state from
+    a prior run, `setup()` creates load-generator clusters / sources / tables,
+    `materialize_views()` lists persist objects to hydrate, and `run()`
+    performs the measurements. The drivers (`run_scenario_strong`,
```
aljoscha: Do we still have these per scenario drivers?

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -2186,175 +2240,1289 @@ def run(self, runner: ScenarioRunner) -> None:
...
-    try:
-        if hard:
-            composition.run("mz", "region", "disable", "--hard", rm=True)
-        else:
-            composition.run("mz", "region", "disable", rm=True)
-    except UIError:
-        # Can return: status 404 Not Found
-        pass
+    The scale point (number of catalog objects, N) is driven externally by
+    `run_scenario_envd_objects_scalability` via init/add_objects/teardown;
```
aljoscha: we still have these scenario runners?

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -2186,175 +2240,1289 @@ def run(self, runner: ScenarioRunner) -> None:
...
+def _make_csv_writer(file: TextIO, fieldnames: list[str]) -> csv.DictWriter:
+    writer = csv.DictWriter(file, fieldnames=fieldnames, extrasaction="ignore")
+    writer.writeheader()
+    return writer
+
+
+def _recreate_cluster_c(
+    runner: ScenarioRunner,
+    replica_size: str,
+    smoke_test: bool = False,
```
aljoscha, line 2357: What do we have this smoke_test thing for?

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -2186,175 +2240,1289 @@ def run(self, runner: ScenarioRunner) -> None:
...
+    """A single point in a scenario's sweep.
+
+    Each field is meaningful only for sweeps that vary it; the rest are
+    ``None``. The scenario interprets its fields in ``apply()`` to set up
+    the world (recreate the cluster, reconfigure envd, add objects). The
+    driver doesn't look at any field except ``label``.
+    """
 
+    label: str
+    cluster_size: str | None = None
```
aljoscha, line 2464: What's the difference between cluster_size and replica_scale?

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -2186,175 +2240,1289 @@ def run(self, runner: ScenarioRunner) -> None:
...
+            target_n=target_n,
+            templates_for=lambda i: (
+                f"CREATE TABLE IF NOT EXISTS {self.PAD_SCHEMA}.pad_t_{i} (a int, b text)",
+            ),
+            label="tables",
+        )
+
+
+class EnvdObjectsScalabilityMvsScenario(EnvdObjectsScalabilityScenario):
+    """N materialized views in the catalog, sharded across pad clusters.
```
aljoscha: We should also mention that each MV has it's own persist source (the table). This is as opposed to say having one table (persist source) that goes into an index, and then many materialized views off that index.

aljoscha: And the same holds for the index scenario below. We can note that as follow-up work we want to also do the one table into index into many indexes scenario

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -2186,175 +2240,1289 @@ def run(self, runner: ScenarioRunner) -> None:
...
+            k = self._pad_clusters_created
+            print(f"  creating pad cluster pad_c_{k} (size {self._pad_replica_size})")
+            runner.run_query(f"DROP CLUSTER IF EXISTS pad_c_{k} CASCADE")
+            runner.run_query(
+                f"CREATE CLUSTER pad_c_{k} SIZE '{self._pad_replica_size}'"
+            )
+            self._pad_clusters_created += 1
+
+    def add_objects(self, runner: ScenarioRunner, target_n: int) -> None:
+        # Build one shard at a time: each shard lives on its own pad cluster
```
aljoscha: Shard naming is confusing here because persist also has the concept of shards. Audit the file for this and find something else.

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -2186,175 +2240,1289 @@ def run(self, runner: ScenarioRunner) -> None:
...
+
+@dataclass(frozen=True)
+class ClusterObjectLimitsKind:
+    """Per-materialization-type knobs for a ClusterObjectLimitsScenario.
+
+    Each materialization (one "test object") is created from one or more SQL
+    statements (indexes need a view + a default index; MVs need a single
+    CREATE MATERIALIZED VIEW). Statement templates take ``{schema}``,
+    ``{base}``, and ``{i}`` placeholders; the ``WHERE id < {i}`` predicate
+    is what makes each view/MV structurally distinct so plan caching can't
```
aljoscha: I don't think it's plan caching that would do this in materialize, but we are doing it because some other things might. Let's drop this, or let's just say to make sure that we get separate dataflows. Audit the reset of the file for this kind of thing and also fix up those places.

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -2186,175 +2240,1289 @@ def run(self, runner: ScenarioRunner) -> None:
...
+    """
+    Scenario for cluster_object_limits. Finds the maximum number of idle
+    materializations (indexes or materialized views, picked via ``kind``)
+    that one cluster can keep fresh.
+
+    Structure:
+      - A single one-row base table `base_t` (never updated) lives in
+        ``PAD_SCHEMA``. Frontiers advance over wall-clock time, so even with
+        no writes the cluster must keep ticking every materialization's
+        write_frontier — which is what saturates an undersized cluster.
```
aljoscha: the saturation is independent of cluster size, so scratch that last bit

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -2186,175 +2240,1289 @@ def run(self, runner: ScenarioRunner) -> None:
...
+        )
+        assert rows is not None and len(rows) == 1
+        total, reporting, max_lag_ms = rows[0]
+        return int(total), int(reporting), float(max_lag_ms)
+
+    def probe_hydrated(self, runner: ScenarioRunner) -> tuple[int, int]:
+        """Return (total, hydrated) for our test objects on cluster ``c``.
+
+        Decoupled from `local_lag` so the freshness loop can wait for the
+        cluster to be *running* first, then judge whether it can keep up —
```
aljoscha: *running* might not be the right framing here. Let's just say "wait for objects to be hydrated"

test/cluster-spec-sheet/mzcompose.py
```diff
@@ -2186,175 +2240,1289 @@ def run(self, runner: ScenarioRunner) -> None:
...
+        timeout_s: int,
+        samples: int,
+        sample_interval_s: float,
+        lag_threshold_ms: int,
+    ) -> tuple[float, bool, int, int]:
+        """
+        Two-phase probe: wait for hydration, then take steady-state lag
+        samples. Returns (max_lag_ms, healthy, reporting, total).
+
+        Splitting the two phases keeps unhealthy probes fast: an overloaded
```
aljoscha: This is not true, an overloaded cluster will also not hydrate quickly, often, so we have to workshop sth here

doc/developer/reports/20260520_spec_sheet_first_numbers.md
```diff
@@ -0,0 +1,206 @@
...
+- **Data sources:** release-qualification builds [1229](https://buildkite.com/materialize/release-qualification/builds/1229) (envd) and [1231](https://buildkite.com/materialize/release-qualification/builds/1231) (cluster), both against `cloud-staging` (aws-eu-west-1).
+
+The spec sheet is a catalogue of idle ceilings — how many things an
+environment or a single cluster can carry when those things aren't doing
+anything. These are upper bounds. Once real workload (writes, non-trivial
+queries, source ingestion) lands on top, the same environment or cluster
+will tolerate fewer objects than this. The point of measuring the idle
+ceiling is that everything else has to fit underneath it.
+
+This report is the third pass over these scenarios; the first pass
```
aljoscha: don't need this paragraph, readers don't need to know the history of the testing

doc/developer/reports/20260520_spec_sheet_first_numbers.md
```diff
@@ -0,0 +1,206 @@
...
+   the steady-state lag check are recorded with `failure_mode = "lag"`;
+   healthy probes have `failure_mode` empty.
+
+## Results — cluster object limits
+
+![Idle materializations per cluster size](./static/20260520_spec_sheet_first_numbers/cluster_object_limits_max_n.png)
+
+Three observations:
+
+**1. The `from_index` variant unlocks ~1.7× more objects on small
```
aljoscha: cut this paragraph

doc/developer/reports/20260520_spec_sheet_first_numbers.md
```diff
@@ -0,0 +1,206 @@
...
+every worker on a cluster owns a slice of every dataflow on that
+cluster and has to tick the frontier on each materialization in its
+slice regardless of cluster size. Adding workers does not reduce
+per-worker idle work; it shards the same per-object cost across more
+workers that each still carry their share, and cross-worker
+coordination grows on top of that. So the prediction for purely idle
+objects is a flat-to-downward ceiling as the cluster grows, and that
+is what we see.
+
+**3. There are spikes worth flagging, not yet acting on.** The
```
aljoscha: also need to distill this one, just mention that we think the one deep outlier is an outlier, leave it at that

doc/developer/reports/20260520_spec_sheet_first_numbers.md
```diff
@@ -0,0 +1,206 @@
...
+on `from_persist_sources` at the same size). The MV curve dips to 193
+at `400cc`, well below its neighbours (481 / 500 / 687). Both look
+like single-cell outliers in a single run rather than a real cliff at
+those sizes — the bisects converged cleanly, but the surrounding
+values don't support a real step-change. The next planned action is a
+repeat run to see whether they reproduce; if they do, both warrant
+investigation. If they don't, our methodology needs more samples per
+size before we publish absolute numbers.
+
+A subtler finding from this run's per-sample data: many of the
```
aljoscha: Also cut this one, say that measurements are still a bit flaky and we'll improve as follow-up. But directionally the results look correct

doc/developer/reports/20260520_spec_sheet_first_numbers.md
```diff
@@ -0,0 +1,206 @@
...
+
+![Adapter DDL latency vs catalog size](./static/20260520_spec_sheet_first_numbers/envd_objects_scalability_ddl_latency.png)
+
+The DDL path slows by roughly 1.7–2× between `N=1` and `N=30000`, with
+the knee around `N=5k–10k`. Tables and MVs track each other closely;
+the slope is essentially the same. The MV baseline at small `N` is
+slightly lower than tables, which is mildly surprising given MVs are
+the "heavier" object — could be variance, could be a real difference
+in the adapter path for MVs at low cardinality, not investigated
+further here.
```
aljoscha: drop this last bit here, no need to speculate

doc/developer/reports/20260520_spec_sheet_first_numbers.md
```diff
@@ -0,0 +1,206 @@
...
+recorded.** The earlier 60-second hydration window timed out
+deterministically on `1600cc` / `3200cc` at the coarse `N=2000`
+step, which yielded a measured "cliff" that was driven by hydration
+time, not steady-state capacity — bigger replicas paradoxically
+reported smaller max-healthy `N`. The window is now 300 seconds (180 s
+first-probe budget after `CREATE CLUSTER c` to absorb cold-start), and
+hydration-timeout failures are written to a new `failure_mode` column
+in the cluster_object_limits CSV (`"hydration_timeout"` vs `"lag"` vs
+empty when healthy). The existing `healthy` 0/1 column is unchanged,
+so prior analysis code keeps working.
```
aljoscha: don't need this section, readers are only interested in the final state

doc/developer/reports/20260520_spec_sheet_first_numbers.md
```diff
@@ -0,0 +1,206 @@
...
+
+One run per data point. Every materialization here is derived from a
+one-row table that never changes — real workloads will lower these
+ceilings, often by a lot. The numbers are also specific to the
+`cloud-staging` environment with its current `environmentd` CPU
+allocation, `max_credit_consumption_rate`, and LaunchDarkly defaults; a
+production environment will land somewhere else. Two of the
+cluster-object-limits cells (`from_index` at `800cc`, `mvs` at `400cc`)
+look like single-run outliers and should be repeated before being
+quoted.
```
aljoscha: also drop the caveats here

ci/release-qualification/pipeline.template.yml
```diff
@@ -704,6 +704,34 @@ steps:
...
+        label: "Cluster spec sheet: Cluster Object Limits (against Staging)"
+        depends_on: build-aarch64
+        timeout_in_minutes: 3600
+        concurrency: 1
+        concurrency_group: 'cluster-spec-sheet-cluster'
+        plugins:
+          - ./ci/plugins/mzcompose:
+              composition: cluster-spec-sheet
+              run: default
+              args: [--cleanup, --target=cloud-staging, cluster_object_limits]
```
def-, line 735: How long does all of this take now btw?

aljoscha: They do take quite long: [https://buildkite.com/materialize/release-qualification/builds/1229](<https://buildkite.com/materialize/release-qualification/builds/1229>), so \~6h for the envd objects scalability and \~3h for the cluster objects scalability one. Is that too long? I can cut it back or make it so we don't run these on every release but only on request? Future improvements to DDL latency will likely help, but no clear roadmap when we can land these

aljoscha: Thoughts on ☝️ ? @def-

def-: It's probably fine. If we get value out of the numbers, let's go for it.

def- (approved):

PR #36541, opened 2026-05-13
Render non-identity join closures

frankmcsherry:
We fail to render non-identity join closures when they only have a projection. This leads to some confusing explain plans, where a projection appears to be missing, but is in fact just suppressed.


ggevay (approved):
Thanks!

PR #36542, opened 2026-05-13
sql: Fix panic in cardinality estimation when EXPECTED GROUP SIZE = 0

mgree:
### Motivation

https://linear.app/materializeinc/issue/SQL-273/unexpected-panic-during-query-optimization-positive-and-representable

### Description

Fixes DBZ from `EXPECTED GROUP SIZE = 0` and hardens one other possible DBZ case (`TableFunc::Wrap` with `width == 0`). Adds the test from the issue.

### Verification

Test went from red to green.


def- (approved):
Thanks, test added lgtm!

PR #36544, opened 2026-05-13
sql: Stable LIR MVP

mgree:
### Motivation

https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/20260311_optimizer_customer_tradeoff.md

### Description

Creates a separate `LirScalarExpr` structure; treats MFP-like structures and other auxiliary structures parametrically, introducing several traits (https://github.com/MaterializeInc/materialize/pull/36759, https://github.com/MaterializeInc/materialize/pull/36647).

Separately spun out https://github.com/MaterializeInc/materialize/pull/36737 and https://github.com/MaterializeInc/materialize/pull/36852.

### Verification

CI green, nightly green.

### DIFF
--- a/src/compute-types/src/explain/text.rs
+++ b/src/compute-types/src/explain/text.rs
@@ -11,10 +11,10 @@
 //!
 //! The format adheres to the following conventions:
 //! 1. In general, every line that starts with an uppercase character
-//!    corresponds to a [`Plan`] variant.
+//!    corresponds to a [`LirRelationExpr`] variant.
 //! 2. Whenever the variant has an attached `~Plan`, the printed name is
 //!    `$V::$P` where `$V` identifies the variant and `$P` the plan.
-//! 3. The fields of a `~Plan` struct attached to a [`Plan`] are rendered as if
+//! 3. The fields of a `~Plan` struct attached to a [`LirRelationExpr`] are rendered as if
 //!    they were part of the variant themself.
 //! 4. Non-recursive parameters of each sub-plan are written as `$key=$val`
 //!    pairs on the same line or as lowercase `$key` fields on indented lines.
@@ -42,13 +42,15 @@ use crate::plan::reduce::{
 };
 use crate::plan::scalar::LirScalarExpr;
 use crate::plan::threshold::ThresholdPlan;
-use crate::plan::{ArrangementStrategy, AvailableCollections, LirId, Plan, PlanNode};
+use crate::plan::{
+    ArrangementStrategy, AvailableCollections, LirId, LirRelationExpr, LirRelationNode,
+};
 
-impl DisplayText<PlanRenderingContext<'_, Plan>> for Plan {
+impl DisplayText<PlanRenderingContext<'_, LirRelationExpr>> for LirRelationExpr {
     fn fmt_text(
         &self,
         f: &mut fmt::Formatter<'_>,
-        ctx: &mut PlanRenderingContext<'_, Plan>,
+        ctx: &mut PlanRenderingContext<'_, LirRelationExpr>,
     ) -> fmt::Result {
         if ctx.config.verbose_syntax {
             self.fmt_verbose_text(f, ctx)
@@ -58,7 +60,7 @@ impl DisplayText<PlanRenderingContext<'_, Plan>> for Plan {
     }
 }
 
-impl Plan {
+impl LirRelationExpr {
     // NOTE: This code needs to be kept in sync with the `Display` instance for
     // `RenderPlan:ExprHumanizer`.
     //
@@ -67,9 +69,9 @@ impl Plan {
     fn fmt_default_text(
         &self,
         f: &mut fmt::Formatter<'_>,
-        ctx: &mut PlanRenderingContext<'_, Plan>,
+        ctx: &mut PlanRenderingContext<'_, LirRelationExpr>,
     ) -> fmt::Result {
-        use PlanNode::*;
+        use LirRelationNode::*;
 
         let mode = HumanizedExplain::new(ctx.config.redacted);
         let annotations = PlanAnnotations::new(ctx.config.clone(), self);
@@ -599,9 +601,9 @@ impl Plan {
     fn fmt_verbose_text(
         &self,
         f: &mut fmt::Formatter<'_>,
-        ctx: &mut PlanRenderingContext<'_, Plan>,
+        ctx: &mut PlanRenderingContext<'_, LirRelationExpr>,
     ) -> fmt::Result {
-        use PlanNode::*;
+        use LirRelationNode::*;
 
         let mode = HumanizedExplain::new(ctx.config.redacted);
         let annotations = PlanAnnotations::new(ctx.config.clone(), self);
@@ -1039,7 +1041,7 @@ fn fmt_mfp_default_text(
     mfp_plan: &MfpPlan<LirScalarExpr>,
     mode: &HumanizedExplain,
     f: &mut fmt::Formatter<'_>,
-    ctx: &mut PlanRenderingContext<'_, Plan>,
+    ctx: &mut PlanRenderingContext<'_, LirRelationExpr>,
 ) -> fmt::Result {
     mode.expr(mfp_plan.safe_mfp().deref(), None)
         .fmt_default_text(f, ctx)?;
@@ -1051,7 +1053,7 @@ fn fmt_mfp_verbose_text(
     mfp_plan: &MfpPlan<LirScalarExpr>,
     mode: &HumanizedExplain,
     f: &mut fmt::Formatter<'_>,
-    ctx: &mut PlanRenderingContext<'_, Plan>,
+    ctx: &mut PlanRenderingContext<'_, LirRelationExpr>,
 ) -> fmt::Result {
     mode.expr(mfp_plan.safe_mfp().deref(), None)
         .fmt_text(f, ctx)?;
@@ -1068,7 +1070,7 @@ fn fmt_temporal_bounds(
     mfp_plan: &MfpPlan<LirScalarExpr>,
     mode: &HumanizedExplain,
     f: &mut fmt::Formatter<'_>,
-    ctx: &mut PlanRenderingContext<'_, Plan>,
+    ctx: &mut PlanRenderingContext<'_, LirRelationExpr>,
 ) -> fmt::Result {
     let (_, lower_bounds, upper_bounds) = mfp_plan.as_parts();
 
@@ -1091,11 +1093,11 @@ fn fmt_temporal_bounds(
     Ok(())
 }
 
-impl DisplayText<PlanRenderingContext<'_, Plan>> for AvailableCollections {
+impl DisplayText<PlanRenderingContext<'_, LirRelationExpr>> for AvailableCollections {
     fn fmt_text(
         &self,
         f: &mut fmt::Formatter<'_>,
-        ctx: &mut PlanRenderingContext<'_, Plan>,
+        ctx: &mut PlanRenderingContext<'_, LirRelationExpr>,
     ) -> fmt::Result {
         if ctx.config.verbose_syntax {
             self.fmt_verbose_text(f, ctx)
@@ -1108,7 +1110,7 @@ impl AvailableCollections {
     fn fmt_default_text(
         &self,
         f: &mut fmt::Formatter<'_>,
-        ctx: &mut PlanRenderingContext<'_, Plan>,
+        ctx: &mut PlanRenderingContext<'_, LirRelationExpr>,
     ) -> fmt::Result {
         let plural = if self.arranged.len() == 1 { "" } else { "s" };
         write!(
@@ -1140,7 +1142,7 @@ impl AvailableCollections {
     fn fmt_verbose_text(
         &self,
         f: &mut fmt::Formatter<'_>,
-        ctx: &mut PlanRenderingContext<'_, Plan>,
+        ctx: &mut PlanRenderingContext<'_, LirRelationExpr>,
     ) -> fmt::Result {
         // raw field
         let raw = &self.raw;
@@ -1158,7 +1160,7 @@ impl AvailableCollections {
 
 /// Format a join implementation chain like `%0:t[#0{a}] » %1:u[#0{c}] » %2[×]`.
 ///
-/// Each position is rendered as `%pos:name` when an underlying [`PlanNode::Get`] can be
+/// Each position is rendered as `%pos:name` when an underlying [`LirRelationNode::Get`] can be
 /// dug out of the corresponding input plan (see [`humanize_input_name`]),
 /// otherwise just `%pos`. `[×]` (U+00D7) marks a cross product (empty lookup
 /// key). A `None` `source_key` renders the source position with no bracketed
@@ -1167,7 +1169,7 @@ fn fmt_join_chain<'a, I>(
     f: &mut fmt::Formatter<'_>,
     humanizer: &dyn ExprHumanizer,
     mode: &HumanizedExplain,
-    inputs: &[Plan],
+    inputs: &[LirRelationExpr],
     source_relation: usize,
     source_key: Option<&'a Vec<LirScalarExpr>>,
     stages: I,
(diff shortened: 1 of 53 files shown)


src/expr/src/linear.rs
```diff
@@ -1852,6 +1852,56 @@ pub mod plan {
...
+
+            let mz_now = MirScalarExpr::CallUnmaterializable(crate::UnmaterializableFunc::MzNow);
+
+            for lb in lower_bounds {
+                // mz_now() >= lb
+                let predicate = MirScalarExpr::CallBinary {
+                    func: crate::BinaryFunc::Gte(crate::func::Gte),
+                    expr1: Box::new(mz_now.clone()),
+                    expr2: Box::new(lb),
+                };
```
antiguru: Here and just below: Might be easier to read using `mz_now.clone().call_binary(crate::func::Gte, lb)`.

src/compute-types/src/plan/lowering.rs
```diff
@@ -1065,7 +1086,11 @@ This is not expected to cause incorrect results, but could indicate a performanc
                 // Determine keys that are not present in `input_keys`.
                 let new_keys = keys
                     .iter()
-                    .filter(|k1| !input_keys.arranged.iter().any(|(k2, _, _)| k1 == &k2))
+                    .filter(|k1| {
+                        !input_keys.arranged.iter().any(|(k2, _, _)| {
+                            *k1 == &k2.iter().map(MirScalarExpr::from).collect_vec()
```
antiguru: Could this be `k1.iter().eq(k2.iter().map(...))`? This might avoid the vector allocation.

src/compute-types/src/plan/scalar.rs
```diff
@@ -0,0 +1,745 @@
...
+
+    let lower = lower.iter().map(MirScalarExpr::from).collect();
+    let upper = upper.iter().map(MirScalarExpr::from).collect();
+    MfpPlan::from_parts(SafeMfpPlan::from_mfp(mir_mfp), lower, upper)
+}
+
+/// Translates a `&Vec<MirScalarExpr>` (or similar) to a `Vec<LirScalarExpr>`.
+///
+/// Panics when it encounters unmaterializable functions.
+pub(crate) fn try_lses_from_mses<'a>(
```
antiguru: I'd expect a `try_` function to return a result, not panic. Usually the non-try variant panics, the try variant returns an error.

src/compute-types/src/plan/threshold.rs
```diff
@@ -56,32 +57,33 @@ impl ThresholdPlan {
...
+    pub fn create_from(arity: usize) -> (Self, (Vec<LirScalarExpr>, Vec<usize>, Vec<usize>)) {
         // Arrange the input by all columns in order.
         let mut all_columns = Vec::new();
         for column in 0..arity {
-            all_columns.push(mz_expr::MirScalarExpr::column(column));
+            all_columns.push(LirScalarExpr::column(column));
         }
         let (permutation, thinning) = permutation_for_arrangement(&all_columns, arity);
-        let ensure_arrangement = (all_columns, permutation, thinning);
+        let ensure_arrangement = (all_columns, permutation.clone(), thinning.clone());
```
antiguru: No need to clone.

src/compute-types/src/plan/threshold.rs
```diff
@@ -56,32 +57,33 @@ impl ThresholdPlan {
...
         let plan = ThresholdPlan::Basic(BasicThresholdPlan {
             ensure_arrangement: ensure_arrangement.clone(),
         });
+
```
antiguru: Stray change.

antiguru (approved):
Looks good, delta some nits.

src/expr/src/linear.rs
antiguru: Done — both bounds now use the builder form: `mz_now.clone().call_binary(lb, crate::func::Gte)` and `…call_binary(ub, crate::func::Lt)`.

src/compute-types/src/plan/threshold.rs
antiguru: Done — `permutation`/`thinning` are moved into the tuple, clones dropped.

src/compute-types/src/plan/threshold.rs
antiguru: Done — removed the stray blank line.

src/compute-types/src/plan/lowering.rs
antiguru: Done, allocation removed. The exact `k1.iter().eq(k2.iter().map(MirScalarExpr::from))` does not compile: it needs `&MirScalarExpr: PartialEq<MirScalarExpr>`, which is not implemented, and `Iterator::eq_by` is still unstable (`iter_order_by`). So I kept it allocation-free with a length check plus `zip_eq`: `k1.len() == k2.len() && k1.iter().zip_eq(k2).all(|(e1, e2)| *e1 == MirScalarExpr::from(e2))`.

src/compute-types/src/plan/scalar.rs
antiguru: Renamed `try_lses_from_mses` -> `lses_from_mses` and kept the panic, with a doc comment stating the invariant (LIR-level expressions never contain unmaterializable functions, so the conversion is total; a panic indicates a lowering bug).

I considered making it genuinely fallible (`-> Result<Vec<LirScalarExpr>, Vec<UnmaterializableFunc>>` with `.expect(...)` at call sites), but every caller relies on the totality invariant and none could meaningfully propagate the error, so the rename is the better fit. The sibling helpers (`mfp_mir_to_lir`, `safe_mfp_mir_to_lir`, `mfp_plan_mir_to_lir`) already follow this convention — no `try_` prefix, panic on unmaterializable funcs — so they are left as-is for consistency.

src/compute-types/src/explain/text.rs
```diff
@@ -24,21 +24,24 @@ use std::fmt;
...
 use mz_expr::explain::{HumanizedExplain, HumanizerMode, fmt_text_constant_rows};
-use mz_expr::{Id, MirScalarExpr};
 use mz_ore::soft_assert_or_log;
 use mz_ore::str::{IndentLike, StrExt, separated};
 use mz_repr::explain::text::DisplayText;
 use mz_repr::explain::{
     CompactScalarSeq, CompactScalars, ExplainConfig, ExprHumanizer, Indices, PlanRenderingContext,
 };
 
+use mz_expr::MfpPlan;
```
ggevay: Stray empty line between this and the above block?

src/compute-types/src/plan/interpret/api.rs
```diff
@@ -18,17 +18,15 @@ use std::fmt::Debug;
...
 use mz_ore::cast::CastFrom;
 use mz_ore::stack::{CheckedRecursion, RecursionGuard, RecursionLimitError};
 use mz_ore::{assert_none, soft_panic_or_log};
 use mz_repr::{Diff, Row, Timestamp};
 
 use crate::plan::join::JoinPlan;
 use crate::plan::reduce::{KeyValPlan, ReducePlan};
+use crate::plan::scalar::LirScalarExpr;
 use crate::plan::threshold::ThresholdPlan;
 use crate::plan::top_k::TopKPlan;
```
ggevay, line 33: `TopKPlan` has some `limit` fields that still have `MirScalarExpr`. Do we want to convert those as well?

src/compute-types/src/plan/lowering.rs
```diff
@@ -248,8 +249,10 @@ impl Context {
                     .arranged
                     .iter()
                     .filter_map(|key| {
-                        mfp.literal_constraints(&key.0)
-                            .map(|val| (key.clone(), val))
+                        mfp.literal_constraints(
+                            &key.0.iter().map(MirScalarExpr::from).collect_vec(),
```
ggevay, line 255: Could we move this `literal_constraints` function from MIR to LIR? Then we wouldn't need to convert back from `LirScalarExpr` to `MirScalarExpr`. (Note that this is some super-old legacy code, from before the `LiteralConstraints` transform existed. It fires very rarely since the `LiteralConstraints` transform exists, but I don't dare remove it due to fear of sneaky regressions.)

src/compute/src/render/context.rs
```diff
@@ -923,10 +921,22 @@ impl<'scope, T: RenderTimestamp> CollectionBundle<'scope, T> {
...
         }
 
-        let max_demand = mfp.demand().last().map(|x| *x + 1).unwrap_or(0);
-        mfp.permute_fn(|c| c, max_demand);
-        mfp.optimize();
-        let mfp_plan = mfp.into_plan().unwrap();
+        // Apply demand-based column pruning. We round-trip through MIR
+        // so temporal bounds are folded back as mz_now() predicates —
+        // this way demand() sees all column references (including those
+        // in temporal bounds), and permute_fn applies uniformly.
```
ggevay, line 960: (As discussed before, it would be great to do a followup where we move such optimization stuff from rendering to some earlier phase.)

src/compute-types/src/plan.rs
```diff
@@ -628,20 +630,35 @@ impl Plan {
         fields(path.segment = "refine_source_mfps")
     )]
     fn refine_source_mfps(dataflow: &mut DataflowDescription<Self>) {
```
ggevay, line 631: Do we want to eventually make `SourceInstanceArguments` have `LirScalarExpr` instead of `MirScalarExpr`? All this back-and-forth conversion between `LirScalarExpr` and `MirScalarExpr` is making me a bit nervous.

ggevay: Also, out of scope for this PR, but is the `typ: SqlRelationType` in `SourceInstanceDesc` right?

ggevay:
(A related thought: Now that we'll have `LirScalarExpr`, how about we finally rename `Plan` to `LirRelationExpr`? This has been bugging me for a long time, but now it seems more relevant than ever.)

src/compute-types/src/plan.rs
```diff
@@ -660,20 +677,72 @@ impl Plan {
...
+                        })
+                        .collect()
                 } else {
-                    common
+                    taken.into_iter().collect()
                 };
-                mfp.optimize();
-                source.arguments.operators = Some(mfp);
+
+            // Second pass: put the MfpPlans back by LirId.
```
ggevay, line 716: From Claude:

Could we harden the put-back pass slightly? If a recorded `LirId` is ever not revisited here (e.g., a future edit changes one of the two parallel traversals but not the other), the node silently keeps the placeholder identity MFP. For filter-only source MFPs (the common case), input and output arity coincide, so the placeholder is a well-formed MFP that just drops the filter — silent wrong results.

Suggestion: use `replacements.remove(&expression.lir_id)` instead of `get` + `clone` (also saves the clone, since each id is matched at most once), and after the loop:

```rust
assert!(
    replacements.is_empty(),
    "refine_source_mfps failed to restore MFPs for {:?}",
    replacements.keys().collect::<Vec<_>>()
);
```

That turns silent plan corruption into a loud failure. (I'd make it a hard assert rather than `soft_panic_or_log`, since continuing with a corrupted plan is worse than failing finalization.)

src/compute-types/src/plan.rs
```diff
@@ -316,7 +318,7 @@ pub enum PlanNode {
         /// become undefined. Additionally, the MFP must be free from temporal
         /// predicates so that it can be readily evaluated.
         /// TODO(ggevay): should we wrap this in [`mz_expr::SafeMfpPlan`]?
-        mfp_after: MapFilterProject,
+        mfp_after: MfpPlan<LirScalarExpr>,
```
ggevay: Could we resolve the above todo, maybe as a followup PR? From Claude:

Make the Reduce mfp_after invariant structural. Render does m.into_nontemporal().expect("Fused Reduce MFPs do not have temporal predicates") (render/reduce.rs). Now that planning happens at lowering, PlanNode::Reduce::mfp_after could be SafeMfpPlan<LirScalarExpr> directly (the existing TODO(ggevay) in plan.rs asks for essentially this), failing at lowering instead of render. Fine as a follow-up PR. (FlatMap's mfp_after legitimately needs MfpPlan — it evaluates temporal bounds.)

src/compute-types/src/plan/lowering.rs
```diff
@@ -271,17 +274,28 @@ impl Context {
                     // present when handling the leftover MFP after this big match.)
                     mfp.permute_fn(|c| permutation[c], thinning.len() + key.len());
                     in_keys.arranged = vec![(key.clone(), permutation.clone(), thinning.clone())];
-                    GetPlan::Arrangement(key.clone(), Some(val.clone()), mfp)
+                    GetPlan::Arrangement(
+                        key.clone(),
+                        Some(val.clone()),
+                        mfp_plan_mir_to_lir(mfp.into_plan().expect("MFP planning failed")),
```
ggevay: Nit:
```
mfp_plan_mir_to_lir(mfp.into_plan().expect("MFP planning failed"))
```
appears 4 times. Maybe worth it to factor it out into a helper func?

src/compute-types/src/plan/join/delta_join.rs
```diff
@@ -92,6 +94,22 @@ impl DeltaJoinPlan {
...
+                            characteristics,
+                        )
+                    })
+                    .collect_vec()
+            })
+            .collect_vec();
+
         // Pick the "first" (by `Ord`) key for the source relation of each path.
         // (This matches the probably arbitrary historical practice from `mod render`.)
         // This needs to be kept in sync with `IndexUsageContext::add_keys`!
```
ggevay, line 115: I'm slightly worried about the

> This needs to be kept in sync with `IndexUsageContext::add_keys`

We'll now rely on the Ord of MirScalarExpr to be in sync with the Ord of LirScalarExpr, right? Anyhow, maybe the proper fix is to port over the whole `IndexUsageContext` machinery from MIR to LIR, but that's a sizable project on its own.

ggevay (approved):
LGTM overall, but wrote some comments. (Some of them could go into followup PRs.)

src/compute-types/src/plan/interpret/api.rs
mgree: Oof, good catch, thank you!

src/compute-types/src/plan.rs
mgree: Eventually, sure! But right now they're used as MIR throughout, which means, e.g., 
[`DataflowDescription<OptimizedMirRelationExpr, ()>`](https://github.com/mgree/materialize/blob/stable-lir-mvp/src/compute-types/src/dataflows.rs#L235-L256) expects MIR. Follow-up, to push LIR further up? Could maybe go with the catalog changes that do the pinning?

https://linear.app/materializeinc/issue/SQL-391/sourceinstanceargumentsare-currently-stored-as-mir-but-should-likely

src/compute-types/src/plan/lowering.rs
mgree: I think a better way forward is to be able to remove it! Maybe a follow-up PR that records a metric of when this is hit, and eventually moving it over to a soft assert, and then an assert? Tricky thing: we don't have any metrics handle where we are in lowering, but we could thread in an `OptimizerMetrics`.

https://linear.app/materializeinc/issue/SQL-393/mapfilterprojectliteral-constraints-should-be-deprecated

src/compute/src/render/context.rs
mgree: https://linear.app/materializeinc/issue/SQL-394/demand-optimization-and-sourcedataflow-integration-happen-too-late

src/compute-types/src/plan/join/delta_join.rs
mgree: We'll always have _some_ amount of keeping things in sync to do at that border, though we could stand to have less! Porting `IndexUsageContext` (and arrangement/index planning, and possibly join planning) to be at the MIR/LIR border sounds good, but is not a small task. 

https://linear.app/materializeinc/issue/SQL-394/demand-optimization-and-sourcedataflow-integration-happen-too-late

src/compute-types/src/plan.rs
mgree: https://linear.app/materializeinc/issue/SQL-395/index-planning-happens-in-mir

src/compute-types/src/plan.rs
mgree: Easy fix, tacked it on.

mgree:
> (A related thought: Now that we'll have `LirScalarExpr`, how about we finally rename `Plan` to `LirRelationExpr`? This has been bugging me for a long time, but now it seems more relevant than ever.)

Went ahead and did this, too! https://github.com/MaterializeInc/materialize/pull/36544/commits/984aefc542d587f894291d565e6d243d288c361b

src/compute-types/src/plan/join/delta_join.rs
ggevay: Ok to leave that to future work!

src/compute-types/src/plan/lowering.rs
ggevay: Yep, metric-and-then-remove sounds very good!

ggevay:
Thanks for addressing the comments, LGTM!

PR #36551, opened 2026-05-14
mcp: hint agents to prefer indexed objects

bobbyiliev:
As discussed adding a short instructions block to the MCP agent endpoint's `initialize` response, nudging agents to prefer indexed objects over materialized views. Pairs with the `indexedColumns` hint already surfaced by `get_data_product_details` so the agent has both session-level guidance and per-tool signals for picking the fast path.

### DIFF
--- a/src/environmentd/src/http/mcp.rs
+++ b/src/environmentd/src/http/mcp.rs
@@ -600,9 +600,10 @@ fn endpoint_instructions(endpoint_type: McpEndpointType) -> Option<String> {
     match endpoint_type {
         McpEndpointType::Agent => Some(concat!(
             "You have access to Materialize data products via MCP. ",
-            "Prefer indexed objects first (served from memory), then materialized views ",
-            "(read from persistent storage). Querying plain views, unindexed sources, or tables ",
-            "triggers query-time computation that can consume significant cluster resources, avoid these.",
+            "Prefer indexed objects (served from memory) over materialized views ",
+            "(read from persistent storage). Indexes are cluster-local; if a data product's ",
+            "cluster differs from your session, pass the `cluster` parameter to `read_data_product` ",
+            "so the index is actually used.",
         ).to_string()),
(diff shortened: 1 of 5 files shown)


jubrad:
@bobbyiliev I'm wondering if we need to be very explicit about either 
1. only showing indexes on the cluster the agent is connected to
2. stressing that indexes are only accessible when they are on the current cluster, and that if a view with an index is queried on a cluster that does not have the index the index will not be used.

bobbyiliev:
@jubrad Yea good call. I just updated the instructions to say indexes are cluster-local and to pass the cluster parameter to `read_data_product` when the agent's session is elsewhere. Docs/code mismatch on that param is tracked in DEX-27.

ggevay:
Do we want to also tweak the comment on `mz_mcp_data_product_details` that says
`The object can be read from any cluster.`?

src/environmentd/src/http/mcp.rs
```diff
@@ -613,7 +598,13 @@ async fn handle_mcp_method(
 /// These guide the AI agent on how to use the server correctly.
 fn endpoint_instructions(endpoint_type: McpEndpointType) -> Option<String> {
     match endpoint_type {
-        McpEndpointType::Agent => None,
+        McpEndpointType::Agent => Some(concat!(
+            "You have access to Materialize data products via MCP. ",
+            "Prefer indexed objects (served from memory) over materialized views ",
```
ggevay: > over materialized views

Note that materialized views can also be indexed. So, this could e.g. say
"over unindexed materialized views"

bobbyiliev: Thanks @ggevay! Both should be addressed now!

ggevay (approved):
Thank you!

jubrad (approved):

mtabebe (approved):

PR #36552, opened 2026-05-14
timely-util: column_pager with policy + lz4

antiguru:
### Motivation

Builds on #36391 (now merged), which introduced `mz_ore::pager`.
Layers a typed `Column` API over the pager with an injected policy that decides, per call, whether to keep resident, page out raw, or page out lz4-compressed.
This is the seam between in-memory columnar buffers and out-of-core storage envisioned in `doc/developer/design/20260504_pager.md`.

### Description

Adds `mz_ore::pager::pageout_with` for explicit-backend dispatch, bypassing the global atomic so layered consumers can route per call without racing other writers.

New `mz_timely_util::column_pager` module:

* `PagingPolicy` trait with `decide(PageHint) -> PageDecision` (`Skip` / `Page { backend, codec }`) and `record(PageEvent)` for budget bookkeeping and metrics.
* `PagedColumn` with three variants: `Resident(Column, ResidentTicket)`, `Paged { handle, meta }`, and `Compressed { inner, meta }` (memory or pager-backed framed bytes).
* `ResidentTicket` is a drop guard that fires `PageEvent::ResidentReleased { bytes }` whether the caller calls `ColumnPager::take` or drops the column without taking — so policy budgets don't leak.
* `ColumnPager::page` drains via `ContainerBytes`. The `Column::Align(Vec)` uncompressed path moves the body `Vec` directly into the pager handle; the other variants serialize and then `Column::clear` in place, retaining the caller's `Typed` allocation for reuse instead of dropping it. The compressed path wraps the target in `FrameEncoder` so `into_bytes` streams straight through lz4 — no intermediate uncompressed `Vec`. The compressed File path pads to u64 alignment; the frame trailer self-delimits so no length prefix or unpad is needed on read.

`TieredPolicy` in `column_pager::policy` is a single process-wide byte budget.
Each `decide` reserves `len_bytes` from one `AtomicUsize` pool — success keeps the column resident, failure spills it via the configured `backend` + `codec` — and `record` credits the bytes back on `ResidentReleased`.
Accounting is a single atomic rather than per-worker `thread_local!` state because resident columns move freely between Timely workers, so a release can run on a different thread than the grant; one pool credits correctly regardless of where the drop lands.
A future thread-aware tiered layout would need a `SendColumn` wrapper that pins the column to its origin thread, or an origin-keyed cross-thread credit channel.

Criterion bench (`cargo bench -p mz-timely-util --bench column_pager`) covers 4 KiB / 256 KiB / 4 MiB × Swap / File × raw / lz4 for both round-trip and operator-loop shapes.
Swap-backend results are labelled `swap-warm` to flag that the bench never builds enough working set to force kernel eviction; a follow-up `column_pager_pressure` bench under `systemd-run --user --scope -p MemoryMax=...` will exercise the cold path.

### Checklist

- [ ] This PR has adequate test coverage / QA is not needed.
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.


src/ore/src/pager/file.rs
```diff
@@ -0,0 +1,380 @@
...
+            continue;
+        };
+        let pid: u32 = match rest.split_once('-').and_then(|(p, _)| p.parse().ok()) {
+            Some(p) => p,
+            None => continue,
+        };
+        if pid == std::process::id() {
+            continue;
+        }
+        if std::path::Path::new(&format!("/proc/{pid}")).exists() {
```
def-: What if you're on macOS?

src/timely-util/src/column_pager.rs
```diff
@@ -0,0 +1,553 @@
...
+                ..
+            }
+        ));
+        let rt = cp.take(paged);
+        assert_eq!(collect_i64(&rt), (0i64..1024).collect::<Vec<_>>());
+    }
+
+    #[mz_ore::test]
+    fn round_trip_file_uncompressed() {
+        let dir = tempfile::tempdir().unwrap();
```
def-: Tests share global `BACKEND` and `SCRATCH_DIR` state; parallel cargo test can race

doc/developer/design/20260504_pager.md
```diff
@@ -0,0 +1,411 @@
...
+
+`Drop` (without `take`).
+`unlink(path)`. The kernel reclaims the inode.
+
+### Scratch directory lifecycle
+
+`set_scratch_dir(root)` creates `{root}/mz-pager-{pid}-{boot_nonce}/`.
+`boot_nonce` is a random 64-bit value sampled at config time so two processes that briefly share the same pid cannot collide.
+On `set_scratch_dir`, the pager runs a reaper that walks `{root}` and removes any sibling `mz-pager-*` subdirectory whose owning pid is no longer alive (`/proc/{pid}` missing on Linux); this reclaims storage from crashed predecessors.
+On clean process exit, a `Drop` on the global pager state removes the per-process subdirectory.
```
def-: Where is that done?

src/ore/src/pager/file.rs
```diff
@@ -0,0 +1,380 @@
...
+static SCRATCH_DIR: OnceLock<PathBuf> = OnceLock::new();
+static SUBDIR: OnceLock<PathBuf> = OnceLock::new();
+static SCRATCH_ID: AtomicU64 = AtomicU64::new(0);
+static SCRATCH_INIT: Once = Once::new();
+
+/// Configures the scratch directory for the file backend. Idempotent across multiple
+/// calls with the same path; logs and ignores subsequent calls with a different path.
+pub fn set_scratch_dir(root: PathBuf) {
+    SCRATCH_INIT.call_once(|| {
+        if let Err(err) = init_subdir(&root) {
```
def-: We might not want to just ignore this error since it will cause follow-up failures and we can only figure them out from the warning then?

def- (commented):
Should we document that you can't do a cross-thread drop of `PagedColumn::Resident` since it would corrupt `TieredPolicy` thread-local accounting?

src/timely-util/src/column_pager.rs
```diff
@@ -0,0 +1,572 @@
...
+use lz4_flex::frame::{FrameDecoder, FrameEncoder};
+use mz_ore::pager::{self, Backend, Handle};
+use timely::bytes::arc::BytesMut;
+use timely::dataflow::channels::ContainerBytes;
+
+use crate::columnar::Column;
+
+// ---------------------------------------------------------------------------
+// Codec
+// ---------------------------------------------------------------------------
```
antiguru: Remove the headers.

src/timely-util/src/column_pager.rs
```diff
@@ -0,0 +1,572 @@
...
+            Some(Codec::Lz4) => {
+                // Stream serialized bytes straight into lz4 — no intermediate
+                // uncompressed `Vec<u8>`.
+                let mut out = Vec::with_capacity(len_bytes / 4);
+                {
+                    let mut enc = FrameEncoder::new(&mut out);
+                    col.into_bytes(&mut enc);
+                    enc.finish().expect("lz4 finish into Vec is infallible");
+                }
+                *col = Column::default();
```
antiguru: We might be dropping the allocation here, don't if not needed.

src/timely-util/src/column_pager.rs
```diff
@@ -0,0 +1,572 @@
...
+                self.policy.record(PageEvent::PagedOut {
+                    bytes_in: len_bytes,
+                    bytes_out: out.len(),
+                    backend,
+                    codec: Some(Codec::Lz4),
+                });
+                let inner = match backend {
+                    Backend::Swap => CompressedInner::Memory(out),
+                    Backend::File => {
+                        let padded = pad_u8_to_u64(out);
```
antiguru, line 309: Look into whether we actually need this, because it's a potential reallocation.

src/timely-util/src/column_pager.rs
antiguru: Done — removed the ASCII section-divider header blocks.

src/timely-util/src/column_pager.rs
antiguru: Fixed: added `Column::clear` (capacity-retaining for the `Typed` variant via `columnar::Clear`) and call it here instead of dropping, so a `Typed` caller keeps its buffer for reuse. The raw path does the same for `Typed`/`Bytes` and still moves `Align` zero-copy into the pager.

src/timely-util/src/column_pager.rs
antiguru: Kept the copy, documented why: the pager's currency is `Vec<u64>` and `out` is already lz4-compressed (~4x smaller), so this copies the small form. Avoiding it would need a byte-oriented pager entry point — not worth widening that surface.

src/timely-util/src/column_pager.rs
antiguru: Addressed: tests route through `pageout_with` (explicit backend, no global `BACKEND` mutation) and share a single scratch dir, so parallel `cargo test` doesn't race.

src/ore/src/pager/file.rs
antiguru: The file backend landed separately via #36391; after rebasing onto main this PR no longer touches `file.rs`.

src/ore/src/pager/file.rs
antiguru: The file backend landed separately via #36391; after rebasing onto main this PR no longer touches `file.rs`.

doc/developer/design/20260504_pager.md
antiguru: The pager design doc landed via #36391; after rebasing onto main this PR no longer modifies it.

DAlperin (approved):
LGTM, thanks

PR #36553, opened 2026-05-14
Revert #36537 and #36514

antiguru:
### Motivation

Reverts #36537. Caused regression tracked in SQL-299. Forward fix in #36548 will take more time; reverting in the interim to unblock.

Also reverts #36514 (container builders for the persist source) because it was merged on top of #36537 and propagates the same generalization through `txn-wal::operator::txns_progress` / `txns_progress_frontiers`, so #36537 can't be reverted in isolation. Once #36548 lands, both #36514 and #36537 can be re-applied.

### Description

* `Revert "Container builders for the persist source (#36514)"`
* `Revert "txn-wal: deasync txns_progress_frontiers + retain remap across close (#36537)"`

Pure `git revert`, no manual edits.

### Verification

`cargo check -p mz-txn-wal -p mz-persist-client -p mz-storage` clean.

### DIFF
--- a/src/compute/src/render.rs
+++ b/src/compute/src/render.rs
@@ -268,25 +268,22 @@ pub fn build_compute_dataflow(
 
                     // Note: For correctness, we require that sources only emit times advanced by
                     // `dataflow.as_of`. `persist_source` is documented to provide this guarantee.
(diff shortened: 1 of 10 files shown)


SangJunBak (approved):

antiguru:
Thanks for the review!

PR #36555, opened 2026-05-14
sql: keep sql_impl resolved IDs separate from statement dependencies SQL-296

jubrad:
## Summary

Fixes the regression from #36480 where views using SQL-implemented functions (`has_table_privilege`, `has_cluster_privilege`, etc.) were rejected with "cannot create view with unstable dependencies."

- `plan()` now returns `(Plan, ResolvedIds)` where the second element contains IDs discovered inside `sql_impl` function bodies
- These are threaded separately through `plan_statement` → `sequence_plan` → `check_plan` → `check_restrict_to_user_objects`
- The main `resolved_ids` used for dependency tracking is never polluted with internal function implementation details
- Added regression test verifying `CREATE VIEW ... has_table_privilege(...)` succeeds

Fixes SQL-296.

### DIFF
--- a/src/adapter/src/catalog/state.rs
+++ b/src/adapter/src/catalog/state.rs
@@ -1009,14 +1009,9 @@ impl CatalogState {
             let session_catalog = state.for_system_session();
 
             let stmt = mz_sql::parse::parse(create_sql)?.into_element().ast;
-            let (stmt, mut resolved_ids) = mz_sql::names::resolve(&session_catalog, stmt)?;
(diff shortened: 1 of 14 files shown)


mtabebe (approved):
This is how I was planning on doing this. Thanks for fixing

jubrad:
> This is how I was planning on doing this. Thanks for fixing
Yeah, the prompt was "pretend you're @mtabebe, fix this issue"

PR #36559, opened 2026-05-15
SS-111 Fix kafka source hanging forever when offset 0 doesn't exist

patrickwwbutler:
Fixes https://linear.app/materializeinc/issue/SS-111/kafka-topics-after-retention-or-compaction-can-become-unavailable

Makes a call to fetch low water marks for all partitions in the topic, and compares against given `start_offsets`, which are populated either by specific `resume_uppers`, which indicates that this source is rehydrating, or by the start offset option when the user creates a kafka source. In either case, if a start offset exists for a partition, and the low watermark is past it, then there is a gap in the data that we can never retrieve, meaning the source is broken.

If there is no `start_offset` for this partition, then this is a new source that we are initially hydrating, and we assume that the user wants to just start from the earliest possible offset, and emit a warning message.


src/storage/src/source/kafka.rs
```diff
@@ -137,10 +137,11 @@ struct PartitionCapability {
     data: Capability<KafkaTimestamp>,
 }
 
-/// The high watermark offsets of a Kafka partition.
+/// The high/low watermark offsets of a Kafka partition.
 ///
-/// This is the offset of the latest message in the topic/partition available for consumption + 1.
-type HighWatermark = u64;
+/// This is the offset of either the first available or the latest message in the topic/partition
```
martykulma, line 142: nit: I think it would be clearer to just label this as `Offset`, as the meaning is entirely local to the function/data structure.

src/storage/src/source/kafka.rs
```diff
@@ -460,6 +413,177 @@ fn render_reader<'scope>(
...
+            let low_watermarks = fetch_partition_info(
+                &consumer,
+                topic.as_str(),
+                config
+                        .config
+                        .parameters
+                        .kafka_timeout_config
+                        .fetch_metadata_timeout,
+                false, // fetch the low watermark
+            ).unwrap_or_else(|e| {
```
martykulma, line 449: It loos like this can fail when the topic doesn't exist, which should emit a definite error.

src/storage/src/source/kafka.rs
```diff
@@ -460,6 +413,177 @@ fn render_reader<'scope>(
...
+                // is no start offset, then we set it to the low watermark and start consuming from
+                // there, assuming the user doesn't care about the messages that have been compacted away.
+                if let Some(start_offset) = start_offsets.get_mut(pid) {
+                    tracing::info!(
+                        source_id = config.id.to_string(),
+                        worker_id = config.worker_id,
+                        num_workers = config.worker_count,
+                        "restored offset {start_offset} for topic {topic} partition {pid} with low watermark {lwm}"
+                    );
+                    if *lwm > ReinterpretCast::reinterpret_cast(*start_offset) {
```
martykulma: wow - I wonder how many of these we have. `ReinterpretCast::reinterpret_cast` is new to me. We use u64::cast_from (also in mz-ore) all over the place, so we probably prefer the latter.

Rather than casting within the body where we initialize the starts, maybe it would make more sense to cast in 2 discrete points - start_offsets is filtered, convert to u64, then do the checks, and convert everything back to i64 before passing to `KafkaSourceReader`. I would opt for `u64/i64::try_into().expect()` as the successful conversion is an invariant and this is not a hot path.

src/storage/src/source/kafka.rs
```diff
@@ -460,6 +413,177 @@ fn render_reader<'scope>(
                 }
             };
 
+            // Start offsets is a map from partition to the next offset to read from.
+            let mut start_offsets: BTreeMap<_, i64> = start_offsets
+                .clone()
+                .into_iter()
+                .filter(|(pid, _offset)| responsible_for_pid(&config, *pid))
+                .map(|(k, v)| (k, v))
```
martykulma: probably don't need this?

src/storage/src/source/kafka.rs
```diff
@@ -460,6 +413,177 @@ fn render_reader<'scope>(
...
+
+            let low_watermarks = fetch_partition_info(
+                &consumer,
+                topic.as_str(),
+                config
+                        .config
+                        .parameters
+                        .kafka_timeout_config
+                        .fetch_metadata_timeout,
+                false, // fetch the low watermark
```
martykulma: It would be clearer to pass `Offset` here so you don't need comments on why true/false.

martykulma (commented):
nice!

src/storage/src/source/kafka.rs
patrickwwbutler: yeah probably not? this is how it was before, but it doesn't seem to do anything? guess I'll delete

src/storage/src/source/kafka.rs
patrickwwbutler: so `Offset` is already used in this context for the rdkafka enum type. Personally I feel like `PartitionWatermark` explains the meaning and purpose pretty well, though? I suppose to be extra clear it could be an enum, but I feel like that might be overkill

src/storage/src/source/kafka.rs
patrickwwbutler: Ah yeah, `try_into().expect()` seems like the right move here. 

unfortunately `u64::cast_from(i64)` isn't defined, reinterpret is actually the only cast defined for these types in `mz_ore`, I suspect because it is a spicy cast that can fail?

martykulma: yeah, I read through the macros too quickly (got my from and to's mixed up). That makes sense that there wouldn't be a i64 -> u64 because of the neg. values.

src/storage/src/source/kafka.rs
```diff
@@ -460,6 +413,192 @@ fn render_reader<'scope>(
...
+            let future_ts =
+                Partitioned::new_range(lower, RangeBound::PosInfinity, MzOffset::from(0));
+            data_cap.downgrade(&future_ts);
+
+            for (pid, lwm) in &low_watermarks {
+                // If a start offset exists for this partition, then either the user specified it
+                // or we restored it from the resume upper. In either case, if the low watermark is
+                // greater than the start offset, we know for certain that the offset we need to
+                // start at has been compacted away or dropped from retention by Kafka. If there
+                // is no start offset, then we set it to the low watermark and start consuming from
```
DAlperin, line 535: tracks

ci/test/pipeline.template.yml
```diff
@@ -448,6 +448,17 @@ steps:
         agents:
           queue: hetzner-aarch64-4cpu-8gb
 
+      - id: kafka-low-watermark
```
DAlperin, line 451: not sure if this best lives here or nightly but we can move in a follow up

DAlperin (approved):
thanks!

PR #36560, opened 2026-05-15
[Source Versioning] Set `enable_create_table_from_source` to default true

patrickwwbutler:
As mentioned in team meeting, we should set this to true for self-managed.


martykulma (approved):
🦈

PR #36565, opened 2026-05-15
Reorganize `MSE::reduce`

frankmcsherry:
Largely Claude-code driven refactor of `MirScalarExpr::reduce`, a hundreds-of-lines-long wall of conditional logic. Broken down into constituent parts, with the ideal structure of `match` statements rather than chained `if/else` blocks. Some understanding of the fall-out of the conditions, and no *expected* errors, but we'll need to check the CI runs.

cc: @mgree, @ggevay


src/expr/src/scalar/reduce/if_then.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+        return;
+    }
+
+    // Lift a common operator out of the two `If` branches. For example,
+    // `IF cond THEN x = y ELSE x = z` becomes `x = IF cond THEN y ELSE z`.
+    //
+    // We must ensure that the two `If` branches have unionable types,
+    // otherwise the lifted `If` could not be typed. See
+    // https://github.com/MaterializeInc/database-issues/issues/9182.
```
ggevay: We lost a bit of comment content here. The original comment was:
```
// We have to also make sure that the expressions that will end up in
// the two `If` branches have unionable types. Otherwise, the `If` could
// not be typed by `typ`. An example where this could cause an issue is
// when pulling out `cast_jsonbable_to_jsonb`, which accepts a wide
// range of input types. (In theory, we could still do the optimization
// in this case by inserting appropriate casts, but this corner case is
// not worth the complication for now.)
// See https://github.com/MaterializeInc/database-issues/issues/9182
```

src/expr/src/scalar/reduce/binary.rs
```diff
@@ -0,0 +1,323 @@
...
+}
+
+/// Decomposes equality with `RecordCreate` on the right-hand side.
+///
+/// Handles two cases:
+/// - `Literal([c1, ...]) = record_create(e1, ...)` → `c1 = e1 AND ...`
+/// - `record_create(a1, ...) = record_create(b1, ...)` → `a1 = b1 AND ...`
+///
+/// `MapFilterProject::literal_constraints` relies on the first transform,
+/// because `(e1,e2) IN ((1,2))` is desugared using `record_create`.
```
ggevay, line 180: A bit of comment content loss also here. From Claude:

Old (line 429–435 of old reduce.rs), for the `record_create(a1,...) = record_create(b1,...)` case:
```
// This is similar to the previous reduction, but this one kicks in also
// when only some (or none) of the record fields are literals. This
// enables the discovery of literal constraints for those fields.
```
New (binary.rs:172–179): the docstring covers both cases as a bulleted list and only mentions `MapFilterProject::literal_constraints` reliance for the first case. The motivation for the second case ("kicks in also when only some/none are literal → enables literal-constraint discovery on those fields") is no longer stated anywhere. This is a non-obvious why — the second case isn't just "the same but for two record_creates", it exists specifically to expose partial literal patterns to later passes. Worth restoring as one extra sentence in the docstring or as an inline comment on the second match arm.

ggevay (approved):
Looks great!

A small thing is that Claude has a tendency to lose code comments during refactorings, so I've specifically instructed my agent to check for these, and it did find 2 occurrences, see below.

antiguru (approved):
Seems good!

src/expr/src/scalar/reduce/binary.rs
frankmcsherry: Should be back in now!

src/expr/src/scalar/reduce/if_then.rs
frankmcsherry: Should be back in now!

frankmcsherry:
Thanks for the reviews folks!

PR #36566, opened 2026-05-15
mcp: add Prometheus metrics for requests, tool calls, and tool call durations

bobbyiliev:
Fixes https://linear.app/materializeinc/issue/DEX-2/mcp-add-observability-metrics

### DIFF
--- a/src/environmentd/src/http/mcp.rs
+++ b/src/environmentd/src/http/mcp.rs
@@ -507,9 +507,10 @@ async fn handle_mcp_request(
 
     let request_id = request.id.clone().unwrap_or(serde_json::Value::Null);
 
-    // Spawn task for fault isolation, with a timeout safety net. The inner
-    // function returns the response paired with a status label so that
-    // requests_total is recorded with the actual outcome.
+    // Spawn task for fault isolation, with a timeout safety net.
+    // `abort_on_drop` propagates the timeout to the task itself; without
+    // it the task orphans and the SQL query keeps running in the
+    // background after the client gives up.
     let metrics_inner = metrics.clone();
     let result = tokio::time::timeout(
         MCP_REQUEST_TIMEOUT,
@@ -523,7 +524,8 @@ async fn handle_mcp_request(
                 metrics_inner,
             )
             .await
-        }),
+        })
+        .abort_on_drop(),
     )
     .await;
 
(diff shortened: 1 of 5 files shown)


src/environmentd/src/http/mcp_metrics.rs
```diff
@@ -0,0 +1,169 @@
...
+            .inc();
+
+        metrics
+            .tool_call_duration
+            .with_label_values(&["agent", "read_data_product"])
+            .observe(0.123);
+
+        let gathered = registry.gather();
+
+        // requests_total: 3 metrics with distinct label sets.
```
jubrad: should this be 2?
requests.get_metric().len(), 2);

src/environmentd/src/http/mcp.rs
```diff
@@ -445,6 +466,7 @@ async fn handle_mcp_request(
     };
     if !enabled {
         debug!(endpoint = %endpoint_type, "MCP endpoint disabled by feature flag");
+        record_request("EndpointDisabled");
```
jubrad: can we use snake case here? endpoint_disabled

src/environmentd/src/http/mcp.rs
```diff
@@ -519,10 +546,12 @@ async fn handle_mcp_request(
                     ))
                     .into(),
                 ),
-            }
+            };
+            (response, "Timeout")
```
jubrad: I think `timeout` here would be more appropriate case matching

src/environmentd/src/http/mcp.rs
```diff
@@ -799,7 +847,19 @@ async fn handle_tools_call(
...
+
+    let status_label = match &result {
+        Ok(_) => "ok",
+        Err(e) => e.error_type(),
+    };
+    metrics
+        .tool_calls
+        .with_label_values(&[endpoint_label, &tool_label, status_label])
+        .inc();
+    timer.observe_duration();
```
jubrad: We should probably just record this at the beginning of the tool call so it doesn't get lost if we timeout?

jubrad (commented):
I think this looks good, a few nits, and maybe one issue where our counts could be inaccurate during timeouts

src/environmentd/src/http/mcp_metrics.rs
bobbyiliev: Good catch. 2 is correct, the comment above was misleading. Fixed.

src/environmentd/src/http/mcp.rs
bobbyiliev: Just updated this. But worth noting that `the error_type()` values from `McpRequestError` stay PascalCase since they're already part of the JSON-RPC error wire format, so the status label is mixed-case overall.

src/environmentd/src/http/mcp.rs
bobbyiliev: Updated!

src/environmentd/src/http/mcp.rs
bobbyiliev: Good point. Just updated this!

jubrad (approved):

PR #36567, opened 2026-05-15
adapter: avoid panic when MV target replica is concurrently dropped

antiguru:
## Summary

Fixes [SQL-304](https://linear.app/materializeinc/issue/SQL-304/thread-coordinator-panicked-at-srcadaptersrccatalogapplyrs122933).

If `ALTER CLUSTER ... SET (REPLICATION FACTOR ...)` drops the replica targeted by an in-flight `CREATE MATERIALIZED VIEW ... REPLICA <name>` while the optimizer is running off-thread, the catalog transaction later panics inside `apply_item_update` (`src/adapter/src/catalog/apply.rs:1229`) because re-parsing the persisted SQL fails with `UnknownClusterReplica`.

Two commits, belt-and-suspenders:

1. **`create_materialized_view_finish`** re-checks the target cluster and replica immediately before building the catalog ops and returns `AdapterError::ChangedPlan` if either has been dropped since planning. This is the tight backstop against a racing non-serialized DDL landing during finish's own awaits.
2. **`create_materialized_view_validate`** now constructs a real `PlanValidity::Checks` (cluster_id, target_replica, resolved dependency IDs, role metadata) instead of the no-op `PlanValidity::RequireRevision`. `sequence_staged` already calls `validity.check(catalog)` at every off-thread → on-thread hop, so this catches concurrent drops between the optimize and finish stages — and as a bonus also catches dropped dependencies and dropped roles for this path.

## Test plan

- [ ] Run the reproducer from SQL-304 (concurrent `CREATE MATERIALIZED VIEW ... IN CLUSTER c REPLICA r2` workers + `ALTER CLUSTER c SET (REPLICATION FACTOR ...)` flipper) and confirm `environmentd` no longer panics; failed creates surface as `ChangedPlan` errors instead.
- [ ] CI

---
_Generated by [Claude Code](https://claude.ai/code/session_015etea7ewkYfwNQvP3JnmE5)_

### DIFF
--- a/src/adapter/src/coord/sequencer/inner/create_materialized_view.rs
+++ b/src/adapter/src/coord/sequencer/inner/create_materialized_view.rs
@@ -329,6 +329,7 @@ impl Coordinator {
                 plan::MaterializedView {
                     expr,
                     cluster_id,
+                    target_replica,
                     refresh_schedule,
                     ..
                 },
@@ -359,8 +360,18 @@ impl Coordinator {
             });
         }
 
-        let validity =
-            PlanValidity::require_transient_revision(self.catalog().transient_revision());
+        // Track the target cluster/replica and resolved dependencies so that
+        // concurrent drops (e.g. `ALTER CLUSTER ... SET (REPLICATION FACTOR
+        // ...)` racing with the off-thread optimizer) are caught between
+        // stages instead of panicking later when the persisted SQL is
+        // re-parsed during catalog application.
+        let validity = PlanValidity::new(
+            self.catalog().transient_revision(),
+            resolved_ids.items().copied().collect(),
+            Some(*cluster_id),
+            *target_replica,
+            session.role_metadata().clone(),
+        );
 
         // Check whether we can read all inputs at all the REFRESH AT times.
         if let Some(refresh_schedule) = refresh_schedule {
--- a/src/adapter/src/coord/sequencer/inner/create_index.rs
+++ b/src/adapter/src/coord/sequencer/inner/create_index.rs
@@ -20,6 +20,7 @@ use mz_sql::ast::ExplainStage;
 use mz_sql::catalog::CatalogError;
 use mz_sql::names::ResolvedIds;
 use mz_sql::plan;
+use mz_sql::session::metadata::SessionMetadata;
 use tracing::Span;
 
 use crate::catalog::CatalogState;
@@ -84,7 +85,7 @@ impl Coordinator {
         resolved_ids: ResolvedIds,
     ) {
         let stage = return_if_err!(
-            self.create_index_validate(plan, resolved_ids, ExplainContext::None),
+            self.create_index_validate(ctx.session(), plan, resolved_ids, ExplainContext::None),
             ctx
         );
         self.sequence_staged(ctx, Span::current(), stage).await;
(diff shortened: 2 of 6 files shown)


src/adapter/src/coord/sequencer/inner/create_materialized_view.rs
```diff
@@ -600,6 +600,25 @@ impl Coordinator {
...
+        // `CREATE MATERIALIZED VIEW ... REPLICA "<name>"` SQL is re-parsed
+        // during catalog application.
+        let Some(cluster) = self.catalog().try_get_cluster(cluster_id) else {
+            return Err(AdapterError::ChangedPlan(format!(
+                "cluster {cluster_id} was concurrently dropped"
+            )));
+        };
+        if let Some(replica_id) = target_replica {
+            if cluster.replica(replica_id).is_none() {
+                return Err(AdapterError::ChangedPlan(format!(
```
ggevay: Could you please use `ConcurrentDependencyDrop` instead of `ChangedPlan`? (One more occurrence above.)

src/adapter/src/coord/sequencer/inner/create_materialized_view.rs
```diff
@@ -359,8 +360,18 @@ impl Coordinator {
             });
         }
 
-        let validity =
-            PlanValidity::require_transient_revision(self.catalog().transient_revision());
```
ggevay, line 363: From the commit msg:

> `PlanValidity::RequireRevision`, which `sequence_staged` treats as a
no-op.

Wow, good catch! This sounds very dangerous! Could you please at least add a comment on `RequireRevision` that calls this out?

Edit: Oh, I see there is a commnet on `require_transient_revision`. But I'd say also having a comment on `RequireRevision` would be good.

ggevay: Also, does CREATE INDEX and CREATE VIEW have the same problem? They are also calling `require_transient_revision`.

ggevay: Actually, why not just remove the whole `RequireRevision`? It seems the only thing it does is blow people's feet off, and 0 useful things. (I guess the idea made some sense back then when it was created 2 years ago, when the Coordinator was more single-threaded than nowadays.)

ggevay (commented):
Nice catch! Wrote some comments.

Also, do I understand correctly that the second commit should completely cover the thing that the first commit checks? If yes, then could we remove the first commit?

antiguru:
Thanks for the review! Force-pushed addressing all four points:

1. **First commit dropped.** Agreed the suspenders cover it — `sequence_staged`'s `validity.check` runs at every off-thread → on-thread hop, so the explicit re-check in `finish` was redundant.
2. **`ConcurrentDependencyDrop` instead of `ChangedPlan`.** Switched in `PlanValidity::check` for cluster/replica/dependency drops. This also addresses the TODO in `error.rs:61`.
3. **CREATE INDEX and CREATE VIEW had the same problem.** Yes — both used `require_transient_revision`. Fixed in the same commit: `create_index_validate` now tracks the target cluster + resolved deps + role metadata; `create_view_validate` tracks resolved deps + role metadata.
4. **`RequireRevision` removed entirely.** `PlanValidity` is now a plain struct (just the former `Checks` fields). No more footgun.

Final state on the branch: one commit with the cross-cutting refactor + MV/View/Index fixes, plus the regression test for the original SQL-304 reproducer.

---

ggevay (approved):
Thank you!

PR #36568, opened 2026-05-15
[sql] fix restrict_to_user_objects bypasses in SHOW COLUMNS and EXPLAIN ANALYZE

mtabebe:
Problem:
SHOW COLUMNS FROM <user_table> and EXPLAIN ANALYZE were bypassing the restrict_to_user_objects check because their inner query resolved IDs were never propagated into sql_impl_resolved_ids.

Solution:
- Add restrict_to_user_objects to the SessionCatalog trait so the planner can query the current session's restriction state during planning.
- Fix show_columns() to call record_sql_impl_ids() after building the inner ShowSelect, so mz_show_columns reaches check_restrict_to_user_objects.
- Fix plan_explain_analyze_object() to perform an owner check first, then propagate the inner query IDs.
- Fix plan_explain_analyze_cluster() to propagate its inner query IDs.

Testing:
- Add regression tests
- Negative tests verify the restriction fires
- Positive test verifies SHOW COLUMNS still works for non-restricted sessions

Fixes SQL-301.

### DIFF
--- a/src/sql/src/plan/statement/dml.rs
+++ b/src/sql/src/plan/statement/dml.rs
@@ -891,12 +891,22 @@ pub fn plan_explain_analyze_object(
         .full_name_str();
     let explainee = plan_explainee(scx, statement.explainee, params)?;
 
-    match explainee {
-        plan::Explainee::Index(_index_id) => (),
-        plan::Explainee::MaterializedView(_item_id) => (),
-        _ => {
-            return Err(sql_err!("EXPLAIN ANALYZE queries for this explainee type",));
+    let check_ownership = |item_id: &CatalogItemId, item_type: &str| -> Result<(), PlanError> {
+        if scx.catalog.restrict_to_user_objects() {
+            let item = scx.catalog.get_item(item_id);
+            if item.owner_id() != *scx.catalog.active_role_id() {
+                let full_name = scx.catalog.resolve_full_name(item.name());
+                return Err(sql_err!("must be owner of {item_type} {full_name}"));
+            }
         }
+        Ok(())
+    };
+    match &explainee {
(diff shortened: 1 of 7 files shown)


ggevay:
Could you please add a centralized explanation to `RESTRICT_TO_USER_OBJECTS` of all the things that this does? There have been several PRs that added handling for various corner cases distributed in many places in the code; it would be good to have a single code comment to read when trying to figure out what exactly `RESTRICT_TO_USER_OBJECTS` does. (And/or maybe also its user-facing description could be extended a bit.)

src/sql/src/plan/statement/dml.rs
```diff
@@ -891,12 +891,22 @@ pub fn plan_explain_analyze_object(
...
+    let check_ownership = |item_id: &CatalogItemId, item_type: &str| -> Result<(), PlanError> {
+        if scx.catalog.restrict_to_user_objects() {
+            let item = scx.catalog.get_item(item_id);
+            if item.owner_id() != *scx.catalog.active_role_id() {
+                let full_name = scx.catalog.resolve_full_name(item.name());
+                return Err(sql_err!("must be owner of {item_type} {full_name}"));
+            }
+        }
+        Ok(())
+    };
```
jubrad, line 903: I think this will work, but am I reading it correctly that this this is restricting explain to only objects that the individual is the owner of? If so, largely this seems fine.. I feel like there's an alternative where we just don't allow explain or show commands, but that's really putting a lot into this falg.

jubrad (approved):

src/sql/src/plan/statement/dml.rs
mtabebe: Yes that is the correct interpretation.

I also don't want to restrict the allow or explain, since I can imagine that it would be helpful to the agent.

PR #36569, opened 2026-05-15
Reapply "[docs] document restrict_to_user_objects for MCP agent isolation (#36563)

mtabebe:
Add a subsection to the MCP agent RBAC docs explaining how a superuser can use `ALTER ROLE ... SET restrict_to_user_objects = true` to prevent an agent role from reading system catalog tables while keeping data product discovery working normally.

### DIFF
--- a/doc/user/content/integrations/mcp-server/mcp-agent-tools.md
+++ b/doc/user/content/integrations/mcp-server/mcp-agent-tools.md
@@ -13,7 +13,7 @@ menu:
 
 ### `get_data_products`
 
-Discover all available data products. Returns a lightweight list with name,
+Discovers all available data products. Returns a lightweight list with name,
 cluster, and description for each product.
 
 **Parameters:** None.
@@ -38,7 +38,7 @@ cluster, and description for each product.
 
 ### `get_data_product_details`
 
-Get the full details for a specific data product, including its JSON schema
+Returns the full details for a specific data product, including its JSON schema
 with column names, types, and descriptions.
 
 | Parameter | Type | Required | Description |
@@ -65,7 +65,7 @@ with column names, types, and descriptions.
 
 ### `read_data_product`
 
-Read rows from a data product.
+Reads rows from a data product.
 
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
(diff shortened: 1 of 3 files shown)


mtabebe:
this was already reviewed, I just clicked merge too early. So holding off on it.

doc/user/content/integrations/mcp-server/mcp-agent.md
```diff
@@ -88,6 +88,33 @@ isolated to:
    roles since role configurations are **not** inherited; only privileges are
    inherited.
 
+#### Restricting access to user objects only {#restrict-to-user-objects}
+
+By default, a role can query any object it has been granted `SELECT` on,
```
kay-kim: So ... this is a little confusing. Isn't this just for the the query tool (which we haven't even mentioned or turned on)?  But ... placed here, the page logic now goes

- Create an environment and functional role
- A role can query anything for which it's been granted SELECT 
- MCP exposes mat views and indexed views as discoverable objects 
- We grant SELECT on these mat views/indexed views

kay-kim (changes requested):

doc/user/content/integrations/mcp-server/mcp-agent.md
```diff
@@ -557,6 +557,33 @@ configuration](/integrations/mcp-server/mcp-agent-config/#enable_mcp_agent_query
 
 {{< /note >}}
 
+### Restricting access to user objects only {#restrict-to-user-objects}
+
+By default, a role can query any object it has been granted `SELECT` on,
```
kay-kim: Can we mention that this is if you enable the query tool? (and link to that). Otherwise, the "By default, ... " sentence would not be true (I think?).

kay-kim (changes requested):

doc/user/content/integrations/mcp-server/mcp-agent.md
kay-kim: Oh ... and can we also mention that on the https://materialize.com/docs/integrations/mcp-server/mcp-agent-tools/#query page link to this section?

doc/user/content/integrations/mcp-server/mcp-agent.md
```diff
@@ -557,6 +557,34 @@ configuration](/integrations/mcp-server/mcp-agent-config/#enable_mcp_agent_query
...
+When the [`query` tool](/integrations/mcp-server/mcp-agent-tools/#query) is
+enabled, a role can query any object it has been granted `SELECT` on by
+default, including system catalog views. To prevent an agent role from reading
+system catalog tables (`mz_catalog`, `mz_internal`, `pg_catalog`,
+`information_schema`), a superuser can set the `restrict_to_user_objects` role
+default:
+
+```mzsql
+ALTER ROLE mcp_agent SET restrict_to_user_objects = true;
+```
```
kay-kim: So ... the `mcp_agent` -- it's a functional role.
Now, in Materialize,  role configurations are not inherited (only role privileges are).  So ... that's also true of this setting, yes?
If so, we probably should set it to the specific agents, as well, yes?

doc/user/content/integrations/mcp-server/mcp-agent.md
```diff
@@ -557,6 +557,34 @@ configuration](/integrations/mcp-server/mcp-agent-config/#enable_mcp_agent_query
 
 {{< /note >}}
 
+### Restricting access to user objects only {#restrict-to-user-objects}
+
+When the [`query` tool](/integrations/mcp-server/mcp-agent-tools/#query) is
+enabled, a role can query any object it has been granted `SELECT` on by
```
kay-kim: I move move the by default to  "a role can, by default, query any object for which it has `SELECT` privileges ..."
That way, it doesn't seem like it's modifying the select privileges people have by default?

doc/user/content/integrations/mcp-server/mcp-agent-tools.md
```diff
@@ -99,6 +99,10 @@ agent run arbitrary SQL against data products. To enable it, set the
 configuration](/integrations/mcp-server/mcp-agent-config/#enable_mcp_agent_query_tool)
 system parameter to `true`.
 
+To restrict which objects an agent role can query (e.g., prevent access to
```
kay-kim: So ... this thing only restricts access for the system tables, yes? So your "To restrict which objects an agent role can query (e.g.,..." seems to indicate a broader restriction capability.

I'll just add a patch to your branch  (also, we seem to go system catalog views sometimes and system catalog tables other times ... I'll switch to objects to be consistent).

mtabebe: Yes you are right only for the system tables

kay-kim (approved):

PR #36572, opened 2026-05-15
Upsert V2 row row spine backed arrangement

DAlperin:
This PR optimizes the memory straegy for upsert v2 by swapping to RowRowSpine for the key-value lookup arrangement, which is able to swap to disk much more efficiently when memory usage goes over the normal memory limits


patrickwwbutler (approved):
Tested in staging on an undersized cluster, saw huge improvements! looks good to me

PR #30083, opened 2024-10-18
compute: spillable MV correction buffer

teskje:
This PR introduces a "correction v2" buffer that differs from the existing one in that it stores data in columnated regions that can be spilled to disk. It follows the general design of the arrangement merge batcher, with the major difference that it sorts updates by time first, in an attempt to more efficiently deal with the presence of updates in
the future (commonly introduced by temporal filters).

The new correction buffer can be switched on through a feature flag, `enable_compute_correction_v2`, and is switched off by default. The plan is to keep it disabled in production but have it available for emergencies where replicas fail to hydrate due to the MV memory spike. Eventually we'll want to make the new correction buffer the default, but we should do more performance testing before that.

### Motivation

  * This PR adds a known-desirable feature.

Part of https://github.com/MaterializeInc/database-issues/issues/8464

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #36577, opened 2026-05-17
compute: store correction v2 chunks in a columnar chunk region

antiguru:
### Motivation

#30083 added a spillable MV correction buffer that uses `StackWrapper`/`ColumnationStack` for the chunk type. That storage has two issues for the spilling use case:

1. The outer `Vec<T>` lives on the regular heap (not lgalloc-backed), so it does not spill.
2. The columnation regions inside grow without a convenient way to pre-size them, so per-chunk memory is hard to predict.

This PR replaces the chunk storage with a columnar chunk region: each chunk becomes a single, fixed-size allocation, so the correction buffer's memory behavior is predictable and spillable.

Part of MaterializeInc/database-issues#8464.

### Changes

**`Chunk` storage.** Each chunk now holds a `ChunkData<D>` enum with two variants:

- `Typed(ContainerOf<(D, Timestamp, Diff)>)` — typed columnar container, used while a chunk is still being filled.
- `Align(Vec<u64>)` — finished chunk encoded into a single aligned allocation.

This mirrors `mz_timely_util::columnar::Column` minus its `Bytes` variant (which wraps `Arc<dyn Any>` and is `!Sync`, breaking the `+ Send` iterator returned by `updates_before`).

**`ChunkBuilder`.** New helper that mints chunks at the same ~2 MiB serialized-size boundary used by `mz_timely_util::columnar::builder::ColumnBuilder` (`SHIP_WORDS = 1 << 18`). When the in-progress container reaches that boundary, its bytes are encoded into a `Vec<u64>` of exactly the required capacity and emitted as a `ChunkData::Align`. Each minted chunk is therefore a single predictably-sized allocation, and remainder data is flushed as a `ChunkData::Typed` on finalize.

**`ChainBuilder`.** Wraps a `ChunkBuilder` and drains minted chunks into a `Chain<D>`. All the chain-producing paths (`Stage::{insert, flush}`, `Cursor::{into_chain, try_unwrap}`, `CorrectionV2::{merge_2, merge_many}`) go through it; `Chain` itself no longer owns a partial chunk or carries a `chunk_capacity`.

**`Data` trait.** Now requires `Columnar` (with `Container: Send + Sync + Clone`) instead of `Columnation`, plus `for<'a> Borrow<Ref<'a>: Eq + Ord>` so the merge/heap code can compare update refs directly without `into_owned` on the hot path. Both `Row` and `DataflowErrorSer` already implement `Columnar`; the latter gained `#[columnar(derive(Eq, PartialEq, Ord, PartialOrd))]` so its auto-generated `Reference` satisfies the new ref-level bound.

**Lifetime helper for `pop_equal`.** `<C as Borrow>::Ref<'a>` is invariant in `'a` through trait projection, so the compiler can't shorten two refs with unrelated lifetimes by variance on its own. A small `refs_eq` helper explicitly reborrows both inputs to a fresh local lifetime via `Columnar::reborrow` before delegating to `Eq`.

**Chain invariant.** The `chain_proportionality` invariant now compares chains by update count rather than chunk count. Chunk count was a faithful proxy under columnation (fixed updates per chunk) but is not anymore: any chain below the ~2 MiB boundary is a single chunk regardless of update count, so the comparison would degenerate to 1-vs-1, collapse the geometric invariant, and turn inserts into O(N^2) up to the boundary. Comparing update counts restores the pre-refactor O(log N) amortization.

### Notes for reviewers

- `compute_correction_v2_chunk_size` changed meaning: it now only sizes the in-memory `Stage` staging buffer and the ship cadence. On-chain chunk byte size is fixed by the builder's ~2 MiB boundary (`SHIP_WORDS`) and no longer follows this knob.
- Correction-size metrics (`mz_persist_sink_correction_*`) now report serialized bytes (`indexed::length_in_bytes`) instead of columnation heap bytes. Balance-based assertions (`correction-buffer-metrics.td`) are unaffected, but absolute byte values shift — heads-up for anyone dashboarding them.
- Added `chain_builder_roundtrips_across_mint_boundary`, exercising the `Align` encode/decode spill path (crossing at least one `mint()`), which previously had no coverage.

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered.
- [ ] This PR has an associated up-to-date design doc, is a design doc, or is sufficiently small to not require a design.
- [ ] If this PR evolves an existing `$T ⇔ Proto$T` mapping (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR.
- [ ] If this PR includes major user-facing behavior changes, I have pinged the relevant PM.

### DIFF
--- a/src/compute/src/sink/correction_v2.rs
+++ b/src/compute/src/sink/correction_v2.rs
@@ -771,14 +771,12 @@ impl<D: Data> ChainBuilder<D> {
     fn push_ref(&mut self, update: Ref<'_, (D, Timestamp, Diff)>) {
         self.builder.push(update);
         self.drain();
-        self.chain.update_count += 1;
     }
 
     /// Push an owned-form update into the builder.
     fn push_owned(&mut self, update: &(D, Timestamp, Diff)) {
         self.builder.push(update);
         self.drain();
-        self.chain.update_count += 1;
     }
 
     /// Push the updates produced by a cursor into the builder.
@@ -1517,5 +1515,22 @@ impl<D: Data> Ord for MergeCursor<D> {
         let (d1, t1, _) = self.0.get();
         let (d2, t2, _) = other.0.get();
         (t1, d1).cmp(&(t2, d2)).reverse()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use mz_repr::{Diff, Timestamp};
+
+    use super::ChainBuilder;
+
+    #[mz_ore::test]
+    fn chain_builder_update_count_matches_items() {
+        let mut builder = ChainBuilder::<i64>::default();
+        for i in 0..10_i64 {
+            builder.push_owned(&(i, Timestamp::new(i as u64), Diff::ONE));
+        }
+        let chain = builder.finish();
+        assert_eq!(chain.update_count, chain.iter().count());
     }
 }
--- a/src/compute/src/sink/correction_v2.rs
+++ b/src/compute/src/sink/correction_v2.rs
@@ -1041,19 +1041,8 @@ impl<D: Data> Cursor<D> {
 ///
 /// Chunks are immutable once created. They are produced by [`ChunkBuilder`], which mints a
 /// new chunk whenever its in-progress columnar container reaches a fixed serialized byte
-/// boundary (~2 MiB, matching the ship granularity used elsewhere in the codebase). This
-/// means each chunk corresponds to a single, predictable allocation; the previous
-/// columnation-backed `Vec<T>`-plus-region representation had no convenient way to pre-size
-/// the storage.
-///
-/// The chunk holds a [`Column`] directly. `Column` became `Send + Sync` in
-/// `timely_bytes` 0.30, which lets [`CorrectionV2::updates_before`] return a `+ Send`
-/// iterator that captures `&self`. Paging chunks out via
-/// [`mz_timely_util::column_pager::ColumnPager`] (introduced in #36552) would need a deeper
-/// rework of the cursor sharing model — `Cursor` holds `Rc<Chunk<D>>` and accesses chunk data
-/// through `&self`, while `pager.take` consumes its `PagedColumn`. Picking that up should
-/// happen at the `Chain` boundary (chains carry `Vec<PagedColumn>`, with chunks materialized
-/// at `into_cursor` time) and is out of scope for the spilling-correctness fix here.
+/// boundary (~2 MiB, matching the ship granularity used elsewhere in the codebase), so each
+/// chunk corresponds to a single, predictably sized allocation.
 struct Chunk<D: Data> {
     /// The contained updates.
     data: Column<(D, Timestamp, Diff)>,
@@ -1459,7 +1448,7 @@ impl<D: Data> Ord for MergeCursor<D> {
 mod tests {
     use mz_repr::{Diff, Timestamp};
 
-    use super::{ChainBuilder, ChunkData};
+    use super::ChainBuilder;
 
     #[mz_ore::test]
     fn chain_builder_update_count_matches_items() {
@@ -1490,12 +1479,12 @@ mod tests {
         }
         let chain = builder.finish();
 
-        // At least one chunk must have been minted into `Align` form, otherwise the spill
-        // path wouldn't be exercised.
+        // At least one chunk must have been minted into `Column::Align` form, otherwise the
+        // spill path wouldn't be exercised.
         let align_chunks = chain
             .chunks
             .iter()
-            .filter(|c| matches!(c.data, ChunkData::Align(_)))
+            .filter(|c| matches!(c.data, mz_timely_util::columnar::Column::Align(_)))
             .count();
         assert!(
             align_chunks > 0,
--- a/src/compute/src/sink/correction_v2.rs
+++ b/src/compute/src/sink/correction_v2.rs
@@ -123,31 +123,102 @@
 //!   chain 2.a: [(b, 2, +1)],
 //!   chain 2.b: [(a, 2, +1), (c, 2, -1)]
 
-use std::borrow::Borrow;
 use std::cmp::Ordering;
 use std::collections::{BinaryHeap, VecDeque};
 use std::fmt;
 use std::rc::Rc;
 
-use columnation::Columnation;
-use differential_dataflow::trace::implementations::BatchContainer;
+use columnar::bytes::indexed;
+use columnar::{Borrow, Clear, Columnar, FromBytes, Index, Len, Push};
 use mz_ore::cast::CastLossy;
 use mz_persist_client::metrics::{SinkMetrics, SinkWorkerMetrics, UpdateDelta};
 use mz_repr::{Diff, Timestamp};
-use mz_timely_util::columnation::ColumnationStack;
 use timely::PartialOrder;
 use timely::progress::Antichain;
 
 use crate::sink::correction::{ChannelLogging, SizeMetrics};
 
 /// Convenient alias for use in data trait bounds.
+///
+/// `D` is constrained to be `Columnar`, so that updates can be stored in a single columnar
+/// region per chunk, and the variable-length payload (e.g. `Row` bytes) lives in the same
+/// allocation as the rest of the chunk.
 pub trait Data:
-    differential_dataflow::Data + Columnation<InnerRegion: Send + Sync> + Send + Sync
+    differential_dataflow::Data
+    + Columnar<Container: Send + Sync + Clone>
+    + Send
+    + Sync
 {
 }
-impl<D: differential_dataflow::Data + Columnation<InnerRegion: Send + Sync> + Send + Sync> Data
-    for D
+impl<D> Data for D where
+    D: differential_dataflow::Data + Columnar<Container: Send + Sync + Clone> + Send + Sync
 {
+}
+
+/// Shorthand for the columnar container that stores `(D, Timestamp, Diff)` triples.
+type TripleContainer<D> = <(D, Timestamp, Diff) as Columnar>::Container;
+/// Shorthand for a borrowed view of [`TripleContainer`].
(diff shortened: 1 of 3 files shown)


src/compute/src/sink/correction_v2.rs
```diff
@@ -123,33 +123,101 @@
...
-    for D
+impl<D> Data for D where
+    D: differential_dataflow::Data + Columnar<Container: Send + Sync + Clone> + Send + Sync
 {
 }
 
+/// Shorthand for the columnar container that stores `(D, Timestamp, Diff)` triples.
+type TripleContainer<D> = <(D, Timestamp, Diff) as Columnar>::Container;
+/// Shorthand for a borrowed view of [`TripleContainer`].
+type TripleBorrowed<'a, D> = <TripleContainer<D> as Borrow>::Borrowed<'a>;
```
antiguru: This should be `columnar::BorrowOf<(D, Timestamp, Diff)>`.

src/compute/src/sink/correction_v2.rs
```diff
@@ -123,33 +123,101 @@
...
+/// Shorthand for a single-update reference produced when reading from a chunk.
+///
+/// Equivalent to `(D::Ref<'_>, Timestamp, Diff)` — `Timestamp` and `Diff` are returned by
+/// value because their columnar `Ref` types are owned, while `D` is returned by `Ref` so
+/// variable-length payloads don't need to be cloned out of the columnar storage.
+type UpdateRef<'a, D> = <TripleBorrowed<'a, D> as Index>::Ref;
```
antiguru: `UpdateRef` should be `columnar::Ref<(D, Timestamp, Diff)>`.

src/compute/src/sink/correction_v2.rs
```diff
@@ -1339,10 +1459,14 @@ impl<D: Data> MergeHeap<D> {
...
-    fn pop_equal(&mut self, data: &D, time: Timestamp) -> Option<(Cursor<D>, Diff)> {
+    fn pop_equal(
+        &mut self,
+        data: <<D as Columnar>::Container as columnar::Borrow>::Ref<'_>,
+        time: Timestamp,
+    ) -> Option<(Cursor<D>, Diff)> {
         let MergeCursor(cursor) = self.0.peek()?;
         let (d, t, r) = cursor.get();
-        if d == data && t == time {
+        if t == time && D::into_owned(d) == D::into_owned(data) {
```
antiguru: Are we missing an `Eq` implementation for `d`/`data` here? The `into_owned` is potentially expensive.

src/compute/src/sink/correction_v2.rs
```diff
@@ -1379,6 +1503,23 @@ impl<D: Data> Ord for MergeCursor<D> {
...
+/// can be expensive for variable-length `D` (e.g. `Row`); merge-heavy code paths that fire this
+/// often should be revisited once a `Ref`-level comparison trait is available.
+fn cmp_keys<D: Data>(
+    t1: Timestamp,
+    d1: <<D as Columnar>::Container as columnar::Borrow>::Ref<'_>,
+    t2: Timestamp,
+    d2: <<D as Columnar>::Container as columnar::Borrow>::Ref<'_>,
+) -> Ordering {
+    match t1.cmp(&t2) {
+        Ordering::Equal => D::into_owned(d1).cmp(&D::into_owned(d2)),
```
antiguru: Same here, missing `Ord`?

src/compute/src/sink/correction_v2.rs
antiguru: Done for `MergeCursor::cmp` (L1522) and `Chain::can_accept` — both now compare refs directly through the new `Ref<'_>: Eq + Ord` bound I added to `Data`.

I couldn't get `pop_equal` to use direct ref equality though: `d` here is `Ref<'a, D>` with `'a` tied to the immutable borrow from `self.0.peek()`, while `data: Ref<'b, D>` has an unrelated input lifetime, and `reborrow_ref<'b, 'a: 'b>` can't unify them without one outliving the other. Compare-via-`into_owned` sidesteps the issue and (as a bonus) ends the immutable borrow before the `self.pop()` call below. I gated the clone behind a `t != time` short-circuit so it only fires on time matches — happy to revisit if there's a cleaner formulation I'm missing.

---

def- (commented):
The update count is double counted:
```diff
diff --git a/src/compute/src/sink/correction_v2.rs b/src/compute/src/sink/correction_v2.rs
index 23ef00ebf9..a243558737 100644
--- a/src/compute/src/sink/correction_v2.rs
+++ b/src/compute/src/sink/correction_v2.rs
@@ -1519,3 +1519,20 @@ impl<D: Data> Ord for MergeCursor<D> {
         (t1, d1).cmp(&(t2, d2)).reverse()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use mz_repr::{Diff, Timestamp};
+
+    use super::ChainBuilder;
+
+    #[mz_ore::test]
+    fn chain_builder_update_count_matches_items() {
+        let mut builder = ChainBuilder::<i64>::default();
+        for i in 0..10_i64 {
+            builder.push_owned(&(i, Timestamp::new(i as u64), Diff::ONE));
+        }
+        let chain = builder.finish();
+        assert_eq!(chain.update_count, chain.iter().count());
+    }
+}
```
fails:
```
---- sink::correction_v2::tests::chain_builder_update_count_matches_items stdout ----

thread 'sink::correction_v2::tests::chain_builder_update_count_matches_items' (1838738) panicked at src/compute/src/sink/correction_v2.rs:1536:9:
  left: 20
```

antiguru:
## Review

Refactor replacing `columnation`/`ColumnationStack` chunk storage with the `columnar` crate. Chunks are now byte-bounded (~2 MiB, `SHIP_WORDS = 1<<18`) and spill into an aligned `Vec<u64>` (`ChunkData::Align`) via `indexed::encode`. Merge/heap paths compare through columnar `Ref`s instead of cloning via `into_owned`. `errors.rs` gains `#[columnar(derive(Eq, PartialEq, Ord, PartialOrd))]` to satisfy the new `Ref: Eq + Ord` bound on `Data`.

### Blocking

* **The one new test never exercises the code this PR exists to fix.** `chain_builder_update_count_matches_items` pushes 10 `i64`s, which stay a single `ChunkData::Typed` chunk — `mint()` only fires near the 2 MiB boundary, so the `Align` encode → `from_bytes` decode roundtrip (the actual spilling path, and the riskiest new code) gets zero coverage. Add a test that pushes enough updates to force at least one `mint()`, then asserts `iter()` roundtrips values/order/diffs correctly across the spill boundary. Without it the titular fix is untested.

### Strong suggestions

* **`CORRECTION_V2_CHUNK_SIZE` quietly changed meaning.** It now only sizes the staging `Vec` and the ship cadence; chain chunk byte-size is hardcoded to `SHIP_WORDS`. The knob no longer controls on-chain chunk size. If intended, note it in the PR body and check whether the config's documented effect is now stale.
* **`Chain::len()` semantics shifted.** It returns chunk count, now byte-bounded rather than proportional to `update_count`. It feeds the chain-proportionality invariant in `insert_inner` (`merge_needed`). For sub-2 MiB chains, every chain is one `Typed` chunk regardless of update count, so the geometric invariant that bounds chain count / amortizes merge cost compares a weaker proxy than before. Confirm the amortization argument still holds, or compare on `update_count`.
* **Metrics report serialized bytes now.** `Chunk::get_size` uses `indexed::length_in_bytes` / `8*len` instead of columnation heap bytes. `correction-buffer-metrics.td` only checks balance, so it won't break — but the absolute `mz_persist_sink_correction_*` byte values shift. Heads-up if anyone dashboards them.

### Nits

* `ChainBuilder::finish`: `if chunk.len() > 0` is redundant — `pending` only ever holds minted (non-empty) chunks. Harmless/defensive.
* `refs_eq` and the `ChunkData` `!Sync`/`Bytes` doc comments are good — they explain the non-obvious lifetime-invariance and `Send`/`Sync` constraints. Keep.

### Good

* Dropping `cached_size`/`invalidate_cached_size` is a clean simplification — the cache existed because columnation `heap_size` chased pointers; the new `length_in_bytes` is O(columns), so caching is no longer worth it.
* Removing `into_owned` clones from `merge_2`/`merge_many`/`pop_equal` is the right call for the hot path.
* No `as` casts, no `anyhow!`, test uses `#[mz_ore::test]`. Single semantic change, single CODEOWNERS area.

**Bottom line:** solid refactor, but the spill path — the point of the PR — has no test. Add one minting-crossing test and I'd consider it mergeable.

src/compute/src/sink/correction_v2.rs
```diff
@@ -1319,6 +1362,22 @@ fn consolidate<D: Data>(updates: &mut Vec<(D, Timestamp, Diff)>) {
...
 
+/// Compare two columnar refs that have unrelated input lifetimes.
+///
+/// `<D::Container as Borrow>::Ref<'a>` is an associated-type projection through a trait, so
+/// the compiler treats it as invariant in `'a` and won't auto-shorten the inputs by variance.
+/// We instead explicitly reborrow both to a fresh, local lifetime `'x` via
+/// [`Columnar::reborrow`] before letting the inner `==` pick up the `for<'a> Ref<'a>: Eq`
+/// bound on [`Data`].
+#[inline]
+fn refs_eq<D: Data>(a: Ref<'_, D>, b: Ref<'_, D>) -> bool {
```
petrosagg, line 1373: I tried to change the Data trait to make equality work without this reborrow helper but it quickly turned into a monster

petrosagg (approved):
lgtm

DAlperin (approved):
LGTM. Probably wants a feature-benchmark run for memory usage

Alphadelta14 (approved):
providing a stamp from console and cloud.
I didn't see obvious fallacies in my skim through.

antiguru:
Ran nightly, no performance regressions (https://buildkite.com/materialize/nightly/builds/16697)

PR #36604, opened 2026-05-18
Support multiple replica sizes per scale in cluster spec sheet

antiguru:
### Motivation

The cluster spec sheet benchmarking tool currently tests only a single replica size per scale. This change enables testing multiple replica sizes at each scale, allowing side-by-side comparison of equivalent cluster configurations (e.g., `Ncc` vs `M.1-*` sizes with the same worker count).

### Description

This PR refactors the replica size selection logic to support multiple sizes per scale:

1. **API Change**: Renamed `replica_size_for_scale()` to `replica_sizes_for_scale()` which now returns `list[str]` instead of a single string. Added a new `replica_size_for_scale()` convenience method that returns the first (default) size for backward compatibility with setup-only clusters.

2. **ManagedTarget Implementation**: Updated to return both `Ncc` and equivalent `M.1-*` sizes where available. Added `M1_REPLICA_SIZES` mapping that pairs each scale with its corresponding M.1 tier based on worker count equivalence (e.g., scale 2 → `M.1-small` with 4 workers).

3. **DockerTarget Implementation**: Updated to return a single-element list for consistency.

4. **Scenario Execution**: Modified `run_scenario_strong()` and `run_scenario_weak()` to iterate over all returned replica sizes, running each scenario once per size. This allows direct performance comparison between equivalent cluster configurations.

5. **Analysis**: Enhanced `analyze_cluster_results_file()` with M.1 size metadata (credits/hour and worker counts) to properly normalize results across different cluster types when analyzing benchmark data.

### Verification

The changes maintain backward compatibility through the new `replica_size_for_scale()` convenience method. Existing tests and benchmarks will continue to work. The new functionality is exercised by the scenario execution loops which now iterate over multiple sizes per scale.

https://claude.ai/code/session_016ELD471W47RyJdWuoLrUc1

### DIFF
--- a/test/cluster-spec-sheet/mzcompose.py
+++ b/test/cluster-spec-sheet/mzcompose.py
@@ -2826,12 +2826,12 @@ class ClusterObjectLimitsScenario(Scenario):
         for replica_scale in REPLICA_SCALES:
             if replica_scale > max_scale:
                 break
-            size = target.replica_size_for_scale(replica_scale)
-            yield ScalePoint(
-                label=f"replica_size={size}",
-                cluster_size=size,
-                replica_scale=replica_scale,
-            )
+            for size in target.replica_sizes_for_scale(replica_scale):
+                yield ScalePoint(
+                    label=f"replica_size={size}",
+                    cluster_size=size,
+                    replica_scale=replica_scale,
+                )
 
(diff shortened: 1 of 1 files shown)


def- (approved):

ggevay (approved):

antiguru:
Thanks all for your help here!

PR #36605, opened 2026-05-18
adapter: Fix `set_statement_execution_cluster` to not panic from a concurrent cluster drop

ggevay:
Fixes [SQL-303](https://linear.app/materializeinc/issue/SQL-303/thread-coordinator-panicked-at-srcadaptersrccatalogstaters89232).

`set_statement_execution_cluster` re-resolved the cluster name from
cluster_id, which panicked when the cluster was concurrently dropped.
Both callers already resolve the name from the snapshot that selected
the cluster, so pass it in directly instead of re-resolving.

(An earlier version of the PR did a different thing, which was not very nice. The current version should be uncontroversial and easier to review.)

### DIFF
--- a/src/adapter/src/coord/sequencer.rs
+++ b/src/adapter/src/coord/sequencer.rs
@@ -170,10 +170,12 @@ impl Coordinator {
                 Err(_) => (None, None),
             };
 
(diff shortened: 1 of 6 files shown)


SangJunBak (approved):
LGTM!

PR #36606, opened 2026-05-18
sql: fix wrong results for JOIN ... USING (col) AS t with RIGHT/FULL joins

ggevay:
Fixes [SQL-279](https://linear.app/materializeinc/issue/SQL-279/wrong-results-with-join-using-col-as-t)


mgree (approved):
LGTM!

PR #36611, opened 2026-05-18
Use JWT detection for OIDC authentication

SangJunBak:
### Motivation

Before we'd explicitly separate SQL password authentication and OIDC authentication through a connection option when a user's AuthenticatorKind(decided by the listener json) is Oidc. However, we've since discovered many external clients like Looker, DBeaver, Terraform, and dbt don't support override PG connection options or make it hard to do so. 

Thus we change the behavior to branch on the authentication method based on if the password is a JWT. There is an edge case where we'll fail if a user's SQL password is a valid JWT. However since this is gated by the Oidc AuthenticatorKind and because it's not as likely, we accept this case.

### DIFF
--- a/src/environmentd/tests/auth.rs
+++ b/src/environmentd/tests/auth.rs
@@ -1390,7 +1390,12 @@ async fn test_auth_base_require_tls_oidc() {
 
     let server = test_util::TestHarness::default()
         .with_tls(server_cert, server_key)
-        .with_oidc_auth(Some(oidc_server.issuer), Some("sub".to_string()), None)
+        .with_oidc_auth(
+            Some(oidc_server.issuer),
+            Some("sub".to_string()),
+            None,
+            None,
+        )
         .start()
         .await;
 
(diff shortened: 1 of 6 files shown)


src/pgwire/src/protocol.rs
```diff
@@ -253,46 +244,76 @@ where
...
                 }
+            } else {
+                let session = match authenticate_with_password(
+                    conn,
+                    &adapter_client,
+                    user,
+                    Password(password),
+                    conn_uuid,
+                    helm_chart_version,
+                    authenticator_kind,
```
mtabebe: Do we need to update the authenticator_kind here?

src/environmentd/tests/auth.rs
```diff
@@ -1803,10 +1807,10 @@ async fn test_auth_oidc_password_fallback() {
     let oidc_header_basic = make_header(Authorization::basic(oidc_user, user_password));
 
     run_tests(
```
mtabebe, line 1823: Maybe I'm missing something but where is the test for password fallback with the authenticator kind set to OIDC?

mtabebe (approved):
In general this change looks good to me, two tiny questions

src/pgwire/src/protocol.rs
SangJunBak: Ah good catch!

Edit: Should most likely just remove authenticator_kind as a parameter in the sub function authenticate_with_password

src/environmentd/tests/auth.rs
SangJunBak: Oh good catch! So usually `with_oidc_auth` above makes the authenticator kind Oidc, but because we chain `with_password_auth` right after, it overrides the authenticator kind back to password.

PR #36612, opened 2026-05-18
MySQL: fix handling of TIMESTAMP when using TEXT COLUMN

martykulma:
When using TEXT COLUMN for a timestamp column, Materialize did not have proper handling for timestamps read from binlog that were using either the legacy (pre-5.6) representation MYSQL_TYPE_TIMESTAMP or the 5.6+ MYSQL_TYPE_TIMESTAMP2.

This resulted in user not being able to ingest invalid timestamps (like MySQL's special "zero" value timestamps "0000-00-00 00:00:00").

- Fixes handling of DATETIME/TIMESTAMP when ingested as TEXT
- New unit tests added
- Tests for DATETIME/TIMESTAMP & DATE as TEXT are split out and enhanced

Fixes [SS-151: MySQL: decode fails for TIMESTAMP column when using TEXT COLUMNS](https://linear.app/materializeinc/issue/SS-151)

### DIFF
--- a/src/mysql-util/src/decoding.rs
+++ b/src/mysql-util/src/decoding.rs
@@ -24,6 +24,27 @@ use mz_repr::{Datum, Row, RowPacker, SqlScalarType};
 use crate::desc::MySqlColumnMeta;
 use crate::{MySqlColumnDesc, MySqlError, MySqlTableDesc};
 
+/// Canonical text for the MySQL zero-date sentinel ('0000-00-00 00:00:00').
+/// In binlog MYSQL_TYPE_DATETIME/MYSQL_TYPE_DATETIME2 encodings, sec=0 *cannot* represent unix
+/// epoch 0.  The TIMESTAMP type's supported range starts at '1970-01-01 00:00:01' UTC
+/// (<https://dev.mysql.com/doc/refman/8.0/en/datetime.html>), so any sec=0 is
+/// unambiguously this sentinel.
+const MYSQL_ZERO_TIMESTAMP: &str = "0000-00-00 00:00:00";
+
+/// Format the zero-date sentinel for a column with the given fractional
+/// precision (matches the Date arm's `{:0precision$}` behavior).
+fn mysql_zero_timestamp(precision: u32) -> String {
+    if precision > 0 {
+        format!(
+            "{}.{}",
+            MYSQL_ZERO_TIMESTAMP,
+            "0".repeat(usize::cast_from(precision))
+        )
+    } else {
+        MYSQL_ZERO_TIMESTAMP.to_string()
+    }
+}
+
 pub fn pack_mysql_row(
     row_container: &mut Row,
     row: MySqlRow,
@@ -325,9 +346,15 @@ fn pack_val_as_datum(
                         }
                     }
                     Some(MySqlColumnMeta::Timestamp(precision)) => {
-                        // TIMESTAMP arrives as three mysql_common::Value variants
-                        // (refs: mysql_common v0.35.5):
-                        //   Value::Date  — binary query response + binlog DATETIME[2]
+                        // Materialize treats DATETIME and TIMESTAMP as MySqlColumnMeta::Timestamp,
+                        // but they have slightly different semantics as far as the range of dates
+                        // they can represent.
+                        // (see https://dev.mysql.com/doc/refman/8.0/en/date-and-time-types.html).
+                        //
+                        // Three mysql_common::Value variants exist, which are mapped to
+                        // [`MySqlColumnMeta::Timestamp`]
+                        // (see https://github.com/blackbeam/rust_mysql_common/blob/2e6f6696de03c91b9fd95a87356d081285290704/src/binlog/value.rs):
+                        //   Value::Date  — MZ snapshot & binlog MYSQL_TYPE_DATETIME/MYSQL_TYPE_DATETIME2
                         //                  (value/mod.rs:443-445, binlog/value.rs:109-161)
                         //   Value::Int   — legacy binlog MYSQL_TYPE_TIMESTAMP, pre-5.6,
                         //                  4-byte unix epoch (binlog/value.rs:87-90)
@@ -356,6 +383,8 @@ fn pack_val_as_datum(
                                 }
                             }
                             // Pre-5.6 unix epoch, no fractional seconds.
+                            // val == 0 is the zero-date sentinel, not epoch 0.
+                            Value::Int(0) => mysql_zero_timestamp(*precision),
                             Value::Int(val) => chrono::DateTime::from_timestamp(val, 0)
                                 .ok_or_else(|| {
                                     anyhow::anyhow!("received invalid timestamp value: {}", val)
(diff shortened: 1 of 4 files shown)


def-:
Fixes https://github.com/MaterializeInc/database-issues/issues/8006 I guess?

def- (changes requested):
Running `bin/mzcompose --find mysql-cdc down && bin/mzcompose --find mysql-cdc run cdc text-columns-timestamp.td` shows me scary temporary retraction errors:
```
> SELECT id, created_at FROM products ORDER BY id;
executing query failed

Caused by:
    0: db error
    1: ERROR: Invalid data in source, saw retractions (1) for row that does not exist: [Int32(5), String("1970-01-01 00:00:00")]
rows didn't match; sleeping to see if dataflow catches up 🕑 50ms
rows match; continuing at ts 1779156671.0067012, took 0.949501151s
```
If we change the handling of the binlog decoding we also have to make the same change in the snapshot code!

martykulma:
> Running bin/mzcompose --find mysql-cdc down && bin/mzcompose --find mysql-cdc run cdc text-columns-timestamp.td shows me scary temporary retraction errors:

Definitely a scary error! I ran this over a dozen times yesterday (against 9.5, 8.4, 8.0), and again for each version this morning, but still no failures.

I definitely missed some thing in looking at this yesterday. I just came across https://dev.mysql.com/doc/refman/8.0/en/datetime.html -- and this has got me worried because the behavior or DATETIME and TIMESTAMP are different, but in MZ we don't [distinguish](https://github.com/MaterializeInc/materialize/blob/949c704a00a84c6ee4aa500bc863a8ed18759c8b/src/mysql-util/src/schemas.rs#L377) -- which is a problem!

def-:
I *think* they are caused by this:

> If we change the handling of the binlog decoding we also have to make the same change in the snapshot code!

martykulma:
> I _think_ they are caused by this:
> 
> > If we change the handling of the binlog decoding we also have to make the same change in the snapshot code!

Seems like the most likely case (the snapshot case is decoding as Value::Date, where binlog will decode as Value::Bytes). I'm not sure why I could not repro the error locally (seems like it should happen fairly consistently). PR is updated with additional settings.

martykulma:
> Fixes [MaterializeInc/database-issues#8006](https://github.com/MaterializeInc/database-issues/issues/8006) I guess?

My understanding is that issue intends to allow keeping the column type a timestamp while allowing processing of "zero" timestamps as NULL.  That is beyond the scope of this PR, which is just allowing MZ to successfully ingest the column as `TEXT`.


@def- - if you have a chance, take another look, thank you!

test/mysql-cdc/text-columns-timestamp.td
```diff
@@ -0,0 +1,130 @@
...
+
+# Verify the column types were rewritten to text by TEXT COLUMNS.
+> SELECT pg_typeof(created_at), pg_typeof(updated_at), pg_typeof(archived_at), pg_typeof(born_at) FROM products LIMIT 1;
+text text text text
+
+# The decode error stalls the source even after the queries above appear to
+# succeed (the snapshot rows are already in dataflow and remain queryable).
+# Asserting on the source status is what makes this a hard regression check.
+# We poll until either the source is fully healthy or surfaces the stall;
+# this avoids a race where the binlog error has not yet been recorded.
```
ublubu: Having not seen the error myself, it's not obvious what this comment means. I feel like I'd have to do some Linear+Slack archaeology, especially if I find it sometime later.

src/mysql-util/src/decoding.rs
```diff
@@ -325,34 +346,85 @@ fn pack_val_as_datum(
                         }
                     }
                     Some(MySqlColumnMeta::Timestamp(precision)) => {
-                        // Some MySQL dates are invalid in chrono/NaiveDate (e.g. 0000-00-00), so
-                        // we need to handle them directly as strings
-                        if let Value::Date(y, m, d, h, mm, s, ms) = value {
```
ublubu, line 330: Checking that I understand:

We used to only handle `Value::Date` from the MySQL binlog, and now we handle `Value::Int` and `Value::Bytes` also.

src/mysql-util/src/decoding.rs
```diff
@@ -481,3 +553,198 @@ fn check_char_length(
...
+    }
+
+    /// Bytes that are valid UTF-8 but not parseable as a unix epoch
+    /// should produce the same structured error as invalid UTF-8 —
+    /// covers the chrono parse failure path that
+    /// `timestamp_value_bytes_invalid_utf8_errors` doesn't reach.
+    #[mz_ore::test]
+    fn timestamp_value_bytes_unparseable_errors() {
+        let col = timestamp_text_col(0);
+        for payload in [&b""[..], &b"not-an-epoch"[..], &b"2024-04-03 10:15:13"[..]] {
```
ublubu, line 754: Maybe my brain stopped working, but I can't tell why `2024-04-03 10:15:13` isn't a timestamp.

_EDIT: found it_
> Value::Bytes("\<sec>"/"\<sec>.\<usec>") — binlog MYSQL_TYPE_TIMESTAMP2 (5.6+)

src/mysql-util/src/decoding.rs
```diff
@@ -481,3 +553,198 @@ fn check_char_length(
...
+    fn timestamp_value_date_with_precision() {
+        let col = timestamp_text_col(6);
+        let s = pack_one(Value::Date(2024, 4, 3, 10, 15, 13, 123456), &col).unwrap();
+        assert_eq!(s, "2024-04-03 10:15:13.123456");
+    }
+
+    #[mz_ore::test]
+    fn timestamp_value_date_zero_date() {
+        // The whole reason TEXT COLUMNS exists for TIMESTAMP: a
+        // zero-date arriving as Value::Date(0,..) should round-trip as
```
ublubu: Are these round-trip tests?

ublubu (approved):
seems reasonable

test/mysql-cdc/text-columns-timestamp.td
martykulma: This is just saying that if there was a decode error for the various timestamps, it should just cause the test to fail. I'll update to be clearer.

src/mysql-util/src/decoding.rs
martykulma: Correct - this is the branch where user requested TEXT COLUMN for the datetime/timestamp field.

When MZ reads the row via snasphot, it will always come in as `Value::Date`, but when it is read via binlog, it might come in as one of the 3, depending on the actual source type and version of MySQL.

src/mysql-util/src/decoding.rs
martykulma: I will reword that.  Round-trip in the sense that a value of "0000-00-00 00:00:00" as displayed by MySQL must be be decoded to the same string when read via binlog.

martykulma:
I reorganized the mzcompose tests for DATE and DATETIME/TIMESTAMP after seeing there was some overlap and some missing test coverage.

This uncovered another bug where a DATETIME/TIMESTAMP with precision ∈ {1..5} was always padding out to precision 6.

def- (approved):
Thanks for adding the tests! I'll take a closer look when I'm back, but don't let that block the merge.

PR #36614, opened 2026-05-18
doc/semantics: Lean 4 mechanization of UnifiedStream error semantics

antiguru:
Lean 4 mechanization of Materialize's error-handling design at `doc/developer/semantics/`, plus the design doc that drives it at `doc/developer/design/20260517_error_handling_semantics.md`.

## Current state

17 Lean modules in `Mz/`, ~3 kloc, 762 build jobs green, no `sorry`.
Five-layer catalog (Datum / Expression / Row / Schema / Collection) on a **schema-indexed GADT** foundation and a **two-diff baseline**.
`Datum : ColType → Type`, `Expr (sch : Schema n) : ColType → Type`; ill-typed expressions are unconstructible.
`Env sch := (i : Fin n) → Datum (sch.types.get i)` — typed lookup, no OOB case.
`Update sch` carries `(row : Env sch, diff : Int, err_diff : Int)`, both retractable.
Cell-scoped errors via `Datum.err _`; row-scoped errors via the err-diff dimension; global-scoped errors (absorbing diff marker) are spec-only.

## What's in the skeleton

* **Scalar layer.**
`Mz.Datum`, `Mz.PrimEval` (`evalAnd` / `evalOr` / `evalNot` / arithmetic / comparison / `evalAndN` / `evalOrN` / `evalCoalesce`), `Mz.Boolean` / `Mz.Laws` truth tables, `Mz.Strict` err-/null-propagation classes, `Mz.Variadic` absorption, `Mz.Coalesce` per-kind equations.
GADT discipline closes the evaluator codomains structurally; no catch-all `_ , _ => .null` routing.
`evalCoalesce` is a two-pass form (`Coalesce.firstConcrete` + `Coalesce.residue`).

* **Expression layer.**
`Mz.Expr` (mutual `Expr` / `ExprList` GADT), `Mz.Eval` (mutual `eval` / `evalList`), `Mz.Subst` (schema-transforming substitution + soundness), `Mz.MightError` (analyzer + per-primitive `evalX_not_err` lemmas, `might_error_sound` mutual with `args_might_error_sound`, `evalCoalesce_not_err_of_some_safe`, variadic safety).

* **Output schema.**
`Mz.OutputType` — `Expr.outputCols` derives per-`Expr` `ColSchema`.
Soundness via `eval_satisfies_outputCols`.
`DatumSatisfies` / `RowSatisfies` predicates; `EnvErrFree_of_RowSatisfies` bridge.
`coalesce_collapse` schema-rider.

* **Collection layer.**
`Mz.Collection` — `Update sch`, `Collection sch`, `filter` / `project` / `cross` / `negate` / `unionAll`.
`Row.refines` / `Update.refines` / `Collection.refines` Smyth-style lift.
`NoRowErr` precondition + propagation through `negate` / `unionAll` / `project` / `cross` / `filter` (`NoRowErr_filter_of_might_error_false` rides on the schema bridge).
`Collection.Equiv` (perm + merge + drop_zero) with `unionAll_comm_equiv` and `negate_unionAll_self` demonstrators.
`filter_cross_pushdown_left_unsound` counterexample.

* **Arity-cast scaffolding.**
`List.Vector.cast_val` / `cast_eq_append_assoc`, `Schema.cast_mk`, `Schema.append_assoc_heq`, `Update.cast`, `Collection.cast`, `cast_rfl`, `crossOne_diff_assoc` / `_err_diff_assoc` / `_diff_eq` / `_err_diff_eq`.
Composable building blocks toward `Collection.cross_assoc`.

* **Equivalence relations.**
`Mz.Equiv` — `Datum.eqErrSet`, `Datum.refines`, compositionality (`refines_cong_binary` + per-op corollaries), `evalAnd_err_err_eqErrSet_comm` and companions.
`Mz.EquivBounded` — bounded-arithmetic counterexample (`evalPlusBounded_assoc_max_*`).
`Mz.Legal` — non-deterministic `LegalEval`, `legal_of_eval`, `legal_plus_err_either`, err-side commutativity, `LegalEquiv` / `LegalSubsume`.

## Design-doc highlights

* Five-layer error scope: cell (`Datum.err`) / row (`err_diff`) / collection (`DataflowError`) / global (absorbing diff) / sink.
* Two-diff `(diff, err_diff)` collection encoding; spec rules for Predicates, Joins, Aggregates, COALESCE's rescue extension over PostgreSQL.
* *Alternatives* section enumerates rejected and open encodings (carrier replacement, absorbing-only diff, per-`EvalError` map, severity-graded `Datum::Error`).
* **Error-category split** subsection: three placements for separating implementation bugs from data-dependent runtime errors — (A) new `Datum.panic`, (B) collection-scoped absorbing marker, (C) `EvalError.implementationBug` payload.
Recommendation: (C) immediate, (B) eventual, (A) only if forced.

## Open

* `Collection.cross_assoc` — arity-cast scaffolding + multiplicity components landed; row component requires Fin-index manipulation under three nested `▸` casts.
* `filter_cross_pushdown_left_*` recovery windows (strict via `NoRowErr`, data via `eraseRowErr`, refinement via `SignOK`) — blocked on iota reduction of `filterOne`'s match auxiliary, see inline note in `Mz/Collection.lean`.
* `eraseRowErr` relation.
* Variadic / `.divide` static-safe-divisor precision direction on `Expr.outputCols`.

## Architecture

`model.md` opens with an "Architecture: how the pieces fit" section and a mermaid diagram showing the load-bearing dependency: GADT structure + analyses → schema facts → preconditions → relation-tagged rewrites.
`transforms.md` is the canonical register of mechanized and open transforms.

## Build

`ci/test/lean-semantics.sh` — Docker (`mz-lean-semantics:latest`), Lean v4.29.1, Mathlib v4.29.1.
`bin/in-image lake build` for one-shot probes; image is prebuilt and cached.


antiguru:
## Foundational gaps

Ranked by what's currently blocking further coverage.

### Tier 1: blocks most remaining modelable passes

* **Stream equivalence relation.** `UnifiedStream` is `List`, so equality is list-equality. Many natural laws only hold up to multiset / bag equivalence:
  * `unionAll_comm` — fails as list eq (concat order).
  * `consolidate_idem` — `consolidate` *reverses* order on noDup inputs (see `SetOps.lean` algorithm), so `consolidate ∘ consolidate ≠ consolidate`.
  * `cross_comm` — Cartesian product is symmetric only as multiset.
  * `unionAll_distrib_cross_right` — interleaves vs groups.

  Fix: define `UnifiedStream.Equiv ≈ List.Perm` (or stronger: zero-diff and same-carrier-sum quotient), or switch to `Multiset (UnifiedRow × DiffWithError Int)`.

* **`Reduce` operator (group-by aggregate).** `Mz/Aggregate.lean` exists on `List Datum` only. No `UnifiedStream` lift. Single largest infra gap — blocks `fusion/reduce`, `reduce_elision`, `reduce_reduction`, `reduction_pushdown`. Needs group-key expressions, aggregate functions (sum / count / min / max / ...), and per-group error handling.

* **`Map` separate from `project`.** `UnifiedStream.project` currently *replaces* the row with `es.map (eval r)`. Rust's `Map` *appends* columns. Without separate `Map`, `MapFilterProject` and `canonicalize_mfp` can't be modeled cleanly.

### Tier 2: specific pass clusters

* **Schema / column types.** Rows are `List Datum` with no type info. Blocks:
  * `non_null_requirements` — needs per-column nullability lattice.
  * `column_knowledge` — needs per-column `{literal, nullable, type}` lattice.
  * Width-correct join statements (width is implicit and unchecked).
  * SQL type-rejection (currently `evalAnd .int .int` returns `.int n` for totality).

* **Other missing operators:**
  * `TopK` (sort + limit, needs ordering on `Row`).
  * `FlatMap` (table-valued function).
  * `Constant` (literal stream — would unlock `fold_constants`).
  * `Get` / `Let` / `LetRec` (named bindings, fixpoint for recursive queries).
  * `ArrangeBy` (key-indexed view — physical but cited by passes).

* **Equivalence-class lattice** for `equivalence_propagation`. Currently only use-sites modelable.

* **Monotonicity / `NoRetraction` predicate** for `monotonic.rs`. Streams that never retract.

### Tier 3: correctness reach

* **Bridge to actual Materialize code.** No translation function \`MirRelationExpr → UnifiedStream\` composition, no soundness statement "the Rust optimizer's rewrites are valid under this semantics." All current theorems are about an idealized model.

* **Real time / frontier semantics.** `TimedUnifiedStream` has `Nat` times. Missing:
  * Partial order on times.
  * `Frontier` (antichain) type.
  * `since` / `upper` bound machinery.
  * Capability / retraction-as-of-time laws.

* **Fixpoint semantics.** No model of `LetRec` / iterate. WITH RECURSIVE is invisible to the spec.

* **Differential dataflow proper.** `cross` is one-shot Cartesian product, not arrangement-based. No worker partitioning. Mostly physical, but the logical retraction semantics deserve modeling.

### Recommendation order

Tier 1 stream-equivalence + Reduce + separate Map unlocks the most. Schema enables a different class of analyses cleanly. Bridge to `MirRelationExpr` is the biggest reach but the only way the spec becomes load-bearing for real review.

doc/developer/semantics/Mz/Datum.lean
```diff
@@ -0,0 +1,53 @@
...
+`src/expr/src/scalar.rs` (`EvalError`).
+-/
+
+namespace Mz
+
+/-- Cell-scoped errors raised by `Datum`-level operations. The
+skeleton's variants are intentionally small; production
+`EvalError` (in `src/expr/src/scalar.rs`) has many more. -/
+inductive EvalError
+  | placeholder
```
mgree: Drop this?

doc/developer/semantics/Mz/Expr.lean
```diff
@@ -0,0 +1,49 @@
...
+iteration.
+-/
+
+namespace Mz
+
+/-- Scalar expression syntax.
+
+* `lit d`: literal datum.
+* `col i`: reference to column `i` in the surrounding environment.
+* `and a b`, `or a b`: binary logical conjunction and disjunction.
```
mgree: No idea why we'd have this AND `andN`.

doc/developer/semantics/Mz/Expr.lean
```diff
@@ -0,0 +1,49 @@
+import Mz.Datum
+
+/-!
+# `Expr`
+
+A minimal subset of `MirScalarExpr` (see `src/expr/src/scalar.rs`).
+
+The skeleton uses binary `and` and `or` rather than the variadic form
```
mgree: This graf is confusing. We have _both_ in the AST, for some reason, but they're both defined in terms of primitive, binary, false-before-error-before-null semantics for AND.

doc/developer/semantics/Mz/UnifiedStream.lean
```diff
@@ -0,0 +1,734 @@
...
+non-error results are dropped. Existing row-scoped `err` records
+pass through unchanged. -/
+def UnifiedStream.filter (pred : Expr) (us : UnifiedStream) : UnifiedStream :=
+  us.flatMap fun ud => match ud with
+    | (_,      .error)        => [ud]
+    | (.err e, d)             => [(.err e, d)]
+    | (.row r, d)             =>
+      match eval r pred with
+      | .bool true => [(.row r, d)]
+      | .err e     => [(.err e, d)]
```
mgree: Should be `(.row r, .error)`.

doc/developer/semantics/Mz/UnifiedStream.lean
```diff
@@ -0,0 +1,734 @@
...
+semiring laws. For other diffs, the predicate is evaluated on
+every real `row`: survivors stay with their original diff, rows
+whose predicate errs become `err` records (diff unchanged —
+multiplicity is preserved through the error route), non-true /
+non-error results are dropped. Existing row-scoped `err` records
+pass through unchanged. -/
+def UnifiedStream.filter (pred : Expr) (us : UnifiedStream) : UnifiedStream :=
+  us.flatMap fun ud => match ud with
+    | (_,      .error)        => [ud]
+    | (.err e, d)             => [(.err e, d)]
```
mgree: Seems weird to have a notion of row-level error as opposed to a row with `Datum::error` in it.

doc/developer/semantics/Mz/UnifiedStream.lean
```diff
@@ -0,0 +1,734 @@
...
+`split` discards the diff component, mapping every carrier record
+to one bag row regardless of multiplicity. This is lossy for diffs
+other than `.val 1` (duplicate rows, retractions, or
+collection-scoped errors). The round trip still goes through
+because `ofBag` only ever produces `.val 1` diffs.
+-/
+
+namespace Mz
+
+inductive UnifiedRow where
```
mgree: I think this is not right: if we're using `Datum::Error`, then I don't think we need a row-level error. (As we talked about, this is the correct model for mz as she exists, but not for the version with `Datum::Error`.)

mgree (commented):
Some notes from our conversation.

antiguru:
## Soundness finding: predicate pushdown over `Cross(L, R)` is unsound under the new diff semantics

The Lean refactor toward the design doc's row-scoped retractable err encoding (`Diff = Int × ErrCount`) surfaced a real semantic gap. The optimizer's `predicate_pushdown` rewrite `filter pred (cross l r) ↝ cross (filter pred l) r` (with `pred` referencing left columns only) does not preserve the output collection's err multiplicities.

### Concrete counterexample

```
l = [(la, val ⟨1, 0⟩)]                 -- 1 valid copy of la, no err counts
r = [(rb, val ⟨0, single e 1⟩)]        -- 0 valid copies of rb + 1 err copy under err e
pred = lit (bool false)                -- trivially false on every row
```

Cross mul rule: `(a, m) * (b, n) = (a*b, a • n + b • m)`. So
`(1, 0) * (0, single e 1) = (0, single e 1)` — the valid `la` joined with the err copy of `rb` produces one err copy of `la++rb` under err `e`. The err propagates through cross as you'd expect.

* **LHS** `filter pred (cross l r)` = `[(la++rb, val ⟨0, single e 1⟩)]`. Filter on a `bool false` eval zeroes val (already 0) and keeps the err count.

* **RHS** `cross (filter pred l) r` = `[(la++rb, val ⟨0, 0⟩)]`. `filter pred l` zeroes left val. Then mul `(0, 0) * (0, single e 1) = (0, 0 • single e 1 + 0 • 0) = (0, 0)`. The err count is **dropped**.

Mismatch: LHS preserves the err under `e`; RHS loses it.

### Root cause

Cross's mul rule scales right-side err counts by **left-side val count**. Filter pushdown zeroes left val before crossing — so right-side err copies that were going to be carried by left's valid copies disappear.

### Why this is invisible today

In the current production model row-scoped errs route to `DataflowError`. The err collection of `r` flows to the join output regardless of the filter — it's a separate stream, not a multiplicity inside the diff. The pushdown rewrite operates only on the data collection, which agrees with the rewrite by construction.

Once row-scoped errs become first-class diff multiplicities (the design's whole point — preserve row content + retract through normal arithmetic), the data path and err path are unified, and the multiplicative rule must account for both. The rewrite as written then changes the output.

### Options

1. **Guard the optimizer.** `predicate_pushdown` should not fire over `Cross(L, R)` when `R` may carry err diffs (or when its `errs` component is non-empty). Static analysis can be conservative: assume any operator output that didn't pass through `escalateRowErrs` may carry err counts.

2. **Add a compensating term.** Pushdown emits `cross filter_dropped_l r_errs` as a side stream and unions into the output. Algebraically:

   `filter pred (cross l r) = cross (filter pred l) r ⊎ (compensation)`

   where the compensation captures what falls through. Cleaner in theory, harder in practice.

3. **Stream equivalence quotient.** Pushdown holds up to consolidation — same logical collection, different list representation. Requires defining `Stream.Equiv` (multiset / bag equivalence) as a first-class relation. This is the missing Tier 1 foundation mentioned in the earlier gaps comment.

4. **Restrict the diff semantics.** If "row-scoped err multiplicities should never survive past a filter on adjacent dimensions," cross mul rule could drop right errs when left val is zero. But that breaks distributivity and the standard diff-semiring laws — likely worse than the bug.

### Recommendation

Option 1 (guard) is the safest near-term move. Option 3 (quotient) is the right long-term answer but requires the foundational work listed in the prior comment.

The design doc should specify which view it picks. Right now §"Joins" says a join predicate's err is symmetric with `WHERE`'s err, but it doesn't address how pushdown rewrites interact with diff-scoped err counts. The model needs that resolution before the optimizer can target the new encoding.

Filed from the Lean attempt: the proof of `filter_cross_pushdown_left` (which previously succeeded under the old absorbing-only `IsPureData` hypothesis) does not go through here. Old hypothesis ruled out the counterexample by construction; the new diff makes it expressible.

antiguru:
## Current state

Two-diff baseline `(row, diff : Int, err_diff : Int)` is the current spec; indexed-arity `StreamN n` pilot runs alongside; schema layer (`ColSchema`, `Schema n`, per-column `nullable`/`errable`, stream-level `rowErrFree`) lands as the discharge mechanism for rewrites unsound in their general form.

**Mechanized at HEAD (`0e86bd4243`, 765 jobs green, no `sorry`):**

* Scalar layer — `Datum`, `Expr` (variadic `andN/orN`, sugar binary `and/or`), `eval`, `evalCoalesce`, `evalPlusBounded` with `EvalError.overflow`. Static analyses: `might_error`, `colReferencesBoundedBy`, `colShift`, `colReferencesUnused`.
* Three equivalence relations on `Datum`: `eqErrSet`, `refines`, `refinesDual` (`Mz/Equiv.lean`). Live bounded-int counterexample at `x = max` discharged against all three (`Mz/EquivBounded.lean`).
* Indexed-arity stream `StreamN n` (`Mz/StreamN.lean`): `filter`, `project`, `cross` with `Vector.congr` arity casts. `crossOne_assoc` and `cross_assoc` proved.
* `filter_cross_pushdown_left` — three soundness windows:
  * `…_unsound` — strict equality fails (concrete `∃` counterexample).
  * `…_data` — closes under data-side erasure (`eraseErr`).
  * `…_strict` — closes under strict equality when right stream is `NoRowErr`.
* Schema (`Mz/Schema.lean`): `ColSchema { nullable, errable }`, `Schema n { cols, rowErrFree }`, `RowSatisfies`, `StreamRecordN.Satisfies`, `StreamN.Satisfies`. Bridge `NoRowErr_of_satisfies_rowErrFree` consumes a schema fact in the pushdown precondition. Schema-driven rewrites: `evalCoalesce_cons_of_concrete` + Expr-level `eval_coalesce_pair_of_a_concrete`, `NoRowErr_cross`, `NoRowErr_filter`.
* Output-schema propagation (`Mz/OutputType.lean`): `DatumSatisfies`, `Expr.outputType` (precise on `.lit` and `.col`, conservative weakest on the rest), `eval_satisfies_outputType` soundness theorem.

**Headline findings from the pilot:**

1. `filter_cross_pushdown_left` is **unsound under strict equality** when the right stream carries `err_diff > 0`. Witnessed concretely.
2. Recoverable under three orthogonal posture choices: data-side erasure (full collapse of err multiplicity); `refines` (errors-as-bottom, "no spurious errors"); schema fact `recR.err_diff = 0` (strict equality back, given the precondition).
3. Per-payload diff `Int × ErrCount` (preserved in branch history) has the **same** soundness gap on cross pushdown — multiplying err multiplicity against data multiplicity is the structural issue, independent of encoding granularity.

## Decisions to make now

### 1. Equivalence relation for the user-facing surface

The Lean evidence shows different optimizer rewrites are sound under different relations. Pick one as the contract:

| Relation       | Rewrites enabled                                       | Rewrites forbidden                              |
| -------------- | ------------------------------------------------------ | ----------------------------------------------- |
| strict `=`     | data-side laws, predicate substitution, bag pushdown   | cross-pushdown on err side, bounded-int reorder |
| `refines`      | + drop-err rewrites (cross pushdown, constant-folding) | introduce-err rewrites                          |
| `refinesDual`  | + introduce-err rewrites (Postgres posture)            | drop-err rewrites                               |
| schema-gated   | strict `=` once schema certifies err absence           | nothing; cost is schema discipline              |

Recommend schema-gated for the runtime surface, with `refines` as the optimizer-internal relation for transformations that move within the err-free fragment. Rejecting `refinesDual` would commit Materialize to "no spurious errors" semantics.

### 2. Per-payload diff vs single `err_diff`

Two-diff `(diff, err_diff)` is the current spec; per-payload `Int × ErrCount` lives in branch history. The per-payload form retains structured error payloads through multiplicities, but has the same cross-pushdown soundness gap. Two-diff is simpler, cheaper, and exposes errors as a flat count.

Recommend confirming two-diff as the runtime carrier, with per-payload form documented as an alternative in the design doc but not pursued unless a forcing function appears.

### 3. Schema commitment on the runtime side

Materialize already carries `RelationType` with a `nullable` flag per column. Adding `errable` per column and a stream-level `rowErrFree` flag is purely additive; the optimizer rules ride on the same shape the Lean model uses.

Three rollout shapes:
* **Eager**: extend `RelationType` now; every operator carries the new bits; rewrites cite the schema immediately.
* **Lazy**: keep `RelationType` as is; derive `errable` / `rowErrFree` from operator-local analysis at rewrite sites.
* **Phased**: extend schema, but populate the new bits only in a few key operators (sources, sinks, filter, cross) where the soundness obligations bite hardest.

Recommend phased — eager has too much surface area to land at once; lazy reproves the same fact at every rewrite site.

### 4. Where cell errors escalate to row errors

The design doc names three error scopes (cell, row, collection). Cell-scoped `Datum::Error` lives inside a row; row-scoped lives in `err_diff`. The unresolved obligation is *which* operators escalate cell to row:
* `WHERE` clauses — yes, by current spec.
* Join predicates — yes, by analogy with `WHERE`.
* Sink output — yes, sinks cannot emit `Datum::Error` to downstream systems.
* Projection — no by spec (transparent), but a `SELECT … FROM t WHERE cond` plan where cond does not escalate is debatable.
* Aggregation — open; current spec is silent.

Recommend codifying escalation policy as a per-operator attribute (transparent / aware / escalating) and documenting it as part of the operator catalog, ahead of the runtime work.

## Next steps

In rough priority order, gated by the decisions above:

1. **Decision 1 outcome → equivalence-relation discipline**. If schema-gated, every new optimizer rewrite must cite a schema fact rather than the raw equivalence. The Lean model is already structured to support this — `eval_satisfies_outputType` is the proof obligation per new constructor.

2. **Decision 3 outcome → `RelationType` extension**. If phased, the first runtime change is adding the `errable` and `rowErrFree` bits to the existing `RelationType` (no operator changes yet), so subsequent operator changes have a place to write them.

3. **Tighten `Expr.outputType` per constructor**. Mechanical; lets the lit / col foundation lift to `.not`, `.plus`, `.minus`, `.times`, `.divide`, `.eq`, `.lt`, `.ifThen`, and (with mutual recursion) the variadic constructors. Each tightening directly unlocks an optimizer rewrite.

4. **Schema propagation through stream operators**. `filter`, `project`, `cross` — `Schema.append` already exists for `cross`; `project` needs the lift of `outputType` over the projection vector; `filter` adjusts `rowErrFree` by predicate err-freedom.

5. **Cell-error-free row schema as a distinct fact**. Different from `rowErrFree`. Both are honest schema facts and gate different rewrites; the optimizer needs both.

6. **Non-deterministic eval relation for the remaining err-side obligations**. The cross-pushdown soundness gap that even schema doesn't close (e.g., right stream genuinely errors) requires a relational eval `evalRel : Env → Expr → Datum → Prop` that admits multiple SQL-conformant evaluation orders. Open; substantial.

The decisions above are not blocking the Lean work — the skeleton can keep accumulating evidence — but they're the points where the runtime side starts diverging from the model. Worth settling before the next round of `Mz/...` lands.

antiguru:
Pushed three commits working through `review.md`'s third-pass findings.

**`0602c7918e`** — code cleanup
* Drop stale `Mz/Equiv.lean` counterexample block (predicate-pushdown entry misread `filterOne`; refines-compositionality entry already mechanized at `refines_cong_binary`).
* Fix `plus_comm_legal_errR_to_errL` bug — body was `plusErrR` repackaged with no operand swap.
* Drop `Datum.refinesDual` and its five theorems (sole consumer was one bounded-int counterexample; doc commits to `LegalEval` as PG-faithful posture).
* Drop `Schema.notNullable` (orphaned).
* Reframe `Mz/WellTyped.lean` intro: kind soundness is unconditional, `WellTyped` unlocks `outputType` precision only.
* Add docstring on `Schema.cellErrFree`.

**`911ebdb255`** — doc reorganization
* Move error-category-split RFC (three placements for the panic / runtime-error split) from `model.md` to design doc's *Alternatives* section.
* Add "Architecture: how the pieces fit" section near top of `model.md` with mermaid diagram showing the dependency graph (analyses → schema facts → preconditions → relation-tagged rewrites).
* Split `model.md` summary table into "Active mechanized layers and relations" and "Sketches and history".
* Reword `LegalEval` section to match actual scope (binary err-payload non-determinism only).
* Consolidate schema-side open obligations to `transforms.md` as canonical "what's open" register.

**`4248dad246`** — structural refactor
* Bundle `RowSatisfies` + `RowSatisfiesKind` into `RowSatisfiesFull`; precision-direction lemmas no longer thread two hypotheses.
* Bundle `Expr.outputType` + `Expr.outputKind` into `FullColSchema` + `Expr.outputSchema`, with `eval_satisfies_outputSchema` the bundled soundness theorem. Additive — existing callers unaffected.
* Rewrite `evalCoalesce` as a two-pass form (`Coalesce.firstConcrete` + `Coalesce.residue`), replacing the `Coalesce.go` state machine. Three focused safety helpers (`firstConcrete_not_err`, `firstConcrete_none_no_concrete`, `residue_eq_null_of_null_mem`) take over from the threaded `Coalesce.go_not_err`.

Build clean at each step (`bin/in-image lake build`, 861 jobs).

antiguru:
## PL review — round 3 verification

Full review at `doc/developer/semantics/review.md`; this comment summarizes the round-3 verification pass against commits `0602c7918e` (code cleanup), `911ebdb255` (doc reorganization), `4248dad246` (structural refactor).

### Round-3 items verified addressed

* **Stale `Mz/Equiv.lean` counterexamples** — deleted. Block at lines 260–277 now points to live theorems (`evalAnd_err_err_eqErrSet_comm`, `evalPlusBounded_assoc_max_*`, `refines_cong_binary`, `filterOne_cross_pushdown_left_unsound` / `filter_cross_pushdown_left_refines`) instead of describing them as open.
* **`plus_comm_legal_errR_to_errL` bug** — fixed at `Mz/Legal.lean:293-297`. Body now `LegalEval.plusErrL h` with conclusion `LegalEval env (.plus b a) (.err er)` — actual operand swap.
* **`Datum.refinesDual`** — dropped. Sole consumer (`evalPlusBounded_assoc_max_refinesDual`) removed from `EquivBounded`. `model.md` prose, summary table, `README.md`, `Eval.lean` docstring all reconciled.
* **`Schema.notNullable`** — dropped.
* **`Mz/WellTyped.lean` intro** — reframed. Module docstring now explicit: kind soundness is unconditional; `WellTyped` unlocks `outputType` precision only.
* **`Schema.cellErrFree` docstring** — added (collection-wide projection of per-column `errable` bits).
* **Error-category-split RFC** — moved to design doc `Alternatives → Error-category split (open future)`. `model.md:489-503` carries a one-paragraph pointer with the decision (stay with `.null`).
* **Architecture section + mermaid** — added at `model.md:22-72`. Renders the load-bearing graph (analyses → schema facts → preconditions → relation-tagged rewrites).
* **Summary table split** — `model.md:766-796` now has two tables: "Active mechanized layers and relations" (planner-citable today, 10 rows) and "Sketches and history" (foundations + historical encodings, 4 rows). `eraseRowErr` and `LegalEval` correctly moved to the second table.
* **`LegalEval` reframed** — `model.md:712-749` now leads with "Scope of the current implementation": binary err-payload non-determinism only; variadic short-circuit, collection lift, data-side commutativity flagged as open before SQL-faithful status.
* **`RowSatisfiesFull` bundle** — landed in `Mz/WellTyped.lean:230-244`. `eval_not_satisfies_precise` consumes it directly via `.toSat` / `.toKind` projections; precision-direction lemmas no longer need two-hypothesis threading.
* **`FullColSchema` + `Expr.outputSchema`** — landed in `Mz/WellTyped.lean:120-136` with bundled soundness `eval_satisfies_outputSchema` at `:462-466`. Existing `outputType` / `outputKind` callers unaffected.
* **`evalCoalesce` two-pass refactor** — `Coalesce.firstConcrete` / `Coalesce.residue` replace the `Coalesce.go` state machine. Three focused helpers (`firstConcrete_not_err`, `firstConcrete_none_no_concrete`, `residue_eq_null_of_null_mem`) replace the prior single `Coalesce.go_not_err`.
* **`transforms.md` schema rows** — added `schema_filter`, `schema_project`, `schema_cross`, `cellErrFree_propagation` rows + open-obligations block; `model.md`'s schema-side open list collapsed to a one-line pointer.

All three commits compile clean (`bin/in-image lake build`, 861 jobs).

### Residual concerns

Items that are still latent or deferred — not regressions, just unfinished work the review still flags.

* **`outputType` precision direction beyond `.not`**. `eval_not_satisfies_precise` demonstrates the pattern; arithmetic / comparison / `.ifThen` precision lemmas (`nullable := OR-of-inputs.nullable` under `WellTyped`) are queued. Bundled `RowSatisfiesFull` / `FullColSchema` is the right tool; pulling the trigger is the next step.
* **`LegalEval` scope vs SQL-faithfulness**. Foundation lands; variadic short-circuit admissibility (`FALSE AND 1/0` admitting both outcomes), data-side binary commutativity, and the collection-level lift are the three follow-ups that make this the *PG-faithful* posture. Until they land, the relation can demonstrate the pattern but cannot underwrite the rewrites that motivated it.
* **`LegalEval` 18-constructor binary-op duplication**. Still present (six ops × three rules). Mechanical complexity that scales with the operator set. A parameterized form would collapse to three rules + a primitive-op tag. Low priority but worth doing before adding more binary primitives.
* **`Collection n` order-sensitivity and retraction**. Documented as caveats in `model.md:339-380`; the `Collection.Equiv` permutation quotient and the consolidation quotient are deferred with the time dimension. These are the two reasons several `=`-tagged rewrites are stronger than they need to be.
* **Cell-error vs row-error escalation policy for `Reduce` / sinks**. Open per `model.md:544-549`. Active operators (filter / project / cross / unionAll / negate) have policies stated.
* **`rowErrFree` schema-inference converse**. `NoRowErr_of_satisfies_rowErrFree` discharges the precondition from a schema fact; the converse direction ("operator output satisfies `rowErrFree` schema when inputs do and operator is statically err-free") is implicit via the `NoRowErr_*` propagation theorems but not stated as a single soundness theorem the planner can cite uniformly.
* **Single source of truth for OOB column semantics**. `Env.get`, `Expr.subst`, and `Expr.outputType (.col i)` all agree on `.null`-on-OOB; the dependency is documented in the `Env.get` docstring but not enforced by a shared definition.

### Posture

Three review passes in, the skeleton is in materially better shape than at the start: every headline finding from passes 1 and 2 is addressed, and the round-3 structural cleanups (bundled satisfaction, output-schema unification, summary-table split, architecture diagram, RFC relocation) noticeably reduce reader friction. The remaining residual concerns are scope-expansion work, not modeling bugs. The skeleton is a credible foundation for the error-handling design doc to cite, and the `transforms.md` / `model.md` split — catalog of mechanized obligations vs working reference — gives a clean home for future iterations.

antiguru:
## PL review — round 4 (post-GADT migration)

Skeleton migrated to a schema-indexed (GADT) model: `Datum : ColType → Type`, `Expr (sch : Schema n) : ColType → Type`, `Update sch` / `Collection sch`, mutual `ExprList sch k` for variadics. Twenty commits across the migration; net ~4 kloc reduction.

### What the migration buys

* Evaluator codomains closed at the type level. `evalAnd : Datum .bool → Datum .bool → Datum .bool` has no catch-all. Algebraic laws (`not_not`, AND / OR truth tables, `coalesce_*` rescue rules) state without the `¬d.IsInt` hypotheses the untyped model required.
* Ill-typed expressions unconstructible. `Expr.not (.lit (.int 5))` fails to type-check. Previous `WellTyped` predicate (477 lines + bundled `RowSatisfiesFull` / `FullColSchema` API) subsumed structurally — entire module deleted, soundness becomes a type-formation property.
* Mixed-kind comparison (`eq` / `lt`) made unspeakable. Shared kind parameter on both operands eliminates the `_, _ => .null` catch-all that previously polluted `nullable`.
* `Coalesce.firstConcrete` / `Coalesce.residue` two-pass form per kind. Bool / int / top each get their own equation set in `Mz/Coalesce.lean`.
* `Env sch := (i : Fin n) → Datum (sch.types.get i)` — typed lookup, no OOB null fallback. `eval_subst` now mutual + unconditional, no totality dependence on `Env.get`. `Subst` schema-transforming (`sch_src → sch_tgt`), strictly more general than the untyped form.
* `outputCols` replaces `outputType` + `outputKind`. The kind dimension lives in the GADT index, so a single function suffices. `eval_satisfies_outputCols` is the unified soundness — but the surrounding `RowSatisfies` / `DatumSatisfies` machinery has been pared back from the previous bundled `*Full` shape.

This is the right architectural move. Headline benefit: the soundness story for the type discipline collapses from "predicate + soundness theorem + threading two hypotheses through every precision lemma" to "the type-checker did it". The single-instance `eval_not_satisfies_precise` from round 3 is now redundant.

### Regressions — mechanized content lost in the migration

Most of the round-1-through-3 mechanizations were *not* ported. Comparing `4248dad246` (head of round-3) to current `HEAD`:

* **`Mz/ColRefs.lean` deleted (690 lines).** Lost: `colReferencesBoundedBy`, `colShift`, `colReferencesUnused`, `eval_append_left_of_bounded`, `eval_append_right_shift`, `eval_replaceAt_of_unused`. These were the substrate for predicate pushdown and column-pruning.
* **`Mz/WellTyped.lean` deleted (477 lines).** Beyond the predicate itself (correctly subsumed), the soundness companions `Expr.kind_of_eval`, `eval_not_satisfies_precise`, `RowSatisfiesFull` bundle, `FullColSchema`, `Expr.outputSchema`, `eval_satisfies_outputSchema` are all gone. Some content (output type derivation) folded into the GADT index; bundled satisfaction did not survive.
* **`Mz/Collection.lean` shrunk to 159 lines (was ~925).** Lost: `cross`, `crossOne`, `castUpdate`, `castCollection`, `crossOne_assoc`, `cross_assoc`, `cross_cons_left`, `cross_append_left`, `cross_nil_left/right`, `filter_append`, `project_append`, `negate_append`, `filterOne_cross_pushdown_left_unsound` (the canonical counterexample), `filterOne_cross_pushdown_left_strict / _data / _refines`, `filter_cross_pushdown_left_strict / _data / _refines`, `Row.refines`, `Update.refines`, `Collection.refines` + lift theorems, `SignOK`, `eraseRowErr`, `eraseErrAll`, `NoRowErr_filter / _cross / _project / _unionAll / _negate`, `NoRowErr_of_satisfies_rowErrFree`. **`cross` is missing entirely** — the operator the whole pushdown obstruction was built around. `README.md:78` flags it as "deferred" but does not call out that it was mechanized two rounds ago.
* **`Mz/Schema.lean` shrunk to 93 lines.** `RowSatisfies` / `Update.Satisfies` / `Collection.Satisfies` / bridge theorems / `NoRowErr` definition / `NoRowErr_of_satisfies_rowErrFree` moved or dropped. `Schema.cellErrFree` retained; `RowSatisfies` moved to `OutputType.lean`.
* **`Mz/Legal.lean` reduced to the bare inductive (91 lines, was ~300).** Lost: `legal_of_eval` (the bridge to deterministic eval), `legal_plus_err_either` (motivating theorem), `plus_comm_legal_errL_to_errR / _errR_to_errL`, `LegalEquiv`, `LegalSubsume`, refl / trans lemmas. The relation now exists as a definition with no theorems — no bridge from deterministic `eval`, no equivalence relation, no commutativity witness.
* **`Mz/MightError.lean` reduced to 224 lines (was 870).** Per-primitive `_not_err` lemmas + analyzer + `EnvErrFree` retained. The `might_error_sound` soundness theorem and the cross-cutting analyses appear gone. Need verification.
* **`Mz/EquivBounded.lean` from 172 → 98 lines.** Some bounded-arithmetic counterexamples or follow-ups dropped.
* **`Mz/Equiv.lean` from ~410 → 133 lines.** `eqErrSet_symm`, `eqErrSet_err_err`, `evalAnd_err_err_eqErrSet_comm`, the counterexample-pointer block, `Datum.err_refines` (kept), `Datum.refines_of_eq` (kept). Compositionality lemmas kept and re-indexed.

Verdict: the GADT structurally improved the model, but the migration was a **rewrite-from-scratch on the type layer**, not a port. Most of the equational / pushdown / non-determinism mechanization was not carried over. The skeleton today proves the *typing discipline* of the scalar layer; it does not currently prove any collection-level rewrite.

### Massive doc / code drift

`model.md`, `transforms.md`, and `README.md` were not synchronized with the deletions. Concrete examples:

* `model.md:22-104` ("Architecture") accurately describes the GADT model.
* `model.md:106-326` (Datum / Expression / Row / Schema sections) describe the **untyped pre-migration model**, contradicting the Architecture section. `Datum` is shown as a flat 4-variant inductive; `Expr` as untyped `inductive Expr`; `Row` as `RowN n = List.Vector Datum n`; the `Mz/WellTyped.lean` references occupy lines 248-299 as if the module still exists.
* `model.md:167-173` references `colReferencesBoundedBy`, `colShift`, `colReferencesUnused` — all in deleted `ColRefs.lean`.
* `model.md:220-236` references `NoRowErr_filter / _project / _unionAll / _negate / _cross` and `filter_cross_pushdown_left_strict` — all deleted.
* `model.md:407-411`, `:650-661`, `:690-707`, `:709-738` reference `Update.refines`, `Collection.refines`, `SignOK`, the unsound + recovery story, `eraseRowErr` — all deleted.
* `model.md:815-832` summary table lists every deleted relation as "mechanized in `Mz/Collection.lean`".
* `transforms.md:81-82` lists four pushdown variants (`_unsound`, `_data`, `_strict`, `_refines`) as mechanized; all four are deleted. The same row references `Mz/ColRefs.lean`. `transforms.md:127` references `eval_replaceAt_of_unused` in `ColRefs.lean`.
* `README.md:74-79` claims `Mz.Equiv` covers compositionality (true), `Mz.Collection` covers `filter`, `project`, `cross [deferred]`, `negate`, `unionAll`, `NoRowErr` — the parenthetical "[deferred]" is the only honest update; the rest understates what's been dropped (no append theorems, no pushdown).

A reader cannot tell from the prose what the skeleton currently proves. `transforms.md` reads as "we have a pushdown story under four windows"; the code says "we have schema-indexed `filter` / `project` / `negate` / `unionAll` with `unionAll_assoc` and `negate_negate` only".

### Naming consistency

* `ColKind` (round 3) → `ColType` (now in `Mz/Schema.lean`). Migration is consistent at the code level. `model.md:240` correctly says `ColType { bool | int | top }`, but `model.md:265-266` still writes `Expr.outputKind` and `type_of_eval` and references the `kind_evalPlus` family from the deleted `WellTyped.lean`. Pick a vocabulary.
* `outputType` → `outputCols`. Code uses the latter; prose still uses the former in many places (`model.md:301`, `transforms.md` schema-driven section).

### Recommendations

The migration shape is right. The execution is half-finished — type discipline rebuilt, equational layer dropped on the floor. Two paths forward:

1. **Port the round-1-through-3 mechanizations to the indexed model.** Most should carry over cleanly: `cross`, the pushdown counterexample, `refines` lift, `SignOK`, `NoRowErr_*` propagation, `LegalEval`'s soundness + `LegalEquiv` + motivating theorems, `might_error_sound`, `Subst`'s consumers, the column-bound analyses. Each was a paragraph of code in the untyped model; the indexed form should be similar.
2. **Update the docs to match what's actually proved today.** `model.md`'s Datum / Expression / Row / Schema sections need a rewrite to match the GADT; the summary table needs to remove every deleted entry; `transforms.md` needs every pushdown row re-tagged "open" (was-mechanized-in-untyped-model). Otherwise downstream readers will cite theorems that no longer exist.

Until path 1 lands, path 2 is mandatory — the current state is doc-fraud.

The full review (rounds 1-3) at `doc/developer/semantics/review.md` is **stale** against this migration; many of its findings are no longer applicable (refines lift, NoRowErr propagation, well-typed bundling — all gone, not addressed). Recommend a fresh `review.md` once paths 1 / 2 settle.

antiguru:
Round-4 review (post-GADT migration) noted regressions: type discipline rebuilt, equational layer dropped. Path 2 (docs) flagged as mandatory; path 1 (port lost mechanizations) as real work.

Working through both. Four commits since the round-4 comment:

* **`c189417`** R1-R3 — restored `eqErrSet` companions (`evalAnd_err_err_eqErrSet_comm`), `Mz/Legal.lean` soundness (`legal_of_eval`, `legal_plus_err_either`, `plus_comm_legal_*`, `LegalEquiv`/`LegalSubsume`), `MightError.might_error_sound` mutual with `args_might_error_sound`, plus Coalesce safety helpers (`firstConcrete_not_err`, `firstConcrete_none_no_concrete`, `residue_eq_null_of_null_mem`, `evalCoalesce_not_err_of_some_safe`, `evalAndN_not_err_of_all_safe`, `evalOrN_not_err_of_all_safe`).
* **`d9d9eecaba`** R4-R5 — restored `Row.refines` / `Update.refines` (Smyth-style) / `Collection.refines` (List.Forall₂) lift; `NoRowErr_negate` / `NoRowErr_unionAll` / `NoRowErr_project` propagation.
* **`de59504e78`** R8 — restored `Collection.Equiv` (perm + merge + drop_zero) with `unionAll_comm_equiv` and `negate_unionAll_self` demonstrators on the schema-indexed Collection.
* **`e95a890b8d`** R10 — `model.md` / `transforms.md` / `README.md` reconciled. Datum / Expression / Row / Schema sections rewritten to reflect the GADT (was the most flagrant "doc-fraud" in the review). Every deleted theorem retagged "open (was mechanized in untyped predecessor)". Summary table refreshed. `outputType` → `outputCols` rename consistent. ColRefs analyses noted as subsumed by `Fin n` indexing.

**Still open** from the round-4 list (deferred regressions):

* `cross` + `crossOne` + `cross_assoc` — needs `Schema.append`-level `Env.append` row composition (type-cast plumbing through `(sch_l.types ++ sch_r.types).get i`). This is the load-bearing follow-up: without it, the entire pushdown story (R7) can't ship.
* `filter_cross_pushdown_left_unsound` / `_strict` / `_data` / `_refines` + `SignOK` + `eraseRowErr` — depend on `cross`. The encoding insight unchanged; the mechanizations need restating against the schema-indexed shape.
* `eval_satisfies_outputCols` — mostly mechanical port, deferred. `Expr.outputCols` derives the schema; soundness was open even before round 4.
* `NoRowErr_filter` — needs `RowSatisfies → EnvErrFree` bridge ported. Also depends on the `might_error_sound` chain working at the collection level.
* `coalesce_collapse` / `eval_coalesce_pair_of_a_concrete` — schema-rider for coalesce identity. Mechanical re-port.

Build clean at every commit (594 jobs). The skeleton now proves: type discipline (GADT), substitution soundness, equality and refinement preorders with compositionality, non-determinism foundation, might-error soundness, retraction-modulo-consolidation equivalence with retraction demonstrator. It still does not prove any pushdown rewrite on the err side — that comes back when `cross` lands.

antiguru:
Three more commits since last update — cross, pushdown counterexample, outputCols soundness.

* **`9090bbcd45`** R6 — `cross` + `crossOne` on schema-indexed Collection. Row composition via `Schema.types_get_append_left` / `_right` helper lemmas (Mathlib's `List.getElem_append` with `dif_pos` / `dif_neg`). Bilinear err-diff rule mechanized. `NoRowErr_cross` propagation (three err-diff terms vanish at zero inputs).
* **`9288f56ec1`** R7 partial — `filter_cross_pushdown_left_unsound` counterexample restored. Concrete witness on `Schema.free 0` with `.lit (.bool false)` predicate showing LHS has err_diff=1, RHS has err_diff=0. The three recovery windows (strict via `NoRowErr r`, data-side via `eraseRowErr`, refinement via `SignOK`) remain open — proofs need substitution-aware predicate evaluation across cross's combined schema.
* **`5f63abd7b4`** R9 — `eval_satisfies_outputCols` soundness. Structural recursion on indexed `Expr`; each constructor closes via the matching `evalX_not_err` lemma from `MightError`. Variadic / `.divide` cases use the conservative `DatumSatisfies.weakest` (matches the conservative `outputCols` rules).
* **`aa8810239f`** doc reconciliation — model.md / transforms.md / README.md now reflect: `cross` mechanized; pushdown counterexample mechanized; recovery windows open; `NoRowErr_cross` propagation mechanized; `eval_satisfies_outputCols` mechanized. Summary table updated.

Status against round-4 review punch list:

| Item | Status |
|---|---|
| eqErrSet companions | ✓ restored |
| LegalEval soundness | ✓ restored |
| MightError.might_error_sound | ✓ restored |
| Row/Update/Collection refines lift | ✓ restored |
| NoRowErr propagation (negate / unionAll / project / cross) | ✓ restored |
| Collection.Equiv + demonstrators | ✓ restored |
| cross + crossOne | ✓ restored |
| filter_cross_pushdown_left_unsound | ✓ restored (counterexample) |
| filter_cross_pushdown_left_strict / _data / _refines | open (recovery proofs) |
| eraseRowErr | open |
| eval_satisfies_outputCols | ✓ restored |
| NoRowErr_filter | open (needs RowSatisfies → EnvErrFree bridge) |
| coalesce_collapse | open |
| cross_assoc | open (arity-cast plumbing) |
| docs reconciled | ✓ |

Build clean at every commit (594 jobs). The skeleton now proves the type discipline, substitution soundness, output-schema soundness, refinement compositionality, non-determinism foundation, might-error soundness, retraction-modulo-consolidation equivalence — *and* the canonical pushdown counterexample under strict equality. Recovery windows + `coalesce_collapse` + `cross_assoc` remain the substantive open items.

antiguru:
Follow-up cycle (commits `b218cf7d8e` → `c74e3aaa15`, 6 commits):

| Item | Status |
|---|---|
| `NoRowErr_filter` + `EnvErrFree_of_RowSatisfies` bridge | ✓ landed |
| `coalesce_collapse` schema-rider + supporting Datum lemmas | ✓ landed |
| `filter_cross_pushdown_left_strict` recovery proof | blocked on iota of `filterOne`'s match auxiliary |
| `cross_assoc` arity-cast plumbing | deferred (Schema HEq under `Nat.add_assoc` + Vector append assoc) |

**Pushdown blockage detail.** Proof structure (induction on `l`, `cross_cons_left` + `filter_append` distribution, `congr` to pointwise `filterOne_crossOne_eq_crossOne_filterOne`, `hLift` rewrite to align scrutinees, case-split on `eval recL.row p_left`, ring-close per case under `NoRowErr_l ∧ NoRowErr_r`) is well-understood and ring-closeable. The mechanization tripped on iota reduction of the match after `cases hev : eval recL.row p_left`. Tried `unfold filterOne`, `simp [filterOne]`, `change`, per-case rewrite lemmas, `generalize` + `clear`, marking `filterOne` `@[reducible]` — none expose the iota-reduced match arm. Fix likely requires marking `filterOne.match_1` reducible separately (not `filterOne` itself), routing via `filterOne.eq_*` auto-gen equation lemmas, or rewriting `filterOne` to dispatch via explicit `Datum.casesOn`. Documented inline in `Mz/Collection.lean`.

Build clean at every commit (594 jobs). The pushdown counterexample (`filter_cross_pushdown_left_unsound`) remains the load-bearing demonstration of the un-recovered direction.

antiguru:
Cross_assoc scaffolding continued (`adb067b67c` → `31513a653b`, 4 commits):

| Block | Status | How |
|---|---|---|
| `Vector.cast_val` | ✓ | `cases h; rfl` (free-var Nat eq) |
| `Vector.cast_eq_append_assoc` | ✓ | `Subtype.ext` + `cast_val` + `List.append_assoc` |
| `Schema.cast_mk` | ✓ | `cases h; rfl` |
| `Schema.append_assoc_heq` | ✓ | `eqRec_heq.symm` → `heq_of_eq` → `cast_mk` rewrite → `congr 1` → 3 closers |
| `Update.cast` / `Collection.cast` | ✓ | `subst h_arity` + `eq_of_heq h_sch` + `subst heq` |
| `cast_rfl` | ✓ | `id` cast is no-op |
| `crossOne_diff_assoc` / `_err_diff_assoc` | ✓ | `ring` on `Int` polynomial identities |
| `crossOne_diff_eq` / `_err_diff_eq` | ✓ | `unfold crossOne; ring` |
| `Collection.cross_assoc` | open | row-component requires Fin-index manipulation under three nested `▸` casts + inductive lift |

Pulled in `Mathlib.Tactic.Ring` (168 extra build jobs, total now 762). Build clean at every step.

The remaining open work on `cross_assoc` is one focused effort on the row component: showing that `(crossOne (crossOne a b) c).row` (a function `Fin (n+m+k) → Datum _`) cast via `Schema.append_assoc_heq` equals `(crossOne a (crossOne b c)).row` (a function `Fin (n+(m+k)) → Datum _`). The case analysis falls cleanly into the three Fin ranges (`< n`, `n ≤ · < n+m`, `≥ n+m`), each landing on the same primitive row lookup; the cast types align via the existing `Schema.types_get_append_left/right` lemmas.

antiguru:
## PL review — round 5

Round-4 findings addressed substantively. The R1-R10 commit series restored cross + crossOne, NoRowErr propagation through every operator, the Smyth refines lift, Collection.Equiv with the load-bearing `negate_unionAll_self` retraction demonstrator, the pushdown counterexample, `eval_satisfies_outputCols` soundness, the `EnvErrFree_of_RowSatisfies` bridge, `coalesce_collapse` schema-rider, full `might_error_sound` mutual proof, arity-cast scaffolding (`Schema.append_assoc_heq` via composable blocks), and crossOne field-level multiplicity associativity (`crossOne_diff_eq`, `crossOne_err_diff_eq` via `ring`). The summary table in `model.md` is honest and the open items are correctly scoped. Net of round-4 regressions: recovered. Plus the GADT discipline as a structural win.

### Findings round 5

**1. `doc/developer/semantics/review.md` is fossilized.** It predates the GADT migration. Names dead modules (`Mz/WellTyped.lean`, `Mz/ColRefs.lean`, `Coalesce.go` state machine), dead types (`ColKind`, `RowSatisfiesKind`, `outputKind`, `refinesDual`, `Schema.notNullable`, `eraseErr`), claims still-unaddressed items that were addressed (e.g. `plus_comm_legal_errR_to_errL` "bug" at lines 339-355 — current `Mz/Legal.lean:156-166` swaps operands correctly in both directions). The line citations are stale. Worse than just out-of-date: someone reading it cold gets a wrong picture of the current model. Either delete it, or fully rewrite it as a round-5 snapshot. Don't leave a dead PL-review file at the canonical path.

**2. Migration-vestige docstrings everywhere.** Twelve modules say "Indexed counterpart of `Mz/X.lean`" but the untyped sibling doesn't exist — the path resolves to the file containing the comment. Self-reference. Examples:

* `Mz/Datum.lean:7-21` — full block on "indexed Datum" vs nonexistent untyped variant, "continues to compile alongside this file during the GADT migration" (no, it doesn't, because there's no other Datum)
* `Mz/PrimEval.lean:11` — "mirroring the layout of the untyped `Mz/PrimEval.lean`"
* `Mz/Eval.lean:21` — "Same shape as `Mz/Eval.lean` modulo the indexing"
* `Mz/Strict.lean:6`, `Mz/Equiv.lean:7`, `Mz/MightError.lean:7`, `Mz/OutputType.lean:11`, `Mz/Collection.lean:11`, `Mz/Legal.lean:7`, `Mz/Coalesce.lean:6`, `Mz/Boolean.lean:7`, `Mz/Laws.lean:7`, `Mz/Variadic.lean:7`, `Mz/EquivBounded.lean:7`

The "indexed counterpart of" framing was correct during the migration when both lived in tree. Two cycles after the `Mz/Indexed/` prefix was dropped (commit `8a6ca2ee37`), the wording is leftover scaffolding. Sweep-and-delete.

**3. `Mz/Schema.lean:9-14` has stale `Mz/Indexed/` paths.** "The actual satisfaction predicates and indexed schema-driven theorems live in `Mz/Indexed/Schema.lean` and `Mz/Indexed/OutputType.lean`. … The GADT migration moved every predicate that touches the value layer to `Mz/Indexed/*`." That directory was dropped — predicates live in `Mz/OutputType.lean`. The whole "import-only" framing reads as transitional state.

**4. `transforms.md:183` vs `:188` contradict.** Line 183: "Soundness theorem is open in the indexed model." Line 188 (same theorem `eval_satisfies_outputCols`): "mechanized in `Mz/OutputType.lean`; structural recursion + `evalX_not_err` lemmas from `MightError`." It is mechanized — line 183 needs to drop the "open" claim.

**5. `model.md:651` typo.** `Row.refines a b := ∀ i : Fin n, (a.get i).refines (b.get i)`. But `Env sch := (i : Fin n) → Datum (sch.types.get i)` is a function, not a vector — the correct call is `a i`, matching the actual definition at `Mz/Collection.lean:531-532`.

**6. `Mz/PrimEval.lean:13-21` lists four groups, prose says three.** "The primitives split into three groups, mirroring the layout of the untyped …" — then lists Binary boolean + Integer arithmetic + Same-kind comparison + Variadic. That's four. Bump the count.

**7. `Mz/Datum.lean:48-58`: `IsNull` predicate is `Prop`, `isNullB` is `Bool`.** Two functions for the same fact, used in different contexts. Fine, but `isNullB`'s docstring "for use in fold contexts where a `Prop` doesn't compose" hides that the only consumer is `Coalesce.residue`'s `rest.any` (one site). If only one call site survives, inline it or note the single consumer in the comment — currently looks more general-purpose than it is.

### Legitimate open work (correctly documented)

Nothing wrong with these — flagged for context only:

* `Collection.cross_assoc` row component — three nested `▸` casts under `Schema.types_get_append_left`/`_right` at two levels. The multiplicity components (`crossOne_diff_eq`/`_err_diff_eq`) close cleanly via `ring`. Honest open.
* Pushdown recovery windows — blocked on iota reduction of `filterOne`'s match auxiliary post `cases hev`. The inline note at `Mz/Collection.lean:498-523` is excellent triage: lists what was tried, suggests three concrete escape routes. The counterexample (`filter_cross_pushdown_left_unsound`) carries the obstruction's story until recovery lands.
* Variadic / `.divide` static-divisor `outputCols` precision — open follow-up. Tied to mutual recursion mirroring `argsMightError`.
* `LegalEval` variadic short-circuit + collection lift — explicitly scoped down to binary err-payload non-determinism. `model.md:692-701` flags this honestly.
* Existing `=`-tagged rewrites' migration to `Collection.Equiv` — deferred until a forcing function appears. Reasonable.

### Verdict

Mechanization content is back where round-3 left it, with GADT discipline added on top. The remaining blockers are real follow-ups, not regressions. The most actionable cleanup is the doc layer — delete or rewrite `review.md`, sweep the "indexed counterpart of" docstrings, fix the `Mz/Indexed/` path references, reconcile the `eval_satisfies_outputCols` open-vs-mechanized contradiction in `transforms.md`. Mechanically straightforward; matters because the docs are the entry point for anyone landing here cold.

antiguru:
Round-5 review findings addressed (commit `94a94beb03`):

| Finding | Action |
|---|---|
| `review.md` fossilized | Deleted (was untracked). |
| 12 modules with "indexed counterpart of `Mz/X.lean`" self-refs | Swept (Boolean, Coalesce, Datum, EquivBounded, Eval, Equiv, Laws, Legal, MightError, OutputType, Strict, Variadic, Collection). |
| `Mz/Schema.lean:9-14` stale `Mz/Indexed/` paths | Header rewritten to point at actual `Mz/OutputType.lean`. |
| `transforms.md:183` vs `:188` contradiction | Dropped the "open" claim at :183; matches the mechanized claim two lines down. |
| `model.md:651` `Row.refines` typo (`.get i` vs `i`) | Fixed. `Env sch` is a function, not a vector. |
| `Mz/PrimEval.lean:13-21` "three groups" lists four | Bumped to four. |
| `Mz/Datum.lean:48-58` `isNullB` consumer note | Sole consumer (`Coalesce.residue`'s `rest.any` in `Mz/PrimEval.lean`) called out in docstring. |

Build clean (762 jobs). 17 files / 70+ insertions / 96- deletions.

The legitimate open items the review flagged (`cross_assoc` row component, pushdown recovery, variadic `outputCols` precision, `LegalEval` SQL-faithfulness, `=`→`Collection.Equiv` migration) remain — they're correctly scoped follow-ups, not regressions.

doc/developer/semantics/Mz/Eval.lean
```diff
@@ -0,0 +1,62 @@
...
+    | _, env, .minus a b  => evalMinus (eval env a) (eval env b)
+    | _, env, .times a b  => evalTimes (eval env a) (eval env b)
+    | _, env, .divide a b => evalDivide (eval env a) (eval env b)
+    | _, env, .eq a b     => evalEq (eval env a) (eval env b)
+    | _, env, .lt a b     => evalLt (eval env a) (eval env b)
+    | _, env, .andN args  => evalAndN (evalList env args)
+    | _, env, .orN args   => evalOrN (evalList env args)
+    | _, env, .ifThen c t e =>
+      evalIfThen (eval env c) (eval env t) (eval env e)
+    | _, env, .coalesce args => evalCoalesce (evalList env args)
```
mgree: These should all be lazily evaluated (to be faithful---it's safe in Lean no matter what, but especially with Datum::Error).

doc/developer/semantics/Mz/PrimEval.lean
```diff
@@ -0,0 +1,174 @@
...
+
+/-- Right-fold variadic AND. Seed value is `.bool true`. -/
+def evalAndN : List (Datum .bool) → Datum .bool
+  | []        => .bool true
+  | d :: rest => evalAnd d (evalAndN rest)
+
+/-- Right-fold variadic OR. -/
+def evalOrN : List (Datum .bool) → Datum .bool
+  | []        => .bool false
+  | d :: rest => evalOr d (evalOrN rest)
```
mgree, line 144: This _seems_ to be doing lazy eval, but notice that we have a list of `Datum`s---we're already evaled. Ditto `coalesce` and `ifThen`.

doc/developer/semantics/Mz/Collection.lean
```diff
@@ -0,0 +1,674 @@
...
+match scrutinizing `Datum.err e✝` (post-cases substitution)
+stays opaque even after `rw [hev]`. The fix likely requires
+either marking the auto-generated `filterOne.match_1` reducible
+(separate from `filterOne` itself), routing via the auto-gen
+`filterOne.eq_*` equation lemmas via `simp only`, or rewriting
+`filterOne` to dispatch via explicit `Datum.casesOn`. Left as
+follow-up.
+
+The counterexample (`filter_cross_pushdown_left_unsound`) remains
+the load-bearing demonstration of the un-recovered direction. -/
```
mgree, line 522: It's not unsound---just inexact!

mgree (commented):
Comments from our conversation.

antiguru:
Three round-5 review threads addressed:

**C1 (`Mz/Eval.lean:53`) + C2 (`Mz/PrimEval.lean:144`): laziness.** Commit `169d01fc17`.

`.ifThen` now dispatches inline on `eval env c` and evaluates only the selected branch:

```lean
| _, env, .ifThen c t e =>
  match eval env c with
  | .bool true  => eval env t   -- only t
  | .bool false => eval env e   -- only e
  | .null       => .null         -- neither
  | .err err    => .err err     -- neither
```

Eager dispatch through `evalIfThen (eval env c) (eval env t) (eval env e)` computed all three branches regardless. The lazy form matters under `Datum.err _` — an err Datum from a discarded branch should not be observable.

Variadics (`.andN`, `.orN`, `.coalesce`) stay eager via `evalList`, with rationale documented in the `Eval.lean` header:

* `evalAndN`'s right-fold + `evalAnd`'s `(.bool false, _) => .bool false` first pattern produce the same value as left-to-right lazy short-circuit on FALSE. Same for `evalOrN` with TRUE. So eager AndN/OrN is *result-faithful* to lazy short-circuit even though it's not structurally lazy.
* `evalCoalesce`'s two-pass form (`Coalesce.firstConcrete` + `Coalesce.residue`) handles the spec's `null > err` residue. Scanning operands after an err head to detect a later null is required by the spec — structural laziness beyond "stop on first concrete" has no win. The eager 2-pass shape captures the spec directly.

ExprList-direct lazy variants (`evalAndN_lazy`, `evalOrN_lazy`) are queued as a follow-up — would need bridge lemmas to existing absorption proofs, modest scope but not free.

**C3 (`Mz/Collection.lean:522`): "not unsound — just inexact!"** Commit `7fcf4dedc8`.

Renamed `filter_cross_pushdown_left_unsound` → `filter_cross_pushdown_left_inexact`. Docstring expanded:

> The rewrite is not unsound in the strict sense — it never produces a wrong row, it only drops an err that the LHS preserves. Under `=` the two sides differ; under `refines` (errors as bottom) the RHS refines the LHS.

References updated in `model.md` / `README.md` / `transforms.md`.

Proof updates: `might_error_sound .ifThen` and `eval_satisfies_outputCols .ifThen` use a case split on `eval env c` (replacing the eager `evalIfThen_not_err` route). `Legal.legal_of_eval .ifThen` uses `conv_lhs => unfold eval` to selectively unfold the LHS before `rw [h]`.

Build clean (762 jobs).

antiguru:
Follow-up to the monadic pilot (`d05697f8d9`):

**Path taken.** A side-by-side comparison of the GADT and monadic stacks (5 modules: `DatumM`, `CollectionM`, `ExprM`, `EvalM`, plus their consumers) showed the monadic encoding's wins (collapsed null/err in 7 strict primitives, ~28 lines saved) do not offset migrating ~30 theorems across 8 modules, and the more interesting SQL operators (variadic absorption, coalesce residue, filter err-diff) sit outside the monad anyway. One carry-over survives: the *structurally lazy variadic walk*.

**Changes in 79aa3431cd.**

* Cherry-picked lazy variadics into the canonical GADT `Mz/Eval.lean`:
  * `evalAndN_lazy` walks `ExprList` and short-circuits on first `.bool false`.
  * `evalOrN_lazy` is the symmetric short-circuit on `.bool true`.
  * `evalCoalesce_lazy` stops on first concrete operand; `evalCoalesce_null_residue` / `evalCoalesce_err_residue` track `null > err` for the no-concrete case.
* Bridge lemmas `evalAndN_lazy_eq_eager` / `evalOrN_lazy_eq_eager` / `evalCoalesce_lazy_eq_eager` discharge the lazy forms to the list-mediated `evalAndN` / `evalOrN` / `evalCoalesce` in `Mz/PrimEval.lean`. Existing algebraic infrastructure (`evalAndN_not_err_of_all_safe`, `evalCoalesce_cons_concrete`, …) keeps applying through one `rw`.
* Consumers updated in `Mz/MightError.lean`, `Mz/OutputType.lean`, `Mz/Subst.lean`.

**Pilot archive.** `Mz/DatumM.lean` retained as a documented reference for the monadic encoding (header notes adopted laziness as the only carry-over). `Mz/CollectionM.lean`, `Mz/ExprM.lean`, `Mz/EvalM.lean` dropped.

**Notes on the proof refactor.**

* The inner polymorphic-k match in `evalCoalesce_lazy` and its residue helpers is extracted into non-mutual step-helpers (`coalesceLazyStep`, `coalesceNullStep`, `coalesceErrStep`) — Lean's compiled mutual auxiliaries do not iota-reduce a polymorphic-discriminant match under `cases`, and lifting the match out sidesteps that. Laziness is preserved via `Unit → Datum k` thunks; the recursive branches are guarded.
* Mutual-block expansion (from 2 partners to 7) shifts iota behavior. Per-constructor equation lemmas `eval_andN` / `eval_orN` / `eval_coalesce` added for `rw` precision; `Mz/Legal.lean`'s `legal_of_eval` arms wrapped with `unfold eval`; the `filter_cross_pushdown_left_inexact` counterexample switches `decide` → `native_decide` (kernel iota stalls on the larger aux).

Build clean: 762 jobs root, 577 for `DatumM` standalone.

Task #54 (`filter_cross_pushdown` recovery proofs) remains pending, independent of this change.

PR #36617, opened 2026-05-19
mcp: surface data product hydration readiness via mz_mcp_data_product_details

bobbyiliev:
Fixes https://linear.app/materializeinc/issue/DEX-30/expose-data-product-hydration-readiness-state-via-mcp.


def- (commented):
The new hydration aggregation in `src/catalog/src/builtin/mz_internal.rs` merges rows by `(object_name, cluster)` even though `hydration_meta` can emit multiple rows for the same data product/cluster when the object has multiple indexes on that cluster

bobbyiliev:
> The new hydration aggregation in `src/catalog/src/builtin/mz_internal.rs` merges rows by `(object_name, cluster)` even though `hydration_meta` can emit multiple rows for the same data product/cluster when the object has multiple indexes on that cluster

Good catch thanks Dennis. I added a per-replica grouping stage that runs `bool_and` over the index dataflows before counting, so a replica only counts as hydrated when every index dataflow for the data product is hydrated on it, and `replica_count` reflects the actual cluster replica count instead of being multiplied by the number of indexes. The `hydrated` boolean was coincidentally still correct before, but the individual counts were wrong.

def- (approved):

ggevay:
So, the [Linear issue](https://linear.app/materializeinc/issue/DEX-30/expose-data-product-hydration-readiness-state-via-mcp) says

> this returned zero rows because there is no matching data

but actually, querying a non-hydrated thing will always just block, never return partial results. How does this affect the motivation for this PR? I guess this also depends on how timeouts in MCP land currently work, i.e., whether the agent would just get blocked for a long time, or would a timeout typically come back with an error.

bobbyiliev:
> So, the [Linear issue](https://linear.app/materializeinc/issue/DEX-30/expose-data-product-hydration-readiness-state-via-mcp) says
> 
> > this returned zero rows because there is no matching data
> 
> but actually, querying a non-hydrated thing will always just block, never return partial results. How does this affect the motivation for this PR? I guess this also depends on how timeouts in MCP land currently work, i.e., whether the agent would just get blocked for a long time, or would a timeout typically come back with an error.

@ggevay ah yes, you're right, the "zero rows" framing was wrong. Reframed the issue and the tool docs around blocking instead. The column still pays for itself, eg an agent can check before firing a read that'd otherwise just block and time out, and it separates no-replicas from warming-up. Let me know what you think! 

ps commit has been pushed but GH seems to be having some trouble atm.

doc/user/content/reference/system-catalog/mz_internal.md
```diff
@@ -570,6 +572,7 @@ with a JSON Schema describing each data product's columns and types.
 | `cluster`     | [`text`] | Cluster where the object computes or its index is hosted. Reads from any cluster work, but only reads on this cluster benefit from the index. |
 | `description` | [`text`] | Index comment if available, otherwise object comment. Used as data product description.   |
 | `schema`      | [`jsonb`]| JSON Schema describing the object's columns and types.                                   |
+| `hydration`   | [`jsonb`]| Readiness summary as a JSON object with `hydrated` (bool), `replica_count` (int), and `hydrated_replica_count` (int). `hydrated` is true only when the cluster has at least one replica and the dataflow is hydrated on every replica. Agents should back off and retry when `hydrated` is false rather than treating an empty read as final. |
```
ggevay: > rather than treating an empty read as final

(Can't happen that an unhydrated object returns an empty result, as discussed above.)

ggevay: The Linear issue has a motivation for the `replica_count` field; would be good to add that here as well:

> if replica_count is 0 that needs operator action, not a retry

src/catalog/src/builtin/mz_internal.rs
```diff
@@ -5342,9 +5348,13 @@ pub static MZ_MCP_DATA_PRODUCT_DETAILS: LazyLock<BuiltinView> = LazyLock::new(||
             "schema",
             "JSON Schema describing the object's columns and types.",
         ),
+        (
+            "hydration",
+            "Readiness summary as a JSON object with `hydrated` (bool), `replica_count` (int), and `hydrated_replica_count` (int). `hydrated` is true only when the cluster has at least one replica and the dataflow is hydrated on every replica. Agents should back off and retry when `hydrated` is false rather than treating an empty read as final.",
```
ggevay: (Same comments here as in the `md`.)

src/environmentd/src/http/mcp.rs
```diff
@@ -658,7 +662,12 @@ fn endpoint_instructions(endpoint_type: McpEndpointType) -> Option<String> {
             "(read from persistent storage). `read_data_product` automatically routes the ",
             "read to the cluster recorded in the data product catalog so indexes are used; ",
             "you only need to set the `cluster` parameter if you intentionally want the ",
-            "read to run on a different cluster (e.g. one with larger or more replicas).",
+            "read to run on a different cluster (e.g. one with larger or more replicas). ",
+            "`get_data_product_details` returns a `hydration` object with `hydrated`, ",
+            "`replica_count`, and `hydrated_replica_count` fields: if `hydrated` is false, ",
+            "the dataflow is still warming up. Back off and retry rather than looping on ",
+            "empty results — an empty answer from a not-yet-hydrated product is not the ",
+            "same as a genuinely empty result.",
```
ggevay: (Same comments also here.)

src/environmentd/src/http/mcp.rs
```diff
@@ -715,7 +724,7 @@ async fn handle_tools_list(
                 ToolDefinition {
                     name: "get_data_product_details".to_string(),
                     title: Some("Get Data Product Details".to_string()),
-                    description: "Get the complete schema and structure of a specific data product. This shows you exactly what fields are available, their types, and what data you can query. Use this after finding a data product from get_data_products() to understand how to query it.".to_string(),
+                    description: "Get the complete schema and structure of a specific data product, plus a `hydration` object reporting whether the dataflow is fully hydrated across the cluster's replicas (`{hydrated, replica_count, hydrated_replica_count}`). This shows you exactly what fields are available, their types, what data you can query, and whether the data product is ready to serve fresh results. Use this after finding a data product from get_data_products() to understand how to query it; if `hydrated` is false, back off and retry rather than treating empty reads as final.".to_string(),
```
ggevay: (Same comments also here.)

test/mcp/mzcompose.py
```diff
@@ -479,3 +479,165 @@ def workflow_default(c: Composition) -> None:
 
         r = post_mcp(c, "agent", jsonrpc("tools/list"))
         assert r.status_code == 200
+
+    # -- hydration: end-to-end coverage for DEX-30 ----------------------------
+    #
+    # Verifies that `mz_internal.mz_mcp_data_product_details` exposes a
+    # `hydration` JSON column that the agent endpoint surfaces through
+    # `get_data_product_details`, so agents can distinguish "still warming
+    # up" from "genuinely empty." Two scenarios:
```
ggevay: (Same comment about "empty" also here.)

doc/user/content/reference/system-catalog/mz_internal.md
```diff
@@ -561,7 +561,9 @@ schema information.
 ## `mz_mcp_data_product_details`
 
 The `mz_mcp_data_product_details` view extends [`mz_mcp_data_products`](#mz_mcp_data_products)
-with a JSON Schema describing each data product's columns and types.
+with a JSON Schema describing each data product's columns and types, and a
+readiness summary that lets agents tell "still warming up" apart from
+"genuinely empty."
```
ggevay: (Same comment about "empty" also here.)

src/environmentd/src/http/mcp.rs
```diff
@@ -68,7 +68,11 @@ const MCP_REQUEST_TIMEOUT: Duration = Duration::from_secs(60);
 
 // Discovery uses the lightweight view (no JSON schema computation).
 const DISCOVERY_QUERY: &str = "SELECT * FROM mz_internal.mz_mcp_data_products";
-// Details uses the full view with JSON schema.
+// Details uses the full view, which also exposes a `hydration` JSON column
+// (`{hydrated, replica_count, hydrated_replica_count}`) so agents can decide
+// whether to back off and retry vs treat empty reads as final. See DEX-30
+// and the comments on `mz_mcp_data_product_details` in
+// `src/catalog/src/builtin/mz_internal.rs`.
```
ggevay: (Same comment about "empty" also here.)

ggevay: Btw. do we need this code comment? There is already some repetition even without this.

src/environmentd/tests/server.rs
```diff
@@ -5547,6 +5547,53 @@ fn test_mcp_agent_with_data_product() {
     assert!(body["result"]["content"][0]["text"].as_str().is_some());
     assert!(body["error"].is_null());
 
+    // The response should expose a `hydration` field per row (5th cell) so
+    // agents can decide whether to back off or treat empty reads as final.
+    // See DEX-30. For an MV that's had time to hydrate on a single-replica
+    // `quickstart` cluster, expect `hydrated: true` with 1/1 replicas.
```
ggevay: (Same comment about "empty" also here.)

ggevay: Also, maybe you could shorten the comment a bit, eliminating some redundancy with other places.

test/mcp/mzcompose.py
ggevay: And again, I guess the comment could be shortened, eliminating repetition.

ggevay (approved):
LGTM, just minor comments.

PR #36620, opened 2026-05-19
SS-152 fix Kafka low watermark logs spam

patrickwwbutler:
Quick patch for https://linear.app/materializeinc/issue/SS-152/investigate-excessive-worker-pid-pairs-in-kafka-test


martykulma (approved):
🦈

PR #36623, opened 2026-05-19
docs: PrivateLink troubleshooting for NLB security group enforcement

materializekatrina:
## Motivation

Customer hit a silent PrivateLink failure: `psql` from their bastion worked, but Materialize source creation failed. Their NLB security group had **Enforce inbound rules on PrivateLink traffic** enabled with rules that only allowed their own test IPs — traffic from Materialize's VPC endpoint arrives with different source IPs, so it was silently dropped.

Our docs don't mention this, and still claim "NLBs do not have associated security groups" (no longer true since AWS added the feature in 2023).

Slack thread: https://materializeinc.slack.com/archives/C08K5GHRV1S/p1779207642559229

## Changes

In each of the three PrivateLink setup shortcodes (PostgreSQL, MySQL, Kafka):

- Soften remark (a.) to "...do not have associated security groups **by default**."
- Add remark (c.) describing the **Enforce inbound rules on PrivateLink traffic** failure mode and how to fix it.

### DIFF
--- a/doc/user/layouts/shortcodes/network-security/privatelink-kafka.md
+++ b/doc/user/layouts/shortcodes/network-security/privatelink-kafka.md
@@ -44,9 +44,11 @@ and retrieve the AWS principal needed to configure the AWS PrivateLink service.
 
     **Remarks**:
 
-    a. Network Load Balancers do not have associated security groups. Therefore, the security groups for your targets must use IP addresses to allow traffic.
+    a. Network Load Balancers do not have associated security groups by default. Therefore, the security groups for your targets must use IP addresses to allow traffic.
 
     b. You can't use the security groups for the clients as a source in the security groups for the targets. Therefore, the security groups for your targets must use the IP addresses of the clients to allow traffic. For more details, check the [AWS documentation](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-register-targets.html).
+
+    c. If you have associated a [security group with your Network Load Balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-security-groups.html) and enabled **Enforce inbound rules on PrivateLink traffic**, the NLB's inbound rules will be applied to traffic arriving from Materialize's VPC endpoint. In that case, the source IPs are private IPs from Materialize's VPC — **not** the IPs of any client (e.g., `kcat` or a Kafka client running from a workstation or bastion host) that you might use to verify connectivity to the brokers directly. Inbound rules that only permit your own test traffic will silently block Materialize, even when your own connectivity tests succeed. To resolve this, either disable **Enforce inbound rules on PrivateLink traffic**, or add inbound rules to the NLB's security group that permit the relevant ports (each broker's listener port and the health check port) from a source that covers Materialize's VPC endpoint traffic.
 
 1. Create a VPC [endpoint service](https://docs.aws.amazon.com/vpc/latest/privatelink/create-endpoint-service.html) and associate it with the **Network Load Balancer** that you’ve just created.
 
--- a/doc/user/layouts/shortcodes/network-security/privatelink-mysql.md
+++ b/doc/user/layouts/shortcodes/network-security/privatelink-mysql.md
@@ -19,9 +19,11 @@
 
     **Remarks**:
 
-    a. Network Load Balancers do not have associated security groups. Therefore, the security groups for your targets must use IP addresses to allow traffic.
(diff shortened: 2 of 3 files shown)


doc/user/layouts/shortcodes/network-security/privatelink-kafka.md
```diff
@@ -44,10 +44,12 @@ and retrieve the AWS principal needed to configure the AWS PrivateLink service.
 
     **Remarks**:
 
-    a. Network Load Balancers do not have associated security groups. Therefore, the security groups for your targets must use IP addresses to allow traffic.
+    a. Network Load Balancers do not have associated security groups by default. Therefore, the security groups for your targets must use IP addresses to allow traffic.
```
kay-kim, line 47: Double-checking ... this is pre-change content ... but since we're here: 
The "Therefore, " sentence , is this the same or different as point b) below?

Will also ask Justin.

doc/user/layouts/shortcodes/network-security/privatelink-kafka.md
```diff
@@ -44,10 +44,12 @@ and retrieve the AWS principal needed to configure the AWS PrivateLink service.
...
 
+    c. If you have associated a [security group with your Network Load Balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-security-groups.html) and enabled **Enforce inbound rules on PrivateLink traffic**, the NLB's inbound rules will be applied to traffic arriving from Materialize's VPC endpoint. In that case, the source IPs are private IPs from Materialize's VPC — **not** the IPs of any client (e.g., `kcat` or a Kafka client running from a workstation or bastion host) that you might use to verify connectivity to the brokers directly. Inbound rules that only permit your own test traffic will silently block Materialize, even when your own connectivity tests succeed. To resolve this, either disable **Enforce inbound rules on PrivateLink traffic**, or add inbound rules to the NLB's security group that permit the relevant ports (each broker's listener port and the health check port) from a source that covers Materialize's VPC endpoint traffic.
```
kay-kim, line 51: So ... we seem to be modifying content shown on 
https://preview.materialize.com/materialize/36623/ingest-data/network-security/privatelink/

Did we want to update the content in https://preview.materialize.com/materialize/36623/ingest-data/postgres/amazon-rds/#b-optional-configure-network-security  page and such?

kay-kim, line 51: Also, maybe make it leaner? And perhaps list the disabling to be the second bullet point(?) since it seems kind of permissive for all inbound traffic and not just materialize's traffic.

```
If you have associated a security group with your Network Load Balancer and
enabled **Enforce inbound rules on PrivateLink traffic**, the security group's
inbound rules also apply to traffic from Materialize's VPC endpoint. Any traffic
not explicitly permitted, including Materialize's, will be silently blocked.

To resolve this either:
- Add inbound rules to the NLB's security group that permit the listener port
  and the health check port from a source covering Materialize's VPC endpoint
  traffic, or
- Disable **Enforce inbound rules on PrivateLink traffic**.
```

jubrad: +1 to kays comment, then I think good to go.

doc/user/layouts/shortcodes/network-security/privatelink-mysql.md
```diff
@@ -19,10 +19,12 @@
 
     **Remarks**:
 
-    a. Network Load Balancers do not have associated security groups. Therefore, the security groups for your targets must use IP addresses to allow traffic.
+    a. Network Load Balancers do not have associated security groups by default. Therefore, the security groups for your targets must use IP addresses to allow traffic.
```
kay-kim, line 22: We seem to talk about NLBs here before we create the NLB.

doc/user/layouts/shortcodes/network-security/privatelink-postgres.md
```diff
@@ -19,10 +19,12 @@
 
     **Remarks**:
 
-    a. Network Load Balancers do not have associated security groups. Therefore, the security groups for your targets must use IP addresses to allow traffic.
+    a. Network Load Balancers do not have associated security groups by default. Therefore, the security groups for your targets must use IP addresses to allow traffic.
```
kay-kim, line 22: Ditto on this page ... we talk about NLBs here and then, we create the NLB in the next step.

kay-kim:
Closing since https://github.com/MaterializeInc/materialize/pull/37056 supersedes this.

PR #36625, opened 2026-05-19
dbt-materialize: support unmanaged clusters in deploy_init

sjwiesman:
The deploy_init macro previously hard-failed when the production cluster was unmanaged, blocking blue/green deployments for any Materialize user not on a managed cluster. Add a dispatch path that, for unmanaged origins, queries mz_cluster_replicas and emits CREATE CLUSTER ... REPLICAS (...) cloning each user-facing replica's SIZE and AVAILABILITY ZONE. Internal replicas are filtered via mz_internal_cluster_replicas. Replicas without a SIZE (orchestrated replicas using COMPUTECTL/STORAGECTL ADDRESSES) cannot be faithfully cloned and now produce a compiler error that names the offending replica.

The new create_unmanaged_cluster macro mirrors the structure of the existing create_cluster macro (pre-existence check, CI tag re-entry, DDL emission, CI tag set) so the managed path is untouched.

### Motivation

Closes <issue>
Fixes <bug>
etc.

### Description

diff *can't* tell a reviewer.

### Verification

### DIFF
--- a/misc/dbt-materialize/CHANGELOG.md
+++ b/misc/dbt-materialize/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Support unmanaged clusters in `deploy_init`. The deployment cluster
   is now created by cloning the production cluster's replicas (including
-  `SIZE` and `AVAILABILITY ZONE`). 
+  `SIZE` and `AVAILABILITY ZONE`).
 
 ## 1.9.9 - 2026-05-18
 
--- a/misc/dbt-materialize/CHANGELOG.md
+++ b/misc/dbt-materialize/CHANGELOG.md
@@ -2,12 +2,9 @@
 
 ## Unreleased
 
-* Support unmanaged origin clusters in `deploy_init`. The deployment
-  cluster is now created by cloning the production cluster's replicas
-  (including SIZE and AVAILABILITY ZONE attributes). Internal replicas
-  are skipped. `deploy_init` errors with a clear message if any
-  user-facing replica lacks a SIZE (orchestrated replicas using
-  COMPUTECTL/STORAGECTL ADDRESSES cannot be cloned).
+* Support unmanaged clusters in `deploy_init`. The deployment cluster
+  is now created by cloning the production cluster's replicas (including
+  `SIZE` and `AVAILABILITY ZONE`). 
 
 ## 1.9.9 - 2026-05-18
 
--- a/misc/dbt-materialize/CHANGELOG.md
+++ b/misc/dbt-materialize/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+* Support unmanaged origin clusters in `deploy_init`. The deployment
+  cluster is now created by cloning the production cluster's replicas
+  (including SIZE and AVAILABILITY ZONE attributes). Internal replicas
+  are skipped. `deploy_init` errors with a clear message if any
+  user-facing replica lacks a SIZE (orchestrated replicas using
+  COMPUTECTL/STORAGECTL ADDRESSES cannot be cloned).
+
 ## 1.9.9 - 2026-05-18
 
 * Add support for the `partition_by` configuration in materialized views,
--- a/misc/dbt-materialize/dbt/include/materialize/macros/ci/create_unmanaged_cluster.sql
+++ b/misc/dbt-materialize/dbt/include/materialize/macros/ci/create_unmanaged_cluster.sql
@@ -1,0 +1,156 @@
+-- Copyright Materialize, Inc. and contributors. All rights reserved.
+--
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License in the LICENSE file at the
+-- root of this repository, or online at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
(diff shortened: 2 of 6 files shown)


misc/dbt-materialize/dbt/include/materialize/macros/ci/create_unmanaged_cluster.sql
```diff
@@ -0,0 +1,156 @@
...
+        {{ exceptions.raise_compiler_error("cluster_name must be provided") }}
+    {% endif %}
+
+    {% if not origin_cluster %}
+        {{ exceptions.raise_compiler_error("origin_cluster must be provided") }}
+    {% endif %}
+
+    {% set deploy_cluster = adapter.generate_final_cluster_name(cluster_name, force_deploy_suffix) %}
+
+    {% if cluster_exists(deploy_cluster) %}
```
bobbyiliev, line 50: Most of this block looks like a copy of `create_cluster.sql. Not sure if it's worth pulling into a shared helper so the two paths can't drift?

misc/dbt-materialize/dbt/include/materialize/macros/ci/create_unmanaged_cluster.sql
```diff
@@ -0,0 +1,156 @@
...
+                    {{ exceptions.raise_compiler_error("
+                        Deployment cluster " ~ deploy_cluster ~ " already exists and is not empty.
+                        This is potentially dangerous as you may end up deploying objects to production you
+                        do not intend.
+
+                        If you are certain the objects in this cluster are supposed to exist, you can ignore this
+                        error by setting ignore_existing_objects to True.
+
+                        dbt run-operation create_unmanaged_cluster --args '{ignore_existing_objects: True}'
+                    ") }}
```
bobbyiliev: The hint suggests `dbt run-operation create_unmanaged_cluster ..` but if I'm reading this correctly the macro also needs `cluster_name` and `origin_cluster`. Would it read better as "re-run `deploy_init` with `ignore_existing_objects: True`"?

misc/dbt-materialize/dbt/include/materialize/macros/ci/create_unmanaged_cluster.sql
```diff
@@ -0,0 +1,156 @@
...
+                {% if row[1] is none %}
+                    {{ exceptions.raise_compiler_error(
+                        "Cannot clone replica '" ~ row[0] ~ "' of unmanaged cluster '"
+                        ~ origin_cluster ~ "': it has no SIZE. Orchestrated replicas "
+                        ~ "(using COMPUTECTL/STORAGECTL ADDRESSES) cannot be cloned by "
+                        ~ "blue/green deployments. Recreate the cluster as managed or "
+                        ~ "replace this replica with a sized one."
+                    ) }}
+                {% endif %}
+            {% endfor %}
```
bobbyiliev, line 73: Maybe we could collect all unsized replicas and raise once, instead of failing on the first one? Just nicer if a user has more than one to fix.

misc/dbt-materialize/dbt/include/materialize/macros/ci/create_unmanaged_cluster.sql
```diff
@@ -0,0 +1,156 @@
...
+            {% for row in replica_results.rows %}
+                {% set parts = ["SIZE " ~ dbt.string_literal(row[1])] %}
+                {% if row[2] is not none %}
+                    {% do parts.append("AVAILABILITY ZONE " ~ dbt.string_literal(row[2])) %}
+                {% endif %}
+                {% do replica_clauses.append(adapter.quote(row[0]) ~ " (" ~ parts | join(", ") ~ ")") %}
+            {% endfor %}
+
+            {% set create_cluster_ddl %}
+                CREATE CLUSTER {{ adapter.quote(deploy_cluster) }} REPLICAS (
```
bobbyiliev, line 95: This uses `adapter.quote(deploy_cluster)` but `create_cluster.sql` line 114 emits the name bare. Should we normalize them while we're here?

misc/dbt-materialize/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+* Support unmanaged origin clusters in `deploy_init`. The deployment
+  cluster is now created by cloning the production cluster's replicas
+  (including SIZE and AVAILABILITY ZONE attributes). Internal replicas
```
bobbyiliev: Tiny nit: "Internal replicas are skipped" might confuse users who don't know what counts as internal. Worth a quick clarifier, or fine to leave?

bobbyiliev (changes requested):

sjwiesman:
@bobbyiliev thank you for the speedy review. I believe I've addressed your comments.

misc/dbt-materialize/CHANGELOG.md
sjwiesman: dropped.

misc/dbt-materialize/dbt/include/materialize/macros/ci/create_unmanaged_cluster.sql
sjwiesman: its worth it. made the change.

bobbyiliev (approved):
Looks good! There seem to be some minor lint CI failures but should be good to go after that.

PR #36627, opened 2026-05-19
Column paged merge batcher

DAlperin:
### Motivation

The merge-batcher's transient state (chains of sealed input chunks awaiting merge / extract) sits between input ingestion and the spine. Under memory pressure that transient peak is what trips OOMs, not the spine itself — the spine has already consolidated.

This PR plugs `ColumnPager` (#36552) into the merge-batcher so those chain entries can live on disk (or compressed in RAM) instead of resident memory, while leaving the hot-path merge / extract logic and on-wire chunk shape unchanged.

### Description

A new `ColumnMergeBatcher` in `mz-timely-util` forks the DD merge-batcher framework so chain entries are `PagedColumn` (`Resident` / `Paged` / `Compressed`) rather than resident `Column`s. Streaming drivers `merge_chains` / `extract_chain` walk those chains via a `FetchIter` that materializes one head per side on demand and hands finished output back to the pager. The per-chunk merge / extract logic itself is the same `Column::merge_from` / `Column::extract` already used by the in-memory `ColumnMerger`.

Compute integration:

- New dyncfgs in `compute-types`: `enable_column_paged_batcher` (off by default — controls whether the installed pager actually evicts) and `column_paged_batcher_budget_fraction` (default 5% of replica memory; floored at 128 MiB so the no-pressure case doesn't page per chunk). Pager backend follows `--scratch-directory` availability the same way `mz_ore::pager` already does (file when a scratch directory is configured, swap otherwise).
- `apply_worker_config` derives `(enabled, total, backend)` from those dyncfgs and routes through `column_pager::apply_tiered_config`, which holds a process-wide `TieredPolicy` singleton and mutates its atomic budget / backend / codec in place. In-flight `ResidentTicket`s keep crediting the same atomic across reconfigures, so operator-driven tunes (or other workers in the same process reapplying the same config) don't orphan accounting onto a stale policy.
- Two arrange call sites swapped from `Col2ValBatcher` to `Col2ValPagedBatcher`: `arrange_collection` in `render/context.rs` and the `JoinStage` arrange in `render/join/linear_join.rs`. Logging arranges stay on the legacy `ColumnMerger` path.
- New `Col2ValPagedBatcher` type alias and a `BuilderInput for Column<((K, V), T, R)>` impl so the DD `OrdValBuilder` can consume the batcher's `Column` output without a container conversion. `RowRowColPagedBuilder` + a couple of small `PushInto` / `PartialEq` impls in `row_spine.rs` get the Row-keyed paths type-checking.
- `Materialized` and `Clusterd` mzcompose services gain `memory_swap` and `mem_swappiness` so feature-benchmarks can configure container swap behavior independently of the batcher.

Observability (last commit on the stack): `column_pager::metrics` registers a `PagerMetrics` struct with the process metrics registry. Counters cover skip / pageout / pagein decisions and bytes through each path; computed gauges expose `mz_column_pager_budget_remaining_bytes` and `mz_column_pager_budget_configured_bytes` against the live `TieredPolicy` atomics. Compute init wires the registration; the timely-util module is registry-agnostic, so callers without a registry (tests, benches, examples) just see no-op observers.

### Verification

- **Unit tests** (`mz-timely-util`, columnar + column_pager modules): 46 tests covering chunker correctness, per-chunk merge / extract, drain, `ColumnMergeBatcher` end-to-end seal under `ColumnPager::disabled()` and a `ForcePagePolicy` that forces every chunk through the pager. New `TieredPolicy::reconfigure` tests cover in-flight ticket preservation across pool resizes, shrink-saturates-at-zero, and live backend/codec swap. Proptests cover merge / extract invariants.
- **Criterion microbench** ([`benches/columnar_merge_batcher.rs`](src/timely-util/benches/columnar_merge_batcher.rs)): compares legacy `ColumnMerger` against the paged path with disabled / swap / lz4 across mixed / collision / disjoint inputs and four cache-tier sizes; prints a throughput summary table.
- **End-to-end example** ([`examples/column_paged_spill.rs`](src/timely-util/examples/column_paged_spill.rs)): drives `arrange_core` over a cancellation workload (positives + negatives at the same time so the spine stays empty and all pressure lives in the batcher). Back-to-back baseline + spill modes with an optional RSS sampler thread.
- **Feature-benchmark scenarios**:
  - `DifferentialJoinColumnPaged` measures steady-state overhead vs. `DifferentialJoin`.
  - `DifferentialJoinHydrationBaseline` / `DifferentialJoinHydrationFile` measure re-hydration time after `REPLICATION FACTOR 0 → 1` toggling. Run under `--this-memory` + `--this-memory-swap` to compare user-space spill against OS swap.

### Risk

**The batcher type swap at the two arrange sites is unconditional.** `enable_column_paged_batcher` only controls whether the pager actually evicts — with the flag off, chunks stay in `PagedColumn::Resident` and behave like the legacy in-memory path, but the surrounding code is the new `ColumnMergeBatcher` (chunker, chain layout, `merge_chains` driver, `extract_chain` driver, `BatcherEvent` accounting). That code path is exercised in every dataflow that hits `arrange_collection` or `JoinStage` from day one of this merging, flag or no flag. If the new batcher misbehaves, mitigation is a revert, not a flag flip.

**Whole-chunk passthrough is lost on the new path.** Legacy `ColumnMerger::merge` does a two-probe passthrough that hands disjoint-key chains directly to output with no per-record work. `merge_chains` doesn't implement that yet (peeking endpoints on a paged head would force materialization with no clean way to undo it). Workloads with disjoint key ranges across chains — monotonic source ingestion, partitioned upserts, time-windowed batches — regress to the per-record merge path even with the feature flag off. The criterion bench compares paged-vs-paged variants and doesn't cover this regression vs. the legacy passthrough; a future change can restore passthrough by gating it on `PagedColumn::Resident` heads or by carrying first/last keys in the pager metadata.

Logging arranges still use the legacy `ColumnMerger` path. Bytes paged to swap / file don't enter the `mz_arrangement_batcher_*_raw` RSS-shaped accounting tables, matching how those tables already treat shipped chunks.


src/timely-util/src/columnar/batcher.rs
```diff
@@ -585,7 +584,13 @@ where
             && !crate::columnar::at_serialized_capacity(&ship_c.borrow())
         {
             let (_, time, _) = self_view.get(*position);
-            T::copy_from(&mut owned_t, time);
```
antiguru, line 588: Restore `copy_from`.

src/timely-util/src/columnar/merge_batcher.rs
```diff
@@ -0,0 +1,945 @@
...
+            logger,
+            operator_id,
+        }
+    }
+
+    fn push_container(&mut self, container: &mut Self::Input) {
+        let pager = self.pager();
+        self.chunker.push_into(container);
+        while let Some(chunk) = self.chunker.extract() {
+            let mut chunk = std::mem::take(chunk);
```
antiguru: Check if we can just use the `chunk` we got from the chunker instead of taking.

src/timely-util/src/columnar/merge_batcher.rs
```diff
@@ -0,0 +1,945 @@
...
+            let bytes = col.length_in_bytes();
+            (records, bytes, bytes, 1)
+        }
+        PagedColumn::Paged { .. } | PagedColumn::Compressed { .. } => (0, 0, 0, 0),
+    }
+}
+
+impl<D, T, R> Batcher for ColumnMergeBatcher<D, T, R>
+where
+    D: Columnar + 'static,
```
DAlperin: are these trait bounds overcomplicated, why ref to owned type?

src/compute/src/compute_state.rs
```diff
@@ -308,6 +308,68 @@ impl ComputeState {
...
+                }
+                "swap" => Backend::Swap,
+                other => {
+                    warn!(
+                        backend = %other,
+                        "unknown column_paged_batcher_backend; using swap"
+                    );
+                    Backend::Swap
+                }
+            };
```
antiguru: Could we simplify this to if the scratch directory is set use file, otherwise swap? Only local testing might have neither, but in this case the swap backend wouldn't be wrong.

src/compute/src/compute_state.rs
```diff
@@ -308,6 +308,68 @@ impl ComputeState {
...
+            let total = usize::cast_lossy(f64::cast_lossy(mem_limit) * fraction).max(128 * MIB);
+
+            let backend_str = COLUMN_PAGED_BATCHER_BACKEND.get(config);
+            let backend = match backend_str.as_str() {
+                "file" => {
+                    if let Some(path) = &self.context.scratch_directory {
+                        // `set_scratch_dir` is process-wide and idempotent
+                        // (per-process subdir under the given root), so calling
+                        // it from every worker on every config apply is safe.
+                        pager::set_scratch_dir(path.clone());
```
antiguru: This should be handled already once rebased.

src/timely-util/src/column_pager.rs
```diff
@@ -215,6 +215,97 @@ impl ColumnPager {
...
+    *GLOBAL_PAGER.write().expect("global pager poisoned") = pager;
+}
+
+/// Process-wide decision counters. Diagnostic only — log a summary every
+/// `DECISION_LOG_INTERVAL` Page decisions so we can tell whether the
+/// pager is actually engaging without per-call log spam.
+static SKIP_COUNT: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
+static PAGE_COUNT: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
+static SKIP_BYTES: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
+static PAGE_BYTES: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
```
antiguru: Probably want metrics here.

antiguru (approved):
Seems good, left some smaller comments. Let's go!

antiguru (commented):
## Code review

Thorough engineering and unusually good docs/test coverage here. My concerns are mostly about a **mismatch between the described design and the actual code**, plus an acknowledged perf regression.

### Major findings

**1. The pager-reconfiguration design in the description isn't implemented.**

The description states `apply_worker_config` routes through `column_pager::apply_tiered_config`, which "holds a process-wide `TieredPolicy` singleton and mutates its atomic budget / backend / codec in place" so "in-flight `ResidentTicket`s keep crediting the same atomic across reconfigures."

That code doesn't exist. `policy.rs` is unmodified — there's no `apply_tiered_config` and no `TieredPolicy::reconfigure` (nor the "reconfigure" tests the Verification section claims). What's actually there (`compute_state.rs`) builds a **brand-new** policy every time:

```rust
let policy = Arc::new(TieredPolicy::new(total, backend, None));
set_global_pager(ColumnPager::new(policy));
```

`apply_worker_config` runs on every `UpdateConfiguration` command, which fires for *any* cluster-wide dyncfg change, not just these three. So each config update resets the new policy's resident-bytes atomic to **0**, while in-flight `PagedColumn::Resident` tickets still hold an `Arc` to the **old** policy and release their bytes there — exactly the orphaning the description claims to avoid. The new policy under-counts true resident usage and may never decide to spill, defeating the feature's purpose when enabled. Blast radius is low today (flag off → everything stays `Resident`), but this needs reconciling before the flag is turned on: either implement the in-place `reconfigure`, or rewrite the description to match the "reinstall" reality and add a test for accounting across reinstalls.

**2. Unconditional batcher swap + lost whole-chunk passthrough (acknowledged).**

Both arrange sites swap to `Col2ValPagedBatcher` unconditionally; the flag only gates eviction. The legacy `ColumnMerger` two-probe passthrough isn't reimplemented in `merge_chains`, so disjoint-key workloads (monotonic ingestion, partitioned upserts, time-windowed batches) regress to per-record merge **even with the flag off**, in every dataflow hitting `arrange_collection`/`JoinStage`. The Risk section is honest about this, but the criterion bench only compares paged-vs-paged and doesn't measure the regression vs the legacy passthrough. Since the swap is unconditional, I'd want a benchmark quantifying the disjoint-key regression vs `main` before merge — that number decides whether this can land ahead of the passthrough follow-up.

### Minor

- **Codec isn't configurable in production.** `compute_state` hardcodes `TieredPolicy::new(total, backend, None)` — no dyncfg for codec, so lz4 (covered in the bench) can't be enabled in prod. The description's "derives `(enabled, total, backend, codec)`" overstates what's wired.
- **Per-worker reinstall isn't idempotent in the accounting sense.** Comments in `column_pager.rs` call the repeated `set_global_pager` calls "idempotent (same Arc-equivalent state)," but each worker constructs a *distinct* `TieredPolicy` with independent atomics; last-writer-wins, and tickets minted against a replaced policy are orphaned. Tied to #1.
- **`file` backend fallback.** Requesting `file` without `--scratch-directory` falls back to swap with a `warn!`. Worth confirming `DifferentialJoinHydrationFile` actually configures a scratch dir, else it silently measures swap.

### Positives

- `account_chunk` correctly reports paged/compressed entries as zero RSS and uses saturating arithmetic defensively.
- `FetchIter::pager()` returning `&'a ColumnPager` tied to the outer lifetime (not `&self`) is a clean way to let drivers re-page output across `next()`.
- The `bounds_checks` module in `row_spine.rs` is a good compile-time guard against batcher/builder output-input drift.

### Recommendation

Solid foundation, but I'd hold merge: the description and implementation disagree on the most safety-relevant piece (pager reconfiguration / budget accounting), and the unconditional passthrough regression lacks a vs-legacy benchmark. Both are blocking for enabling the flag; fine to resolve in follow-ups as long as the description is corrected so reviewers aren't misled about what's wired.

---

PR #36628, opened 2026-05-19
Deletes features flags for shipped features

martykulma:
These features are enabled globally, so removing these flags from the code.

- enable_copy_to_expr - support for `COPY TO 's3://..'`
- enable_aws_msk_iam_auth - support for IAM authentication with MSK
- enable_copy_from_remote - support for `COPY FROM <url>`
- enable_iceberg_sink - support for Iceberg table as a SINK


patrickwwbutler (approved):
LGTM

PR #36629, opened 2026-05-19
edu-27: part 1 of docs updates per edu findings

kay-kim:


doc/user/content/headless/materialize-window-functions.md
```diff
@@ -0,0 +1,18 @@
+---
+headless: true
+---
+For indexed views and materialized views that contain [window
+functions](/sql/functions/#window-functions) (including aggregate functions used
+with an `OVER` clause), when an input record in a partition is
+added/removed/changed, Materialize **recomputes the results from scratch** for
+that partition (instead of using incremental computation).
```
kay-kim, line 8: I did remove the [example blurb about 10 partitions with 100 records](https://github.com/MaterializeInc/materialize/pull/36629/changes#diff-c975e611330dc867d327f6c655c5f05129a8af9a1bc19caf8bd91fd2d74d2bde)  because it kind of the work is proportional to running the window function over 1000 records instead of 10 x (window function over 100) which depending on the window function could be more than running window function over 1000 records.

ggevay (approved):

PR #36631, opened 2026-05-20
adapter: rename staged_cancellation to connection_cancel_watches

aljoscha:
The map of per-connection cancel watch channels was named for its historical sole consumer (staged sequencing in `sequence_staged` / `handle_spawn`), but the mechanism itself is generic: any code path that wants a connection-scoped "has this been cancelled?" signal can install/remove an entry. `handle_privileged_cancel` already pokes the sender from outside the staged path.

Rename the field to `connection_cancel_watches` and rewrite the doc comment to describe the abstraction (a per-connection watch consumed by whatever has cancellable work in flight) rather than naming a single caller. Also tighten the docs on `sequence_staged` / `handle_spawn` to point at the renamed map.

### DIFF
--- a/src/adapter/src/coord.rs
+++ b/src/adapter/src/coord.rs
@@ -1885,9 +1885,14 @@ pub struct Coordinator {
     /// to stage Batches in Persist that we will then link into the shard.
     active_copies: BTreeMap<ConnectionId, ActiveCopyFrom>,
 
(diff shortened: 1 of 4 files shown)


ggevay (approved):

aljoscha:
tyty! :raised_hands:

PR #36633, opened 2026-05-20
Sql-317: Attach metric name labels

SangJunBak:
Attaches names of clusters/objects from the catalog via rules per metric. See commit messages for details

### Motivation
Fixes sql-317. This was one of the requirements for the project given otherwise, users would need to do joins in promql against some metric that represents the catalog (what we do today in Cloud). However, this join technique wouldn't work in promsql->datadog.


src/compute-client/src/metrics.rs
```diff
@@ -74,119 +74,157 @@ pub struct ComputeControllerMetrics {
...
 impl ComputeControllerMetrics {
     /// Create a metrics instance registered into the given registry.
     pub fn new(metrics_registry: &MetricsRegistry, shared: ControllerMetrics) -> Self {
-        ComputeControllerMetrics {
+        let metrics = ComputeControllerMetrics {
             commands_total: metrics_registry.register(metric!(
                 name: "mz_compute_commands_total",
                 help: "The total number of compute commands sent.",
                 var_labels: ["instance_id", "replica_id", "command_type"],
+                rules: [instance_name_rule(), replica_name_rule()],
```
mtabebe, line 100: Is it possible to infer the var_labels from the rules? That is, it seems like duplicate information here?

src/http-util/src/lib.rs
```diff
@@ -159,7 +173,20 @@ pub async fn handle_prometheus(
         ContentType::text()
     };
 
-    Ok((TypedHeader(content_type), buf).into_response())
+    let mut resp = (TypedHeader(content_type), buf).into_response();
+    if wants_enrich_rules(&headers) {
+        let rules_by_metric = registry.rules_by_metric();
+        if !rules_by_metric.is_empty() {
+            let json = serde_json::to_string(&rules_by_metric)
```
mtabebe: Ok so the way I understand this is that we always return these rules that get built up and stored. That is, the size of the map that we return is based on the number of metrics with rules. Are there any concerns with size that we are adding to the response header?

How big is the size of each rule? How many rules are we adding?

In any case, do we need to bound the size here?

src/environmentd/src/http/metrics_public.rs
```diff
@@ -10,31 +10,99 @@
...
+        self.0
+            .try_get_cluster_replica(cid, rid)
+            .map(|r| r.name.clone())
+    }
+
+    fn object_name(&self, global_id: &str) -> Option<String> {
+        let gid = GlobalId::from_str(global_id).ok()?;
+        self.0
+            .try_get_entry_by_global_id(&gid)
+            .map(|e| e.name().item.clone())
```
mtabebe: Are we intentionally dropping the schema here. Will it be confusing if there are two objects that share a name across schemas?

mtabebe (commented):
Small clarification comments/questions.

I really appreciate the thorough tests here both unit and integration

src/compute-client/src/metrics.rs
SangJunBak: I think it might be a bit tricky since var_labels needs to be a slice since order matters. e.g. for
```
var_labels: ["instance_id", "replica_id", "collection_id", "quantile"],
```
we expect a slice when we record as well.

src/http-util/src/lib.rs
SangJunBak: Ah that's a good point, I think this scales with the number of metrics * number of rules. Each rule looks like
```
{"kind":"replica_name_lookup","cluster_id_label":"instance_id","replica_id_label":"replica_id","output_label":"replica_name"}
```
so it's about ~100 bytes per rule. Generally headers are limited to 8-16KB, so 160 rules spread across different metrics would reach this cap. I think it'd be a good idea to bound the size! I also think we can either:
1. gzip/encode the header before sending it since a lot of the information is repetitive
2. Add a dummy metric that represents the rules then make sure we don't include it in the final payload
3. Create our own proto format to include rules then include it in the body

I'm leaning towards 1 but what do you think?

src/environmentd/src/http/metrics_public.rs
SangJunBak: I think the database/schema would be useful actually! Good callout, I'll add it

src/compute-client/src/metrics.rs
mtabebe: Hmm good point. Can we enforce that if instance_id is there that matching rule is there? (For all the rules)

mtabebe: (And vice versa)

src/http-util/src/lib.rs
mtabebe: There is a 4th option ... a separate metrics/rules endpoint that is fetched on demand/on startup. But that is probably overly complex, but could be a longer term direction if we end up with a lot of rules.

I think the pragmatic approach is compression, since there is a lot of duplicate information. A comment explaining why would be helpful

src/http-util/src/lib.rs
```diff
@@ -54,6 +61,27 @@ fn wants_enrich_rules(headers: &HeaderMap) -> bool {
     headers.contains_key(MATERIALIZE_ACCEPT_ENRICH_RULES_HEADER)
 }
 
+/// Serializes `value` as JSON, gzips it, and base64-encodes the result for
+/// transport in [`MATERIALIZE_ENRICH_RULES_HEADER`].
+///
+/// The same handful of enrichment rules repeat across nearly every metric, so
```
mtabebe, line 67: 👍

src/ore/src/metrics/rule.rs
```diff
@@ -26,20 +26,32 @@ use std::collections::BTreeMap;
...
+
 /// Resolves IDs to their names
 pub trait NameLookup {
     /// Returns the name of the cluster with the given ID, if it exists.
     fn cluster_name(&self, cluster_id: &str) -> Option<String>;
     /// Returns the name of the replica with the given (cluster, replica) IDs.
     fn replica_name(&self, cluster_id: &str, replica_id: &str) -> Option<String>;
-    /// Returns the name of the catalog object with the given global ID.
-    fn object_name(&self, global_id: &str) -> Option<String>;
+    /// Returns the fully-qualified name of the catalog object with the given
```
mtabebe, line 46: 👍

mtabebe (approved):
Thanks for adressing!

PR #36635, opened 2026-05-20
Remove `enable_guard_subquery_tablefunc` feature flag

frankmcsherry:
The `enable_guard_subquery_tablefunc` feature flag was introduced when we brought the table func that stands in to guard subquery sizes. That seems to have worked out well, and the flag is no longer needed.


ggevay (approved):
Thank you!

frankmcsherry:
Thank YOU! :D

antiguru (approved):
Thank you!

Looks like we didn't have any tests for this, so no need for backwards-compat changes.

PR #36640, opened 2026-05-20
adapter: Prevent `enable_variadic_left_join` from firing on cross-joins

mgree:
### Motivation

Fixes https://linear.app/materializeinc/issue/SQL-286/enable-variadic-left-join-lowering-causes-wrong-results.

### Description

When trying to apply the variadic left join lowering (`attempt_left_join_magic`), we try to identify a column in the right-hand side of the `LEFT JOIN` that is equated to a `bound_input` in the left-hand side. The code correctly checked that the right-hand side was in the right relation, but only ensured that the left-hand side was uncorrelated. When presented with a cross join (as in the original bug report, `SELECT a.k, b.x, b.y, b.k, c.k FROM (a LEFT JOIN b ON b.x = b.y) LEFT JOIN c ON b.k = c.k`, we must plan `a LEFT JOIN b ON b.x = b.y` as a cross-join), we will attempt the lowering---when in fact we should not.

The fix is simple: make sure that the left column is actually in the left-hand side.

**UPDATE**: @def- found another, somewhat related bug. When there are self-equations `decompose_equations` will return them, and then later logic can generates a `left_vals` that doesn't distinguish between equations between left and right (useful) and just equations on the right (useless). I've renamed `decompose_equations` to `decompose_left_to_right_equations` and made it only return equations that cross its new `lhs_cutoff` parameter.

**UPDATE 2**: @def- found _another_ bug, this time induced by my putative fix. Conservatively, we should not run the VOJ lowering when there are any "non-crossing" equations. (It's not impossible, but the logic is tricky enough that I don't want to attempt it.) We can make a follow-up issue if that seems worthwhile, but there is a `TODO` in the code for now.

### Verification

Repros from the bug report and followup bug report went red to green. No need for other SLT rewrites (i.e., doesn't break existing examples).

### DIFF
--- a/src/sql/src/plan/lowering/variadic_left.rs
+++ b/src/sql/src/plan/lowering/variadic_left.rs
@@ -196,8 +196,9 @@ pub(crate) fn attempt_left_join_magic(
                 inc_metrics("voj_5");
                 return Ok(None);
             }
-            // Only columns not from the outer scope introduce bindings.
-            if left >= oa {
+            // Only columns not from the outer scope introduce bindings (`oa <= left`)
+            // And `left` needs to be a column in the left relation (`left < oa + ba`)
+            if oa <= left && left < oa + ba {
                 if let Some(bound) = bound_input {
                     // If left references come from different inputs, bail out.
                     if bound_to[left] != bound {
--- a/src/sql/src/plan/lowering/variadic_left.rs
+++ b/src/sql/src/plan/lowering/variadic_left.rs
@@ -176,7 +176,7 @@ pub(crate) fn attempt_left_join_magic(
         // then we can fish out values from that input. If it equates values
         // across multiple inputs, we would need to fish out valid tuples and
         // no idea how we would get those w/o doing a join or a cartesian product.
-        let equations = if let Some(list) = decompose_equations(&on) {
+        let equations = if let Some(list) = decompose_equations(&on, oa + ba) {
             list
         } else {
             tracing::debug!(case = 4, index, "attempt_left_join_magic");
@@ -395,7 +395,11 @@ use mz_expr::func::variadic::{And, Or};
 use mz_expr::{BinaryFunc, VariadicFunc};
 
 /// If `predicate` can be decomposed as any number of `col(x) = col(y)` expressions anded together, return them.
-fn decompose_equations(predicate: &MirScalarExpr) -> Option<Vec<(usize, usize)>> {
+/// Excludes equations where both columns are `>= lhs_cutoff`.
+fn decompose_equations(
+    predicate: &MirScalarExpr,
+    lhs_cutoff: usize,
+) -> Option<Vec<(usize, usize)>> {
     let mut equations = Vec::new();
 
     let mut todo = vec![predicate];
@@ -415,9 +419,9 @@ fn decompose_equations(predicate: &MirScalarExpr) -> Option<Vec<(usize, usize)>>
                 if let (MirScalarExpr::Column(c1, _name1), MirScalarExpr::Column(c2, _name2)) =
                     (&**expr1, &**expr2)
                 {
-                    if c1 < c2 {
+                    if c1 < c2 && *c1 < lhs_cutoff {
                         equations.push((*c1, *c2));
-                    } else {
+                    } else if *c2 < lhs_cutoff {
                         equations.push((*c2, *c1));
                     }
                 } else {
--- a/src/sql/src/plan/lowering/variadic_left.rs
+++ b/src/sql/src/plan/lowering/variadic_left.rs
@@ -176,7 +176,7 @@ pub(crate) fn attempt_left_join_magic(
         // then we can fish out values from that input. If it equates values
         // across multiple inputs, we would need to fish out valid tuples and
         // no idea how we would get those w/o doing a join or a cartesian product.
-        let equations = if let Some(list) = decompose_equations(&on, oa + ba) {
+        let equations = if let Some(list) = decompose_left_to_right_equations(&on, oa + ba) {
             list
         } else {
             tracing::debug!(case = 4, index, "attempt_left_join_magic");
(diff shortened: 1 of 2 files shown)


def- (changes requested):
I added an extra test that fails:
```
--- test/sqllogictest/variadic_outer_join.slt
2026-05-21T01:56:57.225736Z ERROR mz_transform: caught a panic during query optimization: type error in transient query:
In the MIR term:
Project (#0, #1)
  Union
    Get l1
    Constant
      - (null)


projection of non-existant columns [0, 1] from type (r_int32?)

With
  cte l0 =
    Constant
      - ()
Return
  Project (#0, #3, #1, #4)
    With
      cte l1 =
        CrossJoin
          Get l0
          Get u1
      cte l3 =
        Project (#0..=#3)
          Map (true)
            With
              cte l2 =
                CrossJoin
                  Get l0
                  Get u2
            Return
              Get l2
      cte l6 =
        Project (#0, #1)
          Map (true)
            With
              cte l5 =
                CrossJoin
                  Get l0
                  Get u3
            Return
              Get l5
      cte l4 =
        Union
          Filter (#1) IS NOT NULL AND (#2) IS NOT NULL
            Get l3
          Project (#2, #1, #0, #5)
            Map (null, null, null, null)
              Threshold
                Union
                  Negate
                    Project (#2, #1)
                      Filter (#1) IS NOT NULL AND (#2) IS NOT NULL
                        Get l3
                  Distinct project=[#0, #1]
                    Project (#0, #1)
                      Union
                        Get l1
                        Constant
                          - (null)
      cte l7 =
        Union
          Filter (#0) IS NOT NULL
            Get l6
          Project (#0, #2)
            Map (null, null)
              Threshold
                Union
                  Negate
                    Project (#0)
                      Filter (#0) IS NOT NULL
                        Get l6
                  Distinct project=[#0]
                    Project (#0)
                      Union
                        Get l1
                        Constant
                          - (null)
    Return
      Project (#0..=#3, #6)
        Map (case when (#5) IS NULL then null else #4 end)
          Filter ((#0 = #4) OR ((#0) IS NULL AND (#4) IS NULL))
            Project (#0..=#5)
              CrossJoin
                Project (#0, #5..=#7)
                  Map (case when (#4) IS NULL then null else #1 end, case when (#4) IS NULL then null else #2 end, case when (#4) IS NULL then null else #3 end)
                    Filter ((#0 = #3) OR ((#0) IS NULL AND (#3) IS NULL)) AND ((#1 = #2) OR ((#1) IS NULL AND (#2) IS NULL))
                      Project (#0..=#4)
                        CrossJoin
                          Get l1
                          Get l4
                Get l7

    SELECT a.k, b.k, b.x, c.k
    FROM (a LEFT JOIN b ON a.k = b.k AND b.x = b.y) LEFT JOIN c ON a.k = c.k
    PlanFailure:test/sqllogictest/variadic_outer_join.slt:197:
    db error: ERROR: internal error: internal error in optimizer: internal transform error: unexpected panic during query optimization: type error in transient query:
    In the MIR term:
    Project (#0, #1)
      Union
        Get l1
        Constant
          - (null)


    projection of non-existant columns [0, 1] from type (r_int32?)

    With
      cte l0 =
        Constant
          - ()
    Return
      Project (#0, #3, #1, #4)
        With
          cte l1 =
            CrossJoin
              Get l0
              Get u1
          cte l3 =
            Project (#0..=#3)
              Map (true)
                With
                  cte l2 =
                    CrossJoin
                      Get l0
                      Get u2
                Return
                  Get l2
          cte l6 =
            Project (#0, #1)
              Map (true)
                With
                  cte l5 =
                    CrossJoin
                      Get l0
                      Get u3
                Return
                  Get l5
          cte l4 =
            Union
              Filter (#1) IS NOT NULL AND (#2) IS NOT NULL
                Get l3
              Project (#2, #1, #0, #5)
                Map (null, null, null, null)
                  Threshold
                    Union
                      Negate
                        Project (#2, #1)
                          Filter (#1) IS NOT NULL AND (#2) IS NOT NULL
                            Get l3
                      Distinct project=[#0, #1]
                        Project (#0, #1)
                          Union
                            Get l1
                            Constant
                              - (null)
          cte l7 =
            Union
              Filter (#0) IS NOT NULL
                Get l6
              Project (#0, #2)
                Map (null, null)
                  Threshold
                    Union
                      Negate
                        Project (#0)
                          Filter (#0) IS NOT NULL
                            Get l6
                      Distinct project=[#0]
                        Project (#0)
                          Union
                            Get l1
                            Constant
                              - (null)
        Return
          Project (#0..=#3, #6)
            Map (case when (#5) IS NULL then null else #4 end)
              Filter ((#0 = #4) OR ((#0) IS NULL AND (#4) IS NULL))
                Project (#0..=#5)
                  CrossJoin
                    Project (#0, #5..=#7)
                      Map (case when (#4) IS NULL then null else #1 end, case when (#4) IS NULL then null else #2 end, case when (#4) IS NULL then null else #3 end)
                        Filter ((#0 = #3) OR ((#0) IS NULL AND (#3) IS NULL)) AND ((#1 = #2) OR ((#1) IS NULL AND (#2) IS NULL))
                          Project (#0..=#4)
                            CrossJoin
                              Get l1
                              Get l4
                    Get l7

    ----
    FAIL: plan-failure=1 success=12 total=13
```

def- (commented):
I added a new failing test:
```
--- test/sqllogictest/variadic_outer_join.slt
    SELECT a.k, b.x, b.y, b.k, c.k
    FROM (a LEFT JOIN b ON a.k = b.k AND b.x = b.y) LEFT JOIN c ON a.k = c.k
    OutputFailure:test/sqllogictest/variadic_outer_join.slt:222
            expected: Values(["1", "NULL", "NULL", "NULL", "NULL"])
            actually: Values(["1", "5", "6", "1", "NULL"])
            actual raw: [Row { columns: [Column { name: "k", table_oid: None, column_id: None, type: Int4 }, Column { name: "x", table_oid: None, column_id: None, type: Int4 }, Column { name: "y", table_oid: None, column_id: None, type: Int4 }, Column { name: "k", table_oid: None, column_id: None, type: Int4 }, Column { name: "k", table_oid: None, column_id: None, type: Int4 }] }]
    ----
    FAIL: output-failure=1 success=17 total=18
```
Claude's explanation in case it helps:

The rename decompose_equations → decompose_left_to_right_equations makes it keep only column equations that straddle the lhs_cutoff, and silently drop any equation whose two columns are on the same side — without bailing out.

So for a join like `... ON a.k = b.k AND b.x = b.y`:
- The cross equation a.k = b.k makes VOJ fire.
- The right-local equation b.x = b.y is dropped and never re-enforced anywhere (the row filter at variadic_left.rs:337 is built solely from the returned equations; the other filters only test RHS IS NOT NULL).

Result: rows that match the cross equation but violate the dropped one leak through — a wrong-results regression. Pre-PR, that same-side pair tripped the voj_5/voj_6 guards, VOJ declined, and generic lowering preserved b.x = b.y correctly.

def- (approved):

PR #36643, opened 2026-05-20
v26.24.3 release notes: clarify TIMESTAMP fix scope

bosconi:
Follow-up to #36641. The original v26.24.3 bug-fix entry read as though #36612 also handled zero-value timestamps (`0000-00-00 00:00:00`). It does not — those still require `TEXT COLUMNS` plus an explicit `CAST` in user queries. This PR adjusts the wording to make that clear.


maheshwarip (approved):

maheshwarip:
@martykulma would love your eyes on this before we merge

martykulma (approved):
lgtm!

PR #36644, opened 2026-05-20
compute: apply temporal bucketing in more situations

ggevay:
See commit messages.

### DIFF
--- a/src/compute-types/src/plan/lowering.rs
+++ b/src/compute-types/src/plan/lowering.rs
@@ -886,7 +886,8 @@ This is not expected to cause incorrect results, but could indicate a performanc
                         AvailableCollections::new_raw(),
                         &keys,
                         arity,
-                        input_future,
+                        // `new_raw` means no arrangement, so no bucketing is needed
+                        false,
                     )
                 } else {
                     input
@@ -919,7 +920,8 @@ This is not expected to cause incorrect results, but could indicate a performanc
                         AvailableCollections::new_raw(),
                         &keys,
                         arity,
-                        input_future,
+                        // `new_raw` means no arrangement, so no bucketing is needed
+                        false,
                     )
                 } else {
                     input
@@ -986,7 +988,7 @@ This is not expected to cause incorrect results, but could indicate a performanc
                         |LoweredExpr {
                              plan,
                              keys,
-                             has_future_updates: future,
+                             has_future_updates: _,
                          }| {
                             // We don't have an MFP here -- install an operator to permute the
                             // input, if necessary.
@@ -996,7 +998,8 @@ This is not expected to cause incorrect results, but could indicate a performanc
                                     AvailableCollections::new_raw(),
                                     &keys,
                                     arity,
-                                    future,
+                                    // `new_raw` means no arrangement, so no bucketing is needed
+                                    false,
                                 )
                             } else {
                                 plan
--- a/src/compute-types/src/plan/lowering.rs
+++ b/src/compute-types/src/plan/lowering.rs
@@ -890,17 +890,7 @@ This is not expected to cause incorrect results, but could indicate a performanc
                     input
                 };
                 // Return the plan, and no arrangements.
-                //
-                // `TopK` itself buckets (via `apply_bucketing_strategy` at the
-                // top of `render_topk`) when the strategy says so, so the
-                // output flag is cleared whenever bucketing actually fires.
-                //
-                // Bucketing absorption: see `strategy_from_future`.
                 let temporal_bucketing_strategy = strategy_from_future(input_future);
-                let has_future_updates = match temporal_bucketing_strategy {
-                    ArrangementStrategy::TemporalBucketing => false,
-                    ArrangementStrategy::Direct => input_future,
-                };
                 let lir_id = self.allocate_lir_id();
                 LoweredExpr {
                     plan: PlanNode::TopK {
@@ -910,7 +900,7 @@ This is not expected to cause incorrect results, but could indicate a performanc
                     }
                     .as_plan(lir_id),
                     keys: AvailableCollections::new_raw(),
-                    has_future_updates,
+                    has_future_updates: false,
                 }
             }
             MirRelationExpr::Negate { input } => {
--- a/src/compute-types/src/plan/lowering.rs
+++ b/src/compute-types/src/plan/lowering.rs
@@ -67,10 +67,6 @@ pub(super) struct Context {
     debug_info: LirDebugInfo,
     /// Whether to enable fusion of MFPs in reductions.
     enable_reduce_mfp_fusion: bool,
-    /// Whether `Union`s with at least one `Negate` input should set
-    /// `consolidate_output = true`. Folded in from the deleted
-    /// `refine_union_negate_consolidation` LIR pass.
-    enable_consolidate_after_union_negate: bool,
 }
 
 impl Context {
@@ -84,7 +80,6 @@ impl Context {
                 id: GlobalId::Transient(0),
             },
             enable_reduce_mfp_fusion: features.enable_reduce_mfp_fusion,
-            enable_consolidate_after_union_negate: features.enable_consolidate_after_union_negate,
         }
     }
 
@@ -1017,15 +1012,12 @@ This is not expected to cause incorrect results, but could indicate a performanc
                     lowered_inputs.push(self.lower_mir_expr(input)?);
                 }
 
-                // Fold the deleted `refine_union_negate_consolidation` pass in
-                // here: a Union with any `Negate` input should consolidate its
+                // A Union with any `Negate` input should consolidate its
                 // output. The lowering is the only place where this decision
                 // can be coupled with the per-input bucketing strategy.
-                let has_negate_input = lowered_inputs
+                let consolidate_output = lowered_inputs
                     .iter()
                     .any(|l| matches!(l.plan.node, PlanNode::Negate { .. }));
-                let consolidate_output =
-                    self.enable_consolidate_after_union_negate && has_negate_input;
 
                 // Per-input bucketing strategies: only meaningful when the
                 // Union consolidates its output, since bucketing only pays off
--- a/src/compute-types/src/plan/lowering.rs
+++ b/src/compute-types/src/plan/lowering.rs
@@ -32,10 +32,8 @@ use crate::plan::{ArrangementStrategy, AvailableCollections, GetPlan, LirId, Pla
 /// Pick an [`ArrangementStrategy`] based on whether the input may contain future-stamped
 /// updates. Future updates are the only case where temporal bucketing pays off.
 ///
-/// Convention: every caller that returns `TemporalBucketing` must also clear
-/// `LoweredExpr::has_future_updates` on the resulting `LoweredExpr`, so that a stack of
-/// bucketing-eligible operators only buckets at the lowest one. A trailing temporal MFP
-/// fused on top naturally re-arms the flag.
+/// Any arrangement or consolidation that absorbs data that can have future updates should be
+/// guarded by a temporal bucketing operator.
 fn strategy_from_future(has_future_updates: bool) -> ArrangementStrategy {
     if has_future_updates {
         ArrangementStrategy::TemporalBucketing
--- a/src/compute-types/src/plan/lowering.rs
+++ b/src/compute-types/src/plan/lowering.rs
@@ -300,11 +300,8 @@ impl Context {
                 // from the underlying binding — applying an MFP doesn't drop future-
                 // timestamped updates that already exist on the input.
                 //
-                // TODO(temporal-bucketing): `has_future_updates` is computed per
-                // dataflow; we don't currently propagate it across `Id::Global`
-                // boundaries (e.g., from an MV's dataflow to its consumer's), so a
-                // downstream-only `Get`-then-`ArrangeBy` won't bucket unless the
-                // consumer has its own local temporal MFP.
+                // Note that global Gets from different dataflows can't have future updates, because
+                // both indexes and materialized views hold back future updates.
                 let has_future_updates = self.has_future_updates.contains(id)
                     || match &plan {
                         GetPlan::Arrangement(_, _, mfp) | GetPlan::Collection(mfp) => {
--- a/src/compute-types/src/plan/lowering.rs
+++ b/src/compute-types/src/plan/lowering.rs
@@ -954,38 +954,22 @@ This is not expected to cause incorrect results, but could indicate a performanc
                 } = self.lower_mir_expr(input)?;
                 let arity = input.arity();
                 let (threshold_plan, required_arrangement) = ThresholdPlan::create_from(arity);
-                // If the conditional `arrange_by` wrap fires, the synthesized
-                // `ArrangeBy` builds a real arrangement (`new_arranged`) and so
-                // its `apply_bucketing_strategy` call at render time will
-                // actually bucket. If the wrap is skipped, the input already
-                // had a suitable arrangement and an upstream operator was
-                // responsible for bucketing.
-                let wrap_fires = !keys
+
+                let (plan, has_future_updates) = if !keys
                     .arranged
                     .iter()
-                    .any(|(key, _, _)| key == &required_arrangement.0);
-                let plan = if wrap_fires {
-                    self.arrange_by(
+                    .any(|(key, _, _)| key == &required_arrangement.0)
+                {
+                    let plan = self.arrange_by(
                         plan,
                         AvailableCollections::new_arranged(vec![required_arrangement]),
                         &keys,
                         arity,
                         input_future,
-                    )
+                    );
+                    (plan, false)
                 } else {
-                    plan
-                };
-
-                // Clear the future flag whenever the wrap actually bucketed
-                // the future-stamped updates.
-                let has_future_updates = if wrap_fires
-                    && matches!(
-                        strategy_from_future(input_future),
-                        ArrangementStrategy::TemporalBucketing
-                    ) {
-                    false
-                } else {
-                    input_future
+                    (plan, input_future)
                 };
 
                 let output_keys = threshold_plan.keys();
--- a/src/compute-types/src/plan/lowering.rs
+++ b/src/compute-types/src/plan/lowering.rs
@@ -31,6 +31,11 @@ use crate::plan::{ArrangementStrategy, AvailableCollections, GetPlan, LirId, Pla
 
 /// Pick an [`ArrangementStrategy`] based on whether the input may contain future-stamped
 /// updates. Future updates are the only case where temporal bucketing pays off.
+///
+/// Convention: every caller that returns `TemporalBucketing` must also clear
+/// `LoweredExpr::has_future_updates` on the resulting `LoweredExpr`, so that a stack of
+/// bucketing-eligible operators only buckets at the lowest one. A trailing temporal MFP
+/// fused on top naturally re-arms the flag.
 fn strategy_from_future(has_future_updates: bool) -> ArrangementStrategy {
     if has_future_updates {
         ArrangementStrategy::TemporalBucketing
@@ -294,6 +299,12 @@ impl Context {
                 // Even with a non-temporal MFP, we must propagate `has_future_updates`
                 // from the underlying binding — applying an MFP doesn't drop future-
                 // timestamped updates that already exist on the input.
+                //
+                // TODO(temporal-bucketing): `has_future_updates` is computed per
+                // dataflow; we don't currently propagate it across `Id::Global`
+                // boundaries (e.g., from an MV's dataflow to its consumer's), so a
+                // downstream-only `Get`-then-`ArrangeBy` won't bucket unless the
+                // consumer has its own local temporal MFP.
                 let has_future_updates = self.has_future_updates.contains(id)
                     || match &plan {
                         GetPlan::Arrangement(_, _, mfp) | GetPlan::Collection(mfp) => {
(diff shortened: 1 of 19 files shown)


ggevay:
https://github.com/MaterializeInc/materialize/pull/36648 is meant to supersede this PR.

Edit: Actually, I've now updated this PR in case we want to move forward with this one.

src/compute-types/src/plan/lowering.rs
```diff
@@ -294,6 +304,12 @@ impl Context {
                 // Even with a non-temporal MFP, we must propagate `has_future_updates`
                 // from the underlying binding — applying an MFP doesn't drop future-
                 // timestamped updates that already exist on the input.
+                //
+                // TODO(temporal-bucketing): `has_future_updates` is computed per
+                // dataflow; we don't currently propagate it across `Id::Global`
+                // boundaries (e.g., from an MV's dataflow to its consumer's), so a
+                // downstream-only `Get`-then-`ArrangeBy` won't bucket unless the
+                // consumer has its own local temporal MFP.
```
antiguru: Correct the comment please. MVs never produce future updates, indexes don't either.

src/compute-types/src/plan/lowering.rs
```diff
@@ -852,21 +868,34 @@ This is not expected to cause incorrect results, but could indicate a performanc
...
+                // `TopK` itself buckets (via `apply_bucketing_strategy` at the
+                // top of `render_topk`) when the strategy says so, so the
+                // output flag is cleared whenever bucketing actually fires.
+                //
+                // Bucketing absorption: see `strategy_from_future`.
+                let temporal_bucketing_strategy = strategy_from_future(input_future);
+                let has_future_updates = match temporal_bucketing_strategy {
+                    ArrangementStrategy::TemporalBucketing => false,
+                    ArrangementStrategy::Direct => input_future,
+                };
```
antiguru: I'm not sure I follow this part: The output of TopK should only be marked as `has_future_updates` if it has an MFP at the end with a temporal filter, but not based on its inputs?

src/compute-types/src/plan/lowering.rs
```diff
@@ -909,11 +939,17 @@ This is not expected to cause incorrect results, but could indicate a performanc
...
                 let arity = input.arity();
                 let (threshold_plan, required_arrangement) = ThresholdPlan::create_from(arity);
-                let plan = if !keys
+                // If the conditional `arrange_by` wrap fires, the synthesized
+                // `ArrangeBy` builds a real arrangement (`new_arranged`) and so
+                // its `apply_bucketing_strategy` call at render time will
+                // actually bucket. If the wrap is skipped, the input already
+                // had a suitable arrangement and an upstream operator was
+                // responsible for bucketing.
+                let wrap_fires = !keys
```
antiguru: Can we rename the variable? `wrap_fires` only makes sense after reading the comment.

src/compute-types/src/plan/lowering.rs
```diff
@@ -31,6 +31,11 @@ use crate::plan::{ArrangementStrategy, AvailableCollections, GetPlan, LirId, Pla
 
 /// Pick an [`ArrangementStrategy`] based on whether the input may contain future-stamped
 /// updates. Future updates are the only case where temporal bucketing pays off.
+///
+/// Convention: every caller that returns `TemporalBucketing` must also clear
+/// `LoweredExpr::has_future_updates` on the resulting `LoweredExpr`, so that a stack of
+/// bucketing-eligible operators only buckets at the lowest one. A trailing temporal MFP
+/// fused on top naturally re-arms the flag.
```
antiguru: I don't think this is true in general? Can make more specific, which is that any arrangement that absorbs data that can have future updates should be guarded by a temporal bucketing operator. The output of an operator can have future updates if it applies a non-safe MFP.

src/compute-types/src/plan/lowering.rs
```diff
@@ -945,14 +993,56 @@ This is not expected to cause incorrect results, but could indicate a performanc
...
+
+                // Fold the deleted `refine_union_negate_consolidation` pass in
+                // here: a Union with any `Negate` input should consolidate its
+                // output. The lowering is the only place where this decision
+                // can be coupled with the per-input bucketing strategy.
+                let has_negate_input = lowered_inputs
+                    .iter()
+                    .any(|l| matches!(l.plan.node, PlanNode::Negate { .. }));
+                let consolidate_output =
+                    self.enable_consolidate_after_union_negate && has_negate_input;
```
antiguru: Can we not fold enable_consolidate_after_union_negate as part of this PR?

src/compute/src/render.rs
```diff
@@ -1481,11 +1504,18 @@ pub trait RenderTimestamp: MzTimestamp + Refines<mz_repr::Timestamp> {
...
         summary: mz_repr::Timestamp,
-    ) -> VecCollection<'scope, Self, Row, Diff>;
+    ) -> VecCollection<'scope, Self, D, Diff>
+    where
+        D: timely::ExchangeData
+            + crate::typedefs::MzData
+            + Ord
+            + Clone
+            + std::fmt::Debug
+            + differential_dataflow::Hashable;
```
antiguru: Can be simplified by using `differential_dataflow::Data` here.

antiguru (commented):
I read most of it, but it's very dense. Overall some nits that comments aren't accurate, and the change mixes some changes (inlining refine_union_negate, explain changes). It seems to do the right thing.

The big issue I see with this change is that it introduces complexity, which we shouldn't have in the first place. It's not just caused by this PR, as my change to make temporal bucketing a property of LIR laid the groundwork for this. With this change, however, it becomes very apparent that this is the wrong approach and we need an alternative.

So, a way forward could be to merge this change and then commit on spending time to implement an alternative that _localizes_ complexity instead of spreading it all around. The idea would be to opt-in to temporal bucketing cluster-wide as a setting one needs to specify, or make it so good it doesn't increase cost. Let's discuss this approach, and how to move forward with this change.

src/compute-types/src/plan/lowering.rs
ggevay: By "fold enable_consolidate_after_union_negate", you mean delete the flag and just make the optimization unconditionally on? We could do that! https://github.com/MaterializeInc/database-issues/issues/8780 also says that we should do it.

I'll add a commit to this PR.

src/compute-types/src/plan/lowering.rs
ggevay: Sorry, fixed!

src/compute-types/src/plan/lowering.rs
ggevay: Ah yes, I've simplified it to `false`. (`TopK` doesn't have have an MFP fused at the end, unlike `Reduce`.)

src/compute-types/src/plan/lowering.rs
ggevay: Simplified `Threshold`:
```
                let (plan, has_future_updates) = if !keys
                    .arranged
                    .iter()
                    .any(|(key, _, _)| key == &required_arrangement.0)
                {
                    let plan = self.arrange_by(
                        plan,
                        AvailableCollections::new_arranged(vec![required_arrangement]),
                        &keys,
                        arity,
                        input_future,
                    );
                    (plan, false)
                } else {
                    (plan, input_future)
                };
```

src/compute-types/src/plan/lowering.rs
ggevay: Yes, simplified the comment. Since this function is about just applying the in-flight flag, I removed that part from the doc comment that talked about propagating the flag downstream.

src/compute-types/src/plan/lowering.rs
ggevay: Done

src/compute/src/render.rs
ggevay: Thanks, done! It ended up as:
```
        D: differential_dataflow::ExchangeData
            + crate::typedefs::MzData
            + differential_dataflow::Hashable;
```

ggevay:
Thank you for the comments! I've addressed all of them.

> The idea would be to opt-in to temporal bucketing cluster-wide as a setting one needs to specify, or make it so good it doesn't increase cost. Let's discuss this approach, and how to move forward with this change.

Yeah, if we could make it so good that it doesn't increase cost, that would be amazing!

ggevay:
Nightly: https://buildkite.com/materialize/nightly/builds/16535

Edit: There are unrelated test failures, plus a test issue due to the flag removal, discussed [on slack](https://materializeinc.slack.com/archives/C01LKF361MZ/p1779628090158369).

Edit 2: Nightly sqlancer after fixing the sqlancer problem: https://buildkite.com/materialize/nightly/builds/16536

src/compute-types/src/plan/lowering.rs
```diff
@@ -852,21 +858,24 @@ This is not expected to cause incorrect results, but could indicate a performanc
                         AvailableCollections::new_raw(),
                         &keys,
                         arity,
-                        input_future,
+                        // `new_raw` means no arrangement, so no bucketing is needed
+                        false,
```
antiguru, line 885: Is this right? Raw means there's no arranged form to read from, only the unarranged form. It can have future updates if `input` has any.

src/compute-types/src/plan/lowering.rs
```diff
@@ -909,20 +919,22 @@ This is not expected to cause incorrect results, but could indicate a performanc
...
                 let (threshold_plan, required_arrangement) = ThresholdPlan::create_from(arity);
-                let plan = if !keys
+
+                let (plan, has_future_updates) = if !keys
                     .arranged
                     .iter()
                     .any(|(key, _, _)| key == &required_arrangement.0)
                 {
-                    self.arrange_by(
+                    let plan = self.arrange_by(
```
antiguru: We need to be careful in these situations because the collection in the input's `CollectionBundle` can still have future updates, and this is what the boolean is supposed to track.

src/compute-types/src/plan/lowering.rs
```diff
@@ -945,14 +957,53 @@ This is not expected to cause incorrect results, but could indicate a performanc
...
+                    lowered_inputs
+                        .iter()
+                        .map(|_| ArrangementStrategy::Direct)
+                        .collect()
+                };
+
+                // If the Union itself buckets each input (consolidating Union
+                // with `TemporalBucketing` strategies), the future-stamped
+                // updates are absorbed here and we should clear the outer
+                // flag for those legs. Otherwise, propagate `any_future`.
```
antiguru: I'm not sure I follow the reasoning behind this comment. The output of union has future updates if `!consolidate_output`. What is this logic trying to achieve?

src/compute-types/src/plan/lowering.rs
```diff
@@ -945,14 +957,53 @@ This is not expected to cause incorrect results, but could indicate a performanc
...
+                    lowered_inputs
+                        .iter()
+                        .map(|l| strategy_from_future(l.has_future_updates))
+                        .collect()
+                } else {
+                    lowered_inputs
+                        .iter()
+                        .map(|_| ArrangementStrategy::Direct)
+                        .collect()
+                };
```
antiguru, line 1007: I feel an empty vector is equivalent and saves memory? The contract with downstream would be that `matches!(strategies.get(idx), Some(ArrangementStrategy::TemporalBucketing)`.

src/compute/src/render/context.rs
```diff
@@ -1353,3 +1360,44 @@ fn walk_cursor<C, F>(
...
+/// `enable_compute_temporal_bucketing` dyncfg is on; otherwise return `oks` unchanged.
+///
+/// Returns `(stream, applied)` where `applied` is true iff bucketing actually fired. Callers
+/// use this flag to avoid double-bucketing the same stream within a single `ensure_collections`
+/// invocation.
+///
+/// Generic over the row data type `D` so the helper can serve both `Row` streams (the
+/// `ArrangeBy` rendering in `ensure_collections`) and `(Row, Row)` streams (the internal
+/// keyed input arrangement built by `render_reduce`).
+pub(crate) fn apply_bucketing_strategy<'scope, T, D>(
```
antiguru: Slight preference to inlining this function, as it's essentially an if around a relatively short block of code.

src/compute/src/render/context.rs
```diff
@@ -1004,7 +1004,21 @@ impl<'scope, T: RenderTimestamp> CollectionBundle<'scope, T> {
...
+        // a bucketing consumer clears `LoweredExpr::has_future_updates`, so its
+        // parent is lowered with `ArrangementStrategy::Direct`.
+        //
+        // This flag is a belt-and-suspenders check for the case where a single
+        // `ensure_collections` call would otherwise fire bucketing twice on the
+        // same collection: once when forming the raw collection (via
+        // `as_collection_core` below) and again when building an arrangement in
+        // `collections.arranged`, or across two arrangements in that loop. The
+        // same `strategy` argument applies to all sites; the flag downgrades
+        // every application after the first to `Direct`.
```
antiguru: I'm not sure this is universally true: If we have to arrange-by operation on the same collection, they'll each bucket independently because they're distinct nodes. The comment seems to imply that in this case we'd only bucket once, but in fact we can only make decisions within the local scope.

antiguru (commented):
Look through it again, thanks for addressing my previous comments.

src/compute-types/src/plan/lowering.rs
ggevay: Do you mean that the `false` is incorrect here, or just that the comment should be adjusted?

I might be missing something, but the `false` looks correct to me. My understanding is as follows:
- The `ArrangeBy` node created by an `arrange_by(...new_raw...)` call doesn't need temporal bucketing because there is no `MergeBatcher`. So, even though we might have future updates here, they won't cause trouble.
- If any downstream operator has a `MergeBatcher`, then the temporal bucketing insertion logic for _that operator_ would insert temporal bucketing there. (Specifically, a few lines below we are creating a `TopK` with `temporal_bucketing_strategy` set to `input_future`, where `input_future` is from before the `ArrangeBy`.)

src/compute-types/src/plan/lowering.rs
ggevay: Ah yes, this is very subtle. You mean that even though we are making an arrangement, which would hold back future updates, somebody downstream might read not the arrangement, but the raw form? Actually, I think there is no issue, because even the raw form would be bucketed in this case. This is because in `ensure_collections` (which is the rendering of `ArrangeBy`) the `Self::arrange_collection` call puts the bucketed thing that it got (from the `apply_bucketing_strategy` call just above it) into `passthrough`. Does this address your concern?

src/compute-types/src/plan/lowering.rs
ggevay: Indeed, simplified the `consolidate_output` `true` branch to simply `false`.

src/compute-types/src/plan/lowering.rs
ggevay: Can do, but I'm not sure it's worth complicating the contract for a minuscule memory gain during lowering. Currently, the contract with downstream is that `temporal_bucketing_strategies` has exactly the same length as `inputs` (see the `zip_eq` in `render_plan_expr`), and the proposal would be adding a layer of "strategy is may or may not be there".

src/compute/src/render/context.rs
ggevay: Ok, inlined.

src/compute/src/render/context.rs
ggevay: Yes, fixed the comment by dropping the 2nd and 3rd paragraphs.

ggevay:
Thanks, addressed the new comments! (Plus did one more simplification: Making `Threshold`'s downstream `has_future_updates` false.)

src/compute-types/src/plan/lowering.rs
antiguru: Correct, but this is local to arrange-by. Other operators don't have the passthrough optimization (they can't; I think only arrange-by enriches the collection bundle, all others consume.)

src/compute-types/src/plan/lowering.rs
antiguru: If I'm reading the code correctly, it installs an arrange-by on `keys` if the requested keys aren't `raw`. I'd assume that the TopK then uses this arrangement, and we should pass input_future instead of `false` here? Also, if we create the arrangement, we don't have to bucket in TopK again, because the arrange-by holds back future updates.

ggevay: I think this call site of `arrange_by` doesn't create an arrangement (very misleadingly, but that's a pre-existing issue). It just creates a raw stream (if there is no raw stream already). And then `TopK` builds its own arrangement (for which we decide whether to do temporal bucketing from `input_future`, which comes from before the `arrange_by`).

antiguru (approved):
Thank you!

src/compute/src/render/top_k.rs
```diff
@@ -50,14 +52,64 @@ use crate::row_spine::{
...
+        if matches!(
+            temporal_bucketing_strategy,
+            ArrangementStrategy::TemporalBucketing
+        ) {
+            let must_consolidate = match &top_k_plan {
+                TopKPlan::MonotonicTop1(p) => p.must_consolidate,
+                TopKPlan::MonotonicTopK(p) => p.must_consolidate,
+                TopKPlan::Basic(_) => true,
+            };
+            soft_assert_or_log!(
```
def-, line 92: I think this assertion is wrong. With this diff:
```diff
diff --git a/src/transform/src/analysis/monotonic.rs b/src/transform/src/analysis/monotonic.rs
index 066da6a36f..47fdfb9b03 100644
--- a/src/transform/src/analysis/monotonic.rs
+++ b/src/transform/src/analysis/monotonic.rs
@@ -88,7 +88,12 @@ impl Analysis for Monotonic {
                 // can result in the future removal of records.
                 // TODO: this could be improved to only restrict if upper bounds
                 // are present, as temporal lower bounds only delay introduction.
-                is_monotonic && !predicates.iter().any(|p| p.contains_temporal())
+                //
+                // MINIMAL CHANGE TO DEMONSTRATE THE render_topk ASSERTION IS REACHABLE:
+                // realize the TODO above (over-approximated: treats *all* temporal
+                // predicates as monotonicity-preserving, not just lower bounds).
+                let _ = predicates;
+                is_monotonic
             }
             MirRelationExpr::Reduce { aggregates, .. } => {
                 // Reduce is monotonic iff its input is, and it is a "distinct"
```
and a simple SLT:
```
# Trigger: render the forbidden TopK plan so the render_topk
# soft_assert_or_log! fires in clusterd.
mode cockroach

statement ok
CREATE SOURCE counter_src FROM LOAD GENERATOR COUNTER;

statement ok
CREATE TABLE counter FROM SOURCE counter_src;

# Monotonic source + temporal lower-bound filter + LIMIT 1 ->
# TopK::MonotonicTop1 { must_consolidate: false } with TemporalBucketing.
statement ok
CREATE MATERIALIZED VIEW mv_topk_mono AS
SELECT counter FROM counter
WHERE counter <= mz_now()
ORDER BY counter
LIMIT 1;

# Read from the MV to force the dataflow to hydrate on the cluster.
query I
SELECT count(*) FROM mv_topk_mono;
----
1
```
I'm seeing it fire:
```
--- test/sqllogictest/zz_trigger_topk_assert.slt
2026-05-29T02:32:25.636470Z  thread 'timely:work-1' panicked at src/compute/src/render/top_k.rs:92:13:
TopK with `TemporalBucketing` should not have `must_consolidate = false`; `RelaxMustConsolidate` only runs on single-time dataflows where `mz_now()` has been const-folded and no temporal bucketing is set
   7: core::panicking::panic_fmt
   8: <mz_compute::render::context::Context<mz_repr::timestamp::Timestamp>>::render_plan_expr
   9: <mz_compute::render::context::Context<mz_repr::timestamp::Timestamp>>::render_letfree_plan
  10: <timely::dataflow::scope::Scope<mz_repr::timestamp::Timestamp>>::region_named::<mz_compute::render::context::CollectionBundle<mz_repr::timestamp::Timestamp>, <mz_compute::render::context::Context<mz_repr::timestamp::Timestamp>>::render_plan::{closure#1}>
  11: <mz_compute::render::context::Context<mz_repr::timestamp::Timestamp>>::render_plan
  12: <timely::dataflow::scope::Scope<mz_repr::timestamp::Timestamp>>::region_named::<(), mz_compute::render::build_compute_dataflow::{closure#3}::{closure#2}>
  13: <timely::worker::Worker>::dataflow_core::<mz_repr::timestamp::Timestamp, (), mz_compute::render::build_compute_dataflow::{closure#3}, alloc::boxed::Box<()>>
  14: mz_compute::render::build_compute_dataflow
  15: <mz_compute::compute_state::ActiveComputeState>::handle_compute_command
  16: <mz_compute::server::Worker>::handle_command
  17: <mz_compute::server::Config as mz_cluster::client::ClusterSpec>::run_worker
```
Based on https://github.com/MaterializeInc/qa-llm-review/blob/master/commit-bugs/done/analysis-commit-92cb1cf559f6be1e6b13b1d9864baab560b7a9af.md

ggevay:
Thanks @def-, indeed this is a potential problem. It can't currently be hit, but could be in the future if the TODO in the `Monotonic` Analysis is resolved.

The fix would be to return an additional boolean from `TopKPlan::create_from` to indicate the monotonic case, and set the `temporal_bucketing_strategy` on the `TopK` node in the lowering accordingly.

However, I'd like to not fix this now, because this code will soon be going away hopefully, as part of simplifying the bucketing. (But if, for whatever reason, we decide that we don't do the simplification very soon, then I'll make sure to fix this.)

PR #36649, opened 2026-05-20
MySQL: Fix zero-value YEAR decoding, improve user docs around zero-value date/time types

martykulma:
This started doc change and new test, but then @kay-kim uncovered a bug via manual testing!  No we have a fix, which is included.

This PR fixes handling of zero-value YEAR
- snapshot decodes this as "0" instead of "0000"
- binlog decodes this as "1900" instead of "0000"

This also works around a mysql_common bug that treat zero-year binlog as 1900, which is an invalid year. It should be 0. See [MySQL crazy set of rules for YEAR](https://dev.mysql.com/doc/refman/8.0/en/year.html).

Additionally:
- adds some clarification to Materialize documentation around the integration with MySQL when using zero-value DATE, DATETIME, TIMESTAMP, and YEAR.
- expands the test coverage for YEAR to verify the documented behavior
- adjusts tests for DATE/DATETIME/TIMESTAMP to ensure snapshot and steady state paths are tested

### DIFF
--- a/doc/user/data/mysql_source_details.yml
+++ b/doc/user/data/mysql_source_details.yml
@@ -79,11 +79,8 @@
     To ingest columns that contain zero values, use [`TEXT
     COLUMNS`](/sql/create-source/mysql/#handling-unsupported-types) to
     decode the affected columns as `text`. The zero values for `date`,
-    `datetime`, and `timestamp` are preserved verbatim as strings (e.g.
-    `"0000-00-00 00:00:00"`).
-
-    NOTE: The zero value for `year` surfaces as the string `"0"` rather than
-    `"0000"`.
+    `datetime`, `timestamp`, and `year` are preserved verbatim as strings
+    (e.g. `"0000-00-00 00:00:00"`, `"0000"`).
 
 - name: mysql-truncation-restriction
   content: |
--- a/doc/user/data/mysql_source_details.yml
+++ b/doc/user/data/mysql_source_details.yml
@@ -68,6 +68,23 @@
     - Use the [`EXCLUDE COLUMNS`](/sql/create-source/mysql/#excluding-columns)
     option to exclude any columns that contain unsupported data types.
 
+    #### Zero values for `date`, `datetime`, and `timestamp`
+
+    MySQL allows the special "zero" values `0000-00-00`, `0000-00-00
+    00:00:00` in `date`, `datetime`, and `timestamp` columns when the server
+    `sql_mode` does not include `NO_ZERO_DATE` or `NO_ZERO_IN_DATE`. These
+    values are not representable in Materialize's corresponding native types,
+    so they will cause ingestion to fail for the affected column.
+
+    To ingest columns that contain zero values, use [`TEXT
+    COLUMNS`](/sql/create-source/mysql/#handling-unsupported-types) to
+    decode the affected columns as `text`. The zero values for `date`,
+    `datetime`, and `timestamp` are preserved verbatim as strings (e.g.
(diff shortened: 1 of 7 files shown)


doc/user/data/mysql_source_details.yml
```diff
@@ -67,6 +67,23 @@
...
+    so they will cause ingestion to fail for the affected column.
+
+    To ingest columns that contain zero values, use [`TEXT
+    COLUMNS`](/sql/create-source/mysql/#handling-unsupported-types) to
+    decode the affected columns as `text`. The zero values for `date`,
+    `datetime`, and `timestamp` are preserved verbatim as strings (e.g.
+    `"0000-00-00 00:00:00"`).
+
+    NOTE: The zero value for `year` surfaces as the string `"0"` rather than
+    `"0000"`.
```
kay-kim:
```suggestion
    {{< note >}}
    The zero value for `year` surfaces as the string `"0"` rather than `"0000"`.
    {{< /note >}}
```

test/mysql-cdc/text-columns-year.td
```diff
@@ -0,0 +1,111 @@
...
+COMMIT;
+
+> SELECT id, event_year FROM events ORDER BY id;
+1 "2024"
+2 "0"
+3 <null>
+4 "1901"
+5 "2155"
+6 "2025"
+7 "0"
```
kay-kim: Should this be 1900?

Image

kay-kim (approved):

test/mysql-cdc/text-columns-year.td
martykulma: Thanks for running it! It should definitely be 0, and the reason the automated test didn't catch it is because it ran so fast, everything was read via the snapshot.  I modified the test and now see the failure.

```
text-columns-year.td:76:1: error: non-matching rows: expected:
[["1", "2024"], ["10", "2155"], ["2", "0"], ["3", "<null>"], ["4", "1901"], ["5", "2155"], ["6", "2025"], ["7", "0"], ["8", "<null>"], ["9", "1901"]]
got:
[["1", "2024"], ["10", "2155"], ["2", "0"], ["3", "<null>"], ["4", "1901"], ["5", "2155"], ["6", "2025"], ["7", "1900"], ["8", "<null>"], ["9", "1901"]]
Poor diff:
- 7 0
+ 7 1900

```

test/mysql-cdc/text-columns-year.td
```diff
@@ -0,0 +1,120 @@
...
+3 <null>
+4 "1901"
+5 "2155"
+
+# Post-snapshot rows exercise the replication / binlog decode path.
+$ mysql-execute name=mysql
+USE public;
+SET SESSION sql_mode = '';
+START TRANSACTION;
+INSERT INTO events VALUES (6, '2025'), (7, 0), (8, NULL);
```
ublubu, line 74: I'm curious, why does the binlog side of the test use `0` in the MySQL statement
while the snapshot side uses `'0000'`?

Is that coincidental?
Just checking that I correctly read https://dev.mysql.com/doc/refman/8.0/en/year.html
and `0` and `'0000'` are interchangeable YEARs in MySQL SQL.

ublubu (approved):
Looks legit. One clarification question inline.

test/mysql-cdc/text-columns-year.td
martykulma: yes. Only 1 or 2 digit strings '0', '00' are interpreted as 2000.  

numeric 0 or '0000' => 0000

PR #36650, opened 2026-05-21
storage: Don't swallow header parse errors

def-:
on empty chunks

### DIFF
--- a/src/storage/src/decode/csv.rs
+++ b/src/storage/src/decode/csv.rs
@@ -124,37 +124,46 @@ impl CsvDecoderState {
                         }
                     };
 
-                    // skip header rows, do not send them into dataflow
-                    if self.next_row_is_header {
-                        self.next_row_is_header = false;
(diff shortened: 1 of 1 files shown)


def-:
Ready for review

martykulma (commented):
Wow.. this thing is hard to read because we keep inspecting result all over the place. 

This is a good opportunity to refactor and clean up where we identify and break out of the loop - e.g.
```
if self.next_row_is_header {
    self.next_row_is_header = false;
    match result {
        Err(e) => break Err(e),
        Ok(Some(row)) => {
            // .. validate header ..
            if chunk.is_empty() { break Ok(None); }
        }
        Ok(None) => unreachable!(), // .. returned early, ln 87
    }
} else {
    break result;
}
```

def-:
Done, ready for review again

martykulma (approved):
thanks @def- !

PR #36653, opened 2026-05-21
expr: lossy text casts don't preserve uniqueness

def-:
text_to_name truncates to 63 bytes (parse_pg_legacy_name) and text_to_varchar(n) with fail_on_len=false truncates to n chars, so both can collapse distinct inputs to the same output. Stop claiming preserves_uniqueness for these cases to prevent downstream consumers of unique-key metadata (ReduceElision, join inference, etc.) from making incorrect transformations.

### DIFF
--- a/src/expr/src/scalar/func/impls/string.rs
+++ b/src/expr/src/scalar/func/impls/string.rs
@@ -55,7 +55,7 @@ fn cast_string_to_pg_legacy_char<'a>(a: &'a str) -> PgLegacyChar {
     PgLegacyChar(a.as_bytes().get(0).copied().unwrap_or(0))
 }
 
-#[sqlfunc(sqlname = "text_to_name", preserves_uniqueness = true)]
+#[sqlfunc(sqlname = "text_to_name", preserves_uniqueness = false)]
 fn cast_string_to_pg_legacy_name<'a>(a: &'a str) -> PgLegacyName<String> {
     PgLegacyName(strconv::parse_pg_legacy_name(a))
 }
@@ -817,7 +817,7 @@ impl EagerUnaryFunc for CastStringToVarChar {
     }
 
     fn preserves_uniqueness(&self) -> bool {
-        !self.fail_on_len || self.length.is_none()
+        self.length.is_none()
     }
 
     fn inverse(&self) -> Option<crate::UnaryFunc> {


src/expr/src/scalar/func/impls/string.rs
```diff
@@ -817,7 +817,7 @@ impl EagerUnaryFunc for CastStringToVarChar {
     }
 
     fn preserves_uniqueness(&self) -> bool {
-        !self.fail_on_len || self.length.is_none()
+        self.length.is_none()
```
ggevay, line 820: I don't understand this one. The PR description says

> fail_on_len=false truncates to n chars

But isn't this true only when length is Some? If yes, then the original condition seems correct.

def-: The original is backwards. With `length=Some, fail_on_len=false` it returns `true` (preserves uniqueness), but in that case it gets truncated, so isn't unique.

ggevay: Yes, sorry! But then would
```
self.fail_on_len || self.length.is_none()
```
be also correct, and tighter?

def-: `fail_on_len=true, format_str` still silently truncates trailing whitespace past the limit, so "abc " and "abc  " at length 3 both produce "abc"

src/expr/src/scalar/func/impls/string.rs
```diff
@@ -55,7 +55,7 @@ fn cast_string_to_pg_legacy_char<'a>(a: &'a str) -> PgLegacyChar {
     PgLegacyChar(a.as_bytes().get(0).copied().unwrap_or(0))
 }
 
-#[sqlfunc(sqlname = "text_to_name", preserves_uniqueness = true)]
+#[sqlfunc(sqlname = "text_to_name", preserves_uniqueness = false)]
```
antiguru, line 58: False is the default, so you can remove the attribute.

antiguru (approved):

src/expr/src/scalar/func/impls/string.rs
def-: I prefer explicit, it's also explicit in many other sqlfuncs in this file.

PR #36661, opened 2026-05-21
expr: fix date_bin off-by-one for sources exactly on a bin boundary before origin

def-:
When `source < origin` and `tm_diff` was an exact multiple of the stride, the unconditional `tm_delta -= stride_ns` shifted the result back a whole extra bin. Also use `checked_sub` so that the `i64::MIN` nanosecond boundary surfaces `DateBinOutOfRange` instead of silently wrapping.

### DIFF
--- a/src/expr/src/scalar/func.rs
+++ b/src/expr/src/scalar/func.rs
@@ -2024,10 +2024,15 @@ where
         )
     })?;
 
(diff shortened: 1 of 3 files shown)


def-:
Ready for review

DAlperin (approved):
LGTM, thanks

antiguru (approved):

PR #36663, opened 2026-05-21
expr: fix non-injective text_to_"char"/text_to_bytea uniqueness annotation

def-:
These casts were marked `preserves_uniqueness = true`, but neither is injective: `text_to_"char"` keeps only the first byte (so 'a' and 'abc' collapse to the same `"char"`), and `text_to_bytea` accepts both hex and traditional encodings for the same bytes. The annotation made `invert_casts_on_expr_eq_literal_inner` rewrite `col::text = lit` to `col = f_inv(lit)`, which can match rows whose `col::text` does not actually equal `lit` — producing wrong query answers via MFP canonicalization and index literal-constraint lookups.


ggevay (approved):
Nice catches!

PR #36669, opened 2026-05-21
mcp: advertise OAuth via RFC 9728 protected resource metadata

bobbyiliev:
First step toward letting MCP clients that expect the standard OAuth discovery
flow (Claude Desktop Custom Connectors, ChatGPT remote MCP, etc.) connect to
our MCP endpoints. Per the [MCP authorization spec](https://modelcontextprotocol.io/specification/draft/basic/authorization),
the resource server MUST implement RFC 9728 Protected Resource Metadata; this
PR adds that piece and the matching 401 challenge.

Fixes https://linear.app/materializeinc/issue/DEX-31/support-oauth-21-authorization-flow-on-the-mcp-endpoints


def- (dismissed):
I added a test that would be important to fix! host injection

src/environmentd/src/http.rs
```diff
@@ -922,16 +1000,45 @@ impl IntoResponse for AuthError {
...
+                    match HeaderValue::from_str(&value) {
+                        Ok(v) => {
+                            headers.append(http::header::WWW_AUTHENTICATE, v);
+                        }
+                        Err(e) => {
+                            warn!(
+                                "skipping Bearer WWW-Authenticate challenge: invalid header \
+                                 value derived from resource_metadata={resource_metadata:?}: {e}",
+                            );
+                        }
```
jubrad, line 1047: Am I reading this right that an invalid header will only be internally logged? It seems like we'd want to surface this? Maybe not IDK for sure.

src/environmentd/src/http.rs
```diff
@@ -1068,10 +1176,34 @@ async fn http_auth(
...
+            Some(oauth_metadata::MCP_SCOPE),
+        )
+    } else {
+        (None, None)
+    };
+    let challenges = WwwAuthenticateChallenges {
+        include_basic,
+        bearer_resource_metadata,
+        bearer_scope,
+    };
```
jubrad: It seems wrong to have http_auth middleware behave differently based on the path.
Generally I think of routers as dealing with Path dependent extension/layer and middleware being moderately generic about the path they're handling.

I'm not super familiar with this part of the code so I don't know if this is the right call here or not

example
```rust
  // MCP routes
  mcp_router = mcp_router
      .layer(Extension(WwwAuthenticateChallenges {
          include_basic: true,
          bearer_resource_metadata: /* computed once */,
          bearer_scope: Some(MCP_SCOPE),
      }))
      .layer(auth_middleware.clone());

  // Profiling / root routes
  base_router = base_router
      .layer(Extension(WwwAuthenticateChallenges {
          include_basic: true,
          ..Default::default()
      }))
      .layer(auth_middleware.clone())
```

jubrad (approved):
A lot of this is somewhat unfamiliar, but from what I understand the code should work and is pretty clean.

I have one NIT on the additions to http_auth.

src/environmentd/src/http/oauth_metadata.rs
```diff
@@ -0,0 +1,678 @@
...
+
+    let adapter_client =
+        match tokio::time::timeout(ADAPTER_WAIT_TIMEOUT, adapter_client_rx.clone()).await {
+            Ok(Ok(client)) => client,
+            Ok(Err(_)) | Err(_) => {
+                metrics.inc("adapter_unavailable");
+                return (StatusCode::SERVICE_UNAVAILABLE, "adapter not ready").into_response();
+            }
+        };
+    let system_vars = adapter_client.get_system_vars().await;
```
jasonhernandez, line 268: Is this call costly to make if we have significant traffic from agents or possibly malicious requests (since this isn't authenticated)?

src/environmentd/src/http/oauth_metadata.rs
```diff
@@ -0,0 +1,207 @@
...
+    // proxy is also responsible for setting `X-Forwarded-*`.
+    let scheme = forwarded_scheme(&req).unwrap_or("https").to_string();
+    let host = match forwarded_host(&req) {
+        Some(h) => h,
+        None => {
+            warn!("oauth-protected-resource: missing Host header on request");
+            return (StatusCode::BAD_REQUEST, "missing Host header").into_response();
+        }
+    };
+    let resource = format!("{scheme}://{host}/api/mcp");
```
SangJunBak: Rather than `format`, maybe we can use the `url` library?

src/environmentd/src/http/oauth_metadata.rs
SangJunBak: The operation shouldn't be expensive given it effectively returns data from in memory. It does queue up a command in the coordinator however.

src/environmentd/src/http.rs
```diff
@@ -1068,10 +1132,26 @@ async fn http_auth(
     }
 
     let path = req.uri().path();
-    let include_www_authenticate_header = path == "/"
+    let include_basic = path == "/"
         || PROFILING_API_ENDPOINTS
             .iter()
-            .any(|prefix| path.starts_with(prefix));
+            .any(|prefix| path.starts_with(prefix))
+        || path.starts_with("/api/mcp/");
```
SangJunBak: Maybe instead of
```
PROFILING_API_ENDPOINTS
.iter()
.any(|prefix| path.starts_with(prefix))
```

we can create a new const like WWW_AUTHENTICATE_ENDPOINTS where it includes PROFILING_API_ENDPOINTS and /api/mcp? Then it's just 

```
WWW_AUTHENTICATE_ENDPOINTS
.iter()
.any(|prefix| path.starts_with(prefix))
```

src/environmentd/src/http/oauth_metadata.rs
```diff
@@ -363,6 +387,48 @@ mod tests {
         assert_eq!(scheme_for(&req), "http");
     }
 
+    /// Defense in depth against header-smuggling: a Host header that
```
SangJunBak: nit: this comment seems a bit verbose and repeats a lot of the comments below.

src/environmentd/src/http/oauth_metadata.rs
```diff
@@ -103,6 +133,50 @@ pub struct ProtectedResourceMetadata {
...
+                name: "mz_oauth_protected_resource_metadata_requests_total",
+                help: "Total number of requests to the OAuth Protected Resource Metadata endpoint.",
+                var_labels: ["status"],
+            )),
+        }
+    }
+
+    fn inc(&self, status: &'static str) {
+        self.requests.with_label_values(&[status]).inc();
+    }
```
SangJunBak: Rather then have it as a string, it'd be better to have `status` be an enum or have separate helper methods per status. Even though we have it commented, it could become stale. Also it helps us gauge the cardinality of the metric

src/environmentd/src/lib.rs
```diff
@@ -385,6 +385,8 @@ impl Listeners {
         let metrics_registry = config.metrics_registry.clone();
         let metrics = http::Metrics::register_into(&metrics_registry, "mz_http");
         let mcp_metrics = http::mcp_metrics::McpMetrics::register_into(&metrics_registry);
+        let oauth_metadata_metrics =
```
SangJunBak, line 388: Why do we have this separate from `mcp_metrics`?

SangJunBak (approved):
Some suggestions but lgtm for the most part!

src/environmentd/src/http.rs
bobbyiliev: Yeah, I went back and forth on this one. The host going into `resource_metadata` is already validated by `resolve_host`, so I think `HeaderValue::from_str` failing here is genuinely unreachable in practice. The alternatives felt worse though, eg. 500-ing breaks the auth flow on a path that's really just constructing a response header, and .expect() would let one weird codepath crash the server. Falling back to Basic seemed like the safest degradation, with the WARN there as a "if this ever fires, something upstream regressed" signal. Happy to bump it to ERROR if you'd rather have it louder.

src/environmentd/src/http/oauth_metadata.rs
bobbyiliev: Considered it but I think `format!` might be safer here. The only variable is `host`, which `resolve_host` already parses through `http::uri::Authority` and rejects `userinfo`, control chars, whitespace, etc. before returning. Open to suggestions though!

src/environmentd/src/http.rs
bobbyiliev: Yeah, agreed, that one was bugging me too. Added a `BearerChallengeConfig` extension and attached it to `mcp_router`. The middleware reads the extension now instead of sniffing the path, so it's path-agnostic again. Also if we ever want to advertise OAuth on another route, it's just attaching the same extension over there.

src/environmentd/src/http/oauth_metadata.rs
bobbyiliev: Good point! Trimmed!

src/environmentd/src/http.rs
bobbyiliev: Ah yes @jubrad  made a similar point here and I pushed a refactor in one of the last commits. So now `/api/mcp` no longer participates in the path-prefix check at all.

It's now opted in via a `BearerChallengeConfig` extension on `mcp_router`, and the middleware just reads the extension. So the remaining `path == "/" || PROFILING_API_ENDPOINTS...` predates this PR. Happy to consolidate those into one const as a follow-up if you'd still like it?

src/environmentd/src/lib.rs
bobbyiliev: Yep they're conceptually separate things, eg `McpMetrics` tracks the MCP request handling itself (per-tool latency, request counts), while `OauthMetadataMetrics` tracks the OAuth Protected Resource Metadata discovery endpoint, which lives in its own module and isn't really an MCP endpoint (it's metadata about how to authenticate to one).

So each metrics struct lives next to its handler, which matches the pattern McpMetrics is already using. I'd rather not couple the two unless we have a concrete reason to.

src/environmentd/src/http/oauth_metadata.rs
bobbyiliev: Good call, done in one of the last commits, introduced a `MetricStatus` enum and the call sites now use `MetricStatus::Ok`, `::Disabled`, etc.

src/environmentd/src/http/oauth_metadata.rs
bobbyiliev: @SangJunBak's right, just to back it up, the adapter wait is bounded by a 2s timeout (`ADAPTER_WAIT_TIMEOUT`), and the response carries `Cache-Control: private, max-age=3600` so well-behaved clients only hit it once an hour.

We do have a precedent for `GlobalConcurrencyLimitLayer` + load_shed on `/api/webhook/*` that might be straightforward to apply here, but I held off because the handler itself is cheap and there's no production signal yet. @jasonhernandez do you think this is worth adding now as defense-in-depth, or fine to defer to a follow up PR?

bobbyiliev: Created a tracking issue for this and we can implement if we agree that this is really neeed: https://linear.app/materializeinc/issue/DEX-37/consider-rate-limit-concurrency-cap-on-oauth-protected-resource

PR #36670, opened 2026-05-21
adapter: introduce UserWriteResponder for PendingWriteTxn::User

aljoscha:
Wrap the per-session `PendingTxn` inside `PendingWriteTxn::User` in a new `UserWriteResponder` enum (single `Session(PendingTxn)` variant for now). Construction and pattern-matching sites are updated to thread the value through `UserWriteResponder::Session(...)`; `DeferredWrite` keeps its raw `pending_txn: PendingTxn` because the deferred-op machinery is session-only and doesn't gain by going through the responder.

This is preparation: the read-then-write OCC path will need to commit writes whose result is delivered via a oneshot rather than retired through an `ExecuteContext`. Introducing the abstraction with one variant first makes the OCC commit additive (a new `Internal` variant plus the call sites that produce/consume it) instead of also having to rename the field at every existing site. No behavior change.

### DIFF
--- a/src/adapter/src/coord/appends.rs
+++ b/src/adapter/src/coord/appends.rs
@@ -131,18 +131,29 @@ pub(crate) enum BuiltinTableUpdateSource {
     Background(oneshot::Sender<()>),
 }
 
+/// Where to deliver the result of a [`PendingWriteTxn::User`] write.
+#[derive(Debug)]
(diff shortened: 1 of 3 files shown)


ggevay (approved):

aljoscha:
tyty! :bow:

PR #36671, opened 2026-05-21
Oidc password authentication on uninitialized Oidc fix

SangJunBak:
Before we were getting throws on password login when Oidc was never initialized. This allows us to not have to check if the oidcManager is available everytime we want Oidc auth details.

### Motivation

Bug found during product review when logging in as mz_system before Oidc was configured

### Verification
Run locally against Oidc instance and login as mz_system. Before you would crash because we were trying to access properties in useAuth that didn't exist. Now it doesn't crash

### DIFF
--- a/console/src/external-library-wrappers/oidc.ts
+++ b/console/src/external-library-wrappers/oidc.ts
@@ -11,9 +11,20 @@
 /**
  * This file is a facade for the react-oidc-context / oidc-client-ts libraries.
  */
-export { AuthProvider, hasAuthParams, useAuth } from "react-oidc-context";
+export {
+  type AuthContextProps,
+  AuthProvider,
+  hasAuthParams,
+  // We should not use this hook directly because it requires AuthProvider to be mounted,
+  // which it only is when OIDC is available. Instead, use AppConfigSwitch.
+  useAuth,
+} from "react-oidc-context";
 
+import { useQuery } from "@tanstack/react-query";
 import { UserManager, WebStorageStateStore } from "oidc-client-ts";
+
+import { apiClient } from "~/api/apiClient";
+import { useAppConfig } from "~/config/useAppConfig";
 
 export interface OidcConfig {
   issuer: string;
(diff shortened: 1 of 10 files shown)


SangJunBak:
I feel like a follow up to this is an e2e test framework that:
1. Runs the Console locally
2. Stands up a Materialize instance with the right config using `mzcompose.ts` 
3. Uses playwright to just login as mz_system using password auth

Then perhaps using frontegg-mock, we could do something similar for oidc.

leedqin (approved):
LGTM

console/src/external-library-wrappers/oidc.ts
```diff
@@ -119,3 +130,31 @@ export class MzOidcUserManager {
...
+        !apiClient.oidcManagerInitializationPromise
+      ) {
+        return null;
+      }
+      return apiClient.oidcManagerInitializationPromise;
+    },
+    enabled: isOidc,
+    staleTime: Infinity,
+    retry: false,
+    retryOnMount: false, // Do not retry on mount, otherwise we will retry indefinitely
```
leedqin, line 158: this is cool! Didn't know retryOnMount was a thing

console/src/platform/auth/Login.tsx
```diff
@@ -154,9 +154,19 @@ const SsoLoginLink = () => {
   const [error, setError] = useState<string | null>(null);
   const auth = useAuth();
 
-  if (!auth) {
-    return null;
-  }
+  const oidcAuthErrorMessage = auth?.error?.message ?? null;
+
+  useEffect(() => {
```
leedqin:
```suggestion
  const oidcError = auth.error?.message? ?? null;
  const displayError =
    error ||
    (oidcAuthErrorMessage &&
      `${oidcAuthErrorMessage}. It looks like there may be an issue with the sign-in configuration. Please review your OIDC settings or check the console logs for more information.`) ||
    null;
    
      return (
    <VStack spacing="2" alignItems="center">
      {displayError&& <Alert variant="error" minWidth="100%" message={error} />}
```

leedqin: I imagine this flow should work
1. . App mounts → `auth.error = null` → `oidcError = undefined` → `displayError = null` → no Alert.
2. `react-oidc-context` populates `auth.error`  → AuthProvider re-renders consumers.
3. `SsoLoginLink` re-renders → `oidcError` is recomputed from `auth.error.message` → `displayError` is the friendly message → Alert appears.
4. User clicks SSO → `handleSsoLogin` runs → if `signinRedirect()` rejects, `setError(...)` runs → next render, `error` (local) wins via `||` short-circuit → Alert swaps to that message.
5. Issuer is fixed and the user reloads → `auth.error` clears → `displayError = null` → Alert disappears.

SangJunBak: Makes sense to me!

PR #36673, opened 2026-05-21
platform-checks: Fix version gates

def-:
Seen in https://buildkite.com/materialize/release-qualification/builds/1234


bosconi (approved):
Looks good.

PR #36674, opened 2026-05-21
[catalog] convert mz_sources table to materialized view

mtabebe:
Problem:
mz_sources was a builtin table populated imperatively in Rust code in builtin_table_updates.rs.

Ideologically we want to move these to views over the catalog.

Solution:
Convert mz_sources into a materialized view that reads directly from mz_catalog_raw. User sources are derived by parsing their stored create_sql. Builtin and log sources, which don't appear in mz_catalog_raw, are exposed via a new internal view mz_internal.mz_builtin_sources

Testing:
- Updated information_schema_tables.slt to expect MATERIALIZED VIEW for mz_sources
- Added mz_builtin_sources to mz_internal.slt and mz_catalog_server_index_accounting.slt
- updated catalog.td to move mz_sources from the tables section into the materialized views section.

Part of SQL-118.

Port of MaterializeInc/materialize#36136 (adapted for the split builtin file structure and current OID range).


src/adapter/src/catalog/open/builtin_schema_migration.rs
```diff
@@ -171,6 +171,12 @@ static MIGRATIONS: LazyLock<Vec<MigrationStep>> = LazyLock::new(|| {
             MZ_CATALOG_SCHEMA,
             "mz_secrets",
         ),
+        MigrationStep::replacement(
+            "26.26.0-dev.0",
```
def-: Should be 26.27.0-dev.0 I assume

def- (changes requested):
Do we intentionally remove the `progress` subsource type? If so we should remove it from the documentation too. If not, we should keep it.

Full nightly will be required for this PR, triggered: https://buildkite.com/materialize/nightly/builds/16516 Lots of related failures, both about `progress` type and the now missing `cluster_id`

src/catalog/src/builtin/mz_catalog.rs
```diff
@@ -888,127 +888,198 @@ pub static MZ_SSH_TUNNEL_CONNECTIONS: LazyLock<BuiltinTable> = LazyLock::new(||
...
+        SELECT
+            data->'key'->>'schema_name' AS schema_name,
+            data->'key'->>'object_name' AS name,
+            's' || (data->'value'->>'catalog_id') AS id
+        FROM mz_internal.mz_catalog_raw
+        WHERE
+            data->>'kind' = 'GidMapping' AND
+            data->'key'->>'object_type' = '2'
+    ),
+    builtin_sources AS (
```
SangJunBak: Given mz_builtin_sources is essentially a hardcoded list of all the builtin logs/sources created on startup, won't we panic/get in an invalid state when its definition changes but someone forgets to put in a builtin item migration for mz_sources? I'm thinking of the following case:

1. Someone is on this version and now they have mz_builtin_sources and the new mz_sources mv
2. On the next version, we add a new builtin source. Now mz_builtin_sources' definition changes to include that new source
3. The developer doesn't think to create a MigrationStep::replacement for mz_sources since it's not obvious we need a migration
4. Because mz_sources is persisted as a builtin materialized view, environmentd expects its definition/fingerprint to be the same

SangJunBak (commented):
Had a concern that might be a bug. Let me know if it's a valid concern!

src/catalog/src/builtin/mz_catalog.rs
mtabebe: Oooh that is a really great point. I hadn't considered that... I think the way to avoid this is to build the SQL values directly and that will trigger a fingerprint  mismatch, which will make environmentd panic. So it will/should be noticeable

SangJunBak (approved):
Nice!

def- (approved):
No further complaints from QA

PR #36675, opened 2026-05-21
catalog-protos: gate Arbitrary derives behind feature/cfg(test)

antiguru:
### Motivation

`proptest-derive` was expanding `Arbitrary` on ~2200 types in `src/catalog-protos/` on every build, even though the impls are only needed by tests in `mz-catalog`.

### Description

`mz-catalog-protos` gains a `proptest` feature. Each `derive(Arbitrary)` and `#[proptest(...)]` field attribute is wrapped in `#[cfg_attr(any(test, feature = "proptest"), ...)]`. A self-dep in `[dev-dependencies]` enables the feature when building the crate's own tests.

`mz-catalog` moves `proptest`/`proptest-derive` to dev-dependencies, gates its own `derive(Arbitrary)` on durable objects behind `cfg(test)`, and restructures the `json_compatible!` macro so the `unsafe impl JsonCompatible` stays unconditional (production migrations rely on it) while `assert_impl_all` and `proptest!` blocks are `cfg(test)`-only.

### Compile-time impact

Measured `mz-catalog-protos` self-time via `cargo --timings`, deps pre-warmed, 3 iterations each, median reported.

| Profile | Before | After | Reduction |
| --- | --- | --- | --- |
| `cargo check -p mz-catalog-protos` | 16.5s | 13.3s | 19% |
| `cargo build -p mz-catalog-protos --profile=optimized` | 41.6s | 16.6s | 60% |

Optimized win dominated by skipped codegen on the ~2200 `Arbitrary` impls. Numbers cover catalog-protos lib only; downstream consumers of these types likely see further savings not measured here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5874,6 +5874,7 @@ dependencies = [
  "anyhow",
  "md-5",
  "mz-audit-log",
+ "mz-catalog-protos",
  "mz-compute-types",
  "mz-controller-types",
  "mz-expr",
--- a/src/catalog-protos/Cargo.toml
+++ b/src/catalog-protos/Cargo.toml
@@ -17,14 +17,17 @@ mz-repr = { path = "../repr" }
 mz-sql = { path = "../sql" }
 mz-storage-types = { path = "../storage-types" }
 num_enum.workspace = true
-proptest.workspace = true
-proptest-derive.workspace = true
+proptest = { workspace = true, optional = true }
+proptest-derive = { workspace = true, optional = true }
 serde.workspace = true
 serde_repr.workspace = true
 
 [dev-dependencies]
+mz-catalog-protos = { path = ".", features = ["proptest"] }
 mz-expr = { path = "../expr" }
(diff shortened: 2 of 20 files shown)


DAlperin (approved):
LGTM

antiguru:
Thanks for the review!

antiguru:
LLVM-IR codegen comparison via `cargo llvm-lines -p mz-catalog-protos`:

| Metric | Before | After | Reduction |
| --- | --- | --- | --- |
| Total LLVM lines | 4,884,312 | 64,970 | 98.7% |
| Total monomorphizations | 232,853 | 7,475 | 96.8% |

Before, top functions are proptest internals: `Weak::drop`, `Arc::new`/`drop`/`clone`, `LazyValueTree::maybe_init`, `TupleUnionValueTree::simplify`, `Strategy::prop_map`. After, those are gone — top functions are normal `RustType` `from_proto`/`into_proto` conversions plus standard BTreeMap/slice sort helpers.

75x more IR before the change explains the 60% optimized-build win.

PR #36681, opened 2026-05-21
ore: box-pin futures inside task::spawn

frankmcsherry:
`mz_ore::task::spawn` was generic over `Fut: Future`, passing the future straight through to `tokio::spawn`. As a result, every distinct call-site future type re-instantiated tokio's full task machinery (`Cell::new`, `RawTask::new`, `Harness::poll_inner`, `Core::poll`, the `catch_unwind` wrapper, ...) — a major source of monomorphization-driven LLVM IR bloat across the workspace.

This change boxes the future once before handing it to `tokio::spawn`, so tokio's task machinery is monomorphized over
`Pin<Box<dyn Future<Output = Fut::Output> + Send>>` (one instantiation per output type) rather than over every distinct future type at every call site.

  Measured on `mz-adapter` via `cargo llvm-lines -p mz-adapter`:

  | Metric | Before | After | Δ |
  |---|---:|---:|---:|
  | LLVM lines | 5,934,717 | 4,992,813 | **-15.9%** |
  | Monomorphization copies | 169,250 | 126,088 | -25.5% |
  | `tokio` task machinery | 896K | 286K | **-68%** |
  | `panic/FnOnce` wrappers | 254K | 70K | **-73%** |
  | Distinct task instantiations | 488 | 120 | -75% |

A full cascade rebuild of `mz-environmentd` after touching `mz-ore` dropped from 238s to 223s of CPU time (-6%). Wall-clock was within noise on this workload because that particular rebuild is highly parallel; scenarios where `mz-adapter` is on the critical path (e.g., incremental edits to adapter, CI on small machines) should see wall-clock benefit as well. The same helper is used throughout the workspace, so other crates that spawn tasks (compute, storage, environmentd, etc.) will see comparable reductions.

Runtime cost: one heap allocation per spawn call. The helper is used at human-action frequency (request handling, background workers, etc.), not on hot per-row / per-batch paths, so the overhead is negligible in practice.

Follow-on opportunities not bundled here:

* `JoinSet::spawn_named` has the same generic shape and would benefit from the same treatment.
* The remaining 120 task instantiations come from distinct `Fut::Output` types; collapsing them further would require erasing the output type too (e.g., returning a oneshot channel rather than `JoinHandle<Out>`), which is a more invasive API change.

Authored by Claude Code 🤖 .

### DIFF
--- a/src/ore/src/task.rs
+++ b/src/ore/src/task.rs
@@ -174,6 +174,10 @@ where
     Fut: Future + Send + 'static,
     Fut::Output: Send + 'static,
 {
+    // Box the future so tokio's task machinery is monomorphized over
+    // `Pin<Box<dyn Future<Output = Fut::Output> + Send>>` per output type,
+    // rather than over every distinct future type at every call site.
+    let future: Pin<Box<dyn Future<Output = Fut::Output> + Send>> = Box::pin(future);
     #[allow(clippy::disallowed_methods)]
     JoinHandle::new(tokio::spawn(future))
 }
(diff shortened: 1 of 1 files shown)


antiguru (approved):
Looks good, thank you.

PR #36686, opened 2026-05-22
workload-replay: Handle larger captured workloads

def-:
Captured workloads from production environments often have features the local replay did not yet support; replaying them previously failed at various points before useful test work could happen.

* config: add extra cluster replica sizes
* objects: accept quoted database names with non-identifier characters (e.g. hyphens) in CREATE CONNECTION; parse double-quoted REFERENCE / EXTERNAL REFERENCE parts so dotted references like `"db-name".schema.table` are split correctly.
* objects: upstream CREATE TABLE statements now use IF NOT EXISTS — the same physical Postgres/MySQL/SQL Server table can back multiple MZ sources and we were failing the second time.
* objects: strip options from captured SQL that don't survive local replay: empty `TEXT COLUMNS = ()` (planner panic), `EXCLUDE COLUMNS` (names upstream columns we don't recreate), Kafka `START OFFSET = (…)` (local topics have 1 partition), `VERSION = N` on CREATE SINK (internal-only option).
* column: extend type coverage so `<scalar>[]` arrays, `interval`, `oid`, `real`, and Materialize-specific list/record/aclitem variants produce sensible non-NULL values; `character` (no size) now emits a single char so it fits CHAR(1).
* column: function-call defaults like `pg_catalog.now()` are skipped in COPY and Kafka paths, where they would be inserted as literal strings and rejected.
* column: avro_type falls back to `string` for non-primitive types so fastavro can parse the generated schema.
* util / mzcompose: workload files can now be stored compressed as `.yml.zst` and are loaded transparently via the new `load_workload` helper. Glob `*.yml` also picks up the compressed variants.

### DIFF
--- a/ci/nightly/pipeline.template.yml
+++ b/ci/nightly/pipeline.template.yml
@@ -306,6 +306,7 @@ steps:
               run: benchmark
               args:
                 - --factor-initial-data=0.01
+                - --skip-large
                 - --compare-against
                 - common-ancestor
         agents:
--- a/ci/release-qualification/pipeline.template.yml
+++ b/ci/release-qualification/pipeline.template.yml
@@ -229,7 +229,7 @@ steps:
         label: "Long Workload Replay (100% initial data)"
         depends_on: build-x86_64
         timeout_in_minutes: 1440
-        parallelism: 3
+        parallelism: 4
         plugins:
           - ./ci/plugins/mzcompose:
               composition: workload-replay
@@ -237,11 +237,9 @@ steps:
               args:
                 - --factor-initial-data=1
                 - --runtime=3600
-                - --compare-against
-                - common-ancestor
                 - --skip-without-data-scale
         agents:
-          queue: hetzner-x86-64-dedi-32cpu-128gb
+          queue: hetzner-x86-64-dedi-48cpu-192gb
 
   - group: SQLsmith
     key: sqlsmith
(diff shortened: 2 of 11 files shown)


def-:
Currently blocked by https://linear.app/materializeinc/issue/CLU-95/thread-coordinator-panicked-at-srccompute-clientsrcas-of & https://github.com/MaterializeInc/database-issues/issues/9656#issuecomment-4523776124

Edit: Rebased on top of potential fix, retrying: https://buildkite.com/materialize/release-qualification/builds/1250

ggevay:
@def-, I'm currently looking into https://linear.app/materializeinc/issue/CLU-95/thread-coordinator-panicked-at-srccompute-clientsrcas-of
Did the test where that happened involve upgrades?

def-:
No upgrades

def-:
Current status: OoM and timeout: https://buildkite.com/materialize/release-qualification/builds/1274#019ed01f-e980-431a-8764-5e3bdc54b5fa

def-:
Current status: Blocked by https://linear.app/materializeinc/issue/PER-10/thread-persist0002-panicked-at-srcpersist-clientsrcinternalgcrs4159 Edit: Fixed

def-:
Test run: https://buildkite.com/materialize/release-qualification/builds/1314

Ready for review

bosconi (approved):
Looks good.

PR #36691, opened 2026-05-22
design: add doc for cluster autoscaling and background reconfiguration

aljoscha:
[Rendered](https://github.com/aljoscha/materialize/blob/design-cluster-autoscaling/doc/developer/design/20260522_cluster_autoscaling.md)

Resolves SQL-315

### DIFF
--- a/doc/developer/design/20260522_cluster_autoscaling.md
+++ b/doc/developer/design/20260522_cluster_autoscaling.md
@@ -1,0 +1,261 @@
+# Cluster Autoscaling and Background Reconfiguration
+
+- Associated: (TBD — link epics/issues)
+
+## The Problem
+
+Two user-facing capabilities motivate this work:
+
+1. **Background graceful cluster reconfiguration.** Today, `ALTER CLUSTER ... SET (SIZE = ...)` with the graceful (zero-downtime) strategy requires the SQL session to remain open for the duration of the reconfiguration — the session holds the wait-for-hydration stage. Long-running reconfigurations are fragile: any process or session interruption — a network blip, a client timeout, an SQL tool closing, an `environmentd` restart — aborts the reconfiguration. The user experience we want is: the statement returns immediately, and the reconfiguration continues in the background, surviving restarts and disconnects.
+
+2. **Hydration burst (autoscaling for hydration).** Users want to express, at cluster creation:
+
+   ```sql
+   CREATE CLUSTER my_cluster (
+     SIZE = '100cc',                       -- steady-state replica size
+     AUTO SCALING STRATEGY = (
+       ON HYDRATION (
+         HYDRATION SIZE = '3200cc'         -- burst replica size while objects are not hydrated
+         LINGER DURATION = '600s'          -- optional duration that the burst replica stays on after steady-state replicas are hydrated
+       ),
+     )
+   );
+   ```
+
+   When the cluster has any object that is not fully hydrated, the cluster controller spins up an additional replica at the configured `HYDRATION SIZE` to accelerate hydration. Once any steady-state replica has fully hydrated all currently-existing objects on the cluster, the burst replica is shut down (after an optional `LINGER DURATION`).
+
+Beneath both capabilities is a shared architectural problem: today, cluster scheduling decisions are made in an ad-hoc fashion. `ON REFRESH` scheduling lives in a single file driven by a coordinator interval; graceful reconfiguration lives as executor stages bound to a SQL session. Adding hydration-burst as another one-off would compound the inconsistency, and the session-bound nature of graceful reconfiguration cannot deliver requirement (1) regardless. We need a coherent place for cluster autoscaling decisions to live.
+
+### Current state
+
+- **Graceful reconfiguration mechanics** live in `src/adapter/src/coord/sequencer/inner/cluster.rs` as a three-stage state machine driven by the executor. Stage 1 creates pending replicas at the new size with a durable `pending: bool` flag. Stage 2 polls for hydration of the pending replicas. Stage 3 drops the old replicas, renames the pending ones (removing the `-pending` suffix), flips `pending` to `false`, and updates the cluster's durable `size` field.
+- **The user's intent during a graceful reconfiguration is not durable.** The `Op::UpdateClusterConfig` that sets the new cluster `size` is deliberately held back until the finalization stage. Mid-reconfig, the durable catalog shows the old `size` and a pending replica at the new size; the *intent* "the user asked for size X" lives only in transient session state (the connection's pending-alter tracker and the executor stage's strategy/timeout).
+- **Existing scheduling policy lives in `cluster_scheduling.rs`.** It runs on a coordinator timer interval, computes decisions for managed clusters with `SCHEDULE = ('on-refresh', ...)`, and sends `Message::SchedulingDecisions` to the coordinator's internal message channel, which then sequences ALTER operations. Decisions are recorded in `mz_audit_events` via `SchedulingDecisionsWithReasonsV2`. Conflicts with an in-flight graceful reconfiguration are absorbed by the scheduler swallowing the planner's `AlterClusterWhilePendingReplicas` reject — an implicit coupling the controller model replaces.
+- **Hydration signal is available in-process from the controller(s).** Per-replica, per-collection hydration state is tracked in-memory by the controller(s) and updated reactively as frontier information flows in from replicas. An in-process API already exists for asking "are all of these collections hydrated on any of these replicas?" and is used today by the graceful reconfiguration wait stage.
+- **Audit log already records scheduling decisions with reasons** (`SchedulingDecisionsWithReasonsV2`). This is the natural place to record additional autoscaling events.
+- **Cluster configuration is fully durable in the catalog**, including `ClusterVariantManaged { size, replication_factor, availability_zones, schedule, logging, optimizer_feature_overrides }`.
+
+## Success Criteria
+
+- A user can issue `ALTER CLUSTER ... SET (SIZE = ...)` and the statement returns immediately. The reconfiguration continues in the background, surviving session disconnects and `environmentd` restarts, and is observable via SQL.
+- A user can bound an individual reconfiguration with the existing `WITH (WAIT ...)` options (`WAIT FOR '...'` or `WAIT UNTIL READY (TIMEOUT '...', ON TIMEOUT COMMIT|ROLLBACK)`); the deadline and the on-timeout action are enforced durably and survive session disconnect and `environmentd` restart along with the reconfiguration itself.
+- A user can specify an `AUTO SCALING STRATEGY` (with `ON HYDRATION` and its `HYDRATION SIZE`) at cluster create or alter. The burst replica is spun up when warranted and shut down when no longer needed, with no further user action.
+- The existing `SCHEDULE = ('on-refresh', ...)` behavior is preserved, ideally as one strategy within a common autoscaling framework rather than as a parallel mechanism.
+- All autoscaling decisions and replica lifecycle changes are recorded with reasons and surfaced to the user.
+- A new autoscaling strategy can be added without restructuring the framework.
+- Operators can disable the hydration-burst strategy across an environment via a break-glass flag, leaving graceful reconfiguration and `ON REFRESH` scheduling (the other strategies in the framework) untouched.
+
+## Out of Scope
+
+- `AUTO SCALING STRATEGY` combined with `SCHEDULE = ('on-refresh', ...)`.
+- More than one concurrent burst replica per cluster.
+- Dedicated push channel for reconfiguration completion (e.g., NOTICE-style).
+
+## Solution Proposal
+
+### Summary
+
+We introduce a **cluster controller** as a dedicated task inside `environmentd`. It is the single decision-maker for the replica set of every cluster. It operates as a reconciler: it reads desired cluster state from the durable catalog, observes the actual replica set and live status signals, computes a desired replica set per cluster by combining a set of **strategies**, and emits catalog-change commands to the Coordinator via a message channel. The Coordinator remains the sole writer of catalog state.
+
+ALTER CLUSTER operations are reshaped so that the user's intent is written durably and immediately to the catalog — as an in-flight target record, leaving the cluster's realized config in place until the controller cuts over; the controller then reconciles from that intent. Graceful reconfiguration becomes a strategy within the controller framework, as does hydration burst. The existing `ON REFRESH` scheduling is lifted into the framework.
+
+### Strategies
+
+A **strategy** is two functions:
+
+- `update_state` of the form `(durable config, observed status, current time) → state updates`. This allows the strategy to update durable catalog state. For example to update a timeout value or to update the durable size of a cluster once a graceful reconfiguration is considered ready.
+- `desired_replicas` of the form `(durable config, observed status, current time) → partial desired state`. This is the contribution it wants to make to the cluster's desired replica set. As of v1, the replica set is only additive, strategies cannot remove existing replicas or influence the desired state of other strategies. As we see below, dropping of existing replicas happens when no strategy "desires" them.
+
+Initial strategies. The implicit baseline is always present; the rest engage per cluster:
+
+- **Implicit baseline.** Desires the replicas implied by the realized config, `replication_factor` replicas at `cluster.size`, with the configured AZ and other cluster shape. It is what lets the policy strategies be purely additive: it holds the steady set, so they only ever add to it.
+
+- **Graceful reconfiguration.** Engaged when `ALTER` writes a durable `reconfiguration` record. This desires `replication_factor` replicas — the target's, since an `ALTER` can change it — at the record's `target` **config shape** (target size, logging, and availability-zone list). When `update_state` observes the target replicas present and hydrated, it updates the cluster configuration (`cluster.size := target`, ...) and clears the `reconfiguration` record. `update_state` also reads the `deadline` and `on_timeout`. Success takes precedence: a tick that sees the target replicas hydrated cuts over even if the deadline has passed. Otherwise, once `now > deadline` with the target not fully hydrated, `update_state` applies `on_timeout` — the default `ROLLBACK` stops contributing this reconfiguration's target replicas and keeps the record as a tombstone, so the strategy stays a no-op until another `ALTER`; `COMMIT` instead cuts over to the still-unhydrated target. Cut-over keys on **hydration** — today's graceful-reconfiguration signal — not a stronger caught-up check: hydration already guarantees correct answers, so a caught-up check would only avoid a brief post-cut-over latency bump, a possible later refinement. One consequence worth noting: an OOM- or crash-looping target replica never hydrates, so it can never cut over — the deadline fires and the default `on_timeout` reverts. No special OOM-loop detection needed.
+
+- **Hydration burst.** When the cluster's `AUTO SCALING STRATEGY` sets `ON HYDRATION (HYDRATION SIZE = ...)`, the cluster is On (`replication_factor > 0`), and no realized-config replica has all current objects hydrated, `update_state` writes a `burst` record (its size and linger duration). When that record is present this desires one extra replica at `HYDRATION SIZE`. When `update_state` notices that at least one steady-state replica is hydrated, it records that timestamp. Once time has passed that timestamp plus linger duration it removes the burst record. Additionally, when there is a burst record and we recorded successful hydration of the steady-state replicas, but the steady-state replicas become un-hydrated again, we reset burst state so that the linger duration can restart after the next successful hydration. Finally, `update_state` clears the `burst` record — regardless of linger — whenever a burst is no longer warranted by current config: the `AUTO SCALING STRATEGY` was removed or its `HYDRATION SIZE` changed, or the cluster was turned off (`replication_factor = 0`). Because `desired_replicas` keys the burst replica purely on the record's presence, this cleanup is what stops a stale record from pinning a burst replica on a cluster that is off or no longer configured for burst; on a `HYDRATION SIZE` change a fresh record is written at the new size on a later tick if a burst is still warranted.
+
+- **`ON REFRESH` scheduling.** Contributes one replica at `cluster.size` while the cluster is inside a refresh window and nothing otherwise, based on REFRESH MV write frontiers and the configured hydration-time estimate. A scheduled cluster's `replication_factor` is normalized to `0`, so the implicit baseline desires no replicas and `ON REFRESH` is the sole contributor of replicas. It does not apply to MANUAL clusters.
+
+Both functions of every strategy are pure: same inputs, same output, no side effects. The controller is the sole mutator, it applies the `update_state` results and the create/drop commands that close the gap between the combined desired set and the actual replicas.
+
+#### Combining strategy outputs
+
+Each strategy's `desired_replicas` contribution is a **multiset of replica slots keyed by config shape** — a count per shape, not a set: the implicit baseline alone wants `replication_factor` slots at one shape. Strategies name no replicas; they describe shapes and counts, and the controller assigns concrete names as it fills slots, so two strategies can never contend over a replica *name*. Contributions combine **per shape, taking the max count across strategies**, so a slot survives as long as *any* strategy desires it and overlapping requests for the same shape don't double it. Strategies are purely additive, so they cannot contradict one another. The controller diffs the combined desired set against the actual replica set and emits the gap: a create for a slot the desired set calls for but actual lacks, a drop for a replica whose shape no strategy still desires. This composes the strategies with no state-machine interaction between them. The controller retains each strategy's individual contribution, not just the merged set, so it can attribute every command it emits: a create records which strategies desired the replica ("required by graceful reconfiguration and hydration burst"), a drop records that none did. These attributions are the reasons carried on the audit events.
+
+### The cluster controller
+
+The controller runs as its own task within `environmentd`. Its inputs are:
+
+- **Durable cluster configuration**, observed via the catalog.
+- **Live status signals** from the compute and storage controllers: hydration state, current frontiers.
+- **A periodic tick**, plus an out-of-band **wake** on events that shouldn't wait for it — notably an `ALTER` that has just written a `reconfiguration` record. Each tick or wake runs one full reconcile pass (both phases); live status changes (e.g. a replica finishing hydration) are observed on the next pass rather than waking the controller themselves.
+
+v1 reads these inputs in-process; a multi-`environmentd` future would swap them for subscriptions to builtin relations exposing the same signals (see [Alternatives Considered](#alternatives-considered)).
+
+Its outputs are **decisions** sent to the Coordinator over a command channel: requests to create or drop a replica, or to update a cluster's durable state (a cut-over, a record write or clear), each with a reason. The Coordinator transacts them through the catalog and orchestrator. The channel is **request/response**: within a tick the controller awaits the Coordinator's confirmation that the `update_state` batch has been transacted before it computes `desired_replicas`.
+
+The controller is not the catalog writer. Coordinator remains the only writer.
+
+The controller does not maintain its own in-memory state, the source of truth is always the catalog plus live signals.
+
+#### Why a separate task rather than running on the coordinator's main loop
+
+The coordinator main loop is a serial command processor and a place we want to do less work in over time, not more. Periodic decision-making, polling of status signals, and computation over potentially many clusters should not block command handling. A separate task with a clean message-channel boundary also makes the controller independently testable and gives us optionality to extract it to a separate process later without further refactoring.
+
+#### Why a reconciler rather than an imperative pipeline
+
+A reconciler computes "desired replica set" purely from durable state plus observable status. This has several properties we want:
+
+- **Crash-safe by construction.** Any state needed to continue an operation across a restart is, by definition, in the catalog or derivable from it.
+- **Composable.** Multiple strategies (graceful reconfig, hydration burst, on-refresh, future ones) each contribute to the desired set; the controller merges them by union. No state-machine interactions between strategies.
+- **Deterministically auditable.** Same inputs always yield the same decisions; audit entries explain each decision from durable state plus observed status.
+
+#### The reconcile tick
+
+Each tick reconciles every cluster in two phases, separated by a barrier:
+
+1. **`update_state`.** The controller reads a fresh catalog snapshot and runs every strategy's `update_state`, collecting their durable writes: cut-overs (`cluster.size := target`), `reconfiguration` and `burst` record writes and clears. It submits that batch to the Coordinator and **awaits confirmation that it has been transacted.**
+2. **`desired_replicas`.** With those writes applied, the controller runs every strategy's `desired_replicas` against the now-current state, unions the contributions into the [combined desired set](#combining-strategy-outputs), diffs it against the actual replica set, and emits the creates and drops that close the gap.
+
+#### State convergence
+
+Each pass the controller **emits only the commands that close the gap**: a create for a replica that should exist but doesn't, a drop for one that shouldn't. It has the observed state, so it does not re-assert the full desired set every tick; it sends work only when there is a delta, keeping needless transactions off the coordinator's serial loop.
+
+Commands name **explicit replicas** — *create replica `<name>` at this config*, *drop replica `<name>`* — rather than asking the Coordinator to make some count of replicas at a config shape exist. The controller derives names deterministically from the observed actual set, picking a fresh name for each create and a specific existing replica for each drop; it never renames a replica to make the sets line up. This keeps re-emission harmless: the controller's view of the actual set lags, and across a crash or re-election a later pass can recompute the same gap, but a create of a name that already exists and a drop of one already gone are both no-ops, and a stale command is reconciled away on the next pass.
+
+Both the create/drop batch and the durable **state** writes `update_state` makes — cutting the realized config over to an in-flight target (`cluster.size := target`, clearing the `reconfiguration` record), writing or clearing the `burst` record — are applied under a **compare-and-append** guard, the pattern we know from persist: each batch carries the controller's view of the durable cluster state it was derived from (the cluster config plus the `reconfiguration` and `burst` records), and the Coordinator applies it only if that state still holds. A user `ALTER` that lands between the controller deriving its commands and the Coordinator applying them changes that state, so the guard fails, the batch is rejected, and the controller recomputes from the new state on the next tick. This is what keeps a drop derived from a pre-`ALTER` snapshot from retiring a replica the `ALTER` has since made desired.
+
+#### Example: a graceful size change
+
+Take a MANUAL cluster at `100cc`, `replication_factor = 2`, serving replicas `r1`, `r2`. The user runs `ALTER CLUSTER ... SET (SIZE = '200cc')`.
+
+1. **`ALTER` returns.** It writes `reconfiguration = { target: 200cc, deadline }` and leaves `cluster.size = 100cc`.
+2. **Reconcile.** The implicit baseline desires *2 replicas at 100cc* (the realized config); graceful reconfiguration desires *2 replicas at 200cc* (the record's target). The desired set is their union, `{2×100cc, 2×200cc}`. Actual is `{r1, r2 @ 100cc}`, matched to the 100cc slots by config; the two 200cc slots are unfilled, so the controller creates them as fresh replicas `r3`, `r4`. Actual is now `{r1, r2 @ 100cc, r3, r4 @ 200cc}` — old and new sets overlapping, all four serving.
+3. **Hydration.** While `r3`, `r4` hydrate, the desired set is unchanged, so the controller does nothing and the `100cc` replicas keep serving.
+4. **Cut-over — the tick's `update_state` phase.** On the first tick where `r3`, `r4` are present and hydrated, `update_state` commits `cluster.size := 200cc` and clears the `reconfiguration` record, and the controller awaits that write before continuing.
+5. **Old set falls out — the same tick's `desired_replicas` phase.** With the cut-over applied, the implicit baseline now desires *2 replicas at 200cc* — matched to `r3`, `r4` — and graceful reconfiguration, with no record, desires nothing. The desired set is `{2×200cc}`; `r1`, `r2` (at `100cc`) are desired by no strategy, so the controller drops them in the same tick. Actual settles at `{r3, r4 @ 200cc}`.
+
+Advancing `cluster.size` at cut-over is the single durable write that retires the old set: it flips the implicit baseline from holding the `100cc` replicas to holding the `200cc` ones, and the old replicas fall out of the union on their own. Because the desired set is matched to actual by config and count, the freshly-named `r3`, `r4` satisfy the new steady state directly. Nothing is renamed, and the cluster never drops below `replication_factor` serving replicas.
+
+### Durable state model
+
+The cluster's durable configuration represents the cluster's **realized, currently-serving state**, what is actually running at steady state. We add additional records in cluster state for use by the strategies:
+
+- `auto_scaling_strategy: Option<AutoScalingStrategy>` — the strategy block (v1: `ON HYDRATION` with its `HYDRATION SIZE` and optional `LINGER DURATION`). This is user-configured *policy*, distinct from the two transition records below, which are `ALTER`/controller-managed *runtime state*.
+- `reconfiguration: Option<ReconfigurationState>` — an in-flight graceful reconfiguration: the `target` (the size / replication-factor / availability-zones / logging the cluster is moving to), a `deadline`, and the `on_timeout` action (`COMMIT` or `ROLLBACK`) to apply if the deadline passes before the target hydrates.
+- `burst: Option<BurstState>` — the analogous record for an active hydration burst: the `burst_size` of the in-flight burst replica, a `linger_duration`, and the timestamp at which we observed the steady-state replicas as hydrated. Burst is controller-initiated (not tied to an `ALTER`), the strategy writes the record when we determine burst is needed. It is cleared when burst tears down on success.
+
+An `ALTER CLUSTER SET (...)` that changes a replica's **config shape** — `SIZE`, logging (`INTROSPECTION ...`), or `AVAILABILITY ZONES` — needs a hydrate-overlap, so it writes the `reconfiguration` record in a transaction and returns; the realized config is left untouched until the controller cuts over. When no reconfiguration is in flight, changes that need no overlap ( replication-factor-only, etc.) skip the record and update the realized config directly. But once a `reconfiguration` record is present, every further `ALTER` instead **folds into it**, overwriting its `target` and deadline. So the realized config is advanced only by the controller at cut-over, and no direct config write ever races an in-flight transition; re-targeting and ALTER-back are an overwrite of the record's `target`. The controller's job is to converge the actual replica set onto the target and, at cut-over, advance the realized config to match.
+
+**Scheduled clusters and `replication_factor`.** For a cluster with a non-MANUAL `schedule`, `replication_factor` is not a user-facing capacity knob: it is held at `0`, and the controller is the sole authority over the replica set. `CREATE CLUSTER` and `ALTER CLUSTER ... SET (SCHEDULE = ...)` both normalize `replication_factor` to `0` when the schedule is non-MANUAL. Otherwise a stale non-zero value would have bootstrap provision a replica the controller immediately drops, a restart-time flap. As a result `mz_clusters.replication_factor` reads `0` for a scheduled cluster even while a replica runs inside a refresh window; `mz_cluster_replicas` is authoritative for what is actually running.
+
+### SQL surface
+
+**Reshaped: `ALTER CLUSTER ... SET (...)`.** Writes the durable `reconfiguration` record when needed and kicks the controller; the realized `cluster.size` is left in place until cut-over. In the default (background) mode, the statement returns immediately and the reconfiguration runs to completion in the background. While the new mode is being rolled out behind a dyncfg gate, the foreground experience is preserved as a thin session-side wait shim over the same mechanism: the session blocks polling the new introspection view until the controller reports reconciliation complete (or hits a timeout). Notably, under this model session disconnect during the wait does *not* abort the reconfiguration. The durable target is already set and the controller carries on. Deprecating the foreground experience after rollout means removing the wait shim, not unwinding any parallel machinery.
+
+**Per-`ALTER` timeout:** The current `ALTER CLUSTER ... SET (...) WITH (WAIT ...)` syntax is retained unchanged and repurposed to drive the durable mechanism. Two forms:
+
+- `WAIT UNTIL READY (TIMEOUT '<dur>', ON TIMEOUT COMMIT|ROLLBACK)` — the hydration-aware form. `TIMEOUT` is added to the transaction timestamp and written as the `reconfiguration` record's `deadline`; `ON TIMEOUT` is written as its `on_timeout`. The controller cuts over once the target replicas hydrate; only if the deadline passes un-hydrated does it apply `on_timeout`.
+- `WAIT FOR '<dur>'` — retained as sugar for `WAIT UNTIL READY (TIMEOUT '<dur>', ON TIMEOUT COMMIT)`. Its legacy contract was "wait the fixed duration, then cut over regardless of hydration"; a change from the legacy behavior is that the controller cuts over the instant the target hydrates, so we drop the "wait the *full* duration even when ready early" property, which carries no value in a background model. The blind cut-over-at-the-deadline it implied is exactly `ON TIMEOUT COMMIT`. A `WAIT FOR '0s'` (equivalently `TIMEOUT '0s'` with `ON TIMEOUT COMMIT`) thus requests an immediate cut-over: the deadline is already past, so the controller commits on its first pass without ever provisioning overlap replicas — the non-graceful, drop-and-recreate resize, now an explicit opt-in rather than the no-`WAIT` default.
+
+`ON TIMEOUT` defaults to **`ROLLBACK`** when omitted — the safe, conservative choice that never surprises a user with downtime by cutting over to a not-yet-hydrated target. This is a change from today's implicit `COMMIT` default, applied uniformly (the foreground wait shim picks it up too). Omitting `WITH (WAIT ...)` entirely on a reconfiguring `ALTER` falls back to the `default_cluster_reconfiguration_timeout` dyncfg for the deadline (so a transition is never unbounded) and the default `ROLLBACK` action. The session-side wait shim does not run a timer of its own; it watches the durable outcome and returns when the controller resolves — a completed cut-over or a rolled-back tombstone — so the in-session UX of "ALTER returns or errors within the timeout" is preserved without the shim and controller ever disagreeing. There is no shim-side timeout at all: the controller resolves the reconfiguration by its deadline, and the shim simply waits for that.
+
+**New cluster option: `AUTO SCALING STRATEGY = (...)`.** Available at `CREATE CLUSTER` and `ALTER CLUSTER ... SET (...)`. The value is a paren-enclosed list mixing strategy specs (currently only `ON HYDRATION (HYDRATION SIZE = '...', LINGER DURATION = '...')`, where `LINGER DURATION` is optional — the dyncfg default applies when it is omitted). The whole block is set or replaced atomically; specifying an empty list — or `ALTER CLUSTER ... RESET (AUTO SCALING STRATEGY)` — disables auto-scaling for the cluster. Fine-grained ALTERs that change one sub-parameter are out of scope for v1.
+
+**Cancellation.** v1 uses ALTER-back semantics: a user cancels an in-flight reconfiguration by issuing another `ALTER CLUSTER` with the original (or any other) size. The new `ALTER` overwrites the `reconfiguration` record's target and deadline; the controller converges on the new target — and if it equals the still-realized `cluster.size` (a true cancel-back), convergence is just dropping the in-flight new replicas and clearing the record.
+
+We may add explicit `ALTER CLUSTER ... CANCEL ALTER` (or equivalent) syntax later; this is not strictly necessary for v1 because the ALTER-back path covers the same need.
+
+**Rejected combinations** (error at parse/plan):
+- `HYDRATION SIZE` equal to the cluster `SIZE` (no-op; rejected for clarity).
+- `AUTO SCALING STRATEGY` combined with `SCHEDULE = ('on-refresh', ...)` in v1.
+
+### Observability
+
+Two distinct surfaces, mirroring the controller's separation of "current state" from "history":
+
+- **Current state — introspection view.** A new builtin view (working name: `mz_internal.mz_cluster_reconfigurations`, naming TBD) reports, per cluster: realized vs. target config shape (size, logging, availability zones) and replication factor, presence of an in-flight reconfiguration, the active deadline (when set), currently-running burst replicas. The *target* is the `reconfiguration` record's target whenever a `reconfiguration` record is present (including a timed-out tombstone), otherwise `cluster.size` itself. During a graceful reconfiguration the two differ; at cut-over `cluster.size` advances to the target and they coincide again. A timed-out reconfiguration shows as a tombstoned record (deadline in the past) with `cluster.size` still at the realized value, and the abandoned target visible in the record.
+
+- **`SHOW CLUSTERS` surfaces in-flight reconfigurations.** Today's `SHOW CLUSTERS` is implemented as SQL over `mz_clusters` and related views; we extend that SQL to join with the new introspection view so a user sees, per cluster, both the current size (the realized `cluster.size`) and the target size (the `reconfiguration` record's target, when one is in flight), plus an indication of whether a reconfiguration is in flight. The exact column set is left to implementation, but the design requires that a single `SHOW CLUSTERS` answers the questions "what did I ask for", "what's actually there now", and "is something in progress, done, or timed out" without requiring the user to know about the introspection view.
+
+- **History — audit log.** Audit events come from three places: the `ALTER` that writes or overwrites a `reconfiguration` record (a graceful reconfiguration *started*, or *cancelled* via ALTER-back), the durable transitions `update_state` commits within a tick (cut-over, timeout-fired, and burst *started* — burst is controller-initiated, so `update_state` writes its record), and the create/drop commands the controller derives from the `desired_replicas` diff. `desired_replicas` itself records nothing directly — it touches the log only through the replica-lifecycle commands its output induces. Extend the existing `SchedulingDecisionsWithReasonsV2` event family (or add closely-related event variants) to cover:
+  - Graceful reconfiguration: started, replica created, replica dropped, finalized, cancelled, timeout fired (for stuck reconfigurations).
+  - Hydration burst: started, replica created, replica dropped.
+
+All audit entries record the cluster, and identifiers of the replicas involved. Reconfiguration start and timeout-fired events also record the active deadline so an operator can correlate "timeout fired" with the originating `ALTER`.
+
+### Failure handling and safety rails
+
+- **Stuck reconfiguration.** The `reconfiguration` record's `deadline` and `on_timeout` come from the `ALTER`'s `WITH (WAIT ...)` (or the `default_cluster_reconfiguration_timeout` dyncfg + default `ROLLBACK` when no `WAIT` is given). *Success precedence* always applies first: if the target replicas all hydrate at or before the deadline, the controller cuts over normally regardless of `on_timeout`. When `now` exceeds the deadline with the target **not** fully hydrated, the controller applies `on_timeout`:
+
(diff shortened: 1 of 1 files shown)


doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,212 @@
+# Cluster Autoscaling and Background Reconfiguration
+
+- Associated: (TBD — link epics/issues)
+
+## The Problem
+
+Two user-facing capabilities motivate this work:
+
+1. **Background graceful cluster reconfiguration.** Today, `ALTER CLUSTER ... SET (SIZE = ...)` with the graceful (zero-downtime) strategy requires the SQL session to remain open for the duration of the reconfiguration — the session holds the wait-for-hydration stage. Long-running reconfigurations are fragile: any process or session interruption — a network blip, a client timeout, an SQL tool closing, an `environmentd` restart — aborts the reconfiguration. The user experience we want is: the statement returns immediately, and the reconfiguration continues in the background, surviving restarts and disconnects.
```
maheshwarip, line 9: yaay!

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,212 @@
...
+- A user can issue `ALTER CLUSTER ... SET (SIZE = ...)` and the statement returns immediately. The reconfiguration continues in the background, surviving session disconnects and `environmentd` restarts, and is observable via SQL.
+- A user can specify `HYDRATION_SIZE` at cluster create or alter. The burst replica is spun up when warranted and shut down when no longer needed, with no further user action.
+- The existing `SCHEDULE = ('on-refresh', ...)` behavior is preserved, ideally as one strategy within a common autoscaling framework rather than as a parallel mechanism.
+- All autoscaling decisions and replica lifecycle changes are recorded with reasons and surfaced to the user.
+- A new autoscaling strategy can be added without restructuring the framework.
+- Operators can disable the burst behavior across an environment via a break-glass flag without disabling other autoscaling.
+
+## Out of Scope
+
+- `HYDRATION_SIZE` combined with `SCHEDULE = ('on-refresh', ...)`. Initial version rejects; see [Open Questions](#open-questions).
```
maheshwarip: agreed

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,212 @@
...
+- More than one concurrent burst replica per cluster. Initial version supports exactly one; revisit if needed.
```
maheshwarip: also agreed

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,212 @@
...
+## Success Criteria
+
+- A user can issue `ALTER CLUSTER ... SET (SIZE = ...)` and the statement returns immediately. The reconfiguration continues in the background, surviving session disconnects and `environmentd` restarts, and is observable via SQL.
+- A user can specify `HYDRATION_SIZE` at cluster create or alter. The burst replica is spun up when warranted and shut down when no longer needed, with no further user action.
+- The existing `SCHEDULE = ('on-refresh', ...)` behavior is preserved, ideally as one strategy within a common autoscaling framework rather than as a parallel mechanism.
+- All autoscaling decisions and replica lifecycle changes are recorded with reasons and surfaced to the user.
+- A new autoscaling strategy can be added without restructuring the framework.
+- Operators can disable the burst behavior across an environment via a break-glass flag without disabling other autoscaling.
+
+## Out of Scope
```
maheshwarip, line 48: I agree with the calls here

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,212 @@
...
+
+- **Cluster controller as a separate process.** Deferred. v1 places it as a task inside `environmentd`. The message-channel boundary between controller and coordinator is clean enough that extraction to a separate process later is a localized refactor, not a redesign.
+
+## Open Questions
+
+1. **Stuck-reconfiguration recovery policy.** When a reconfiguration's pending replicas have not hydrated within the system timeout, do we (a) park the reconfiguration indefinitely with a clear signal in the introspection view for an operator to act, (b) auto-cancel and revert to the prior steady state, or (c) make the policy a dyncfg with one of (a)/(b) as the default? Same question for stuck burst replicas.
+
+2. **`HYDRATION_SIZE` + `SCHEDULE = ('on-refresh', ...)` combination.** v1 rejects this combination. Semantically the combination is interesting (every refresh window, burst comes up first to accelerate hydration, steady catches up, then the schedule turns the cluster off), but our strong recommendation is to keep this rejected indefinitely: there is currently no appetite to invest further in the `SCHEDULE` syntax, and supporting the combination would expand its surface area.
```
maheshwarip: agreed

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,212 @@
...
+
+3. **Multiple burst replicas (one per steady replica vs. one total).** v1 supports exactly one burst replica per cluster regardless of replication factor. Our strong recommendation is to keep it that way: the burst replica is by design transient, and provisioning N burst replicas to mirror an N-replica steady set would multiply cost for diminishing benefit — burst tear-down only requires one steady replica to have caught up. Revisit only if real-world usage proves the single-burst model insufficient.
```
maheshwarip: agreed

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,212 @@
...
+
+7. **Foreground/synchronous graceful reconfiguration retention.** Our strong recommendation is to deprecate the current foreground (session-bound) mechanism in favor of the background model. During rollout, the foreground experience is preserved as a thin session-side wait shim over the background mechanism (see [SQL surface](#sql-surface)) — *not* by retaining the existing parallel state machine. This means the `pending: bool` flag on replicas and the associated three-stage machinery can be removed up front; deprecating the foreground experience later is simply deleting the wait shim. The one behavioral difference vs. today is that session disconnect during the wait no longer aborts the reconfiguration (arguably a feature; the durable target stays set and the controller continues).
+
+8. **Hydration burst during graceful reconfiguration.** Should burst kick in while a graceful reconfig is in flight (target size differs from current replicas)? Leaning toward no: the new-size replicas are themselves transient hydration capacity, and stacking burst on top risks confusing billing and behavior. Burst resumes once the reconfig settles.
```
maheshwarip: Nope. If a user type ALTER CLUSTER SET SIZE 200cc, that shouldn't trigger a burst. It should trigger a 200cc replica. Once the 200cc replica is hydrated, retire the original replica.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,212 @@
...
+
+The following behaviors fall out of the design rather than being its headline outcomes. They are user-observable and worth flagging in release notes and user-facing documentation.
+
+- **Burst and reconfiguration-transient replicas appear in billing and metering identically to ordinary replicas.** A user with `HYDRATION_SIZE` set sees additional billing during hydration windows; a user issuing a background `ALTER CLUSTER` sees additional billing during the overlap between the old and new replica sets.
+- **Background `ALTER CLUSTER` returns immediately** after writing the new target to the catalog. The actual replica transition happens asynchronously and is observable via the new introspection view. This matches the existing pattern for other async DDL (e.g., `CREATE INDEX` returns once the catalog entry exists; hydration happens afterwards).
+- **`SHOW CLUSTERS` reports the new (target) size immediately on ALTER**, not the old size. Mid-reconfiguration the durable cluster configuration already reflects the user's intent, so `SHOW CLUSTERS` does too. This is a change from today's behavior, where the old size is reported until the graceful reconfiguration finalizes.
```
maheshwarip: Hmm. What I would really like is for SHOW CLUSTERS to tell me whether a reconfiguration is in flight or not, tell me what the current size is, and what the target size is

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,212 @@
...
+### Migration
+
+- **Add `hydration_size: Option<String>`** to the durable cluster configuration. Standard catalog migration with `None` default for existing clusters.
+- **Remove `pending: bool`** from the durable replica location. Forward-only migration: drop the field from the proto/Rust types and stop reading it. Any pending replicas existing at the moment of upgrade are picked up by the new controller, which converges them via the diff (a pending-flagged replica at the new size is functionally identical to "a replica the controller is tracking as part of a reconfiguration").
+
+## Notable user-facing changes
+
+The following behaviors fall out of the design rather than being its headline outcomes. They are user-observable and worth flagging in release notes and user-facing documentation.
+
+- **Burst and reconfiguration-transient replicas appear in billing and metering identically to ordinary replicas.** A user with `HYDRATION_SIZE` set sees additional billing during hydration windows; a user issuing a background `ALTER CLUSTER` sees additional billing during the overlap between the old and new replica sets.
```
maheshwarip: Hmm this makes sense. But is this new behavior? I assumed that this was how it always worked!

maheshwarip: Maybe I was wrong!

aljoscha: Nah, it is how it worked, but put it in there because the bursting is new. For graceful reconfig it was always like this

doc/developer/design/20260522_cluster_autoscaling.md
aljoscha: So the nope is a confirmation of my "leaning towards no", yes? 😅

doc/developer/design/20260522_cluster_autoscaling.md
aljoscha: Yeah, I can see why you would like that. I think we can do something good there!

doc/developer/design/20260522_cluster_autoscaling.md
maheshwarip: Gotcha gotcha. Ok, no concerns!

doc/developer/design/20260522_cluster_autoscaling.md
maheshwarip: Correct :)

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -145,6 +145,8 @@ Two distinct surfaces, mirroring the controller's separation of "current state"
 
 - **Current state — introspection view.** A new builtin view (working name: `mz_internal.mz_cluster_reconfigurations`, naming TBD) reports, per cluster: desired vs. actual size and replication factor, presence of an in-flight reconfiguration, currently-running burst replicas, per-replica hydration progress, and the strategies' current decisions and reasons. Pull-only: users `SELECT` to learn status. No push notifications in v1.
 
+- **`SHOW CLUSTERS` surfaces in-flight reconfigurations.** Today's `SHOW CLUSTERS` is implemented as SQL over `mz_clusters` and related views; we extend that SQL to join with the new introspection view so a user sees, per cluster, both the target size (from the durable cluster config) and the current size (derived from the actual replica set), plus an indication of whether a reconfiguration is in flight. The exact column set is left to implementation, but the design requires that a single `SHOW CLUSTERS` answers the questions "what did I ask for", "what's actually there now", and "is something in progress" without requiring the user to know about the introspection view.
```
aljoscha: @maheshwarip what do you think?

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+     AUTO SCALING STRATEGY = (
+       ON HYDRATION (
+         HYDRATION_SIZE = '3200cc'         -- burst replica size while objects are not hydrated
+       ),
+       COOL DOWN = '10m'                   -- shared across all strategies in the block
+     )
+   );
+   ```
+
+   When the cluster has any object that is not fully hydrated, the cluster controller spins up an additional replica at the configured `HYDRATION_SIZE` to accelerate hydration. Once any steady-state replica has fully hydrated all currently-existing objects on the cluster, the burst replica is shut down. The block's `COOL DOWN` bounds how often the controller may take consecutive scaling actions, providing built-in hysteresis.
```
ggevay: Is it enough to wait for hydration, or do we want something like the 0dt cutover's caught-up check? Specifically:
- We might want to wait for frontiers to catch up (either to the wall clock, or to the hydration replica).
- But we might not need to bother with looking for OOM loops that happen on both replicas and ignoring them, since this feature is intended for functioning clusters. That is, any OOM loop should prevent the cutover, unlike with the 0dt cutover.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+- **Cluster configuration is fully durable**, including `ClusterVariantManaged { size, replication_factor, availability_zones, schedule, logging, optimizer_feature_overrides }`.
+
+## Success Criteria
+
+- A user can issue `ALTER CLUSTER ... SET (SIZE = ...)` and the statement returns immediately. The reconfiguration continues in the background, surviving session disconnects and `environmentd` restarts, and is observable via SQL.
+- A user can specify an `AUTO SCALING STRATEGY` (with `ON HYDRATION` and its `HYDRATION_SIZE`) at cluster create or alter. The burst replica is spun up when warranted and shut down when no longer needed, with no further user action.
+- The existing `SCHEDULE = ('on-refresh', ...)` behavior is preserved, ideally as one strategy within a common autoscaling framework rather than as a parallel mechanism.
+- All autoscaling decisions and replica lifecycle changes are recorded with reasons and surfaced to the user.
+- A new autoscaling strategy can be added without restructuring the framework.
+- Operators can disable the burst behavior across an environment via a break-glass flag without disabling other autoscaling.
```
ggevay: What does this mean? I mean, if it's "across an environment", then what "other autoscaling" is there?

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+
+- **Cluster controller as a separate process.** Deferred. v1 places it as a task inside `environmentd`. The message-channel boundary between controller and coordinator is clean enough that extraction to a separate process later is a localized refactor, not a redesign.
+
+## Open Questions
+
+1. **Stuck-reconfiguration recovery policy.** When a reconfiguration's pending replicas have not hydrated within the system timeout, do we (a) park the reconfiguration indefinitely with a clear signal in the introspection view for an operator to act, (b) auto-cancel and revert to the prior steady state, or (c) make the policy a dyncfg with one of (a)/(b) as the default? Same question for stuck burst replicas.
+
+2. **`AUTO SCALING STRATEGY` + `SCHEDULE = ('on-refresh', ...)` combination.** v1 rejects this combination. Semantically the combination is interesting (every refresh window, burst comes up first to accelerate hydration, steady catches up, then the schedule turns the cluster off), but our strong recommendation is to keep this rejected indefinitely: there is currently no appetite to invest further in the `SCHEDULE` syntax, and supporting the combination would expand its surface area.
```
ggevay: I agree, no need to support this combination!

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+- The existing `SCHEDULE = ('on-refresh', ...)` behavior is preserved, ideally as one strategy within a common autoscaling framework rather than as a parallel mechanism.
+- All autoscaling decisions and replica lifecycle changes are recorded with reasons and surfaced to the user.
+- A new autoscaling strategy can be added without restructuring the framework.
+- Operators can disable the burst behavior across an environment via a break-glass flag without disabling other autoscaling.
+
+## Out of Scope
+
+- `AUTO SCALING STRATEGY` combined with `SCHEDULE = ('on-refresh', ...)`. Initial version rejects; see [Open Questions](#open-questions).
+- More than one concurrent burst replica per cluster. Initial version supports exactly one; revisit if needed.
+- Per-`ALTER` timeout configuration. The current session-bound `ALTER ... WAIT UNTIL READY (TIMEOUT = ...)` semantics do not carry over directly to background reconfiguration. A system-wide timeout (via dyncfg) is used instead in v1.
```
ggevay: Which semantics don't carry over? Cancelling the reconfiguration by cancelling the command certainly doesn't carry over, but the `TIMEOUT` and the `ON TIMEOUT` settings could carry over, right?

Or is it better to make these global so that the user can influence an in-progress reconfiguration? That is, instead of having to give IDs to reconfigurations and make a new SQL command for resolving (cancelling or rolling forward) an in-progress reconfiguration, one can simply alter the global timeout settings.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+1. **Background graceful cluster reconfiguration.** Today, `ALTER CLUSTER ... SET (SIZE = ...)` with the graceful (zero-downtime) strategy requires the SQL session to remain open for the duration of the reconfiguration — the session holds the wait-for-hydration stage. Long-running reconfigurations are fragile: any process or session interruption — a network blip, a client timeout, an SQL tool closing, an `environmentd` restart — aborts the reconfiguration. The user experience we want is: the statement returns immediately, and the reconfiguration continues in the background, surviving restarts and disconnects.
+
+2. **Hydration burst (autoscaling for hydration).** Users want to express, at cluster creation:
+
+   ```sql
+   CREATE CLUSTER my_cluster (
+     SIZE = '100cc',                       -- steady-state replica size
+     AUTO SCALING STRATEGY = (
+       ON HYDRATION (
+         HYDRATION_SIZE = '3200cc'         -- burst replica size while objects are not hydrated
```
ggevay: nit: For consistency with other SQL syntax, we shouldn't have that underscore in `HYDRATION_SIZE`. Our [SQL design principles](https://www.notion.so/materialize/SQL-design-principles-2e2069797f7c46b8b4c7dd10043afdcd?source=copy_link#9d4a155e751f4cbeab7988bfb338b3f6) also mentions this.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+- All autoscaling decisions and replica lifecycle changes are recorded with reasons and surfaced to the user.
+- A new autoscaling strategy can be added without restructuring the framework.
+- Operators can disable the burst behavior across an environment via a break-glass flag without disabling other autoscaling.
+
+## Out of Scope
+
+- `AUTO SCALING STRATEGY` combined with `SCHEDULE = ('on-refresh', ...)`. Initial version rejects; see [Open Questions](#open-questions).
+- More than one concurrent burst replica per cluster. Initial version supports exactly one; revisit if needed.
+- Per-`ALTER` timeout configuration. The current session-bound `ALTER ... WAIT UNTIL READY (TIMEOUT = ...)` semantics do not carry over directly to background reconfiguration. A system-wide timeout (via dyncfg) is used instead in v1.
+- Push notifications for reconfiguration completion. Pull-only via introspection in v1.
```
ggevay: But a user could `SUBSCRIBE` to the introspection, right?

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+- Cross-cluster scaling decisions (e.g., scaling cluster A based on cluster B's load).
+
+## Solution Proposal
+
+### Summary
+
+We introduce a **cluster controller** as a dedicated task inside `environmentd`. It is the single decision-maker for the replica set of every cluster. It operates as a reconciler: it reads desired cluster state from the durable catalog, observes the actual replica set and live status signals, computes a desired replica set per cluster by combining a set of **strategies**, and emits catalog-change commands to the Coordinator via a message channel. The Coordinator remains the sole writer of catalog state.
```
ggevay, line 58: The term "cluster controller" seems like a natural choice, but I'm wondering if people would get confused all the time with our existing use of the word "controller". Edit: See e.g. my next comment already :D

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+
+ALTER CLUSTER operations are reshaped so that the user's intent is written durably and immediately to the catalog; the controller then reconciles from that intent. Graceful reconfiguration becomes a strategy within this framework, as does hydration burst. The existing `ON REFRESH` scheduling is lifted into the framework.
+
+### The cluster controller
+
+The controller runs as its own task within `environmentd`. Its inputs are:
+
+- **Durable cluster configuration**, observed via the catalog (delivered as updates or snapshotted).
+- **Live status signals** from the compute and storage controllers: replica readiness, hydration state, replica failures, current frontiers. Hydration is consumed directly from the controller's in-process state.
```
ggevay: In theory, both the "Durable cluster configuration" and the "Live status signals" could be obtained also through subscribing to builtin relations. That setup could be more future-proof for when we have multiple envd processes, where not all of them have controller (It has begun... I mean, the old sense of "controller") state for all clusters.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+- **A periodic tick** for time-based strategies.
+
+Its outputs are **decisions** sent as messages to the Coordinator's existing internal command channel. Each decision is a request to create, drop, or modify a replica, accompanied by a reason. The Coordinator transacts them through the catalog and orchestrator as it does today for `Message::SchedulingDecisions`.
+
+The controller is not the catalog writer. Coordinator remains the only writer. The controller's commands must be idempotent (replayable on retry without harm), because crashes and re-elections may cause the controller to recompute and re-emit a decision that was already partially applied.
```
ggevay: Could you write a bit more on how to make these idempotent? For example, how will we make a replica creation idempotent? I can imagine either
- making the command a compare-and-swap, e.g., a command could be: "if exactly these replicas exist for the cluster, then create this replica"
- or adding some reconciliation-like logic, e.g. a command could be: "I'd like at least `r` number of replicas with size `s` for this cluster".

ggevay (commented):
Releasing some comments, will continue after lunch.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+
+### The cluster controller
+
+The controller runs as its own task within `environmentd`. Its inputs are:
+
+- **Durable cluster configuration**, observed via the catalog (delivered as updates or snapshotted).
+- **Live status signals** from the compute and storage controllers: replica readiness, hydration state, replica failures, current frontiers. Hydration is consumed directly from the controller's in-process state.
+- **A periodic tick** for time-based strategies.
+
+Its outputs are **decisions** sent as messages to the Coordinator's existing internal command channel. Each decision is a request to create, drop, or modify a replica, accompanied by a reason. The Coordinator transacts them through the catalog and orchestrator as it does today for `Message::SchedulingDecisions`.
```
ggevay: > Each decision is a request to create, drop, or modify a replica, accompanied by a reason

Plus it can also happen that a decision is to modify a cluster's configuration, not just its replicas, right? For example, for `ON TIMEOUT = 'ROLLBACK'`, it would observe if a reconfiguration replica is taking too long to hydrate and drop not just the replica, but also roll back the cluster's desired size.

ggevay:
(I think the "[Rendered](https://github.com/aljoscha/materialize/blob/1e2fd58686968da8c62e041c875843c779078462/doc/developer/design/20260522_cluster_autoscaling.md)" link in the PR description is pointing to a slightly outdated version.)

aljoscha:
> (I think the "[Rendered](https://github.com/aljoscha/materialize/blob/1e2fd58686968da8c62e041c875843c779078462/doc/developer/design/20260522_cluster_autoscaling.md)" link in the PR description is pointing to a slightly outdated version.)

fixed!

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+- **Crash-safe by construction.** Any state needed to continue an operation across a restart is, by definition, in the catalog or derivable from it.
+- **Composable.** Multiple strategies (graceful reconfig, hydration burst, on-refresh, future ones) each contribute to the desired set; the controller merges them with explicit precedence. No state-machine interactions between strategies.
+- **Deterministically auditable.** Same inputs always yield the same decisions; audit entries explain each decision from durable state plus observed status.
+
+### Durable state model
+
+The cluster's durable configuration represents the user's **desired target state**, not in-flight mechanics:
+
+- `size`, `replication_factor`, `availability_zones`, `schedule`, `logging`, `optimizer_feature_overrides` — preserved.
+- `auto_scaling_strategy: Option<AutoScalingStrategy>` — new optional field carrying the strategy block (currently only `ON HYDRATION` with its `HYDRATION_SIZE`, plus the shared `COOL DOWN`). The exact in-catalog shape is left to the implementer; the design only requires that it durably captures everything the user expressed in the SQL block.
```
ggevay: So, a strategy can be either explicitly recorded, like the "Hydration burst" strategy, or it can be implicit, like a graceful reconfiguration? That is, would a graceful reconfiguration be visible only from the fact that the replica set doesn't match the cluster's desired state?

(Also, does the "currently only" refer to some prototype implementation, which hasn't yet lifted the REFRESH stuff?)

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+
+The actual replica set lives in the catalog as before, with each replica's durable size, allocation, AZ, etc. The **`pending: bool` flag is removed.** Its existing roles fall away in the new model:
+
+- It was used to identify "in-progress reconfiguration" replicas — derivable from the diff between cluster config and actual replicas.
+- It was used by `cluster_scheduling.rs` to skip clusters mid-reconfig — that scheduler folds into the controller, which has full visibility into all strategy intents.
+- It gated bootstrap cleanup of orphaned pending replicas. With everything durable and the reconciler running, no bootstrap cleanup is needed: the controller's first reconcile-pass on startup naturally converges actual to desired.
+
+### Strategies
+
+A **strategy** is a function of the form `(durable cluster config, observed status, current time) → partial intent`. Each strategy returns, for a cluster, the contribution it wants to make to the cluster's desired replica set: replicas to add, replicas to keep, replicas to drop, and the conditions that govern transitions (e.g., "drop replica R once replica R' has hydrated all current objects").
```
ggevay: > to the cluster's desired replica set

(And maybe also to the cluster's overall config, as noted elsewhere for the case when we are rolling back a graceful cluster reconfiguration.)

doc/developer/design/20260522_cluster_autoscaling.md
ggevay: I'm not sure it's good to make graceful reconfiguration implicit, as it would be an asymmetry among strategies. Also, it might make it complex to see at a glance whether a graceful reconfiguration is in progress. For example, some other strategies might exclude a graceful reconfiguration, but some might not:
- `REFRESH` excludes it (as per the design decision above), so if there is a `REFRESH` strategy, then not having enough replicas does _not_ mean that there is a graceful reconfiguration in progress.
- I guess the "Hydration burst" strategy doesn't necessarily have to exclude the graceful reconfiguration.

Edit: Although, the latter concern could be addressed with adding good observability, e.g., with the proposed smartening of `SHOW CLUSTERS`. But I'd say that asymmetry argument still stands.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+## Success Criteria
+
+- A user can issue `ALTER CLUSTER ... SET (SIZE = ...)` and the statement returns immediately. The reconfiguration continues in the background, surviving session disconnects and `environmentd` restarts, and is observable via SQL.
+- A user can specify an `AUTO SCALING STRATEGY` (with `ON HYDRATION` and its `HYDRATION_SIZE`) at cluster create or alter. The burst replica is spun up when warranted and shut down when no longer needed, with no further user action.
+- The existing `SCHEDULE = ('on-refresh', ...)` behavior is preserved, ideally as one strategy within a common autoscaling framework rather than as a parallel mechanism.
+- All autoscaling decisions and replica lifecycle changes are recorded with reasons and surfaced to the user.
+- A new autoscaling strategy can be added without restructuring the framework.
+- Operators can disable the burst behavior across an environment via a break-glass flag without disabling other autoscaling.
+
+## Out of Scope
```
ggevay, line 48: Are multi-replica (in steady state) clusters in scope? I think the current state of the design document doesn't consider them. For example, in that case a hydration burst might want to wait for multiple replicas to be hydrated (or caught up) rather than any (non-burst) replica being hydrated (or caught up).

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+
+3. **Multiple burst replicas (one per steady replica vs. one total).** v1 supports exactly one burst replica per cluster regardless of replication factor. Our strong recommendation is to keep it that way: the burst replica is by design transient, and provisioning N burst replicas to mirror an N-replica steady set would multiply cost for diminishing benefit — burst tear-down only requires one steady replica to have caught up. Revisit only if real-world usage proves the single-burst model insufficient.
+
+4. **Cancellation syntax.** v1 uses ALTER-back. Do we want explicit `ALTER CLUSTER ... CANCEL ALTER` syntax for clarity?
+
+5. **Strategy roadmap.** No concrete next strategies are committed. Our strong recommendation is to keep the strategy interface minimal — sufficient for graceful reconfiguration, hydration burst, and `ON REFRESH` — and resist extending it preemptively. Capability gets added when concrete needs arrive.
```
ggevay: There have been requests for automated cluster pause/resume for reasons other than REFRESH MVs, e.g., automatically spinning down a cluster when nobody is using it, and spinning it up when somebody queries it. See: https://github.com/MaterializeInc/database-issues/issues/6966

And another strategy could be automatically scaling up/down based on load: https://github.com/MaterializeInc/database-issues/issues/6821

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+3. **Hydration burst** is additive on top of the gated steady set: if the steady set is non-empty and at least one steady object isn't hydrated, add one burst replica.
+
+This ordering ensures determinism and avoids strategy fights. If a future strategy needs different precedence, it is introduced into this ordering explicitly in a follow-up.
+
+### SQL surface
+
+**Reshaped: `ALTER CLUSTER ... SET (...)`.** Always writes the new durable target and kicks the controller. In the default (background) mode, the statement returns immediately and the reconfiguration runs to completion in the background. While the new mode is being rolled out behind a dyncfg gate, the foreground experience is preserved as a thin session-side wait shim over the same mechanism: the session blocks polling the new introspection view until the controller reports reconciliation complete (or hits a timeout). Notably, under this model session disconnect during the wait does *not* abort the reconfiguration — the durable target is already set and the controller carries on. Deprecating the foreground experience after rollout means removing the wait shim, not unwinding any parallel machinery.
```
ggevay: > polling

(Or subscribing I guess)

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+
+**New cluster option: `AUTO SCALING STRATEGY = (...)`.** Available at `CREATE CLUSTER` and `ALTER CLUSTER ... SET (...)`. The value is a paren-enclosed list mixing strategy specs (currently only `ON HYDRATION (HYDRATION_SIZE = '...')`) with shared parameters (`COOL DOWN = '...'`). The whole block is set or replaced atomically; omitting it on ALTER (or specifying an empty list) disables auto-scaling for the cluster. Fine-grained ALTERs that change one sub-parameter are out of scope for v1. `COOL DOWN` is itself optional; if omitted, a system-wide default (via dyncfg) applies.
```
ggevay: > Fine-grained ALTERs that change one sub-parameter are out of scope for v1.

That might be a bit confusing for users, because other cluster options don't fall off when a different cluster option is ALTERed. I'm not sure how much this behavior would simplify the implementation, but if not much, then I think it would be better to be consistent with how the other cluster options behave.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+
+**Cancellation.** v1 uses ALTER-back semantics: a user cancels an in-flight reconfiguration by issuing another `ALTER CLUSTER` with the original (or any other) size. The controller observes the new target and converges. No "remembered original size" is needed because the durable `cluster.size` is the desired state; the user issues the size they want.
```
ggevay: > No "remembered original size" is needed

Well, `ON TIMEOUT = 'ROLLBACK'` would need to remember, as mentioned elsewhere. But I guess it could see it by looking at the set of replicas and seeing a replica that still has the original size.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+
+**Rejected combinations** (error at parse/plan):
+- `HYDRATION_SIZE` equal to the cluster `SIZE` (no-op; rejected for clarity).
+- `AUTO SCALING STRATEGY` combined with `SCHEDULE = ('on-refresh', ...)` in v1. See [Open Questions](#open-questions).
+
+### Observability
+
+Two distinct surfaces, mirroring the controller's separation of "current state" from "history":
+
+- **Current state — introspection view.** A new builtin view (working name: `mz_internal.mz_cluster_reconfigurations`, naming TBD) reports, per cluster: desired vs. actual size and replication factor, presence of an in-flight reconfiguration, currently-running burst replicas, per-replica hydration progress, and the strategies' current decisions and reasons. Pull-only: users `SELECT` to learn status. No push notifications in v1.
```
ggevay: > Pull-only

(Can be also `SUBSCRIBE`d, right?)

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+
+- **Keep `pending: bool` on replicas.** Considered. Removed because, in the reconciler model, it carries no information not already present in the diff between cluster config and actual replicas, and its presence implies semantics (lifecycle marker) that the controller already owns. The implementation cost of removing it (verifying no other consumers, removing bootstrap cleanup) is one-time; carrying it forward indefinitely is ongoing complexity.
+
+- **Cluster controller as a separate process.** Deferred. v1 places it as a task inside `environmentd`. The message-channel boundary between controller and coordinator is clean enough that extraction to a separate process later is a localized refactor, not a redesign.
+
+## Open Questions
```
ggevay, line 243: Btw. how would the strategies interact with blue-green deploys? E.g., would the blue-green cutover happen already when a burst replica hydrates? I guess yes, if we don't add explicit handling of burst replicas to the blue-green code (the dbt stuff). But this might be undesirable: A user might want the blue-green cutover only when the steady-state replica is stable.

doc/developer/design/20260522_cluster_autoscaling.md
ggevay: > is a function of the form `(durable cluster config, observed status, current time) → partial intent

How does the implementation of Hysteresis know that something happened recently? Will it look at something like `mz_cluster_replica_history` and try to guess what happened? Or will it look at the audit log to not have to guess what strategy fired recently? If the latter, than we should add the audit log to the input of strategies.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+
+- **Hysteresis.** The user-facing `COOL DOWN` parameter on the `AUTO SCALING STRATEGY` block is the primary hysteresis knob: it bounds how often the controller may take consecutive scaling actions on a cluster. If a cluster's `AUTO SCALING STRATEGY` omits `COOL DOWN`, a system-wide default (via dyncfg) applies. The hydration-burst trigger is intentionally high-thrash in v1 (any non-hydrated object causes burst, including newly added objects on an already-running cluster); `COOL DOWN` is what prevents rapid up/down churn in pathological cases.
+
+- **Break-glass dyncfg.** A dedicated flag disables the hydration-burst strategy environment-wide without disabling graceful reconfiguration or `ON REFRESH`. Operations can flip this if burst behavior misfires.
+
+- **Feature gate.** The full feature ships behind a dyncfg for staged rollout. Both `AUTO SCALING STRATEGY` SQL acceptance and the background reconfiguration behavior are gated.
```
ggevay: (It might have to be in `feature_flags!` instead of a `dyncfg`, because we need `enable_for_item_parsing`.)

doc/developer/design/20260522_cluster_autoscaling.md
ggevay, line 243: How does all this interact with 0dt upgrades when we have two non-read-only envs in the future? The catalog is shared between the old and new envs, so a strategy making a decision affects both envs. Would this mean that they'd need to look at the states of both envs before making a decision? (I think you also mentioned this in one of our 1:1s.)

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+- No bootstrap cleanup of "pending" replicas is needed (there are no pending-flagged replicas; the diff drives convergence).
+
+### Migration
+
+- **Add `auto_scaling_strategy: Option<AutoScalingStrategy>`** to the durable cluster configuration, durably capturing whatever the user expressed in the `AUTO SCALING STRATEGY` SQL block (for v1: optional `ON HYDRATION { hydration_size }`, optional `COOL DOWN`). Standard catalog migration with `None` default for existing clusters.
+- **Remove `pending: bool`** from the durable replica location. Forward-only migration: drop the field from the proto/Rust types and stop reading it. Any pending replicas existing at the moment of upgrade are picked up by the new controller, which converges them via the diff (a pending-flagged replica at the new size is functionally identical to "a replica the controller is tracking as part of a reconfiguration").
```
ggevay: (This makes it sound like we'll continue in-progress old-style reconfigurations at the moment when we migrate to the new code, but I think we'll actually roll them back, because the durably recorded size will still be the old size, because the old-style reconfiguration doesn't durably record the new size.)

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Notable user-facing changes
+
+The following behaviors fall out of the design rather than being its headline outcomes. They are user-observable and worth flagging in release notes and user-facing documentation.
+
+- **Burst and reconfiguration-transient replicas appear in billing and metering identically to ordinary replicas.** A user with an `AUTO SCALING STRATEGY` set sees additional billing during hydration windows (at the configured `HYDRATION_SIZE`); a user issuing a background `ALTER CLUSTER` sees additional billing during the overlap between the old and new replica sets.
```
ggevay: Isn't this already the case with the old-style graceful cluster reconfiguration? As far as I know, the billing pipeline operates based on the set of replicas that actually exist, which would include overlapping replicas during old-style graceful cluster reconfigurations.

doc/developer/design/20260522_cluster_autoscaling.md
ggevay: One more issue if we make the graceful reconfiguration settings global is that if a user has development clusters and prod clusters in the same env, then they would probably need different reconfiguration settings. I guess during development, they would often not need graceful reconfigurations, and might even actively don't want it, e.g. to save costs (or just for simplicity).

doc/developer/design/20260522_cluster_autoscaling.md
aljoscha: Yeah, I think the name's just to useful and I don't like the alternatives I have in the last section 🙈

doc/developer/design/20260522_cluster_autoscaling.md
aljoscha: adding a section about that, but would keep as future work

doc/developer/design/20260522_cluster_autoscaling.md
aljoscha: Yeah, this keeps being a problem, and I decided to largely punt it so we can make progress on this. I think there is a way where we either ignore the other env, or state is in the catalog and everyone needs to consider what the others are doing/whether they are ready. 🙈

.bk.yaml
mtabebe, line 1: Why is this here?

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,233 @@
...
+- The existing `SCHEDULE = ('on-refresh', ...)` behavior is preserved, ideally as one strategy within a common autoscaling framework rather than as a parallel mechanism.
+- All autoscaling decisions and replica lifecycle changes are recorded with reasons and surfaced to the user.
+- A new autoscaling strategy can be added without restructuring the framework.
+- Operators can disable the hydration-burst strategy across an environment via a break-glass flag, leaving graceful reconfiguration and `ON REFRESH` scheduling (the other strategies in the framework) untouched.
+
+## Out of Scope
+
+- `AUTO SCALING STRATEGY` combined with `SCHEDULE = ('on-refresh', ...)`. Initial version rejects; see [Open Questions](#open-questions).
+- More than one concurrent burst replica per cluster. Initial version supports exactly one; revisit if needed.
+- Dedicated push channel for reconfiguration completion (e.g., NOTICE-style). The introspection view is `SUBSCRIBE`-able, which is enough for clients that want push semantics; we're not building a separate notification surface in v1.
```
mtabebe: I think a nice variant (particularly for testing) is a version that explicitly IS blocking. This could just be syntactic sugar around the two building blocks  but I think it could be useful

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,233 @@
...
+#### Why a separate task rather than running on the coordinator's main loop
+
+The coordinator main loop is a serial command processor and a place we want to do less work in over time, not more. Periodic decision-making, polling of status signals, and computation over potentially many clusters should not block command handling. A separate task with a clean message-channel boundary also makes the controller independently testable and gives us optionality to extract it to a separate process later without further refactoring.
+
+#### Why a reconciler rather than an imperative pipeline
+
+A reconciler computes "desired replica set" purely from durable state plus observable status. This has several properties we want:
+
+- **Crash-safe by construction.** Any state needed to continue an operation across a restart is, by definition, in the catalog or derivable from it.
+- **Composable.** Multiple strategies (graceful reconfig, hydration burst, on-refresh, future ones) each contribute to the desired set; the controller merges them with explicit precedence. No state-machine interactions between strategies.
```
mtabebe: This makes sense, and I can't think of something better, but one challenge will be: how do you explain the decisions to the user after they create dumb policies, and how altering one affects the other.

I am guessing this is just a problem we solve with docs, but I wonder if we could somehow surface conflicting strategies.

doc/developer/design/20260522_cluster_autoscaling.md
mtabebe: Right, I'm not sure I understand this ... somewhere something has to know the previous state right? Maybe you are saying this is somehow implicitly remembered as part of a catalog mutation?

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,233 @@
...
...ely do *not* port 0dt's OOM-tolerance logic here: for burst the desired behavior is the opposite — a steady replica that is OOM-looping or otherwise unable to keep up should *prevent* tear-down, which falls out naturally from the catch-up check. The strategy block's `COOL DOWN` bounds how soon the controller may take a follow-on scaling action after a transition. Burst is bounded by a durable `burst_deadline` — from the optional `ON HYDRATION` `TIMEOUT`, else the dyncfg floor — written when the burst replica is created. If the deadline passes before tear-down, the strategy drops the burst replica and **retains the deadline as a tombstone**, keeping burst suppressed while the same objects stay un-hydrated; this is the same deadline-as-tombstone behavior reconfiguration uses (see [Failure handling](#failure-handling-and-safety-rails)). The tombstone resets when the un-hydrated set clears or on a reconfiguring `ALTER`, so a burst that cannot help a too-small steady set stops thrashing rather than re-launching every cool-down. One documented edge: a newly-added object does not re-engage burst while an earlier object remains stuck — resolving or resizing past the stuck object resets it.
```
mtabebe: > at least one steady replica has all currently-existing cluster objects fully hydrated

Are these somehow snapshotted when the reconfiguration starts? Is this an existing choice?

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,233 @@
...
+- **Deterministically auditable.** Same inputs always yield the same decisions; audit entries explain each decision from durable state plus observed status.
+
+#### Command idempotency
+
+Commands from the controller to the coordinator are **desired-state assertions** ("ensure a replica with config X exists for cluster C under name N", "ensure replica N is dropped"), not imperative actions. Each strategy contributes a fixed-shape set of replicas under deterministic names derived from the cluster and the strategy's intent (today's `r1..rN` for steady managed replicas is an example of this style), so a recomputed command is byte-identical to the previous one.
+
+The coordinator's `CreateClusterReplica` handler today errors on a duplicate name (`SqlCatalogError::DuplicateReplica`). For the controller path we extend it to check-by-name within the transaction: if a replica under the asserted name already exists with the asserted config, the transaction commits as a no-op; if it exists with a different config, that's a controller bug and is surfaced as an error.
```
mtabebe: So does this mean we allow two concurrent alter cluster commands?

doc/developer/design/20260522_cluster_autoscaling.md
aljoscha: currently we just wait for one to be hydrated, but we could change that in the future, imo

doc/developer/design/20260522_cluster_autoscaling.md
aljoscha: you mean a version of the ALTER that is blocking? We do have that, and the doc mentions that we can keep that using a shim. I mentioned we only wanted to keep initially but could also keep indefinitely

doc/developer/design/20260522_cluster_autoscaling.md
aljoscha: agreed, we could show that in SHOW CLUSTER once (if ever) we allow multiple strategies

doc/developer/design/20260522_cluster_autoscaling.md
aljoscha: we don't, and one "cancels" any previous one, but with these commands going to the Coordinator there can be races, there is concurrency

.bk.yaml
aljoscha: hrm, snuck in by accident 🙈

doc/developer/design/20260522_cluster_autoscaling.md
jubrad: If we're waiting objects to be hydrated across multiple replicas we might need to filter out unbilled replicas. First past not worrying about multiple replicas makes sense to me.

doc/developer/design/20260522_cluster_autoscaling.md
ggevay: Ok :)

doc/developer/design/20260522_cluster_autoscaling.md
ggevay: I think this comment is still open.

doc/developer/design/20260522_cluster_autoscaling.md
ggevay: ~Fine to not overthink multi-replica clusters for now, but then it would be good to mention them in the "Out of Scope" section.~ Edit: Actually, I see the "Example: a graceful size change" section does a multi-replica example since then.

doc/developer/design/20260522_cluster_autoscaling.md
ggevay: still open

doc/developer/design/20260522_cluster_autoscaling.md
ggevay: still open

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Summary
+
+We introduce a **cluster controller** as a dedicated task inside `environmentd`. It is the single decision-maker for the replica set of every cluster. It operates as a reconciler: it reads desired cluster state from the durable catalog, observes the actual replica set and live status signals, computes a desired replica set per cluster by combining a set of **strategies**, and emits catalog-change commands to the Coordinator via a message channel. The Coordinator remains the sole writer of catalog state.
+
+ALTER CLUSTER operations are reshaped so that the user's intent is written durably and immediately to the catalog — as an in-flight target record, leaving the cluster's realized config in place until the controller cuts over; the controller then reconciles from that intent. Graceful reconfiguration becomes a strategy within the controller framework, as does hydration burst. The existing `ON REFRESH` scheduling is lifted into the framework.
+
+### Strategies
+
+A **strategy** is two functions:
```
ggevay, line 64: When are these functions called? Simply periodically?

Edit: Ok, I see elsewhere "and kicks the controller", so then it seems certain operations also trigger a call of these functions, to not have to wait for the period, right?

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,244 @@
...
+
+Both functions of every strategy are pure: same inputs, same output, no side effects. The controller is the sole mutator, it applies the `update_state` results and the create/drop commands that close the gap between the combined desired set and the actual replicas.
+
+#### Combining strategy outputs
+
+The desired replica set for a cluster is the **union** of every strategy's `desired_replicas` contribution, the implicit baseline included. Strategies are purely additive, so they cannot contradict one another. A replica survives as long as *any* strategy desires it. The controller diffs the union against the actual replica set and emits the gap: a create for a replica the union calls for but actual lacks, a drop for a replica no strategy desires. This composes the strategies with no state-machine interaction between them. The controller retains each strategy's individual contribution, not just the merged union, so it can attribute every command it emits: a create records which strategies desired the replica ("required by graceful reconfiguration and hydration burst"), a drop records that none did. These attributions are the reasons carried on the audit events.
```
ggevay: > The desired replica set for a cluster is the **union** of every strategy's `desired_replicas` contribution

I'm not sure about the semantics of this `union`. The later

> A replica survives as long as *any* strategy desires it.

makes it sound like it's set semantics. However, does that mean that a strategy can't request more than on replica with the same size and configuration (which would need multiset semantics)?

ggevay: Ah, replica names are mentioned later in the document, so maybe set semantics are fine. (But then does it even happen that two strategies want to create the same replica with the same name?)

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,244 @@
...
+
+**Per-`ALTER` timeout: `WITH (TIMEOUT = '...')`.** Available on `ALTER CLUSTER ... SET (...)`. The duration is added to the transaction timestamp and written as the `reconfiguration` record's `deadline`. Resolution at the controller is `WITH (TIMEOUT = ...)` → system default (dyncfg). Omitting `WITH` on a reconfiguring `ALTER` falls back to that system default rather than leaving the transition unbounded. The session-side wait shim defaults its own wait to the same value, preserving today's foreground UX of "ALTER returns or errors within TIMEOUT"; the behavioral delta is that when the shim's wait expires the durable reconfiguration *continues* until the controller hits the same deadline, rather than the session aborting it.
```
ggevay: > The session-side wait shim defaults its own wait to the same value

Why does the session-side wait shim need its own timeout? Couldn't it just detect when the controller times out, and then time out the statement in reaction to that? If it has its own timeout, then maybe the two timeouts might slightly disagree, and then the statement might return with an error when in fact the change actually happened.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,244 @@
...
+
+- **`SHOW CLUSTERS` surfaces in-flight reconfigurations.** Today's `SHOW CLUSTERS` is implemented as SQL over `mz_clusters` and related views; we extend that SQL to join with the new introspection view so a user sees, per cluster, both the current size (the realized `cluster.size`) and the target size (the `reconfiguration` record's target, when one is in flight), plus an indication of whether a reconfiguration is in flight. The exact column set is left to implementation, but the design requires that a single `SHOW CLUSTERS` answers the questions "what did I ask for", "what's actually there now", and "is something in progress, done, or timed out" without requiring the user to know about the introspection view.
+
+- **History — audit log.** Extend the existing `SchedulingDecisionsWithReasonsV2` event family (or add closely-related event variants) to cover:
+  - Graceful reconfiguration: started, replica created, replica dropped, finalized, cancelled, timeout fired (for stuck reconfigurations).
+  - Hydration burst: started, replica created, replica dropped.
```
ggevay, line 183: So, both `update_state` and `desired_replicas` function calls can result in audit log events? Also, does `desired_replicas` have its own audit log event type, or does it affect the audit log only through replica creations/drops?

doc/developer/design/20260522_cluster_autoscaling.md
ggevay: (Still open)

ggevay (approved):
Thank you, it looks great to me!

Wrote some minor comments.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,244 @@
...
+#### The reconcile tick
+
+Each tick reconciles every cluster in two phases, separated by a barrier:
+
+1. **`update_state`.** The controller reads a fresh catalog snapshot and runs every strategy's `update_state`, collecting their durable writes: cut-overs (`cluster.size := target`), `reconfiguration` and `burst` record writes and clears. It submits that batch to the Coordinator and **awaits confirmation that it has been transacted.**
+2. **`desired_replicas`.** With those writes applied, the controller runs every strategy's `desired_replicas` against the now-current state, unions the contributions into the [combined desired set](#combining-strategy-outputs), diffs it against the actual replica set, and emits the creates and drops that close the gap.
+
+#### State convergence
+
+Each pass the controller **emits only the commands that close the gap**: a create for a replica that should exist but doesn't, a drop for one that shouldn't. It has the observed state, so it does not re-assert the full desired set every tick; it sends work only when there is a delta, keeping needless transactions off the coordinator's serial loop.
```
mtabebe, line 122: 👍

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,244 @@
...
+
+Advancing `cluster.size` at cut-over is the single durable write that retires the old set: it flips the implicit baseline from holding the `100cc` replicas to holding the `200cc` ones, and the old replicas fall out of the union on their own. Because the desired set is matched to actual by config and count, the freshly-named `r3`, `r4` satisfy the new steady state directly. Nothing is renamed, and the cluster never drops below `replication_factor` serving replicas.
+
+### Durable state model
+
+The cluster's durable configuration represents the cluster's **realized, currently-serving state**, what is actually running at steady state. We add additional records in cluster state for use by the strategies:
+
+- `auto_scaling_strategy: Option<AutoScalingStrategy>` — the strategy block (v1: `ON HYDRATION` with its `HYDRATION SIZE` and optional `TIMEOUT`). This is user-configured *policy*, distinct from the two transition records below, which are `ALTER`/controller-managed *runtime state*.
```
mtabebe: This makes sense to me

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,244 @@
...
+- `reconfiguration: Option<ReconfigurationState>` — an in-flight graceful reconfiguration: the `target` (the size / replication-factor / availability-zones / logging the cluster is moving to) and a `deadline`.
+- `burst: Option<BurstState>` — the analogous record for an active hydration burst: the `burst_size` of the in-flight burst replica, a `linger_duration`, and the timestamp at which we observed the steady-state replicas as hydrated. Burst is controller-initiated (not tied to an `ALTER`), the strategy writes the record when we determine burst is needed. It is cleared when burst tears down on success.
```
mtabebe, line 146: I like the notion of linger

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,244 @@
...
+
+**New cluster option: `AUTO SCALING STRATEGY = (...)`.** Available at `CREATE CLUSTER` and `ALTER CLUSTER ... SET (...)`. The value is a paren-enclosed list mixing strategy specs (currently only `ON HYDRATION (HYDRATION SIZE = '...', LINGER DURATION = '...')`, where `LINGER PERIOD` is optional — the dyncfg default applies when it is omitted). The whole block is set or replaced atomically; omitting it on ALTER (or specifying an empty list) disables auto-scaling for the cluster. Fine-grained ALTERs that change one sub-parameter are out of scope for v1.
```
mtabebe: linger period or linger duration?

mtabebe (approved):

doc/developer/design/20260522_cluster_autoscaling.md
jubrad: Yeah, we're already doubling our cluster count during upgrade rollouts with 0dt. I'd be worried if we also launched much larger hydration replicas on top of that.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Open Questions
+
+1. **`AUTO SCALING STRATEGY` + `SCHEDULE = ('on-refresh', ...)` combination.** v1 rejects this combination. Semantically the combination is interesting (every refresh window, burst comes up first to accelerate hydration, steady catches up, then the schedule turns the cluster off), but our strong recommendation is to keep this rejected indefinitely: there is currently no appetite to invest further in the `SCHEDULE` syntax, and supporting the combination would expand its surface area.
+
+2. **Multiple burst replicas (one per steady replica vs. one total).** v1 supports exactly one burst replica per cluster regardless of replication factor. Our strong recommendation is to keep it that way: the burst replica is by design transient, and provisioning N burst replicas to mirror an N-replica steady set would multiply cost for diminishing benefit — burst tear-down only requires one steady replica to have caught up. Revisit only if real-world usage proves the single-burst model insufficient.
```
jubrad, line 247: I think this is reasonable.
It is possible that burst replicas could lead to oddities with HA, If someone had 2 replicas, a hydration replica was created, then they bumped the replicas up to 3. min domains would be satisfied with the hydration replica and the third replica could be created on the same AZ as one of the other two permanent replicas. The likelihood of this being a real issue is very low, if we wanted to deal with it we'd have to exclude this from counting towards the cluster replicas for min domains / topology spread.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,244 @@
...
+
+5. **Introspection view naming and exact column layout.** Working name `mz_internal.mz_cluster_reconfigurations` is a placeholder.
+
+6. **Foreground/synchronous graceful reconfiguration retention.** Our strong recommendation is to deprecate the current foreground (session-bound) mechanism in favor of the background model. During rollout, the foreground experience is preserved as a thin session-side wait shim over the background mechanism (see [SQL surface](#sql-surface)) — *not* by retaining the existing parallel state machine. This means the `pending: bool` flag on replicas and the associated three-stage machinery can be removed up front; deprecating the foreground experience later is simply deleting the wait shim. The one behavioral difference vs. today is that session disconnect during the wait no longer aborts the reconfiguration (arguably a feature; the durable target stays set and the controller continues).
```
jubrad, line 255: Yeah, I think this is the right call.

doc/developer/design/20260522_cluster_autoscaling.md
```diff
@@ -0,0 +1,244 @@
...
...g "current" from the replica set).** Considered and rejected. In this alternative `ALTER` writes the new size straight into `cluster.size` and the controller diffs against it, with "what is actually serving" derived from the replica set. It is appealing — `cluster.size` is the obvious home for "the target," re-targeting is just another write, and the cut-over needs no config write (only dropping the old replicas). We rejected it because it makes `cluster.size` a *wish* rather than a *fact*: mid-reconfiguration `mz_clusters.size` shows a size nothing is serving yet, and a **timed-out** reconfiguration leaves it showing the unrealized target *indefinitely* (until the user `ALTER`s again) while the cluster serves the old size. Every reader then has to derive "what is actually running" from the replica set and is told never to trust `cluster.size` alone, and the timeout-park state needs a tombstone precisely to hold that persistent divergence stable. Instead we keep `cluster.size` the **realized fact** and put the in-flight target in the `reconfiguration` record. This costs the "see the new size in `SHOW CREATE` the instant `ALTER` returns" property, which we judged the lesser concern.
```
jubrad: > it makes `cluster.size` a *wish* rather than a *fact*

I'm not so sure about this. If I run `create cluster foo (size='xlarge', replication factor = 3);` in todays main branch. that is actually expressing an intent, *not* a fact. The underlying provisioning system may not actually be able to fulfill the request, perhaps we just don't have capacity in the cluster for 3 more xlarge replicas. This only becomes a fact when the replicas for the cluster report that they are ready, and can stop being a fact at any point.


> Every reader then has to derive "what is actually running" from the replica set and is told never to trust `cluster.size` alone

Yes I think that it is accurate that you can't trust cluster.size for the operational state of the cluster, but you should be able to trust it for the intended state. One thing you can do here (or that we do on the cloud side), is set some metadata (k8s status) on resources to indicate that their operational state is in sync with their target state. Additional metadata could be placed in the object to help provide context on the current operational state, for for instance a

aljoscha: > in todays main branch. that is actually expressing an intent, not a fact. The underlying provisioning system may not actually be able to fulfill the request, perhaps we just don't have capacity in the cluster for 3 more xlarge replicas.

I agree that this is true, but from a users perspective, and the perspective of parts of the system what's durable state is fact. When I run `CREATE CLUSTER REPLICA ...` as a user, that replica exists even though that might not have percolated all the way through the different layers. The thing about wish and fact here is talking about the catalog/controller interaction, and the difference between the target being encoded in `cluster.size` vs. an in-flight reconfiguration record. 

And we keep having things like `mz_cluster_replica_status...` for looking into the actual runtime status of replicas.

aljoscha:
@mtabebe @ggevay @jubrad I know you approved already, but I pushed changes that hopefully address all your follow-up comments. In case you want to look, but I might merge today regardless.

Thanks for the reviews! 🙇‍♂️

PR #36692, opened 2026-05-22
[SS-69] use updated iceberg-rust fork. new auth hooks for S3 Tables REST and S3 FileIO

ublubu:
In preparation for sinking to GCP's Lakehouse, I've pulled in the stable release `v0.9.0` for `iceberg-rust` and added a `RequestAuthenticator` trait for customizing auth support.

Auth options:
- AWS SigV4 (_now in this PR, previously in our `iceberg-rust` fork_)
- GCP (_via `gcp_auth` in an upcoming PR_)

----

~~This PR is still in _Draft_ state because it depends on `iceberg-rust` changes that have not yet been reviewed. (e.g. https://github.com/MaterializeInc/iceberg-rust/pull/1)~~

### DIFF
--- a/src/storage-types/src/connections.rs
+++ b/src/storage-types/src/connections.rs
@@ -184,11 +184,21 @@ impl RequestAuthenticator for Sigv4Authenticator {
             .and_then(|b| b.as_bytes())
             .map(|b| b.to_vec())
             .unwrap_or_default();
-        let headers: Vec<(&str, &str)> = req
+        let headers = req
             .headers()
             .iter()
-            .map(|(k, v)| (k.as_str(), v.to_str().unwrap_or("")))
-            .collect();
+            .map(|(k, v)| {
+                Ok((
+                    k.as_str(),
+                    v.to_str().map_err(|_| {
+                        iceberg::Error::new(
+                            iceberg::ErrorKind::DataInvalid,
+                            format!("header '{}' value is not all visible ASCII", k),
+                        )
+                    })?,
+                ))
+            })
+            .collect::<iceberg::Result<Vec<(&str, &str)>>>()?;
         let signable = SignableRequest::new(
             req.method().as_str(),
             req.url().as_str(),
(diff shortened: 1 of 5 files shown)


patrickwwbutler (approved):
Seems fine to me! I would just make sure you test the changes with a real connection, and not just in testdrive

src/storage-types/src/connections.rs
```diff
@@ -129,6 +136,96 @@ impl AwsCredentialLoad for AwsSdkCredentialLoader {
...
+            .into();
+        let body: Vec<u8> = req
+            .body()
+            .and_then(|b| b.as_bytes())
+            .map(|b| b.to_vec())
+            .unwrap_or_default();
+        let headers: Vec<(&str, &str)> = req
+            .headers()
+            .iter()
+            .map(|(k, v)| (k.as_str(), v.to_str().unwrap_or("")))
```
martykulma: Any cases where `HeaderValue` is not ASCII? My understanding is this will end up as empty string in that case, which may cause subtle errors/failures.

src/storage-types/src/connections.rs
```diff
@@ -129,6 +136,96 @@ impl AwsCredentialLoad for AwsSdkCredentialLoader {
...
+        let params = v4::SigningParams::builder()
+            .identity(&identity)
+            .region(&self.region)
+            .name(&self.signing_name)
+            .time(SystemTime::now())
+            .settings(SigningSettings::default())
+            .build()
+            .map_err(sigv4_err)?
+            .into();
+        let body: Vec<u8> = req
```
martykulma: If the body is empty, does this still sign the body?

src/storage-types/src/connections.rs
ublubu: I've made this into an error instead.

src/storage-types/src/connections.rs
ublubu: It's treating empty body the same as absent body, which Claude tells me is fine. In any case, the shape of `SignableBody` suggests the same.

But we were turning all streaming request bodies into `""`, so I made that an error as well.

PR #36695, opened 2026-05-22
[SS-69] Iceberg catalog/sink can use a GCP connection

ublubu:
**In this PR:**
- `iceberg-rust` `RequestAuthenticator` using `gcp_auth`
- Iceberg catalog connection can use a GCP connection now.
  - Iceberg sink _just works_.
- new Nightly CI group "GCP" with a GCP test
  - uses https://github.com/MaterializeInc/i2/pull/3279

N.B. GCP BigLake doesn't perform Iceberg maintenance yet, so long-running Iceberg sinks will fail when the table metadata grows too large.
(ref: https://cloud.google.com/blog/products/data-analytics/improved-interoperability-for-your-apache-iceberg-lakehouse)

_stacked on #36694_

### DIFF
--- a/ci/nightly/pipeline.template.yml
+++ b/ci/nightly/pipeline.template.yml
@@ -959,6 +959,20 @@ steps:
           - ./ci/plugins/mzcompose:
               composition: secrets-local-file
 
+  - group: GCP
+    key: gcp
+    steps:
+      - id: gcp-real
+        label: GCP (Real)
+        depends_on: build-aarch64
+        timeout_in_minutes: 30
+        agents:
+          # because of GCP access
+          queue: linux-aarch64-small
+        plugins:
+          - ./ci/plugins/mzcompose:
+              composition: gcp
+
   - group: "Platform checks"
     key: platform-checks
     steps:
--- a/src/storage-types/src/connections.rs
+++ b/src/storage-types/src/connections.rs
@@ -915,6 +915,7 @@ impl IcebergCatalogConnection<InlinedConnection> {
                     GCS_CREDENTIALS_JSON.to_owned(),
                     base64::engine::general_purpose::STANDARD.encode(creds_json),
                 );
+                // We supplied a service account key. Don't look elsewhere for GCP credentials.
                 props.insert(GCS_DISABLE_VM_METADATA.to_owned(), "true".to_owned());
                 props.insert(GCS_DISABLE_CONFIG_LOAD.to_owned(), "true".to_owned());
                 if let Some(project_id) = service_account.project_id() {
--- a/src/storage-types/src/connections.rs
+++ b/src/storage-types/src/connections.rs
@@ -22,10 +22,14 @@ use aws_sigv4::http_request::{SignableBody, SignableRequest, SigningSettings, si
 use aws_sigv4::sign::v4;
 // Aliased to avoid colliding with `mz_ccsr::tls::Identity`.
 use aws_smithy_runtime_api::client::identity::Identity as AwsIdentity;
+use base64::Engine;
 use http::{HeaderName, HeaderValue};
 use iceberg::Catalog;
 use iceberg::CatalogBuilder;
-use iceberg::io::{S3_ACCESS_KEY_ID, S3_DISABLE_EC2_METADATA, S3_REGION, S3_SECRET_ACCESS_KEY};
+use iceberg::io::{
+    GCS_CREDENTIALS_JSON, GCS_DISABLE_CONFIG_LOAD, GCS_DISABLE_VM_METADATA, GCS_USER_PROJECT,
+    S3_ACCESS_KEY_ID, S3_DISABLE_EC2_METADATA, S3_REGION, S3_SECRET_ACCESS_KEY,
+};
 use iceberg_catalog_rest::{
     REST_CATALOG_PROP_URI, REST_CATALOG_PROP_WAREHOUSE, RequestAuthenticator, RestCatalogBuilder,
 };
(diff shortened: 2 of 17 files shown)


ci/nightly/pipeline.template.yml
```diff
@@ -961,6 +961,24 @@ steps:
...
+    key: gcp
+    steps:
+      - id: gcp-real
+        label: GCP (Real)
+        depends_on: build-aarch64
+        timeout_in_minutes: 30
+        retry:
+          automatic:
+            - exit_status: 1
+              limit: 1
```
def-: Why auto-retry? Do we expect it to be flaky?

ublubu: I don't expect it to be flaky. I'll turn off retries.

src/storage-types/src/connections/gcp.rs
```diff
@@ -95,3 +106,59 @@ impl GcpConnectionValidationError {
...
+            .map_err(|e| {
+                iceberg::Error::new(
+                    iceberg::ErrorKind::Unexpected,
+                    format!("gcp_auth: {}", e.display_with_causes()),
+                )
+            })?;
+        Ok(token.as_str().to_string())
+    }
+
+    async fn invalidate(&self) -> iceberg::Result<()> {
```
patrickwwbutler, line 182: i might be missing something but should we not still have some sort of manual invalidate in case we need to force expire a token? I guess if that need doesn't arise during our usual workflow then maybe it's fine?

patrickwwbutler (approved):
Overall looks good besides the one question!

def- (commented):
How worried are we about someone being able to exfiltrate the GCP OAuth2 access token? I would argue we should be quite worried and only allow passing it to actual GCP URLs. Currently this works:
```diff
diff --git a/test/iceberg/gcp-connection-validation.td b/test/iceberg/gcp-connection-validation.td
new file mode 100644
index 0000000000..a3a4435bf2
--- /dev/null
+++ b/test/iceberg/gcp-connection-validation.td
@@ -0,0 +1,41 @@
+# Copyright Materialize, Inc. and contributors. All rights reserved.
+#
+# Use of this software is governed by the Business Source License
+# included in the LICENSE file at the root of this repository.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0.
+
+# Regression test for SS-69: an Iceberg REST catalog connection that
+# authenticates with a GCP connection must only send its reusable, broadly-
+# scoped bearer token to a Google-operated host. Without a plan-time host check,
+# a principal with only USAGE on the GCP connection can aim the catalog URL at a
+# host they control and exfiltrate the service-account token.
+
+# Disable default validation so the accepted case below doesn't connect; the
+# host check is enforced at plan time, independent of validation.
+$ postgres-execute connection=postgres://mz_system:materialize@${testdrive.materialize-internal-sql-addr}
+ALTER SYSTEM SET enable_default_connection_validation = false
+
+> CREATE SECRET gcpsa AS 'dummy'
+
+> CREATE CONNECTION gcpconn TO GCP (SERVICE ACCOUNT KEY = SECRET gcpsa)
+
+# Exploit: a non-Google catalog host must be rejected before any token is sent.
+! CREATE CONNECTION evil TO ICEBERG CATALOG (
+    CATALOG TYPE = 'rest',
+    URL = 'https://attacker.example.com/iceberg/v1/restcatalog',
+    GCP CONNECTION = gcpconn,
+    WAREHOUSE = 'gs://anything'
+  )
+contains:must point URL at a *.googleapis.com host
+
+# Positive control: a real BigLake host is accepted, so the allowlist doesn't
+# break the feature.
+> CREATE CONNECTION biglake TO ICEBERG CATALOG (
+    CATALOG TYPE = 'rest',
+    URL = 'https://biglake.googleapis.com/iceberg/v1/restcatalog',
+    GCP CONNECTION = gcpconn,
+    WAREHOUSE = 'gs://anything'
+  )
diff --git a/test/iceberg/mzcompose.py b/test/iceberg/mzcompose.py
index 78aa241741..990848d1fb 100644
--- a/test/iceberg/mzcompose.py
+++ b/test/iceberg/mzcompose.py
@@ -80,6 +80,22 @@ def workflow_smoke(c: Composition) -> None:
     )


+def workflow_gcp_connection_validation(c: Composition) -> None:
+    """Regression test for SS-69: an Iceberg REST catalog connection that
+    authenticates with a GCP connection must only target Google-operated catalog
+    hosts. A GCP access token is a reusable, broadly-scoped bearer credential
+    with no audience binding, so before this validation a principal with only
+    USAGE on a GCP connection could point the catalog URL at an attacker host and
+    exfiltrate the connection's service-account token. This exercises connection
+    planning only and needs no Iceberg backend."""
+    c.down(destroy_volumes=True)
+    c.up("materialized")
+
+    c.run_testdrive_files(
+        "gcp-connection-validation.td",
+    )
+
+
 def workflow_mode_append(c: Composition) -> None:
     key = _setup(c)
```
and allows someone with only `USAGE` on the GCP connection object to send the oauth token to attacker.example.com.

src/storage-types/src/connections/gcp.rs
ublubu: We don't actually use this function, so it shouldn't matter. Anyway, `gcp_auth` hides its innards from us. Hopefully it's smart enough to just "do the right thing."

src/storage-types/src/connections.rs
```diff
@@ -813,29 +861,71 @@ impl IcebergCatalogConnection<InlinedConnection> {
...
-                // `iceberg-rust` forwards these props to `opendal`.
-                // N.B. This is not confirmed to work with other catalog & storage implementations.
-                customized_credential_load: None,
-            }))
+                props.insert(
+                    GCS_CREDENTIALS_JSON.to_owned(),
+                    base64::engine::general_purpose::STANDARD.encode(creds_json),
+                );
+                props.insert(GCS_DISABLE_VM_METADATA.to_owned(), "true".to_owned());
+                props.insert(GCS_DISABLE_CONFIG_LOAD.to_owned(), "true".to_owned());
```
martykulma, line 920: a comment about why these are set to true would be help!

martykulma (approved):
nice!

PR #36697, opened 2026-05-22
expr-derive-impl: support stateful unary/binary in sqlfunc macro

antiguru:
Extends `#[sqlfunc(...)]` to accept stateful structs and arena access for unary and binary functions. Bumps `EagerUnaryFunc::call` to take `&'a self` and `&'a RowArena`, mirroring the variadic shape. Adds `skip_display = true` modifier for structs needing custom `Display`.

Migrates 19 hand-written `LazyUnaryFunc` impls (compound-type casts, regex match/split, record_get, etc.) in `src/expr/src/scalar/func/impls/*` to the macro form.

Remaining hand-written impls: the 6 short-circuit variadics (`And`, `Or`, `Coalesce`, `ErrorIfNull`, `Greatest`, `Least`) plus `CaseLiteral`, which require true lazy semantics over `MirScalarExpr`.

Prep work for collapsing per-sqlfunc monomorphizations into vtable dispatch ([prompt](doc/developer/prompts/sqlfunc-dyn-dispatch.md)).

### DIFF
--- a/src/expr-derive-impl/src/lib.rs
+++ b/src/expr-derive-impl/src/lib.rs
@@ -141,6 +141,28 @@ mod test {
 
     #[cfg_attr(miri, ignore)] // unsupported operation: extern static `pidfd_spawnp` is not supported by Miri
     #[mz_ore::test]
+    fn insta_test_unary_self_arena() {
+        let attr = quote! {
+            CastArrayToString,
+            sqlname = "arraytostr",
+            preserves_uniqueness = true,
+            introduces_nulls = false,
+        };
+        let item = quote! {
+            fn cast_array_to_string<'a>(
+                &self,
+                a: Array<'a>,
+                temp_storage: &'a RowArena,
+            ) -> Result<&'a str, EvalError> {
+                unimplemented!()
+            }
+        };
+        let (output, input) = super::test_sqlfunc(attr, item);
+        insta::assert_snapshot!("unary_self_arena", output, &input);
+    }
+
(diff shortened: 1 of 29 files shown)


antiguru:
Closing in favour of a reworked stack. Posted by Claude Code.

The capability this adds is still wanted, but this branch is not the right vehicle for it any more. Two reasons.

The macro work here is written three times, once per generator arm. `skip_display`, the `has_self` emission branch, the `call_expr`, and deriving the struct name from `struct_ty` each appear in `unary_func`, `binary_func`, and `variadic_func`. That triplication is the thing the capability was meant to reduce, so the replacement canonicalizes the three arms behind one generator first, and adds the capability to that single path afterwards.

The conversions are no longer correct against `main`. This branch is based on 2026-05-22, and it converts the compound-type casts by hardcoding `Box<MirScalarExpr>`. That was right at the time. It is wrong now: [#37961](https://github.com/MaterializeInc/materialize/pull/37961) introduced the `<E = MirScalarExpr>` parameter on 2026-08-12 to finish the MIR and LIR separation, and `LirScalarExpr` also implements `Eval`. 26 of the 29 files this branch touches have changed on `main` since.

Design document: `doc/developer/design/20260917_sqlfunc_canonicalization.md`, landing as the first pull request of the replacement stack.

Two decisions from this branch are carried over unchanged, because they were the right calls: `skip_display` as the mechanism for state-dependent names, and giving `EagerUnaryFunc::call` the `RowArena` that binary and variadic already had.

PR #36701, opened 2026-05-24
storage-types: use scheme default port for CSR SSH tunnel

def-:
Previously `url::Url::port()` returned `None` for default ports, so HTTPS schema registry URLs without an explicit port forwarded SSH tunnel traffic to port 80, causing TLS handshake failures.


martykulma (approved):
nice - thanks @def- !

PR #36703, opened 2026-05-24
storage-types: redact credentials in ContentSource Debug

def-:
Use SensitiveUrl for ContentSource::Http and skip the AwsConnection field in Debug so reconcile log lines no longer leak url passwords or inline AWS credentials. Also narrow the storage_state reconcile log to ids only.

Carry SensitiveUrl through HttpOneshotSource and HttpObject so the `tracing::info!(?object, "found objects")` line in the HTTP oneshot source also redacts. Move URL-userinfo Basic auth into a Basic auth header before handing the URL to reqwest, so reqwest::Error::Display cannot leak the password on connect/DNS failures.

### DIFF
--- a/src/storage-operators/src/oneshot_source/http_source.rs
+++ b/src/storage-operators/src/oneshot_source/http_source.rs
@@ -16,10 +16,10 @@ use bytes::Bytes;
 use derivative::Derivative;
 use futures::TryStreamExt;
 use futures::stream::{BoxStream, StreamExt};
+use mz_ore::url::SensitiveUrl;
 use reqwest::Client;
 use reqwest::dns::{Addrs, Name, Resolve, Resolving};
 use serde::{Deserialize, Serialize};
-use url::Url;
 
 use crate::oneshot_source::util::IntoRangeHeaderValue;
 use crate::oneshot_source::{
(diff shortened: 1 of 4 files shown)


def-:
Ready for review

src/storage-operators/src/oneshot_source/http_source.rs
```diff
@@ -87,20 +87,49 @@ fn check_not_redirect(response: &reqwest::Response) -> Result<(), StorageErrorX>
...
+/// so the password does not surface via `reqwest::Error::Display` (which
+/// renders the URL using `Url::Display` and would otherwise leak it on
+/// connect/DNS failures).
+///
+/// `Url::username` and `Url::password` return values in their
+/// percent-encoded form, but `RequestBuilder::basic_auth` base64-encodes its
+/// arguments verbatim. Decode here so credentials containing characters that
+/// must be percent-encoded in a URL (`@`, spaces, `:`, ...) authenticate
+/// correctly.
+fn build_request(
```
martykulma: This appears to be doing the same thing that [reqwest already does in RequestBuilder](https://github.com/seanmonstar/reqwest/blob/11489b34eda6d32b15ad4033e62beba2ee401350/src/async_impl/request.rs#L172-L186).  Was there an observed leak of this information?

martykulma (commented):
Nice - the SensitiveUrl change looks good.  Not sure the `build_request` method and deps are needed, but not sure how we detected a credentials leak.

src/storage-operators/src/oneshot_source/http_source.rs
def-: True, using that now.

martykulma (approved):
lgtm, thanks @def- !

PR #32087, opened 2025-04-03
[sql_server] Create all `*Source` related types, support `CREATE CONNECTION ... TO SQL SERVER`

ParkMyCar:
This PR adds all of the various SQL Server 'Source' and 'Connection' types, and leaves the implementations of Purification and Source Rendering as TODOs. Given the number of new types required, and files that need to be touched, I figured it was easier to make this into its own PR.

The two main types this PR adds are:
1. `SqlServerSource` (analogous to `PostgresSourceConnection`)
    * Implements the `SourceConnection`, `AlterCompatible`, and `SourceRender` traits.
2. `SqlServerConnectionDetails` (analogous to `PostgresConnection`)
    * Implements `IntoInlineConnection` and `AlterCompatible`.

Note: I slightly differed from the naming convention here because to me `PostgresSourceConnection` and `PostgresConnection` were similar enough that I kept getting the two confused.

We also add several types related to these two:
1. `SqlServerSource`
    * `SqlServerSourceExtras` - like the MySQL source, this type is empty but include to conform with other sources.
    * `SqlServerSourceExportDetails` - details of the specific upstream tables we'll be replicating.
    * `GenericSourceConnection::SqlServer` variant - enum wrapper around all `SOURCE` types.
2. `SqlServerConnectionDetails`
    * `Connection::SqlServer` variant - enum wrapper around all `CONNECTION` types.

### Feature Changes

Everything in this PR is still gated behind the `enable_sql_server_source` flag. When this flag is enabled, the only feature related change is `CREATE CONNECTION ... TO SQL SERVER` is now supported!

Otherwise purification in the Adapter and implementation of the `SourceRender` trait are left as a TODO which I will include in a follow-up PR.

### Testing

This PR also adds two tests to exercise the new connection behavior, and will eventually be used to exercise CDC:

1. `tests/sql-server-cdc` - testdrive based mzcompose workflows, asserts creating and validating a `CONNECTION` is successful
2. `tests/platform-checks` - a new `SqlServerCdc` check is added to ensure creating and dropping connections works across restarts and upgrades

### Motivation

Progress towards https://github.com/MaterializeInc/database-issues/issues/8762

### Tips for reviewer

Overall a lot of this is just plumbing and cargo culting what already exists for Postgres and MySql. I would focus on the changes in the `storage-types` crate, that's where the most net new code is.

### Checklist

[...]

PR #36704, opened 2026-05-24
storage-types: implement AlterCompatible for SqlServerSourceExportDetails

def-:
The variant was added in #32087 without updating
SourceExportDetails::alter_compatible, so any non-identical pair of SqlServer details fell through to the wildcard and returned an error, which crashes clusterd reconciliation via
.expect("only alter compatible ingestions permitted"). Mirror the MySql pattern: defer field-level checks to the source render operators.


martykulma (approved):
looks reasonable, thanks @def- !

PR #36705, opened 2026-05-24
expr: direct-unwrap LazyBinaryFunc for sqlfunc-generated structs

antiguru:
Stacked on #36697.

Sqlfunc-generated binary structs now emit an explicit `impl LazyBinaryFunc` whose body calls `try_from_result` per argument instead of going through the generic `<(T0, T1) as InputDatumType>::try_from_iter`. The tuple `try_from_iter` was the single largest contributor to mz-expr's LLVM IR (93k lines / 218 copies) and produced a costly iterator state machine in every binary variant's dispatch path.

The blanket `impl<T: EagerBinaryFunc> LazyBinaryFunc for T` is removed; the two remaining hand-written `EagerBinaryFunc` impls (`ListLengthMax`, `RegexpReplace`) use a new `lazy_via_eager_binary!` declarative macro that pays the tuple cost only for those two shapes. Unary and variadic paths are unchanged.

Measurements:

| metric | before | after | Δ |
|---|---|---|---|
| `cargo llvm-lines -p mz-expr` total | 1,289,072 | 1,232,567 | -4.4% |
| `<(T0,T1)>::try_from_iter` | 93,259 / 218 | 0 | killed |
| per-variant binary dispatch (LLVM) | ~608 | ~472 | -22% |
| per-variant binary dispatch (asm) | 949 | 726 | -23.5% |
| total binary dispatch asm | 201,214 | 153,888 | -47k insns |

No behavior change: `cargo test -p mz-expr` (49), `cargo test -p mz-expr-derive-impl` (14 snapshot tests), `test_runner` datadriven suite, and `cargo check --workspace` all pass. Snapshots regenerated via `cargo insta accept`.

Prompt at [`doc/developer/prompts/sqlfunc-dyn-dispatch.md`](doc/developer/prompts/sqlfunc-dyn-dispatch.md).

### DIFF
--- a/src/expr-derive-impl/src/lib.rs
+++ b/src/expr-derive-impl/src/lib.rs
@@ -141,6 +141,28 @@ mod test {
 
     #[cfg_attr(miri, ignore)] // unsupported operation: extern static `pidfd_spawnp` is not supported by Miri
     #[mz_ore::test]
+    fn insta_test_unary_self_arena() {
+        let attr = quote! {
+            CastArrayToString,
+            sqlname = "arraytostr",
+            preserves_uniqueness = true,
+            introduces_nulls = false,
+        };
+        let item = quote! {
+            fn cast_array_to_string<'a>(
+                &self,
+                a: Array<'a>,
+                temp_storage: &'a RowArena,
+            ) -> Result<&'a str, EvalError> {
+                unimplemented!()
+            }
+        };
+        let (output, input) = super::test_sqlfunc(attr, item);
+        insta::assert_snapshot!("unary_self_arena", output, &input);
+    }
+
+    #[cfg_attr(miri, ignore)] // unsupported operation: extern static `pidfd_spawnp` is not supported by Miri
+    #[mz_ore::test]
     fn insta_test_binary_arena() {
         let attr = quote! {test = true};
         let item = quote! {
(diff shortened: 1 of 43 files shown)


antiguru:
Closing in favour of a reworked stack. Posted by Claude Code.

This depends on [#36697](https://github.com/MaterializeInc/materialize/pull/36697), which is being closed for the same reason: both branch from 2026-05-22 and their conversions predate [#37961](https://github.com/MaterializeInc/materialize/pull/37961), which parameterized `UnaryFunc` over an `Eval` implementor to finish the MIR and LIR separation.

The change itself is still wanted. Removing the blanket `impl<T: EagerBinaryFunc> LazyBinaryFunc for T` and emitting an explicit per-struct implementation that unwraps each argument with `try_from_result`, rather than routing through the tuple `InputDatumType::try_from_iter`, is carried into the replacement stack as its own pull request, together with the `lazy_via_eager_binary!` shim for `ListLengthMax` and `RegexpReplace`.

The measurements recorded here are not being reused. They were taken against a May 2026 tree, before #37961 reshaped the dispatch path, so the replacement re-measures on the current tree and reports what it actually finds. If the win no longer holds, the pull request either lands on the narrower argument that per-struct emission is more explicit than a blanket implementation, or is dropped.

Design document: `doc/developer/design/20260917_sqlfunc_canonicalization.md`.

PR #36706, opened 2026-05-24
expr: recover pushdown for day-only interval predicates

antiguru:
### Motivation

After #36702 (merged) marked `add/sub_timestamp{,_tz}_interval` as `(false, false)` to fix the persist-filter-pushdown audit panic (database-issues#9656), filter pushdown was lost for **literal day-only interval** temporal predicates like `t_col - INTERVAL '1' day < literal` — exactly the pattern temporal filters were built to optimize. The non-monotonicity that motivated the parent fix only manifests when the interval has a non-zero `months` component (calendar-month arithmetic with day-clamping); for `months == 0` the operation reduces to adding a fixed number of microseconds and is straightforwardly monotone in `t`.

### Description

Adds a dynamic-monotonicity escape hatch to the abstract interpreter so the `(false, false)` static annotation can be safely upgraded at runtime when the spec proves the interval is harmless.

`SpecialBinary` (renamed to `AbstractFunc` per review) grows from a single `map_fn` override into an enum with two variants:

```rust
struct AbstractFunc {
    handler: AbstractFuncHandler,
    pushdownable: (bool, bool),  // (left, right): see Trace
}

enum AbstractFuncHandler {
    /// Existing: replace the spec computation entirely.
    Override(fn(ResultSpec, ResultSpec) -> ResultSpec),
    /// New: use the default flat-map machinery, but compute the per-argument
    /// monotonicity verdict from the input specs.
    DynamicMonotone(fn(&ResultSpec, &ResultSpec) -> (bool, bool)),
}
```

`AddTimestampInterval`, `AddTimestampTzInterval`, `SubTimestampInterval`, `SubTimestampTzInterval` now get an `AbstractFunc` whose `DynamicMonotone` handler returns `(months_zero, months_zero)` when the right argument is a single known interval with `months == 0`, and `(false, false)` otherwise. The static `is_monotone()` annotation from #36702 stays `(false, false)`, so anywhere this dynamic check isn't consulted (e.g. `Trace`'s static pushdownability gate) the conservative answer still applies. `AbstractFunc::pushdownable` is set to `(true, false)` so `Trace` routes `t_col +/- INTERVAL_lit` predicates through pushdown regardless of months — at runtime the dynamic check decides whether to narrow.

A `Values::as_single() -> Option<Datum>` helper centralises the "is this a single known value?" check, so the test against `Interval::months` is robust to future variants of `Values` (e.g. a small-discrete-set representation).

### Verification

- `interpret::tests::test_timestamp_plus_interval_dynamic_monotone` — three-scenario regression test covering tight narrowing (day-only interval, range straddling the literal), provable elimination (day-only, range strictly above), and conservative fallback (month-bearing literal).
- `interpret::tests::proptest_timestamp_plus_interval_monotone_when_months_zero` — proptest that exercises the monotonicity claim against the actual `add_timestamp_interval` impl by sampling random timestamps and zero-month intervals, checking that input ordering is preserved in the output.
- `test/sqllogictest/filter-pushdown.slt` — end-to-end test asserting `pushdown=...` is emitted for `WHERE t - INTERVAL '1' day < literal`.

### Known limitations

The dynamic upgrade only triggers when the interval is a *constant* with statically-zero months. Day-only column-valued intervals don't get pushdown today — that's a smaller win to leave for later. CASE expressions over multiple day-only interval literals also don't recover (the CASE folds into a `Values::Within` range, not a single value), but that's the right cost trade-off given the conservative static annotation: the existing slt tests at lines 322-353 still reflect that.

### DIFF
--- a/src/expr/src/interpret.rs
+++ b/src/expr/src/interpret.rs
@@ -123,6 +123,21 @@ impl<'a> Values<'a> {
             },
         }
     }
+
+    /// Returns the sole datum in this value set, if it is known to be a single
+    /// value. Returns `None` otherwise (for empty sets, ranges with distinct
+    /// endpoints, structured constraints, and the unconstrained set).
+    ///
+    /// Prefer this over pattern-matching on [Values::Within] directly when you
+    /// only need the "single known value" case: it's robust against future
+    /// variants of [Values] (e.g. a small-set representation) automatically
+    /// degrading to "not a single value" rather than silently mis-matching.
+    fn as_single(&self) -> Option<Datum<'a>> {
+        match self {
+            Values::Within(a, b) if a == b => Some(*a),
+            _ => None,
+        }
+    }
 }
 
 /// An approximation of the set of values an expression might have, including whether or not it
@@ -602,18 +617,43 @@ impl SpecialUnary {
     }
 }
 
-/// A binary function we've added special-case handling for; including:
-/// - A two-argument function, taking and returning [ResultSpec]s. This overrides the
-///   default function-handling logic entirely.
+/// The abstract-domain counterpart of a [BinaryFunc]: a binary function
+/// we've added special-case handling for; including:
+/// - Either a complete override of [ResultSpec] computation, or a way to
+///   compute monotonicity dynamically from the input specs.
 /// - Metadata on whether / not this function is pushdownable. See [Trace].
-struct SpecialBinary {
-    map_fn: for<'a> fn(ResultSpec<'a>, ResultSpec<'a>) -> ResultSpec<'a>,
+///
+/// Note: today a function can have *either* a handler override *or* a
+/// dynamic-monotonicity verdict, but not both. If a future function wants
+/// both, promote [AbstractFuncHandler] from an enum to a struct with two
+/// optional fields.
+struct AbstractFunc {
+    handler: AbstractFuncHandler,
+    /// `(left, right)`: per-argument pushdownability hint consumed by
+    /// [Trace]. `true` for an argument means the function preserves enough
+    /// structure that, with sufficient information about that argument's
+    /// range, the output spec can be predicted — i.e. the predicate is a
+    /// pushdown candidate when that argument is constant or a tight range.
     pushdownable: (bool, bool),
 }
 
-impl SpecialBinary {
+/// How an [AbstractFunc] computes the output [ResultSpec].
+enum AbstractFuncHandler {
+    /// Completely override the spec computation; the default flat-map machinery
+    /// is bypassed.
+    Override(for<'a> fn(ResultSpec<'a>, ResultSpec<'a>) -> ResultSpec<'a>),
+    /// Use the default flat-map machinery, but with a monotonicity verdict that
+    /// depends on the input specs. This lets us claim monotonicity for cases
+    /// the static `LazyBinaryFunc::is_monotone` annotation can't safely claim:
+    /// for instance, `t + INTERVAL '1' day` is monotone in `t`, but `t + i`
+    /// generally isn't (the calendar-month / day-clamping arithmetic in
+    /// `add_timestamp_interval` is non-monotone when `i.months != 0`).
+    DynamicMonotone(fn(&ResultSpec<'_>, &ResultSpec<'_>) -> (bool, bool)),
+}
+
+impl AbstractFunc {
     /// Returns the special-case handling for a particular function, if it exists.
-    fn for_func(func: &BinaryFunc) -> Option<SpecialBinary> {
+    fn for_func(func: &BinaryFunc) -> Option<AbstractFunc> {
         /// Eager in the same sense as `func.rs` uses the term; this assumes that
         /// nulls and errors propagate up, and we only need to define the behaviour
         /// on values.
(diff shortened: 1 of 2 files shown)


antiguru:
Re Dov's request — here's the proptest. My local shell got wedged before I could push, but the diff is small and lives in `src/expr/src/interpret.rs` just after `test_timestamp_plus_interval_dynamic_monotone`. Posting here so it survives a session reset; will commit and push once I can run `cargo test` again locally.

```rust
/// Proptest companion to [`test_timestamp_plus_interval_dynamic_monotone`]:
/// the dynamic-monotonicity handler in [`SpecialBinary`] claims that
/// `add_timestamp_interval(t, i)` is monotone in `t` whenever `i.months == 0`
/// (the only case it actually claims monotonicity for at runtime: the
/// matches above require the right argument to be a single value with
/// `months == 0`). This proptest verifies that claim directly against the
/// function impl by sampling random timestamps and zero-month intervals
/// and checking that input ordering is preserved in the output.
#[mz_ore::test]
#[cfg_attr(miri, ignore)]
fn proptest_timestamp_plus_interval_monotone_when_months_zero() {
    use mz_repr::adt::interval::Interval;
    use mz_repr::{Datum, RowArena, SqlScalarType, arb_datum_for_scalar};
    use proptest::prelude::*;

    let timestamp_strat = arb_datum_for_scalar(SqlScalarType::Timestamp { precision: None });
    // Lex order on `Interval` does *not* match total-microseconds order when
    // both days and micros vary independently (e.g. `{0, 0, 86_400_000_001}`
    // is lex-less than `{0, 1, 0}` but evaluates to a strictly larger
    // timestamp), so we only claim monotonicity for *fixed* zero-month
    // intervals — which is exactly what the DynamicMonotone handler does.
    // The proptest accordingly varies `t` with `i` held constant.
    let zero_month_interval_strat = (any::<i32>(), any::<i64>())
        .prop_map(|(days, micros)| Interval { months: 0, days, micros });

    let expr = MirScalarExpr::CallBinary {
        func: AddTimestampInterval.into(),
        expr1: Box::new(MirScalarExpr::column(0)),
        expr2: Box::new(MirScalarExpr::column(1)),
    };
    let arena = RowArena::new();

    proptest!(|(
        t1 in timestamp_strat.clone(),
        t2 in timestamp_strat,
        i in zero_month_interval_strat,
    )| {
        let t1 = match t1 { PropDatum::Timestamp(t) => t, _ => unreachable!() };
        let t2 = match t2 { PropDatum::Timestamp(t) => t, _ => unreachable!() };
        let i = Datum::Interval(i);
        let r1 = expr.eval(&[Datum::Timestamp(t1), i], &arena);
        let r2 = expr.eval(&[Datum::Timestamp(t2), i], &arena);
        // Only compare when both calls succeed; the monotonicity claim
        // applies only within the success domain.
        if let (Ok(Datum::Timestamp(r1)), Ok(Datum::Timestamp(r2))) = (r1, r2) {
            prop_assert_eq!(t1.cmp(&t2), r1.cmp(&r2));
        }
    });
}
```

A couple of caveats worth noting before this lands:

1. The interval generator uses `any::<i32>()` and `any::<i64>()` directly; `Interval::duration_as_chrono` calls `Duration::try_days(...).unwrap()`, which is fine for `i32::MAX` days (~5.8M years, well below chrono's i64-ms ceiling) but the subsequent `+ Duration::microseconds(...)` could overflow at the extremes. Both function returns and panics are observable as test failures so we'll catch it; if it's flaky we may need to bound the strategy.
2. Only `add_timestamp_interval` is exercised. `sub_*` and the `_tz` variants share the same `add_timestamplike_interval` / `sub_timestamplike_interval` implementations, so the property transfers, but if we want belt-and-braces I can fan out to a parameterized version that runs all four `BinaryFunc::{Add,Sub}{Timestamp,TimestampTz}Interval` variants.

---

src/expr/src/interpret.rs
```diff
@@ -598,14 +598,28 @@ impl SpecialUnary {
...
 /// A binary function we've added special-case handling for; including:
-/// - A two-argument function, taking and returning [ResultSpec]s. This overrides the
-///   default function-handling logic entirely.
+/// - Either a complete override of [ResultSpec] computation, or a way to
+///   compute monotonicity dynamically from the input specs.
 /// - Metadata on whether / not this function is pushdownable. See [Trace].
 struct SpecialBinary {
-    map_fn: for<'a> fn(ResultSpec<'a>, ResultSpec<'a>) -> ResultSpec<'a>,
+    handler: SpecialBinaryHandler,
     pushdownable: (bool, bool),
```
mgree, line 637: Not your code, but would be good to say the bools correspond to which arguments are pushdown-able. Took me a minute to reconstruct.

src/expr/src/interpret.rs
```diff
@@ -598,14 +598,28 @@ impl SpecialUnary {
...
 }
 
+/// How a [SpecialBinary] computes the output [ResultSpec].
+enum SpecialBinaryHandler {
+    /// Completely override the spec computation; the default flat-map machinery
```
mgree, line 642: I would call this "AbstractFunc", i.e., this is the function in the abstract domain that corresponds to some binary function.

src/expr/src/interpret.rs
```diff
@@ -688,19 +702,56 @@ impl SpecialBinary {
...
+        /// annotation has to pick the conservative answer; this dynamic check
+        /// recovers filter pushdown for the common case of literal
+        /// `INTERVAL '<N>' day`-style predicates.
+        fn timestamp_plus_interval_monotone(
+            _left: &ResultSpec<'_>,
+            right: &ResultSpec<'_>,
+        ) -> (bool, bool) {
+            let months_zero = matches!(
+                &right.values,
+                Values::Within(Datum::Interval(a), Datum::Interval(b))
```
mgree: Comparing against a particular `Values::` value feels brittle... what if we add the discrete sets? Add a predicate to `Values`?

src/expr/src/interpret.rs
```diff
@@ -688,19 +702,56 @@ impl SpecialBinary {
...
+                    if a == b && a.months == 0,
+            );
+            if months_zero {
+                (true, true)
+            } else {
+                (false, false)
+            }
```
mgree:
```suggestion
            (months_zero, months_zero)
```

src/expr/src/interpret.rs
```diff
@@ -1836,6 +1899,193 @@ mod tests {
...
 
+    /// Companion test to `test_add_timestamp_interval_non_monotone`: when the
+    /// interval argument is a literal with `months == 0`, the function reduces
+    /// to a pure linear shift in microseconds and *is* monotone in the
+    /// timestamp. The dynamic-monotonicity handler in `SpecialBinary` should
+    /// recover the tight output range in that case, so that filter pushdown
+    /// can still narrow predicates like `t - INTERVAL '1' day < literal`.
+    #[mz_ore::test]
+    #[cfg_attr(miri, ignore)]
+    fn test_timestamp_plus_interval_dynamic_monotone() {
```
mgree, line 2210: Can we add an SLT (or other end-to-end test) to make sure that we're actually pushing things down?

src/expr/src/interpret.rs
```diff
@@ -598,14 +598,28 @@ impl SpecialUnary {
...
+    handler: SpecialBinaryHandler,
     pushdownable: (bool, bool),
 }
 
+/// How a [SpecialBinary] computes the output [ResultSpec].
+enum SpecialBinaryHandler {
+    /// Completely override the spec computation; the default flat-map machinery
+    /// is bypassed.
+    Override(for<'a> fn(ResultSpec<'a>, ResultSpec<'a>) -> ResultSpec<'a>),
+    /// Use the default flat-map machinery, but with a monotonicity verdict that
```
mgree, line 645: I don't know if we'll ever have a function that both (a) wants a special abstraction function and (b) has dynamically determined monotonicity properties. If so, this arrangement doesn't quite cut it. Would be an easy enough refactor when that arises to have a struct with both of these as optional fields.

mgree (approved):
Approach seems fine to me! Some thoughts on naming and clarity, and a test suggestion.

antiguru:
Pushed `43314aa91a` rebasing on main and addressing the review:

- **`pushdownable: (bool, bool)`**: added a `(left, right)` doc comment.
- **`SpecialBinary` → `AbstractFunc`**: renamed, plus `SpecialBinaryHandler` → `AbstractFuncHandler`.
- **Brittle `Values::Within(Datum::Interval(..), ..)` match**: added `Values::as_single() -> Option<Datum>` and use it. Future variants of `Values` (e.g. a discrete-set representation) will degrade to "not a single value" rather than silently mis-matching.
- **`if months_zero { (true, true) } else { (false, false) }`**: collapsed to `(months_zero, months_zero)`.
- **End-to-end pushdown test**: added one to `test/sqllogictest/filter-pushdown.slt` next to the existing `t < literal` test, exercising `t - INTERVAL '1' day < literal` and asserting the `pushdown=` line is emitted.
- **Override + DynamicMonotone coexistence**: added a doc note on `AbstractFunc` flagging that today a function has *either* a handler override or a dynamic-monotonicity verdict; when a future function wants both, the suggested refactor is to promote `AbstractFuncHandler` from an enum to a struct with two `Option` fields. Leaving the enum form until that day arrives.

Skipped local compile/test (disk in the dev container was full) — letting CI exercise the proptest and the new SLT.

---

antiguru:
Thanks for the review!

PR #36729, opened 2026-05-26
adapter: extract begin_statement_logging helper from frontend_peek

aljoscha:
The new statement and reused-outer-context branches in `try_frontend_peek` are otherwise identical bookkeeping. Lift them into a `PeekClient::begin_statement_logging` method so that another frontend-sequenced caller (the upcoming OCC read-then-write path) can share the same logic.


ggevay (approved):
LGTM

PR #36731, opened 2026-05-26
mcp: skip auto-route when role lacks USAGE on the data product's cluster

bobbyiliev:
Follow-up to https://github.com/MaterializeInc/materialize/pull/36619#issuecomment-4540116380

### DIFF
--- a/src/environmentd/src/http/mcp.rs
+++ b/src/environmentd/src/http/mcp.rs
@@ -84,6 +84,8 @@ enum McpRequestError {
     ToolNotFound(String),
     #[error("Data product not found: {0}")]
     DataProductNotFound(String),
+    #[error("{0}")]
+    ClusterPrivilegeMissing(String),
     #[error("Query validation failed: {0}")]
     QueryValidationFailed(String),
     #[error("Query execution failed: {0}")]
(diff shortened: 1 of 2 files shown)


src/environmentd/src/http/mcp.rs
```diff
@@ -1011,28 +1011,63 @@ async fn read_data_product(
...
+    let has_cluster_usage: bool = lookup_row
+        .and_then(|row| row.get(1))
+        .and_then(|v| v.as_bool())
+        .unwrap_or(false);
+
+    // Override beats everything. Otherwise auto-route to the catalog
+    // cluster only when the role has `USAGE` on it; without `USAGE` the
+    // `SET CLUSTER` would succeed but the subsequent `SELECT` would fail
+    // with a `permission denied`, so we leave the read on the session
+    // default — slower (no index) but correct.
```
ggevay: > so we leave the read on the session
>    // default — slower (no index) but correct

I'm a bit worried that this implicit fallback would lead to silent RBAC problems, where the symptom is just slow queries, but people (and agents) wouldn't realize that there is an RBAC problem. So, maybe we'd want to fail the call in this case? And then if an agent really wants to execute on a cluster that has no index, then it could still do it with an explicit override.

def- (approved):
Thanks Bobby, test change looks reasonable

src/environmentd/src/http/mcp.rs
bobbyiliev: Good call @ggevay, done. It now fails with a clear `ClusterPrivilegeMissing` error naming the cluster, and the agent can pass an explicit `cluster` override to read elsewhere if it really wants to.

src/environmentd/src/http/mcp.rs
ggevay: Stale?

ggevay (approved):
LGTM, just a minor comment left.

PR #36732, opened 2026-05-26
pgcopy,pgwire: fix CSV decoding of quoted NULL markers and end-of-copy marker

def-:
The csv crate strips quoting during parsing, so `decode_copy_format_csv` could not distinguish a quoted NULL marker from an unquoted one (silent data corruption: e.g. with default params, a literal `""` decoded to SQL NULL instead of the empty string), and the end-of-copy marker check treated a quoted `"\."` value as the bare `\.` terminator (silent data loss: the row and every subsequent row were dropped).

Switch the decoder to csv-core so we can inspect each field's leading input byte and recover per-field quote state; the NULL-marker and end-of-copy checks both now require the field to be unquoted. The pgwire COPY row scanner had the same quote-blind end-marker check; switch it to match against the raw input bytes for the in-progress record.

### DIFF
--- a/src/pgcopy/src/copy.rs
+++ b/src/pgcopy/src/copy.rs
@@ -11,7 +11,6 @@ use std::borrow::Cow;
 use std::io;
 
 use bytes::BytesMut;
-use csv::{ByteRecord, ReaderBuilder};
 use itertools::Itertools;
 use mz_repr::{
     Datum, RelationDesc, Row, RowArena, RowRef, SharedRow, SqlColumnType, SqlRelationType,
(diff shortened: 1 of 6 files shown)


src/pgcopy/src/copy.rs
```diff
@@ -690,73 +689,143 @@ pub fn decode_copy_format_csv(
...
+    // We use csv-core (rather than the higher-level csv crate) so we can
+    // recover per-field "was this field quoted?" information by inspecting the
+    // first byte of each field's input. csv unquotes during parsing, which
+    // makes a quoted empty string indistinguishable from an unquoted empty
+    // field — and PostgreSQL COPY ... FORMAT CSV semantics need that
+    // distinction to honor the NULL marker.
+    let mut input = data;
+    let mut output = vec![0u8; data.len().max(1024)];
+    let mut out_pos = 0;
+    let mut fields: Vec<(usize, usize, bool)> = Vec::new();
```
ggevay: Could you please introduce a struct instead of using a 3-tuple?

src/pgcopy/src/copy.rs
```diff
@@ -690,73 +689,143 @@ pub fn decode_copy_format_csv(
...
+                    out_pos = 0;
                 }
             }
+            csv_core::ReadFieldResult::OutputFull => {
+                let new_len = output.len().saturating_mul(2).max(out_pos + 1);
+                output.resize(new_len, 0);
+            }
+            csv_core::ReadFieldResult::InputEmpty => {
+                // We've consumed all input; loop again with the now-empty
+                // slice so csv-core can flush any buffered trailing field.
```
ggevay: At first I misunderstood this comment as `ReadFieldResult::InputEmpty` meaning "We've consumed all input". Could be clarified to something like:
```
    // InputEmpty means csv-core consumed all our input mid-field and wants more.
    // We have no more to give, so the next iteration calls it with the now-empty
    // slice; per its documented termination protocol, that yields the
    // partial field (if any) as a final `Field`, then `End`.
```

src/pgwire/src/protocol.rs
```diff
@@ -3364,12 +3385,16 @@ impl CopyRowScanner {
         self.end_marker_end = self
             .end_marker_end
             .and_then(|end| end.checked_sub(split_pos));
+        // Splits always occur at a completed-row boundary, so the in-progress
+        // record (if any) starts at the new beginning of the buffer.
```
ggevay: Maybe worth recording the invariant also in a soft_assert. (Otherwise, the `saturating_sub` would silently paper over a violation.) Something like:
```
soft_assert_or_log!(
    self.record_start >= split_pos,
    "split bisected an in-progress CSV record: record_start={} < split_pos={}",
    self.record_start,
    split_pos,
);
```

test/pgtest-mz/copy-from-csv.pt
```diff
@@ -381,7 +382,133 @@ DataRow {"fields":["user:materialize","1","one"]}
...
+CopyIn {"format":"text","column_formats":["text","text"]}
+CommandComplete {"tag":"COPY 4"}
+ReadyForQuery {"status":"I"}
+RowDescription {"fields":[{"name":"a"},{"name":"b"}]}
+DataRow {"fields":["","c"]}
+DataRow {"fields":["a","NULL"]}
+DataRow {"fields":["b",""]}
+DataRow {"fields":["d","\\."]}
+CommandComplete {"tag":"SELECT 4"}
+ReadyForQuery {"status":"I"}
```
ggevay, line 514: One more case to test:
```

# Obscure edge case: the NULL marker is itself set to `\.`, colliding with the
# end-of-copy marker's byte sequence. The two are still distinguished:
#   - An unquoted `\.` *field* matches the NULL marker => SQL NULL.
#   - A quoted `"\."` field is the literal string `\.` (quoting suppresses
#     both NULL-marker and end-marker interpretation).
#   - A bare `\.` on its own line (a single-field record) is the end-of-copy
#     marker regardless of the NULL setting, and subsequent rows are dropped.
send
Query {"query": "CREATE TABLE t9 (i INT8, t TEXT)"}
----

until
ReadyForQuery
----
CommandComplete {"tag":"CREATE TABLE"}
ReadyForQuery {"status":"I"}

send
Query {"query": "COPY t9 FROM STDIN WITH (FORMAT CSV, NULL '\\.')"}
CopyData "1,\\.\n"
CopyData "2,\"\\.\"\n"
CopyData "3,after\n"
CopyData "\\.\n"
CopyData "4,ignored\n"
CopyDone
Query {"query": "SELECT i, t FROM t9 ORDER BY i"}
----

until
ReadyForQuery
ReadyForQuery
----
CopyIn {"format":"text","column_formats":["text","text"]}
CommandComplete {"tag":"COPY 3"}
ReadyForQuery {"status":"I"}
RowDescription {"fields":[{"name":"i"},{"name":"t"}]}
DataRow {"fields":["1","NULL"]}
DataRow {"fields":["2","\\."]}
DataRow {"fields":["3","after"]}
CommandComplete {"tag":"SELECT 3"}
ReadyForQuery {"status":"I"}
```

ggevay (approved):
LGTM, minor comments

PR #22753, opened 2023-10-27
Collapse `generate_series` when output columns are projected away

frankmcsherry:
This PR updates projection pushdown for `FlatMap` nodes to notice example cases where output columns are unused, and the number of records can be determined from the data. This is essentially only in support of optimizing `generate_series`, which has historically forced complete unpacking for something like
```sql
select count(*) from generate_series(1, 1000000);
```
This now optimizes to the constant `1000000` rather than a dataflow that deploys to build one million rows and count them.

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered.
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] This PR includes the following [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note):
  -

PR #36735, opened 2026-05-26
Rowwise subquery eval

frankmcsherry:
## Summary

A lowering strategy for correlated subqueries that reference only the current row: instead of the stateful keyed `Reduce` joined back to the outer relation, the subquery is housed in a new `TableFunc::EvalRelation` and driven row-by-row by a stateless `FlatMap`. The housed `MirRelationExpr` is interpreted per input row by `mz_expr::relation::eval` (a closed-expression evaluator factored out of `FoldConstants`, whose per-operator helpers it now shares), with the input row bound to a distinguished placeholder local.

Gated behind `enable_rowwise_subquery_lowering`, default off; disposition is a trial — disable or remove absent a positive response.

> **Not mergeable as-is.** Rebased onto current `main` and green everywhere except the stable-LIR schema tests, which this design conflicts with. See [Blocker](#blocker-evalrelation-vs-the-stable-lir-schema) below — that question wants an answer before the trial can start.

## Design notes

- **Eligibility.** A cheap HIR pre-filter (`is_rowwise_candidate`) rejects subqueries that reference stored collections; the authoritative check (`is_rowwise_closed`) runs on the trial-decorrelated MIR and requires every leaf to be a constant or an internally-bound local, no recursion, no arrangements, and no unmaterializable functions (those are resolved by a later pass that cannot see into housed bodies). On failure the lowering falls through to the keyed path unchanged.
- **Housed bodies are optimized.** Housed relations are opaque to every other transform, so a dedicated early pass (`HousedRelationOptimization`) runs `ProjectionPushdown` inside each housed body — pruning unused columns and, critically, applying its FlatMap rewrites: a housed `count(*)` over `generate_series(1, <input col>)` collapses to a `repeat_row_non_negative` and, with the count-aware aggregate folding of #36776 (since merged to `main`), evaluates in constant time per row. This is what resolves the OoM report on this PR.
- **CSE.** Housed-relation local ids are canonically renumbered so structurally-equivalent subqueries produce byte-identical `EvalRelation`s, letting `RelationCSE` share their FlatMaps. Single-valued `Map`/`Project`-only housed bodies strength-reduce back to plain `Map`/`Project` in `FlatMapElimination`.
- **Error semantics.** The interpreter propagates evaluation errors (including in join-equivalence constraints), and the housed `guard_subquery_size` preserves the 'more than one record produced in subquery' error, pinned in the slt.

## Blocker: `EvalRelation` vs the stable LIR schema

`src/compute-types/tests/lir_schema.rs` (#37814, August) traces the serde surface reachable from `LirRelationExpr` and asserts that a named set of unstable types is *not* reachable — `MirRelationExpr` among them, because plan pinning (`ALTER CLUSTER … FREEZE`, per `doc/developer/design/20260311_optimizer_customer_tradeoff.md`) stores LIR durably in the catalog and replays it across releases.

`TableFunc` is shared by both levels: MIR's `FlatMap` holds one and so does `LirRelationExpr::FlatMap` (`plan.rs:376`). So `TableFunc::EvalRelation { relation: Box<MirRelationExpr>, .. }` puts a serialized MIR subtree inside pinned plans, and the next MIR change invalidates them — exactly the failure mode pinning exists to prevent. All three tests in that file fail (they trip first on incomplete enum tracing for `MirRelationExpr` and `Result`; the forbidden-list assertion is what fires once tracing completes).

There is no cheap exemption. `#[serde(skip)]` would drop the housed body from pinned plans; an opaque-bytes encoding trips the companion assertion that only `StableRow` and `StableEvalError` may serialize as bytes, and MIR has no proto encoding to guard with the buf lint. The registry traces types rather than values, so "plans containing `EvalRelation` are simply not pinnable" does not help either.

The fix that follows the existing pattern — `MirScalarExpr`/`AggregateExpr`/`Row`/`EvalError` are all already forbidden and all already have stable LIR twins:

- parameterize `TableFunc<R = MirRelationExpr>`, as `UnaryFunc<E = MirScalarExpr>` already is, so LIR instantiates it at a stable housed-relation type;
- build that type from `LirScalarExpr`, `LirAggregateExpr` and `StableRow`, covering exactly the fragment `eval_relation` interprets and `is_rowwise_closed` enforces, and convert at MIR→LIR lowering;
- make the four per-operator helpers generic over `E: Eval` (`src/expr/src/scalar/eval.rs:20`, implemented for both scalar types) so one interpreter still serves both constant folding and row-wise evaluation.

That is a fair amount of work, and it buys a new durable format with its own migration obligation, for a flag that is default-off and scoped as a month-long trial. Hence raising it here rather than building it speculatively: **is the stable-LIR contract meant to exclude MIR unconditionally, or is a housed interpreted relation a case for replan-on-unfreeze?**

## Testing

`test/sqllogictest/rowwise_subquery_lowering.slt` (flag on) documents which idioms house and which fall back, with plan and result assertions: unnest/dedup/re-aggregate capabilities, the stored-collection and unmaterializable fallbacks, error parity for the >1-row scalar subquery guard, and the housed-body optimization's constant-time evaluation of huge correlated series. Unit tests cover the interpreter (`relation/eval.rs`) and the housed-body optimization.

After the rebase, locally: `cargo check` clean across mz-expr/mz-transform/mz-sql/mz-adapter; 206 unit tests green in those crates; sqllogictest 352/352 over `rowwise_subquery_lowering.slt`, `table_func.slt` and main's `transform/generate_series_collapse.slt`, with no plan drift; and `rowwise_subquery_lowering.slt` clean under `--auto-index-selects`. `lir_schema`'s three tests fail, per the blocker above.

## Rebase notes (2026-09-02)

Merged `main` after three months (1130 commits). What changed in the branch:

- **#36776 (count-aware aggregates) has since landed on `main`** and evolved further there, so the copies this branch carried are dropped in favour of main's: `reduce.rs`, `projection_pushdown.rs` (plus its testdata and `.spec`), `aoc_1202.slt`, the catalog index-accounting slt, and the OID block. `main` also moved `generate_series_collapse.slt` under `test/sqllogictest/transform/`, so the stale copy at the old path is deleted rather than duplicated.
- `fold_constants.rs` keeps its deletion of the per-operator helpers, which now live in `mz_expr::relation::eval`; main's only change to them since the merge base was the 36776 count-awareness this branch had already ported there.
- **`mz-expr` no longer depends on differential dataflow** (#38170 pruned it as unused), so `eval.rs` consolidates with a small local helper rather than reintroducing the dependency.
- The retitled reduce non-positive-multiplicity message is also pinned in parallel workload's negative-accumulation allowlist, which `main` added since; updated there.
- `postgres/rowtypes.slt` is back to main's content — that was unrelated churn swept in by a blanket `--rewrite-results`.

## Known gaps / follow-ups

- Guard-bearing scalar subqueries over derived tables and FROM items currently fall back to the keyed path: HIR plans FROM clauses as `Join`s rooted at unit constants with local bindings, which the candidate pre-filter rejects. Loosening it is recorded as a follow-up (investigated; left out pending attribution of one anomaly).
- `EXPLAIN` renders housed relations as an opaque `eval_relation`; a verbose-mode rendering would help evaluation of the flag.
- `EvalRelation::eval` interprets with no intermediate size limit; the housed-body optimization addresses the known pathological case, but a backstop limit may still be prudent.


def- (commented):
This query:
```sql
SELECT (SELECT count(*) FROM generate_series(1, n)) AS c FROM (VALUES (10000000000)) AS t(n);
```
used to run on clusterd with fixed memory, with this PR it runs on environmentd and OoMs it. Even after a timeout and getting cancelled it just keeps running:
```
materialize=> SELECT (SELECT count(*) FROM generate_series(1, n)) AS c FROM (VALUES (10000000000)) AS t(n);
ERROR:  canceling statement due to statement timeout
HINT:  Consider increasing the maximum allowed statement duration for this session by setting the statement_timeout session variable. For example, `SET statement_timeout = '120s'`.
```

frankmcsherry:
Makes sense. The dataflow evaluator is better optimized than the scalar evaluator (which is mostly programmed to bail out when it sees lots of data, rather than perform the work efficiently). We were unfortunately unable to merge https://github.com/MaterializeInc/materialize/pull/22753 which is the fix for the demo query, but tbh it's mostly cosmetic (in that, no one does this outside of tests, and the PR fixes behavior that only happens in tests). If you actually want to produce 10B records from one row .. probably worth imagining that you might have 10B records from that row. If it were `array_agg` rather than `COUNT(*)` of course it wouldn't work.

I'm checking out what it would take to fix the underlying issue (at the time it was "rewrite all of our aggregation operators, because they are implemented 'in unary'"). I'm also ok with "this doesn't work great, but folks get all the memory savings on the queries they are issuing."

The *tl;dr* is that we have two implementations of `COUNT(*)` in MZ: one in `fold_constants` that uses the SQL implementation based on row iterators, and on in DD that just adds up numbers. The former is much worse, but we don't have DD in scalar evaluation. =/

frankmcsherry:
Ah, and fwiw it may run in finite memory (seems to) but also has taken about 10m to run so far on a 100cc, and still going. I'm not sure either of these are good. :D. (edit: it took 20m).

frankmcsherry:
Lucky you! I am here to break all of your tests. :D

https://github.com/MaterializeInc/materialize/pull/36776

test/sqllogictest/postgres/rowtypes.slt
```diff
@@ -438,19 +438,19 @@ create type testtype3 as (a int, b text);
...
 
 statement ok
 create type testtype5 as (a int);
 
-query error no overload
+query error operator does not exist: testtype1 < testtype5
 select row(1, 2)::testtype1 < row(1)::testtype5;
 
-query error no overload
+query error operator does not exist: testtype1 <> testtype5
```
frankmcsherry: No idea what these are about, or why this PR picked them up.

frankmcsherry: From 🤖 :
```
  These aren't related to the subquery-lowering work — they got swept in by the
  blanket --rewrite-results in the "Update SLTs" commit.
  
  The 4 lines flip the expected error for comparing two distinct composite types
  (e.g. testtype1 < testtype3) from no overload to operator does not exist: 
  testtype1 < testtype3. That message change comes from the older
  operator-resolution / BinaryFunc refactor that's been on main for a while;
  src/sql/src/func.rs is byte-identical between this branch and origin/main, so
  none of it is our doing.

  main still has the old no overload text, but that's actually stale — it fails
  against the current binary. It just never gets noticed because
  test/sqllogictest/postgres/*.slt runs with --no-fail and doesn't gate CI. So the
   new text here is simply the correct expectation; keeping it is fine (and
  slightly better than main).
```

PR #36736, opened 2026-05-26
expr: fix Values::union for Nested specs with disjoint keys

DAlperin:
## Summary

- Fix a persist filter-pushdown wrong-results bug ([PER-6](https://linear.app/materializeinc/issue/PER-6/persist-filter-pushdown-returns-wrong-results-for-case-expressions)) where a query that used CASE to pick between two JSON columns and then extracted a key only present in one branch was either silently returning the wrong count, or panicking under audit (`persist filter pushdown correctness violation!`).
- Root cause: `Values::union` for two `Nested` specs in `src/expr/src/interpret.rs` only iterated the left-hand map and silently discarded right-hand-only keys, while keeping left-hand-only keys verbatim. The semantics of `Nested(map)` is that keys absent from `map` are unconstrained; the union of two Nested specs must therefore drop any key missing from one side, because the missing side contributes `anything` at that key and `x ∪ anything = anything`. The buggy code instead produced an *under*-approximation, which downstream `->>` extraction turned into a non-nullable spec, causing `IS NULL` predicates to be folded away to `false` and parts to be pruned that contained matching rows.
- Rewrite the union to keep only keys present in both sides (per-key unioned), dropping single-side keys to fall back to the Nested default. This is the dual of how the existing `intersect` implementation handles the same asymmetry.

## Test plan

- [x] `cargo nextest run -p mz-expr interpret` — new `test_nested_union_partial_overlap` and `test_case_over_jsonb_columns` pass; pre-existing `test_equivalence` proptest still passes (the fix only makes the abstract result a strictly looser over-approximation, which preserves the correctness invariant).
- [x] Regression test added to `test/sqllogictest/explain/pushdown.slt` asserting the query from the issue returns the expected count of 3.
- [ ] CI: full sqllogictest run on the regression case.

### DIFF
--- a/src/expr/src/interpret.rs
+++ b/src/expr/src/interpret.rs
@@ -56,17 +56,24 @@ impl<'a> Values<'a> {
             (Values::Within(a0, a1), Values::Within(b0, b1)) => {
                 Values::Within(a0.min(b0), a1.max(b1))
             }
-            (Values::Nested(mut a), Values::Nested(mut b)) => {
-                a.retain(|datum, values| {
-                    if let Some(other_values) = b.remove(datum) {
-                        *values = values.clone().union(other_values);
+            (Values::Nested(a), Values::Nested(mut b)) => {
+                // `Nested(map)` treats keys missing from `map` as fully unconstrained, so a
+                // key present in only one side of the union must be treated as `anything`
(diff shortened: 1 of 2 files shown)


antiguru (approved):
Looks great, thank you!

PR #36737, opened 2026-05-26
refactor: Use BTreeSet::last() and ::first() to compute max (logarithmic instead of linear)

mgree:
### Motivation

Noticed a few places in the code that call `.iter().max()` and `.iter().min()` on a `BTreeSet`---this is linear. `BTreeSet`s are already neatly ordered, with `.first()` returning the minimum element and `last()` returning the maximum.

### Description

Refactored several calls.

One of the calls, on closer inspection, could entirely be replaced by a call to `itertools`'s `.minmax()`, which should save us the allocation of a `BTreeSet` entirely.

### Verification

CI should be green. I can't imagine this actually moves the needle (the sets are typically <100 elements), so not doing further measurement.

### DIFF
--- a/src/compute/src/render/context.rs
+++ b/src/compute/src/render/context.rs
@@ -921,7 +921,7 @@ impl<'scope, T: RenderTimestamp> CollectionBundle<'scope, T> {
             return self.as_specific_collection(key.as_deref(), config_set);
         }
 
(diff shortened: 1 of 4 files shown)


antiguru (approved):
Looks good, thanks!

PR #36740, opened 2026-05-26
persist: prevent duplicate rollup re-insertion in `add_rollup`

DAlperin:
## Summary

Fixes the `gc.rs:569` panic tracked in [PER-16](https://linear.app/materializeinc/issue/PER-16/panic-at-gcrs569-duplicate-rollup-insertion-within-a-single-gc-run) by tightening `StateCollections::add_rollup` rather than softening the GC assertion.

`add_rollup` previously only checked `self.rollups.get(&rollup_seqno)`. The `apply_unbatched_idempotent_cmd` retry path captures `(rollup_seqno, rollup)` and replays the closure on each retry. If the original CaS attempt actually succeeded but returned an indeterminate error, and a concurrent GC then removed the just-committed rollup map entry before the retry runs, the retry observes `None` and re-inserts the same `PartialRollupKey` under a new state seqno. That produces a second `Insert` for the same key in the diff stream and — once both entries are later GC'd — a second `Delete` for the same key. A subsequent `find_removable_blobs` walks both deletes in a single pass and trips `assert!(rollups_to_delete.insert(...))`, SIGABRTing the persist worker.

Fix it at the source: the `None` arm now refuses the insert when the rollup's key is already referenced anywhere in `state.rollups`, returning `Continue(false)` so callers treat it as a duplicate write (and delete the orphaned blob, as they already do for same-seqno collisions). This makes the GC-side invariant ("at most one Delete per `PartialRollupKey` per pass") naturally hold, leaving the existing hard assert in place as a meaningful tripwire for any future regression.

## Tests

* New `state.rs` unit test `add_rollup_rejects_duplicate_key` covering fresh insert, same-seqno idempotent retry, and the new duplicate-key-different-seqno rejection path.
* Existing `gc_rollups` / `regression_gc_behind` datadriven coverage continues to exercise the happy path.

## Test plan

- [x] `cargo check -p mz-persist-client --tests`
- [x] `cargo clippy --all-targets -p mz-persist-client -- -D warnings`
- [x] `bin/fmt`
- [x] `cargo test -p mz-persist-client --lib add_rollup_rejects_duplicate_key`
- [x] `cargo test -p mz-persist-client --lib datadriven` (covers `gc` / `gc_rollups` / `regression_gc_behind`)

Fixes PER-16

### DIFF
--- a/src/persist-client/src/internal/state.rs
+++ b/src/persist-client/src/internal/state.rs
@@ -1408,6 +1408,31 @@ where
         let applied = match self.rollups.get(&rollup_seqno) {
             Some(x) => x.key == rollup.key,
             None => {
+                // PER-16: refuse to insert a rollup at a seqno that GC has
+                // already physically removed. `apply_unbatched_idempotent_cmd`
+                // replays the captured `(rollup_seqno, rollup)` tuple on
+                // every retry. If the original CAS committed indeterminately
+                // and a concurrent GC then removed the rollup, the retry
+                // would otherwise observe an empty map entry and re-insert
+                // the same `PartialRollupKey` at the same seqno, producing
+                // a second Insert (and, later, a second Delete) for that
+                // key in the diff stream and tripping the `assert!` in
+                // `find_removable_blobs`.
+                //
+                // We use the smallest seqno currently in `self.rollups` as
+                // the GC watermark. GC's `remove_rollups` always keeps the
+                // latest rollup `<= seqno_since` and removes the older
+                // ones, so the surviving minimum is strictly above every
+                // seqno that has been removed. (Unlike `last_gc_req`, this
+                // is only advanced by an actual physical removal, not by
+                // `maybe_gc` deciding to *request* one — `last_gc_req` runs
+                // ahead of removals and would falsely reject legitimate
+                // late `add_rollup` calls for older seqnos.)
+                if let Some(min_kept) = self.rollups.keys().next() {
+                    if rollup_seqno < *min_kept {
+                        return Continue(false);
+                    }
+                }
                 self.active_rollup = None;
                 self.rollups.insert(rollup_seqno, rollup.to_owned());
                 true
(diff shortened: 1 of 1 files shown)


src/persist-client/src/internal/state.rs
```diff
@@ -1408,6 +1408,16 @@ where
...
             None => {
+                // PER-16: refuse to insert a rollup whose key is already
+                // referenced anywhere in self.rollups. Without this check,
+                // the apply_unbatched_idempotent_cmd retry path for
+                // add_rollup can re-insert a rollup that was committed in a
+                // prior attempt and subsequently removed by a concurrent
+                // GC, producing duplicate Insert events in the diff stream
+                // and (later) duplicate Delete events that trip gc.rs:569.
+                if self.rollups.values().any(|r| r.key == rollup.key) {
+                    return Continue(false);
```
mtabebe: 👍

src/persist-client/src/internal/state.rs
```diff
@@ -4423,4 +4433,70 @@ pub(crate) mod tests {
...
+    /// Regression for PER-16. `add_rollup` must refuse to insert a rollup
+    /// whose `PartialRollupKey` is already referenced elsewhere in
+    /// `state.rollups`. Without this check, an `apply_unbatched_idempotent_cmd`
+    /// retry of `add_rollup` that runs after a concurrent GC removed the
+    /// rollup's original map entry will re-insert it under the same key,
+    /// producing duplicate Insert events in the diff stream and (later)
+    /// duplicate Delete events that trip the `assert!` in `gc.rs`'s
+    /// `find_removable_blobs`.
+    #[mz_ore::test]
+    fn add_rollup_rejects_duplicate_key() {
```
mtabebe: This is a really nice test

src/persist-client/src/internal/state.rs
```diff
@@ -1408,6 +1408,16 @@ where
         let applied = match self.rollups.get(&rollup_seqno) {
             Some(x) => x.key == rollup.key,
             None => {
+                // PER-16: refuse to insert a rollup whose key is already
```
mtabebe: Not related to the change but thanks for the detailed bug in linear, reading through it helped me understand the invariants and how they were being violated

mtabebe (approved):

antiguru (approved):
Looks great, thank you!

def- (commented):
I don't think this is a complete fix:
```diff
diff --git a/src/persist-client/src/internal/state.rs b/src/persist-client/src/internal/state.rs
index 0a5641adbb..acf7317ba1 100644
--- a/src/persist-client/src/internal/state.rs
+++ b/src/persist-client/src/internal/state.rs
@@ -4499,4 +4499,45 @@ pub(crate) mod tests {
         assert_eq!(state.collections.rollups.get(&other_seqno), Some(&other));
         assert!(!state.collections.rollups.contains_key(&new_seqno));
     }
+
+    /// Regression for PER-16, modeling the *actual* retry interleaving the
+    /// current-state-only duplicate guard in `add_rollup` does not cover.
+    ///
+    /// `Machine::add_rollup` replays the same captured `(seqno, rollup)` tuple
+    /// on every `apply_unbatched_idempotent_cmd` retry. If the first CAS
+    /// committed indeterminately and a concurrent GC's `remove_rollups` then
+    /// deleted the rollup's only reference, the retry must still be refused:
+    /// re-inserting the same key recreates the Insert/Delete pattern that trips
+    /// the `assert!` in `gc.rs`'s `find_removable_blobs`. Unlike
+    /// `add_rollup_rejects_duplicate_key`, this reaches the bug state purely via
+    /// `add_rollup` -> `remove_rollups` -> `add_rollup`, with no pre-seeded
+    /// duplicate-key map.
+    #[mz_ore::test]
+    fn add_rollup_idempotent_across_gc_removal() {
+        let mut state = TypedState::<String, String, u64, i64>::new(
+            DUMMY_BUILD_INFO.semver_version(),
+            ShardId::new(),
+            "".to_owned(),
+            0,
+        );
+
+        let seqno = SeqNo(10);
+        let rollup = HollowRollup {
+            key: PartialRollupKey::new(seqno, &RollupId::new()),
+            encoded_size_bytes: None,
+        };
+        let add = |state: &mut StateCollections<u64>| state.add_rollup((seqno, &rollup));
+
+        // First attempt commits the rollup; a concurrent GC then removes it.
+        assert_eq!(add(&mut state.collections), Continue(true));
+        let _ = state
+            .collections
+            .remove_rollups(&[(seqno, rollup.key.clone())]);
+        assert!(state.collections.rollups.is_empty());
+
+        // The retry replays the captured tuple. Even though no entry references
+        // the key anymore, re-inserting it must be refused.
+        assert_eq!(add(&mut state.collections), Continue(false));
+        assert!(state.collections.rollups.is_empty());
+    }
 }
```
still fails:
```
thread 'internal::state::tests::add_rollup_idempotent_across_gc_removal' (287962) panicked at src/persist-client/src/internal/state.rs:4540:9:
  left: Continue(true)
 right: Continue(false)
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed::<core::ops::control_flow::ControlFlow<mz_persist_client::internal::state::NoOpStateTransition<bool>, bool>, core::ops::control_flow::ControlFlow<mz_persist_client::internal::state::NoOpStateTransition<bool>, bool>>
   4: mz_persist_client::internal::state::tests::add_rollup_idempotent_across_gc_removal::test_impl
   5: mz_persist_client::internal::state::tests::add_rollup_idempotent_across_gc_removal
   6: mz_persist_client::internal::state::tests::add_rollup_idempotent_across_gc_removal::{closure#0}
   7: <mz_persist_client::internal::state::tests::add_rollup_idempotent_across_gc_removal::{closure#0} as core::ops::function::FnOnce<()>>::call_once
```

DAlperin:
@def- good point, fixed with checking a low water mark

PR #36742, opened 2026-05-27
adapter: wrap statement logging in an RAII StatementLoggingGuard

aljoscha:
Convert `PeekClient::begin_statement_logging` (extracted in the previous change) from returning a raw `Option<StatementLoggingId>` to returning a `StatementLoggingGuard` whose `Drop` impl emits
`StatementEndedExecutionReason::Aborted` if the statement was begun but no terminal outcome was logged.

The guard exposes:
- `id()` to get the underlying `StatementLoggingId` for callers that manage the lifecycle manually (existing peek sequencing).
- `defuse()` to hand off responsibility (e.g. for streaming responses whose end is logged later by the coordinator).

`try_frontend_peek` keeps its existing manual `log_ended_execution` shape: it `defuse()`s the guard immediately and continues to drive the streaming/non-streaming branches itself. Otherwise the guard's `Drop` would race the coordinator's eventual `end_statement_execution` and panic.

This is preparation: the upcoming OCC read-then-write path will rely on the guard so that early-returns and panics during the SUBSCRIBE+retry loop don't strand a `BeganExecution` event without an `EndedExecution`.

### DIFF
--- a/src/adapter/src/peek_client.rs
+++ b/src/adapter/src/peek_client.rs
@@ -421,12 +421,14 @@ impl PeekClient {
     /// Set up statement logging for a frontend-sequenced operation.
     ///
     /// If `outer_ctx_extra` is `None`, begins a new statement execution log
-    /// entry. If `outer_ctx_extra` is `Some` (e.g. EXECUTE/FETCH), reuses the
-    /// existing logging id and retires the outer context.
+    /// entry. If `outer_ctx_extra` is `Some` (e.g. EXECUTE/FETCH), reuses and
+    /// retires the existing logging context.
     ///
-    /// The end of execution must be logged separately by the caller (via
-    /// [`Self::log_ended_execution`]) once the terminal outcome is known, or
-    /// handed off to the coordinator for streaming responses.
+    /// Returns a [`StatementLoggingGuard`]. Callers must
+    /// [`defuse`](StatementLoggingGuard::defuse) the guard when handing off
+    /// logging responsibility (or, in future, retire it explicitly on a
+    /// terminal outcome). Dropping the guard without retiring it emits an
+    /// `Aborted` end-execution event.
     pub(crate) fn begin_statement_logging(
         &self,
         session: &mut Session,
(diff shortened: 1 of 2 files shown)


ggevay:
Triggered a Nightly, because that has more statement logging tests: https://buildkite.com/materialize/nightly/builds/16609

aljoscha:
@ggevay nightly passed as well

src/adapter/src/peek_client.rs
```diff
@@ -453,12 +454,15 @@ impl PeekClient {
                 None
             }
         } else {
-            // We're executing in the context of another statement (e.g. FETCH),
-            // so take ownership of the outer context and inherit its logging id
-            // (if any). The end of execution will be logged by the caller.
```
ggevay, line 458: Was deleting this comment and the one a few lines above intentional?

src/adapter/src/peek_client.rs
```diff
@@ -546,6 +553,63 @@ impl PeekClient {
...
+#[must_use = "StatementLoggingGuard must be explicitly retired or handed off; \
+              otherwise `Drop` will log the statement as Aborted"]
+pub(crate) struct StatementLoggingGuard {
+    /// `None` if the statement was not sampled for logging.
+    id: Option<StatementLoggingId>,
+    coordinator_client: Client,
+    now: mz_ore::now::NowFn,
+}
+
+impl StatementLoggingGuard {
```
ggevay, line 584: The PR description and the commit msg mention also `retire` and `retire_with_result`, which are not in the current diff.

src/adapter/src/peek_client.rs
```diff
@@ -546,6 +553,63 @@ impl PeekClient {
...
+    /// event. Use when another component (e.g. the coordinator, for streaming
+    /// peek / subscribe responses) will log the end asynchronously.
+    pub(crate) fn defuse(mut self) {
+        self.id = None;
+    }
+
+    fn emit(&mut self, reason: statement_logging::StatementEndedExecutionReason) {
+        let Some(id) = self.id.take() else {
+            return;
+        };
```
ggevay, line 600: The caller (in `drop`) also checks this. Do we want both checks?

src/adapter/src/peek_client.rs
```diff
@@ -546,6 +553,63 @@ impl PeekClient {
     }
 }
 
+/// RAII guard owning a frontend statement-logging lifecycle.
+///
+/// Created by [`PeekClient::begin_statement_logging`]. The guard ensures that
+/// every statement for which `BeganExecution` was logged also receives a
+/// corresponding `EndedExecution`, even on early-return, panic, or mid-flight
+/// drop of the enclosing future. If the guard is dropped without being
```
ggevay: By saying "every statement", this seems to overstate its responisibilities a bit. Maybe mention that this is `defuse`d sometimes, in which case some other code will ensure logging `EndedExecution`.

ggevay (approved):
LGTM, just minor comments.

aljoscha:
thanks for the review, and good catches! I pushed a changed commit and will merge once that's green again 🙇‍♂️

PR #36745, opened 2026-05-27
workload-replay: harden workload anonymization

jasonhernandez:
## Motivation

`bin/mz-workload-anonymize` scrubs identifiers and string literals from workload captures so they can be shared. While looking at it I found several gaps where sensitive data leaks through or the output is silently corrupted. This PR fixes the cheap, high-value ones and adds a fail-loud verification backstop. It deliberately does **not** attempt the parser-level rewrite (see "Out of scope").

## What changed

**Privacy leaks**
- **Connections and sinks now have their literals anonymized.** Previously their `create_sql` only had identifiers replaced, never literals — so hostnames, usernames, regions, broker lists, bucket/path URLs, and topic names were emitted verbatim. Types and indexes also get literal anonymization now, for consistency.

**Correctness / robustness**
- **ReDoS fix:** the string-literal regex `'(?:[^']*(?:'')?)*'` had a nested quantifier (catastrophic-backtracking shape). Rewrote as the linear `'(?:[^']|'')*'`.
- **KeyError fix:** child sources looked up `child["schema"]`/`["database"]` directly in the mapping, crashing on any builtin/uncaptured name. Now uses `.get`, matching the query path.

**Footgun**
- **No more silent overwrite:** output defaulted to overwriting the input capture. Now requires an explicit `-o` (`-` for stdout) or the new `--in-place` flag.

**Fail-loud verification (`--verify`, on by default)**
- After anonymizing, re-scans the output for whole-word survivals of original identifiers and for string literals not reduced to a `literal_N` placeholder, and **refuses to write** if any are found. `--no-verify` overrides.
- Cluster `create_sql` is exempt from the literal check: its literals (SIZE, replication factor, AZs) are non-sensitive config that replay must preserve.

## Out of scope (parser-level, tracked by existing TODOs)

These require resolving SQL with the Mz parser instead of text substitution, and are left for follow-up:
- Case folding and word-boundary matching in the identifier regex.
- The global `mapping` collapsing duplicate column/table names across schemas (distinct objects can map to one name).
- Dollar-quoted strings, SQL comments, and numeric literals are still not anonymized.

The `--verify` pass is meant as the safety net for this residual risk: anything it can detect becomes a hard error rather than a silent leak.

## Testing

- `bin/fmt` clean; `ruff check` clean; `py_compile` clean.
- Manual runs on a synthetic capture: confirmed connection broker/username, sink topic, table defaults, and query predicate literals are scrubbed; confirmed the no-output-target error, `--in-place`, `--no-verify` bypass, and that `--verify` catches an intentionally-leaked literal.

### DIFF
--- a/misc/python/materialize/cli/mz_workload_anonymize.py
+++ b/misc/python/materialize/cli/mz_workload_anonymize.py
@@ -34,6 +34,68 @@ def keywords() -> set[str]:
     return result
 
 
+def _iter_sql(obj: Any, path: str = "") -> Any:
+    """Yield (location, sql) for every create_sql/sql string in the workload."""
+    if isinstance(obj, dict):
+        for key, value in obj.items():
+            child_path = f"{path}.{key}"
+            if key in ("create_sql", "sql") and isinstance(value, str):
+                yield child_path, value
+            else:
+                yield from _iter_sql(value, child_path)
+    elif isinstance(obj, list):
+        for i, value in enumerate(obj):
+            yield from _iter_sql(value, f"{path}[{i}]")
+
+
+def verify_anonymized(
+    new: dict[str, Any], mapping: dict[str, str], args: argparse.Namespace
+) -> list[str]:
+    """Best-effort scan of anonymized output for data that should have been scrubbed.
+
+    This is a backstop for the heuristic text substitution, not a proof: it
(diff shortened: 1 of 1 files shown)


jasonhernandez:

PR #36747, opened 2026-05-27
 console: fix cluster size ordering

leedqin:
Fixes [CNS-80. ](https://linear.app/materializeinc/issue/CNS-80/cluster-sizes-are-out-of-order)`ORDER BY credits_per_hour` to sort cluster replica sizes by cost ascending.

### Verification
image

### DIFF
--- a/console/src/api/materialize/cluster/__snapshots__/availableClusterSizes.test.ts.snap
+++ b/console/src/api/materialize/cluster/__snapshots__/availableClusterSizes.test.ts.snap
@@ -5,6 +5,10 @@ exports[`queries > buildAvailableClusterReplicaSizesQuery produces the expected 
   "parameters": [
     "mz_%",
   ],
-  "sql": "select "size" from "mz_cluster_replica_sizes" where "size" not like $1",
+  "sql": "select "size" from "mz_cluster_replica_sizes" where "size" not like $1 order by CASE
+          WHEN size ~ '^[0-9]+([a-zA-Z]*cc[xX]?|C)$' THEN '0_legacy'
+          WHEN size LIKE '%-%' THEN '1_' || split_part(size, '-', 1)
+          ELSE '2_tshirt'
+        END, "credits_per_hour"",
 }
 `;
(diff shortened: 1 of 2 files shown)


def- (changes requested):
I don't think that's the correct fix. We should group by kind of cluster replica size first, so that we get "25cc, 50cc, 100cc, 200c", and only after that "M.1 nano, M.1 micro, ..."

leedqin:
@def- fixed this here 
image

console/src/api/materialize/cluster/__snapshots__/availableClusterSizes.test.ts.snap
```diff
@@ -5,6 +5,12 @@ exports[`queries > buildAvailableClusterReplicaSizesQuery produces the expected
   "parameters": [
     "mz_%",
   ],
-  "sql": "select "size" from "mz_cluster_replica_sizes" where "size" not like $1",
+  "sql": "select "size" from "mz_cluster_replica_sizes" where "size" not like $1 order by 
+        CASE
+          WHEN size LIKE '%cc%' THEN '0_legacy'
+          WHEN size LIKE '%-%' THEN '1_' || split_part(size, '-', 1)
```
def-: This might lead to weird results in self-managed with custom replica names. But probably no good way around it?

def- (approved):

console/src/api/materialize/cluster/availableClusterSizes.ts
```diff
@@ -25,6 +25,16 @@ export const buildAvailableClusterReplicaSizesQuery = () => {
       .select("size")
       // We don't want to show the mz_probe size in the UI
       .where("size", "not like", "mz_%")
+      // Group sizes by family (legacy cc, t-shirt, M.1, D.1, ...) so related
+      // sizes stay contiguous, then sort by cost within each family.
+      .orderBy(
+        sql<string>`CASE
```
leedqin, line 31: @def- used regex here which I think would be more robust for self managed as well

PR #36749, opened 2026-05-27
workload-replay: add tests for the workload anonymizer

jasonhernandez:
**Third in the stack** — base `workload-anonymize-parser` (#36746), which is itself stacked on #36745. Merge those first.

## Motivation

The anonymizer (`mz_workload_anonymize.py`) had **zero automated tests** despite being a privacy tool with a heuristic core — the worst combination. Everything in #36745 and #36746 was verified by hand. This codifies those checks as regression tests.

## What's covered

A new `misc/python/materialize/cli/mz_workload_anonymize_test.py` (15 tests), colocated so the existing `pytest --doctest-modules misc/python` CI step picks it up with no pipeline changes:

- **End-to-end** anonymization of a structurally complete workload — identifiers scrubbed, anonymized names present.
- **Leak regressions** (the #36745 fixes): connection host/user, sink topic, and column-default literals are scrubbed.
- Query string-literal redaction.
- **Cluster SIZE preserved** (non-sensitive config replay needs).
- The **no-output-target error** and **`--in-place`** overwrite.
- **`--no-literals`** keeps literals while still anonymizing identifiers.
- **`verify_anonymized`**: catches surviving identifiers and literals, accepts both `'<REDACTED>'` and `'literal_N'` placeholders, and exempts cluster literals.
- **`redact_literals_via_parser`** returns `None` (fallback signal) without the binary; the regex fallback warns.

Most tests force the regex fallback (monkeypatch `_locate_redactor`) so they're deterministic regardless of whether the `mz-sql-anonymize` helper is built. One test exercises the parser path (numeric-literal redaction) and is `skipif`'d when the binary is absent — so CI (no binary) runs 14 and skips 1; locally all 15 run.

## Testing

- All 15 pass locally (`pytest`); `bin/fmt` and `ruff check` clean.
- Confirmed the file is collected by pytest's default `*_test.py` discovery.

### DIFF
--- a/misc/python/materialize/cli/mz_workload_anonymize_test.py
+++ b/misc/python/materialize/cli/mz_workload_anonymize_test.py
@@ -1,0 +1,270 @@
+# Copyright Materialize, Inc. and contributors. All rights reserved.
+#
+# Use of this software is governed by the Business Source License
+# included in the LICENSE file at the root of this repository.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0.
+
+"""Tests for the workload anonymizer (`mz-workload-anonymize`).
+
+These exercise the pure-Python anonymization logic and the regex-based literal
+fallback, which run without a built `mz-sql-anonymize` binary. One test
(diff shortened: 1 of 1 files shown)


jasonhernandez:

PR #36753, opened 2026-05-27
DEP-13: Generate documentation on prometheus metrics

SangJunBak:
Generates a YAML based our metrics in our src directory using the `syn` crate. Used to autogenerate metrics documentation.  

Docs page is hidden from the nav menu but you can view via https://preview.materialize.com/materialize/36753/manage/monitor/appendix-metrics/#top

### Motivation

Fixes https://linear.app/materializeinc/issue/DEP-13/auto-generate-metrics-documentation

### Description

diff *can't* tell a reviewer.

### Verification


def- (commented):
The QA LLM review flagged a few potential issues. I haven't looked into them closer:

## Finding 1 (MEDIUM): `subsystem:` is parsed but discarded, so the docs publish metric names that don't exist

`src/metrics-catalog/src/main.rs` parses the `subsystem` field through the
unknown-field arm (`_ => { let _: Expr = input.parse()?; }`) and throws it
away. But Prometheus composes the fully-qualified metric name as
`namespace_subsystem_name` (`Opts::subsystem` → `fq_name`), so the `name:`
field alone is not the exposed name when `subsystem` is present.

Concrete instance shipped in this PR's generated `doc/user/data/metrics.yml`:

```yaml
- name: request_duration_seconds
- name: requests_active
- name: requests_total
  source: src/environmentd/src/http/metrics.rs
```

These come from `src/environmentd/src/http/metrics.rs:41-59`, which passes
`subsystem: component` with `component = "mz_http"`
(`src/environmentd/src/lib.rs:386`). The actually exposed metrics are
`mz_http_requests_total`, `mz_http_requests_active`, and
`mz_http_request_duration_seconds`. A user searching their Prometheus for the
documented names finds nothing, and the documented names never appear on any
scrape endpoint.

**Fix:** capture `subsystem` in `MetricArgs` and render the name as
`{subsystem}_{name}`, using the existing non-literal fallback (glob to `*`)
when the subsystem is not a string literal — here that yields
`*_requests_total`, which is at least truthful under the page's stated `*`
convention.

## Finding 2 (MEDIUM): test-only metrics leak into the user-facing catalog — visible in this PR's own output

The committed `doc/user/data/metrics.yml` contains:

```yaml
- name: test_counter
  help: a counter for testing
  source: src/ore/src/metrics/tests.rs
- name: test_counter_third_party
  help: an third_party counter for testing
  source: src/ore/src/metrics/tests.rs
```

The tool's `#[cfg(test)]` filtering has two gaps that combine here:

1. `has_cfg_test` is only consulted in `visit_item_mod`, i.e. it only skips
   *inline* `#[cfg(test)] mod tests { ... }` blocks. An out-of-line test
   module (`#[cfg(test)] mod tests;` with the body in `tests.rs`) puts the
   attribute in the parent file while the walker visits `tests.rs` as an
   independent file with no attribute in sight. This is an established pattern
   in the repo (e.g. `src/catalog/src/durable/persist.rs:11`,
   `src/catalog/src/durable/upgrade.rs:64`), so the leak will recur as such
   modules gain metrics.
2. `IGNORED_DIRS` prunes only *directories* named `tests`/`benches`/`examples`;
   a *file* named `tests.rs` passes through.

The module doc states the tool visits "every `metric!` invocation outside of
`#[cfg(test)]`" — the shipped output already violates that, publishing
test fixtures as real Materialize metrics on a reference page.

**Fix:** skip files named `tests.rs` in the walker, and/or honor `cfg(test)`
attributes on all item kinds (fns, impls), not just `ItemMod`.

Related minor (LOW, latent): `has_cfg_test` matches by substring
(`tokens.to_string().contains("test")`), so a production module gated
`#[cfg(not(test))]` — or a feature whose name merely contains "test" (e.g.
`"latest"`) — would be wrongly *excluded* from the docs. No current
`metric!`-containing module hits this, but the substring match is fragile;
prefer actually inspecting the cfg predicate for a bare `test` ident.

## Finding 3 (MEDIUM): `metric!` invocations inside `macro_rules!` bodies are silently missing — entire production metric families absent

`syn`'s visitor does not descend into the token bodies of `macro_rules!`
definitions, so any `metric!` written inside a declarative helper macro is
never collected, with no warning. Two production families are missing from the
generated catalog:

- `src/metrics/src/lgalloc.rs` (`metrics_size_class!` / `map_metrics!`):
  ~25 metrics — `mz_metrics_lgalloc_allocations_total`,
  `mz_metrics_lgalloc_area_total_bytes`, `..._file_size_bytes`,
  `..._vm_mapped_bytes`, etc. The only lgalloc entry in the yml is the one
  direct `metric!` call ("The time it took to update lgalloc stats").
- `src/metrics/src/rusage.rs` (`metrics!`): the `mz_metrics_rusage_*` family.

These are real scraped metrics (lgalloc memory metrics in particular are
operationally relevant), yet the page that claims to "list the Prometheus
metrics exposed by Materialize" omits them silently. Because the omission is
deterministic, the CI drift check passes forever and nothing surfaces it.

**Fix options:** (a) special-case: also visit `ItemMacro` bodies whose path is
`macro_rules`, scanning their token streams for `metric!` groups and parsing
the body with the existing `MetricArgs` parser, rendering
`concat!(stringify!($ns), "_", stringify!($m))`-style names via the existing
`*` glob fallback; or (b) cheaper: emit a build-time warning/failing count when
a file contains the literal `metric!` more times than the visitor collected,
so omissions are at least loud instead of silent; or (c) hand-curate the two
families into the yml via a small allowlist appended by the generator.

src/metrics-catalog/src/main.rs
```diff
@@ -0,0 +1,476 @@
...
+    /// `format!("mz_persist_{}_test_metric", name)`.
+    #[mz_ore::test]
+    fn format_template_to_glob_replaces_placeholder_with_star() {
+        assert_eq!(
+            format_template_to_glob("mz_persist_{}_test_metric"),
+            "mz_persist_*_test_metric"
+        );
+    }
+
+    // Each pair of brace is escaped, so nothing globs
```
mtabebe, line 380: This series of tests is ... something. I don't fully understand the rules here

doc/user/data/metrics.yml
```diff
@@ -0,0 +1,1614 @@
...
+- name: request_duration_seconds
+  help: How long it takes for a request to complete in seconds.
+  source: src/environmentd/src/http/metrics.rs
+- name: requests_active
+  help: Number of currently active/open http requests.
+  source: src/environmentd/src/http/metrics.rs
+- name: requests_total
+  help: Total number of http requests received since process start.
+  source: src/environmentd/src/http/metrics.rs
+- name: test_counter
```
mtabebe: Claude told me that this leaked in ... 

> from src/ore/src/metrics/tests.rs. The visitor only skips mod items with #[cfg(test)], but tests.rs is wired in elsewhere (e.g. via #[cfg(test)] mod tests; in a parent), so its file-level items aren't gated when parsed standalone

val-materialize:
**Suggestion: categorize the metrics instead of one flat table**
The generated page currently renders all 358 metrics as a single alphabetical Metric | Description table. That's hard to scan, and it's dominated by one family: persist_* plus column_pager_* account for ~231 of the 358 rows (~65%), so everything else is buried.

Since the catalog is auto-generated, I'd avoid any hand-maintained grouping. The generator already records source: (the file path) for every metric — that's the natural, deterministic key to categorize on. Map source crate → category once in the generator, emit a category field into metrics.yml, and render one sub-table per category (H2/H3 headings, alphabetical within each). New metrics then sort themselves automatically; only a brand-new crate would need a one-line mapping addition.

Proposed top-level categories (by source crate / name prefix), with counts from the current output:

- Persistence (persist) — ~231 metrics. Big enough that it should be its own H2 with H3 sub-sections: blob/consensus storage, compaction, state machine & cmd I/O, column pager, audit. (Sub-section by the next name segment after persist_.)
- Compute & dataflow (compute, dataflow, arrangement, index_peek) — ~45
- Adapter / SQL coordinator (adapter, coord, catalog, optimizer, parse, linearize, determine_timestamp, group_commit, active_*) — ~30
- Sources & ingestion (mysql, postgres, kafka) — ~16
- Runtime / memory (jemalloc, memory_limiter, metrics/lgalloc, otel) — ~17
- HTTP / frontend / MCP (http, parameter_frontend, mcp, *_postgres_connpool) — ~7
- Balancer / networking (balancer) — ~7
- Authentication (auth) — ~5

src/metrics-catalog/src/main.rs
```diff
@@ -0,0 +1,476 @@
...
+use quote::ToTokens;
+use serde::Serialize;
+use syn::parse::{Parse, ParseStream};
+use syn::punctuated::Punctuated;
+use syn::visit::Visit;
+use syn::{Attribute, Expr, Ident, ItemMod, Lit, Macro, Meta, Token, braced, bracketed};
+use walkdir::WalkDir;
+
+// Known directories that we want to ignore when walking the source tree.
+static IGNORED_DIRS: &[&str] = &["tests", "benches", "examples"];
```
mtabebe: Do you need "target" in here?

mtabebe (approved):
Small comments, but generally looks good to me

src/metrics-catalog/src/main.rs
mtabebe: To be more precise, I don't understand when braces are escaped vs not and why some level of results in * vs others

SangJunBak: Fair enough! Pretty much, `format!("My name is {}", "bob")` gives back `My name is bob` but the way you escape curly braces is another curly brace. i.e. `format!("My name is {{}}", "bob")` would give back `My name is {}`. Thus you always require an odd number of brace pairs to actually substitute, which is why the glob `*` only appears then

Can make this more transparent in the comments however.

mtabebe: Right,  I understand why you want to do formatting/bracing. I just don't understand why you would want {{{{}}}} level formatting?

SangJunBak: I don't think you would, but still gotta respect the rules 😂

kay-kim (commented):
Just 
- a trivial nit (not a biggie ...but just flagging in case): We tend to use uppercase sometimes and lowercase other times.

- general question: Things like counts and such, do these ever reset?

SangJunBak:
> *  We tend to use uppercase sometimes and lowercase other times.

I think this is just an artifact of the fact that because these were meant to be internal, there was no need to polish it. This can be changed however\~

> * general question: Things like counts and such, do these ever reset?

I believe counters and histograms (bucketed counters) reset on restart. However that's why most queries will always wrap the metrics with `rate` rather than by itself, unless it's a gauge metric

val-materialize:
oops didn't mean to leave that as a reply to Dennis' comment ^

def- (commented):
What about all the mz_tokio_* metrics?

SangJunBak:
@def- Ah good catch. Accounted for here [`5a31e42` (this PR)](https://github.com/MaterializeInc/materialize/pull/36753/changes/5a31e42c2032d5c140db7b9d666e52654bba654a)

SangJunBak:
@mtabebe wanted to call out that since your approval, I've attached 3 more commits starting from https://github.com/MaterializeInc/materialize/pull/36753/changes/021e0f138440d1141aac0d3003528d10411df2c5. These mainly surround the metrics that aren't defined via the `metric!` macro. Might be worth a re-review

src/metrics-catalog/src/main.rs
```diff
@@ -214,12 +216,18 @@ fn expr_to_string(e: &Expr) -> String {
...
-    attrs.iter().any(|attr| {
-        attr.path().is_ident("cfg")
-            && matches!(&attr.meta, Meta::List(list) if list.tokens.to_string().contains("test"))
-    })
+/// Whether `attrs` mark an item as test-only, so the user-facing catalog should
+/// skip it (and everything inside it). This deliberately just checks whether any
+/// attribute mentions `test`, which catches both test-runner attributes
+/// (`test`, `mz_ore::test`, `tokio::test`, …) and `cfg(test)` gates.
+///
+/// It's a coarse over-approximation: a feature name that
```
mtabebe, line 224: 👍

src/metrics-catalog/src/main.rs
```diff
@@ -303,6 +336,15 @@ fn run() -> anyhow::Result<()> {
         });
     }
 
+    // Pull tokio metrics
+    for (name, help) in mz_ore::metrics::describe_runtime_metrics() {
+        entries.push(MetricDoc {
+            name,
+            help,
+            source: "src/ore/src/metrics.rs".to_owned(),
```
mtabebe: Can we have this be like the ignored dirs? A configurable vector of paths?

mtabebe (commented):
A minor comment

PR #36754, opened 2026-05-27
Fix timestamp overflow when applying timezone offset

antiguru:
### Motivation

Fixes CLU-93: applying a fixed timezone offset to a near-maximum timestamp could cause a panic instead of properly returning a "timestamp out of range" error.

### Description

The `timezone_timestamp` function was performing unchecked subtraction when applying a timezone offset to a timestamp. When the offset subtraction would overflow the valid timestamp range, this could panic or produce incorrect results.

The fix introduces a new `checked_sub_with_leapsecond` helper function (mirroring the existing `checked_add_with_leapsecond`) that safely performs the subtraction with proper overflow checking. This function:
- Temporarily removes the nanosecond component to avoid precision loss
- Uses `checked_sub_signed` to safely perform the subtraction
- Returns `None` if the operation would overflow
- Recovers the nanosecond component in the result

The caller now properly converts the `Option` result to an `EvalError::TimestampOutOfRange` error instead of panicking.

### Verification

Added a regression test in `timezone.slt` that verifies applying a `+15:00` offset to a near-maximum timestamp returns a "timestamp out of range" error rather than panicking. Existing timezone tests continue to pass.

https://claude.ai/code/session_019Bu679om3fCETwm3nMLmtN

### DIFF
--- a/src/expr/src/scalar/func/impls/timestamp.rs
+++ b/src/expr/src/scalar/func/impls/timestamp.rs
@@ -762,11 +762,14 @@ fn checked_sub_with_leapsecond(lhs: &NaiveDateTime, rhs: &FixedOffset) -> Option
     let nanos = lhs.nanosecond();
     let lhs = lhs.with_nanosecond(0).unwrap();
     let rhs = rhs.local_minus_utc();
-    lhs.checked_sub_signed(match chrono::Duration::try_seconds(i64::from(rhs)) {
-        Some(dur) => dur,
-        None => return None,
-    })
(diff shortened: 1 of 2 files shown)


ggevay (approved):

ggevay:
Btw. the newly added `checked_sub_with_leapsecond` needs the same fix as https://github.com/MaterializeInc/materialize/pull/36755 does for `checked_add_with_leapsecond`, right?

antiguru:
Yes, fixed. Thank you!

PR #36756, opened 2026-05-27
Fix FlatMap column references when using non-leading arrangements

antiguru:
### Motivation

Fixes database-issues#11280: `FlatMap` operations were producing wrong results when their input was only available as an arrangement on a non-leading column.

### Description

When a table is arranged on a non-leading column (e.g., column `c` in a table with columns `a, b, c`), the arrangement exposes columns in arrangement order (key columns first, then value columns) rather than logical order. The lowering code was forwarding this arrangement to `FlatMap` operations without properly permuting column references.

The fix addresses two issues:

1. **Prefer unarranged input**: When available, use the raw unarranged collection which presents columns in logical order, avoiding permutation entirely.

2. **Permute MFP references**: When an arrangement must be used, permute not only the table function argument expressions (as before) but also the MFP (Map-Filter-Project) that runs after the table function. The MFP still references input columns at positions `0..input_arity`, which must be adjusted to account for the arrangement's reordering. Table function output columns are left unchanged since they're appended in a fixed order.

The key insight is that the MFP runs after the table function and still uses logical column positions for input references, so it requires the same permutation as the table function arguments.

### Verification

Added regression test `test/sqllogictest/github-11280.slt` that covers:
- `FlatMap` with `LATERAL generate_series` on a table with an index on a non-leading column
- Multiple index configurations (on column `c` and column `b`)
- Predicates on both the table function output and input columns

The test verifies that queries return correct results regardless of which non-leading column is indexed.

https://claude.ai/code/session_01DeVqtkFQNP1dxdkSCQ5Gxu

### DIFF
--- a/src/compute-types/src/plan/lowering.rs
+++ b/src/compute-types/src/plan/lowering.rs
@@ -616,10 +616,8 @@ impl Context {
                         // positions `0..input_arity`). Table-function output columns at
                         // positions `input_arity..` are appended after the table function in
                         // a fixed order, so we leave those references alone.
-                        let perm_map: BTreeMap<usize, usize> =
-                            permutation.iter().cloned().enumerate().collect();
                         for expr in &mut exprs {
-                            expr.permute_map(&perm_map);
+                            expr.permute(permutation);
                         }
                         let input_arity = permutation.len();
                         let mfp_input_arity = mfp.input_arity;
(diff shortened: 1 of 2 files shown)


def- (approved):
Thanks for adding the test!

def-:
Sorry, I just realized that this doesn't fix https://linear.app/materializeinc/issue/CLU-70/wrong-results-from-flatmap-mfp-not-being-permuted-when-input-is fully, fixing here: https://github.com/MaterializeInc/materialize/pull/36765

PR #36760, opened 2026-05-27
console: add subscribe health atom

leedqin:
Environment health polls `SELECT mz_version()` on mz_catalog_server every 5s per tab. Once a global SUBSCRIBE is streaming, the environment is already proven reachable, so the poll is redundant.

Derive health from subscribe state (`subscribeDerivedHealthAtom`) and back the poll off to 30s while subscribes are healthy; keep 5s during bootstrap and when subscribes go down, so recovery and the crashed/blocked banner still surface.

This frees up browser connection slots that 5 second health poll check was taking. Tested against workload capture for production analytics workload. Health  checks requests on a 6 tab stress test have around 45 concurrent in flight requests in 2 mins. This change brought it down to 7-9 concurrent requests. Worse case requsest queue time shows one health request stuck at  240s behind a saturated pool. This change brought it down to 1 ms.

Fixes [CNS-83](https://linear.app/materializeinc/issue/CNS-83/fix-health-check-polling-in-the-console)

### DIFF
--- a/console/src/store/environments.ts
+++ b/console/src/store/environments.ts
@@ -341,6 +341,19 @@ export const mergeEnvironmentsWithHealth = atom(
   },
 );
 
+// Interval used while bootstrapping, before any environment is healthy.
+const BOOTSTRAP_POLL_INTERVAL_MS = 5_000;
+
+/** True once every enabled environment is healthy. */
+const allEnvironmentsHealthy = (environments?: EnvironmentsWithHealth) => {
+  const enabled = [...(environments?.values() ?? [])].filter(
+    (env): env is EnabledEnvironment => env.state === "enabled",
+  );
+  return (
+    enabled.length > 0 && enabled.every((e) => e.status.health === "healthy")
+  );
+};
+
 /**
  * Polls for region information and environment health.
  */
@@ -357,7 +370,13 @@ export const usePollEnvironmentHealth = (options: { intervalMs: number }) => {
   // eslint-disable-next-line @tanstack/query/exhaustive-deps
   useSuspenseQuery({
     queryKey: [...environmentQueryKeys.environmentHealth(cloudRegions)],
-    refetchInterval: options.intervalMs,
+    // Back off to the slower interval once healthy; a steady environment is
+    // already proven reachable, so frequent health checks just burn connection
+    // slots. Poll fast until then so bootstrap transitions surface promptly.
+    refetchInterval: (query) =>
+      allEnvironmentsHealthy(query.state.data)
+        ? options.intervalMs
+        : BOOTSTRAP_POLL_INTERVAL_MS,
     queryFn: async () => {
       return Sentry.startSpan(
         {
--- a/console/src/components/AppInitializer.tsx
+++ b/console/src/components/AppInitializer.tsx
@@ -29,7 +29,7 @@ import IncidentStatusWidget from "./IncidentStatusWidget/IncidentStatusWidget";
 /** Initializes global features, rendered after frontegg token is available if Cloud */
 const useAppInitializer = () => {
   useTrackFocus();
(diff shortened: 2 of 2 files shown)


console/src/store/environments.ts
```diff
@@ -341,13 +342,23 @@ export const mergeEnvironmentsWithHealth = atom(
...
+const HEALTHY_POLL_INTERVAL_MS = 30_000;
+
 /**
  * Polls for region information and environment health.
  */
 export const usePollEnvironmentHealth = (options: { intervalMs: number }) => {
   const mergeEnvironments = useSetAtom(mergeEnvironmentsWithHealth);
   const cloudRegions = useAtomValue(cloudRegionsSelector);
   const appConfig = useAtomValue(appConfigAtom);
+  const subscribeHealth = useAtomValue(subscribeDerivedHealthAtom);
```
SangJunBak: I do want to call out I think the region creation flow relies on polling `fetchEnvironmentsWithHealth` to transition from "booting" to "ready". I think 30 seconds is fine, but instead of relying on a subscribe, I wonder if it's just simpler to change the overall health check poll to 30s/1 minute and pass in 30_000 into `usePollEnvironmentHealth`

SangJunBak:
Left a comment. Lemme know what you think!

console/src/store/environments.ts
leedqin: I was just concerned in the event that suddenly we stop polling for 30 seconds and then the console doesn't try to reconnect in 5 seconds so the user doesn't immediately see the console being unable to connect to environmentd. Not necessarily the worst but I thought it was a slight regression so I used a subscribe atom too. 

My initial idea was to just use the subscribe atom but then in case the request is blocked, an active subscribe will not catch that unless the socket disconnects and connects again.

SangJunBak: > My initial idea was to just use the subscribe atom but then in case the request is blocked, an active subscribe will not catch that unless the socket disconnects and connects again.

Yeah I don't think this will work given we rely on the API call to the region controller which you can't subscribe on for the following flow:
> I do want to call out I think the region creation flow relies on polling fetchEnvironmentsWithHealth to transition from "booting" to "ready". 

Regarding:
> I was just concerned in the event that suddenly we stop polling for 30 seconds and then the console doesn't try to reconnect in 5 seconds so the user doesn't immediately see the console being unable to connect to environmentd.

Yeah but how this actually displays in the Console is the "environment not ready" toast which we already _deafen_. All other queries need to connect to environmentd anyways so they don't actually require this health check to operate correctly.

jubrad: @SangJunBak doesn't that work if we just poll for initial connection? Also I think we it would be better to do this with by fetching the status of the environment from the region api rather than waiting for a subscribe to work right?

jubrad (approved):
LGTM, might want to make sure Jun is on board, but this seems like a good approach.

console/src/store/environments.ts
SangJunBak: > doesn't that work if we just poll for initial connection?

Hmm wouldn't what work? And what "initial connection" are we referring to? Are we talking about the initial subscribe connection?

>Also I think we it would be better to do this with by fetching the status of the environment from the region api rather than waiting for a subscribe to work right?

If we're talking about the region creation flow, I agree that we can do this separately by fetching the status of the environment from the region api rather than waiting for a subscribe to work. 

Overall, I feel like the added complexity of increasing the poll interval based on global subscription health doesn't offer much than to simply increase the poll interval. Furthermore, if we were to introduce another global subscribe, we'd have to keep `subscribeHealth` updated with it.

leedqin: That makes sense! Updated the polling to 30 seconds in the latest commit

SangJunBak (approved):

PR #36762, opened 2026-05-28
persist-committer: in-envd consensus committer service

antiguru:
Introduces a centralized persist committer service hosted by `environmentd`.

Each environment funnels all persist consensus traffic through a single CockroachDB connection pool instead of every clusterd opening its own.
A new `mz-persist-committer` crate exposes the `Consensus` trait over gRPC, backed by a monotonic in-memory `ShardCache`.
Clusterds and envd itself implement `Consensus` via a new `RpcConsensus` that targets the committer.
Routing is gated behind the `persist_consensus_use_committer` LD flag (default off).

Design doc: `doc/developer/design/20260527_persist_committer.md`.
Plan: `docs/superpowers/plans/2026-05-27-persist-committer.md`.

Status: draft.
Open before merge: cross-crate integration test, full metric instrumentation, validation that `set_committer_channel` ordering versus `open_consensus` is safe under all production startup paths, and a security review of accepting arbitrary persist consensus traffic over the committer socket.


doc/developer/design/20260527_persist_committer.md
```diff
@@ -0,0 +1,361 @@
...
+Existing persist test suites pass against the new transport with no semantic changes.
+
+## Out of scope
+
+* The persist Blob layer (S3/MinIO).
+  Blob uses cheap HTTP connections and does not exhibit the same pressure.
+* Coalescing or merging the *payloads* of `compare_and_set` requests across callers.
+  Caller-provided new-state depends on caller-observed old-state, so payload merging is not semantically sound.
+* Cross-environment consensus sharing.
+  Each environment retains its own consensus DB.
```
antiguru, line 29: This is an interesting point. In our cloud deployment, each region shares a consensus DB. I think each environment should have its own persistd, and no persistd is shared between environments. This isolates environments from another.

antiguru, line 29: This is an interesting point. In our cloud deployment, each region shares a consensus DB. I think each environment should have its own persistd, and no persistd is shared between environments. This isolates environments from another.

doc/developer/design/20260527_persist_committer.md
```diff
@@ -0,0 +1,361 @@
...
+* A server-streaming `Subscribe` RPC that pushes per-shard diffs.
+  An earlier draft included one; no client consumed it, so v1 ships without subscriptions and the cache is refreshed only on `CompareAndSet` mismatch.
+  Diff propagation continues to flow through the existing persist pubsub channel, unchanged by this work.
+
+## Solution proposal
+
+### Overview
+
+Introduce a **persist committer**: an in-process service hosted by `environmentd` that owns the single `PostgresConsensus` connection pool for the environment and exposes the `Consensus` trait over a new gRPC service.
```
antiguru, line 38: I think we should pivot to **persistd** external to environmentd, at the moment just hosting the committer, but in the future it could host committer, pubsub, and maybe even the compactor.

doc/developer/design/20260527_persist_committer.md
```diff
@@ -0,0 +1,361 @@
...
+Clusterds (and `environmentd` itself) use a new RPC-backed `Consensus` implementation that routes every operation to the committer.
+The committer maintains a small in-memory cache of the latest head per shard to serve `head` reads without round-tripping to CRDB.
```
antiguru, line 40: I think the cost of a head query against the metadata DB is so low that it doesn't justify the cost of the cache. Let's remove and add if it is actually needed.

doc/developer/design/20260527_persist_committer.md
```diff
@@ -0,0 +1,361 @@
...
+
+Combined with monotonic insert and the no-fast-reject CaS rule (below), the cache cannot serve a stale value that causes incorrect behavior; the worst case is that a follow-up CaS by this committer mismatches, refreshes, and the caller retries.
+There is no periodic refresh task — an earlier draft included one keyed off active subscribers, but with no subscriber consumer the freshness loop never ran in practice; we removed it rather than ship an inert loop.
+
+### Connectivity and failover
+
+Each `environmentd` runs its own committer.
+Clusterds connect to the leader `environmentd` (the same one that issues controller commands) and reconnect to the new leader during failover, reusing the existing leader-discovery mechanism.
+In-flight RPCs against the demoted committer fail with a retryable error, and the persist client's existing `next_listen_batch_retryer` handles the backoff.
+The new leader's cache starts empty and warms naturally on first access.
```
antiguru, line 149: This is different in the setup with an external persistd.

doc/developer/design/20260527_persist_committer.md
```diff
@@ -0,0 +1,361 @@
...
+The replica makes no progress until reconnect, which matches today's behavior when CRDB is down; controllers are also unreachable in that scenario, so the cluster is effectively idle anyway.
+
+When the committer cannot reach CRDB, the underlying `PostgresConsensus` error propagates verbatim to in-flight RPCs.
+Read-side calls can still be served from the cache under the monotonic guarantee for any shard already warm.
+
+Overload protection — a per-shard outstanding-request queue that rejects with `RESOURCE_EXHAUSTED` past a threshold — is described under follow-ups; v1 ships in-flight gauges but no enforced bound.
+
+### Configuration
+
+All flags are LaunchDarkly-backed and evaluated dynamically:
```
antiguru, line 166: Note that the persistd needs to be either spawned by environmentd and needs to pass through the configuration flags, or it needs its own LaunchDarkly client. The latter is likely the better architecture as it truly decouples the environment and persistd.

doc/developer/design/20260527_persist_committer.md
```diff
@@ -0,0 +1,361 @@
...
+Cheaper alternatives, ordered by ceremony:
+* `tower::Service` over a custom length-prefixed bincode/postcard codec on raw framed TCP. Drops protobuf but keeps the async-server abstraction.
+* `capnp-rpc` (Cap'n Proto): zero-copy reads, fast, but more setup than bincode.
+* Hand-rolled binary protocol over `tokio::net::TcpStream`. Maximum perf, maximum maintenance.
+
+Decision (v1): keep gRPC.
+The work to swap transports is unblocked and orthogonal to everything else in this design, and gRPC matches the pubsub precedent.
+Benchmark the committer once it has miles; if the per-RPC overhead is a meaningful share of total persist latency, swap to a leaner transport.
+
+## Alternatives
```
antiguru, line 247: Another alternative worth mentioning is pgbouncer, with the limits that it cannot add any smartness to the service later.

doc/developer/design/20260527_persist_committer.md
```diff
@@ -0,0 +1,361 @@
...
+* Unit tests in `mz-persist-committer` covering monotonic-insert (random-order property test), cache LRU eviction, and the determinacy-preserving error envelope on the wire.
+* Unit tests in `mz-persist-client::rpc_consensus` covering the client-side roundtrip and the `tonic::Status` → `Indeterminate` mapping for transport failures.
+* A `test/persist-committer/` mzcompose composition that runs `environmentd` plus a clusterd against an external `persistd`, exercising the standalone deployment end-to-end.
+* A `test/persistd/` mzcompose composition that smoke-tests the binary directly and exercises envd restart while `persistd` keeps running.
+* A platform check (`mz-platform-checks`) that verifies shard state survives `environmentd` restart with the committer enabled.
+* Parallel-workload runs with the committer enabled to surface panics or deadlocks under concurrent DDL/DML.
+
+Parameterizing the entire existing persist correctness suite against `RpcConsensus` is appealing but is not part of this work; it is filed as a follow-up.
+
+## Minimal viable prototype
```
antiguru, line 216: Need a section on testing: Just functional testing, performance testing, and correctness under failures (random restarts of persistd). Byzantine failures are out of scope.

src/persist-committer/src/server.rs
```diff
@@ -0,0 +1,560 @@
...
+    ProtoPersistConsensus, ProtoPersistConsensusServer,
+};
+use crate::proto::{
+    ProtoCompareAndSetOk, ProtoCompareAndSetRequest, ProtoCompareAndSetResponse, ProtoDeterminacy,
+    ProtoHeadOk, ProtoHeadRequest, ProtoHeadResponse, ProtoListKeysOk, ProtoListKeysRequest,
+    ProtoListKeysResponse, ProtoOperationError, ProtoScanOk, ProtoScanRequest, ProtoScanResponse,
+    ProtoTruncateOk, ProtoTruncateRequest, ProtoTruncateResponse, ProtoVersionedData,
+    proto_compare_and_set_response, proto_head_response, proto_list_keys_response,
+    proto_scan_response, proto_truncate_response,
+};
```
antiguru, line 232: Either move to separate module, or combine use statements with the ones at the top of the file.

src/persist-committer/src/startup.rs
```diff
@@ -0,0 +1,110 @@
...
+    );
+    let in_process_consensus: Arc<dyn Consensus + Send + Sync> =
+        Arc::new(InProcessConsensus::new(Arc::clone(&committer)));
+    let service = committer.into_service();
+    let server_task = mz_ore::task::spawn(|| "persist_committer::grpc_server", async move {
+        tracing::info!(listen_addr = %listen_addr, "persist committer gRPC server listening");
+        let res = Server::builder()
+            .add_service(service)
+            .serve(listen_addr)
+            .await;
```
antiguru, line 116: Do we have to configure connection timeouts somewhere? Elsewhere we try to protect us from crashed clients by enabling keep-alives.

src/persist/src/location.rs
```diff
@@ -217,6 +217,33 @@ impl ExternalError {
...
+    /// `undefined_table`).
+    ///
+    /// The standalone `persistd` process keys on this. Its consensus schema is
+    /// created once, at startup, by [`crate::postgres::PostgresConsensus::open`].
+    /// If the backing store is reset out from under a running `persistd` the
+    /// relation vanishes and every consensus op then fails indefinitely with
+    /// this code, which no in-process retry can clear. `persistd` treats it as
+    /// fatal and exits so its supervisor restarts it and recreates the schema.
+    /// The in-process committer embedded in `environmentd` deliberately does
+    /// not act on this.
```
antiguru, line 232: Ideally we find a better solution than this. The tables should never disappear, and the only reason we have it is because tests kill and restart envd with a reset metadata DB, but are not aware of persistd.

src/persistd/src/bin/persistd.rs
```diff
@@ -0,0 +1,326 @@
...
+    tracing: TracingCliArgs,
+}
+
+fn humantime_duration_parser(s: &str) -> Result<Duration, String> {
+    humantime::parse_duration(s).map_err(|e| e.to_string())
+}
+
+fn main() {
+    let args: Args = cli::parse_args(CliConfig {
+        env_prefix: Some("MZ_"),
```
antiguru, line 102: Check other binaries: Should this be `MZ_` or `MZ_PERSISTD_`?

src/persistd/src/bin/persistd.rs
antiguru, line 1: Need to expose a metrics and healthcheck endpoint. Check if we can use the generic mz-prof infra?

antiguru:
## Benchmark: does the committer help?

Added a `committer-comparison` workflow to `test/limits` to gather motivating data. It sweeps cluster/MV count across metadata-store backends (cockroach, postgres) with and without the committer, measuring hydration time (`mz_compute_hydration_statuses`) and write-frontier freshness (`mz_wallclock_lag_history`).

**Workload:** N managed clusters × 16 trivial MVs over a tiny ticking table (load is per-tick persist writes, not data). "no committer" = `persist_consensus_use_committer=off` (direct CaS); "persistd" = committer in a persistd companion. Run locally (32 core, 32G envd). Single run.

### The committer rescues the postgres backend (hydration/throughput)

Time to create + hydrate all MVs, and the worst per-object hydration time, at 1600 MVs:

| backend, 1600 MVs | wall (1s tick) | hyd tail (1s) | wall (500ms tick) | hyd tail (500ms) |
|---|--:|--:|--:|--:|
| psql / no committer | 326s | **76s** | **399s** | **214s** |
| psql / persistd | 108s | ~1s | 112s | ~1s |
| crdb / no committer | 146s | ~1s | 152s | ~1s |
| crdb / persistd | 161s | ~1s | 162s | ~1s |

Direct CaS to postgres degrades super-linearly with scale and write rate: the concurrent cold-start CaS burst from ~100 clusterds blows up Postgres SSI predicate-lock tracking (`ERROR: not enough elements in RWConflictPool`), triggering a retry storm. persistd funnels all CaS through one writer and stays flat. **persistd is what makes postgres-metadata viable as a consensus backend under load.** crdb tolerates direct CaS (committer is a slight cost there).

### Where crdb falls apart, the committer is a *regression*

Pushing the tick to 100ms (~10 writes/s per MV), steady-state write-frontier lag:

| crdb, 100ms tick | 400 MVs (~4k w/s) | 800 (~8k) | 1600 (~16k) |
|---|--:|--:|--:|
| no committer | ~2s | ~2s | **mean 56s / p95 79s** |
| persistd | ~2s | ~3s | **mean 92s / p95 122s** |

crdb's knee is ~16k writes/s, and there the single committer is a *throughput bottleneck* (worse than direct CaS, which fans out across crdb's connection pool). So the committer is a postgres-enabler, not a universal win.

### Caveats

* Freshness below ~1s is unmeasurable: `mz_wallclock_lag_history` floors lag to whole seconds (`WallclockLag::Seconds(as_secs())`, `instance.rs:567`). The freshness divergence only shows once lag reaches tens of seconds (crdb collapse above).
* "no committer" still runs envd's in-process committer for envd-internal writes; only the workload (clusterd) writes go direct.
* Single run, no variance bars — but the 76s→214s psql tail and the crdb collapse are far outside noise.
* foundationdb: harness supports `--stores foundationdb`, but single-node fdb wouldn't come up healthy locally (the canonical `test/foundationdb` fails identically here), so no fdb numbers — needs a CI run.

Run: `bin/mzcompose --find limits run committer-comparison [--stores …] [--tick-interval …] [--sizes …]`.

antiguru:
## Correction: the postgres result was a misconfigured-Postgres artifact

The earlier benchmark concluded "persistd is what makes postgres-metadata viable under load."
That is wrong.
The psql collapse (399s wall / 214s hydration tail at 1600 MVs, 500ms tick) was `RWConflictPool` *exhaustion* from default Postgres predicate-lock sizing, not a fundamental limit of direct compare-and-append.

### Raising the predicate-lock limits alone closes the gap

Starting `postgres-metadata` with:

```
max_pred_locks_per_transaction = 1024
max_pred_locks_per_relation   = 10000
max_pred_locks_per_page       = 512
```

| 1600 MVs, 500ms tick | no committer | persistd |
|---|--:|--:|
| original (default PG limits) | **399s wall / 214s hyd tail** | 112s / ~1s |
| pred-locks raised | **119.8s / ~1.05s** | 114.5s / ~1.03s |
| pred-locks + `persist_use_postgres_tuned_queries` | 122.3s / ~1.0s | 116.0s / ~1.05s |

Direct CaS on a correctly-configured Postgres matches persistd.
The committer provides **no hydration-throughput benefit** here.

### Why the tuned query does not help

`persist_use_postgres_tuned_queries=true` switches CaS to the `SELECT … FOR UPDATE` variant, but the numbers are unchanged and Postgres still logs SSI pivots:

```
ERROR: could not serialize access due to read/write dependencies among transactions
DETAIL: Reason code: Canceled on identification as a pivot, during write.
```

Persist forces `SET SESSION CHARACTERISTICS AS TRANSACTION ISOLATION LEVEL SERIALIZABLE` on every consensus connection (`src/postgres-client/src/lib.rs:143`), so SSI runs regardless of `FOR UPDATE` or the server default (`read committed`).
These pivots are absorbed by persist's retries and cost no measurable wall time once the lock pool is not starved.

### Root-cause lever (not pursued here)

The CaS is a single atomic `INSERT … SELECT … WHERE (subselect) = expected`, and its correctness is already guaranteed by `PRIMARY KEY(shard, sequence_number)`.
It does not need SERIALIZABLE on Postgres — that was inherited from CockroachDB.
Dropping the consensus connection to READ COMMITTED would eliminate the pivots entirely, but requires `compare_and_set` to treat a unique-key violation as `ExpectationMismatch` instead of relying on the current 0-rows-affected path.

### What still stands

The crdb high-rate regression (100ms tick, ~16k writes/s) is unaffected: there the single committer is a throughput bottleneck versus crdb's connection-pool fan-out.
The committer still needs a motivation other than postgres hydration throughput — memory offload / envd-OOM avoidance at high MV count remains a real benefit.

The harness now starts `postgres-metadata` with the raised predicate-lock limits and sets `persist_use_postgres_tuned_queries`, so future runs reflect a correctly-configured Postgres.
A postgres run at 100ms tick (the rate where crdb collapses) is in progress; numbers to follow.

antiguru:
## Postgres at 100ms tick (~16k writes/s): same knee as crdb

Ran the postgres backend at 100ms tick with the corrected configuration (raised predicate-lock limits + tuned queries).
This is the rate where crdb's steady-state freshness collapsed earlier.

| psql, 100ms tick | 400 MVs (~4k w/s) | 800 (~8k) | 1600 (~16k) |
|---|--:|--:|--:|
| no committer — lag mean / p95 | ~2s / 2s | ~2s / 2s | **42s / 53s** |
| persistd — lag mean / p95 | ~2s / 2s | ~2s / 2s | **52s / 64s** |
| no committer — hyd tail | 0.2ms | 1.0s | 3.1s |
| persistd — hyd tail | 0.2ms | 1.0s | 1.1s |

Postgres hits the **same ~16k writes/s knee crdb does**.
Below it (4k–8k w/s) lag sits at the 2s tick floor for both variants.
At 16k w/s steady-state freshness collapses to tens of seconds, and persistd is *worse* (52s vs 42s mean lag) — the single committer is a throughput bottleneck versus direct CaS fanning out across the connection pool, exactly as on crdb.

One nuance in the committer's favor: under sustained 16k w/s it keeps the per-object hydration tail tight (1.1s vs 3.1s at 1600 MVs), because the serialized writer avoids retry-induced hydration stalls.
But it trades that for worse aggregate lag.

Net: with a correctly-configured Postgres, the committer is a throughput bottleneck at high write rate on **both** backends and a no-op on hydration throughput at moderate rate.
It is not a postgres-enabler.
Its motivation has to come from elsewhere (memory offload / envd-OOM avoidance at high MV count).

antiguru:
## Pool-exhaustion repro: the committer turns a cascading crash into graceful degradation

Added a `pool-exhaustion` workflow to `test/limits` (`bin/mzcompose --find limits run pool-exhaustion`) that reproduces the production incident where the persist consensus connection pool to Postgres saturates: consensus ops block waiting for a free pool slot, reader-lease heartbeats stall, leases expire, and processes halt.

This is a different failure than the hydration benchmark above. The production logs showed ~21k `Timeout occurred while waiting for a slot to become available` (on `fetch_state::scan` and `downgrade_since`), heartbeats taking minutes, mass lease expiry, and `dataflow reconciliation failed` — **not** `RWConflictPool` exhaustion or SSI pivots.

### Reproduction levers

A fast local Postgres never reproduces it — even `connection_pool_max_size=2` absorbs ~16k ops/s because each CaS is sub-millisecond. Two levers are needed:

* `--pool-sizes` shrinks `persist_consensus_connection_pool_max_size` (production default 50).
* `--postgres-cpu` (default 1) restricts the metadata store so each consensus op holds its connection longer, draining the pool.

With Postgres pinned to 1 CPU and `pool_size=2`: 800 MVs survives cleanly, 1600 MVs overloads.

### Committer off vs on, at the overload point (1600 MVs)

| | outcome | slow heartbeats | lease expired | force-expiring |
|---|---|--:|--:|--:|
| committer **off** | **envd overload crash** | 2221 | 50 | 3903 |
| committer **on** | **survived, no crash** | 396 | 0 | 1 |

Routing consensus through the in-envd committer removes the per-clusterd pool fan-out, so the pool-saturation cascade does not occur: no lease expiry, no force-expiry storm, no envd halt.

The trade-off matches the hydration benchmark: committer-on at 1600 MVs did not finish hydrating in time (`hydrate_ok=false`, 2556s vs 1529s) — the single committer against a 1-CPU Postgres is a throughput bottleneck.

### Takeaway

The committer **converts the failure mode** under metadata-store overload:

* **without committer** — hard cascading failure: transactions hang, leases expire, envd halts (the production incident).
* **with committer** — soft graceful degradation: slower, but stable, with no lease loss and no crash.

This is the incident-grounded motivation for the committer: graceful degradation under metadata-store overload, not raw throughput (under contention it is slower, and at very high write rate it is a bottleneck on both backends, per the earlier comments).

antiguru:
## Handoff notes: the two `test/limits` benchmark workflows

This PR's benchmark work lives in two `test/limits` workflows. Both run locally (32-core box, no balancerd/frontegg — all SQL as `mz_system` on port 6877). Below is how to run each, which dimensions matter, and the data gathered so far.

The harness starts `postgres-metadata` with raised SSI predicate-lock limits and `persist_use_postgres_tuned_queries=on` so Postgres is correctly configured (see the correction comment above — the original psql collapse was `RWConflictPool` exhaustion from default limits, not a fundamental direct-CaS limit).

---

### 1. `committer-comparison` — hydration throughput & steady-state freshness

```
bin/mzcompose --find limits run committer-comparison \
  [--stores cockroach,postgres-metadata] [--tick-interval 500ms] \
  [--sizes 25,50,100] [--mvs-per-cluster 16] [--memory 32G]
```

Sweeps {cockroach, postgres} x {no committer, persistd} over a clusters x MVs workload. Lower `--tick-interval` raises per-MV write rate (500ms ~ 2 w/s, 100ms ~ 10 w/s per MV).

**Dimensions to look at** (CSV columns):
* `batch_hydrate_wall_s` — wall time to create + hydrate all MVs.
* `hydration_ms_max/_p95` — worst per-object hydration tail.
* `lag_ms_*` — steady-state write-frontier lag. **Caveat: floored to whole seconds** (`mz_wallclock_lag_history`, `WallclockLag::Seconds`), so only meaningful once lag reaches tens of seconds.

**Data:**

| 1600 MVs, 500ms tick | no committer | persistd |
|---|--:|--:|
| psql (default PG limits — misconfigured) | 399s / 214s tail | 112s / ~1s |
| psql (corrected: pred-locks + tuned queries) | 120s / ~1s | 115s / ~1s |
| crdb | 146s / ~1s | 161s / ~1s |

| 1600 MVs, 100ms tick (~16k w/s) — lag mean / p95 | no committer | persistd |
|---|--:|--:|
| psql | 42s / 53s | 52s / 64s |
| crdb | 56s / 79s | 92s / 122s |

Takeaways: on a correctly-configured Postgres the committer gives **no** hydration-throughput benefit; at high write rate (~16k w/s) the single committer is a **throughput bottleneck** on both backends.

---

### 2. `pool-exhaustion` — consensus connection-pool overload (reproduces the prod incident)

```
bin/mzcompose --find limits run pool-exhaustion \
  [--committer off|on] [--pool-sizes 4,16,50] [--postgres-cpu 1] \
  [--pool-max-wait 60s] [--sizes 50,100] [--mvs-per-cluster 16] \
  [--load-seconds 120] [--tick-interval 100ms]
```

Reproduces the production failure where each clusterd's consensus connection pool to Postgres saturates → consensus ops block waiting for a slot → reader-lease heartbeats stall → leases expire → processes halt. (Prod logs: ~21k `waiting for a slot to become available` on `fetch_state::scan`/`downgrade_since`, minute-long heartbeats, mass lease expiry, `dataflow reconciliation failed` — **not** `RWConflictPool` or SSI pivots.)

**Key levers** (a fast local Postgres never reproduces it — 2 connections absorb ~16k sub-ms ops/s):
* `--postgres-cpu 1` — restricts the metadata store so each consensus op holds its connection longer. **This is the lever that makes it reproducible.**
* `--pool-sizes` — `persist_consensus_connection_pool_max_size` (prod default 50). Small values force exhaustion sooner in committer-off mode.
* `--pool-max-wait` — `persist_consensus_connection_pool_max_wait` (prod 60s). The `slot_wait` log only emits when this times out; lower it (e.g. `5s`) to surface it explicitly.
* `--committer on|off` — off matches prod (clusterds direct to PG, own pools); on routes through the in-envd committer.

**Dimensions to look at** (CSV columns; counts are log-line occurrences scraped from the materialized container, reported as per-size deltas):
* `overloaded` — envd lost its connection (hard failure / terminal state).
* `hydrate_ok` — did all MVs hydrate within timeout.
* `slot_wait` — `waiting for a slot to become available` (the prod-dominant signature).
* `heartbeat_took` / `lease_expired` / `force_expiring` — the lease-expiry cascade.
* `reconciliation_failed` / `halting` / `lost_lease` — fleet-wide fallout.

**Data** (Postgres pinned to 1 CPU; envd has 28 CPUs):

| pool, committer, 1600 MVs | overloaded | hydrate_ok | wall | heartbeat_took | lease_exp | force_exp |
|---|---|---|--:|--:|--:|--:|
| pool=2, **off** | **yes (crash)** | no | 1529s | 2221 | 50 | 3903 |
| pool=2, **on** | no | no | 2556s | 396 | 0 | 1 |
| pool=50, **on** | no | no | 2612s | 848 | 0 | 0 |

* At 800 MVs every variant survives clean (the knee is between 800 and 1600 MVs at pool=2 / 1 CPU).
* `slot_wait` confirmation: pool=2, 1 CPU, `--pool-max-wait 5s`, 1200 MVs → **slot_wait=493** (the exact prod signature).

**Takeaway:** the committer **converts the failure mode** under metadata-store overload — committer-off cascades into a hard crash (hang → lease expiry → envd halt, the prod incident); committer-on degrades gracefully (slower, but stable, no lease loss, no crash). With the committer on, pool size barely matters (pool=2 ~ pool=50) because the bottleneck is the slow Postgres downstream of the single committer, not the pool.

---

### Open questions / suggested next runs

* **committer-off at the production pool default (50), 1 CPU** — does it also cascade, or only at the artificial pool=2? This decides whether the committer's protection matters at realistic pool sizes. (Only off/pool=2 has been run.)
* Sweep `--postgres-cpu` (1 → 2 → 4) to find where the overload knee moves; map the safe operating envelope.
* The deeper root-cause lever (not pursued): persist forces `SERIALIZABLE` on the consensus connection (`src/postgres-client/src/lib.rs:143`), but the single-statement CaS is already made atomic by `PRIMARY KEY(shard, sequence_number)`. Dropping to READ COMMITTED on Postgres would remove SSI pivots, but needs `compare_and_set` to treat a unique-key violation as `ExpectationMismatch`.

Result CSVs are local artifacts (gitignored); regenerate with the commands above. Both workflows are committed to this branch.

antiguru:
## CaS coalescing prototype: reverses the high-rate regression

Follow-up to the 100ms-tick result where the committer was a throughput bottleneck at ~16k writes/s (psql 52s/64s, crdb 92s/122s mean/p95 lag vs better direct CaS).
A local prototype of the "single-writer batched CaS" direction from the design doc makes the committer the best variant on both backends at the same load.

### Results

1600 MVs (100 clusters × 16), `default_timestamp_interval = 100ms`, ~14-16k CaS/s sustained, steady-state wallclock lag:

| variant | crdb mean/p95 | psql mean/p95 |
|---|---|---|
| no committer (direct CaS) | 56s / 79s | 42s / 53s |
| persistd | 92s / 122s | 52s / 64s |
| persistd + coalescing (batch 64, concurrency 16) | **46s / 62s** | **37s / 44s** |

* 400/800 MVs sit at the 2s tick-quantization floor on all variants; divergence only appears at the 16k w/s knee.
* All variants hydrate; coalesced psql hydration tail is 4s vs 1.1s baseline (whole-batch conflict retries during the DDL burst), p95 0.1ms.
* Committer steady state on psql: 14.3k CaS/s at 9.5ms mean, batch latency ~3.5ms, adaptive batch size ~3 (grows under backlog, small when the store keeps up).
* Residual tens-of-seconds lag at this rate is the single-node metadata store's knee, not the committer.

### Design

Coalescing is recovered behind the `Consensus` trait, invisible to the committer:

* `Consensus::compare_and_set_multi(Vec<(key, VersionedData)>) -> Result<Vec<CaSResult>>`, default impl loops over `compare_and_set`.
* `PostgresConsensus` implements it as one UNNEST statement over arrays, `RETURNING shard` identifying committed elements; first writes (seqno 0) fall back to the singleton path.
* `CoalescingConsensus` wraps any `Consensus`: `compare_and_set` enqueues, a drainer dispatches batches of distinct shards, pipelined up to a concurrency limit.
* A shard is in at most one in-flight batch, preserving per-shard FIFO; same-shard followers wait.
* The committer interposes the wrapper between itself and the backing store when enabled, so its caching logic is unchanged.

### Three findings that apply to the productized batch RPC

* Serial batch dispatch caps throughput at `batch_size / batch_latency` (~800 CaS/s observed); pipelined shard-disjoint batches are required.
* A per-shard `ORDER BY seqno DESC LIMIT 1` subselect inside a multi-row statement decorrelates into a full-history scan per element on both CRDB and Postgres, and the cost grows with the truncation backlog it causes (observed 65ms → 500ms/statement, 17s CaS mean).
  Expressing "expected is head" as `EXISTS (shard, seqno = expected) AND NOT EXISTS (shard, seqno > expected)` plans as semi + anti lookup join, O(1) per element, and keeps CRDB's 1PC `auto commit`.
* Batch errors must surface as indeterminate per element: loop-based implementations partially commit before failing, and reporting a determinate failure for a committed element trips the `AlreadyCommitted` assertion in `Machine::compare_and_append` and aborts envd.

Prototype is local to the `persist-committer-design` worktree and not yet pushed; data in `committer_coalesce_100ms.csv`.

antiguru:
## Correction + stability analysis for the coalescing results

The "steady state" lag numbers in my previous comment (#issuecomment-4633742702) are cold-start snapshots of diverging backlogs, not equilibria.
Rerunning with a 120s settle window before sampling shows lag at 1600 MVs growing roughly with wall time on every implementation (e.g. coalesced crdb p95 62s → 195s, coalesced psql 44s → 144s).
Above the 2s tick floor nothing is stable, so per-variant lag magnitudes only rank who diverges slower; the meaningful metric is binary: holds the floor or not.

### Stability per implementation (settled windows, 100ms tick)

| MVs (writes/s) | crdb direct | crdb persistd | crdb +coalesce | psql direct | psql persistd | psql +coalesce |
|---|---|---|---|---|---|---|
| 800 (8k) | ✗ (7s) | ✗ (10s) | ✓ floor | ✓ floor | ✓ floor | ✓ floor |
| 1200 (12k) | ✗ | ✗ | ✗ | ✗ | ✗ | ✗ |

* Every implementation's inflection point lies between 8k and 12k writes/s on this box; coalescing does not move it.
* Coalescing is the only crdb configuration that holds the floor at 8k w/s; both crdb baselines sit slightly above it.
* Above the knee, coalescing tracks direct CaS and outpaces the plain committer, but all three diverge.
* The committer's earlier 2× high-rate regression is gone with coalescing; its remaining value above the knee is graceful degradation, not added capacity.

Data: `committer_coalesce_100ms.csv` (short window), `committer_coalesce_100ms_settled.csv`, `committer_knee_settled{,_coal}.csv`.

### Crashes observed at 1600 MVs under sustained divergence

* `psql / persistd` (coalescing **disabled**, plain committer): envd abort, `gc.rs:417 earliest state fetched during GC did not have corresponding rollup`, state seqno v13197 between rollups v13068/v13264.
  Not caused by the coalescing changes (the wrapper is not instantiated in that variant); `psql / no committer` survived the same load in the same run, which points at the committer path — but both are single observations.
  Worth investigating before merge; likely candidates are the committer cache serving stale heads to GC's state fetches under heavy backlog.
* `psql / +coalesce`: envd OOM-killed at the 32G container limit while 1600 diverging MVs accumulate correction state — load-shedding territory, not a consensus bug.

### Future work

* FoundationDB comparison: a single-node fdb metadata store would not come up healthy on this box previously; retry (possibly in CI) now that the harness sweeps stability rather than lag magnitude.
* Resolve the knee more finely (bisection between 800 and 1200 MVs) if the exact per-implementation capacity number matters.

Coalescing prototype is now committed on this branch (`persist: add batched CaS and committer-side coalescing`, `limits: add coalescing variants to committer comparison`).

PR #36764, opened 2026-05-28
sql-pretty: Fix output of passwords and AS OF

def-:
Noticed in https://materializeinc.slack.com/archives/C085QFLB072/p1779932199965099

### DIFF
--- a/src/sql-pretty/tests/lossless.rs
+++ b/src/sql-pretty/tests/lossless.rs
@@ -1,0 +1,63 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! Round-trip tests asserting that pretty-printing preserves AST nodes that the
+//! generic `AstDisplay` redacts or omits. Unlike `parser.rs`, which compares
+//! `to_ast_string_stable()` (and so drops passwords on both sides), this compares
+//! the full parsed AST, proving the value actually survives the round trip.
+
+#![cfg(not(target_arch = "wasm32"))]
+
+use mz_sql_parser::parser::parse_statements;
+use mz_sql_pretty::pretty_str_simple;
+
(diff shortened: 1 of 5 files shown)


src/sql-pretty/tests/lossless.rs
antiguru, line 1: We could probably add fuzz testing to ensure we don't miss statement parts?

antiguru (approved):
Thanks!

src/sql-pretty/tests/lossless.rs
def-: Yes, but I'll do that outside of this PR since it will probably end up being more involved.

justjake:
yippee! 🎉

test/sqllogictest/distinct_arrangements.slt
```diff
@@ -1057,12 +1057,12 @@ JOIN mz_introspection.mz_dataflow_operator_dataflows mdod ON mash.operator_id =
...
 ----
-AccumulableErrorCheck  3
+AccumulableErrorCheck  4
 Arrange␠ReduceMinsMaxes  1
-ArrangeAccumulable␠[val:␠empty]  3
+ArrangeAccumulable␠[val:␠empty]  4
 ArrangeBy[[Column(0,␠"operator_id")]]  2
 ArrangeBy[[Column(1,␠"dataflow_id")]]  2
-ReduceAccumulable  3
+ReduceAccumulable  4
```
ggevay, line 1065: Hmm, do you know why this change happened?

def-: It was something else that hit main, not caused by my PR, same failure appeared in Nightly today: https://buildkite.com/materialize/nightly/builds/16598
So it's some change in this range: https://github.com/MaterializeInc/materialize/compare/f1513fb122...25f0e42865
If it's important I can bisect later.

ggevay: Ah, thanks! Maybe it's https://github.com/MaterializeInc/materialize/commit/3aa14b6c5b700e9dcbfe29ad258c75398dcec36d. I'll run before and after to confirm.

ggevay: Confirmed, ok!

PR #36765, opened 2026-05-28
compute-types: fix FlatMap column refs for non-column arrangement keys

def-:
When a FlatMap reads from an arrangement whose key is not a set of distinct column references (an expression, functional, or repeated-column index), the arranged row is wider than the logical input row because it carries the extra key value(s). The renderer hands the MFP the whole arranged row and appends the table-function output after it, so the MFP must reference the output columns at that wider position and use the wider input arity. Leaving them at their logical positions produced silently wrong results: an empty result when an output column was filtered, or a wrong-column projection otherwise.

Account for the arranged width (thinning.len() + key.len()) when permuting the MFP, and extend github-11280.slt to cover expression, functional, and repeated-column keys.

Closes CLU-70

Sorry for only noticing after https://github.com/MaterializeInc/materialize/pull/36756 merged!

### DIFF
--- a/src/compute-types/src/plan/lowering.rs
+++ b/src/compute-types/src/plan/lowering.rs
@@ -607,23 +607,36 @@ impl Context {
                     // in logical order, so no permutation is required.
                     let input_key = if keys.raw {
                         None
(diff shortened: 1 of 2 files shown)


antiguru (approved):
Thank you!

PR #36767, opened 2026-05-28
rust: Fix some compiler warnings about imports only being required with proptest

def-:


ggevay (approved):
Thank you!

antiguru (approved):
Thank you! My bad, missed this. Should we have a lint that checks that non-test builds don't have warnings?

def-:
Sure, let's do that: https://github.com/MaterializeInc/materialize/pull/36774

PR #36769, opened 2026-05-28
ci: Improve retries when Github is slow

def-:
Works together with https://github.com/MaterializeInc/i2/pull/3301


bosconi (approved):
Looks good.

PR #36770, opened 2026-05-28
slt: Fix variadic_outer_join.slt plan

def-:
https://github.com/MaterializeInc/materialize/pull/36640 wasn't rebased and conflicted with something else. Seen in https://buildkite.com/materialize/test/builds/124015

Nightly run: https://buildkite.com/materialize/nightly/builds/16604


ggevay (approved):
Thank you!

PR #36775, opened 2026-05-28
[SCIM] make oidc_group_claim and oidc_group_role_sync_strict user-modifiable

mtabebe:
Allow environment superusers to configure OIDC group-sync settings via ALTER SYSTEM SET without needing mz_system access. This is appropriate since these control how a customer's own OIDC claims map to roles.

Fixes SQL-321

### DIFF
--- a/src/sql/src/session/vars.rs
+++ b/src/sql/src/session/vars.rs
@@ -76,6 +76,7 @@ use std::time::Duration;
 use chrono::{DateTime, Utc};
 use derivative::Derivative;
 use imbl::OrdMap;
+use mz_adapter_types::dyncfgs::{OIDC_GROUP_CLAIM, OIDC_GROUP_ROLE_SYNC_STRICT};
 use mz_build_info::BuildInfo;
 use mz_dyncfg::{ConfigSet, ConfigType, ConfigUpdates, ConfigVal};
 use mz_persist_client::cfg::{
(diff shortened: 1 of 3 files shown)


jubrad (approved):

def- (commented):
Both `oidc_group_claim` and `oidc_grouP_role_sync_strict` are dyncfg-backed, and all dyncfgs register with `user_visible: false`. The `ALTER SYSTEM SET` authorization path calls `var.visible()`, which rejects `user_visible: false` vars for anyone who isn't `mz_system`/`mz_support`. So an environment superuser (a Superuser that is not internal) still can't set or use them. Is that intentional?

Reproducer:
```diff
$ postgres-connect name=mz_system url=postgres://mz_system:materialize@${testdrive.materialize-internal-sql-addr}

$ postgres-execute connection=mz_system
CREATE ROLE env_superuser SUPERUSER LOGIN;

$ postgres-connect name=env_superuser url=postgres://env_superuser@${testdrive.materialize-sql-addr}

# Control: a user-modifiable + user-visible var works, proving env_superuser is a
# real superuser (not a "must be a superuser" rejection).
$ postgres-execute connection=env_superuser
ALTER SYSTEM SET cluster TO quickstart;

# Regression: fails on the buggy code with "unrecognized configuration parameter".
$ postgres-execute connection=env_superuser
ALTER SYSTEM SET oidc_group_claim = 'my_groups';
ALTER SYSTEM SET oidc_group_role_sync_strict = on;

> SHOW oidc_group_claim
my_groups

> SHOW oidc_group_role_sync_strict
on
```
Fails with:
```
>> ALTER SYSTEM SET oidc_group_claim = 'my_groups';
^^^ +++
oidc-group-sync-vars.td:45:1: error: executing postgres query: db error: ERROR: unrecognized configuration parameter "oidc_group_claim"
```

jubrad:
Yeah, I though this would work as well, but this should follow the pattern for vars in `src/sql/src/session/vars/definitions.rs` right?

PR #36776, opened 2026-05-28
Count aware aggregates

frankmcsherry:
## Summary

Optimizations that together let a row-wise `count(*)` over an integer `generate_series` evaluate in O(1) instead of materializing one row per generated value, plus a stable escape hatch for tests that rely on the enumeration actually happening.

- **Count-aware aggregate eval.** `AggregateFunc::eval` now consumes `Item = (Datum, Diff)` instead of bare `Datum`. `count` sums the diffs, multiplicity-insensitive aggregates (min/max/any/all) drop them, signed integer sums fold `value * diff` in i128 (mirroring the dataflow's accumulable rendering), and the rest expand by count as before. Avoids `fold_reduce_constant` expanding a compact `(row, N)` into N rows before aggregation. On non-positive multiplicities (invalid collections only), the rendered basic path now passes signed diffs through to the aggregate — moving it toward the accumulable path's behavior, with the companion error-check reduction still reporting the inconsistency; `fold_constants` continues to refuse such collections.

- **Collapse unused integer `generate_series`.** When a `generate_series(start, stop, step)` FlatMap's output column is not demanded, projection_pushdown rewrites it into a `RepeatRowNonNegative` emitting the same row count. With all-literal arguments the cardinality is computed exactly at rewrite time in i128 and emitted as a literal count, declining when it exceeds i64 (the original enumerates such series without error — its checked stepping terminates exactly where overflow would occur — so the rewrite must not introduce one). With column-valued bounds and a non-zero literal step, a sign-guarded `CASE` computes `floor((stop - start) / step) + 1`; under the emptiness guard the i64 arithmetic can only error when both bounds are extreme with opposite signs (span >= 2^63), which we accept and document.

- **`mz_unsafe.generate_series_unoptimized`.** An int64 `generate_series` the optimizer promises to leave as an enumeration: no transform may match the new `TableFunc::GenerateSeriesUnoptimized` variant to apply cardinality shortcuts, though its argument expressions still simplify. This gives stress tests a stable way to rely on the enumeration work happening (robust against future rewrites, e.g. closed-form aggregates of a series). It lives in `mz_unsafe` to make the contract structural: not a supported surface, and bug reports must not depend on it.

## Testing

- `test/sqllogictest/generate_series_collapse.slt` is a boundary specification for the collapse: covered (int4/int8, two-argument form, literal steps of either sign, exact literal folds including spans too wide for i64 arithmetic, NULL bounds) and not covered (value-dependent demand, non-literal step, zero step, counts beyond i64, timestamp series, WITH ORDINALITY, and — by contract — `generate_series_unoptimized`), asserting plan shapes and results. Notably `count(g)` of the series column still collapses: the optimizer rewrites count of a non-nullable column to `count(*)` first.
- Transform-level tests pin the projection_pushdown rewrite including the literal-fold and decline cases.

This reproduces the work of #22753, but with the assistance of 🤖 to grind through the aggregate eval work.

cc: @def- : `mz_unsafe.generate_series_unoptimized` is the escape hatch we discussed — tests whose load relies on enumeration can migrate to it and become immune to this and future series optimizations. Note that `count(x)` is NOT a durable alternative (it rewrites to `count(*)` and collapses); value-dependent aggregates like `sum(x)` work today but could fall to closed forms eventually.


def- (commented):
with this setup:
```sql
CREATE TABLE t (lo int8, hi int8);
INSERT INTO t VALUES (-5000000000000000000, 5000000000000000000);
```

On current main:
```
materialize=> SELECT count(*) FROM t, generate_series(t.lo, t.hi, 2000000000000000000) g;
 count
-------
     6
(1 row)
```

On this PR:
```
materialize=> SELECT count(*) FROM t, generate_series(t.lo, t.hi, 2000000000000000000) g;
ERROR:  Evaluation error: numeric field overflow
```

src/transform/tests/testdata/projection-pushdown
```diff
@@ -184,6 +184,88 @@ Project (#2)
     Project (#0, #2)
       Get x
 
+# Unused generate_series with a non-unit positive step collapses
+build apply=ProjectionPushdown
```
ggevay: It would be good to have these as `.spec` tests instead, because the mzreflect/lowertest-based optimizer tests are deprecated.

test/sqllogictest/transform/generate_series_collapse.slt
ggevay, line 1: Maybe move under `test/sqllogictest/transform`.

src/transform/src/movement/projection_pushdown.rs
```diff
@@ -516,6 +523,119 @@ impl ProjectionPushdown {
...
+/// where the original would not: `stop - start` overflows when the span is at
+/// least `2^63`, which under the emptiness guard requires both bounds extreme
+/// with opposite signs — and for the original plan to have been *feasible*
+/// rather than effectively non-terminating, also a step of around `10^10` or
+/// more. We accept this corner.
+///
+/// Null inputs are handled by `RepeatRowNonNegative` itself: like
+/// `generate_series`, it is `empty_on_null_input`, so a null count yields no
+/// rows.
+fn collapse_unused_generate_series(func: &mut TableFunc, exprs: &mut Vec<MirScalarExpr>) {
```
ggevay, line 569: Comment from Claude:

Document the +1 overflow corner in collapse_unused_generate_series. The doc comment presents the stop - start (SubInt64) span-overflow as the corner where synthesized arithmetic can error. But the final AddInt64(_, 1) can also overflow when the quotient is exactly i64::MAX — reachable with step = ±1 and a span just under 2^63 (e.g. column bounds 0, i64::MAX), which does not trip SubInt64. It's in the same "infeasible anyway" category (such a series has > i64::MAX rows, exactly what the literal path declines), so erroring is acceptable — but the comment currently reads as exhaustive about the corners, and this one isn't mentioned.

ggevay (approved):
LGTM!

PR #36778, opened 2026-05-29
Fix physical monotonicity analysis for MFP and FlatMap

antiguru:
### Motivation

CLU-68: The `SingleTimeMonotonic` physical-monotonicity interpreter propagated
the input's monotonicity through `mfp` and `flat_map` without consulting the
MFP's predicates or the table function's properties. The MIR monotonicity
analysis (`src/transform/src/analysis/monotonic.rs`) already accounts for both,
so the LIR interpreter was inconsistent with it.

### Severity

The practical correctness impact is low, and this is best understood as
defense-in-depth / consistency rather than a live bug fix:

- This analysis only ever **relaxes** the safe default. Single-time peeks set
  `must_consolidate = true` on hierarchical reductions / TopK by default; a
  positive physical-monotonicity result is what lets that be relaxed to
  `false`. A spuriously monotonic result is therefore the only way to cause
  trouble.
- The only LIR-reachable operator that can actually emit negative diffs is
  `repeat_row` with a negative count, which requires the unsafe
  `enable_repeat_row` flag and is not meaningfully representable in a
  single-time peek (a consolidated `-1` row has no SQL meaning).
- `preserves_monotonicity()` also returns `false` for `AclExplode`,
  `MzAclExplode`, and `GuardSubquerySize`, but those classifications are
  conservative — they do not emit negative diffs.

So no known query produces wrong results today; the change keeps the LIR
interpreter honest and aligned with the MIR analysis.

### Description

Corrects `mfp` and `flat_map` in `SingleTimeMonotonic`:

1. **`flat_map`** — now requires `func.preserves_monotonicity()` (the
   substantive check; e.g. `repeat_row` does not preserve it) in addition to
   propagating the input status. Mirrors `MirRelationExpr::FlatMap`.

2. **Temporal-predicate checks** (`mfp`, and the post-MFP of `flat_map`) — now
   reject MFPs containing temporal predicates (`mz_now()`). Note this is
   *conservative*: an MFP cannot itself negate diffs, and temporal predicates
   only introduce retractions *across* timestamps, while a one-shot dataflow
   runs at a single time. We keep the check as defense-in-depth and to mirror
   `MirRelationExpr::Filter`.

The code comments document this severity nuance inline.

### Verification

Added unit tests covering:
- `flat_map` with a monotonicity-preserving function (`GenerateSeriesInt64`)
- `flat_map` with a non-preserving function (`RepeatRow`)
- `flat_map` with a temporal predicate in the post-MFP
- `mfp` with and without temporal predicates

https://claude.ai/code/session_013EEMmczcmK98tcSa9B9vvQ

### DIFF
--- a/src/compute-types/src/plan/interpret/physically_monotonic.rs
+++ b/src/compute-types/src/plan/interpret/physically_monotonic.rs
@@ -125,12 +125,15 @@ impl Interpreter for SingleTimeMonotonic<'_> {
         &self,
         _ctx: &Context<Self::Domain>,
         input: Self::Domain,
-        _mfp: &MapFilterProject,
+        mfp: &MapFilterProject,
         _input_key_val: &Option<(Vec<MirScalarExpr>, Option<Row>)>,
     ) -> Self::Domain {
-        // In a single-time context, we just propagate the monotonicity
-        // status of the input
-        input
+        // In a single-time context, we propagate the monotonicity status of the
+        // input, unless the MFP contains temporal predicates. Temporal predicates
+        // (i.e., `mz_now()`) can result in the future removal of records, thus
+        // breaking physical monotonicity. This matches the judgment made for
+        // `MirRelationExpr::Filter` in the MIR monotonicity analysis.
+        PhysicallyMonotonic(input.0 && !mfp.has_temporal_predicates())
     }
 
     fn flat_map(
@@ -139,12 +142,19 @@ impl Interpreter for SingleTimeMonotonic<'_> {
         _input_key: &Option<Vec<MirScalarExpr>>,
         input: Self::Domain,
         _exprs: &Vec<MirScalarExpr>,
-        _func: &TableFunc,
-        _mfp: &MapFilterProject,
+        func: &TableFunc,
+        mfp: &MapFilterProject,
     ) -> Self::Domain {
-        // In a single-time context, we just propagate the monotonicity
-        // status of the input
-        input
+        // In a single-time context, we propagate the monotonicity status of the
+        // input, but only if the table function preserves the append-only
+        // property of its input (e.g., `repeat_row` does not). Additionally, the
+        // MFP applied after the table function must not contain temporal
+        // predicates, as those can result in the future removal of records. This
+        // matches the judgment made for `MirRelationExpr::FlatMap` (combined with
+        // `MirRelationExpr::Filter`) in the MIR monotonicity analysis.
+        PhysicallyMonotonic(
+            input.0 && func.preserves_monotonicity() && !mfp.has_temporal_predicates(),
+        )
     }
 
     fn join(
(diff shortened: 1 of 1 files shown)


src/compute-types/src/plan/interpret/physically_monotonic.rs
```diff
@@ -125,12 +125,19 @@ impl Interpreter for SingleTimeMonotonic<'_> {
...
-        input
+        // In a single-time context, we propagate the monotonicity status of the
+        // input, conservatively treating an MFP with temporal predicates as
+        // breaking physical monotonicity. Note that an MFP cannot itself produce
+        // negative diffs (map/filter/project never negate), so this is stricter
+        // than strictly necessary for a single-time dataflow: temporal predicates
+        // (`mz_now()`) only introduce retractions *across* timestamps, and a
+        // one-shot dataflow runs at a single time. We keep the check anyway as
+        // defense-in-depth and to mirror the judgment made for
+        // `MirRelationExpr::Filter` in the MIR monotonicity analysis.
```
DAlperin, line 139: This strikes me as correct, but it is possible to imagine a case where this causes noticably worse performance for things like COPY TO where the query uses a temporal predicate. Or even just peeks with temporal predicates… Since what we were doing before wasn't strictly wrong either. <br><br>This is a good change, just slightly unfortunate it can introduce a regression

DAlperin (approved):

src/compute-types/src/plan/interpret/physically_monotonic.rs
antiguru: So, once we come here, we should've replaced `mz_now` with a concrete timestamp so the temporal filter essentially disappears. Let me check, but I think there shouldn't be a performance regression (and otherwise, it's the right thing as it'd be a correctness issue otherwise).

antiguru: Confirmed — your reasoning holds, so there's no performance regression from the temporal-predicate checks. Tracing it through:

- `SingleTimeMonotonic` runs only via `RelaxMustConsolidate`, which `finalize_dataflow` invokes solely under `if dataflow.is_single_time()` (`compute-types/src/plan.rs:564-593`) — i.e. one-shot peeks and `COPY TO`.
- In those pipelines, `ExprPrepOneShot::prep_scalar_expr` (`adapter/src/optimize/dataflows.rs:216`) const-folds every `CallUnmaterializable`, including `MzNow`, to a literal. This runs *before* `finalize_dataflow` in both paths (`peek.rs`: prep at 305, finalize at 417; `copy_to.rs`: prep at 306, finalize at 354).
- So `mz_now() < col` has become `<literal> < col` by the time we get here — no temporal predicates survive into LIR, and `has_temporal_predicates()` is always `false`. The new `mfp`/`flat_map` temporal checks are provable no-ops in this context (pure defense-in-depth).

This is the same invariant already pinned in `compute/src/render/top_k.rs:71-81` ("`ExprPrepOneShot` constant-folds `mz_now()` to the dataflow `as_of` before lowering, so no temporal predicates survive into LIR"), backed by a `soft_assert_or_log!`.

The only behavioral change is the `flat_map` `func.preserves_monotonicity()` check, which forces consolidation for non-monotone table functions (`repeat_row` et al.) feeding a monotonic reduce/TopK — that's the correctness fix, and it doesn't touch the temporal-predicate cases @DAlperin was concerned about.

---

PR #36779, opened 2026-05-29
Fix generate_subscripts to respect array lower bounds

antiguru:
### Motivation

Fixes database-issues#11268: `generate_subscripts` was not respecting custom lower bounds on arrays. When an array was created with a custom lower bound (e.g., via `array_fill` with a lower bound parameter), `generate_subscripts` would incorrectly start from 1 instead of the array's actual lower bound.

### Description

The issue was in `generate_subscripts_array` in `src/expr/src/relation/func.rs`. The function was calling `generate_series` with only the `lower_bound` and `length` parameters, but `generate_series` expects the upper bound as the second argument, not the length.

The fix:
1. Extract `lower_bound` and `length` from the dimension into separate variables
2. Calculate the actual `upper_bound` as `lower_bound + length - 1` (since subscripts are inclusive on both ends)
3. Add overflow checking when computing the upper bound
4. Pass the correct `lower_bound` and `upper_bound` to `generate_series`

This ensures that `generate_subscripts` correctly generates subscripts from the array's actual lower bound through its upper bound, inclusive.

### Verification

Added two regression tests in `test/sqllogictest/table_func.slt`:
1. Single-dimensional array with custom lower bound 5 and length 3, expecting subscripts 5, 6, 7
2. Multi-dimensional array with custom lower bounds (10, 20) and lengths (2, 3), expecting subscripts 20, 21, 22 for dimension 2

Both tests verify the fix works correctly for arrays created with `array_fill` using custom lower bounds.

https://claude.ai/code/session_01C2n4d7MhwFRXVezyxs3DgQ

### DIFF
--- a/src/expr/src/relation/func.rs
+++ b/src/expr/src/relation/func.rs
@@ -2960,16 +2960,21 @@ fn generate_subscripts_array(
             .try_into()
             .map_err(|_| EvalError::Int32OutOfRange((dim - 1).to_string().into()))?,
     ) {
-        Some(requested_dim) => Ok(Box::new(generate_series::<i32>(
-            requested_dim.lower_bound.try_into().map_err(|_| {
+        Some(requested_dim) => {
+            let lower_bound: i32 = requested_dim.lower_bound.try_into().map_err(|_| {
                 EvalError::Int32OutOfRange(requested_dim.lower_bound.to_string().into())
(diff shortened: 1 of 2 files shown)


DAlperin (approved):
LGTM, thanks

PR #36780, opened 2026-05-29
Fix array_lower/array_upper to respect custom lower bounds

antiguru:
### Motivation

Closes https://github.com/MaterializeInc/database-issues/issues/11278

Both `array_lower()` and `array_upper()` ignored the array's custom lower bound:

- `array_lower()` always returned `1`, regardless of the array's actual lower bound.
- `array_upper()` returned the dimension's *length* rather than its upper bound (`lower_bound + length - 1`).

For arrays with the default lower bound of `1` these happen to be correct (since `upper_bound == length`), so the bug only surfaced for arrays with a non-default lower bound, e.g. those built by `array_fill()`:

```
SELECT array_lower(array_fill(0, ARRAY[3], ARRAY[5]), 1);  -- returned 1, should be 5
SELECT array_upper(array_fill(0, ARRAY[3], ARRAY[5]), 1);  -- returned 3, should be 7
```

PostgreSQL returns `5` and `7` respectively.

### Description

- **`array_lower()`**: now returns the dimension's actual lower bound via `dim.dimension_bounds()` instead of hardcoding `1`. The return type changed from `Option<i32>` to `Result<Option<i32>, EvalError>` (matching `array_upper()`) so a bound that does not fit in `i32` is reported as an error.
- **`array_upper()`**: now returns the dimension's upper bound via `dim.dimension_bounds()` instead of `dim.length`.
- Reduced the `sqlfunc` attribute lists on both functions to just `is_infix_op = true`; `output_type`, `sqlname`, `propagates_nulls`, and `introduces_nulls` are inferred by the macro from the signature and matched the previous explicit values.

### Verification

- Added a unit test (`array_lower_upper_respect_lower_bound`) in `src/expr/src/scalar/func.rs` covering default, custom positive, negative, and out-of-range dimensions for both functions.
- Added test cases in `test/sqllogictest/array_fill.slt` for single-dimension custom (`5`), negative (`-3`), and multi-dimensional (`4`, `3`) lower bounds.

https://claude.ai/code/session_01A4oEvWwfEb6yE8FYgPAfpQ

### DIFF
--- a/src/expr/src/scalar/func.rs
+++ b/src/expr/src/scalar/func.rs
@@ -2732,12 +2732,7 @@ fn array_length<'a>(a: Array<'a>, b: i64) -> Result<Option<i32>, EvalError> {
     })
 }
 
-#[sqlfunc(
-    output_type = "Option<i32>",
-    sqlname = "array_lower",
-    propagates_nulls = true,
-    introduces_nulls = true
-)]
+#[sqlfunc(is_infix_op = true)]
 // TODO(benesch): remove potentially dangerous usage of `as`.
 #[allow(clippy::as_conversions)]
 fn array_lower<'a>(a: Array<'a>, i: i64) -> Result<Option<i32>, EvalError> {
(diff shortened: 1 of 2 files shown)


test/sqllogictest/array_fill.slt
```diff
@@ -372,3 +372,37 @@ ORDER BY o, i;
 0	-2	null
 0	-1	null
 0	0	null
+
+# array_lower/array_upper must respect the lower bounds of the array
+# (https://github.com/MaterializeInc/database-issues/issues/11278).
```
antiguru: Remove mention of the issue number.

src/expr/src/scalar/func.rs
antiguru:
```suggestion
   is_infix_op = true
```

antiguru:
```suggestion
   is_infix_op = true
```

DAlperin (approved):
LGTM, thanks

PR #36781, opened 2026-05-29
compute: fix SUM(float8) for large finite values

antiguru:
### Motivation

Fixes CLU-75 / database-issues#11265: `SUM(float8)` returns incorrect results for large finite values when read from a table.

```sql
CREATE TABLE float_boundary (f1 float8);
INSERT INTO float_boundary VALUES (1.1e31), (-1.1e31);
SELECT SUM(f1) FROM float_boundary;
-- before:  -5.960464477539063e-8
-- after:   0
```

### Root cause

The accumulable float sum maps each input onto a fixed-point `i128` domain (scaled by `2^24`) and combines accumulators with **wrapping** arithmetic so the representation forms a commutative group, as required for incremental maintenance with retractions.

The per-datum conversion, however, used a **saturating** `as i128` cast. Values whose scaled magnitude exceeds `i128`'s range (`|n| > ~1.01e31`) saturated to `i128::MAX` or `i128::MIN`. Because `i128`'s range is asymmetric (`MIN = -(MAX + 1)`), summing two cancelling values such as `1.1e31` and `-1.1e31` produced `MAX + MIN = -1`, surfacing as `-5.96e-8` (`= -1 / 2^24`) instead of `0`. (Without a table the query is constant-folded through the scalar path and was already correct.)

### Changes

- Add `float_to_fixed_point`, which computes `trunc(n * 2^24)` reduced modulo `2^128` (wrapping) directly from the `f64` bit pattern. This makes the conversion a group homomorphism consistent with the wrapping accumulation, so large finite values whose mathematical sum is representable now yield the correct result. Working in the integer domain also avoids forming the intermediate `f64` product `n * FLOAT_SCALE`, which could itself overflow to infinity.
- Replace the remaining raw `as` casts on the finalize path with `cast_lossy`, adding the `i128 -> f64` and `f64 -> f32` impls in `mz_ore::cast`.
- Add Rust unit tests for `float_to_fixed_point` (agreement with the old cast in range, truncation toward zero, subnormals, cancellation of out-of-range values, full accumulate/finalize path) and a `float.slt` regression test mirroring the issue.

Note: sums whose **mathematical result** is itself outside the representable fixed-point range (e.g. `SUM` of `1e31 + 1e31 = 2e31`) remain approximate — the fixed-point domain simply cannot represent them — and continue to log the existing overflow warning. This change makes the *representable* cases (including exact cancellation) correct.

### Tip

`cargo test -p mz-compute --lib render::reduce::tests`

https://claude.ai/code/session_01HXYJ6TAz8GawjxdC6TNr8R

---
_Generated by [Claude Code](https://claude.ai/code/session_01HXYJ6TAz8GawjxdC6TNr8R)_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6086,6 +6086,7 @@ dependencies = [
  "mz-storage-types",
  "mz-timely-util",
  "mz-txn-wal",
+ "num-traits",
  "prometheus",
  "proptest",
  "prost",
--- a/src/compute/Cargo.toml
+++ b/src/compute/Cargo.toml
@@ -39,6 +39,7 @@ mz-storage-operators = { path = "../storage-operators" }
 mz-storage-types = { path = "../storage-types" }
 mz-timely-util = { path = "../timely-util" }
 mz-txn-wal = { path = "../txn-wal" }
+num-traits.workspace = true
 prometheus.workspace = true
 prost.workspace = true
 scopeguard.workspace = true
--- a/src/compute/src/render/reduce.rs
+++ b/src/compute/src/render/reduce.rs
@@ -12,7 +12,6 @@
 //! Consult [ReducePlan] documentation for details.
 
 use std::collections::BTreeMap;
-use std::sync::LazyLock;
 
 use columnation::{Columnation, CopyRegion};
 use dec::OrderedDecimal;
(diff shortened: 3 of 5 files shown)


DAlperin (approved):
LGTM, thanks

PR #36782, opened 2026-05-29
adapter: bump with_0dt_deployment_max_wait default to 1 year

aljoscha:
This matches the default we use in LD, and is a safer choice as the comment explains.

Contributes to SQL-322


def- (approved):

aljoscha:
tyty! 🙇

PR #36784, opened 2026-05-29
docs: 1 year wait time for upgrading

def-:
Follow-up to https://github.com/MaterializeInc/materialize/pull/36782


kay-kim (commented):
Q: So ... this change won't go out until next Thursday/Friday's release?

def-:
It just missed the cut-off for 26.27, so going to be in 26.28, which will be released on June 12 for self-managed. If we want it out earlier, we'd have to cut a new rc for 26.27.

kay-kim:
Thanks much. ❤️   Just wanted to know which release so that we don't mere before then.

aljoscha (approved):

PR #36789, opened 2026-05-29
Mark v26.26 as released

def-:
We also need to trigger the self-managed deploy again. Edit: Doing that now: https://buildkite.com/materialize/publish-helm-charts/builds/168


bosconi (approved):
Looks good.

PR #36790, opened 2026-05-29
release: fix mark_release_complete for python-frontmatter 1.3.0

def-:
The `complete-release` CI job is failing with:

```
File ".../materialize/release/mark_release_complete.py", line 37, in main
    frontmatter.dump(metadata, f, sort_keys=False)
TypeError: a bytes-like object is required, not 'str'
```

`python-frontmatter` was bumped 1.1.0 → 1.3.0 in #36710 (2026-05-25). In 1.1.0, `frontmatter.dump()` encoded the content to bytes (`.encode(encoding)`) before writing, so the doc file was opened in binary mode (`"wb"`). In 1.3.0, `dump()` writes a `str` directly and expects a text-mode file, so writing to the `"wb"` handle now raises `TypeError`.

### DIFF
--- a/misc/python/materialize/release/mark_release_complete.py
+++ b/misc/python/materialize/release/mark_release_complete.py
@@ -33,10 +33,10 @@ def main():
     metadata["patch"] = int(args.patch)
     if "rc" in metadata:
         del metadata["rc"]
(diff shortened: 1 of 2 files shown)


bosconi (approved):
Looks good.

PR #36791, opened 2026-05-29
testdrive: Fix flaky savings check in expected_group_size_tuning

def-:
Flake seen in https://buildkite.com/materialize/nightly/builds/16627

Test run: https://buildkite.com/materialize/nightly/builds/16634


bosconi (approved):
Looks good.

PR #36795, opened 2026-05-29
SS-16: E2E testing for AWS Glue schema registry

martykulma:
Adds E2E testing for AWS Glue schema registry with a Kafka source.

AWS Glue supports multiple registries, including a default. Schemas exist within registries. mzcompose.py takes care of registry setup and teardown, testdrive takes care of schema creation (glue-schema-create).

- local/test pipeline use [Moto](https://github.com/getmoto/moto) to emulate schema registry (default workflow)
- nightly pipeline uses AWS (aws workflow)

This is not an exhaustive list of tests, we are missing availability/connectivity tests (MZ restart, glue schema registry unreachable, etc.).  I'll draft a separate PR for those.

### DIFF
--- a/ci/nightly/pipeline.template.yml
+++ b/ci/nightly/pipeline.template.yml
@@ -949,6 +949,23 @@ steps:
           - ./ci/plugins/mzcompose:
               composition: aws-localstack
 
+      - id: aws-glue-schema-registry-real
+        label: AWS Glue Schema Registry (Real)
+        depends_on: build-aarch64
+        timeout_in_minutes: 30
+        retry:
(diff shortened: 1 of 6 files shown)


patrickwwbutler (approved):
LGTM Seems pretty comprehensive to me, though it feels inevitable that we'll have to add to it eventually as we discover new ways to break things

martykulma:
> LGTM Seems pretty comprehensive to me, though it feels inevitable that we'll have to add to it eventually as we discover new ways to break things

💯 - TFTR!

PR #36797, opened 2026-05-29
workload-replay: anonymize subsource child keys, broaden verify

jasonhernandez:
**Fourth in the stack** — base `workload-anonymize-tests` (#36749). Found by actually running the tool against a real Materialize Cloud capture.

## The bug

Captures with Postgres/MySQL CDC sources store each source's subsources (`children`) in a dict **keyed by the child's fully-qualified `database.schema.name`**. That key was built during the mapping pass from `child["database"]`/`child["schema"]`, which are only remapped *later* (pass 2) — so the key kept the **original database and schema names**, even though the child's own fields and every other position were correctly anonymized.

Real example from the capture, anonymized output **before** this fix:
```
children:
  qa_canary_environment.public_loadgen.child_25:   # <-- original db + schema leaked in the key
    name: child_25
    database: db_5      # correctly anonymized
    schema: schema_18   # correctly anonymized
```
After: `db_5.schema_18.child_25`.

8 distinct original database/schema names leaked this way in a 5-minute prod capture.

## Why verify didn't catch it

`verify_anonymized` only scanned `create_sql`/`sql` *values*, never structural dict keys — so it reported all-clear. This PR broadens it to also scan dict keys, with two guards to avoid false positives (both learned the hard way while writing it):

1. **Reserved format keys** (`RESERVED_FORMAT_KEYS`): the workload file's own field names (`sql`, `cluster`, `transaction_id`, `tables`, …). Without this, a user column named `transaction_id` (which maps to `column_N`) makes verify flag every query record's own `transaction_id` field. Their *values* are still checked — only the structural key name is exempt.
2. **Scalar values are not identifier-scanned.** A kept literal — e.g. a column default `'secret note'` under `--no-literals` — can contain a word matching a renamed column (`note`) without being a leak. SQL text is still scanned (as before).

## Validation

- Re-ran against the real prod capture: non-keyword identifier survivors dropped **8 → 0** (the one remaining match is the legitimate `transaction_id` *format key*, not data). `--verify` passes and writes; before the fix it correctly **refused to write**.
- 4 new regression tests (19 total): end-to-end child-key anonymization, verify catching a leaked child key, and the two false-positive guards.
- `bin/fmt`, `ruff` clean.

### DIFF
--- a/misc/python/materialize/cli/mz_workload_anonymize.py
+++ b/misc/python/materialize/cli/mz_workload_anonymize.py
@@ -80,6 +80,61 @@ def keywords() -> set[str]:
     return result
 
 
+# Keys that are part of the workload file format itself, not user identifiers.
+# The verify pass must not treat these as leaks when a user object happens to
+# share a name with one of them (e.g. a column named `transaction_id`). Their
+# *values* are still checked; only the structural key name is exempt.
+RESERVED_FORMAT_KEYS = frozenset(
+    {
+        # top level
+        "databases",
+        "clusters",
+        "queries",
+        "mz_workload_version",
+        # schema-level containers
+        "tables",
+        "views",
+        "materialized_views",
+        "indexes",
+        "types",
+        "connections",
+        "sources",
+        "sinks",
+        # object / column fields
+        "create_sql",
+        "name",
+        "type",
+        "schema",
+        "database",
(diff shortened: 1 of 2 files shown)


jasonhernandez:

PR #36798, opened 2026-05-29
Fix issue where SHOW CREATE SINK returns a version which can not be used to create a new sink

peterdukelarsen:
### Motivation

Closes https://linear.app/materializeinc/issue/SS-160/show-create-sink-includes-internal-version-n-round-trip-fails

### Description

Scrubs the version output by `SHOW CREATE SINK`. Test simply asserts the version isn't output

### Verification

1. I confirmed locally that the round trip with the internal version fails as described in the ticket.
2. I wrote the fix/test to scrub the version.
3. I ran the test against the old version of the code locally to confirm it would have failed before.


patrickwwbutler (approved):
LGTM! congrats on first issue!

PR #36799, opened 2026-05-29
workload-replay: require the parser by default for query literals

jasonhernandez:
**Fifth in the stack** — base `workload-anonymize-subsource-keys` (#36797).

## Change

Flips the anonymizer to **fail-closed** on literal redaction. Previously, if the `mz-sql-anonymize` binary was missing (or an individual captured statement didn't parse), the tool silently fell back to a regex that only redacts single-quoted strings — leaving **numbers, dollar-quoted strings, and comments** in query SQL exposed, with just a warning. For a privacy tool, fail-open is the wrong default.

New `--require-parser` flag, **on by default**:
- Parser binary unavailable → **error**, with build instructions, instead of regex fallback.
- Parser present but some captured queries don't parse → **error**, reporting how many.
- `--no-require-parser` opts back into the regex fallback (which now warns it's active).

Also moved output-target resolution to the top of `main()` so an invalid invocation (`no -o`, no `--in-place`) fails immediately, independent of the parser check.

## Behavior summary

| Situation | Before | After (default) | `--no-require-parser` |
|---|---|---|---|
| Binary missing | regex + warning | **error** | regex + warning |
| Statement doesn't parse | regex for that stmt | **error** | regex for that stmt |
| Binary present, all parse | parser | parser | parser |

## Testing

- 20 tests pass. New `test_require_parser_errors_without_binary` covers the fail-closed default; the existing regex-path tests now pass `--no-require-parser` via the `run_tool` helper (which defaults to it so tests are deterministic whether or not the binary is built).
- `bin/fmt`, `ruff` clean.
- README updated with the new default and flag.

### DIFF
--- a/misc/python/materialize/cli/mz_workload_anonymize.py
+++ b/misc/python/materialize/cli/mz_workload_anonymize.py
@@ -581,10 +581,12 @@ def main() -> int:
     # strings, intervals, dollar-quoted and escape strings) where the regex only
     # caught single-quoted strings.
     #
-    # By default (--require-parser) the regex is NOT an acceptable substitute:
-    # if the parser binary is unavailable, or any individual statement does not
-    # parse, the tool errors rather than silently emitting weaker redaction.
-    # --no-require-parser opts into the regex fallback for those cases.
+    # --require-parser gates only the wholesale case: if the parser binary is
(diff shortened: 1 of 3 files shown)


jasonhernandez:

PR #36801, opened 2026-05-29
workload-replay: anonymize SQL on the AST, not by regex

jasonhernandez:
**Sixth in the stack** — base `workload-anonymize-require-parser` (#36799). The big one: replaces text-regex SQL rewriting with AST-based rewriting using Materialize's own parser.

## Why

Running the tool against a **real production capture** (565 queries) showed the regex identifier substitution **corrupted ~12% of queries** — 65 failures (47 `SELECT`, 18 `FETCH`). The raw SQL parsed 100%; our own regex broke it by matching identifiers as substrings and inside string literals. The strict `--require-parser` mode then (correctly) refused to emit the broken output, so the tool couldn't complete on a real capture at all.

Text substitution was never the right tool — and we already shell out to the real parser. So do all rewriting on the AST.

## What changed

**`mz-sql-anonymize` (Rust)** now parses each statement and rewrites the AST via `VisitMut`:
- **Identifier renaming as whole tokens.** Reaches object/cluster/type references by overriding the visitors for `Raw`'s `AstInfo` associated types (`visit_item_name_mut`, `visit_cluster_name_mut`, `visit_data_type_mut`, …), whose generic defaults are no-ops. No substring corruption, no in-string rewrites, no word-boundary/case guesswork. (Mz identifiers are case-sensitive → exact match is correct.)
- **Query-literal redaction** on the AST — numbers, hex strings, intervals included (the regex only caught single-quoted strings).
- **Config preserved**: `CREATE CLUSTER`/`CLUSTER REPLICA`/`ALTER CLUSTER` and `SET`/`RESET`/`SET TRANSACTION`/`ALTER SYSTEM` keep their literals (sizes, timeouts) — replay needs them.

Protocol: `{mapping, rename_identifiers, redact_literals, statements}` → rewritten SQL per statement (or `null` if unparseable).

**Python** sends all cluster/DDL/query SQL through the helper, and:
- still scrubs **DDL `create_sql` literals with the blanket regex**, because option strings (broker addresses, hosts) are *typed AST fields* the visitor — and the engine's own redacted Display — don't treat as redactable literals;
- applies the structural identifier mapping (column types, child schema/db, query routing) directly (not SQL);
- falls back to regex only when the binary is unavailable (gated by `--require-parser`) or a statement doesn't parse.

`verify` now also exempts preserved config-statement literals (matching the helper), so a kept `SET … = '5s'` isn't flagged.

## Validation (same production capture, re-captured)

| Metric | Regex (before) | AST (after) |
|---|---|---|
| Anonymized queries that fail to re-parse | **65 / 565 (12%)** | **0 / 623** |
| Identifier leaks (non-keyword originals) | — | 0 (only the `transaction_id` *format key*) |
| Query numeric literals remaining | leaked | 0 |
| Cluster SIZE / SET timeout | preserved | preserved |

Default mode (`--require-parser`) now **completes** on a real capture, where before it errored.

## Tests
- 8 Rust unit tests (token-level rename, qualified refs, no substring/in-string corruption, query-literal redaction incl. numbers, cluster/SET preservation, parse-failure → null).
- 22 Python tests, incl. a skip-if-unbuilt end-to-end AST rename test and the config-literal verify exemption.
- `bin/fmt`, `clippy`, `ruff`, test-attribute lint clean; no `Cargo.lock` version bumps (adds `serde`, already present).

## Note for reviewers
This supersedes #36746's query-only redaction approach. The stack tells the story (heuristic → parser-literals → tests → subsource fix → require-parser → full AST); if preferred I can squash the SQL-rewriting PRs before merge.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7815,6 +7815,7 @@ version = "0.0.0"
 dependencies = [
  "mz-ore",
  "mz-sql-parser",
+ "serde",
  "serde_json",
 ]
 
--- a/misc/python/materialize/cli/mz_workload_anonymize.py
+++ b/misc/python/materialize/cli/mz_workload_anonymize.py
@@ -37,26 +37,42 @@ def _locate_redactor() -> list[str] | None:
     return None
 
 
-def redact_literals_via_parser(sqls: list[str]) -> list[str | None] | None:
-    """Redact literals in each SQL string using Materialize's own parser.
+def anonymize_sql_via_parser(
+    sqls: list[str],
+    mapping: dict[str, str],
+    rename_identifiers: bool,
+    redact_literals: bool,
+) -> list[str | None] | None:
+    """Rename identifiers and/or redact literals in each SQL string via the AST.
 
-    Returns a list aligned with the input, where each element is the redacted
-    SQL or None if that statement could not be parsed. Returns None for the
-    whole batch if the helper binary is unavailable or errors, signaling the
-    caller to fall back to regex-based redaction.
+    Uses the `mz-sql-anonymize` helper, which parses each statement with
+    Materialize's own parser, renames identifier tokens per `mapping`, and (when
+    `redact_literals`) replaces literal values with `'<REDACTED>'`. Doing this on
+    the AST avoids the corruption a text regex causes (substring matches,
+    in-string rewrites, broken syntax).
+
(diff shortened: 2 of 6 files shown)


jasonhernandez:

PR #36803, opened 2026-05-29
workload-replay: harden and AST-anonymize captured workloads

jasonhernandez:
`bin/mz-workload-anonymize` scrubs identifiers and literals from captured workloads so they can be shared. It relied on text-regex substitution, which both **leaked sensitive data** and **corrupted SQL**. This moves all SQL rewriting onto Materialize's own parsed AST. The regex path is gone: the parser is now required, and the tool refuses to write if it's missing or any statement fails to parse.

## New crate `src/sql-anonymize` (`mz-sql-anonymize`)
Reads a JSON request and rewrites each statement on the AST via `VisitMut`:
- **Renames identifiers as whole tokens**, reaching object/cluster/type references that the generic visitors miss by overriding `visit_item_name_mut` & friends. No substring corruption, no in-string rewrites. Builtins and query-local names (CTEs, aliases) are left alone.
- **Redacts query literals** to `'<REDACTED>'` (which replay binds as `NULL`), and **numbers where they're data** (queries and view/MV/index bodies) to the neutral literal `1`. Numbers in option positions (cluster sizes, ports, replication factors, column defaults) and intervals/durations are kept, since replay needs them valid.
- **Consistently renames DDL string values** (Kafka topics, upstream databases, connection hosts) to a stable salted `redacted_<hash>` token, so each original lines up everywhere replay cross-references it. The salt is per-run and random, so the tokens can't be reversed by a dictionary attack on the shared output. Inline Avro/Protobuf schemas are anonymized in place so they stay decodable.
- **Preserves config literals** (`CREATE`/`ALTER CLUSTER`, `CLUSTER REPLICA`, `SET`/`RESET`/`ALTER SYSTEM`): sizes, timeouts, isolation, that replay needs and aren't sensitive.

It also has a **collect mode** that reports object names referenced but not in the mapping, so the Python tool can name objects created or dropped *during* the capture window (dbt-built views, transient clusters) that the catalog snapshot never saw.

## Built-in keep-lists
`system_catalog_identifiers.txt` and `load_generator_identifiers.txt` list built-in object/column/function/type names that are never renamed (renaming them breaks replay, e.g. a view over `mz_internal.mz_role_auth` must keep `password_hash`). Each file's header records the SQL it was generated from.

## Anonymizer (Python)
Drives the helper for all SQL, applies the structural identifier mapping to non-SQL fields (column types, child schema/db, query routing) directly, renames consistently (the same name always maps the same way), rebuilds source-child dict keys from the mapped names (these previously leaked the originals), and redacts bound query parameters (`$1, $2, …`) to `<REDACTED>`, which replay binds as `NULL`.

## Safety net
A **verify pass** (`--verify`, on by default) re-scans the output for surviving original identifiers and non-redacted literals (including numeric-string values like SSNs and bound params), and refuses to write if anything leaks. Output now requires an explicit `-o`/`--in-place` rather than silently overwriting the input.

## Tests
Rust unit tests cover the rewrite behaviors above (whole-token rename, literal/number redaction, the body-vs-option distinction, config preservation, collect mode). Python tests cover end-to-end anonymization, leak regressions (connection/sink literals, subsource keys, numeric-string predicates, bound params), and the verify guards. A session fixture builds the helper so the end-to-end tests always run instead of being skipped when the binary is absent.


def- (changes requested):
> requires the parser by default (--require-parser); falls back to regex, with a warning, only when the binary is unavailable or a statement doesn't parse.

I would remove the regex-based approach instead. There is no reason to keep it.

Bound query parameter values currently don't get anonymized, but they probably should be.

def- (approved):

def-:
I have reworked parts so have to ask for another review

jasonhernandez (commented):
@def- I approve of your commit on top of this. I'm going to merge it! Thanks for working together on this!

PR #36804, opened 2026-05-29
Update columnar, timely, and differential dependencies

antiguru:
## Motivation

Bump the dataflow dependency family to their latest releases:

| crate | old | new |
| --- | --- | --- |
| `columnar` | 0.12.1 | 0.13.0 |
| `timely` | 0.29.0 | 0.30.0 |
| `differential-dataflow` | 0.23.0 | 0.24.0 |
| `differential-dogs3` | 0.23.0 | 0.24.0 |

These releases carry breaking changes that ripple through the batcher/arrange machinery and the timely communication layer.

## What changed

**columnar 0.13** — `AsBytes` now requires `SLICE_COUNT` and `get_byte_slice`, and the `chain` helper was removed. Updated the manual `AsBytes` impls for `Overflows`, `Timestamps`, and `Rows` (the `Rows` impl now relies on the default `as_bytes` built from `get_byte_slice`).

**differential 0.24** — removes the `Batcher::Input` / `push_container` API and the `merge_batcher::container` compatibility layer (`InternalMerge`, `InternalMerger`). The batcher now consumes already-chunked input via `PushInto<Output>`, `seal` returns `(chain, description)`, and chunking moves into the operator via a separate `ContainerBuilder`/chunker. `MergeBatcher` collapses to a single `Merger` type parameter.

- Reimplemented `ColInternalMerger` as a chunk-list `Merger` over `ColumnationStack` chunks (the old `InternalMerge`-based impl is gone). The `merge` drain phase keeps differential 0.23's zero-copy behavior: a partially-consumed head is copied (or swapped when untouched) and the remaining full chunks of the longer chain pass through to the output without copying; consumed chunks recycle into the stash. An earlier draft copied drained tails element-by-element, which regressed the MinMax feature benchmark by ~20% wallclock and +18% clusterd memory.
- Migrated the pageable `ColumnMergeBatcher` to the new `Batcher` trait.
- `mz_arrange` / `mz_arrange_core` / `consolidate_pact` / `consolidate_named*` / `consolidate_and_pack` gain an explicit `Chu` chunker type parameter, mirroring upstream's `arrange_core`. Call sites name the chunker: `ColumnationChunker<_>` for `Vec` input, `batcher::Chunker<_>` for `Column` input, and `batcher::ColumnChunker<_>` for the paged `Column` path.
- Updated `MergeBatcherWrapper` (temporal bucketing), the upsert-v2 source stash, and an interchange test to drive the chunker explicitly.
- `TraceBox.trace` is now read via `trace()`; `ShutdownButton::press_on_drop` was removed, replaced by a small local press-on-drop guard.

**timely 0.30** — adds a spill-policy parameter to the process allocators and bundles refill/spill/log hooks into `Hooks` for `initialize_networking_from_sockets`. `BytesRefill` now yields `Send` buffers, so `LgallocHandle` gains an `unsafe impl Send` (it exclusively owns its allocation).

https://claude.ai/code/session_01CrNqefrHBNbfJetgjRHsmH

---
_Generated by [Claude Code](https://claude.ai/code/session_01CrNqefrHBNbfJetgjRHsmH)_

### DIFF
--- a/src/cluster/src/communication.rs
+++ b/src/cluster/src/communication.rs
@@ -164,7 +164,7 @@ where
     }
 
     // Choose the intra-process allocator flavor up front.
-    // TODO(CLU-99): wire our zero-copy pager into the timely 0.30 spill policy
+    // TODO(CLU-99): wire our zero-copy pager into the timely spill policy
     // instead of passing `None` here.
     let process_allocators = if enable_zero_copy_binary {
         ProcessBuilder::new_bytes_vector(workers, refill.clone(), None)
@@ -172,9 +172,9 @@ where
         ProcessBuilder::new_typed_vector(workers, refill.clone(), None)
     };
 
-    // Since timely 0.30 the refill, spill policy, and per-thread logger hook are
-    // bundled into `Hooks`. We don't install a comm logger, and the spill policy
-    // is left unset pending CLU-99.
+    // Timely bundles the refill, spill policy, and per-thread logger hook into
+    // `Hooks`. We don't install a comm logger, and the spill policy is left
+    // unset pending CLU-99.
     let hooks = Hooks {
         log_fn: Arc::new(|_| None),
         refill,
--- a/src/cluster/src/communication.rs
+++ b/src/cluster/src/communication.rs
@@ -84,6 +84,7 @@ use mz_ore::cast::CastFrom;
 use mz_ore::netio::{Listener, Stream};
 use mz_ore::retry::Retry;
 use regex::Regex;
+use timely::communication::Hooks;
 use timely::communication::allocator::ProcessBuilder;
 use timely::communication::allocator::generic::AllocatorBuilder;
 use timely::communication::allocator::zero_copy::bytes_slab::BytesRefill;
@@ -163,20 +164,24 @@ where
     }
 
     // Choose the intra-process allocator flavor up front.
+    // TODO(CLU-99): wire our zero-copy pager into the timely 0.30 spill policy
+    // instead of passing `None` here.
     let process_allocators = if enable_zero_copy_binary {
-        ProcessBuilder::new_bytes_vector(workers, refill.clone())
+        ProcessBuilder::new_bytes_vector(workers, refill.clone(), None)
     } else {
-        ProcessBuilder::new_typed_vector(workers, refill.clone())
+        ProcessBuilder::new_typed_vector(workers, refill.clone(), None)
     };
 
-    match initialize_networking_from_sockets(
-        process_allocators,
-        sockets,
-        process,
-        workers,
+    // Since timely 0.30 the refill, spill policy, and per-thread logger hook are
+    // bundled into `Hooks`. We don't install a comm logger, and the spill policy
+    // is left unset pending CLU-99.
+    let hooks = Hooks {
+        log_fn: Arc::new(|_| None),
         refill,
-        Arc::new(|_| None),
-    ) {
+        spill: None,
+    };
+
+    match initialize_networking_from_sockets(process_allocators, sockets, process, workers, hooks) {
         Ok((tcp_builders, guard)) => {
             info!(process = process, "successfully initialized network");
             let builders = tcp_builders
--- a/src/compute/src/render.rs
+++ b/src/compute/src/render.rs
@@ -180,9 +180,7 @@ pub use join::LinearJoinSpec;
 
 /// Guard that presses a differential [`ShutdownButton`] when dropped.
 ///
-/// differential-dataflow 0.24 removed `ShutdownButton::press_on_drop`, so we
-/// reconstruct the press-on-drop behavior locally: dropping this guard releases
-/// the imported trace's capabilities.
+/// Dropping this guard releases the imported trace's capabilities.
 struct PressOnDrop<T>(ShutdownButton<T>);
 
 impl<T> Drop for PressOnDrop<T> {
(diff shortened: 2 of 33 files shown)


src/timely-util/src/columnar/merge_batcher.rs
```diff
@@ -175,6 +173,26 @@ fn account_chunk<C: Columnar>(entry: &PagedColumn<C>) -> (usize, usize, usize, u
...
+{
+    /// Accept a chunk of already-sorted, consolidated updates (produced by the
+    /// operator's chunker), page it out, and fold it into the chains.
+    fn push_into(&mut self, mut chunk: Column<(D, T, R)>) {
+        if chunk.borrow().len() == 0 {
+            return;
+        }
+        let pager = self.pager();
+        let paged = pager.page(&mut chunk);
+        self.insert_chain(VecDeque::from([paged]));
```
antiguru: We need to ensure that `paged` is consolidated and sorted. Where is the code that does this?

src/timely-util/src/columnar/merge_batcher.rs
```diff
@@ -203,25 +219,11 @@ where
...
-    }
-
-    fn seal<B: Builder<Input = Self::Output, Time = Self::Time>>(
+    fn seal(
         &mut self,
         upper: Antichain<Self::Time>,
-    ) -> B::Output {
+    ) -> (Vec<Self::Output>, Description<Self::Time>) {
         let pager = self.pager();
-        // Finish chunker, fold any tail chunks in.
```
antiguru, line 263: Where do we ensure that our staging area is empty now?

src/cluster/src/communication.rs
```diff
@@ -164,19 +165,20 @@ where
...
-        process,
-        workers,
+    // Since timely 0.30 the refill, spill policy, and per-thread logger hook are
+    // bundled into `Hooks`. We don't install a spill policy or comm logger.
+    let hooks = Hooks {
+        log_fn: Arc::new(|_| None),
         refill,
-        Arc::new(|_| None),
-    ) {
+        spill: None,
```
antiguru, line 181: File a linear issue to wire up our pager with `spill` here.

src/timely-util/src/columnar/merge_batcher.rs
antiguru: The chunk arrives here already sorted and consolidated. Since differential 0.24 the batcher no longer chunks its own input — that's the operator's `ContainerBuilder`/chunker. For this pageable batcher that chunker is `columnar::batcher::ColumnChunker` (wired up via the new `ArrangeChunker` impl), which sorts each input container and folds equal `(data, time)` runs before emitting a chunk to `push_into`. `insert_chain`/`merge_by` then only have to *preserve* that order across chunks, not re-establish it — same invariant the framework relies on for differential's own `VecMerger`.

I've added a doc comment on `push_into` (pushed) spelling this out and pointing at the chunker.

---

src/timely-util/src/columnar/merge_batcher.rs
antiguru: The staging area moved out of the batcher and into the operator along with the chunker. The operator drains it with `chunker.finish()` right before calling `seal` — see `consolidate_pact` in `timely-util/src/operator.rs` and differential's own `arrange_core`, both of which do `while let Some(chunk) = chunker.finish() { batcher.push_into(...) }` before sealing. So by the time `seal` runs, every input record has already reached the batcher via `push_into` and there's nothing left to flush internally. Added a comment on `seal` (pushed) noting this.

---

src/cluster/src/communication.rs
antiguru: Filed [CLU-99](https://linear.app/materializeinc/issue/CLU-99/wire-up-zero-copy-pager-with-timely-030-spill-policy) and added a `TODO(CLU-99)` here (pushed).

---

antiguru:
Marking as ready for review, but needs another round of reviews, and nightly tests.

src/compute/src/render.rs
```diff
@@ -176,6 +178,19 @@ mod top_k;
 pub use context::CollectionBundle;
 pub use join::LinearJoinSpec;
 
+/// Guard that presses a differential [`ShutdownButton`] when dropped.
+///
+/// differential-dataflow 0.24 removed `ShutdownButton::press_on_drop`, so we
+/// reconstruct the press-on-drop behavior locally: dropping this guard releases
+/// the imported trace's capabilities.
```
DAlperin: style note that I've tried to beat out of my claude (only somewhat sucesfully) but claude really likes to reference things that were previously changed or deleted. This generally is not helpful to future human readers. 


```suggestion
/// Guard that presses a differential [`ShutdownButton`] when dropped.
///
/// Dropping this guard releases the imported trace's capabilities.
```

src/interchange/src/envelopes.rs
```diff
@@ -150,11 +152,20 @@ mod tests {
     fn batch_from_tuples(
         mut tuples: Vec<((String, String), u64, Diff)>,
         upper: u64,
-    ) -> <ValBuilder<String, String, u64, Diff> as differential_dataflow::trace::Builder>::Output
-    {
+    ) -> <ValBuilder<String, String, u64, Diff> as Builder>::Output {
+        // Since differential-dataflow 0.24 the batcher consumes already-chunked
+        // input via `PushInto` and chunking is the caller's responsibility.
```
DAlperin: for another example

DAlperin (approved):
Changes make sense to me

src/compute/src/render.rs
antiguru: Applied the suggestion and swept the rest of the diff for the same pattern in e21a798946 — all "since differential 0.24 / timely 0.30" style comments now describe the current state only.

src/interchange/src/envelopes.rs
antiguru: Fixed in e21a798946.

antiguru:
Thanks for the review!

PR #36805, opened 2026-05-30
ci-builder: Bump cargo-* versions

def-:
Inspired by https://materializeinc.slack.com/archives/C01LKF361MZ/p1780090028190879 to fix cargo-deny issues in most recent version, discovered by @peterdukelarsen


antiguru (approved):
Thank you!

PR #36807, opened 2026-05-30
ci: parallelize dev.materialize.com docs build and upload

def-:
Closes https://linear.app/materializeinc/issue/QAR-76/deploy-devsite-build-step-duration-has-increased-noticeably


antiguru (approved):
Looks good, thank you.

PR #36810, opened 2026-05-30
txn-wal: deasync txns_progress_frontiers

antiguru:
Convert `txns_progress_frontiers` from an async `AsyncOperatorBuilder` to a synchronous `OperatorBuilderRc`, and remove the async implementation entirely.

Third attempt: #36537/#36538 were reverted and #36548 never landed. The SQL-299 data-loss bug is fixed by construction — buffered passthrough data is emitted at the pre-activation capability, before any capability downgrade. The PER-4 stall fix (retain the last remap entry after the remap input closes) and the replica-targeted-select-abort fix (gate the passthrough frontier on `waiting_for_remap`) are preserved.

Tests: three regression tests (SQL-299, PER-4, `SELECT AS OF MAX`), each confirmed to fail when its fix is reverted, plus an oracle-free fuzz test asserting no data loss and no premature shutdown over random interleavings. `as_of_until` is relaxed to the deasync stream-timestamp contract.

Design: `doc/developer/design/20260529_txn_wal_frontiers_deasync.md`.

### DIFF
--- a/src/txn-wal/src/operator.rs
+++ b/src/txn-wal/src/operator.rs
@@ -1313,11 +1313,39 @@ mod tests {
                 Event::Progress(progress) => Either::Left(progress),
                 Event::Messages(ts, data) => Either::Right((ts, data)),
             });
-        let expected = vec![
-            (3, vec![("2".to_owned(), 3, 1), ("3".to_owned(), 3, 1)]),
-            (3, vec![("4".to_owned(), 4, 1)]),
+        // Aggregate the captured records, ignoring the stream-level
+        // timestamp on each batch. The operator emits each container at
+        // whatever capability it currently holds (which is determined by
+        // its scheduling cadence and the upstream frontiers it has
+        // observed), so the per-batch `ts` is not deterministic and not
+        // part of the operator's contract. Per-record `(key, time, diff)`
+        // tuples are what callers see, and the differential invariant
+        // (stream `ts <= record time`) is checked separately below.
+        let mut actual_records: Vec<(String, u64, i64)> = actual_events
+            .iter()
+            .flat_map(|(_ts, data)| data.iter().cloned())
+            .collect();
+        actual_records.sort();
+        let expected_records: Vec<(String, u64, i64)> = vec![
+            ("2".to_owned(), 3, 1),
+            ("3".to_owned(), 3, 1),
+            ("4".to_owned(), 4, 1),
         ];
-        assert_eq!(actual_events, expected);
+        assert_eq!(actual_records, expected_records);
+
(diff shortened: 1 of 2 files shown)


src/txn-wal/src/operator.rs
```diff
@@ -244,166 +251,181 @@ where
...
-                if r.physical_upper.less_equal(cap.time()) {
-                    if cap.time() < &r.logical_upper {
-                        cap.downgrade(&r.logical_upper);
+        };
+        // Whether the remap input has reached the empty antichain.
+        let mut remap_closed = false;
+
+        move |frontiers| {
+            if shutdown_handle.local_pressed() {
+                capability = None;
```
aljoscha: I was wondering if this is correct? Don't we get spurious frontier downgrades from this, downstream?

src/txn-wal/src/operator.rs
```diff
@@ -244,166 +251,181 @@ where
...
-            match event {
-                // NB: Ignore the data_cap because this input is disconnected.
-                AsyncEvent::Data(_data_cap, mut data) => {
-                    // NB: Nothing to do here for `until` because both the
-                    // `shard_source` (before this operator) and
-                    // `mfp_and_decode` (after this operator) do the necessary
-                    // filtering.
+            debug!("{} remap {:?} remap_closed={}", name, remap, remap_closed);
+
+            // Emit buffered passthrough data at the current (pre-downgrade)
```
aljoscha: I found "buffered data" slightly confusing here because the operator itself is not buffering it, it's just consuming it's input. Which yes, is buffered and yada yada, but still.

aljoscha (approved):
I read the doc, to get the reasoning and facts and the operator impl. They look good! Only skimmed the tests.

src/txn-wal/src/operator.rs
antiguru: Good catch — you were right. `builder_async` holds capabilities until *all* workers have pressed (on the local press it only stops scheduling the logic future and draining inputs); dropping on `local_pressed` alone could let the downstream frontier advance past this worker's discarded input while other workers' instances still feed downstream. Fixed in 4b10171d7b: the operator now mirrors the two-phase shutdown via `build_reschedule` — wedge (retain capability, leave inputs undrained, reschedule) on the local press, drop and drain once `all_pressed()`.

src/txn-wal/src/operator.rs
antiguru: Fair — reworded to "pending input" / "in-flight" in 4b10171d7b.

antiguru:
Thanks for the review!

PR #36811, opened 2026-05-30
sqlsmith: allowlist date/timestamp out-of-range parse errors

antiguru:
### Motivation

SQLsmith flagged a query that casts an out-of-range value to `date`:

```
ERROR: "14714-12-31 18:01:00+00 BC" is out of range for type date
```

Rejecting an out-of-range or malformed value is correct behavior — this is
just SQLsmith generating nonsense input — so the message should be on the
known-errors allowlist rather than treated as a bug.

### Description

Add two entries to the SQLsmith `known_errors` allowlist:

- `is out of range for type date`
- `is out of range for type timestamp` (also covers `timestamptz`)

This matches how the other out-of-range parse errors (e.g. numeric) are
already handled.

Per @def-'s review, this drops the earlier approach of changing the
SQLSTATE codes in `src/adapter/src/error.rs`; we just allowlist these
messages instead.

### DIFF
--- a/misc/python/materialize/sqlsmith.py
+++ b/misc/python/materialize/sqlsmith.py
@@ -101,6 +101,8 @@ known_errors = [
     "coalesce types text and text list cannot be matched",  # Bad typing for ||
     "coalesce types text list and text cannot be matched",  # Bad typing for ||
     "is out of range for type numeric: exceeds maximum precision",
+    "is out of range for type date",
+    "is out of range for type timestamp",
     "CAST does not support casting from ",  # TODO: Improve type system
     "SET clause does not support casting from ",  # TODO: Improve type system
     "coalesce types integer and interval cannot be matched",  # TODO: Implicit cast from timestamp to date in (date - timestamp)


antiguru:
Observed in https://buildkite.com/materialize/nightly/builds/16645#019e76af-7b03-4a9e-b42f-8df826d12f66

def-:
There are probably WAY more errors like that, I didn't care much about the error codes we use, and instead just ignored the known errors with https://github.com/MaterializeInc/materialize/blob/bcae6e708752bf7adffa8fc8ea085d8d63ee6b01/misc/python/materialize/sqlsmith.py#L13

antiguru:
Switched to the allowlist-only approach per your suggestion, @def-. Dropped the SQLSTATE changes in `error.rs`/`strconv.rs` (and the regression test) and force-pushed a single commit that just adds `is out of range for type date` / `is out of range for type timestamp` to the `known_errors` allowlist. Updated the PR title/description to match.

---

def- (approved):
LGTM. Fixing the error types might still be useful if we care about that!

antiguru:
File https://linear.app/materializeinc/issue/SQL-326/parsecast-errors-return-sqlstate-xx000-instead-of-proper-data for a proper fix!

PR #36812, opened 2026-05-30
ci: Make dev docs output less busy, minor optimizations

def-:
Follow-up to https://github.com/MaterializeInc/materialize/pull/36807

Output was > 500 MB.

Closes https://linear.app/materializeinc/issue/QAR-76/deploy-devsite-build-step-duration-has-increased-noticeably


bosconi (approved):
Looks good.

PR #36814, opened 2026-05-31
adapter: assign real SQLSTATEs to all EvalError variants

antiguru:
### Motivation

`AdapterError::code` fell back to `SqlState::INTERNAL_ERROR` (`XX000`) for every
`EvalError` except a handful of special cases:

```rust
AdapterError::Eval(_) => SqlState::INTERNAL_ERROR,
```

But the overwhelming majority of evaluation errors are user-facing **data
exceptions**, not internal server errors. The catch-all mislabeled ordinary
user mistakes — `1 / 0`, `sqrt(-1)`, out-of-range casts, scalar subqueries
returning multiple rows, and many more — as `XX000`. SQLsmith surfaced one
instance (out-of-range `date` cast, SQL-326), but the wildcard was the real
bug.

Fixes SQL-326.

### Description

Replace the wildcard with an **exhaustive** `eval_error_code` match over every
`EvalError` variant, choosing the SQLSTATE PostgreSQL uses where an equivalent
exists:

| Class of error | SQLSTATE |
| --- | --- |
| division by zero | `22012` |
| out-of-range numeric / int / float | `22003` |
| out-of-range datetime / `NaN` timestamp | `22008` |
| interval overflow | `22015` |
| malformed datetime input | `22007` |
| malformed numeric/text/base64/hex input | `22P02` |
| invalid byte sequence for encoding | `22021` |
| invalid argument values (timezone, datepart, hash algo, …) | `22023` |
| invalid regular expression | `2201B` |
| invalid LIKE escape | `22025` |
| NULL where not permitted | `22004` |
| array subscript / dimension errors | `2202E` |
| string too long | `22001` |
| limits (LIKE pattern, array size, recursion, length) | `54000` |
| multi-row / negative-row scalar subquery | `21000` |
| math domain (sqrt of negative, complex result) | `2201F` |
| unsupported features | `0A000` |

Parse failures keep their kind/type-aware mapping via `parse_error_code`. Only
genuinely internal conditions (`Internal`, `InvalidCatalogJson`, the
pretty/redact printer failures) remain `INTERNAL_ERROR`.

The match is **deliberately exhaustive with no wildcard**, so adding a new
`EvalError` variant becomes a compile error until it is assigned a code — we
can't silently regress to `XX000` again.

### Scope / limitation

This covers errors raised through `AdapterError` — i.e. constant folding and
optimization (`OptimizerError::EvalError` delegates to
`AdapterError::Eval(_).code()`). Errors produced **during dataflow execution**
are stringified into `PeekResponse::Error` and surface as
`AdapterError::Unstructured` (`XX000`); giving those structured codes would
require threading the error kind through the compute boundary and is left as
follow-up. Concretely, `SELECT 1/0` reports `22012` but `SELECT 1/0 FROM
<relation>` still reports `XX000`.

### Verification

`test_parse_error_codes` (in `src/environmentd/tests/sql.rs`) covers
out-of-range date, malformed date, out-of-range float, malformed integer,
division by zero, `sqrt(-1)`, and a multi-row scalar subquery. The pgtest and
HTTP datadriven snapshots (`test/pgtest{,-mz}/*.pt`,
`src/environmentd/tests/testdata/http/{ws,post}`) are updated for the
constant-folded `division by zero` cases that now report `22012` (the
dataflow-execution cases there intentionally stay `XX000`, per the limitation
above). Also drops the three SQLsmith allowlist entries the issue named;
#36815 (stacked) prunes the rest of the now-redundant family.

Opened as a draft to let CI run the new regression test and the SQLsmith job.


aljoscha (approved):
Looks good on the surface but I didn't verify that the error codes are correct. Would defer to @def- for the testing aspect and verification of this.

def- (approved):
Testing side lgtm!

antiguru:
Thanks for the reviews!

PR #36845, opened 2026-06-01
adapter: introduce internal-only subscribes

aljoscha:
Add an `internal: bool` field on `ActiveSubscribe` so that callers can mark a subscribe as system-internal. Internal subscribes are not advertised via `mz_subscriptions` (no row write/retract on set up or tear down).

The existing user-facing call site in `sequence_subscribe` sets the flag to `false`, preserving today's behavior.

Also demote the "drop_compute_sinks called on nonexistent sink" log from `error!` to `debug!`: an internal subscribe may be torn down through its own path while session-disconnect cleanup is in flight, and that race is benign.

This is preparation for the upcoming frontend OCC read-then-write path, which uses a long-lived internal SUBSCRIBE to maintain accumulated diffs.

### DIFF
--- a/src/adapter/src/active_compute_sink.rs
+++ b/src/adapter/src/active_compute_sink.rs
@@ -115,6 +115,9 @@ pub struct ActiveSubscribe {
     pub start_time: EpochMillis,
     /// How to present the subscribe's output.
     pub output: SubscribeOutput,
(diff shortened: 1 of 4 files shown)


ggevay (approved):

aljoscha:
tyty, as always! 🙇‍♂️

PR #36848, opened 2026-06-01
expr: infer non-nullable columns from regex capture groups

antiguru:
### Motivation

Columns derived from regex capture groups were unconditionally marked nullable, with a long-standing `TODO` referencing database-issues#612 noting "we can do better." A capture group that is **not** nested inside an optional construct always participates whenever the regex matches at all, so its extracted column is non-nullable.

### Changes

- **`src/repr/src/adt/regex.rs`**: add `Regex::non_nullable_capture_groups()`, which parses the pattern AST (via `regex-syntax`, already a workspace dep) and returns the indices of capture groups guaranteed to participate in a match. A group is considered nullable only when nested inside an optional context: a `?`/`*` quantifier, a `{0,n}`/`{0}`-style repetition, or an alternation branch (`a|b`). If the pattern somehow fails to parse (shouldn't happen for a regex that compiled), it conservatively returns the empty set, preserving the previous all-nullable behavior.
- **`src/expr/src/relation/func.rs`**: `regexp_extract` now uses the analysis instead of hardcoding `nullable: true`; removes the `TODO`.
- **`src/storage-types/src/sources/encoding.rs`**: `FORMAT REGEX` source encoding uses the same analysis.

The traversal only recurses through nodes a match cannot skip over, so every capture group it reaches is non-nullable; it stops descending at optional repetitions and alternations.

**Soundness**: the runtime decode path only emits a row on a full regex match, and a structurally-required group always captures when the whole regex matches, so a non-nullable column never receives a null at runtime.

### Tests

- Unit tests in `regex.rs` covering `?`, `*`, `+`, `{0,n}`, `{1,n}`, top-level vs. nested alternation, named/non-capturing groups, plus the exact production patterns from the testdrive suite.
- New `mz_columns`-based nullability assertion in `test/sqllogictest/regex.slt` (replacing the old TODO).
- Updated `SHOW COLUMNS` expectations in both `regex-sources.td` variants (`ip`/`ts`/`path`/`code` → non-nullable; `search_kw`/`product_detail_id` remain nullable since they live in alternation branches).

Resolves the TODO for database-issues#612.

https://claude.ai/code/session_01193gm1qtyQcSHSaZNMbx6e

---
_Generated by [Claude Code](https://claude.ai/code/session_01193gm1qtyQcSHSaZNMbx6e)_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7571,6 +7571,7 @@ dependencies = [
  "prost-build",
  "rand 0.9.4",
  "regex",
+ "regex-syntax",
  "ryu",
  "serde",
  "serde_json",
--- a/misc/python/materialize/checks/all_checks/regex.py
+++ b/misc/python/materialize/checks/all_checks/regex.py
@@ -22,12 +22,10 @@ class RegexpExtractNonNullable(Check):
     """
 
     def initialize(self) -> Testdrive:
-        return Testdrive(
-            dedent("""
+        return Testdrive(dedent("""
             > CREATE TABLE regexp_extract_nn_table (f1 STRING);
             > INSERT INTO regexp_extract_nn_table VALUES ('ab');
-            """)
-        )
+            """))
 
     def manipulate(self) -> list[Testdrive]:
         return [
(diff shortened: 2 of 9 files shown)


ggevay (approved):

antiguru:
Thanks for the review!

PR #36851, opened 2026-06-01
[SSO] support dot-separated paths in oidc_group_claim

mtabebe:
Problem:
Frontegg-issued JWTs cannot place group memberships at a top-level groups claim. The only way to surface group info on the Frontegg path is to configure a JWT-generation webhook that injects a nested customClaims object `{ "customClaims": { "groups": ["analytics", "platform_eng"] } }`

The existing OIDC group sync only looks up oidc_group_claim as a single top-level key, so there is no way to reach customClaims.groups.

Solution:
Allow oidc_group_claim to be a dot-separated path into nested JSON objects. A bare claim name like groups continues to behave exactly as before. A path like customClaims.groups iterates through the intermediate object before applying the existing array/string coercion.

Testing
12 new unit tests in covering nested arrays, nested single strings, deeply nested paths, etc.

### DIFF
--- a/src/adapter-types/src/dyncfgs.rs
+++ b/src/adapter-types/src/dyncfgs.rs
@@ -142,11 +142,13 @@ pub const OIDC_GROUP_ROLE_SYNC_ENABLED: Config<bool> = Config::new(
     "Enable OIDC JWT group-to-role membership sync on login.",
 );
 
-/// The JWT claim name that contains group memberships.
+/// The JWT claim path that contains group memberships. May be a bare claim
+/// name (e.g. `groups`) or a dot-separated path into nested objects (e.g.
+/// `customClaims.groups`).
 pub const OIDC_GROUP_CLAIM: Config<&'static str> = Config::new(
     "oidc_group_claim",
     "groups",
-    "JWT claim name containing group memberships for role sync.",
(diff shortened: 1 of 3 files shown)


jubrad (approved):

SangJunBak:
Not blocking but it'd be nice to do this for `OIDC_AUTHENTICATION_CLAIM` too.

SangJunBak:
One thing I'm realizing is we still haven't hooked up the group claims logic with the Frontegg authenticator, something that's necessary if we want to have group role mapping in Materialize Cloud. The changes we've been are done in `oidc.rs`, but this serves as the authenticator kind for self managed OIDC. For Cloud, we use AuthenticatorKind::Frontegg and use the code in `src/frontegg-auth/src/auth.rs`. We'll most likely need to extract the group claim from that authenticator kind too

SangJunBak (approved):

mtabebe:
> Not blocking but it'd be nice to do this for `OIDC_AUTHENTICATION_CLAIM` too.

Sure it can be a separate PR

jubrad:
https://github.com/MaterializeInc/materialize/pull/36851#issuecomment-4596094121
^well this explains why it wasn't working  for me :)

PR #36853, opened 2026-06-01
Fix panic when attempting to set default_timestamp_interval to 0

peterdukelarsen:
### Motivation

Closes [ss-147](https://linear.app/materializeinc/issue/SS-147/thread-coordinator-panicked-at-srcadaptersrccoordddlrs56155-period)

### Description

Adds a non-zero constraint to the "default_timestamp_interval" system parameter. This prevents an issue where setting it to 0 causes a panic and the database won't start up until you go in and update the parameter by different means.

### Verification

1. Reproduced the issue locally, updated code and confirmed we received an error instead of crashing when the `ALTER ...` statement is issued
2. Added basic sqllogictest tests

### DIFF
--- a/src/sql/src/session/vars/constraints.rs
+++ b/src/sql/src/session/vars/constraints.rs
@@ -11,6 +11,7 @@
 
 use std::fmt::Debug;
 use std::ops::{RangeBounds, RangeFrom, RangeInclusive};
+use std::time::Duration;
 
 use mz_repr::adt::numeric::Numeric;
 use mz_repr::bytes::ByteSize;
@@ -18,6 +19,8 @@ use mz_repr::bytes::ByteSize;
 use super::{Value, Var, VarError};
 
 pub static NUMERIC_NON_NEGATIVE: NumericNonNegNonNan = NumericNonNegNonNan;
+
+pub static NON_ZERO_DURATION: NonZeroDuration = NonZeroDuration;
 
 pub static NUMERIC_BOUNDED_0_1_INCLUSIVE: NumericInRange<RangeInclusive<f64>> =
     NumericInRange(0.0f64..=1.0);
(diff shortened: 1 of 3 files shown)


src/sql/src/session/vars/constraints.rs
```diff
@@ -114,6 +117,25 @@ impl DomainConstraint for NumericNonNegNonNan {
...
+
+impl DomainConstraint for NonZeroDuration {
+    type Value = Duration;
+
+    fn check(&self, var: &dyn Var, d: &Duration) -> Result<(), VarError> {
+        if d.is_zero() {
+            Err(VarError::InvalidParameterValue {
+                name: var.name(),
+                invalid_values: vec![format!("{:?}", d)],
+                reason: "only supports durations greater than zero".to_string(),
```
ublubu: It looks a little weird to `is_zero()` and then say `greater than zero`. I had to double-check that Duration can't be negative 😅

test/sqllogictest/alter.slt
```diff
@@ -205,6 +205,21 @@ ALTER SYSTEM SET max_timestamp_interval = '10s';
...
+
+simple conn=mz_system,user=mz_system
+ALTER SYSTEM SET default_timestamp_interval = '0';
+----
+db error: ERROR: parameter "default_timestamp_interval" cannot have value "0ns": only supports durations greater than zero
+
+simple conn=mz_system,user=mz_system
+ALTER SYSTEM SET default_timestamp_interval = '-1';
+----
+db error: ERROR: parameter "default_timestamp_interval" requires a "duration" value
```
ublubu, line 221: I realize this isn't a regression, but this error message is interesting, especially given that the above tests use values like `1` and `0`, which don't appear to be durations either.

ublubu (approved):
Looks good. It does give me a new question, though. (inline)

antiguru (approved):

src/sql/src/session/vars/constraints.rs
peterdukelarsen: Ah whoops, I'll change this to "only supports non-zero durations" - I had originally named this struct `PositiveDuration` which I later realized was a little weird/redundant given Durations are never negative. But then I forgot to update this reason field.

test/sqllogictest/alter.slt
peterdukelarsen: Poked around here a bit, and it looks like default unit for a `Duration` in our parsing logic is "ms", see here: https://github.com/MaterializeInc/materialize/blob/8f3bf9dd6298b6e82e01af3c615e3a976848c2c7/src/sql/src/session/vars/value.rs#L309-L310.

This seems to be consistent across durations, but is documented on a per-field level. I.e. see docs for "idle_in_transaction_session_timeout" under https://materialize.com/docs/sql/alter-system-set/.

Discussed offline that this may also align with postgres.

PR #36856, opened 2026-06-01
SS-16 AWS Glue Support: SQL

martykulma:
Introduces SQL `CREATE CONNECTION` for Glue Schema Registry, which takes an AWS CONNECTION and a registry name.

- Adds a feature flag, which is currently false because things are incomplete.
- SLT tests for new syntax


patrickwwbutler (approved):
LGTM! pretty standard sql changes here it seems

PR #36857, opened 2026-06-02
sql-pretty: add indent & config in wasm

justjake:
### Motivation

I want to shape Materialize SQL for rendering in the terminal as part of my migration tooling.

For this use-case, 4-space indent is quite greedy for horizontal screen space, so I would prefer a 2-space indent. However everyone's opinion is different, so it should be an option.

### Description

Note: I am not a Rust programmer, so this diff is mostly Codex. I reviewed the code and it looks good to me, but I am not attuned to the aesthetics of Rust. Thank you for your patience.

Add an `indent` field to PrettyConfig and use it when nesting pretty-printer documents. Provide Default for PrettyConfig so existing callers only specify the fields they override.

I converted several stand-alone functions into methods to make indent available to them.

Then in wasm, add `prettyStrConfig` and `prettyStrsConfig` exports that accept width,
 indent, and formatMode and turns it into `mz_sql_pretty::PrettyConfig`. 

I considered tacking the indent on as a third arg, but I think `f(string, number, number)` is a bad signature - too easy to confuse the number args. I considered overloading the current signature but overloads are a bit weird and I didn't want to suggest weird code. Happy to do if y'all like them (and bindgen supports...).

I added a little JavaScript test to verify the whole package works.

### Verification

- Rust unit tests.
- New JavaScript test.

### DIFF
--- a/misc/wasm/src/sql-pretty-wasm/src/lib.rs
+++ b/misc/wasm/src/sql-pretty-wasm/src/lib.rs
@@ -23,6 +23,92 @@ use wasm_bindgen::prelude::*;
 static ALLOCATOR: LockedAllocator<FreeListAllocator> =
     LockedAllocator::new(FreeListAllocator::new());
 
+#[wasm_bindgen(typescript_custom_section)]
+const TS_TYPES: &str = r#"
+export type PrettyFormatMode = "simple" | "simpleRedacted" | "stable";
+
+export interface PrettyConfig {
+    width?: number;
+    indent?: number;
+    formatMode?: PrettyFormatMode;
+}
+"#;
+
+#[wasm_bindgen]
+extern "C" {
+    #[wasm_bindgen(typescript_type = "PrettyConfig")]
+    pub type JsPrettyConfig;
+
+    #[wasm_bindgen(method, getter, structural)]
+    fn width(this: &JsPrettyConfig) -> JsValue;
+
+    #[wasm_bindgen(method, getter, structural)]
+    fn indent(this: &JsPrettyConfig) -> JsValue;
+
+    #[wasm_bindgen(method, getter, structural, js_name = formatMode)]
+    fn format_mode(this: &JsPrettyConfig) -> JsValue;
+}
+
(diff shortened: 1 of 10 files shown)


misc/wasm/src/sql-pretty-wasm/src/lib.rs
```diff
@@ -23,6 +23,92 @@ use wasm_bindgen::prelude::*;
...
 
+#[wasm_bindgen(typescript_custom_section)]
+const TS_TYPES: &str = r#"
+export type PrettyFormatMode = "simple" | "simpleRedacted" | "stable";
+
+export interface PrettyConfig {
+    width?: number;
+    indent?: number;
+    formatMode?: PrettyFormatMode;
+}
```
justjake, line 34: Is this legit? The types aren't auto-generated?

justjake:

antiguru:
Reviewed. Mostly a clean, behavior-preserving refactor (default `indent` = old `TAB` = 4, no `TAB` left, all `PrettyConfig` constructors updated). No correctness bugs found. Two low-severity notes:

**`misc/wasm/src/sql-pretty-wasm/src/lib.rs` (the `width`/`indent` helpers)** — the four raw casts (`usize::MAX as f64`, `width as usize`, `isize::MAX as f64`, `indent as isize`) trip `as_conversions = "warn"`, which this workspace opts into (`misc/wasm/Cargo.toml`). It isn't CI-blocking (root `cargo clippy` doesn't cover the `misc/wasm` workspace) and `mz_ore`'s cast helpers aren't reachable here (`mz-ore` is only a non-wasm dev-dep of `mz-sql-pretty`, and `CastLossy` lacks an `f64→isize` impl anyway). Suggest an explicit `#[allow(clippy::as_conversions)]` with a one-line justification rather than pulling in `mz-ore`.

**`bin/wasm-build`** — dropping `exec` and appending the `node` test run changes the script's contract: `ci/deploy/npm.py` calls `bin/wasm-build` to build the published package, so the npm deploy now also runs `test.mjs` and fails if `node` lacks `--experimental-wasm-modules` or the test fails. Likely intended, just flagging the coupling.

PR #36858, opened 2026-06-02
ssh-util: drop FIPS path from openssh-static, keep non-FIPS build

jasonhernandez:
## Summary

Carves the FIPS work out of #35858 and back into a separate draft (#36859), which is gated on AWS-LC-FIPS 3.x completing NIST CMVP validation. The static OpenSSH image **stays on main** because it is the prerequisite for the distroless migration of environmentd/clusterd, independent of FIPS.

This addresses the post-merge QA review on #35858 (https://github.com/MaterializeInc/materialize/pull/35858#issuecomment-4582060423), which found the FIPS path could not actually produce a validated module yet.

### Changes

- **Dockerfile**: remove the `AWS_LC_FIPS` build arg and `-DFIPS=1` cmake path; the image now builds only a non-FIPS static `ssh` against a regular AWS-LC release. Bump OpenSSH to `V_10_3_P1`, which natively stubs `BN_set_flags()` for AWS-LC, so the prior `-DBN_FLG_CONSTTIME=0` shim is dropped (it was a footgun: a global value-define of a security-critical OpenSSL macro to `0`).
- **Keep the configure-CPPFLAGS hardening fix** from the QA review (Finding 3): inject `-DBN_FLG_CONSTTIME=0` at configure time so OpenSSH's hardening/optimization flags (`-O2`, `-fstack-protector-strong`, `-fPIE`, `-ftrapv`, ...) are preserved. A `make CFLAGS=...` override silently replaced them, producing an unhardened binary.
- **tunnel.rs**: remove the dormant `MZ_FIPS` SSH-config enforcement (`fips_mode_enabled` / `write_fips_ssh_config` and the call site).

Opened as a **draft** at the author's request.

Part of SEC-236.

## Test plan

- [x] `cargo check -p mz-ssh-util` passes (rustc 1.96.0)
- [x] `cargo fmt -p mz-ssh-util --check` clean
- [x] Docker build of `misc/images/openssh-static/` produces a working static binary (verified: statically-linked `OpenSSH_10.3p1` against AWS-LC 1.54.0, 3.9M)

### DIFF
--- a/misc/images/openssh-static/Dockerfile
+++ b/misc/images/openssh-static/Dockerfile
@@ -9,13 +9,15 @@
 
 # Build a statically-linked OpenSSH `ssh` client binary using AWS-LC
 # as the crypto backend. AWS-LC is a faster, smaller alternative to
-# OpenSSL that also supports FIPS 140-3 validation when needed.
+# OpenSSL. The static binary (~3-6MB) is the prerequisite for migrating
+# environmentd/clusterd to distroless containers, which lack /usr/bin/ssh.
 #
 # OpenSSH natively supports AWS-LC as a crypto backend (no patches needed).
 # See: https://github.com/openssh/openssh-portable/blob/master/INSTALL
 #
-# To enable FIPS mode, build with: --build-arg AWS_LC_FIPS=1
-# (requires Go for the FIPS delocator)
+# FIPS support is built separately; see the SEC-236 FIPS draft (it requires a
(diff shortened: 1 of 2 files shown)


jasonhernandez:
@def- per your feedback / review. I'm removing the incomplete scaffolding for FIPS support and just focusing on getting rid of the openssh dependency to unblock more use of distroless for now.

def- (approved):

PR #36862, opened 2026-06-02
sql: return NULL from pg_cancel_backend(NULL) instead of panicking

aljoscha:
`SELECT pg_cancel_backend(null)` panicked in the planner with `Datum::unwrap_int32 called on Null` because the side-effecting-function plan_fn unconditionally unwrapped the argument as an int32.

PostgreSQL treats pg_cancel_backend as strict (NULL input -> NULL
output), so match that: PgCancelBackend now carries an Option<u32>
connection id, the planned return type is nullable, and a None id short-circuits to a NULL result.

Note on duplication: the NULL short-circuit and its explanatory comment are duplicated across sequence_side_effecting_func (old peek sequencing) and execute_side_effecting_func (frontend peek path). This mirrors the existing duplication of the cancel logic between those two functions -- the old peek-sequencing path is slated for removal (see the existing TODO(peek-seq)), at which point the duplicate goes away with it. Rather than introduce a shared helper for code that is about to be deleted, the behavior is kept duplicated alongside the code it lives next to.

Fixes SQL-295.

### DIFF
--- a/src/environmentd/tests/sql.rs
+++ b/src/environmentd/tests/sql.rs
@@ -3098,6 +3098,40 @@ fn test_pg_cancel_backend() {
     );
 }
 
+/// Regression test for SQL-295: `pg_cancel_backend(NULL)` used to panic while
+/// planning the call (`Datum::unwrap_int32 called on Null`). Like in
+/// PostgreSQL, a `NULL` argument must return `NULL` instead.
+#[mz_ore::test]
+fn test_pg_cancel_backend_null() {
+    mz_ore::test::init_logging();
+    let server = test_util::TestHarness::default().start_blocking();
+
+    let mut client = server.connect(postgres::NoTls).unwrap();
+
+    // A `NULL` argument returns `NULL`, matching PostgreSQL, rather than
+    // panicking. Cover the literal `NULL`, an explicitly-typed `NULL`, a `NULL`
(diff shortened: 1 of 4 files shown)


src/environmentd/tests/sql.rs
```diff
@@ -3098,6 +3098,40 @@ fn test_pg_cancel_backend() {
     );
 }
 
+/// Regression test for SQL-295: `pg_cancel_backend(NULL)` used to panic while
+/// planning the call (`Datum::unwrap_int32 called on Null`). Like in
+/// PostgreSQL, a `NULL` argument must return `NULL` instead.
+#[mz_ore::test]
+fn test_pg_cancel_backend_null() {
```
ggevay, line 3105: Could be merged into the existing `test_pg_cancel_backend`, to avoid the overhead of starting up one more Materialize.

ggevay (approved):
LGTM, just a minor comment

src/environmentd/tests/sql.rs
aljoscha: hehe, I had it folded into that one first, but then thought keeping it separate is better to keep tests smaller and more focused :sweat_smile:

ggevay: Oh, ok. I'm fine with either option :)

aljoscha:
tyty! :bowing_man:

PR #36863, opened 2026-06-02
CLO-81 cancel rollout after timeout

alex-hunt-materialize:
Cancel rollout after timeout exceeded.

### Motivation

Resolves https://linear.app/materializeinc/issue/CLO-81/deploy-should-cancel-environment-rollout-after-a-configurable-timeout

### Description

Adds a new `rollout_request_timeout` field to the Materialize CRD, with a default of 24 hours. If the rollout runs for longer than that time, it gets aborted.

There is a significant change to the timestamp of the UpToDate condition. We now track its last transition time as when the value changes, not every time the reason changes. I don't think this affects anything else, but I think this is actually closer to how most K8S things use this field, and it's required to not reset the timer.

### Verification

New rust tests, and new orchestratord nightly test.


alex-hunt-materialize:
@bosconi Did you get the distribution of actual times? I wrote this with 24h timeout, but I'd like to confirm that's the appropriate number before we merge.

doy-materialize (approved):

PR #36864, opened 2026-06-02
Fix `ConstantVisitor` to recurse into `TableFactor`s

ggevay:
Fixes https://linear.app/materializeinc/issue/SQL-298/non-constant-insert-select-are-classified-as-constant

### DIFF
--- a/test/testdrive/insert-select.td
+++ b/test/testdrive/insert-select.td
@@ -47,6 +47,49 @@ contains:column "i" is of type integer but expression is of type text
 ! INSERT INTO t SELECT * FROM t;
 contains:cannot be run inside a transaction block
 
+# Regression tests for ConstantVisitor (SQL-298)
+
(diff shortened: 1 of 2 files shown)


test/testdrive/insert-select.td
```diff
@@ -47,6 +47,50 @@ contains:column "i" is of type integer but expression is of type text
 ! INSERT INTO t SELECT * FROM t;
 contains:cannot be run inside a transaction block
 
+# Regression tests for ConstantVisitor
+# https://linear.app/materializeinc/issue/SQL-298/non-constant-insert-select-are-classified-as-constant
```
aljoscha: maybe shorten to just "SQL-298"

aljoscha (approved):

test/testdrive/insert-select.td
ggevay: Done

PR #36865, opened 2026-06-02
Error gracefully if a `SUBSCRIBE` key has duplicate columns

ggevay:
Fixes https://linear.app/materializeinc/issue/SQL-307/subscribe-envelope-upsert-key-a-a-crashes-environmentd

### DIFF
--- a/src/sql/src/plan/statement/dml.rs
+++ b/src/sql/src/plan/statement/dml.rs
@@ -17,7 +17,7 @@ use std::collections::{BTreeMap, BTreeSet};
 
 use itertools::Itertools;
 use mz_arrow_util::builder::ArrowBuilder;
-use mz_expr::RowSetFinishing;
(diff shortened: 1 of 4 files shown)


src/sql/src/plan/statement/dml.rs
```diff
@@ -1748,6 +1756,14 @@ pub fn plan_subscribe(
             if !map_exprs.is_empty() {
                 return Err(PlanError::InvalidKeysInSubscribeEnvelopeDebezium);
             }
+            let mut seen = BTreeSet::new();
```
aljoscha: light code deduplication

aljoscha (approved):
looks good, flagged one bit of code duplication

src/sql/src/plan/statement/dml.rs
ggevay: Factored it out into a function.

PR #36866, opened 2026-06-02
sql-parser: fix `AstDisplay` panic for `CREATE TABLE FROM SOURCE` with constraints only

ggevay:
CreateTableFromSourceStatement's AstDisplay entered the parenthesized block whenever columns were specified OR constraints were non-empty, but the inner match panicked on `TableFromSourceColumns::NotSpecified`. A statement like
```
    CREATE TABLE t (PRIMARY KEY (c)) FROM SOURCE foo (REFERENCE bar)
```
would crash as soon as the AST was formatted (e.g. via statement logging). Treat `NotSpecified` as a no-op and gate the ", " separator so constraints-only output stays valid SQL. Adds round-trip tests for the constraints-only case and for columns + constraints (both Defined and Named).

### DIFF
--- a/src/sql-parser/src/ast/defs/statement.rs
+++ b/src/sql-parser/src/ast/defs/statement.rs
@@ -1760,7 +1760,7 @@ impl<T: AstInfo> AstDisplay for CreateTableFromSourceStatement<T> {
             f.write_str(" (");
 
             match columns {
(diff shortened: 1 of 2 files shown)


martykulma (approved):
thanks @ggevay - lgtm!

PR #36867, opened 2026-06-02
mz-test: note `CI=1` trap, `--optimized`

ggevay:
Make the testing skill mention some gotchas in how to run testdrive:
- use `--optimized` in the canonical command;
- unset `CI=1`, which was auto-set in my Junie env (RustRover's AI agent), creating some confusion;
- argument ordering.

### DIFF
--- a/.agents/skills/mz-test/SKILL.md
+++ b/.agents/skills/mz-test/SKILL.md
@@ -49,10 +49,18 @@ You can use `-v` for printing each query before it is run, e.g. to figure out wh
 Run testdrive files with:
 
 ```
-bin/mzcompose --find testdrive run default -- FILENAME.td
+bin/mzcompose --optimized --find testdrive run default -- FILENAME.td
 ```
 
 `FILENAME.td` is a file in `test/testdrive/`, relative to that directory (not the repo root).
+Global `mzcompose` flags like `--optimized`/`--dev` must go before the `run`
+subcommand (they are top-level flags, not `run` flags).
+
+If `CI=1` is set in your shell (agent runtimes such as Junie export
+it by default for unattended-mode hygiene), `mzcompose` will refuse to build
+images locally and exit with `Expected builds to be available, the build
+probably failed, so not proceeding`. Either `unset CI` or set
+`CI_ALLOW_LOCAL_BUILD=1` before invoking `bin/mzcompose`.
 
 Some compositions (e.g. platform-checks, upgrade, pg-cdc multi-version) run the
 same `.td` file against multiple Materialize versions. When a change alters


.agents/skills/mz-test/SKILL.md
```diff
@@ -49,10 +49,18 @@ You can use `-v` for printing each query before it is run, e.g. to figure out wh
...
-bin/mzcompose --find testdrive run default -- FILENAME.td
+bin/mzcompose --optimized --find testdrive run default -- FILENAME.td
 ```
 
 `FILENAME.td` is a file in `test/testdrive/`, relative to that directory (not the repo root).
+Global `mzcompose` flags like `--optimized`/`--dev` must go before the `run`
+subcommand (they are top-level flags, not `run` flags).
+
+If `CI=1` is set in your shell (agent runtimes such as Junie export
+it by default for unattended-mode hygiene), `mzcompose` will refuse to build
```
def-, line 60: Huh! Why does Junie set `CI=1` if it's not running in CI?

.agents/skills/mz-test/SKILL.md
```diff
@@ -49,10 +49,18 @@ You can use `-v` for printing each query before it is run, e.g. to figure out wh
 Run testdrive files with:
 
 ```
-bin/mzcompose --find testdrive run default -- FILENAME.td
+bin/mzcompose --optimized --find testdrive run default -- FILENAME.td
```
def-, line 52: This is the default already, why do we need to specify it explicitly?

ggevay: Oh, ok, sorry, then no need! (I thought something else is the default, because it started building a lot of things for me yesterday, but maybe I just had my build dir empty.)

Btw. we also have this stale help text, which suggests that `--release` is the default:
```
        build_mode.add_argument(
            "--release",
            action="store_true",
            help="build Rust binaries with the release profile (default)",
        )
```

.agents/skills/mz-test/SKILL.md
ggevay: In Junie's own words:

> Junie's runtime injects CI=1, TERM=dumb, GIT_TERMINAL_PROMPT=0, DEBIAN_FRONTEND=noninteractive into every shell tool invocation as a generic "you're running unattended, pick CI-friendly defaults" hint.

I think it wants to nudge away various tools from getting into interactive mode, and it seems that [some tools](https://claude.ai/share/d6bb127e-86ca-4f30-bc68-01b976717dc6) understand `CI=1` as a hint to not go interactive.

So, it would be good to resolve this somehow. Either with adding the above mention to the skill, or maybe changing our code to use not `CI`, but e.g. `BUILDKITE`. Btw. `BUILDKITE` is set automatically by Buildkite:
https://buildkite.com/docs/pipelines/configure/environment-variables

.agents/skills/mz-test/SKILL.md
def-: Yep, we should update that! The actual default is a bit complex. Maybe Junie ran into doing a release build because of CI=1 actually...

.agents/skills/mz-test/SKILL.md
def-: I'd consider it ok to use BUILDKITE instead of CI in all of our tools instead. But who knows if Junie will next discover setting that to BUILDKITE=1 :D

.agents/skills/mz-test/SKILL.md
ggevay: I think that's `CI_LTO` which makes the default `--release`:
```
Profile.RELEASE if ui.env_is_truthy("CI_LTO") else Profile.OPTIMIZED
```

.agents/skills/mz-test/SKILL.md
ggevay: Fair point! I'll just try to add a mention of this in some local, Junie-specific `md` for now.

PR #36868, opened 2026-06-02
pgwire: error instead of panicking on binary encoding of unsupported types

aljoscha:
COPY ... TO STDOUT WITH (FORMAT binary) over a query whose result contained a list, map, or aclitem column panicked the connection's worker thread:

    encode_binary should never trigger a to_sql failure:
    "binary encoding of list types is not implemented"

These types have no binary encoding, so `Value::encode_binary` returns an error, which the surrounding `.expect()` turned into a panic. The invariant that `.expect()` relies on is that callers pre-validate types with `can_encode_binary`, but two gaps let unsupported types through:

  * COPY never validated its format at all. The extended-query Bind handler does validate result column formats, but it deliberately skips COPY TO statements -- a COPY portal's result formats describe the CopyData wrapper, not the COPY format chosen via WITH (FORMAT ...) -- and nothing validated the COPY format itself.

  * The Bind handler's check only inspected the top-level scalar type, so a binary result column whose type was a record/array/range *containing* a list, map, or aclitem slipped through and hit the same panic via the ordinary extended-query path (e.g. SELECT ROW(1, LIST[1, 2]) requested with a binary result format, as sqllogictest does).

Fix both:

  * Replace `can_encode_binary` with `binary_encoding_error`, which recurses into container types and returns the offending leaf type's error message. `can_encode_binary` is kept as a thin wrapper. This is now the single source of truth, shared by the COPY path, the Bind handler, and the Kafka BYTES sink.

  * Add a pre-stream check in `copy_rows` that rejects un-encodable types with a clean error before sending CopyOutResponse, mirroring the existing Parquet handling.

With every caller validating recursively, the `.expect()` invariant now genuinely holds for all entry points.

Tests:

  * test/pgtest-mz/copy-binary-unsupported.pt covers COPY binary of list, map, aclitem, and a record containing a list, plus the text-format success case. It lives in pgtest-mz because list/map do not exist in PostgreSQL and our aclitem message differs from PostgreSQL's.

  * test/sqllogictest/record.slt covers the recursive Bind-path cases that previously panicked the server.

Fixes SQL-323

### DIFF
--- a/src/pgrepr/src/value.rs
+++ b/src/pgrepr/src/value.rs
@@ -969,8 +969,8 @@ mod tests {
     #[mz_ore::test]
     #[cfg_attr(miri, ignore)] // numeric/decimal contexts unsupported under miri
     fn proptest_binary_encoding_error_matches_encode_binary() {
-        let strat = any::<SqlScalarType>()
-            .prop_flat_map(|ty| (Just(ty.clone()), arb_datum_for_scalar(ty)));
+        let strat =
+            any::<SqlScalarType>().prop_flat_map(|ty| (Just(ty.clone()), arb_datum_for_scalar(ty)));
         proptest!(ProptestConfig::with_cases(256), |((ty, prop_datum) in strat)| {
             // `binary_encoding_error` is a precondition for callers of
             // `encode_binary`: if it returns `Ok`, then encoding must succeed
--- a/src/pgrepr/src/value.rs
+++ b/src/pgrepr/src/value.rs
@@ -434,7 +434,7 @@ impl Value {
     /// format](Format::Binary).
     pub fn encode_binary(&self, ty: &Type, buf: &mut BytesMut) -> Result<(), io::Error> {
         // NOTE: If implementing binary encoding for a previously unsupported `Value` type,
-        // please update the `can_encode_binary` method below.
+        // please update the `binary_encoding_error` method below.
         let is_null = match self {
             Value::Array { dims, elements } => {
                 let ndims = pg_len("number of array dimensions", dims.len())?;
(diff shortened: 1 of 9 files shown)


src/pgwire/src/protocol.rs
```diff
@@ -1451,32 +1451,15 @@ where
         }) {
             if let Some(desc) = stmt.desc().relation_desc.clone() {
                 for (format, ty) in result_formats.iter().zip_eq(desc.iter_types()) {
-                    match (format, &ty.scalar_type) {
-                        (Format::Binary, mz_repr::SqlScalarType::List { .. }) => {
+                    if let Format::Binary = format {
+                        if let Err(msg) = mz_pgrepr::Value::binary_encoding_error(&ty.scalar_type) {
                             return self
                                 .send_error_and_get_state(ErrorResponse::error(
                                     SqlState::PROTOCOL_VIOLATION,
```
ggevay, line 1458: A pre-existing issue, which we could maybe also fix here: We have been using `SqlState::PROTOCOL_VIOLATION`, while Postgres uses `SqlState::UNDEFINED_FUNCTION` for this situation where it doesn't have a binary encoding for a type.

src/pgrepr/src/value.rs
```diff
@@ -586,51 +586,63 @@ impl Value {
     }
 
     /// Static helper method to pre-validate that a given Datum corresponding to
```
ggevay: nit: There is no "given" Datum here; the function takes only a type. So, I'd suggest to just drop the word "given".

src/pgrepr/src/value.rs
```diff
@@ -586,51 +586,63 @@ impl Value {
...
+            SqlScalarType::RegProc => Ok(()),
+            SqlScalarType::RegType => Ok(()),
+            SqlScalarType::UInt32 => Ok(()),
+            SqlScalarType::UInt64 => Ok(()),
+            SqlScalarType::Float32 => Ok(()),
+            SqlScalarType::Float64 => Ok(()),
+            SqlScalarType::Numeric { .. } => Ok(()),
+            SqlScalarType::MzTimestamp => Ok(()),
+            SqlScalarType::MzAclItem => Ok(()),
+            SqlScalarType::AclItem => Err("binary encoding of aclitem types does not exist"),
```
ggevay: nit:

> aclitem types

I'd use singular, as there is only one aclitem type.

src/pgrepr/src/value.rs
```diff
@@ -942,8 +954,44 @@ pub fn values_from_row(row: &RowRef, typ: &SqlRelationType) -> Vec<Option<Value>
...
+    /// Property test: [`Value::binary_encoding_error`] agrees with the actual
+    /// behavior of [`Value::encode_binary`] for every `(SqlScalarType, Datum)`
+    /// pair the proptest infrastructure can generate.
+    ///
+    /// This guards against future drift between the static predicate and the
+    /// encoder: any new `SqlScalarType` variant whose classification disagrees
+    /// with what `encode_binary` actually does will surface here.
+    #[mz_ore::test]
+    #[cfg_attr(miri, ignore)] // numeric/decimal contexts unsupported under miri
+    fn proptest_binary_encoding_error_matches_encode_binary() {
```
ggevay, line 971: (I pushed up this proptest.)

ggevay (approved):
LGTM, just minor comments.

Plus I pushed up a proptest.

src/pgwire/src/protocol.rs
aljoscha: 😂 you will like this follow up commit that I already have sitting there: https://github.com/aljoscha/materialize/tree/sql-323-binary-encoding-pg-diagnostic

src/pgrepr/src/value.rs
aljoscha: tyty! 🙇‍♂️

PR #36871, opened 2026-06-02
Document limitations of self-managed SSO

val-materialize:
Added current limitations for self-managed SSO regarding supported protocols.

### DIFF
--- a/doc/user/content/security/self-managed/sso.md
+++ b/doc/user/content/security/self-managed/sso.md
@@ -24,7 +24,11 @@ separately using [role-based access control (RBAC)](/security/self-managed/acces
 {{</ note >}}
 
 {{< note >}}
-**Current limitations:** Self-managed SSO supports OIDC only. SAML, SCIM (automated user provisioning), and IdP group-to-role mapping are not currently supported. 
+**Current limitations:**
+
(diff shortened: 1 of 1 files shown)


doc/user/content/security/self-managed/sso.md
```diff
@@ -23,6 +23,10 @@ SSO handles **authentication** only. Permissions within the database are managed
 separately using [role-based access control (RBAC)](/security/self-managed/access-control/).
 {{</ note >}}
 
+{{< note >}}
+**Current limitations:** Self-managed SSO supports OIDC only. SAML, SCIM (automated user provisioning), and IdP group-to-role mapping are not currently supported.
```
kay-kim: So ... the first sentence makes it seem like lack of SCIM and IdP group-to-role mapping is an SSO limitation.  
Also, the `(automated user provisioning)` is slightly confusing because we state below that we support automated user provisioning.

val-materialize: Good catches, both fair. Reframed as three separate bullets to untangle auth / provisioning / authorization, and dropped the "(automated user provisioning)"

kay-kim (approved):
lgtm!

PR #36872, opened 2026-06-02
clusterd: derive CLUSTERD_PROCESS from pod hostname for distroless

jasonhernandez:
## Summary

Distroless clusterd has no shell entrypoint to set `CLUSTERD_PROCESS` (the process ordinal), which `entrypoint.sh` derived from the pod hostname before the distroless migration (#36099) removes it. clusterd now derives it itself when running under Kubernetes and the value is unset: the ordinal is the last `-`-delimited segment of the StatefulSet pod name, matching how orchestrator-kubernetes parses the process id back out of pod names.

No signal handling is added. tini (added to the distroless image in #36099) runs as PID 1 and forwards signals, so clusterd is a child process and SIGTERM takes its default action rather than being ignored by PID 1.

Part of the distroless migration: #36099 (image), #36101 (UID/GID gating), #36876 (removes the unused CTP FQDN check, stacks on this). Supersedes the closed #36100.

## Merge order

Land before or with #36099. Without it, distroless clusterd has no process ordinal and won't start. Safe to land first: on the current Ubuntu images `entrypoint.sh` still sets `CLUSTERD_PROCESS`, and the in-process derivation only runs when it is unset.

## Test plan

- [x] `cargo check -p mz-clusterd`
- [ ] Confirm a distroless clusterd pod in k8s picks up its ordinal and joins the cluster

### DIFF
--- a/src/clusterd/src/lib.rs
+++ b/src/clusterd/src/lib.rs
@@ -174,8 +174,39 @@ struct Args {
     enable_storage_introspection_logs: bool,
 }
 
+/// The process ordinal for a StatefulSet pod, taken from the trailing
+/// `-`-delimited segment of its hostname (e.g.
+/// "mz5ncn-cluster-s1-replica-s1-gen-1-0" → "0"). This mirrors how
+/// orchestrator-kubernetes recovers the process id from pod names.
+///
+/// Returns `None` when the trailing segment is not a non-negative integer, so
+/// an unexpected hostname leaves `CLUSTERD_PROCESS` unset rather than set to a
+/// value that fails to parse as the process index.
+fn process_ordinal_from_hostname(hostname: &str) -> Option<&str> {
+    let ordinal = hostname.rsplit('-').next()?;
+    ordinal.parse::<usize>().ok().map(|_| ordinal)
+}
+
 pub fn main() {
     mz_ore::panic::install_enhanced_handler();
+
+    // Derive `CLUSTERD_PROCESS` (the process ordinal) from the pod hostname
+    // when running under Kubernetes and it was not set explicitly. The
+    // distroless image has no shell entrypoint to do this, so clusterd does it
+    // itself.
+    if std::env::var("KUBERNETES_SERVICE_HOST").is_ok()
+        && std::env::var("CLUSTERD_PROCESS").is_err()
+    {
+        if let Ok(hostname) = std::env::var("HOSTNAME") {
+            if let Some(ordinal) = process_ordinal_from_hostname(&hostname) {
+                // SAFETY: `set_var` is called before any threads are spawned.
+                // `install_enhanced_handler` above only registers a panic hook.
+                // That hook spawns a thread only on panic, which cannot happen
+                // before this call.
+                unsafe { std::env::set_var("CLUSTERD_PROCESS", ordinal) };
+            }
+        }
+    }
 
     let args = cli::parse_args(CliConfig {
         env_prefix: Some("CLUSTERD_"),
@@ -476,4 +507,32 @@ fn is_connection_error(e: &std::io::Error) -> bool {
             | std::io::ErrorKind::ConnectionAborted
             | std::io::ErrorKind::ConnectionReset
     )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[mz_ore::test]
+    fn test_process_ordinal_from_hostname() {
+        // A StatefulSet pod name ends in the process ordinal.
+        assert_eq!(
+            process_ordinal_from_hostname("mz5ncn-cluster-s1-replica-s1-gen-1-0"),
+            Some("0")
+        );
+        assert_eq!(
+            process_ordinal_from_hostname("mz5ncn-cluster-s1-replica-s1-gen-1-11"),
+            Some("11")
+        );
+        // A bare numeric hostname is its own ordinal.
+        assert_eq!(process_ordinal_from_hostname("7"), Some("7"));
+
+        // A trailing segment that is not a non-negative integer yields `None`,
+        // so `CLUSTERD_PROCESS` stays unset rather than being set to a value
+        // that fails to parse as the process index.
+        assert_eq!(process_ordinal_from_hostname("clusterd"), None);
+        assert_eq!(process_ordinal_from_hostname("replica-abc"), None);
+        assert_eq!(process_ordinal_from_hostname("replica-"), None);
+        assert_eq!(process_ordinal_from_hostname(""), None);
+    }
 }


src/orchestrator-kubernetes/src/lib.rs
```diff
@@ -1001,15 +1001,42 @@ impl NamespacedOrchestrator for NamespacedKubernetesOrchestrator {
...
+                    field_ref: Some(ObjectFieldSelector {
+                        field_path: "metadata.name".to_string(),
+                        ..Default::default()
+                    }),
+                    ..Default::default()
+                }),
+                ..Default::default()
+            },
+            EnvVar {
+                name: "CLUSTERD_GRPC_HOST".to_string(),
```
antiguru: What does the CLUSTERD_GRPC_HOST env symbol actually do? We only use GRPC for persist pubsub now, and I'm not sure it uses this.

jasonhernandez: Thanks - digging into it - I don't think it is used. I reworked this draft PR to just add the SIGTERM handler (which should unblock switching clusterd to distroless). #36876 removes some more cruft.

src/clusterd/src/lib.rs
```diff
@@ -174,9 +174,69 @@ struct Args {
...
 }
 
+/// On Linux, PID 1 has special signal semantics: the kernel will not
+/// deliver signals whose disposition is SIG_DFL (the default). Since
+/// distroless containers run the binary directly as PID 1 (no tini),
+/// signals like SIGTERM from Kubernetes pod termination would be silently
+/// ignored without explicit handlers. This function registers a handler
+/// that restores the default disposition and re-raises, producing the
+/// expected termination behavior.
+fn install_termination_signal_handlers() {
```
antiguru: If we need this elsewhere consider moving to mz-ore.

antiguru (approved):
Looks fine. Do we need the SIGTERM handler in other binaries, too? Then I'd recommend extracting common logic into mz-ore.

src/clusterd/src/lib.rs
```diff
@@ -174,9 +174,69 @@ struct Args {
...
+    use nix::sys::signal;
+
+    extern "C" fn handle_signal(signum: i32) {
+        let action = signal::SigAction::new(
+            signal::SigHandler::SigDfl,
```
def-: Linux flags an init process as `SIGNAL_UNKILLABLE`, so `SIG_DFL` doesn't work and clusterd would keep running on SIGTERM. With this diff:
```diff
diff --git a/test/cluster/mzcompose.py b/test/cluster/mzcompose.py
index f1f3042d21..b96c555b51 100644
--- a/test/cluster/mzcompose.py
+++ b/test/cluster/mzcompose.py
@@ -15,6 +15,7 @@ usual clusterd included in the materialized container).
 import json
 import random
 import re
+import subprocess
 import time
 from collections.abc import Callable
 from copy import copy
@@ -2346,6 +2347,36 @@ def workflow_test_clusterd_death_detection(c: Composition) -> None:
         assert "replica task failed: timed out" in envd.stdout


+def workflow_test_clusterd_sigterm_pid1(c: Composition) -> None:
+    clusterd = Clusterd(name="clusterd1")
+    clusterd.config["entrypoint"] = []
+
+    with c.override(clusterd):
+        c.up("materialized", "clusterd1")
+        container_id = c.container_id("clusterd1")
+        assert container_id is not None, "clusterd1 is not running"
+
+        c.kill("clusterd1", signal="SIGTERM", wait=False)
+
+        try:
+            exit_code = int(
+                subprocess.run(
+                    ["docker", "wait", container_id],
+                    check=True,
+                    capture_output=True,
+                    text=True,
+                    timeout=15,
+                ).stdout
+            )
+        except subprocess.TimeoutExpired:
+            raise AssertionError(
+                "clusterd ignored SIGTERM as PID 1: the SIG_DFL re-raise was "
+                "dropped (SIGNAL_UNKILLABLE), so the process never terminated"
+            )
+
+        assert exit_code == 143, f"expected exit 143 (128 + SIGTERM), got {exit_code}"
+
+
 class Metrics:
     metrics: dict[str, str]
```
Running `bin/mzcompose --find cluster run test_clusterd_sigterm_pid1` on this PR gives me:
```
==> mzcompose: test case workflow-test_clusterd_sigterm_pid1 failed: builtins.AssertionError: clusterd ignored SIGTERM as PID 1: the SIG_DFL re-raise was dropped (SIGNAL_UNKILLABLE), so the process never terminated
  File "/home/deen/git/materialize-review-file/test/cluster/mzcompose.py", line 2379, in workflow_test_clusterd_sigterm_pid1
    subprocess.run(
    ~~~~~~~~~~~~~~^
        ["docker", "wait", container_id],
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<3 lines>...
        timeout=15,
        ^^^^^^^^^^^
    ).stdout
    ^
  File "/home/deen/.local/share/uv/python/cpython-3.13.2-linux-x86_64-gnu/lib/python3.13/subprocess.py", line 558, in run
    stdout, stderr = process.communicate(input, timeout=timeout)
                     ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/deen/.local/share/uv/python/cpython-3.13.2-linux-x86_64-gnu/lib/python3.13/subprocess.py", line 1221, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
                     ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/deen/.local/share/uv/python/cpython-3.13.2-linux-x86_64-gnu/lib/python3.13/subprocess.py", line 2131, in _communicate
    self._check_timeout(endtime, orig_timeout, stdout, stderr)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/deen/.local/share/uv/python/cpython-3.13.2-linux-x86_64-gnu/lib/python3.13/subprocess.py", line 1265, in _check_timeout
    raise TimeoutExpired(
    ...<2 lines>...
            stderr=b''.join(stderr_seq) if stderr_seq else None)
subprocess.TimeoutExpired: Command '['docker', 'wait', '5ab6305d27a9f932d98b726629e96d8037e998d2f9c73873a2a95547e41184b9']' timed out after 15 seconds

During handling of the above exception, another exception occurred:

  File "/home/deen/git/materialize-review-file/misc/python/materialize/mzcompose/composition.py", line 709, in test_case
    yield
  File "/home/deen/git/materialize-review-file/misc/python/materialize/cli/mzcompose.py", line 857, in handle_composition
    composition.workflow(
2315                   "listen": "0.0.0.0:2101",
    ~~~~~~~~~~~~~~~~~~~~^
        workflow_name, *args.unknown_subargs[1:], *extra_args
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/deen/git/materialize-review-file/misc/python/materialize/mzcompose/composition.py", line 597, in workflow
    func(self)
    ~~~~^^^^^^
  File "/home/deen/git/materialize-review-file/test/cluster/mzcompose.py", line 2388, in workflow_test_clusterd_sigterm_pid1
    raise AssertionError(
    ...<2 lines>...
    )
AssertionError: clusterd ignored SIGTERM as PID 1: the SIG_DFL re-raise was dropped (SIGNAL_UNKILLABLE), so the process never terminated
```
So what does this actually achieve?

def- (changes requested):

jasonhernandez:

Part of a set that ships together: **#36099** (distroless image), **#36101** (UID/GID gating), **#36876** (removes the unused CTP FQDN check, stacks on this).

**Merge order:** land **before or with** #36099. Without the in-process `CLUSTERD_PROCESS` derivation here, distroless clusterd has no process ordinal and won't start. Safe to land first: current Ubuntu images still set `CLUSTERD_PROCESS` via `entrypoint.sh`, and the in-process path only runs when it is unset.

No signal handling is included: `tini` (added to the distroless image in #36099) runs as PID 1 and forwards signals, so clusterd is a child process and SIGTERM takes its default action. This supersedes the earlier SIGTERM-handler framing, which assumed clusterd would be PID 1.

src/clusterd/src/lib.rs
jasonhernandez: it turns out that tini is the right choice here and I was able to land a static build of it instead of trying to build our own init.

src/clusterd/src/lib.rs
jasonhernandez: I took a different approach where we're keeping tini, but we're going to statically build that so we don't need dependencies in the distroless images.

def- (approved):

PR #36874, opened 2026-06-02
cluster: fix false positives in clusterd-not-receiving-commands alert

antiguru:
### Motivation

Fixes [CLU-103](https://linear.app/materializeinc/issue/CLU-103/fix-false-positives-in-clusterd-command-alert).

The `clusterd-not-receiving-commands` alert reads the server-side "last command received" timestamp and fires when it goes stale. The alert was silently broken after the switch from gRPC to CTP (#32897), and once re-pointed at the metric it began firing false positives on healthy clusters.

**Root cause:** the server updated the timestamp only on fully decoded command messages (`message_received`). CTP exchanges keepalives every second on otherwise idle connections, but those empty keepalive frames are dropped before decoding and never reach `message_received`. So on a healthy-but-idle cluster that hadn't received a command in a while, the metric went stale and the alert false-fired.

### Changes

* Update the timestamp on any inbound bytes (`bytes_received`) instead, which includes keepalives. This turns the metric into a connection-liveness signal: it stays fresh as long as the controller connection is healthy and only goes stale once the connection is actually broken — the failure mode the alert was built to detect (the original ~38h General Mills incident).
* Rename the metric `mz_grpc_server_last_command_received` → `mz_cluster_server_last_command_received`, since it is no longer gRPC-specific.
* Move `ClusterServerMetrics` / `PerClusterServerMetrics` out of the vestigial `mz_service::grpc` module into the CTP `transport` module, and delete the now-empty `grpc.rs`. (Persist PubSub's gRPC is separate code and is unaffected.)
* Add a regression test to `workflow_test_replica_metrics` asserting the metric stays fresh on an idle but healthy connection.

### Note for reviewer / rollout

The alert query lives in the external monitoring repo, not in this tree. It must be repointed from `mz_grpc_server_last_command_received` to `mz_cluster_server_last_command_received` for the alert to keep working — the old name is no longer emitted.

### Checklist

https://claude.ai/code/session_01FH8199FBgaosaqv6RcmZ98


materializekatrina (approved):
TY!

DAlperin:
LGTM in principle but if we can, lets just drop the alias? @materializekatrina can correct me if I'm wrong but it should be pretty easy to just fix the alert once this merges

materializekatrina:
Yep, fine to drop the alias. Once it merges I'd want to watch the new metric for a week or two before wiring an alert back up, just to confirm it's reliable. The current alert is already silenced to avoid the false positives, so there's no harm in removing it in the meantime.

antiguru:
Great, this simplifies things! I'll drop the alias and also clean up around the code a bunch. Thank y'all for the reviews.

PR #36876, opened 2026-06-02
clusterd: remove unused CTP server-FQDN validation

jasonhernandez:
Removes the CTP `server_fqdn` handshake check that `CLUSTERD_GRPC_HOST` fed. Despite the name it has **nothing to do with gRPC** (which is now only persist pubsub). In the CTP handshake, clusterd advertised its FQDN and the controller compared it against the address it dialed, failing on mismatch (`transport.rs::handshake`). That check is:

- **optional** — only fires when both sides set a value, so it's already a no-op when the host is unset;
- **narrow** — guards only against reaching a misrouted/stale replica (DNS / pod-IP reuse);
- **misnamed**, which is what prompted the review question on the previous PR.

The distroless migration removes `entrypoint.sh`, which set `CLUSTERD_GRPC_HOST` via `hostname --fqdn`. Rather than re-plumb that, this removes the feature: drops `--grpc-host`/`CLUSTERD_GRPC_HOST`, the `server_fqdn` field from the CTP `Hello`, the `host_from_address` helper, and the `test_handshake_fqdn_mismatch` test.

This is rebased to a single commit on `main` and stands alone: it touches only `clusterd`/`service`, not `entrypoint.sh` or any Dockerfile, so it can merge ahead of the distroless image PR (#36099).

### Notes
- Wire change: `Hello` loses a field. CTP version-gates the handshake (mismatched versions fail and reconnect), so this is safe across a release boundary.
- `test_metrics` byte-count bounds were loosened (the handshake shrank).

This is the "rip it out" answer to the FQDN question; it makes #36100 (in-process resolve) unnecessary.

## Test plan
- [x] `cargo check -p mz-clusterd -p mz-service -p mz-compute-client -p mz-storage-controller` (rustc 1.96.0)
- [x] `cargo test -p mz-service --test transport` — green across 10 consecutive runs
- [x] Confirm controller↔replica CTP connects in a k8s/kind cluster — verified via cloudtest on kind: `test_smoke` (`test_wait`: default cluster `u1` replica reached `Ready`, i.e. the CTP handshake succeeded; `test_sql`) and `test_testdrive` (multi-replica: `c1` with `r1`/`r2`, `c2` at replication factor 2, Kafka source, materialized views) all pass. Confirmed the deployed `clusterd` binary no longer exposes `--grpc-host`.

### DIFF
--- a/src/clusterd/src/lib.rs
+++ b/src/clusterd/src/lib.rs
@@ -81,12 +81,6 @@ struct Args {
         default_value = "127.0.0.1:6878"
     )]
     internal_http_listen_addr: SocketAddr,
-    /// The FQDN of this process, for GRPC request validation.
-    ///
-    /// Not providing this value or setting it to the empty string disables host validation for
-    /// GRPC requests.
-    #[clap(long, env = "GRPC_HOST", value_name = "NAME")]
-    grpc_host: Option<String>,
 
     // === Timely cluster options. ===
     /// Configuration for the storage Timely cluster.
@@ -406,7 +400,6 @@ async fn run(args: Args) -> Result<(), anyhow::Error> {
         None,
     );
 
-    let grpc_host = args.grpc_host.and_then(|h| (!h.is_empty()).then_some(h));
     let cluster_server_metrics = ClusterServerMetrics::register_with(&metrics_registry);
 
     let mut storage_timely_config = args.storage_timely_config;
@@ -451,7 +444,6 @@ async fn run(args: Args) -> Result<(), anyhow::Error> {
         transport::serve(
             args.storage_controller_listen_addr,
             BUILD_INFO.semver_version(),
-            grpc_host.clone(),
             Duration::MAX,
             storage_client_builder,
             cluster_server_metrics.for_server("storage"),
@@ -483,7 +475,6 @@ async fn run(args: Args) -> Result<(), anyhow::Error> {
         transport::serve(
             args.compute_controller_listen_addr,
             BUILD_INFO.semver_version(),
-            grpc_host.clone(),
             Duration::MAX,
             compute_client_builder,
             cluster_server_metrics.for_server("compute"),
--- a/src/service/src/transport.rs
+++ b/src/service/src/transport.rs
@@ -66,29 +66,12 @@ impl<Out: Message, In: Message> Client<Out, In> {
         idle_timeout: Duration,
         metrics: impl Metrics<Out, In>,
     ) -> anyhow::Result<Self> {
-        let dest_host = host_from_address(address);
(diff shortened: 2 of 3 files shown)


jasonhernandez:

Removes the unused CTP server-FQDN check, whose only supplier (`CLUSTERD_GRPC_HOST`, set by the shell entrypoint) goes away with the distroless migration (#36099).

**Stacks on #36872.** Off the critical path: a missing `CLUSTERD_GRPC_HOST` already makes the check a no-op (it only fires when both sides set a value), so this is pure dead-code cleanup and can land any time after #36872 rather than being gated on the cutover.

Part of the set: #36099 (image), #36101 (UID/GID gating), #36872 (clusterd process ordinal). Supersedes the FQDN half of the closed #36100.

def- (commented):
From the QA LLM review:
## MEDIUM - Removing the CTP host check leaves no process-identity validation for partitioned replica connections

This PR removes `--grpc-host`/`CLUSTERD_GRPC_HOST`, stops passing a server identity into `transport::serve`, and reduces the CTP `Hello` handshake to only a semver check (`src/service/src/transport.rs:395-431`). After this change, a controller that dials a stale or misrouted replica address accepts any clusterd process running the same Materialize semver.

That is a real correctness contract in this code path, not just defense in depth. Managed Kubernetes replicas build an ordered address list from StatefulSet pod DNS names (`src/orchestrator-kubernetes/src/lib.rs:688-698`). The controller then constructs a `Partitioned` client from that ordered list, and the connection index determines which command shard is sent and how responses are reassembled (`src/service/src/client.rs:141-148`, `:158-164`). Compute makes this especially sensitive because all commands except `Hello` and `UpdateConfiguration` are sent only to shard 0 (`src/compute-client/src/service.rs:368-379`).

Before this PR, managed clusterd pods set `CLUSTERD_GRPC_HOST` from `hostname --fqdn`, so the handshake rejected a connection where the dialed host did not match the server's own FQDN. That caught the exact stale-DNS or pod-IP-reuse case called out in the PR description. After this PR, the same stale address can silently connect to a different clusterd process of the same semver. The new connection also cancels any existing CTP connection on that process (`src/service/src/transport.rs:144-151`), so the failure mode can corrupt the controller's partition mapping and disrupt the correctly connected replica at the same time.

The Kubernetes orchestrator explicitly uses zero termination grace and documents that duplicate pod identities and delayed network packets are part of the model (`src/orchestrator-kubernetes/src/lib.rs:1207-1228`). In that environment, pod DNS and pod-IP reuse are not hypothetical enough to remove the only peer identity check without replacing it. A safer fix is to keep the check until CTP carries an explicit controller-assigned process identity, preferably including at least cluster, replica, process ordinal, and generation, and rejects peers whose identity does not match the address slot being connected.

PR #36877, opened 2026-06-02
optimizer: include panic location and backtrace in caught panic errors

antiguru:
### Motivation

When a transform panics, `catch_unwind_optimize` demotes it to an internal optimizer error. Today we only get the panic *message* — no location, no backtrace — so an error like:

```
internal error in optimizer: internal transform error: unexpected panic during query optimization: index out of bounds: the len is 2 but the index is 18446744073709551615
```

gives no hint about *where* in the optimizer the panic happened, making these bugs hard to track down.

The panic location and backtrace are only available inside the panic handler (which runs *before* the stack is unwound), but `mz_ore`'s enhanced panic handler returns early — without printing or reporting — while catching an unwind. So that context was discarded, and `catch_unwind_str` only recovered the payload message. There was even a long-standing `TODO(teskje): collect and log a backtrace from the panic site` at the call site.

### Changes

* **`mz_ore::panic`** — add an opt-in capture path:
  * `catch_unwind_with_details(...)` works like `catch_unwind_str`, but on panic returns a `CaughtPanic { message, location, backtrace }`.
  * While inside such a call, the enhanced panic handler stashes the panic location and a backtrace into a thread-local before letting the unwind proceed, instead of throwing them away.
  * This is a *parallel* infra: the backtrace is only captured when a panic actually fires (the exceptional case), so there's **no per-optimization cost**, and other `catch_unwind` callers are completely unaffected. If the enhanced handler isn't installed, the message is still recovered and `location`/`backtrace` are simply absent.

* **`mz_transform`** — `catch_unwind_optimize` now uses `catch_unwind_with_details`:
  * The **user-facing error** includes at least the panic location: `... unexpected panic during query optimization: <msg> (at <file:line:col>)`.
  * The **full backtrace** is logged via `tracing::error!` (resolving the old TODO).

### Tips for reviewer

The behavioral change for existing `catch_unwind`/`catch_unwind_str` callers is nil — they don't set the `CAPTURE_PANIC_DETAILS` thread-local, so the handler's early-return path is unchanged for them.

This PR was drafted by Claude Code at the request of @mh.

https://claude.ai/code/session_01Av4N3HccSLB4Y3EAjvdngx

---
_Generated by [Claude Code](https://claude.ai/code/session_01Av4N3HccSLB4Y3EAjvdngx)_

### DIFF
--- a/src/ore/src/panic.rs
+++ b/src/ore/src/panic.rs
@@ -19,6 +19,7 @@ use std::any::Any;
 use std::backtrace::Backtrace;
 use std::borrow::Cow;
 use std::cell::RefCell;
+use std::fmt;
 use std::fs::File;
 use std::io::{self, Write as _};
 use std::os::fd::FromRawFd;
@@ -39,6 +40,50 @@ use crate::iter::IteratorExt;
 thread_local! {
     /// Keeps track of how many `catch_unwind` calls we are inside.
     static CATCHING_UNWIND: RefCell<usize> = const { RefCell::new(0) };
+
+    /// Keeps track of how many [`catch_unwind_with_details`] calls we are inside.
+    ///
+    /// When non-zero, the enhanced panic handler records details (the panic
+    /// location and a backtrace) about caught panics into [`CAUGHT_PANIC_DETAILS`]
+    /// before letting the unwind proceed.
+    static CAPTURE_PANIC_DETAILS: RefCell<usize> = const { RefCell::new(0) };
+
+    /// Details about the most recently caught panic, recorded by the enhanced
+    /// panic handler when [`CAPTURE_PANIC_DETAILS`] is non-zero. Consumed by
+    /// [`catch_unwind_with_details`].
+    static CAUGHT_PANIC_DETAILS: RefCell<Option<PanicDetails>> = const { RefCell::new(None) };
+}
+
+/// Details about a caught panic, captured at the panic site by the enhanced
+/// panic handler. See [`catch_unwind_with_details`].
+#[derive(Clone, Debug, Default)]
+pub struct PanicDetails {
+    /// The source code location at which the panic occurred, if known.
+    pub location: Option<String>,
+    /// A backtrace captured at the panic site, if one was captured.
+    pub backtrace: Option<String>,
+}
+
+/// A panic recovered by [`catch_unwind_with_details`], bundling the panic
+/// message with the location and backtrace captured at the panic site.
+#[derive(Clone, Debug)]
+pub struct CaughtPanic {
+    /// The panic message.
+    pub message: Cow<'static, str>,
+    /// The source code location at which the panic occurred, if known.
+    pub location: Option<String>,
+    /// A backtrace captured at the panic site, if one was captured.
+    pub backtrace: Option<String>,
+}
+
+impl fmt::Display for CaughtPanic {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.message)?;
+        if let Some(location) = &self.location {
+            write!(f, ", at {location}")?;
+        }
+        Ok(())
+    }
 }
 
 #[cfg(feature = "async")]
@@ -93,6 +138,24 @@ pub fn install_enhanced_handler() {
         #[cfg(not(feature = "async"))]
         let catching_unwind_async = false;
         if catching_unwind != 0 || catching_unwind_async {
+            // We're letting the unwind proceed without printing or reporting the
+            // panic. The location and backtrace would normally be lost here,
+            // because `catch_unwind` only recovers the panic payload (the
+            // message). If a caller has opted in via `catch_unwind_with_details`,
+            // stash those details now so they can be attached to the resulting
+            // error. We only pay the cost of capturing a backtrace when a panic
+            // actually occurs, which is the exceptional case.
+            let capture_details = CAPTURE_PANIC_DETAILS.with(|v| *v.borrow()) != 0;
+            if capture_details {
+                let location = panic_info.location().map(|loc| loc.to_string());
+                let backtrace = Backtrace::force_capture().to_string();
+                CAUGHT_PANIC_DETAILS.with(|details| {
+                    *details.borrow_mut() = Some(PanicDetails {
+                        location,
+                        backtrace: Some(backtrace),
+                    });
+                });
+            }
             return;
         }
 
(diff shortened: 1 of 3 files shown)


ggevay (commented):
Edit: Sorry, AI hiccup :D Will post again shortly.

src/transform/src/lib.rs
```diff
@@ -335,11 +335,26 @@ where
...
+            let backtrace = panic
+                .backtrace
+                .as_deref()
+                .unwrap_or("<backtrace unavailable>");
+            tracing::error!(
+                "caught a panic during query optimization: {} (at {})\n{}",
+                panic.message,
+                location,
+                backtrace,
+            );
```
antiguru, line 351: Pass variables verbatim to `error!` instead of encoding in the string. This'll lead to better error messages.

src/ore/src/panic.rs
```diff
@@ -39,6 +40,50 @@ use crate::iter::IteratorExt;
...
+
+/// A panic recovered by [`catch_unwind_with_details`], bundling the panic
+/// message with the location and backtrace captured at the panic site.
+#[derive(Clone, Debug)]
+pub struct CaughtPanic {
+    /// The panic message.
+    pub message: Cow<'static, str>,
+    /// The source code location at which the panic occurred, if known.
+    pub location: Option<String>,
+    /// A backtrace captured at the panic site, if one was captured.
```
ggevay: Format drift: `CaughtPanic`'s `Display` produces `"<msg>, at <loc>"` (comma), while `catch_unwind_optimize` in `src/transform/src/lib.rs` formats it as `"<msg> (at <loc>)"` (parentheses). Consider aligning the two so callers can just `format!("...: {panic}")` and log scrapers see one shape.

src/ore/src/panic.rs
```diff
@@ -39,6 +40,50 @@ use crate::iter::IteratorExt;
...
+    /// location and a backtrace) about caught panics into [`CAUGHT_PANIC_DETAILS`]
+    /// before letting the unwind proceed.
+    static CAPTURE_PANIC_DETAILS: RefCell<usize> = const { RefCell::new(0) };
+
+    /// Details about the most recently caught panic, recorded by the enhanced
+    /// panic handler when [`CAPTURE_PANIC_DETAILS`] is non-zero. Consumed by
+    /// [`catch_unwind_with_details`].
+    static CAUGHT_PANIC_DETAILS: RefCell<Option<PanicDetails>> = const { RefCell::new(None) };
+}
+
```
ggevay, line 56: Minor: `PanicDetails::backtrace` is `Option<String>` but it's always populated to `Some(..)` at the only write site, and `PanicDetails` itself sits behind an `Option` in the thread-local. The double-nullability is mildly redundant — either drop the inner `Option` or add a comment explaining the intent. Also consider `#[non_exhaustive]` since this struct seems likely to grow (thread name, timestamp, ...).

src/ore/src/panic.rs
```diff
@@ -39,6 +40,50 @@ use crate::iter::IteratorExt;
 thread_local! {
     /// Keeps track of how many `catch_unwind` calls we are inside.
     static CATCHING_UNWIND: RefCell<usize> = const { RefCell::new(0) };
+
+    /// Keeps track of how many [`catch_unwind_with_details`] calls we are inside.
+    ///
```
ggevay, line 45: `Backtrace::force_capture()` ignores `RUST_BACKTRACE`, whereas the aborting handler above honors `RUST_BACKTRACE=full` vs short. Defensible here (operators debugging a rare optimizer panic want everything), but worth a one-line note in the doc comment on `catch_unwind_with_details` so future callers aren't surprised by the inconsistency.

src/ore/src/panic.rs
```diff
@@ -39,6 +40,50 @@ use crate::iter::IteratorExt;
...
+/// Details about a caught panic, captured at the panic site by the enhanced
+/// panic handler. See [`catch_unwind_with_details`].
+#[derive(Clone, Debug, Default)]
+pub struct PanicDetails {
+    /// The source code location at which the panic occurred, if known.
+    pub location: Option<String>,
+    /// A backtrace captured at the panic site, if one was captured.
+    pub backtrace: Option<String>,
+}
+
```
ggevay, line 67: nit: this downcast ladder is a verbatim copy of `catch_unwind_str`'s. Extracting a small helper (e.g. `fn downcast_panic_message(opaque: Box<dyn Any + Send>) -> Cow<'static, str>`) would keep the two in sync as future payload types (e.g. `panic_any`) are added.

src/transform/src/lib.rs
```diff
@@ -335,11 +335,26 @@ where
         }
         Ok(result) => result.map_err(|e| e),
```
ggevay: Pre-existing, not introduced by this PR, but while you're touching this arm: `result.map_err(|e| e)` is a no-op and could just be `result`.

src/transform/src/lib.rs
antiguru: Done in 16010b6 — `location`, `backtrace`, and `message` are now passed as structured fields rather than interpolated into the message string.

---

ggevay (commented):
Some AI comments:

A few non-blocking suggestions on the new infra. Behavior looks correct for the stated goal; comments below are about robustness/clarity for future callers.

src/ore/src/panic.rs
antiguru: Aligned in c28110a — `Display` now produces `"<msg> (at <loc>)"`, and `catch_unwind_optimize` just does `format!("...: {panic}")`.

---

src/ore/src/panic.rs
antiguru: Done in c28110a. `PanicDetails` is purely internal, so I made it private and changed `backtrace` to a plain `String` (it's always populated at the write site). The remaining nullability lives on the public `CaughtPanic`, where `backtrace: Option<String>` is meaningful (absent when the enhanced handler isn't installed). Also added `#[non_exhaustive]` to `CaughtPanic`.

---

src/ore/src/panic.rs
antiguru: Added a note to the doc comment in c28110a calling out that this always captures a full backtrace via `force_capture` and intentionally does not honor `RUST_BACKTRACE`.

---

src/ore/src/panic.rs
antiguru: Extracted `downcast_panic_message` in c28110a; both `catch_unwind_str` and `catch_unwind_with_details` now call it.

---

src/transform/src/lib.rs
antiguru: Simplified to `Ok(result) => result` in c28110a.

---

src/ore/src/panic.rs
antiguru: Reverted the `#[non_exhaustive]` part in 615d893 — agreed, since all clients are in-tree it would just let consumers silently miss new fields instead of failing to compile. Kept the `PanicDetails` cleanup (private, `backtrace: String`).

---

antiguru:
Thanks, I addressed your comments. Mind taking another look?

ggevay (approved):
Thanks!

antiguru:
Thanks for the review!

PR #36878, opened 2026-06-02
console: show real balancerd host in the OIDC connect modal

leedqin:
Read balancerd_dns_names from app-config and use
the first entry as the SQL host in the OIDC connect modal and the MCP server hostname. Falls back to the <host> placeholder when the field is absent. Drop the "Replace <host>..." instruction from the intro when a real host is available.

### Motivation

Fixes [CNS-87](https://linear.app/materializeinc/issue/CNS-87/use-balancerd-dns-names-in-console-app-config-to-show-host-name-in)

### Verification
- Reading the host name from app config
image

- Host name in MCP connect modal
image

- When no host name is available 
image

### DIFF
--- a/console/src/components/McpConnectInstructions.tsx
+++ b/console/src/components/McpConnectInstructions.tsx
@@ -61,12 +61,15 @@ const McpConnectInstructions = ({
 
   if (!envAddress) return null;
 
-  // Cloud: HTTPS with the environment's HTTP address hostname.
-  // Self-managed: Use a placeholder since the MCP endpoint may be behind a
-  // load balancer or custom domain that we can't determine from the console.
+  const balancerdHost =
+    appConfig.mode === "self-managed"
+      ? appConfig.balancerdDnsNames?.[0]
+      : undefined;
   const baseUrl = isCloud
     ? `https://${envAddress.split(":")[0]}`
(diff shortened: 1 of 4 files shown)


console/src/components/McpConnectInstructions.tsx
```diff
@@ -61,12 +61,15 @@ const McpConnectInstructions = ({
 
   if (!envAddress) return null;
 
-  // Cloud: HTTPS with the environment's HTTP address hostname.
-  // Self-managed: Use a placeholder since the MCP endpoint may be behind a
-  // load balancer or custom domain that we can't determine from the console.
+  const balancerdHost =
+    appConfig.mode === "self-managed"
```
Alphadelta14, line 65: do we not propagate the same for cloud though?

Alphadelta14 (approved):

console/src/components/McpConnectInstructions.tsx
SangJunBak: I don't think so given we don't host the Console for Cloud. Also I don't think we define balancerdExternalCertificateSpec in Cloud either.

SangJunBak (approved):
In the sample `app-config.json`, we should probably default that to an empty array!

PR #36886, opened 2026-06-03
persist: diagnose lease state on missing-blob fetch

antiguru:
### Motivation

Investigating `halting process: unhandled error in compute_import: batch fetcher could not fetch batch part` halts ([PER-21](https://linear.app/materializeinc/issue/PER-21/1036-follow-up-investigate-halting-process-unhandled-error-in-compute)).
The halt message guesses `lost lease?` but never checks, so each occurrence requires a manual investigation that usually can't distinguish a lost lease from a garbage collection bug.

### Description

When a fetch hits a missing blob, refresh the shard state (with a timeout, since a consensus partition may be the very reason the lease was lost) and look up the minting reader before halting:

* reader expired out of state → the lease was lost, e.g. because the process failed to heartbeat it within the lease duration (machine sleep, CPU/memory starvation, partition from consensus)
* reader still present → the seqno hold did not protect the blob, which points at a GC or lease-tracking bug

To make this possible on the fetch side, `LeasedBatchPart` and `ExchangeableBatchPart` now carry the id of the reader that leased the part; fetcher workers and processes previously had no way to know whose lease protected a part.
This mirrors the existing check in `Listen::next`, which distinguishes the same two cases when a batch advances past the listen frontier.

### DIFF
--- a/src/persist-client/src/fetch.rs
+++ b/src/persist-client/src/fetch.rs
@@ -259,7 +259,7 @@ where
     }
 
     /// Diagnoses a missing-blob fetch failure for a part leased by the given
-    /// reader. See [missing_blob_diagnostics].
+    /// reader. See the free function `missing_blob_diagnostics`.
     pub async fn missing_blob_diagnostics(&self, reader_id: &LeasedReaderId) -> String {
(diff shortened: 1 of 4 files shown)


DAlperin (approved):
LGTM thanks

PR #36887, opened 2026-06-03
SS-169 Update handling for flaky kafka connections when fetching low watermarks

patrickwwbutler:
We found that the orevious strategy of connection handling led to some problematic behavior when dealing with flaky connections that led to unhealthy sources and unfresh data, so this updates the handling to behave more like the way `fetch_partition_info` normally handles connection errors, by emitting a `stalled` health status update for the kafka connection and the SSH tunnel.


martykulma (approved):
🐟

PR #36891, opened 2026-06-03
pager: recycle scratch files and pool page-in read buffers

DAlperin:
## What

Profile-driven optimization of the file-backed pager (`mz_ore::pager` file
backend) that spills columnar merge-batcher data. A CPU profile of a production
clusterd put `mz_ore::pager::file::try_take_file` as the #2 on-CPU frame across
the whole process (~1m18s in a 15-minute window). This PR removes ~46s of that
with two independent changes, plus the design doc.

Take-path cost breakdown from the profile:

| Cost | Before | After this PR |
|---|---|---|
| unlink / inode-evict | 35.6s | ~0 (recycle) |
| read copy (`copy_to_user`, irreducible) | 22.8s | 22.8s |
| destination-buffer zero + page-fault | 13.4s | ~2.7s (memset only) |
| `File::open` | 4.3s | 4.3s (left for a follow-up) |

## Tier 1 — recycle scratch files instead of unlinking (`ore/pager`)

The backend unlinked each scratch file on take/drop. Unlink was dominated not
by the directory op but by `ext4_evict_inode -> truncate_inode_pages_final` —
synchronously tearing down the page cache the read had just populated, on the
worker thread. For a write-once / read-once / delete temp file the whole inode
lifecycle is pure overhead.

Instead, take/drop returns the slot to a capped free-list, keeping the inode and
its warm page cache for the next `pageout` to overwrite from offset 0. Reads are
length-bounded so a stale tail is never observed. Slots over a size bound or
past the cap fall back to unlink, bounding retained (writeback-reclaimable) page
cache. The no-fd-retained invariant is preserved.

## Buffer warmth — warm page-in read buffers (`timely-util`)

`ColumnPager::take` read each file-backed payload into a freshly allocated
`Vec<u64>`, so the kernel faulted in and zeroed cold anonymous pages on the
read's `copy_to_user` — the bulk of the 13.4s above. A per-worker
(thread-local) pool of warm read buffers lets file-backed take read into
resident memory; consumed `Column::Align` chunks return their backing at the
merge batcher's single `recycle_capped` chokepoint. The pool is bounded like the
existing chunk stash. Swap-backed reads keep their own path via a new
`Handle::is_file()`. (The existing `recycle_chunk` stash could not be reused: it
keeps only `Column::Typed` and drops `Align`/`Bytes`, which is exactly what
`take` produces.)

## Design

`doc/developer/design/20260603_pager_pooling.md` documents both changes and a
third tier — an open-fd cache keyed by slot id — that is designed but
intentionally **not** implemented here. It is the smallest remaining lever
(~8s, hit-rate-gated) and is gated on a re-profile after these land.

## Tests

Added to `mz_ore::pager::file`: recycle-on-take, recycle-on-drop, slot reuse with
stale-tail safety, oversize-not-recycled, and free-list cap. Added to
`column_pager`: warm-buffer reuse and cap/oversize rejection. Existing
file-backed round-trip tests in both crates now exercise the recycle and pooling
paths.

Part of [CLU-65](https://linear.app/materializeinc/issue/CLU-65).

### DIFF
--- a/doc/developer/design/20260603_pager_pooling.md
+++ b/doc/developer/design/20260603_pager_pooling.md
@@ -1,0 +1,229 @@
+# Pager file + buffer pooling
+
+* Associated: [CLU-65](https://linear.app/materializeinc/issue/CLU-65/pager) (follow-up to the original pager design, `20260504_pager.md`).
+
+## The problem
+
+The original pager design parked cross-handle file pooling as out of scope:
+
+> Cross-handle file pooling (one shared scratch file with a free-list). Each
+> handle owns one named file in the scratch directory; pooling is a follow-up
+> if inode pressure shows up.
+
+Inode pressure shows up. A 15-minute on-CPU profile of a production `u6`
+clusterd dominated by columnar merge-batcher work puts
+`mz_ore::pager::file::try_take_file` second across the entire process at
+**1m17.8s cumulative**. The cost splits four ways:
+
+| Cost | Time | Share | What it is |
+|---|---|---|---|
+| `FileInner::drop` → `remove_file` (unlink) | 35.6s | 46% | Deleting the scratch file after read |
+| `read_at` (pread + `copy_to_user`) | 22.8s | 29% | The read itself |
+| `Vec<u64>::resize` (zero-fill) | 13.4s | 17% | Zeroing `dst` before reading into it |
+| `File::open` | 4.3s | 5% | Path walk + open syscall |
+
+Two findings drive this design:
+
+1. **Unlink is the dominant cost, and it is not the directory operation.** Of
+   the 35.6s, ~24.8s is `iput_final → evict → ext4_evict_inode`, and ~18.4s of
+   that is `truncate_inode_pages_final` — the kernel synchronously tearing down
+   every page-cache page that the immediately-preceding read populated, on the
+   worker thread, in the hot loop. For a write-once / read-once / delete temp
+   file the entire ext4 inode lifecycle is pure overhead: we create an inode,
+   populate its page cache, then destroy both, per spill.
+
+2. **An open-fd cache would not help the hot path.** The merge batcher drains
+   chains via `FetchIter::next`, which calls `ColumnPager::take` exactly once
+   per entry: open, read fully, delete. Each scratch file is touched by exactly
+   one open and one read in its lifetime. There is no second access to amortize
+   an open against, so retaining file descriptors keyed by handle buys nothing
+   here and reintroduces the fd-exhaustion risk the original design eliminated.
+   `read_at_many` (the write-once / read-many random-access API) is the only
+   shape an fd cache would help, and it is absent from this workload.
+
+The redundant zero-fill (row 3) is mostly irreducible in isolation: skipping
+`dst.resize(.., 0)` only saves the ~2.7s userspace memset. The
+`clear_page`/`alloc_anon_folio` faulting (~10s) is the kernel zeroing fresh
+anonymous pages on first write-fault, which `read_at`'s `copy_to_user` triggers
+regardless of whether we pre-zeroed. That cost is recoverable only by keeping
+the destination memory warm across takes (buffer reuse), not by reordering the
+zero.
+
+## Success criteria
+
+* The common merge-batcher churn (pageout and take interleaved within a merge)
+  stops paying inode create + unlink + page-cache evict per spill.
+* Peak open file descriptors stay bounded by a small constant, independent of
+  how many handles are simultaneously paged out. The original design's core
+  invariant — fd count decoupled from handle count — is preserved.
+* The destination-buffer faulting cost is addressed for the steady state
+  without holding unbounded resident memory off the pager's budget.
+* No change to existing `mz_ore::pager` signatures or the
+  handle-survives-backend-flip contract. The buffer-warmth change adds one
+  accessor — `Handle::is_file` — so the `ColumnPager` layer can route a warm
+  buffer only for file-backed takes (the swap backend keeps its allocation
+  resident, so a warm buffer there is wasted). The routing lives in the caller
+  rather than inside the pager because a pager-internal pool can't see the
+  caller's recycling chokepoint — the merge batcher's per-chunk reuse — where
+  the warm buffer has to come from and return to.
+* Crash-safety story is unchanged: stale scratch is reclaimed with the pod's
+  ephemeral volume; pooling adds no new persistent state.
+
+## Out of scope
+
+* Async / `io_uring`. Still sync syscalls.
+* The swap backend. Pooling is a file-backend concern only.
+* Compression interaction. The pool stores opaque `Vec<u64>` payloads exactly
+  as today; the `ColumnPager` codec layer is unaffected.
+
+## Solution proposal
+
+Two independent, separately-shippable changes. Tier 1 attacks the 35.6s unlink;
+Tier 2 attacks the 4.3s open; the buffer story attacks the ~13s faulting. They
+compose but do not depend on each other.
+
+### Tier 1 — inode free-list (reuse files instead of unlinking)
+
+Today `FileInner::drop` calls `remove_file`, which unlinks the dirent and, as
+the last link drops, evicts the inode and truncates its page cache. Instead,
+maintain a per-process free-list of **scratch slots** — named files that exist
+on disk but back no live handle.
+
+* `take` / drop, instead of `remove_file`, returns the slot id to the
+  free-list. The file stays linked; its inode and (warm) page cache survive.
+* `pageout` pops a free slot if one is available and writes the new payload at
+  offset 0. Reads are already bounded by `len_u64s`, so stale bytes past the
+  new length are never observed — no `ftruncate` needed.
+* The free-list is capped (à la the merge batcher's `STASH_CAP` /
+  `MAX_RECYCLE_BYTES`). Beyond the cap, or for payloads larger than a recycle
+  threshold, fall back to today's create-new / unlink-on-drop path so we don't
+  hoard inodes or pin oversized files.
+
+This preserves the no-fd-retained invariant exactly: a slot on the free-list
+holds no descriptor, and a slot backing a live handle holds no descriptor. It
+removes the unlink, the inode eviction, and the synchronous page-cache
+truncate — the entire 35.6s — for every spill that lands a recycled slot.
+Because pageout and take interleave during merges (take frees a slot, the next
+pageout immediately reuses it), steady-state reuse is high with a tiny cap.
+
+Reuse count being bounded by interleaving (not by the cap alone) is the same
+dynamic that makes `STASH_CAP = 2` sufficient for chunk recycling: during a
+phase that pages out N chunks before taking any, the free-list drains and the
+overflow falls back to fresh files; that long-tail keeps today's behavior.
+
+### Tier 2 — open-fd cache keyed by slot id
+
+Tier 1 still pays `File::open` on both pageout and take (4.3s on take in the
+profile; roughly the same again on pageout, so an ~8s ceiling). The fd cannot
+live in the handle: the spilled working set is thousands of cold chunks, each a
+live handle, and the no-fd-retained invariant exists precisely so fd count does
+not track handle count. The fd also cannot live only in the free-list slot,
+because it is needed at pageout *and* at the later take, with the cold in-use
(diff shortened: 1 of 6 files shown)


doc/developer/design/20260603_pager_pooling.md
```diff
@@ -0,0 +1,214 @@
...
+
+**Architecture (decided): thread-local free-list + cache.** Both the Tier 1
+free-list and the fd cache are made thread-local rather than process-global.
+This drops the hot-path mutex, gives each timely worker perfect locality over
+its own churn (a slot recycled by worker A is never popped by worker B, which
+would miss in B's cache), and matches the one-batcher-per-worker structure. The
+fd budget is then `N × workers`; `N` of 8–16 per worker covers the churn working
+set and stays well under the process ulimit. An open-vs-hit counter is exported
+as a metric — the only way to know in production whether `N` is sized right, and
+the hook tests use to assert reuse deterministically.
```
antiguru: Let's consider alternatives here and measure. The thread-local pool can degenerate if we have many threads, which can be true when using the pager from another runtime than Timely. Lgalloc solves this problem by using work-stealing, which allows the thread-local pool to refill from a global pool once it is exhausted, and it allows us to keep the local pool small.

Additionally, the file operations are amortized to once per 2MiB of data (roughly), so a mutex might be just fine.

antiguru: I see, the implementation deviated from the design!

src/ore/src/pager/file.rs
```diff
@@ -90,6 +90,69 @@ pub(crate) fn alloc_scratch_id() -> u64 {
     SCRATCH_ID.fetch_add(1, Ordering::Relaxed)
 }
 
+/// Free-list of scratch slots: named files that still exist on disk but back no
+/// live handle. `take`/drop returns a slot here instead of unlinking, so the
+/// inode and its warm page cache survive for the next `pageout` to overwrite.
+/// This avoids the synchronous inode eviction and page-cache truncate that
+/// dominate unlink cost for the write-once / read-once / delete spill pattern.
+static FREE_SLOTS: Mutex<Vec<u64>> = Mutex::new(Vec::new());
```
antiguru: To aid readability, introduce a newtype for the file ID.

src/ore/src/pager/file.rs
```diff
@@ -161,8 +240,20 @@ pub(crate) fn pageout_file(chunks: &mut [Vec<u64>]) -> Handle {
...
+fn write_chunks(path: &Path, chunks: &[Vec<u64>], reused: bool) -> std::io::Result<()> {
+    // Reused slots already have a file; open it and overwrite from offset 0.
+    // Deliberately no `truncate`: shrinking would re-evict the tail pages — the
+    // cost recycling exists to avoid — and reads are length-bounded, so a stale
+    // tail is never observed. Fresh slots use `create_new` to catch id-reuse
+    // bugs: a fresh id must never name an existing file.
+    let mut opts = File::options();
+    opts.write(true);
+    if reused {
+        opts.create(true);
```
antiguru, line 276: It is really confusing that `OpenOptions::create` has different semantics compared to `File::create`. But the implementation looks correct.

src/timely-util/src/columnar/batcher.rs
antiguru, line 1: Seems like an unrelated fix, extract to separate PR.

src/timely-util/src/columnar/merge_batcher.rs
```diff
@@ -630,7 +636,11 @@ pub fn extract_chain<D, T, R, SinkShip, SinkKeep>(
     T: Columnar + Default + Clone + PartialOrder,
     for<'a> columnar::Ref<'a, T>: Copy + Ord,
     R: Columnar + Default + Semigroup + for<'a> Semigroup<columnar::Ref<'a, R>>,
-    SinkShip: FnMut(PagedColumn<(D, T, R)>),
+    // The ship side is consumed by the seal builder immediately, so it's handed
+    // over resident rather than paged: round-tripping it through the pager would
+    // write the whole ship side to scratch and read it straight back. The keep
+    // side persists to the next round, so it stays paged.
+    SinkShip: FnMut(Column<(D, T, R)>),
```
antiguru: 🧊🚢

src/timely-util/src/column_pager.rs
antiguru, line 1: Seems this could be a separate PR, too!

antiguru (commented):
Looks good, left some inline comments. Main concern is that this should probably be more than one PR?

doc/developer/design/20260603_pager_pooling.md
```diff
@@ -0,0 +1,214 @@
...
+Correctness falls out cleanly: `O_RDWR` fds serve both pwrite and pread; recycle
+overwrites the *same inode* in place (pwrite at offset 0), so a cached fd never
+goes stale; fresh ids come from a monotonic counter and unlinked ids never
+reissue, so there is no inode-reuse-under-stale-fd hazard. The single
+invalidation rule: when Tier 1 unlinks a slot (oversize, cap overflow, or the
+write-error path), `remove` + close its cached fd first, or the open fd pins the
+inode and defeats the unlink. The cache incidentally also speeds the
+write-once / read-many `read_at_many` path, which re-opens per call today.
+
+**Architecture (decided): thread-local free-list + cache.** Both the Tier 1
```
antiguru: This says thread-local is decided for "both the Tier 1 free-list and the fd cache", but the Tier 1 implementation in this PR uses a process-global `Mutex` (and the "Concurrency" bullet below says process-global). The implementation is the right call: `Handle` is `Send`, so a handle can drop on a different thread than the one that paged it out — a thread-local free-list would strand slots on the dropping thread, away from the worker whose pageouts could reuse them. Worth updating this paragraph to match the implementation and record that reasoning, since it also constrains the Tier 2 fd cache (a cached fd for a slot recycled on another thread misses by construction).

doc/developer/design/20260603_pager_pooling.md
```diff
@@ -0,0 +1,214 @@
...
+## Success criteria
+
+* The common merge-batcher churn (pageout and take interleaved within a merge)
+  stops paying inode create + unlink + page-cache evict per spill.
+* Peak open file descriptors stay bounded by a small constant, independent of
+  how many handles are simultaneously paged out. The original design's core
+  invariant — fd count decoupled from handle count — is preserved.
+* The destination-buffer faulting cost is addressed for the steady state
+  without holding unbounded resident memory off the pager's budget.
+* No change to the public `mz_ore::pager` API or the handle-survives-backend-flip
```
antiguru: This criterion is no longer met as stated: the PR adds `pub fn Handle::is_file`, which leaks the backend variant through the public API so `ColumnPager` can route warm buffers. The addition seems fine, but the doc should say "no change to existing signatures" or similar, and maybe note why the routing lives in the caller rather than inside the pager (e.g. a pager-internal pool can't know the caller's recycling chokepoint).

src/ore/src/pager/file.rs
```diff
@@ -90,6 +90,69 @@ pub(crate) fn alloc_scratch_id() -> u64 {
...
+/// the high-water mark of payloads written to it (reuse overwrites from offset
+/// 0 without truncating); refusing to recycle past this bound keeps every
+/// free-list file — and thus retained page cache — under the threshold. Any
+/// larger write lands on a slot that is then unlinked on drop, so it never
+/// re-enters circulation. Mirrors the merge batcher's recycle threshold.
+const MAX_RECYCLE_BYTES: usize = 1 << 22;
+
+/// Pops a recycled slot id, or `None` if the free-list is empty. The returned
+/// id names an existing file at `scratch_path(id)` holding stale bytes that the
+/// caller is expected to overwrite.
```
antiguru: nit: "names an existing file" isn't always true — if `try_take_file` fails with ENOENT (file unlinked out from under us), the inner's drop still recycles the slot, so a free-listed id can name a missing file. Harmless because the reused-slot open uses `create(true)`, but the doc should say "usually existing" or call out the ENOENT case so the `create(true)` choice below stays explained.

src/ore/src/pager/file.rs
```diff
@@ -90,6 +90,69 @@ pub(crate) fn alloc_scratch_id() -> u64 {
...
+        .pop()
+}
+
+/// Returns slot `id` to the free-list, retaining its file for reuse. Returns
+/// `false` (and keeps nothing) if the list is already at [`FREE_LIST_CAP`], so
+/// the caller unlinks instead.
+fn push_free_slot(id: u64) -> bool {
+    let mut slots = FREE_SLOTS
+        .lock()
+        .expect("mz_ore::pager FREE_SLOTS poisoned");
```
antiguru: `push_free_slot` runs from `FileInner::drop`, so a poisoned `FREE_SLOTS` turns into a panic-in-drop → abort. The free-list can't be left in a bad state by a panic mid-push (worst case a slot id is lost), so recovering with `unwrap_or_else(|e| e.into_inner())` — as the test `SERIAL` lock already does — is strictly safer here.

src/ore/src/pager/file.rs
```diff
@@ -90,6 +90,69 @@ pub(crate) fn alloc_scratch_id() -> u64 {
...
+/// the cap mostly absorbs burst phase transitions. Tuning candidate.
+const FREE_LIST_CAP: usize = 16;
+
+/// Don't recycle a slot whose payload exceeds this. A slot's on-disk file is
+/// the high-water mark of payloads written to it (reuse overwrites from offset
+/// 0 without truncating); refusing to recycle past this bound keeps every
+/// free-list file — and thus retained page cache — under the threshold. Any
+/// larger write lands on a slot that is then unlinked on drop, so it never
+/// re-enters circulation. Mirrors the merge batcher's recycle threshold.
+const MAX_RECYCLE_BYTES: usize = 1 << 22;
```
antiguru, line 135: nit: `1 << 22` now exists in three places (here, `BODY_POOL_MAX_BYTES` in `column_pager.rs`, `MAX_RECYCLE_BYTES` in `merge_batcher.rs`), with comments saying they mirror each other but nothing enforcing it. Either share a constant or note explicitly that they may diverge independently.

For the record, I checked the invariant the comment claims and it holds: an oversize payload can still *pop* a free slot (no size check at pageout) and grow its file past the bound, but that handle's `len_u64s` then exceeds the threshold, so drop unlinks and the grown file never re-enters the free-list. Free-listed files stay ≤ `MAX_RECYCLE_BYTES`. Might be worth a sentence here, since the pop-before-size-check path is the non-obvious case.

src/timely-util/src/columnar.rs
```diff
@@ -171,7 +171,19 @@ where
 /// so this is the page count we round up to. Picked to match
 /// [`builder::ColumnBuilder`]'s output granularity so chunks shipped from the
 /// merger and chunks shipped from the builder are sized comparably.
-const SHIP_WORDS: usize = 1 << 18;
+pub(crate) const SHIP_WORDS: usize = 1 << 18;
```
antiguru: nit: the `pub(crate)` widening looks unnecessary — `RESERVE_WORD_BUDGET` is defined in this module, and `batcher.rs` only references `RESERVE_WORD_BUDGET` and `serialized_words`.

antiguru (commented):
Tier 1 and the buffer pool look correct; I verified the sharp edges. Oversize reuse can grow a free slot's file past the bound, but that handle then unlinks on drop, so free-listed files stay ≤ `MAX_RECYCLE_BYTES`. Empty handles return before the recycle branch and never pollute the list. `CompressedInner::Paged` is only ever built via `pageout_with(Backend::File, ...)`, so the unconditional warm buffer there is sound. The global mutex (vs the design doc's thread-local) is the right call since handles drop cross-thread.

Two asks on scoping/description, both about the last two commits, which the PR body doesn't mention:

* The `merge_from` pre-reserve change is an OOM fix ("this OOM'd a replica"), not a profile optimization. It deserves its own PR — separate revert/backport unit — or at least a section in the description.
* Dropping the seal ship-side round-trip changes observable behavior beyond speed: ship chunks no longer fire `observe_pageout`/`observe_pagein` or `PagingPolicy` events, ship output flips `Align` → `Typed`, and typed allocations now flow downstream instead of staying in the stash. Worth a paragraph.

One stale comment the diff can't anchor on: `ColumnPager::take`'s doc (column_pager.rs:490) still says "file backend unlinks"; it now recycles.

Remaining inline comments are doc-consistency nits in the design doc and `pager/file.rs`.

PR #36892, opened 2026-06-03
sql-227: Fix statement-logging throttling when sample rate is changed

SangJunBak:
In the frontend peek, we mutate mark a prepared statement as logged before the throttling check. Whileas in the original coordinator path, we wait until after. This causes a bug where if we initially throttled a statement, subsequent non-throttled executions will cause a prepared statement to never be logged in mz_prepared_statement_history
### Motivation

Fixes https://linear.app/materializeinc/review/sql-227-fix-statement-logging-throttling-when-sample-rate-is-changed-0624907ce490

### DIFF
--- a/src/environmentd/tests/server.rs
+++ b/src/environmentd/tests/server.rs
@@ -750,7 +750,7 @@ fn test_statement_logging_prepared_statement_throttling() {
 
 /// throttling the first execution of a prepared statement must not make
 //  every subsequent execution of that statement invisible in
-// `mz_recent_activity_log`.
+//  `mz_recent_activity_log`.
 #[mz_ore::test]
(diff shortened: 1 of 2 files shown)


src/environmentd/tests/server.rs
```diff
@@ -748,6 +748,112 @@ fn test_statement_logging_prepared_statement_throttling() {
     run_throttling_test(true);
 }
 
+/// throttling the first execution of a prepared statement must not make
+//  every subsequent execution of that statement invisible in
+// `mz_recent_activity_log`.
```
ggevay: nit: Doc comment formatting is inconsistent.

ggevay (approved):
Thank you, LGTM!

PR #36893, opened 2026-06-03
edu-45: refactor isolation levels page (part1 excludes vetting real-t…

kay-kim:
…ime recency)
https://preview.materialize.com/materialize/36893/reference/isolation-level/


doc/user/content/reference/isolation-level.md
```diff
@@ -0,0 +1,195 @@
...
+Serializable also guarantees that the result of concurrently executing
+transactions is equivalent to some serial execution of those transactions. A
+serial execution is one in which each transaction completes before the next one
+begins. However, Serializable does not guarantee
+[linearizability](https://jepsen.io/consistency/models/linearizable); that is,
+it does not guarantee that the serial order matches the real-time order of the
+transactions. For example, if transaction **T1** completes before transaction
+**T2** begins in real time, the result may be equivalent to a serial execution
+in which **T2** executes before **T1**.
+
```
kay-kim, line 51: I removed the section about consistent snapshot availability guarantee  when Materialize can determine that a transaction accesses only a single object ... because not sure about its utility.

Its utility seems limited to suggesting that if you're making single-statements, people should use auto-commit mode vs wrapping the single-statement in explicit transaction block  but, in practice, how likely is it that people are wrapping single-statements in an explicit transaction block?

doc/user/content/reference/isolation-level.md
```diff
@@ -0,0 +1,195 @@
...
+- However, once that data is included in the results of a read transaction, all
+  subsequent read transactions are guaranteed to see it.
+
+### Logical timestamp selection {#strict-serializable-logical-timestamp-selection}
+
+{{% include-headless "/headless/logical-timestamp-selection-strict-serializable"
+%}}
+
+### Real-time recency
```
kay-kim, line 93: I did not vet the real-time recency section.  That's for some far-off future.

def- (approved):

PR #36896, opened 2026-06-04
persist: match schema modulo nullability on the write path

antiguru:
### Motivation

Fixes the `compare_and_append_batch` panic in [CLU-105](https://linear.app/materializeinc/issue/CLU-105):

```
thread 'timely:work-3' panicked at src/persist-client/src/write.rs:745:21:
fetching a schema id should only fail when the shard is tombstoned
```

`WriteHandle::compare_and_append_batch` calls `try_register_schema()`; when that returns `None` it asserts the shard must be a tombstone. The assert fires because the shard is **live** yet registration returned `None`.

`StateCollections::register_schema` returns `None` in two cases: the shard is a tombstone (expected), **or** the shard already has a registered schema and the writer's schema does not exactly match it. The lookup compared `RelationDesc` for full structural equality, **including each column's `nullable` bit**. Persist encodes all top-level columns as nullable at the arrow/parquet level (database-issues#2488), so two schemas that differ only in column nullability are byte-compatible — but the exact-equality lookup treated them as different, returned `None` on a live shard, and tripped the assert.

This surfaced via #36848 (regex non-nullable-column inference): across a restart/upgrade an old build registers all-nullable columns while the new build re-derives some as non-nullable, panicking the persist sink (e.g. an MV using `regexp_extract`, or a `FORMAT REGEX` source).

The assertion itself predates the change that triggered it — it was added in Nov 2025; the regression range only surfaces the latent fragility.

### Changes

- `StateCollections::register_schema`: when the exact `RelationDesc` lookup fails, fall back to matching on the arrow `DataType` (which persist forces nullability- and position-invariant) and reuse the existing `SchemaId`.
- Adds a persist-client regression test with a null-invariant codec.

This is a **compatibility check, not a schema migration**: no new `SchemaId` is minted, nothing is written to `state.schemas`, no migration action is recorded, and no data is rewritten. It deliberately does not touch the `compare_and_evolve_schema` / `backward_compatible` machinery.

### Known limitation (deliberately out of scope)

This removes the panic but does not address the underlying **time-travel** hazard: #36848 tightens the logical desc (nullable → non-nullable) and propagates "non-null" to consumers, while retained history / `AS OF` reads can still surface NULLs the old build wrote. Persist storing everything arrow-nullable means persist never loses those NULLs — the discrepancy is at the catalog/optimizer level, and is the same class of issue replacement MVs and optimizer changes already cause. The durable fix (MVs/sources not propagating nullability/key info across the persist boundary, at a possible optimizer perf cost) is tracked separately.

### Tests

- New `register_schema_nullability_invariant` test in `src/persist-client/src/schema.rs` (passes).
- `cargo check -p mz-persist-client` passes.

https://claude.ai/code/session_01BMTv48z3QnXG4NC83zwVGt

---
_Generated by [Claude Code](https://claude.ai/code/session_01BMTv48z3QnXG4NC83zwVGt)_

### DIFF
--- a/src/persist-client/src/internal/state.rs
+++ b/src/persist-client/src/internal/state.rs
@@ -1589,6 +1589,28 @@ where
         let existing_id = self.schemas.iter().rev().find(|(_, x)| {
             K::decode_schema(&x.key) == *key_schema && V::decode_schema(&x.val) == *val_schema
         });
+
+        // If there's no exact match, fall back to matching on the arrow
+        // `DataType`. Persist encodes all top-level columns as nullable and
+        // identifies columns positionally (see database-issues#2488), so two
+        // schemas that produce the same `DataType` are byte-compatible and can
+        // safely share a `SchemaId`. In particular, schemas that differ only in
+        // column nullability are equivalent as far as persist is concerned.
+        //
+        // Without this, an otherwise-harmless nullability change -- e.g.
+        // improved non-null inference for a source's columns across an upgrade
+        // -- would fail to find its already-registered schema, return `None`,
+        // and trip the "fetching a schema id should only fail when the shard is
+        // tombstoned" assertion on the write path.
+        let existing_id = existing_id.or_else(|| {
+            let key_data_type = mz_persist_types::columnar::data_type::<K>(key_schema).ok()?;
+            let val_data_type = mz_persist_types::columnar::data_type::<V>(val_schema).ok()?;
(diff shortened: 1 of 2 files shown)


ggevay:
The [Checks 0dt upgrade across four versions 2](https://buildkite.com/materialize/nightly/builds/16694#019e91f4-2cb2-4456-a241-250d6a1c96f7) fail looks relevant.

PR #36898, opened 2026-06-04
compute: make correction buffer reads slice-proportional

antiguru:
### Motivation

MVs using the v2 correction buffer can stall during hydration when the input has to catch up through many distinct timestamps.
`CorrectionV2::updates_before` restructures the entire buffer on every call — Rc-wrapping every chunk, merging to the upper, rebuilding remainders, and re-merging chains for the chain invariant — so each read costs O(total buffered chunks) instead of O(drained slice), and a catch-up through T timestamps pays O(T²/chunk_capacity).
A micro-benchmark reproduces this: at 65536 timestamps (one day of 1s ticks at 16 updates/tick) a stepwise drain takes 13.5s per worker against CorrectionV1's 0.47s, with the ratio growing in T (~T^1.8).

### Description

Restructures `CorrectionV2` so reads only do work proportional to the drained slice, keeping the public interface, metrics, introspection logging, and dyncfgs unchanged.
`CorrectionV1` remains the fallback via `enable_correction_v2`. Rebased on top of #36577, so chunks use the columnar chunk region storage.

* Times at or beyond the largest read upper live in a `BucketChain` of exponentially growing time-range buckets, each holding (time, data)-sorted chains under the chain invariant. Reads peel only the buckets below their upper; far-future updates (e.g. temporal-filter retractions) are never touched.
* Chains split at the upper at chunk granularity (`Chain::split_at_time`), reusing whole chunks and copying at most one straddling chunk per chain.
* Emitted-but-not-yet-cancelled updates live in a designated `emitted` chain outside any invariant, so reads never re-merge future chains. (We also evaluated a destructive `take_before`/`insert_batch` interface; it measured CPU-neutral because feedback cancellation forces the same merge either way, so the interface stays as-is.)
* Times below the since are advanced with a hybrid strategy, decided by a bounded count of distinct stale times: few stale times — the steady state, since the previously emitted chain was written just before the since advanced past it — use the existing `Cursor::advance_by` run-splitting; many stale times (a since jump, e.g. a sink restarting with an old as-of) are collapsed in one sort-and-consolidate pass. The always-bulk variant regressed the ReplicaExpiration feature benchmark by 31%; with the hybrid it is 3.3% faster than the merge base.
* `BucketChain` exposes bucket iterators for metrics reporting, and `restore` is invoked with bounded fuel per buffer operation — incomplete restoration is picked up by the next operation, so reads never stall the operator. Builds on the `restore` fast path from #36897.

Benchmark results (stepwise drain with persist feedback, 16 updates per timestamp, medians, on top of #36577's columnar chunks):

| T | v1 | v2 before | v2 after |
|---:|---:|---:|---:|
| 1024 | 5.2 ms | 32 ms | ~16 ms |
| 4096 | 21 ms | 104 ms | ~64 ms |
| 16384 | 85 ms | 858 ms | 270 ms |

Scaling is now linear in T (was ~T^1.8); the since-jump scenario improves ~3× and inserts are unchanged. ("v2 before" measured pre-#36577; the columnar chunk storage adds a constant on this microbenchmark but does not change the asymptotics.)

### Verification

New micro-benchmark `cargo bench -p mz-compute --features bench --bench correction` covering insert, stepwise drain with feedback, since jumps, and a temporal-filter pattern; the `bench` feature only gates `pub` visibility.
New tests assert emission equivalence with `CorrectionV1` under a stepwise-drain-with-feedback workload, since-jump collapse onto the since, and that reads never observe times at or beyond their upper.
The ReplicaExpiration feature benchmark scenario was verified locally against the merge base (3.3% faster, less memory).

### DIFF
--- a/src/compute/src/sink/correction_v2.rs
+++ b/src/compute/src/sink/correction_v2.rs
@@ -28,9 +28,22 @@
 //! Stored updates are of the form `(data, time, diff)`, where `time` and `diff` are fixed to
 //! [`mz_repr::Timestamp`] and [`mz_repr::Diff`], respectively.
 //!
-//! [`CorrectionV2`] holds onto a list of [`Chain`]s containing [`Chunk`]s of stashed updates. Each
-//! [`Chunk`] is a columnation region containing a fixed maximum number of updates. All updates in
+//! [`CorrectionV2`] holds onto a list of `Chain`s containing `Chunk`s of stashed updates. Each
+//! `Chunk` is a columnation region containing a fixed maximum number of updates. All updates in
 //! a chunk, and all updates in a chain, are ordered by (time, data) and consolidated.
+//!
+//! Chains live in three places:
+//!
+//!  * A [`BucketChain`] partitions times at or beyond the `boundary` (the largest read `upper`
+//!    seen so far) into buckets of exponentially growing time ranges, each holding a list of
+//!    chains. Reads only touch the buckets below their `upper`, so the bulk of the buffered
+//!    updates — in particular far-future retractions produced by temporal filters — is left
+//!    alone.
+//!  * `pending_low` holds chains at times below the `boundary`, mostly insertions arriving
+//!    through the persist feedback.
+//!  * `emitted` is a single chain holding the updates returned by the last read. Updates must
+//!    stay in the buffer until their feedback retractions arrive, and keeping them separate from
+//!    the bucket chain means reads never have to re-merge future updates.
 //!
 //! ```text
 //!       chain[0]   |   chain[1]   |   chain[2]
@@ -49,13 +62,13 @@
 //!       (c, 3, +1) |              |
 //! ```
 //!
-//! The "chain invariant" states that each chain has at least `chain_proportionality` times as
+//! The "chain invariant" states that each chain in a bucket has at least `chain_proportionality` times as
 //! many updates as the next one. This means that chain sizes will often be powers of
 //! `chain_proportionality`, but they don't have to be. For example, for a proportionality of 2,
 //! the chain sizes `[11, 5, 2, 1]` would satisfy the chain invariant.
 //!
 //! Note that the invariant is maintained on update counts, not chunk counts. Chunks are
-//! byte-bounded (see [`ChunkBuilder`]), so chunk count is not proportional to update count and
+//! byte-bounded (see `ChunkBuilder`), so chunk count is not proportional to update count and
 //! would be a poor proxy: any chain below the chunk byte boundary is a single chunk regardless
 //! of how many updates it holds, which would let the geometric invariant collapse and break the
 //! O(log N) amortization of inserts.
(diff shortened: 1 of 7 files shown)


antiguru (commented):
Took a careful pass over the rewrite — the design (BucketChain + dedicated `emitted` chain, peel-below-upper reads, hybrid stale-time strategy) is sound, and I couldn't find correctness issues. The module docs and the differential equivalence-against-V1 test are excellent. A few non-blocking observations/questions:

**1. `updates_before` no longer short-circuits when `upper <= since`.** V1 had an early `if !PartialOrder::less_than(&self.since, upper) { return empty }`. The new version always runs `consolidate_before`, so a read with `upper <= since` eagerly peels buckets, merges, advances `boundary`, and populates `emitted` — only to have the `take_while` return nothing. I convinced myself it's correct (inserts clamp times to `since`, so nothing can later appear below it, and the time-5/since-10 case is covered by `upper_not_beyond_since`), but it's eager work + state mutation where the old path was a no-op. Is dropping the guard intentional? A one-line comment noting that normal reads / `consolidate_at_since` always have `upper > since` would help.

**2. `update_metrics` walks every bucket and chain on each insert and read.** It's bounded — roughly O(#buckets × chains/bucket) ≈ O(log(time-range) × log N), so it does *not* reintroduce the O(total) cost this PR removes — but it's a non-trivial per-op constant now that the rest of the read path is slice-proportional. Flagging in case it shows up in the insert microbenchmark; otherwise fine as-is.

**3. `bench` feature widens `sink` / `correction` / `correction_v2` to `pub`.** Gated behind a non-default feature used only by the bench, so acceptable — just want to confirm nothing builds `mz-compute` with `--features bench` in a production/release configuration, since it exposes these internals publicly.

**4. Tuning constants.** `MAX_STALE_RUNS = 32` is well-explained. `RESTORE_FUEL = 1_000_000` is generous enough that restoration effectively always completes in one shot — a sentence on how that value was picked (and what a realistic per-op chunk count looks like) would help future tuners reason about the "incomplete restoration is picked up next op" path actually getting exercised.

None of these block; #1 is the only one I'd like a quick confirmation on. Nice work.

---

antiguru:
Thanks for the careful pass. Addressed in the latest push (`57e7931d5d`, also rebased on current main):

**1. `upper <= since` short-circuit.** Confirmed the intent and restored the guard rather than just commenting. All buffered times are advanced to at least `since`, so a read with `upper <= since` can only yield times `>= since >= upper` — i.e. nothing — which is exactly what the `take_while` produced, just after eagerly peeling/merging/advancing `boundary`. The guard returns empty up front and matches V1's no-op. Normal reads and `consolidate_at_since` (which reads at `since + 1`) always pass `upper > since`, so the guard never short-circuits a non-empty read. Comment added explaining this.

**2. `update_metrics` bucket/chain walk.** Agreed it's bounded (O(#buckets × chains/bucket) ≈ O(log(range) × log N)) and doesn't reintroduce the O(total) cost. Left as-is; it didn't show up in the insert microbenchmark (inserts are ~unchanged vs the pre-rewrite v2). If it ever matters we can maintain a running total incrementally, but that's premature now.

**3. `bench` feature.** Confirmed — the only consumer is `[[bench]] correction` with `required-features = ["bench"]`; nothing builds `mz-compute` with `--features bench` in any release/production path (it's not in the default features and CI only enables it for the bench step). The widened visibility is compiled out of every shipped artifact.

**4. `RESTORE_FUEL`.** Expanded the doc: `restore` spends one unit per bucket split and a single `peel` leaves at most `BucketTimestamp::DOMAIN` (64) buckets to re-split, so the budget completes restoration in one call for any realistic buffer. It's deliberately generous — the "picked up next op" path is a correctness safety net for pathological bucket counts, not a hot path — and the comment now says to lower it if restoration ever needs to interleave with other work.

Also applied your inline suggestion to bind the `take_while` iterator to a `let` before wrapping it.

src/compute/src/sink/correction_v2.rs
```diff
@@ -254,137 +301,236 @@ impl<D: Data> CorrectionV2<D> {
...
         }
 
         self.consolidate_before(upper);
 
-        // There is at most one chain that contains updates before `upper` now.
-        result = self
-            .chains
+        // `emitted` can contain times at or beyond `upper`: when `upper` is not beyond the
+        // `since`, stale times below `upper` advance to the `since`. Yield only times before
+        // `upper`; `emitted` is time-sorted, so `take_while` suffices.
```
DAlperin: I think this is a good check but technically unreachable. We early return right above

DAlperin (approved):
Very cool. 1 nit LGTM

antiguru:
@DAlperin good catch — you're right that the `take_while` is unreachable given the `upper > since` guard added right above. Traced every path that writes `emitted` in `consolidate_before`: all split at `upper` (remainder → `pending_low`), and with `upper > since` advancing stale times to `since` can't lift them to/beyond `upper`, so `emitted` is always entirely `< upper` after consolidation.

Replaced the dead filter (`440a6740cb`): yield all of `emitted` directly, and encode the invariant as a `soft_assert_or_log!` (matching the idiom in `render/sinks.rs`) rather than a hard assert — a violation would mean writing updates beyond the batch upper to persist, so it's worth catching in CI/staging and logging in prod, but not worth crashing a replica over. Comment explains why the invariant holds.

PR #36899, opened 2026-06-04
compute: always export memory limiter metrics, gate enforcement

antiguru:
### Motivation

The `mz_memory_limiter_%` metrics are absent in self-managed: the limiter only started when clusterd received `--heap-limit`, which the controller only passes for swap-enabled sizes with a nonzero disk limit. The self-managed helm chart forces `disk_limit=0` (the unlimited-swap special case), so the limiter never ran there and its metrics were never registered.

Investigation in https://linear.app/materializeinc/issue/CLU-102.

### Description

Decouples observation from enforcement:

* The limiter task always starts, so the usage gauges (`vm_rss`, `vm_swap`, `memory_usage`) are exported even without an announced heap limit.
* Process termination (exit 167) is gated behind a known limit and a new `memory_limiter_enforce` dyncfg (default `true`), so cloud behavior is unchanged and self-managed runs in observe-only mode.
* `mz_memory_limiter_memory_limit_bytes` reports 0 when no limit is announced.
* `memory_limiter_interval = 0` remains the full-disable switch.

### Verification

New unit tests cover that observe-only mode (no limit) and `enforce = false` never terminate the process while still updating the usage metrics; existing limiter tests updated for the new config shape.

### DIFF
--- a/misc/python/materialize/mzcompose/__init__.py
+++ b/misc/python/materialize/mzcompose/__init__.py
@@ -510,6 +510,7 @@ UNINTERESTING_SYSTEM_PARAMETERS = [
     "memory_limiter_usage_bias",
     "memory_limiter_burst_factor",
     "catalog_info_metrics_reconcile_interval",
+    "memory_limiter_enforce",
     "compute_server_maintenance_interval",
     "compute_dataflow_max_inflight_bytes_cc",
     "compute_flat_map_fuel",
--- a/src/compute-types/src/dyncfgs.rs
+++ b/src/compute-types/src/dyncfgs.rs
@@ -238,6 +238,14 @@ pub const MEMORY_LIMITER_BURST_FACTOR: Config<f64> = Config::new(
     "Multiplicative burst factor to the memory limiter's limit.",
 );
 
+/// Whether the memory limiter enforces the memory limit by terminating the process. When
+/// disabled, the limiter only observes memory usage and exports metrics.
+pub const MEMORY_LIMITER_ENFORCE: Config<bool> = Config::new(
+    "memory_limiter_enforce",
+    true,
+    "Whether the memory limiter enforces the memory limit by terminating the process.",
+);
+
 /// Enable lgalloc for columnation.
 pub const ENABLE_COLUMNATION_LGALLOC: Config<bool> = Config::new(
     "enable_columnation_lgalloc",
@@ -506,6 +514,7 @@ pub fn all_dyncfgs(configs: ConfigSet) -> ConfigSet {
         .add(&MEMORY_LIMITER_INTERVAL)
         .add(&MEMORY_LIMITER_USAGE_BIAS)
         .add(&MEMORY_LIMITER_BURST_FACTOR)
+        .add(&MEMORY_LIMITER_ENFORCE)
         .add(&ENABLE_LGALLOC_EAGER_RECLAMATION)
         .add(&ENABLE_COLUMNATION_LGALLOC)
         .add(&COMPUTE_SERVER_MAINTENANCE_INTERVAL)
--- a/src/compute/src/memory_limiter.rs
+++ b/src/compute/src/memory_limiter.rs
@@ -156,6 +156,11 @@ struct LimiterTask {
     config: LimiterConfig,
     /// The amount of burst budget remaining.
     burst_budget_remaining: usize,
+    /// Whether we have already logged that the limit was exceeded.
+    ///
+    /// Used to log only on the transition to over-limit, rather than at every check interval
+    /// while enforcement is disabled.
+    limit_exceeded_logged: bool,
     /// The time of the last check.
     last_check: Instant,
     /// Metrics tracked by the limiter task.
@@ -169,6 +174,7 @@ impl LimiterTask {
         let mut task = Self {
             config: LimiterConfig::disabled(),
             burst_budget_remaining: 0,
+            limit_exceeded_logged: false,
             last_check: Instant::now(),
             metrics,
         };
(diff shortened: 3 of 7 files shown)


DAlperin (commented):
One question around if we really need the flag, but otherwise looking good

src/compute-types/src/dyncfgs.rs
```diff
@@ -238,6 +238,14 @@ pub const MEMORY_LIMITER_BURST_FACTOR: Config<f64> = Config::new(
...
 );
 
+/// Whether the memory limiter enforces the memory limit by terminating the process. When
+/// disabled, the limiter only observes memory usage and exports metrics.
+pub const MEMORY_LIMITER_ENFORCE: Config<bool> = Config::new(
+    "memory_limiter_enforce",
+    true,
+    "Whether the memory limiter enforces the memory limit by terminating the process.",
+);
+
```
DAlperin, line 248: It's not immediately obvious to me when someone would use this flag especially if it defaults to true. If they wanted to announce a memory limit but not enforce it? 0 seems like a fine sentinel.

antiguru: The flag is meant as a runtime kill-switch for enforcement, which the limit sentinel can't provide.

The announced limit is a deploy-time, per-size property: a replica gets `--heap-limit` only for swap-enabled sizes with a nonzero `disk_limit`, and changing that means reprovisioning the replica. The `enforce` dyncfg lets us stop process termination at runtime, with no reprovision, while still sampling and exporting the usage metrics. That's the intended use: if the limiter starts terminating replicas incorrectly in cloud, we flip one flag to fall back to observe-only instantly, and we keep the `memory_limit_bytes` signal we'd want during the incident.

It's also distinct from the existing `memory_limiter_interval = 0`, which disables sampling entirely and so drops the metrics too.

Defaulting to `true` keeps cloud behavior unchanged; the flag only matters when you want a known limit reported but not acted on.

misc/python/materialize/mzcompose/__init__.py
```diff
@@ -510,6 +510,7 @@ def get_default_system_parameters(
     "memory_limiter_usage_bias",
     "memory_limiter_burst_factor",
     "catalog_info_metrics_reconcile_interval",
+    "memory_limiter_enforce",
```
bosconi: nit: `action.py` adds the same flag directly after `memory_limiter_burst_factor`; this list splits it from the other `memory_limiter_*` entries, which makes the two lists harder to diff against each other when a limiter flag is added or removed.

```suggestion
    "memory_limiter_enforce",
    "catalog_info_metrics_reconcile_interval",
```

src/compute/src/memory_limiter.rs
```diff
@@ -382,6 +418,61 @@ mod tests {
...
+    #[cfg_attr(miri, ignore)] // reads /proc/self/status
+    fn check_without_enforce_does_not_terminate() {
+        let mut task = task_for_test();
+        // A zero limit guarantees the current usage exceeds it.
+        task.apply_config(LimiterConfig {
+            interval: Duration::from_secs(1),
+            memory_limit: Some(0),
+            burst_budget: 0,
+            enforce: false,
+        });
```
bosconi: This test reaches the exhausted-budget branch only if `excess_bs` comes out greater than zero, and `excess_bs` is `usize::cast_lossy(memory_usage as f64 * last_check.elapsed())`. `apply_config` sets `last_check` to now, so `elapsed` here is just the wall-clock time to run one `check`. If that product truncates to 0, then `burst_budget_remaining >= excess_bs` holds with both sides at 0, the subtract branch is taken instead, `limit_exceeded_logged` stays false, and the assertion below fails. The margin depends on the test process RSS and on machine speed, neither of which the test controls.

Backdating `last_check` makes the branch deterministic:

```suggestion
        });
        // Backdate `last_check` so the excess byte-seconds charged by the next
        // `check` exceed the zero burst budget regardless of machine speed.
        task.last_check = Instant::now() - Duration::from_secs(1);
```

src/compute/src/memory_limiter.rs
```diff
@@ -232,19 +258,26 @@ impl LimiterTask {
...
+                // Burst budget exhausted.
+                self.burst_budget_remaining = 0;
+                if !self.limit_exceeded_logged {
+                    warn!(
+                        memory_usage,
+                        memory_limit, "memory utilization exceeded configured limits",
+                    );
+                    self.limit_exceeded_logged = true;
+                }
+                if self.config.enforce {
```
bosconi: With `enforce` false there is no exported signal that the limiter would have terminated the replica. The only evidence is the `warn!` above, which is now emitted once per over-limit episode, and `mz_memory_limiter_burst_budget_byteseconds` sitting at 0, which reads the same as a replica under its limit with `memory_limiter_burst_factor = 0`.

The stated purpose of the flag is to keep the limiter signal while enforcement is off during an incident, and a single log line is hard to alert on and is lost once it falls out of the log retention window. Consider incrementing a counter here, for example `mz_memory_limiter_limit_exceeded_total`, before the `enforce` branch, so observe-only mode is visible in the metrics the rest of this PR is about.

test/cluster/mzcompose.py
```diff
@@ -6037,6 +6037,35 @@ def setup_workload():
...
+            ALTER SYSTEM SET memory_limiter_burst_factor = 0;
+            ALTER SYSTEM SET memory_limiter_enforce = false;
+            """,
+            port=6877,
+            user="mz_system",
+        )
+        setup_workload()
+        c.up("clusterd1")
+
+        c.testdrive("> SELECT count(*) FROM mv\n1000000")
```
bosconi: Test 4 asserts only that the MV hydrates. It infers from Test 2 that the workload exceeds a 205 MiB limit, but it does not observe that in its own run, so if the workload's memory profile ever changes it will pass for the wrong reason and stop covering the enforcement gate.

Asserting that the limiter logged the over-limit condition during this run makes the test self-contained. Test 2 emits the same message, and the container logs survive the kill and restart, so compare counts rather than presence:

```suggestion
        # Test 2 logged the same message before exiting and the container logs
        # survive the restart, so compare occurrence counts rather than presence.
        exceeded = "memory utilization exceeded configured limits"
        before = c.invoke("logs", "clusterd1", capture=True).stdout.count(exceeded)

        setup_workload()
        c.up("clusterd1")

        c.testdrive("> SELECT count(*) FROM mv\n1000000")

        after = c.invoke("logs", "clusterd1", capture=True).stdout.count(exceeded)
        assert after > before, "limiter never observed the limit being exceeded"
```

PR #36903, opened 2026-06-04
Reapply iterative `Visitor` refactor (#36852) with SQL-332 fix

antiguru:
### Motivation

The iterative, child-based `Visitor` refactor (#36852) was reverted in #36905 because it introduced a regression: SQL-332, where `mz_now()` used as a direct argument to a value window function (e.g. `first_value`, `last_value`) or an aggregate window function (e.g. `max`) was no longer recognized as temporal, so the query failed to receive a timestamp near the current time. For example, `SELECT first_value(mz_now()) OVER () < '3000-01-01'` returned `false` instead of `true`.

This PR re-lands the refactor with that regression fixed, so it doesn't ship again.

### Description

This PR consists of two commits:

1. **Reapply "refactor: Change all `Visitor`s to be iterative, child-based" (#36852)** — reverts the revert (#36905), restoring the iterative traversal.

2. **Fix temporal detection (SQL-332)** — the root cause of the regression. The iterative `Visit` traversals walk expressions via the new `VisitChildren::children()` / `children_mut()` iterators. For `ValueWindowExpr` and `AggregateWindowExpr`, the new `children()` impls descended *into* the argument expression and yielded *its* children, rather than yielding the argument expression itself — which is what the canonical `visit_children` does, and what the impl-level doc comments promise (*"Yields `args`; does not descend into it"*).

   `contains_temporal()` runs over this traversal (via `visit_post`). When the argument is a leaf like `mz_now()` (a `CallUnmaterializable` with no children), descending skipped it entirely, so the `mz_now()` node was never visited and the query was planned as a constant.

   The fix yields the argument expression itself via `std::iter::once(...)`, matching `visit_children`.

### Verification

Green CI. Added regression tests in `test/sqllogictest/temporal.slt` covering `mz_now()` as the direct argument of `first_value`, `last_value`, and `max` window functions — each asserts the value is `< '3000-01-01'`, confirming the query receives a current timestamp rather than `u64::MAX`.

https://claude.ai/code/session_017Xe6H3p7ZETe4LW5bqPiWZ

### DIFF
--- a/src/expr/src/visit.rs
+++ b/src/expr/src/visit.rs
@@ -195,7 +195,7 @@ pub trait Visit {
     /// Optionally, `pre` can return which children, if any, should be visited
     /// (default is to visit all children).
     ///
-    /// It is improtant for safety that `pre` is (a) safe code and (b) returns children only.
+    /// It is important for safety that `pre` is (a) safe code and (b) returns children only.
     fn visit_mut_pre_post<F1, F2>(&mut self, pre: &mut F1, post: &mut F2)
     where
         F1: FnMut(&mut Self) -> Option<Vec<&mut Self>>,
--- a/src/expr/src/visit.rs
+++ b/src/expr/src/visit.rs
@@ -11,7 +11,7 @@
 //!
 //! Recursive types can implement the [`VisitChildren`] trait, to
 //! specify how their recursive entries can be accessed. The extension
-//! trait [`Visit`] then adds support for recursively traversing
+//! trait [`Visit`] then adds support for iteratively traversing
 //! instances of those types.
 //!
 //! # Naming
@@ -34,12 +34,6 @@
 //!   * no suffix: recursively visit children in pre- and post-order
 //!     using a ~Visitor~` that encapsulates the shared context.
 
-use std::marker::PhantomData;
-
-use mz_ore::stack::{CheckedRecursion, RecursionGuard, RecursionLimitError, maybe_grow};
-
-use crate::RECURSION_LIMIT;
-
 /// A trait for types that can visit their direct children of type `T`.
 ///
 /// Implementing [`VisitChildren<Self>`] automatically also implements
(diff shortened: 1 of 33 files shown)


src/expr/src/visit.rs
```diff
@@ -233,554 +194,297 @@ pub trait Visit {
...
-    /// Does not enforce a recursion limit.
-    ///
-    /// The function `pre` runs on `self` before it runs on any of the children.
-    /// The function `post` runs on children first before the parent.
     ///
-    /// Optionally, `pre` can return which children, if any, should be visited
-    /// (default is to visit all children).
-    #[deprecated = "Use `visit_mut_pre_post` instead."]
-    fn visit_mut_pre_post_nolimit<F1, F2>(&mut self, pre: &mut F1, post: &mut F2)
+    /// It is improtant for safety that `pre` is (a) safe code and (b) returns children only.
```
ggevay: Typo in "improtant".

ggevay:
Triggered a Nightly subset: https://buildkite.com/materialize/nightly/builds/16708

src/sql/src/plan/hir.rs
```diff
@@ -708,18 +747,33 @@ impl VisitChildren<HirScalarExpr> for ValueWindowExpr {
...
     }
 
     fn try_visit_mut_children<F, E>(&mut self, mut f: F) -> Result<(), E>
     where
         F: FnMut(&mut HirScalarExpr) -> Result<(), E>,
-        E: From<RecursionLimitError>,
     {
         f(&mut self.args)
     }
+
```
ggevay, line 736: Could the above `visit_children`, `visit_mut_children`, `try_visit_children`, `try_visit_mut_children` be deleted? Their trait defaults call `children`/`children_mut`, which should have the same behavior.

And the same applies for `impl VisitChildren<HirScalarExpr> for AggregateWindowExpr {`.

(But this doesn't apply to other things, e.g., `impl VisitChildren<HirScalarExpr> for WindowExprType {`, because there the explicit implementations are saving Vec allocations.)

ggevay (approved):
LGTM if Nightly is green, just minor comments.

antiguru:
Thanks for the review! Let's try again.

PR #36908, opened 2026-06-04
sql: remove refresh strategies and scheduled clusters from public docs

sjwiesman:
## Motivation

Materialized-view refresh strategies (`REFRESH ON COMMIT` / `AT` / `EVERY`) and the scheduled-cluster feature (`SCHEDULE = ON REFRESH`, `HYDRATION TIME ESTIMATE`) should no longer be surfaced in the public documentation. The feature remains private preview; this PR moves all of its documentation onto a single **unlisted, non-indexed** page that is only reachable via a direct link.


val-materialize (approved):
LGTM - Agree we don't need to wait on GM to merge

PR #36909, opened 2026-06-04
design: recorders — recorded, non-deterministic, at-least-once transformations

antiguru:
## What

A design doc proposing **recorders**: a way to house transformations that must **record a decision made at processing time and never recompute it** — non-deterministic, side-effecting, or time-dependent logic (including UDFs). The guarantee is **at-least-once application** of the function and **at-most-once persistence** of its result (committed once via a per-commit CAS).

Write-side complement to the `CHANGES` table function (#36869, draft); a deliberate revival of the removed Continual Tasks feature (database-issues#9694 / PR #35967), pulled for *lack of design consensus and incompleteness*, **not** because the model is unsound.

Docs: `doc/developer/design/20260604_recorders.md` (conceptual) and `20260604_recorders_implementation.md` (feasibility). The model matches the WG "Recorders" design (Notion); these docs are the long-form version.

## The calculus

- a **TVC** — a normal collection; change implicit.
- a **dTVC** — a changelog carrying its change *as data*: a **time** and a **diff** column (ordinary, user-named).

`differentiate` (`CHANGES`) and `integrate` (`INTEGRATE`) are a **carrier-preserving inverse pair**, both in the input's timeline (domain A). The cross-domain, non-deterministic leg — picking the commit time `T`, sampling frozen values like `now()` — lives **only in `RECORD`**. `freeze` is `sample-and-hold`.

## The surface (no new collection kind)

```sql
-- a regular table that declares its carriers + an associated progress collection.
CREATE TABLE enriched (
  a BIGINT, b BIGINT, c BIGINT, val TEXT,
  change_ts mz_timestamp NOT NULL, change_diff bigint NOT NULL
) WITH (PROGRESS, TIMESTAMP = change_ts, DIFF = change_diff);   -- inits progress to (0,0)

-- the one new standing object; bare query like an MV, INTO like a SINK.
CREATE RECORDER enrich INTO enriched AS
  SELECT e.*, d.val
  FROM CHANGES(events AS OF AT LEAST mz_now() - INTERVAL '1 hour'
               USING TIME change_ts, DIFF change_diff) e
  JOIN dim d ON e.fk = d.key;            -- bare TVC ref ⇒ frozen lookup

-- argument-free: carriers + progress come from the table.
CREATE MATERIALIZED VIEW enriched_now AS SELECT * FROM INTEGRATE(enriched);

-- bounding is ordinary DELETE; integral-preserving = data-domain compaction.
```

- **The store is a regular table** that declares its carriers + an associated **progress collection** via `WITH (PROGRESS, …)` (initialized to `(0,0)`; empty progress = sealed frontier). No `DELTA TABLE` collection kind.
- **`change_diff` is data, not the persist multiplicity.** `CHANGES` emits each change as a `+1` row with the signed count in the column.
- **`INTEGRATE(table)` is argument-free** — carriers and the progress collection come from the table; it **accumulates `DIFF` per row, thresholding at `max(0, Σ)`** internally (safe by construction; `repeat_row` not exposed). A stateful reduce, memory ∝ live output.
- **The progress collection is owned by the *table***, engine-written / user-read-only (the `CREATE SOURCE … EXPOSE PROGRESS AS` precedent); `INTEGRATE` *reclocks* through it. A `RECORD` writer only writes a *lane*. Taking a table (not arbitrary SQL) is what keeps the progress lookup unambiguous.
- **Application shaping** — dedup, first-seen, upsert, top-k — lives in the **`RECORD` body**, not in `INTEGRATE`; first-seen/upsert bodies are self-referential (Tier 3).
- **Bounding is ordinary `DELETE`**; the integral-preserving form is **data-domain compaction** — a `clamp + GROUP BY/SUM` *reduce*, not a bare `UPDATE`.

## Decisions worth reviewers' attention

- **Per-table progress collection + per-writer lanes.** `B` is the table's single shard frontier ⇒ exactly one progress collection per table; all `RECORD`s and `INTEGRATE`s share it (multiple readers consistent by construction). Multi-writer = per-writer lanes `R_i: B → X_i`, table A-completeness `= meet_i R_i[B_upper]` at read time (a merged frontier can't work). Drop the last writer → **freeze or seal** (open).
- **Optimizer barrier (hard invariant):** the recorded table is authoritative; never recomputed from the `RECORD` body's inputs.
- **Freeze is typing, not a keyword** — bare TVC ref in a `RECORD` body = frozen; must be **diagnosable, never silent** (`EXPLAIN` + plan-time `NOTICE`). In a body, `now()` (frozen wall-clock) ≠ `mz_now()` (domain-A event time).
- **Aging domain decided:** retention defaults to **wall-clock (domain B)**, event-age (A) opt-in. **Where `RETAIN HISTORY` sits** (table vs `INTEGRATE` consumer vs recorder — leaning `INTEGRATE`) is open.
- **GDPR erasure** = `DELETE` + advancing `since`; a cascade `DELETE` alone is not erasure.

## SQL-consistency pass

An agent reviewed the surface against Materialize's grammar (file:line) and prior art:
- Carriers via a `USING` clause / table `WITH (…)`, not `=>` named args (MZ has no named-arg support; `=>` already means map-entry). `CHANGES`'s `AS OF` parses like `SUBSCRIBE (query) AS OF …`. Dropped the `AS RECORD(...)` wrapper; `IN DOMAIN` → `WITH (TIMELINE = …)`.
- **Deliberate divergences:** `DIFF` is a signed multiplicity, not an op-enum (cf. Snowflake `METADATA$ACTION`, Flink `+I/-U`, Debezium `c,u,d`); `CHANGES` knowingly collides with Snowflake's `CHANGES` clause.

## Status

**Draft, for discussion.** De-risked by the `CHANGES` PR (#36869) and the `BEGIN CONTINUAL TRANSACTION` prototype. Gating dependency: the (unbuilt) OCC timestamped-write substrate — `INTEGRATE` and bounding DML are off that critical path. Collapsing to a regular table + progress collection removes the speculative `DELTA TABLE` kind (impl M2) and reshapes engine-owned compaction (M3/M5). Open agenda: co-design the `CHANGES` carrier spelling with #36869, the `RECORDER` keyword, commit-timestamp policy, `RETAIN HISTORY` placement, freeze-vs-seal on drop-all.

Stakeholders: @antiguru + the WG Continual Tasks folks (Aljoscha, Seth, Frank).

https://claude.ai/code/session_015YFH7J7PaEqkBSrAQYaq8H


antiguru (commented):
## Review — *Recorders* design doc

Strong, unusually well-grounded design. The conceptual framing is the best part; my substantive concerns are concentrated in **freeze-by-typing ergonomics**, the **multi-writer domain-A frontier**, and one **success-criterion vs. mechanism mismatch** on compliance erasure. None are blockers for a discussion draft — they're the agenda.

### Strengths

- **The calculus reduces surface area honestly.** Framing stream-table join, upsert, finalization, dedup, and compliance as *compositions* of four operations — rather than bespoke features — is the right answer to exactly why CTs bloated and stalled.
- **Intellectual honesty about feasibility.** H1 states the gating OCC commit substrate is unbuilt ("zero hits in the tree") — confirmed. H2 admits the relaxed evaluation rule removes the *fixpoint* but **not** the lagged self-read machinery.
- **Code references are accurate.** Line numbers check out (`group_commit` at `appends.rs:344`, `append_table` at `storage-controller/src/lib.rs:2082`); the CT as-of special-casing the doc wants to reuse is genuinely still live in `compute-client/src/as_of_selection.rs`. The `continual_task.rs` render module is gone but the doc correctly notes it's recoverable from base commit `add050bf8`. The salvage list is credible.
- **The z-transform duality** (`D = 1−z⁻¹`, `I = D⁻¹`, freeze = sample-and-hold) is a genuinely clarifying mental model, and makes "`INTEGRATE` carries no non-determinism, hence no new object kind" obvious.

### Substantive concerns / questions

**1. Freeze-by-typing is the sharpest usability footgun — strengthen the diagnostics story.**
The same SQL — `JOIN dim d` — means a *maintained join* in a plain MV but a *silently frozen, sampled-once snapshot* inside a `RECORD` body, with freeze as the **default** and tracking the opt-in. This is spooky-action-at-a-distance: identical join syntax flips semantics based on the enclosing object kind, with no marker at the join site. The plan-time taint rule protects plain MVs from accidental freeze, but nothing protects a `RECORD` author from *accidentally* freezing a reference they meant to track — and they get no error. I'd push Open Question 4 further: consider an explicit marker (even if redundant for the type checker) or strong `EXPLAIN`/notice output naming every frozen reference in a body. Arguably the *safe* (tracked) reading should be the default and freeze the explicit one.

**2. The multi-writer domain-A completion frontier is underspecified.**
Several `RECORD` writers into one delta table is marked *Decided*, and the table-owned reclock "recovers the precise A→B mapping." But `INTEGRATE` must drive `v`'s domain-A frontier, which is a *meet* across all active writers' committed-through A-times — one slow/idle writer holds back the whole table's A-completeness. The doc never states this meet explicitly, nor what happens when a writer is dropped (does the A-frontier jump forward?) or stalls indefinitely. Given multi-writer is decided, the cross-writer A-frontier computation deserves specification, not just an assertion that it "works."

**3. The `mz_now()`/aging domain (A vs B) is flagged open, but it's the most consequential unresolved semantic — resolve it early.**
Whether retention `mz_now() < mz_timestamp + W` is *event-age* (domain A, stalls when input idles) or *wall-clock* (domain B, always advances) changes user-visible behavior dramatically and dictates which temporal-filter machinery you reuse (M3/M4). This isn't polish-phase; it gates the bounding design and the user mental model. Worth pulling forward to a Phase 0/1 decision.

**4. `mz_timestamp`-surfaced-as-data instability may quietly undermine the headline dedup/first-seen use case.**
The pitfall is honestly documented: `… AS first_seen_at` inherits the `max(mz_timestamp, upper)` clamp and can *advance as `since` ticks forward*, stable only within `RETAIN HISTORY`. But "first seen at" is precisely the column users will project and expect to be immutable — tying its stability to a retention knob makes a correctness-flavored property depend on an ops setting. This is more than "document it"; it warrants a louder warning / lint (gestured at in Open Q5), and the dedup example shouldn't be read as endorsing a stable `first_seen_at` it can't guarantee.

**5. Make the "DELTA TABLE is not recomputable from inputs" optimizer barrier a stated hard invariant.**
The impl doc (Open Q7) correctly says the optimizer must **not** treat the delta table as recomputable from the original inputs, but **may** treat `INTEGRATE` as recomputable over the delta table. This is correctness-critical — if the optimizer ever "sees through" `RECORD` to re-derive frozen/non-deterministic values, the determinism-boundary argument collapses. It currently reads as a passing remark; promote it to a named invariant in the conceptual doc's "determinism boundary" section, which is the load-bearing soundness argument.

**6. Success-criterion vs. mechanism mismatch on compliance erasure.**
Success Criteria says deleting a user **"physically erases"** their recorded rows. "Bounding growth" then correctly explains true erasure requires advancing `since` (forfeiting `AS OF`/replay in that range), and that a consolidating retraction at event-time alone leaves rows visible to earlier `AS OF` reads — "insufficient for GDPR." These are consistent only if you read the criterion as "consolidating delete + `since` advancement." As written it overpromises; restate the criterion with the `since`-advancement caveat so it can't be read as "a cascade `DELETE` alone is GDPR-compliant."

### Minor / editorial

- **Heavy duplication between the two docs.** The time-domains/reclock explanation appears nearly verbatim in both the conceptual doc ("Time domains and reclocking") and the impl doc (Open Q7) — ~40 echoed lines. The impl doc should reference rather than restate.
- **`RECORDER` vs `RECORD` writer** terminology drifts (example says `CREATE RECORDER … INTO`, prose says "the `RECORD` writer", title says "Recorders"). Self-flagged as open, but worth picking one for readability even in draft.
- **Tier table:** Tier 1 "exactly-once into persist" still rests on the per-commit CAS (stated elsewhere); worth noting in the table.
- The `DELETE (SELECT …) FROM enriched` cascade example reads as pseudo-syntax — fine given "syntax TBD," but maybe label it inline.

### Recommendation
Approve the *direction* for discussion — the doc is in excellent shape for driving WG consensus. Before it graduates from draft I'd want: (a) the freeze-default safety question resolved, (b) the multi-writer A-frontier specified, (c) the `mz_now()` domain chosen, (d) the compliance success-criterion reworded to match the mechanism. Items 2–4 are also the most likely to bite the implementation, so resolving them now de-risks Phase 1.

---

antiguru:
## Design review: derive the store from requirements, and record the alternative it beats

This distills a long design discussion (mh + Claude Code) that stress-tested the recorded-table model against a simpler alternative. The conclusion **supports the design's store choice**, but on a rationale the doc does not currently state. Capturing the derivation, the rejected alternative, and the resulting doc fixes here so the next reviewer doesn't spend the same four rounds rediscovering them.

### The derivation the doc should contain

The doc currently justifies the store by minimality framing ("no new collection kind", precedent for the progress collection). Those framings are contestable, and the discussion contested them at length. What actually settles the store choice is a chain of requirements, none of which is written down as the rationale today:

1. **Consistency requires domain-A read placement.** An event at `t` must be visible in the recorded output's reconstruction at `t`. No derived object in Materialize moves data in time, an MV presents rows at exactly its inputs' timestamps and expresses processing delay as frontier lag. A recorder placing outputs at commit time would be the first in-system derivation to break that, and there would be *no* timestamp at which `output@t = f(input@t)` holds. That eventually-consistent seam is exactly what Kafka-loopback and Postgres-round-trip users suffer today, and "stay inside Materialize *and keep consistency*" is the feature's pitch.
2. **DML on the recorded collection is a hard requirement.** An MV can be DML-less because recompute is its escape hatch (drop, fix, rebuild). A recorder's output is definitionally non-recomputable, so bounding, correction, and erasure have no other mechanism. All collections must have a way to bound their size.
3. **DML must be ordinary, data-plane DML.** DML execution is moving fully onto replicas, with no data transiting envd and no new data channels into replicas. So DML on a recorded collection must use the same write substrate as all DML: txn-wal, from a replica, which the envd horizontal-scalability work (multi-process txn-wal) makes available there.
4. **A txn-wal-registered shard cannot lag.** Registration advances the shard's logical upper with the txns timeline, en masse, at system pace (`src/txn-wal/src/lib.rs`). A shard whose upper advances at the recorder's pace cannot be registered.
5. Therefore the store's physical timeline is domain B, the event times must ride as data, and completeness must be reconstructed out-of-band. **The carriers, the progress collection, and `INTEGRATE` are forced by 1–4, not chosen.**

Corollary worth stating as a design property: because logical placement is data and serialization is the txns shard, a recorder, a compaction `DELETE`, and an erasure `DELETE` on three different clusters can interleave arbitrarily and never distort placement. A lost CAS delays a physical write by one tick and shifts nothing, since the shard timestamp was never load-bearing. This race-immunity is the strongest single argument for placement-as-data and it appears nowhere in the doc.

### The alternative to record in Alternatives (rejected at steps 3–4)

**Own-shard recorder output**: the recorder writes actual inserts/retractions (diff as persist multiplicity) into a collection whose shard timeline *is* domain A, MV-style, upper driven by the recorder's frontier via replica-side `compare_and_append` on a raw shard (the shipped MV-sink / CT-sink pattern). The table is the integrated collection, `CHANGES(output)` recovers the changelog, the shard's own upper is the progress.

Why it's attractive: the user surface collapses to `CREATE RECORDER` + `CHANGES` + an explicit freeze marker. No carriers in schemas, no `WITH (PROGRESS, …)`, no user-visible progress object, no `INTEGRATE` on reads, no integral-preserving DML discipline, upsert retractions consolidate natively, and multi-writer becomes `UNION ALL` views where the meet of frontiers is ordinary dataflow frontier propagation.

Why it fails, progressively:
- *Naive form* (DML as an independent CAS writer on the shard): a lost CAS at `t+1` forces the recorder to re-place deltas for input times `≤ t` at `t+2`, so the frontier over-claims completeness for one tick, permanently wrong at that slice. When placement is encoded in the shard timestamp, a physical write conflict becomes a logical placement shift.
- *Serialized form* (a single commit loop unions body diffs and DML diffs per tick, DML read at `t` applied at `t+1`, non-negativity guaranteed because a successful CAS proves no interleaving): correct, but the loop must live on a replica (data plane), and then DML diffs must reach the recorder's sink, which requires either a new data channel through envd (ruled out by the platform direction) or cluster co-location with a rendezvous at the sink (a worse special case). The registered-table model gets DML from any cluster through the ordinary substrate.
- A like-for-like state note the doc could carry: the registered-table model buys a *blind* writer by deferring retraction/DML collision handling to read time, `INTEGRATE`'s `max(0, Σ)` threshold is the read-side safety net for write-side blindness, paid per reading dataflow. The own-shard model pays it once, writer-side, with exact semantics. This is a conscious trade (readers pay, writer stays blind) and should be stated as one.

### Implementation-doc corrections

- **Option B contradicts the data-plane requirement as written.** It says the body's proposed diffs are drained by "a control-plane loop" that commits, with a "dataflow → coordinator subscribe". The recorder must live entirely on a replica. The corrected architecture is a replica-side commit loop participating in multi-process txn-wal. The real Option A/B fork is then raw-CAS vs txn-wal, and it is settled by step 4 above, not by the data+progress atomicity argument alone.
- **Name multi-process txn-wal as a gating dependency** alongside the OCC substrate, including replica-side write-timestamp coordination and the oracle write-back (`apply_write`) path from replicas. Today the only production txn-wal writer is the storage controller's table worker in envd (`src/storage-controller/src/persist_handles.rs`), and replicas hold txn-wal machinery read-only (`src/compute/src/compute_state.rs`, `TxnsContext`/`TxnsCache`). The design is sequenced behind the horizontal-scalability deliverables and should say so.
- The rejection of in-band progress markers argues its single-CAS upside is "moot" because each `RECORD` commits table + progress together. That mootness holds only under txn-wal commits, worth noting, since it was nearly decisive under a raw-CAS architecture.

### Surface ergonomics (the concepts must exist, so make them cheaper)

1. **Cut `USING TIME …, DIFF …` from `CHANGES`.** Standard table-alias column lists already parse on the #36869 branch (`CHANGES(t AS OF …) AS e (a, b, …, change_ts, change_diff)`), and a recorder body's SELECT list can rename `mz_timestamp`/`mz_diff`. Column correspondence with the target table is what `INSERT INTO … SELECT` already does, and the table's `WITH (…)` clause already declares which columns are carriers. This also decouples this design from #36869's surface entirely (see companion comment there).
2. **Make freeze explicit and mandatory, not silent-with-a-NOTICE.** A bare TVC reference in a recorder body should be an error naming the two fixes: `CHANGES(dim …)` (tracked) or `FROZEN dim` (sampled). `LATERAL` is grammar precedent for a prefix table-factor modifier. This is the direction-safe choice: explicit-to-default is a compatible relaxation later, silent-to-explicit is a breaking change. It also marks, in the query text, exactly where the consistency guarantee ends: row placement is time-consistent, frozen values are deliberately processing-time facts. The same argument covers surfacing `now()` (frozen wall-clock sample) vs `mz_now()` (domain-A event time) loudly.
3. **Answer the open DML-guardrail question with an engine rewrite.** `DELETE FROM t WHERE <carrier-only predicate>` is exactly the data-domain-compaction case, and the planner can recognize the shape and execute the clamp + `GROUP BY`/`SUM` reduce instead of the literal delete. Predicates on data columns are *intended* integral changes (erasure, correction) and pass through. This converts the most dangerous documented hazard ("a hand-written DELETE that is not integral-preserving silently changes INTEGRATE's result", acknowledged unenforceable) into correct behavior for the most common bounding statement.
4. **Defer multi-writer lanes and freeze-vs-seal past v1.** Single writer per table. txn-wal serializing all writers doesn't change that the per-writer-lane meet machinery only pays off once a second writer exists, and the only named use case is webhook demux.
5. **Name the progress collection.** `WITH (PROGRESS)` creates an object with no name, and therefore no SHOW/reference/DROP story. The cited precedent spelled it `EXPOSE PROGRESS AS <name>`, and the subsource history is a cautionary tale about auto-created unnamed subobjects.
6. **Fix the window-vs-driver rule.** The doc's rule "a `CHANGES` over a long-lived input must be bounded with `AS OF AT LEAST mz_now() − W` *and* a temporal filter" cannot apply as written to recorder bodies: under #36869's maintained mode, window expiry produces retraction deltas, and a recorder recording driver deltas would record un-recordings, draining the recorded output on a rolling basis. A recorder needs a consume-mode `CHANGES` whose input hold trails the recorder's own progress frontier, with no aging retractions. Details in the companion comment on #36869.

---
Distilled from a design-review session. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_013bPwEhF2LCYTWmQbjTdUM9

---

PR #36913, opened 2026-06-05
timely-util: stop re-sorting the whole reclock remap_trace each invocation

DAlperin:
STEP 5 of the reclock operator re-consolidated and re-sorted the entire remap binding trace on every invocation: consolidate_updates (sort by (FromTime, IntoTime)) followed by a re-sort into the IntoTime order STEP 4 requires -- two full O(n log n) passes in conflicting orders. This dominated CPU during source hydration, when [remap_since, remap_upper) (and thus the trace) grows large because remap_upper races ahead of remap_since.

The operator already maintains the trace in IntoTime order, so replace the two passes with a single sort_by((IntoTime, FromTime)) + in-place consolidate. sort_by (driftsort) detects the already-sorted tail as a natural run and only sorts the small prefix that advance_by just collapsed onto remap_since, then merges -- a single ~O(n) pass instead of two O(n log n) passes (it removes the log factor and the redundant second sort; it is not sub-linear, since the merge still scans the whole trace). Factor the logic into consolidate_remap_trace with a unit test.

### DIFF
--- a/src/timely-util/src/reclock.rs
+++ b/src/timely-util/src/reclock.rs
@@ -177,7 +177,7 @@ use std::iter::FromIterator;
 
 use differential_dataflow::difference::Semigroup;
 use differential_dataflow::lattice::Lattice;
-use differential_dataflow::{AsCollection, ExchangeData, VecCollection, consolidation};
+use differential_dataflow::{AsCollection, ExchangeData, VecCollection};
 use mz_ore::Overflowing;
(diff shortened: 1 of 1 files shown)


antiguru (approved):
Looks fine

DAlperin:
Closing in favor of https://github.com/MaterializeInc/materialize/pull/34267

PR #36916, opened 2026-06-05
pgwire: match PostgreSQL's diagnostic for missing binary output functions

aljoscha:
When a client requests a binary result or COPY binary format for a type that has no binary output function, Materialize now reports the same diagnostic as PostgreSQL: SQLSTATE 42883 (undefined_function) with the message `no binary output function available for type <t>`. Verified against postgres:16:

    COPY (SELECT '=r/postgres'::aclitem) TO STDOUT WITH (FORMAT binary);
    ERROR:  42883: no binary output function available for type aclitem

PostgreSQL reports this recursively too, e.g. for aclitem[] and records, which our recursive check mirrors.

08P01 (protocol_violation) would be the wrong code here: protocol_violation means the client sent something malformed, but the client did nothing wrong -- it legitimately requested a binary result or COPY format and the server simply has no binary output function for the type. Adopting PostgreSQL's message and SQLSTATE means drivers and tools that branch on the standard error code -- and users who already know the PostgreSQL diagnostic -- behave consistently against Materialize.

list and map don't exist in PostgreSQL, so there's no upstream wording to copy for them; we use the same phrasing with the type name ("type list" / "type map") for consistency.

Only the user-facing diagnostics returned by `binary_encoding_error` and the SQLSTATE at the two pgwire call sites change. The "should never happen" backstop strings inside `encode_binary` itself are left as-is: they are unreachable because every caller validates first, and they sit next to the comments explaining why a binary encoding isn't implemented.

### DIFF
--- a/src/pgrepr/src/value.rs
+++ b/src/pgrepr/src/value.rs
@@ -593,6 +593,11 @@ impl Value {
     /// types) supports binary encoding, or `Err(reason)` describing the first
     /// unsupported type encountered. Container types are checked recursively, so
     /// e.g. a record or array that contains a `list` is rejected.
+    ///
+    /// The error messages mirror PostgreSQL's `no binary output function
+    /// available for type <t>` so that drivers and users see a familiar
+    /// diagnostic. Callers should report these errors with the SQLSTATE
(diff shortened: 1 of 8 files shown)


ggevay (approved):
Thank you!

PR #36895, opened 2026-06-04
Refactor peek optimizer dispatch with PeekOptimizer enum

antiguru:
### Motivation

The peek sequencing state machine currently uses `Either<peek::Optimizer, copy_to::Optimizer>` to dispatch between `SELECT`/`EXPLAIN` and `COPY TO` optimization paths. This approach is verbose and requires pattern matching at multiple sites. Consolidating this into a dedicated enum improves code clarity and reduces duplication.

### Description

This PR introduces a `PeekOptimizer` enum that wraps the two optimizer variants and provides a unified interface for the shared optimization pipeline:

1. **New `PeekOptimizer` enum** (`src/adapter/src/optimize.rs`):
   - `Select(peek::Optimizer)` variant for `SELECT`/`EXPLAIN` statements
   - `CopyTo(copy_to::Optimizer)` variant for `COPY TO` statements
   - Methods: `cluster_id()`, `optimize()`, `into_select()`, `into_copy_to()`

2. **New `PeekGlobalLirPlan` enum** (`src/adapter/src/optimize.rs`):
   - Tagged result type that preserves which optimization path was taken
   - `Select(peek::GlobalLirPlan)` and `CopyTo(copy_to::GlobalLirPlan)` variants

3. **Shared `optimize_oneshot()` helper** (`src/adapter/src/optimize.rs`):
   - Factors out the duplicated HIR → local MIR → resolve → global LIR pipeline
   - Uses generic bounds to work with both optimizer types
   - Eliminates ~30 lines of duplicated optimization logic

4. **Updated state machine** (`src/adapter/src/coord.rs`, `src/adapter/src/coord/sequencer/inner/peek.rs`):
   - Replaces `Either<peek::Optimizer, copy_to::Optimizer>` with `PeekOptimizer` in all peek stage structs
   - Simplifies pattern matching in the optimization stage
   - Uses `into_select()` and `into_copy_to()` to recover concrete optimizers after optimization

5. **Removed unused imports**:
   - Removes `itertools::Either` from `coord.rs` and `peek.rs` (no longer needed)
   - Removes unused `Optimize` trait import from `peek.rs`

### Verification

This is a refactoring that preserves the existing optimization pipeline behavior. The changes:
- Consolidate duplicated code paths into a single `optimize_oneshot()` function
- Replace verbose `Either` pattern matching with cleaner enum dispatch
- Maintain the same optimization semantics for both `SELECT` and `COPY TO` paths

[...]

PR #36921, opened 2026-06-05
adapter: reuse optimize_oneshot in frontend_peek pipelines

antiguru:
### Motivation

Follow-up to #36895, which introduced `optimize::PeekOptimizer` and factored
the shared one-shot optimization pipeline into `optimize::optimize_oneshot`.

The frontend peek path (`frontend_peek.rs`) still hand-rolled the same
`HIR -> local MIR -> resolve -> global LIR` sequence in both its `COPY TO`
and `SELECT`/`EXPLAIN` arms — the exact sequence now encapsulated by
`optimize_oneshot`. This makes that helper `pub(crate)` and calls it from
both arms.

I deliberately did **not** push the `PeekOptimizer` enum into this file: its
dispatch is a three-way `match` that also includes `Subscribe` (a different
pipeline shape), and the `Select`/`CopyTo` arms each need the concrete
optimizer downstream. The right shared unit here is the pipeline helper, not
the dispatch wrapper.

### Tips for reviewer

As a side effect, the `SELECT` arm's `pipeline` closure disappears. It only
existed to corral multiple `?` operators into a local `Result` so a failure
could still be routed to `EXPLAIN BROKEN`; that is now a single fallible call
whose result we capture with `.map_err(AdapterError::from)`, preserving the
broken-explain handling.

No behavior change — pure refactor. `cargo check -p mz-adapter` is clean.

### Checklist

- [x] This PR has adequate test coverage (no behavior change; covered by existing peek/copy-to tests).
- [x] This PR does not need a release note.

https://claude.ai/code/session_018q73DM2B1dQAb6MQbqTHHk

---
_Generated by [Claude Code](https://claude.ai/code/session_018q73DM2B1dQAb6MQbqTHHk)_

### DIFF
--- a/src/adapter/src/frontend_peek.rs
+++ b/src/adapter/src/frontend_peek.rs
@@ -889,19 +889,18 @@ impl PeekClient {
                         span.in_scope(|| {
                             let _dispatch_guard = explain_ctx.dispatch_guard();
 
-                            // COPY TO path
-                            // HIR ⇒ MIR lowering and MIR optimization (local)
-                            let local_mir_plan =
-                                optimizer.catch_unwind_optimize(raw_expr.clone())?;
-                            // Attach resolved context required to continue the pipeline.
-                            let local_mir_plan = local_mir_plan.resolve(
-                                timestamp_context.clone(),
-                                &session_meta,
-                                stats,
-                            );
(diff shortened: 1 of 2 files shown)


ggevay (approved):
Thank you, LGTM!

PR #36922, opened 2026-06-05
[ss-159] Accept empty exclude columns for consistency with text columns

peterdukelarsen:
### Motivation

While working on https://github.com/MaterializeInc/materialize/pull/36890 I noticed that EXCLUDE COLUMNS has the same behavior as TEXT COLUMNS. For consistency's sake I've implemented the same fix.

### Description

Removes panic/error log for empty exclude columns because we supported and want to support empty exclude columns.

### Verification
Testdrive tests asserting legacy behavior aligns with expectations and new behavior matches it.

### DIFF
--- a/test/pg-cdc-old-syntax/empty-text-columns.td
+++ b/test/pg-cdc-old-syntax/empty-text-columns.td
@@ -7,13 +7,9 @@
 # the Business Source License, use of this software will be governed
 # by the Apache License, Version 2.0.
 
-#
 # Regression test for ss-159: an empty `TEXT COLUMNS ()` on the legacy
 # `CREATE SOURCE .. FOR TABLES` (subsource-creating) syntax is treated as
-# though the option were omitted and must succeed. This differs from the
-# `CREATE TABLE .. FROM SOURCE` syntax, where an empty set is rejected with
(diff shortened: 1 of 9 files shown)


test/pg-cdc-old-syntax/empty-text-columns.td
```diff
@@ -7,13 +7,9 @@
 # the Business Source License, use of this software will be governed
 # by the Apache License, Version 2.0.
 
-#
 # Regression test for ss-159: an empty `TEXT COLUMNS ()` on the legacy
 # `CREATE SOURCE .. FOR TABLES` (subsource-creating) syntax is treated as
-# though the option were omitted and must succeed. This differs from the
-# `CREATE TABLE .. FROM SOURCE` syntax, where an empty set is rejected with
-# `invalid TEXT COLUMNS option value: cannot be empty`.
```
peterdukelarsen, line 15: I caught some stale comments that slipped through on the old PR from when I was planning to make this an error. Generally not a fan of mixing changes, but it seemed related and trivial enough.

martykulma (approved):
sweet - thanks @peterdukelarsen !

PR #36932, opened 2026-06-08
[SS-104] Mysql drop create corruption

peterdukelarsen:
### Motivation
There's a race of sorts between dropping and recreating a table with a certain name in MYSQL and materialize processing the drop. In particular, on DROP TABLE materialize reaches back to MYSQL to diff its sources with the tables in MYSQL to determine what's been dropped without needing to do too much SQL parsing. Unfortunately, we can't differentiate (in MYSQL at least) between the table being there and the table having been dropped and a new table with the same name and compatible schema having been recreated.

Fixes [SS-104](https://linear.app/materializeinc/issue/SS-104/mysql-drop-create-data-corruption)

### Description
To fix this issue, we parse the dropped tables from the mysql binlog and poison any sources (that haven't already been poisoned) with matching table names.


src/storage/src/source/mysql/replication/events.rs
```diff
@@ -18,18 +23,30 @@ use mz_storage_types::sources::mysql::GtidPartition;
...
 /// Returns the MySqlTableName for the given table name referenced in a
 /// SQL statement, using the current schema if the table name is unqualified.
 fn table_ident(name: &str, current_schema: &str) -> Result<MySqlTableName, TransientError> {
     let stripped = name.replace('`', "");
     let mut name_iter = stripped.split('.');
-    match (name_iter.next(), name_iter.next()) {
+    mysql_table_name(name_iter.next(), name_iter.next(), current_schema, name)
+}
+
+fn mysql_table_name(
```
peterdukelarsen, line 44: Small pure refactor here to pull out the mysql table name building.

src/storage/src/source/mysql/replication/events.rs
```diff
@@ -39,6 +56,33 @@ fn table_ident(name: &str, current_schema: &str) -> Result<MySqlTableName, Trans
     }
 }
 
+fn table_ident_from_object_name(
```
peterdukelarsen, line 63: This function has the same intent as `table_ident` except handles the object name type parsed out of the sql by sqlparser rather than a raw string.

The `ObjectName` type is more flexible than the constraints for an identifier in mysql. i.e. it has a function type, which appears to only be supported in snowflake. It also has a vector for identifier components, however a table identifier from the binlog should only have 1 or 2 components - never 0 or more than 2

src/storage/src/source/mysql/replication/events.rs
```diff
@@ -112,38 +156,40 @@ pub(super) async fn handle_query_event(
...
-                    (dropped_output.output_index, Err(err.into())),
-                    new_gtid.clone(),
-                    Diff::ONE,
-                );
-                let size = std::mem::size_of_val(&update);
-                ctx.data_output.give_fueled(&gtid_cap, update, size).await;
-                ctx.errored_outputs.insert(dropped_output.output_index);
+            for (table_name, outputs) in tables_to_drop {
+                for output in outputs {
+                    let err = DefiniteError::TableDropped(table_name.to_string());
```
peterdukelarsen, line 188: Same message we would have gotten on a missing table calling verify_schemas: https://github.com/MaterializeInc/materialize/blob/097450b02012cbf1de0d5c1f10ae3282a07a2343/src/storage/src/source/mysql/schemas.rs#L56

src/storage/src/source/mysql/replication/events.rs
```diff
@@ -109,41 +153,46 @@ pub(super) async fn handle_query_event(
...
+                            dropped output: {output:?}: {err:?}");
+                    let gtid_cap = ctx.data_cap_set.delayed(new_gtid);
+                    let update = (
+                        (output.output_index, Err(err.into())),
+                        new_gtid.clone(),
+                        Diff::ONE,
+                    );
+                    let size = std::mem::size_of_val(&update);
+                    ctx.data_output.give_fueled(&gtid_cap, update, size).await;
+                    ctx.errored_outputs.insert(output.output_index);
```
peterdukelarsen, line 197: this block of code here is the same, just indented one layer deeper.

peterdukelarsen:
You can see test runs here: https://github.com/MaterializeInc/materialize/pull/36961. This will be the follow-up PR once this merges.
You can see the test-only diff here: https://github.com/peterdukelarsen/materialize/pull/2/changes

src/storage/src/source/mysql/replication/events.rs
```diff
@@ -179,10 +180,9 @@ pub(super) async fn handle_query_event(
                 .collect();
             is_complete_event = true;
             for (table_name, outputs) in sources_to_drop {
+                tracing::info!(%id, "timely-{worker_id} DDL change dropped outputs: {outputs:?}");
```
martykulma, line 186: may want to include `table_name`

peterdukelarsen: Took this feedback at first, but it's actually included as one of the fields of the outputs so I don't think we need it.

martykulma: nice!

src/storage/src/source/mysql/replication/events.rs
```diff
@@ -39,6 +57,33 @@ fn table_ident(name: &str, current_schema: &str) -> Result<MySqlTableName, Trans
...
+    current_schema: &str,
+) -> Result<MySqlTableName, TransientError> {
+    let processed_name_parts: Vec<String> = name.0.iter().map(|part| match part {
+        ObjectNamePart::Identifier(ident) => Ok(ident.value.clone()),
+        // Functions for table name identifiers are a snowflake-specific concept, unexpected for mysql so we should fail hard.
+        ObjectNamePart::Function(_) => Err(TransientError::Generic(anyhow::anyhow!(
+            "Invalid table name from QueryEvent, function identifiers not supported in mysql: {}", name
+        ))),
+    }).collect::<Result<_, _>>()?;
+    if !(processed_name_parts.len() > 0 && processed_name_parts.len() <= 2) {
```
martykulma: hard to reason about - there are only 2 valid conditions. possible alternatives `!= 1 && != 2` or `.is_empty() || .len() > 2`

martykulma (approved):
lgtm!

src/storage/src/source/mysql/replication/events.rs
peterdukelarsen: done, thanks!

PR #36933, opened 2026-06-09
Add mz-aws-glue-schema-registry crate with basic client

martykulma:
Adds an AWS Glue schema registry client and create. Follows a pattern similar to the existing Confluent schema registry client.

This PR implements the following APIs that will be used during the setup of a GSR connection and a kafka source using GSR:
* `Client::get_registry`
  * used to verify a registry exists at `CREATE CONNECTION` time.
* `Client::get_schema_version_by_id`
  * source decode: fetch a writer schema by the UUID embedded in each record's Glue wire-format header.
* `Client::get_schema_version_latest_by_name`
  * DDL planning: lookup and pin a reader schema to the registry's current "latest" version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -792,6 +792,30 @@ dependencies = [
 ]
 
 [[package]]
+name = "aws-sdk-glue"
+version = "1.142.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc77647907307c70ffd751db85653804552e4a3c27f054d3af7a0874ef4dfe22"
+dependencies = [
+ "aws-credential-types",
+ "aws-runtime",
+ "aws-smithy-async",
+ "aws-smithy-http",
+ "aws-smithy-json",
+ "aws-smithy-observability",
+ "aws-smithy-runtime",
+ "aws-smithy-runtime-api",
+ "aws-smithy-types",
+ "aws-types",
+ "bytes",
+ "fastrand 2.3.0",
+ "http 0.2.12",
+ "http 1.4.0",
+ "regex-lite",
+ "tracing",
+]
+
+[[package]]
 name = "aws-sdk-kms"
 version = "1.104.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5680,6 +5704,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "mz-aws-glue-schema-registry"
+version = "0.0.0"
+dependencies = [
+ "anyhow",
+ "aws-sdk-glue",
+ "aws-smithy-runtime-api",
+ "aws-types",
+ "thiserror 2.0.18",
+ "tracing",
+ "uuid",
+]
+
+[[package]]
 name = "mz-aws-secrets-controller"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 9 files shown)


src/aws-glue-schema-registry/src/client.rs
```diff
@@ -0,0 +1,269 @@
...
+            description: output.description,
+            lifecycle_status: output.status.map(RegistryLifecycleStatus::from_sdk),
+        })
+    }
+
+    /// Fetch a schema version by its UUID.
+    ///
+    /// This is the source-decode path: the UUID is read from the Glue
+    /// wire-format header, and the returned `SchemaVersion::definition`
+    /// carries the writer schema (Avro JSON, for our usage).
```
peterdukelarsen, line 67: Just to confirm I understand: "Avro JSON, for our usage" means that for the Glue schema registry we will only support AVRO source data. So if the schema type returned by Glue was protobuf we'd want to error out.

I'm assuming that check will be handled by the consuming code, which seems reasonable to me.

src/aws-glue-schema-registry/src/client.rs
```diff
@@ -0,0 +1,269 @@
...
+/// document the Avro parser can ingest directly. The remaining fields are
+/// informational and exist for debug logging.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SchemaVersion {
+    pub schema_version_id: Option<String>,
+    pub schema_arn: Option<String>,
+    /// The format-specific schema text (Avro JSON, JSON Schema, etc.).
+    pub definition: Option<String>,
+    /// Glue's data-format tag — e.g. `"AVRO"` or `"JSON"`. Kept as a string
+    /// rather than the SDK enum so the SDK type doesn't leak to callers.
```
peterdukelarsen: Why not define a new enum and convert from the sdk type to our internal type like you did with `RegistryLifecycleStatus`?

Cargo.lock
```diff
@@ -791,6 +791,30 @@ dependencies = [
...
+ "aws-smithy-runtime",
+ "aws-smithy-runtime-api",
+ "aws-smithy-types",
+ "aws-types",
+ "bytes",
+ "fastrand 2.3.0",
+ "http 0.2.12",
+ "http 1.4.0",
+ "regex-lite",
+ "tracing",
```
peterdukelarsen, line 815: Overall impact of the new dependency here appears to be two net-new dependencies in spite of this long list of transitive deps, right? Just aws-sdk-glue and mz-aws-glue-schema-registry.

This all seems nicer than the mz-ccsr because you don't need to hand-roll the sdk.

peterdukelarsen (commented):
High-level approach seems reasonable to me here -- a package that wraps the full SDK and re-exposes only the minimal set of endpoints and types.

src/aws-glue-schema-registry/src/client.rs
martykulma: yes

src/aws-glue-schema-registry/src/client.rs
martykulma: ah - when I saw that definition was a string, I assumed format was as well.  Nice find!

Cargo.lock
martykulma: yeah, those transitive deps are fairly standard for AWS (based on the other crates in the file).

src/aws-glue-schema-registry/src/client.rs
martykulma: fixed

peterdukelarsen (approved):

martykulma:
TFTR!

PR #36936, opened 2026-06-09
tests: Fix lots of flakes happening over last 1.5 weeks

def-:
Based on:
- https://buildkite.com/materialize/nightly/builds/16684
- https://buildkite.com/materialize/test/builds/124732
- https://buildkite.com/materialize/test/builds/124721
- https://buildkite.com/materialize/test/builds/124705
- https://buildkite.com/materialize/test/builds/124969

### DIFF
--- a/misc/python/materialize/mzcompose/services/polaris.py
+++ b/misc/python/materialize/mzcompose/services/polaris.py
@@ -88,7 +88,15 @@ class Polaris(Service):
                 "interval": "1s",
                 "timeout": "5s",
                 "retries": 10,
-                "start_period": "2s",
+                # The Quarkus JVM plus its JDBC connection to Postgres can take
+                # well over a minute to start listening on the management port
+                # when the CI host is under heavy load (every container starting
+                # at once). A short start_period made `docker compose up --wait`
+                # declare Polaris unhealthy, which triggered an `up` retry that
+                # re-ran polaris-bootstrap and failed with "realm already
+                # bootstrapped" (exit 3). Give Polaris generous time to come up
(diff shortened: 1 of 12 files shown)


misc/python/materialize/mzcompose/services/polaris.py
```diff
@@ -88,7 +88,15 @@ def __init__(
...
-                "start_period": "2s",
+                # The Quarkus JVM plus its JDBC connection to Postgres can take
+                # well over a minute to start listening on the management port
+                # when the CI host is under heavy load (every container starting
+                # at once). A short start_period made `docker compose up --wait`
+                # declare Polaris unhealthy, which triggered an `up` retry that
+                # re-ran polaris-bootstrap and failed with "realm already
+                # bootstrapped" (exit 3). Give Polaris generous time to come up
+                # so it reliably passes its health check on the first attempt.
+                "start_period": "120s",
```
Alphadelta14, line 99: I'd recommend increasing the interval so that your 10 retries are not running out in 10 seconds and that you are considerably started within your normal retries.
12 retries with a 10s interval would give you the same results but if we are successfully healthy in 60 seconds, we have to wait half as long :)

Alphadelta14 (approved):

misc/python/materialize/mzcompose/services/polaris.py
def-: That's not how the start period work I think. We have an interval of 1s, so in those 120s we check every second whether Polaris is up already, and then continue.

PR #36938, opened 2026-06-09
mcp: advertise Frontegg authorization server via RFC 9728

bobbyiliev:
Follow-up to #36669 that adds the Frontegg arm of MCP OAuth discovery. Introduces `--frontegg-oauth-issuer-url` (env `FRONTEGG_OAUTH_ISSUER_URL`) on environmentd; when set on a Frontegg listener, `/.well-known/oauth-protected-resource` advertises that URL as the authorization server and 401 Bearer challenges point clients at it. Cloud-side plumbing in MaterializeInc/cloud#12773.

### DIFF
--- a/src/environmentd/src/http.rs
+++ b/src/environmentd/src/http.rs
@@ -172,6 +172,7 @@ pub struct HttpConfig {
     /// today, and an attacker reaching the server directly can otherwise
     /// poison the published metadata URLs.
     pub http_host_name: Option<String>,
+    pub frontegg_oauth_issuer_url: Option<String>,
     pub concurrent_webhook_req: Arc<tokio::sync::Semaphore>,
     pub metrics: Metrics,
     pub metrics_registry: MetricsRegistry,
@@ -230,6 +231,7 @@ impl HttpServer {
             active_connection_counter,
             helm_chart_version,
             http_host_name,
+            frontegg_oauth_issuer_url,
             concurrent_webhook_req,
             metrics,
             metrics_registry,
@@ -243,6 +245,14 @@ impl HttpServer {
     ) -> HttpServer {
         let tls_enabled = tls.is_some();
         let webhook_cache = WebhookAppenderCache::new();
+
+        // Compute OAuth discovery once per listener so the Bearer challenge
+        // and the discovery handler always agree, and the middleware doesn't
+        // re-derive (and re-allocate the Frontegg issuer) on each request.
+        let oauth_discovery = Arc::new(oauth_metadata::McpOAuthDiscovery::for_authenticator(
+            authenticator_kind,
+            frontegg_oauth_issuer_url.as_deref(),
+        ));
 
         // Create secure session store and manager
         let session_store = TowerSessionMemoryStore::default();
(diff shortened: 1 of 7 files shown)


src/environmentd/src/http.rs
```diff
@@ -959,6 +968,9 @@ where
 pub(crate) struct BearerChallengeConfig {
     /// OAuth scope advertised in the `Bearer` challenge's `scope=` parameter.
     pub scope: &'static str,
+    /// Precomputed once at server build time and shared with the discovery
+    /// handler so the two never disagree.
```
SangJunBak: "two never disagree" is a bit vague and I don't quite understand what that means

src/environmentd/src/http.rs
```diff
@@ -589,6 +597,7 @@ impl HttpServer {
                 .layer(auth_middleware.clone())
                 .layer(Extension(BearerChallengeConfig {
                     scope: oauth_metadata::MCP_SCOPE,
+                    discovery: Arc::clone(&oauth_discovery),
```
SangJunBak, line 597: To simplify, could we just merge BearerChallengeConfig into DiscoveryConfig and rename to like `McpOAuthConfig` then maybe have an ` impl McpOAuthConfig {get_scope}` method that gives back ` oauth_metadata::MCP_SCOPE,`? That way we don't need to plumb `http_host_name` into `auth_middleware`

src/environmentd/src/http/oauth_metadata.rs
```diff
@@ -178,32 +178,45 @@ impl OauthMetadataMetrics {
...
-            | listeners::AuthenticatorKind::Password
-            | listeners::AuthenticatorKind::Sasl
-            | listeners::AuthenticatorKind::None => Self::Disabled,
+    /// advertisement behavior. Frontegg requires a valid issuer URL;
+    /// otherwise falls back to `Disabled`.
+    pub(crate) fn for_authenticator(
+        kind: listeners::AuthenticatorKind,
+        frontegg_oauth_issuer_url: Option<&str>,
+    ) -> Self {
+        match (kind, frontegg_oauth_issuer_url) {
```
SangJunBak, line 205: Something I didn't catch earlier is for Oidc, we could technically be using password authentication even when the listener kind if Oidc. Below we map to the actually auth strategy in `get_authenticator` (it depends if in http, they send back a bearer token or username/password credentials). It would be technically more correct to only attach a challenge when we're using Authenticator::Oidc instead of AuthenticatorKind::Oidc, however I also don't think it matters too much? What do you think?

SangJunBak: Furthermore, this also means when using Authenticator::Password, we'll get the OAUTH challenge header if they don't pass in any credentials in the request but wanna use our towersession/login flow.  I don't think this is too big of a deal however as long as it doesn't trigger some dialog box on the browser.

SangJunBak (changes requested):

src/environmentd/src/http.rs
bobbyiliev: Rewrote the doc to spell it out!

src/environmentd/src/http.rs
bobbyiliev: Done. Merged both into `McpOAuthConfig` with a `scope()` method. Also the `http_host_name` no longer threads through `auth_middleware`/`http_auth`, since the middleware reads it from the same extension.

src/environmentd/src/http/oauth_metadata.rs
bobbyiliev: I think `AuthenticatorKind` is right here. RFC 9728 advertises what the resource accepts, not what this request used. If the listener accepts OIDC, the discovery doc and Bearer challenge should reflect that even when the caller picked Basic. Happy to revisit though!

Bearer does go out alongside Basic. Browsers only pop a dialog for Basic, so this should be silent to the tower-sessions flow. Worst case curl sees an extra header. Lmk if you think it should be gated tighter!

SangJunBak: Makes sense to me! Thanks for the thorough response

SangJunBak (approved):
LGTM

PR #36939, opened 2026-06-09
[environmentd] skip group_claim lookup for non-Frontegg auth

mtabebe:
The HTTP auth path unconditionally called group_claim_for, which issues a round-trip to the coordinator.

Only the Frontegg arm consumes the result, so unauthenticated probes like `/api/livez`, `/api/readyz`, and `/metrics` coupled liveness to coordinator health. Gate the fetch on `Authenticator::Frontegg` to mirror the pgwire path.

Add a regression test in `test/http-auth` that asserts liveness probes do not round-trip the coordinator.

Closes SQL-343

### DIFF
--- a/test/http-auth/mzcompose.py
+++ b/test/http-auth/mzcompose.py
@@ -98,3 +98,33 @@ def workflow_default(c: Composition) -> None:
                 "permission denied"
                 in r.json()["results"][0]["error"]["message"].lower()
             )
+
+    check_livez_coordinator_coupling(c)
+
+
+def check_livez_coordinator_coupling(c: Composition) -> None:
+    """Regression test for SQL-343: previously having group_claim_for above
+    the authenticator match in http_auth, so unauthenticated internal
+    endpoints (/api/livez, /api/readyz, /metrics) began round-tripping
+    the coordinator (`Command::GetSystemVars`) on every request for a value
+    only the Frontegg arm uses.
+    """
+    c.up("materialized")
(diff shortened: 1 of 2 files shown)


test/http-auth/mzcompose.py
```diff
@@ -98,3 +98,33 @@ def sql_user(url, query, user=None):
...
+    the authenticator match in http_auth, so unauthenticated internal
+    endpoints (/api/livez, /api/readyz, /metrics) began round-tripping
+    the coordinator (`Command::GetSystemVars`) on every request for a value
+    only the Frontegg arm uses.
+    """
+    c.up("materialized")
+    base = f"http://localhost:{c.port('materialized', 6878)}"
+
+    # Coordinator-side count of GetSystemVars commands, scraped from /metrics.
+    def get_system_vars_count() -> int:
```
SangJunBak, line 116: It does feel a bit brittle to assert on the metrics endpoint. e.g. if there's some background task that does this coordinator call in the future unrelated to the fetches, then this test would fail. However I can't think of a better way to test this regression! I do know instead of fetching the metrics api, tests in `server.rs` dependency inject the metrics registry and assert against that. But testing it via mzcompose seems fine too!

SangJunBak (approved):

test/http-auth/mzcompose.py
mtabebe: I agree, I think it is best to just leave it and then if it ends up being flaky, fix it?

PR #36940, opened 2026-06-09
[sql] fix REVOKE/GRANT ALL on TABLE <view/mv/source> losing non-SELECT privileges

mtabebe:
plan_update_privilege computed the ACL mode against the object's actual type (View/MaterializedView/Source) before remapping the reference type to TABLE.

Compute the ACL mode against the (post-remap) reference type so that the TABLE syntax expands `ALL` to the full TABLE privilege set (SELECT|INSERT|UPDATE| DELETE), matching PostgreSQL semantics. Update the SLT case that previously encoded the buggy behavior, and add a regression block exercising view/mv/ source × GRANT ALL / REVOKE ALL / REVOKE ALL PRIVILEGES.

Update the Privileges platform check to spell out the privileges explicitly (`INSERT, SELECT, UPDATE, DELETE`) for views, materialized views, and sources instead of `ALL PRIVILEGES`. The check previously used `GRANT ALL ON TABLE` against those object types and expected only SELECT to be granted, i.e. it encoded the bug.

Fixes SQL-232

### DIFF
--- a/misc/python/materialize/checks/all_checks/privileges.py
+++ b/misc/python/materialize/checks/all_checks/privileges.py
@@ -39,6 +39,13 @@ class Privileges(Check):
         return s
 
     def _grant_privileges(self, role: str, i: int, expensive: bool = False) -> str:
+        # For views, materialized views, and sources, the explicit privilege
+        # list is intentional. SQL-232 fixed `GRANT/REVOKE ALL ON TABLE
+        # <view/mv/source>` to expand to the full TABLE set (arwd) instead of
+        # only SELECT. Using `ALL PRIVILEGES` here would yield different
+        # `SHOW GRANTS` output across pre- and post-fix versions, breaking
+        # upgrade checks. Spelling out the privileges keeps the check
+        # version-stable.
         s = dedent(f"""
             $ postgres-execute connection=postgres://materialize@${{testdrive.materialize-sql-addr}}
             GRANT ALL PRIVILEGES ON DATABASE privilege_db{i} TO {role}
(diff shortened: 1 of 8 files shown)


misc/python/materialize/checks/all_checks/privileges.py
```diff
@@ -47,13 +47,13 @@ def _grant_privileges(self, role: str, i: int, expensive: bool = False) -> str:
             GRANT ALL PRIVILEGES ON CONNECTION privilege_csr_conn{i} TO {role}
             GRANT ALL PRIVILEGES ON TYPE privilege_type{i} TO {role}
             GRANT ALL PRIVILEGES ON TABLE privilege_t{i} TO {role}
-            GRANT ALL PRIVILEGES ON TABLE privilege_v{i} TO {role}
-            GRANT ALL PRIVILEGES ON TABLE privilege_mv{i} TO {role}
+            GRANT INSERT, SELECT, UPDATE, DELETE ON TABLE privilege_v{i} TO {role}
+            GRANT INSERT, SELECT, UPDATE, DELETE ON TABLE privilege_mv{i} TO {role}
```
ggevay, line 58: Could you please add a comment to explain that explicit privilege lists are used instead of ALL PRIVILEGES so results are identical across pre-/post-fix versions in upgrade scenarios. Without it, someone "simplifying" back to ALL PRIVILEGES would pass same-version checks but break upgrade checks confusingly.

test/sqllogictest/privilege_grants.slt
```diff
@@ -1131,6 +1131,120 @@ false
...
+----
+s  joe=arwd/materialize
+
+statement ok
+REVOKE ALL PRIVILEGES ON TABLE s FROM joe
+
+query TT
+SELECT name, privilege FROM item_privileges WHERE name = 's' AND privilege LIKE 'joe=%'
+----
+
```
ggevay, line 1247: Maybe also add a test for `GRANT/REVOKE ALL ON ALL TABLES IN SCHEMA` affecting views/MVs/sources correctly.

ggevay:
There is a minor issue that after the PR `GRANT/REVOKE ALL ON TABLE <view>` will emit the `NonApplicablePrivilegeTypes` notice. I think it would be good to suppress that somehow.

Btw. we might want to also update the user docs: doc/user/content/sql/grant-privilege.md:16-18 says ALL expands to all _applicable_ privileges for the object type — after this fix, for views/MVs/sources via TABLE syntax it expands to the full table set (arwd), most of which isn't _applicable_. The view/MV/source tab also documents the privilege choice as <SELECT | ALL [PRIVILEGES]> even though INSERT/UPDATE/DELETE are accepted.

And a further glitch found by Claude:

Default privileges still diverge from Postgres. ALTER DEFAULT PRIVILEGES … GRANT ALL ON TABLES + CREATE VIEW yields the full set in Postgres (arwdDxtm) but only r in Materialize, because get_applicable_privileges (src/catalog/src/memory/objects.rs) masks to the actual object type. Pre-existing and deliberate, not regressed by this PR — but empirically confirmed against your Postgres, so worth filing if PG parity is the goal.

ggevay (approved):
Thank you!

PR #36941, opened 2026-06-09
[mz-deploy] declarative SQL project deployment tooling

sjwiesman:
## Motivation

This PR introduces `mz-deploy`, a CLI for managing Materialize projects declaratively. You define your clusters, roles, network policies, connections, secrets, and models as a directory of `.sql` files; `mz-deploy` compiles them into a deployment plan, diffs that plan against the live environment, and executes blue/green schema migrations using Materialize's zero-downtime deployment primitives.

The goal is safe, testable, repeatable deployments: the project source is the source of truth, changes are type-checked and planned before they touch an environment.

  ## What's included

  The work lands in three commits:

  1. **The `mz-deploy` crate** (`src/mz-deploy`).

  2. **mzcompose integration test** (`test/mz-deploy`)

  3. **VS Code extension** (`misc/vscode-ext`)
  
  ## Testing

  Adds the `test/mz-deploy` mzcompose integration suite with example projects
  covering deploys, multi-profile resolution, connection/secret updates, system
  dependencies, and error cases. The crate's unit tests cover the compiler,
  client, and type-contract logic.

  ## Notes

The crate is `publish = false` and not wired into any release artifact, so there are no user-facing release notes for this change. Publishing will come as a follow on PR.


sjwiesman:
@def- Could you take a look at the second commit and just help ensure mzcompose is wired up properly to run in CI.

ci/test/pipeline.template.yml
```diff
@@ -291,6 +291,18 @@ steps:
...
+    label: "mz-deploy"
+    depends_on: build-x86_64
+    timeout_in_minutes: 30
+    inputs: [src/mz-deploy, test/mz-deploy]
+    plugins:
+      - ./ci/plugins/mzcompose:
+          composition: mz-deploy
+          ci-builder: stable
+    agents:
+      queue: hetzner-x86-64-8cpu-16gb
```
def-: Why not 4cpu-8gb? Edit: I see, it's currently building, added another comment about that.

ci/test/pipeline.template.yml
```diff
@@ -291,6 +291,18 @@ steps:
     agents:
       queue: hetzner-x86-64-dedi-32cpu-128gb-cargo-test
 
+  - id: mz-deploy
```
def-: Please add this towards the end, not beginning. At least I am used to the usual order of tests so it's nice to not move things around too much. I also assume this test will run pretty quickly, so being at the end doesn't hurt.

test/mz-deploy/mzcompose.py
```diff
@@ -0,0 +1,1355 @@
...
+        f.write(
+            f'[admin]\nhost = "127.0.0.1"\nport = {mz_system_port}\nusername = "mz_system"{rbac_options}\n\n'
+            f'[default]\nhost = "127.0.0.1"\nport = {mz_port}\nusername = "deploy_user"{rbac_options}\n\n'
+            f'[staging]\nhost = "127.0.0.1"\nport = {mz_port}\nusername = "deploy_user"{rbac_options}\n\n'
+            f'[dev]\nhost = "127.0.0.1"\nport = {mz_port}\nusername = "dev_user"{rbac_options}\n\n'
+            f'[monitor]\nhost = "127.0.0.1"\nport = {mz_port}\nusername = "monitor_user"{rbac_options}\n'
+        )
+
+
+def run_mz_deploy(
```
def-, line 111: I'm not happy with having it build for each run, that will take a lot of time. Why don't we build mz-deploy as a mzbuild-based container and use that?

def- (changes requested):

def- (commented):
Also, if it's ok, would you care to look at the QA LLM review? It seems t o have some good points. It would also be a good proving ground for the new mz-deploy mzcompose test to see if it can easily construct a regression test for each of those issues:

## FINDING 1 — HIGH: Simple-CTE scope handling diverges from Materialize, mis-classifying a real-table reference as a CTE reference (silently-wrong SQL / failed deploy)

**Location:**
- `src/mz-deploy/src/project/resolve/normalize/visitor.rs:306-311` (`visit_query_mut`)
- `src/mz-deploy/src/project/resolve/cte_scope.rs:20-24` (incorrect justifying comment) — the same wrong claim also remains in `src/mz-deploy/src/project/analysis/deps.rs:21-25`
- `src/mz-deploy/src/project/resolve/normalize/tests.rs` (`test_cte_shadowing_external_table` locks in the wrong behavior)

**The bug.** `visit_query_mut` pushes **all** CTE names of a `WITH` block into scope
*before* visiting any CTE body:

```rust
fn visit_query_mut(&mut self, node: &mut Query<Raw>) {
    let names = CteScope::collect_cte_names(&node.ctes);   // ALL names (Simple + MutRec)
    self.cte_scope.push(names);
    visit_mut::visit_query_mut(self, node);                // visits ctes, then body
    self.cte_scope.pop();
}
```

The justifying comment asserts "self-references in simple CTEs are SQL errors that
Materialize rejects" — but Materialize's own resolver (`fold_query` in
`src/sql/src/names.rs:1665-1683`, mirrored by `plan_ctes` in
`src/sql/src/plan/query.rs`) resolves each `CteBlock::Simple` body **before**
inserting that CTE's name into scope. So in valid, accepted SQL such as:

```sql
CREATE VIEW v AS
WITH products AS (SELECT id FROM products WHERE active)   -- inner `products` = the real table
SELECT * FROM products;                                   -- outer `products` = the CTE
```

the inner `products` resolves to the **catalog object**. mz-deploy does the
opposite: the inner reference is treated as a CTE reference and left un-rewritten.

**Impact.**
- *Type-check path (fail-closed):* the bare inner reference can't be resolved against
  the flat-named type-check catalog → a valid model fails to compile/deploy.
- *Staging (blue/green) and dev-overlay paths (fail-open, silent):* the inner
  reference should be rewritten to the staging/overlay copy of the upstream; left
  un-rewritten it silently points at **production**, so staged objects are
  validated/hydrated against the wrong upstream — defeating the tool's core safety
  guarantee.

The common idiom `WITH x AS (SELECT ... FROM x WHERE ...)` is enough to trigger it.
The existing test `test_cte_shadowing_external_table` actively enforces the buggy
behavior, so CI will not catch this.

**Suggested fix.** Mirror `fold_query`: introduce simple-CTE names *incrementally* —
for `CteBlock::Simple`, visit CTE *i*'s body with only CTEs `0..i` in scope, then
push CTE *i*'s name; only `MutuallyRecursive` blocks push all names up front. Correct
the comments in `cte_scope.rs` and `analysis/deps.rs`, and fix
`test_cte_shadowing_external_table` to assert the inner reference *is* qualified.

---

## FINDING 2 — MEDIUM: `ObjectId` keys parsed references with `Ident::to_string()` (quoted) but the deploy set with bare strings, so quoting-requiring object names are mis-classified as external and not staged

**Location:**
- `src/mz-deploy/src/project/ir/object_id.rs:124-159` (`from_item_name` — root cause)
- `src/mz-deploy/src/project/ir/compiled.rs:126-130` (`with_names` — the bare-keyed side)
- `src/mz-deploy/src/project/resolve/normalize/transformers.rs` (`is_external`, the consumer)

**The bug.** The deploy set keys `ObjectId.object` with the bare name (e.g. `order`),
while parsed references use `Ident`'s `AstDisplay`, which adds double quotes for
uppercase names, special characters, and **reserved keywords** (`order`, `user`,
`table`, …). `{object: "order"}` ≠ `{object: "\"order\""}`, so
`objects_to_deploy.contains(...)` is `false`, `is_external` returns `true`, and the
reference is **not suffixed** for staging — it silently resolves to the
**production** object rather than the staged copy. The same mismatch affects
`external_dependencies` / `replacement_objects` membership.

**Impact.** For any deployed object whose name requires quoting (a model file named
`order.sql` suffices — plain lowercase ASCII), intra-deployment references are not
rewritten during staging, so the staged graph reads from production, undermining
blue/green validation.

**Suggested fix.** Key `ObjectId` consistently on the bare identifier value: use
`ident.as_str().to_string()` in `from_item_name`/`from_raw_item_name` (and for the
database/schema components, and in `analysis/deps.rs`).

---

## FINDING 3 — MEDIUM: Staging-object discovery uses an unescaped `_` (and unvalidated `deploy_id`) in a SQL `LIKE` pattern, over-matching objects that are then `DROP … CASCADE`d by `abort`/rollback

**Location:**
- `src/mz-deploy/src/client/introspection.rs:659-700` (`get_staging_schemas`, `get_staging_clusters`)
- `src/mz-deploy/src/client/deployment_ops.rs` (hydration readiness gate)
- Reached destructively from `src/mz-deploy/src/cli/commands/abort.rs` and the stage-failure rollback, via `drop_staging_schemas`/`drop_staging_clusters` (`DROP SCHEMA/CLUSTER … CASCADE`).

**The bug.** `pattern = format!("%_{}", deploy_id)` is bound as a `LIKE` pattern, but
`_` is the single-character wildcard, so it means "ends with `<any-char><deploy_id>`",
and `deploy_id` is not charset-validated (the CLI-help "alphanumeric, hyphens,
underscores" rule is documentation only), so a `deploy_id` containing `_` or `%`
injects further wildcards.

**Impact.** With a custom `--deploy-id prod`, `abort prod` builds pattern `%_prod`,
which matches a production schema named `analytics_prod` and `DROP SCHEMA … CASCADE`s
it. A `deploy_id` containing `%` matches almost everything → mass cascade drop.
Default auto-generated ids are hex/git-sha, so the exposure is custom deploy ids.

**Suggested fix.** Escape LIKE metacharacters and anchor the separator
(`... LIKE $1 ESCAPE '\'` with `%\_<escaped-id>`), or avoid `LIKE` with
`right(name, length($1)) = $1`. Independently, enforce the documented
`^[A-Za-z0-9_-]+$` charset on `deploy_id` at the CLI boundary.

---

## FINDING 4 — MEDIUM: `trim_end_matches(staging_suffix)` strips the suffix repeatedly, computing the wrong production name when a prod schema/cluster legitimately ends in `_<deploy_id>` (wrong cutover swap / wrong CASCADE drop)

**Location:** `src/mz-deploy/src/cli/commands/promote.rs:100, 167, 401, 697, 732, 751, 957, 981, 999, 1059` — all recover a production name from a staging name via `trim_end_matches(&staging_suffix)`.

**The bug.** `str::trim_end_matches` strips **all** trailing occurrences. With
`deploy_id = prod` and production schema `analytics_prod`, staging is
`analytics_prod_prod`, and recovery yields `analytics` (stripped twice) instead of
`analytics_prod`.

**Impact.** The cutover runs `ALTER SCHEMA ... "analytics" SWAP WITH
"analytics_prod_prod"` — the wrong production schema. If `analytics` exists, the
swap silently cuts over into it, and `drop_old_resources` then `DROP … CASCADE`s the
real production schema → **data loss**. Same defect for cluster names and the
dependent-sink repoint target. Exposure is custom `deploy_id`s coinciding with an
existing production name suffix.

**Suggested fix.** Strip exactly once:
`x.strip_suffix(staging_suffix).unwrap_or(x)` at every listed site.

test/mz-deploy/mzcompose.py
sjwiesman: Did not realize that's what was happening! Will fix

sjwiesman:
@def- QA LLM review was great. Fixed the issues.

def-:
## Finding 1 (MEDIUM): LIKE-escaping fix applied incompletely — hydration-status queries still use unescaped `%_<deploy_id>` patterns

The delta fixes the unescaped-LIKE over-matching bug in
`src/mz-deploy/src/client/introspection.rs`, but the identical pattern remains in
`src/mz-deploy/src/client/deployment_ops.rs`:

- `get_deployment_hydration_status` — `deployment_ops.rs:990`:
  `let pattern = format!("%_{}", deploy_id);` feeding `hydration_status_query`, which
  uses plain `c.name LIKE $1` in three CTEs (`deployment_ops.rs:912,924,940`) with no
  `ESCAPE` clause.
- `subscribe_deployment_hydration` — `deployment_ops.rs:1537`: same unescaped pattern
  inside a `SUBSCRIBE`.

Since `_` is a single-character LIKE wildcard, deploy id `prod` produces pattern
`%_prod`, which matches any cluster name ending in `<any char>prod` — e.g. an unrelated
production cluster `dataprod` — not just names ending in the literal `_prod`. A deploy
id containing `%` would match nearly every cluster.

These queries drive `mz-deploy wait` (cli/commands/wait.rs:92,150,208,223,449,512) and
`mz-deploy list` — the readiness gate before promotion. Over-matching sweeps unrelated
clusters into the deployment-status computation: an unrelated cluster that is
`hydrating`/`lagging`/`failing` (or merely replica-less) blocks `wait` indefinitely or
reports bogus failure for a healthy deployment. The direction is fail-safe (a superset
can only make readiness stricter, never falsely ready), and unlike the fixed
introspection.rs queries nothing here feeds a `DROP`, hence MEDIUM rather than HIGH.

**Fix**: hoist `staging_suffix_like_pattern` out of `introspection.rs` (e.g. to
`pub(super)` in the client module), use it in `get_deployment_hydration_status` and
`subscribe_deployment_hydration`, and add `ESCAPE '\'` to the three `LIKE $1`
predicates in `hydration_status_query`.

---

## Finding 2 (MEDIUM): `--no-docker` test mode — per-test `DISCARD ALL` resets the shared session, dropping the `_mz_deploy_server` cluster pin after the first test

New in this delta (`1bf4b7024b`): `mz-deploy test --no-docker` runs all unit tests
through a **single** shared connection created by
`Client::connect_with_profile(settings.connection().clone())`
(cli/commands/test.rs, `run_tests`). Two facts collide:

1. `connect_with_profile` pins the session cluster to `_mz_deploy_server` via the libpq
   startup `options` parameter (client/connection.rs:145–150). Pgwire applies startup
   options as plain `session.vars_mut().set(..., LOCAL=false)` — *not* as session
   defaults (src/pgwire/src/protocol.rs:557–571).
2. `run_single_test` ends every test with `DISCARD ALL` (test.rs:~590), which maps to
   `session.reset()` → `vars.reset_all()` — resetting **every** session var to its
   default (src/adapter/src/coord/sequencer.rs:547–552, src/adapter/src/session.rs:825).

On the default Docker path this was harmless: each test gets a fresh client from
`runtime_client`, so the session dies after the `DISCARD ALL` anyway. On the new shared
profile connection, the first test's `DISCARD ALL` permanently strips the
`cluster=_mz_deploy_server` pin and any profile-configured `options` (e.g. the
`enable_session_rbac_checks` used in the mzcompose profiles). Result:

- Test 1 runs on `_mz_deploy_server`; tests 2..N run on the server-default cluster
  (typically `quickstart`). In real regions where the default cluster has been dropped,
  every test after the first fails with "unknown cluster"; where it exists, the test
  workload silently lands on the user's default (often production-serving) cluster.
- It also contradicts the documented contract on `connect_with_profile` itself: "The
  unit-test runtime uses `connect_with_profile_no_pin` instead."

The mzcompose CI run won't notice: `quickstart` exists there and the asserted tests
still produce the same pass/fail names.

**Fix**: give each test its own connection in `--no-docker` mode (mirroring the Docker
path's per-test `runtime_client`), or keep the shared connection but replace
`DISCARD ALL` with `DISCARD TEMP` (which drops the temp views without resetting session
vars). The per-test-connection variant also fixes Finding 3.

---

## Finding 3 (MEDIUM): `--no-docker` — setup-statement failure skips cleanup, leaking temp views that cascade false failures across the rest of the suite

In `run_single_test` (cli/commands/test.rs), the setup loop returns early on any failed
statement:

```rust
for (i, sql) in sql_statements[..sql_statements.len() - 1].iter().enumerate() {
    if let Err(e) = client.batch_execute(sql).await {
        ...
        return Ok(TestOutcome::Failed(ExecutionFailure::Error(...)));  // no DISCARD ALL
    }
}
```

`DISCARD ALL` only runs on the fall-through path after the assertion query. Every test
lowers to `CREATE TEMPORARY VIEW` statements with **fixed names**: mocks are named by
the mocked object's FQN (e.g. `"materialize.public.users"`) and the expected-rows view
is literally named `expected` (cli/commands/test/lower.rs:689–775). On the Docker path
the leak was unobservable — the session (and its temp schema) died with the per-test
client. On the new shared `--no-docker` connection, one test with a SQL error in its
setup (a malformed mock, a type mismatch in EXPECTED — routine for a test tool) leaves
its temp views behind, and every subsequent test then fails at
`CREATE TEMPORARY VIEW expected` with "view already exists". One genuinely broken test
poisons the remainder of the suite with spurious failures (never spurious passes — the
collision aborts before the assertion).

CI won't catch this: the mzcompose `unit-tests` case (`test/mz-deploy/mzcompose.py`)
only exercises assertion-failure tests (`*_wrong_expectation`, `*_wrong`), which do
reach `DISCARD ALL`.

**Fix**: run the cleanup on all exit paths of `run_single_test` (e.g. a small
`async`-drop guard or restructuring so `DISCARD TEMP`/`DISCARD ALL` executes before
every `return` that follows the first `CREATE`), or adopt the per-test connection from
Finding 2 which makes leaks structurally impossible.

sjwiesman:
My LLM has addressed the concerns discovered by your LLM 🙇

maheshwarip (approved):

sjwiesman:
@def- any other concerns?

def- (approved):
It's 74k lines, so slightly concerned! But don't let me stand in the way of progress

def-:
Turns out that rerunning the QA LLM review yet another time, even without changes, finds different bugs (I also made some changes to the QA LLM review script, so that could also be related):

## Finding 1 — HIGH: GRANT/COMMENT-only changes on views & materialized views are silently ignored

**Files:**
- `src/mz-deploy/src/project/analysis/deployment_snapshot.rs:157` (`compute_typed_hash`)
- `src/mz-deploy/src/project/analysis/changeset/diff.rs:32` (`find_changed_objects`)
- `src/mz-deploy/src/cli/commands/stage.rs:431` (`select_stage_objects`), `:1182` (`deploy_single_object`)

### What's wrong
The incremental diff that decides which objects to (re)deploy is driven purely by a
content hash. `compute_typed_hash` hashes only `db_obj.stmt` and `db_obj.indexes`:

```rust
db_obj.stmt.hash(&mut hasher);
// ... sort + hash indexes ...
for index in &indexes { index.hash(&mut hasher); }
```

But `compiled::DatabaseObject` (`ir/compiled.rs:411`) also carries
`grants: Vec<GrantPrivilegesStatement>` and `comments: Vec<CommentStatement>`, which are
**not** part of the hash. The diff (`find_changed_objects`) compares only these hashes;
its own doc comment even states that comment edits "do not produce different hashes and
therefore do not appear in the diff."

For **views and materialized views** (the compute objects that flow through the hash-based
incremental diff), grants and comments are applied **only** inside `deploy_single_object`
(`stage.rs:1214-1232`), which runs **only** for objects in `change_set.objects_to_deploy`
(= changed objects). Grants are applied *additively* on the freshly-staged object — there
is no separate full reconcile pass for views/MVs (the `grants::reconcile` path runs only
for tables/sources/secrets/connections/clusters/network-policies — confirmed by grepping
all call sites).

### Consequence
If an operator edits **only** the `GRANT …` or `COMMENT …` statements of an existing
view/MV `.sql` file (without touching the view definition), the object's hash is unchanged,
it is excluded from `objects_to_deploy`, and the grant/comment change is **never applied** —
yet `mz-deploy` reports the deployment as successful.

- Adding a grant (`GRANT SELECT ON v TO new_role;`) → silently not applied; `new_role`
  still cannot read `v`. (Fails closed.)
- **Removing a grant to revoke access** → silently not revoked; the privilege persists even
  though the operator believes the declarative source-of-truth removed it. (**Fails open —
  security relevant.**)

This directly defeats the crate's stated purpose ("the project source is the source of
truth"). It is trivially and commonly triggered (granting access to an existing view is a
routine operation).

### Suggested fix
Include `grants` and `comments` in `compute_typed_hash`, deterministically sorted exactly
as `indexes` already are (e.g. sort by rendered `to_string()`, then `hash`). That makes a
grant/comment-only edit produce a changed hash so the object enters `objects_to_deploy` and
its grants/comments are re-applied through the normal staging path.

---

## Finding 2 — HIGH: Resuming `promote` after the swap commits silently skips sink-repointing and old-resource cleanup, with no recovery path

**Files:**
- `src/mz-deploy/src/cli/commands/promote.rs:393-398` (`generate_deployment_plan`, PostSwap branch)
- `src/mz-deploy/src/cli/commands/promote.rs:786-809` (`execute_atomic_swap` — state marker flip is inside the swap txn)
- `src/mz-deploy/src/cli/commands/promote.rs:530-547` (`run_post_swap_steps` — the skipped gates)

### What's wrong
The apply-state marker is flipped to `PostSwap` **inside** the swap transaction
(`execute_atomic_swap` swaps the `apply_<id>_pre`/`apply_<id>_post` state schemas in the same
`BEGIN`/`COMMIT` as the schema and cluster swaps, lines 785-809). So the moment the
production cutover commits, the persisted `apply_state` is already `PostSwap`.

On a resume (re-invoking `promote <id>` after a crash / `Ctrl-C` / OOM / network blip / deploy
timeout between the swap `COMMIT` and the end of cleanup), `generate_deployment_plan` takes the
`PostSwap` branch and hard-codes **empty** staging sets:

```rust
let (staging_schemas, staging_clusters, staging_suffix) = if apply_state == ApplyState::PostSwap {
    verbose!("Resuming post-swap: skipping conflict check and resource gathering");
    (BTreeSet::new(), BTreeSet::new(), format!("_{}", deploy_id))   // <-- both empty
} else { gather_resources_and_check_conflicts(...).await? };
```

`run_post_swap_steps` then gates the two remaining production-cutover steps on those sets:

```rust
if !plan.staging_schemas.is_empty() { repoint_dependent_sinks(client, plan).await?; }     // SKIPPED
client.deployments().update_promoted_at(&plan.deploy_id).await?;                          // marks promoted
if !plan.staging_schemas.is_empty() || !plan.staging_clusters.is_empty() {
    drop_old_resources(client, plan).await;                                               // SKIPPED
}
```

`execute_pending_sinks` and `apply_replacement_mvs` *do* resume correctly (they use their own
persisted vectors and are idempotent — `execute_pending_sinks` skips already-existing sinks via
`check_sinks_exist`), so the resume runs cleanly to `update_promoted_at` + `cleanup_apply_state`
and reports **"Deployment completed successfully!"** A subsequent invocation then fails
`validate_staging` with `DeploymentAlreadyPromoted`. There is **no recovery path**.

### Consequence
A crash in the (multi-round-trip, network-bound) window after swap-commit but before
cleanup completes leads to a "successful" promote in which:

1. **`repoint_dependent_sinks` never runs.** Existing sinks that depend on the swapped-out
   objects are not repointed to the promoted objects. After `ALTER SCHEMA … SWAP`, the old
   objects now live in `<schema>_<deploy_id>`; the sink keeps reading those stale, no-longer-
   maintained definitions and silently emits diverging data to its external destination —
   permanently.
2. **`drop_old_resources` never runs.** The old production schemas (`<schema>_<deploy_id>`)
   and clusters (`<cluster>_<deploy_id>`) leak permanently (wasted compute), and will collide
   with the next deployment that reuses the same deploy id.

This directly contradicts the design's own documented failure-mode contract
(`doc/developer/design/20260610_mz_deploy.md:954,956`: "promote after swap, before cleanup →
rerun `promote <id>` — post-swap steps retried"; "old-world drop failure → rerun `promote <id>`").

### Suggested fix
In the `PostSwap` branch of `generate_deployment_plan`, re-derive `staging_schemas` and
`staging_clusters` from the still-persisted deployment records (the same
`get_schema_deployments` / `get_deployment_clusters` used to build the marker before swap)
instead of zeroing them. Both `repoint_dependent_sinks` and `drop_old_resources` are already
written to be safe to re-run (`find_sinks_depending_on_schemas` returns nothing once the old
schemas are gone; `DROP … IF EXISTS` is idempotent), so retrying them on resume is correct.
Move `update_promoted_at`/cleanup to occur only after both steps succeed.

---

## Finding 3 — MEDIUM: `abort` and stage-rollback drop schemas/clusters by name suffix across all databases, with no scoping to the deployment — can drop unrelated production objects

**Files:**
- `src/mz-deploy/src/cli/commands/abort.rs:87-126`
- `src/mz-deploy/src/cli/commands/stage.rs` (`rollback_staging_resources`)
- `src/mz-deploy/src/client/introspection.rs:658-700` (`get_staging_schemas`/`get_staging_clusters`), `:819-852` (`drop_staging_schemas`/`drop_staging_clusters`)
- `src/mz-deploy/src/client.rs:79` (`staging_suffix_like_pattern`)

### What's wrong
`get_staging_schemas` / `get_staging_clusters` select **every** `mz_schemas` / `mz_clusters`
row whose name matches `LIKE '%\_<deploy_id>' ESCAPE '\'` — with **no** join to
`_mz_deploy` deployment records and **no** database qualification — then feed the results
straight into `DROP SCHEMA … CASCADE` / `DROP CLUSTER … CASCADE`:

```rust
let pattern = staging_suffix_like_pattern(deploy_id);  // "%\_<deploy_id>"
// SELECT … FROM mz_schemas s JOIN mz_databases d … WHERE s.name LIKE $1 ESCAPE '\'
```

The escape pattern correctly avoids matching `fooprod` for deploy id `prod`, but it still
matches any object whose name ends in the *literal* `_<deploy_id>`. `abort` only checks that
the deployment row exists and is unpromoted; it never constrains the drop set to objects
actually recorded for the deployment.

### Consequence
With an operator-chosen deploy id that collides with a real naming suffix (e.g. `prod`,
`v2`, `blue`, `green`, `release`), a pre-existing, unrelated **production** schema literally
named e.g. `metrics_prod` / `analytics_v2` (in *any* database) or a production cluster
`etl_prod` is matched and, on `mz-deploy abort <id>` or an automatic stage rollback, issued a
`DROP SCHEMA "<db>"."metrics_prod" CASCADE` / `DROP CLUSTER "etl_prod" CASCADE`. That is
production data loss / production compute teardown for an object that was never part of the
deployment.

This is the exact naming-collision hazard that `promote.rs` defends against elsewhere
(`strip_staging_suffix`, escaped LIKE), but abort/rollback do not intersect the LIKE match
with the deployment's recorded production schemas/clusters.

### Suggested fix
Constrain the drop set to schemas/clusters actually recorded for the deploy id (intersect the
LIKE result with `get_schema_deployments` + `get_deployment_clusters` for that deployment, or
record exact staging names at stage time and drop only those).

---

## Finding 4 — MEDIUM: `deploy_id` is never validated, yet interpolated raw (`"{}"`) into swap/drop SQL in `promote.rs` (inconsistent with the rest of the crate)

**Files:**
- `src/mz-deploy/src/bin/mz-deploy/main.rs:300-302` (doc claims validation that doesn't exist)
- `src/mz-deploy/src/cli/commands/stage.rs:201-203` (`staging_suffix = format!("_{}", stage_name)`, no validation)
- `src/mz-deploy/src/cli/commands/promote.rs:752, 770, 786-791, 929-938, 1050-1058, 1093, 1109` (raw `"{}"` interpolation)

### What's wrong
The `--deploy-id` / `--name` argument's doc comment says *"Must contain only alphanumeric
characters, hyphens, and underscores,"* but there is **no `value_parser` and no runtime
validation anywhere** (confirmed by grep — only the doc comment exists). The value flows into
`staging_suffix = format!("_{}", stage_name)` and into catalog object names.

The whole crate consistently uses `quote_identifier` (which wraps in `"…"` and doubles
embedded `"`) to build SQL — *except* `promote.rs`, which builds every swap/drop/replacement
statement with raw double-quote interpolation, embedding `deploy_id` directly or via the
staging suffix, e.g.:

```rust
let pre_schema = format!("apply_{}_pre", deploy_id);
let state_swap_sql = format!(
    "ALTER SCHEMA _mz_deploy.\"{}\" SWAP WITH \"{}\";", pre_schema, post_schema);   // promote.rs:786
// and: ALTER SCHEMA "{}"."{}" SWAP WITH "{}"  (:752), ALTER CLUSTER "{}" SWAP WITH "{}" (:770),
//      ALTER MATERIALIZED VIEW … APPLY REPLACEMENT … (:929), ALTER SINK … SET FROM … (:1050),
//      DROP SCHEMA "{}"."{}" CASCADE (:1093), DROP CLUSTER "{}" CASCADE (:1109)
```

Because `create_apply_state_schemas` builds the *same* names with `quote_identifier`
(`deployment_ops.rs:1059-1060`), a `deploy_id` containing `"` (or `.`) is created safely at
`stage` time but produces broken / boundary-crossing SQL at `promote` time.

### Consequence
A `deploy_id` containing `"` succeeds through `stage` (quote-safe) but corrupts the swap/drop
SQL at the dangerous mid-cutover moment, leaving a deployment that can be staged but never
cleanly promoted or aborted. `deploy_id` is operator-supplied (and auto-generated values are
safe hex), so this is a self-inflicted footgun rather than an external-attacker injection
vector — hence MEDIUM, not HIGH — but it is a real correctness/robustness gap made worse by
the false documented contract and the quoting inconsistency with the rest of the crate.

### Suggested fix
Either enforce the documented contract with a `value_parser` that rejects anything outside
`[A-Za-z0-9_-]`, or route every name in `promote.rs` through `quote_identifier` like the rest
of the crate. Doing both is cheap.

---

## Finding 5 — LOW: Sink output connections and Avro/CSR format connections are not extracted as dependencies

**File:** `src/mz-deploy/src/project/analysis/deps.rs:432-453` (`extract_dependencies` for `CreateSink`/`CreateSource`)

For a sink, only `s.from` and `s.in_cluster` are recorded as dependencies; the sink's output
connection (`CreateSinkConnection::Kafka { connection, .. }` / `Iceberg { catalog_connection,
aws_connection, .. }`) and the Avro/CSR schema-registry connection in
`format: Option<FormatSpecifier>` (for both sinks and sources) are never extracted.

Cross-type apply ordering is saved by the fixed phase order (connections → sources → … →
sinks), so this rarely breaks deploys. The concrete impact is a validation gap: an
external sink/format connection that is not declared in `project.toml` never appears in
`discovered`, so `validate_dependencies` (`deps.rs:519`) fails to flag it as an undeclared
dependency, and the failure surfaces later as a raw DB error at execution time instead of a
clean validation error. Low severity (degraded diagnostics, not data loss).

sjwiesman:
Will follow up with these. At least finding one was purposeful. Though open to reconsidering. Redeploying a whole mat view just to update a comment feels expensive.

PR #36942, opened 2026-06-09
compute/repr: add `ToDatumIter::extend_datums` and port fill sites

frankmcsherry:
Add `ToDatumIter::extend_datums(&self, target, max: Option<usize>)` with a default that matches the existing `target.extend(x.to_datum_iter()[.take(n)])` pattern (zero behavior change for `Row` and the `&T` blanket forwards through to the inner impl). Override it on row-spine's `DatumSeq` with a tight `read_datum` push loop, so the read-item representation is branched on once per row rather than once per datum (the `Iterator::extend`/`take` machinery).

Port the fitting fill sites to `extend_datums`: render context (key/val, with the `max_demand` cap), linear- and delta-join, reduce, and the peek-result iterator. Genuine non-fits left as-is: `reduce.rs`'s repeat-first-datum, `encode.rs`'s collect-for-assert.

Independent of dictionary compression; the dictionary-arrangements PR will add a codec-aware `DatumSeq` override on top.

### Motivation

Closes <issue>
Fixes <bug>
etc.

### Description

diff *can't* tell a reviewer.

### Verification

### DIFF
--- a/src/compute/src/compute_state/peek_result_iterator.rs
+++ b/src/compute/src/compute_state/peek_result_iterator.rs
@@ -214,16 +214,14 @@ where
         let arena = RowArena::new();
 
         let key_item = self.cursor.key(&self.storage);
-        let key = key_item.to_datum_iter();
         let row_item = self.cursor.val(&self.storage);
-        let row = row_item.to_datum_iter();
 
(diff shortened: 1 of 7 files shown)


antiguru (approved):
Checks out, thank you!

PR #36943, opened 2026-06-09
adapter: fix statement-logging double-end panics on concurrent DROP CLUSTER

ggevay:
### Motivation

A frontend-sequenced peek whose cluster is dropped concurrently could log its
statement-execution end twice, panicking the coordinator at
`end_statement_execution` and aborting environmentd. Two distinct races hit the
identical panic — one fast-path, one slow-path — both fixed here.

This seems to have happened in [#incident-1070](https://materializeinc.slack.com/archives/C0B8QU8FGCR).

### Description

Two commits, one per path — reviewable independently, apart from the two tests'
cross-references to each other.

A frontend peek must be the sole end-logger unless the coordinator already has a
genuine end of its own. The fast and slow paths break this differently — in one
the coordinator's extra end is real, in the other it is spurious — so each is
fixed differently:

- **Fast path**: `DROP CLUSTER` in the `RegisterFrontendPeek`→`client.peek()`
  window makes the coordinator retire the peek with a *real* end (`Canceled`)
  while the frontend also logs `Errored`. Fix: `UnregisterFrontendPeek` reports
  whether the peek was still registered; if already retired, the frontend skips
  logging (`end_already_logged`).
- **Slow path**: `implement_peek_plan` errors before taking its
  `ExecuteContextGuard`, so the guard's `Drop` emits a *spurious* `Aborted` on
  top of the frontend's `Errored`. Fix: defuse the guard on error.

Also documents `implement_peek_plan`'s `ctx_extra` ownership contract.

### Verification

Two new `test/cluster` workflows that fail on any environmentd abort: a
slow-path stress test, and a deterministic fast-path test using a new
`peek_after_register_before_issue` failpoint. Both pass with the fixes;
reverting the fast-path fix makes its workflow fail as expected.

Nightly: https://buildkite.com/materialize/nightly/builds/16727

### DIFF
--- a/src/adapter/src/command.rs
+++ b/src/adapter/src/command.rs
@@ -328,11 +328,16 @@ pub enum Command {
 
     /// Unregister a pending peek that was registered but failed to issue.
     /// This is used for cleanup when `client.peek()` fails after `RegisterFrontendPeek` succeeds.
-    /// The `ExecuteContextExtra` is dropped without logging the statement retirement, because the
-    /// frontend will log the error.
+    ///
+    /// Sends back whether the peek was still registered. If it was, its
+    /// `ExecuteContextExtra` is dropped without logging the statement
+    /// retirement, because the frontend will log the error. If it was *not*
+    /// still registered, then a concurrent teardown (e.g. a `DROP CLUSTER`)
+    /// already removed and retired it, which logged the end of execution; the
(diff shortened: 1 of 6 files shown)


ggevay:
Superseded by https://github.com/MaterializeInc/materialize/pull/36956

PR #36947, opened 2026-06-09
doc: Add design for scoped feature flags (per-cluster and per-replica)

antiguru:
### Motivation

This design document describes scoped feature flags in Materialize: per-cluster and per-replica LaunchDarkly overrides. It addresses two concrete use cases:

1. **Per-cluster optimizer flags**: clusters such as `mz_catalog_server` should be able to run with a different optimizer feature set without affecting the rest of the environment.
2. **Per-replica flags by size family**: replica size families (legacy t-shirt sizes, `D.1`, etc.) should support different configurations (e.g. legacy sizes keep `lgalloc`, while `D.1` enables the persist pager and LZ4 compression).

Today feature flags are evaluated once per environment, with no way to target a specific cluster or replica.

### Description

The design extends the existing LaunchDarkly integration with two new scope classes, resolved at boundaries that already carry the right context:

- **Cluster-coherent flags** — evaluated replica-free via a `cluster` context kind and applied at plan time, feeding `OptimizerFeatureOverrides` (not replacing it; `EXPLAIN WITH` still needs per-statement overrides). Replica-free evaluation makes incoherent per-replica plan divergence unrepresentable by construction.
- **Replica-local flags** — evaluated per replica via a `replica` context kind (carrying size, size family, and owning-cluster attributes), pushed to the compute controller's per-replica dyncfg layer (which reaches storage on `clusterd` via the shared persist client `ConfigSet`).

Key design decisions:

1. **Dual context kinds** (`cluster`, `replica`) to enforce the coherence boundary.
2. **Required scope declaration** on every synced parameter (`ParameterScope::{Environment,Cluster,Replica}`), driving evaluation, validation, and docs.
3. **Durable, id-keyed cache** written solely by the sync loop — two catalog collections (`cluster_system_configurations`, `replica_system_configurations`), so scoped values survive an `environmentd` restart and an LD outage. There is no user-facing `ALTER SYSTEM ... FOR CLUSTER` DDL. Because catalog ids are never reused, stale entries are inert and GC is hygiene, not correctness.
4. **Dual id/name attributes** on contexts: name/family predicates survive recreate (role intent), `*_id` pins die with the incarnation (patch intent).
5. **Differs-from-env-wide recording** (compared on the resolved/typed value, not the `variation_detail` reason), which keeps the collections sparse and lets a cluster-scoped LD value beat a manual `CREATE CLUSTER ... FEATURES` pin.
6. **Feature gate** `enable_scoped_system_parameters` (dyncfg, default `false`): strictly opt-in, checked at the single `sync_scoped_params` chokepoint, so default behavior is exactly today's environment-wide-only resolution.
7. **Synchronous create-time resolution**: a newly created cluster/replica's overrides are folded into the create transaction (so they persist synchronously and a fresh object's first configuration/plan already carries them), because render-frozen flags make the next-tick window a correctness problem.

The document also covers LaunchDarkly's billing model (server-side service connections, not MAU), self-managed behavior, worked examples, alternatives, and open questions.

### Status

The design has been **implemented and merged** across #37079, #37080, #36959, #37151, #37159, and #37158. This document has been verified against the merged implementation and reconciled with it (durable persistence, the catalog-implication-driven replica push, synchronous create-time persistence, and the shipped `ReplicaAllocation::family()` fallback), so it now serves as the up-to-date design record rather than a pre-implementation proposal.

https://claude.ai/code/session_01S9fiehWEbC4BEXEq8p7LP9

### DIFF
--- a/doc/developer/design/20260609_scoped_feature_flags.md
+++ b/doc/developer/design/20260609_scoped_feature_flags.md
@@ -227,8 +227,35 @@ overrides at plan time, at the (already cluster-aware) sequencing sites in
 
 Precedence for replica-local flags, lowest to highest:
 `Global < ReplicaSizeFamily < Replica(id)` (a specific replica pin beats a size
-family rule). Cluster-coherent flags resolve `Global < Cluster`. Interaction with
-manually-set `CREATE CLUSTER ... FEATURES` is covered in Open Questions.
+family rule).
+
+For cluster-coherent flags the ordering is **specificity-then-source**, lowest
+to highest:
+
+```
+env-wide LD  <  manual CREATE CLUSTER ... FEATURES  <  cluster-scoped LD
+```
+
+i.e. a more specific scope wins, and at equal specificity **LD beats manual**.
+This is consistent with the existing global behavior — LD already overrides a
+manual `ALTER SYSTEM SET` wherever LD serves a value (the sync loop rewrites the
+param every tick), and `FEATURES` is likewise a system-only operator knob, so it
+gets the same treatment. It also *preserves* today's behavior that a manual
+`FEATURES` override beats the env-wide defaults (confirmed by `override_from` in
+`src/repr/src/optimize.rs`): env-wide LD is not cluster-specific, so a deliberate
+per-cluster manual pin still beats it, while a cluster-specific *LD* rule beats
+the manual pin.
+
+Crucially this is decided **per feature, only where LD actually serves a value**,
+mirroring the global case (where a manual value survives exactly when LD is
+silent). The implementation uses the LD evaluation *reason* from
+`variation_detail` to distinguish "LD has an opinion" (`RULE_MATCH` /
+`TARGET_MATCH` / `FALLTHROUGH`) from "LD is silent" (`FLAG_NOT_FOUND` / error);
+where LD is silent, the manual `FEATURES` value stands.
+
+The accepted trade-off, stated plainly: an operator's manual `FEATURES` pin can
+be overridden by a cluster-scoped LD rule. That is the same bargain operators
+already accept for `ALTER SYSTEM`, so it is consistent rather than novel.
 
 ### Worked examples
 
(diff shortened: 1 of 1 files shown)


doc/developer/design/20260609_scoped_feature_flags.md
```diff
@@ -0,0 +1,534 @@
...
+
+1. **Per-cluster overrides (cluster-coherent flags).** We want the
+   `mz_catalog_server` cluster (the builtin cluster known internally as `s2`) to
+   run with a different set of flags than the rest of the environment,
+   independent of the environment-wide values. The motivating class here is
+   **optimizer flags**: there is broad interest in being able to flip optimizer
+   features per cluster from LD, e.g. enabling a feature on the catalog server
+   (or a specific user cluster) without affecting everyone else.
+
+2. **Per-replica overrides (replica-local flags), keyed by size family.** We are
```
mgree, line 25: How does this interact with self-managed, where (as I understand it) customers can create their own replica sizes?

doc/developer/design/20260609_scoped_feature_flags.md
```diff
@@ -0,0 +1,534 @@
...
+
+```rust
+enum ParameterScope {
+    /// Environment-wide only; no cluster/replica overrides. The default, so all
+    /// existing synced parameters are unchanged.
+    Environment,
+    /// Cluster-coherent: env-wide base + per-cluster overrides. Evaluated with
+    /// the `cluster` context (replica-free) and resolved at plan time via
+    /// `OptimizerFeatureOverrides`. e.g. optimizer features.
+    Cluster,
```
mgree, line 262: Is there some way to uses this information to automate the `OptimizerConfig`/`OptimizerFeatureOverrides`? That would remove a lot of friction from making optimizer feature flags.

antiguru: I think yes, optimizer feature flags are just cluster-scoped flags. With my proposal, we likely wouldn't need the overrides anymore. That said, I'm not very familiar with the optimizer overrides.

ggevay: We need the overrides also for EXPLAIN, i.e., `EXPLAIN WITH (<flag>)` being able to tell us how the plan _would_ look like if a flag were flipped.

doc/developer/design/20260609_scoped_feature_flags.md
```diff
@@ -0,0 +1,624 @@
...
+
+Because ids are never reused, an incarnation pin keyed by `*_id` cannot
+accidentally carry over; because predicates key on stable names/families, role
+intent survives recreate by design. The operator picks the behavior by choosing
+the attribute. This is why we expose **both** identifiers on each context.
+
+### Storage: a durable cache, written solely by the sync loop
+
+Scoped overrides are **persisted**, so they survive an `environmentd` restart and
+stay in effect while LaunchDarkly is slow to sync or unavailable. This is
```
DAlperin, line 313: Do they? I mean this seems reasonable but isn't this also status quo? Right now if LD is slow to sync it would fall back to the default in code and you might lose the environment level setting from LD

antiguru: We store whatever we synced from LD in the catalog, so on restart we should come up with the same flags we had the last time we synced.

aljoscha (approved):
this is already implemented and used, so stamping as is

antiguru:
Thanks for the review!

PR #36952, opened 2026-06-10
sql: Error with table functions and unsupported FILTER/OVER/DISTINCT

def-:
Closes SQL-346

### DIFF
--- a/src/sql-parser/src/parser.rs
+++ b/src/sql-parser/src/parser.rs
@@ -8338,7 +8338,7 @@ impl<'a> Parser<'a> {
 
     fn parse_rows_from(&mut self) -> Result<TableFactor<Raw>, ParserError> {
         self.expect_token(&Token::LParen)?;
-        let functions = self.parse_comma_separated(Parser::parse_named_function)?;
+        let functions = self.parse_comma_separated(Parser::parse_windowless_function)?;
         self.expect_token(&Token::RParen)?;
         let (with_ordinality, alias) = self.parse_table_function_suffix()?;
         Ok(TableFactor::RowsFrom {
@@ -8371,6 +8371,23 @@ impl<'a> Parser<'a> {
     fn parse_named_function(&mut self) -> Result<Function<Raw>, ParserError> {
         let name = self.parse_raw_name()?;
         self.parse_function(name)
+    }
+
+    /// Parses a function call in a position that only permits "windowless"
+    /// function syntax (PostgreSQL's `func_expr_windowless`): DISTINCT,
+    /// FILTER, and OVER are not part of the grammar here, which guarantees
+    /// the planner's invariant that table functions never carry them.
+    fn parse_windowless_function(&mut self) -> Result<Function<Raw>, ParserError> {
+        let name = self.parse_raw_name()?;
+        self.expect_token(&Token::LParen)?;
+        let args = self.parse_optional_args(false)?;
+        Ok(Function {
+            name,
+            args,
+            filter: None,
+            over: None,
+            distinct: false,
+        })
     }
 
     fn parse_derived_table_factor(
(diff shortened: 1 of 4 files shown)


SangJunBak:
So the first question I had is **why would we `assert` here in the first place but bail in other places**? Digging into it, we expect the parser to hardcode `filter: None, over: None, distinct: false` for `FROM f()` (parser.rs:8333-8340). And for evaluating `(SELECT generate_series(...)) `, we only lift a table function out of the SELECT when filter/over/distinct are all empty `(query.rs:6392-6398)`. **So the assert is a true invariant there.** 

However, **`ROW FROM` is an edge case here** and allows us to pass non nullish values for filter, distinct, and over. I think this change makes sense and we should bail instead of panic. However, I think we should also error at the parser level, **[something that Postgresql does](https://github.com/postgres/postgres/blob/master/src/backend/parser/gram.y#L14915)**. We could maybe do something like extending `parse_rows_from`. Something Claude recommended:
```rs
// Before:
  fn parse_rows_from(&mut self) -> Result<TableFactor<Raw>, ParserError> {
        self.expect_token(&Token::LParen)?;
        let functions = self.parse_comma_separated(Parser::parse_named_function)?;
        self.expect_token(&Token::RParen)?;
        let (with_ordinality, alias) = self.parse_table_function_suffix()?;
        Ok(TableFactor::RowsFrom {
            functions,
            alias,
            with_ordinality,
        })
    }

// After:
fn parse_rows_from(&mut self) -> Result<TableFactor<Raw>, ParserError> {
    self.expect_token(&Token::LParen)?;
    let functions = self.parse_comma_separated(Parser::parse_windowless_function)?;
    self.expect_token(&Token::RParen)?;
    let (with_ordinality, alias) = self.parse_table_function_suffix()?;
    Ok(TableFactor::RowsFrom {
        functions,
        alias,
        with_ordinality,
    })
}

/// Parses a function call in a position that only permits "windowless"
/// function syntax (PostgreSQL's `func_expr_windowless`): DISTINCT, FILTER,
/// and OVER are not part of the grammar here, which guarantees the planner's
/// invariant that table functions never carry them.
fn parse_windowless_function(&mut self) -> Result<Function<Raw>, ParserError> {
    let name = self.parse_raw_name()?;
    self.expect_token(&Token::LParen)?;
    let args = self.parse_optional_args(false)?;
    Ok(Function {
        name,
        args,
        filter: None,
        over: None,
        distinct: false,
    })
}
```

For example in Postgresql, 
```sql
postgres=# SELECT * FROM ROWS FROM (generate_series(1, 2) FILTER (WHERE true), generate_series(3, 4));
ERROR:  syntax error at or near "FILTER"
LINE 1: SELECT * FROM ROWS FROM (generate_series(1, 2) FILTER (WHERE...
                                                       ^
postgres=# SELECT * FROM ROWS FROM (generate_series(1, 2) OVER (), generate_series(3, 4));
ERROR:  syntax error at or near "OVER"
LINE 1: SELECT * FROM ROWS FROM (generate_series(1, 2) OVER (), gene...
                                                       ^
postgres=# SELECT * FROM ROWS FROM (generate_series(DISTINCT 1, 2), generate_series(3, 4));
ERROR:  DISTINCT specified, but generate_series is not an aggregate function
LINE 1: SELECT * FROM ROWS FROM (generate_series(DISTINCT 1, 2), gen...
                                 ^
```

SangJunBak (changes requested):
Overall changes makes sense to me! I would recommend extending it such that we're catching it at the parser level too! https://github.com/MaterializeInc/materialize/pull/36952#issuecomment-4687435100

def-:
Thanks Jun, done!

SangJunBak (approved):
Thanks for doing this change!

PR #36953, opened 2026-06-10
Add LaunchDarkly flag consistency check

antiguru:
### Motivation

Materialize synchronizes system parameters (system variables, dyncfgs, and feature flags — everything `SystemVars::iter_synced()` returns) from LaunchDarkly at runtime. If a synchronized parameter has no corresponding LaunchDarkly flag, `client.variation` silently falls back to the compiled-in default and the parameter can never be controlled in cloud. Conversely, a deployment without LaunchDarkly runs on the compiled-in default, so a LaunchDarkly value that diverges from that default means cloud behaves differently from the out-of-the-box default. Nothing checks for either today.

### Description

A new nightly CI step (`test/launchdarkly-flag-consistency`) that compares Materialize's synchronized parameters against the LaunchDarkly project. It **fails** on four kinds of discrepancy, each gated by a curated known-exceptions allowlist so the build is green on today's baseline and only a *new* discrepancy turns it red:

- **Missing** — a synchronized parameter exists in Materialize but has no flag in LaunchDarkly (it can't be controlled in cloud), unless it is in `KNOWN_MISSING_FROM_LD`.
- **Stale** — a LaunchDarkly flag is no longer a synchronized parameter in the current build or the last published release (a candidate for archival), unless it is in `KNOWN_STALE_LD_FLAGS`. The last release is considered to avoid flagging a flag a deployed older version still needs.
- **Cloud vs default** — a flag's production LaunchDarkly default differs from the compiled-in default (e.g. a feature enabled in cloud but off by default), unless it is deliberate cloud-only tuning in `INTENTIONAL_LD_OVERRIDES`.
- **Cross-environment** — a flag's served default differs between LaunchDarkly environments (production vs staging), which usually only happens during a staged rollout, unless it is in `KNOWN_CROSS_ENV_DIVERGENCES`.

The allowlists capture the accepted baseline and are expected to shrink over time; the check also prints any allowlist entry that is no longer a discrepancy so it can be pruned (e.g. a cross-environment rollout self-flags once it reaches production). The current baseline seeds: 231 missing, 35 stale, 35 cloud-vs-default overrides, 38 cross-environment. The cloud-vs-default group includes a commented TODO subset of feature flags that default *on* in code but are served *off* in production cloud — allowlisted for now and to be triaged (reconcile the compiled-in default, or flip production).

How it works:
- Boots the current build (and, best-effort, the last published release) with an empty `system_parameter_defaults` and no LaunchDarkly SDK key, so `SHOW ALL` as `mz_system` reports the true compiled-in defaults rather than mzcompose's CI overrides or LaunchDarkly-synced values. Volumes are destroyed between boots so the older binary doesn't trip over newer persist/catalog state.
- Fetches flag keys+tags from the LaunchDarkly list endpoint, and per-flag targeting (for the flags present in both) to determine the value each environment serves by default.
- Compares values via a shared `values_equivalent` helper that normalizes representation differences: durations (`1 min` == `60s`, `10 s` == raw-milliseconds `10000`), byte sizes (`1GB` == `1073741824`), and boolean spellings (`on` == `true`), so formatting differences aren't reported.
- Filters out per-session variables (kept in sync with `src/sql/src/session/vars.rs`), the `enable_launchdarkly` kill switch, and `ci-test`-tagged throwaway flags.

Configuration via env vars: `LAUNCHDARKLY_PROJECT` (default `default`) and `LAUNCHDARKLY_ENVIRONMENTS` (default `production,staging`; the first is the cloud baseline compared against the compiled-in default).

**Pipeline integration** (`ci/nightly/pipeline.template.yml`): new `launchdarkly-flag-consistency` step on `linux-aarch64-small` (has `LAUNCHDARKLY_API_TOKEN`), depends on `build-aarch64`. It is blocking — a discrepancy outside the allowlists fails the step. (`--no-fail` remains as a local-only override for downgrading failures to warnings.)

### Verification

Exercised by the nightly pipeline (requires `LAUNCHDARKLY_API_TOKEN` and LaunchDarkly API access). Trial runs against the real `default` project confirmed it surfaces genuine drift and that the allowlists make the baseline green. The `values_equivalent` duration/byte/boolean normalization is unit-tested locally.

https://claude.ai/code/session_01Qm3eZHzYhDYoMe4okpebri

### DIFF
--- a/ci/nightly/pipeline.template.yml
+++ b/ci/nightly/pipeline.template.yml
@@ -1436,6 +1436,17 @@ steps:
           - ./ci/plugins/mzcompose:
               composition: launchdarkly
 
+      - id: launchdarkly-flag-consistency
+        label: "LaunchDarkly flag consistency"
+        depends_on: build-aarch64
+        timeout_in_minutes: 30
+        agents:
+          # Requires LAUNCHDARKLY_API_TOKEN
+          queue: linux-aarch64-small
+        plugins:
+          - ./ci/plugins/mzcompose:
+              composition: launchdarkly-flag-consistency
+
   - group: E2E
     key: e2e
     steps:
--- a/ci/nightly/pipeline.template.yml
+++ b/ci/nightly/pipeline.template.yml
@@ -1446,6 +1446,10 @@ steps:
         plugins:
           - ./ci/plugins/mzcompose:
               composition: launchdarkly-flag-consistency
+              # Warn-only for now: many synchronized parameters legitimately
+              # have no LaunchDarkly flag. Drop --no-fail once the backlog of
+              # missing flags has been triaged.
+              args: [--no-fail]
 
   - group: E2E
     key: e2e
--- a/ci/nightly/pipeline.template.yml
+++ b/ci/nightly/pipeline.template.yml
@@ -1446,10 +1446,6 @@ steps:
         plugins:
           - ./ci/plugins/mzcompose:
               composition: launchdarkly-flag-consistency
-              # Warn-only for now: many synchronized parameters legitimately
-              # have no LaunchDarkly flag. Drop --no-fail once the backlog of
-              # missing flags has been triaged.
-              args: [--no-fail]
 
   - group: E2E
     key: e2e
--- a/test/launchdarkly-flag-consistency/mzcompose.py
+++ b/test/launchdarkly-flag-consistency/mzcompose.py
@@ -382,12 +382,18 @@ def _as_bool(value: Any) -> bool | None:
 
 
 def values_equivalent(a: Any, b: Any) -> bool | None:
-    """Whether two values represent the same setting, comparing durations by
+    """Whether two values represent the same setting. Durations are compared by
     value so representation differences ("1 min" vs "60s", "10 s" vs the
     raw-milliseconds "10000") are not treated as differences. `a` and `b` may be
-    Materialize's `SHOW` strings or LaunchDarkly's typed values. Returns `None`
-    when the two cannot be compared confidently (e.g. one side is missing, or a
-    byte size like "1GB" vs the raw number 1073741824)."""
+    Materialize's `SHOW` strings or LaunchDarkly's typed values.
+
+    Returns `None` when the two cannot be compared confidently: when one side is
+    missing, or one side is a bool/duration the other cannot be parsed as.
+
+    Byte sizes are *not* normalized: a unit-bearing value like "1GB" is compared
+    textually against the equivalent raw byte count (1073741824) and so is
+    reported as differing even though the two are equal. Suppress such flags via
+    INTENTIONAL_LD_OVERRIDES if the noise is undesirable."""
     if a is None or b is None:
         return None
     if isinstance(a, bool) or isinstance(b, bool):
--- a/test/launchdarkly-flag-consistency/mzcompose.py
+++ b/test/launchdarkly-flag-consistency/mzcompose.py
@@ -164,19 +164,6 @@ CI_TEST_TAG = "ci-test"
 IGNORED_MZ_PARAMETERS: set[str] = set()
 
 
-class Flag:
-    """A LaunchDarkly flag, reduced to what this check needs."""
-
-    def __init__(self, tags: list[str], defaults: dict[str, Any | None]) -> None:
-        self.tags = tags
-        # The value LaunchDarkly serves to an unmatched context (targeting off
-        # -> off variation, otherwise the fallthrough variation), per checked
-        # environment. A value is `None` when it could not be determined
-        # unambiguously (e.g. a percentage rollout) or the environment is
-        # unknown to the flag.
-        self.defaults = defaults
-
-
 def synced_parameters(c: Composition) -> dict[str, str]:
     """Return the synchronized system parameters of the running `materialized`
     service as a mapping from name to its (default) value, derived from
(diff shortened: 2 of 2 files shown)


antiguru:
Nightly run in https://buildkite.com/materialize/nightly/builds/16741

ci/nightly/pipeline.template.yml
```diff
@@ -1436,6 +1436,21 @@ steps:
...
+        agents:
+          # Requires LAUNCHDARKLY_API_TOKEN
+          queue: linux-aarch64-small
+        plugins:
+          - ./ci/plugins/mzcompose:
+              composition: launchdarkly-flag-consistency
+              # Warn-only for now: many synchronized parameters legitimately
+              # have no LaunchDarkly flag. Drop --no-fail once the backlog of
+              # missing flags has been triaged.
+              args: [--no-fail]
```
def-: What good is adding it to Nightly if it can't fail? I would instead add a list of known unsynchronized parameters and fail for each difference other than that.

test/launchdarkly-flag-consistency/mzcompose.py
```diff
@@ -0,0 +1,650 @@
...
+              default in the production environment differs from Materialize's
+              compiled-in default. Self-managed deployments have no LaunchDarkly
+              and run on the compiled-in default, so this means cloud and
+              self-managed behave differently (e.g. a feature enabled in cloud
+              but still off by default in self-managed). Parameters that
+              deliberately tune cloud-only infrastructure are listed in
+              `INTENTIONAL_LD_OVERRIDES` and excluded.
+  * WARN   -- a flag's served default differs *between* LaunchDarkly
+              environments (e.g. production vs staging), which usually only
+              happens during a staged rollout.
```
def-: I think those are useless. Who will look at warnings in CI runs? We should surface each failure if possible

test/launchdarkly-flag-consistency/mzcompose.py
```diff
@@ -0,0 +1,650 @@
...
+#
+# Keep this in sync with the session variables defined in
+# `src/sql/src/session/vars.rs`:
+#   * the `vars` array in `SessionVars::new`,
+#   * the `SESSION_SYSTEM_VARS` set, and
+#   * the `mz_version` / `is_superuser` pseudo-variables added by
+#     `SessionVars::iter`.
+# Matched case-insensitively, so the historical capitalization of e.g.
+# `DateStyle` does not matter.
+SESSION_VARIABLES = {
```
def-, line 118: Isn't there a way to list all session variables dynamically? Keeping it in sync manually will definitely fail at some point. There should at least be a linter, but I prefer dynamic.

test/launchdarkly-flag-consistency/mzcompose.py
```diff
@@ -0,0 +1,650 @@
...
+IGNORED_MZ_PARAMETERS: set[str] = set()
+
+# Parameters whose production LaunchDarkly value is *deliberately* different from
+# the compiled-in default, because they tune cloud-specific infrastructure that
+# does not apply to self-managed deployments. These are excluded from the
+# "cloud vs self-managed" divergence warning (but still participate in the
+# cross-environment prod-vs-staging check). Curate this list; everything else
+# that diverges is surfaced so a feature accidentally left off in self-managed
+# is noticed.
+INTENTIONAL_LD_OVERRIDES: set[str] = {
```
def-, line 468: Do we also want to encode what the specific values are?

test/launchdarkly-flag-consistency/mzcompose.py
```diff
@@ -0,0 +1,650 @@
...
+        ","
+    )
+    if env.strip()
+]
+PRODUCTION_ENVIRONMENT = (
+    LAUNCHDARKLY_ENVIRONMENTS[0] if LAUNCHDARKLY_ENVIRONMENTS else ""
+)
+
+# The internal SQL port, on which we can connect as `mz_system`.
+MZ_SYSTEM_PORT = 6877
```
def-: Just using 6877 directly is more intuitive (to me)

test/launchdarkly-flag-consistency/mzcompose.py
```diff
@@ -0,0 +1,650 @@
...
+    "hours": 3600.0,
+    "d": 86400.0,
+    "day": 86400.0,
+    "days": 86400.0,
+}
+
+_DURATION_TOKEN = re.compile(r"(\d+(?:\.\d+)?)\s*([a-zµ]+)")
+
+
+def parse_duration_seconds(
```
def-, line 706: It feels like we shouldn't be parsing this here. I don't have a good solution though. Edit: Same for everything else under the values_equivalent comparison. Can't we make Materialize do the comparsion?

test/launchdarkly-flag-consistency/mzcompose.py
antiguru: Yup, same here. I'll file a follow-up issue to detect session variables, at the moment there is none other than looking at Rust code, or Rust code looking at Python code.

test/launchdarkly-flag-consistency/mzcompose.py
antiguru: I would not; as it causes churn if we modify them in LD.

test/launchdarkly-flag-consistency/mzcompose.py
antiguru: The issue is that without parsing, we get mostly mismatches. `1s` vs. `1 s`, `10s` vs `10000`. Not great, but seems maintainable?

ci/nightly/pipeline.template.yml
antiguru: Removed the check and added a list of known failures so it's not immediately red. I'll file an issue to fix the known failures.

test/launchdarkly-flag-consistency/mzcompose.py
antiguru: Removed.

def- (approved):
Thanks

PR #36954, opened 2026-06-10
timely-util: serialize resident merge-batcher chunks into a fitting buffer

antiguru:
### Motivation

The column-paged merge batcher routes every shipped chunk through `ColumnPager::page`.
With paged spill gated off in production the pager is the disabled (always-resident) policy, so `page`'s `Skip` arm took the whole `Column::Typed` body into the merge chain.
That body arrives carrying `Column::merge_from`'s worst-case `reserve_for` capacity, sized for the unconsolidated union of both merge inputs.
Heavy consolidation collapses the record count but leaves the surplus capacity live in the chain across merge rounds.

Production heap profiles showed this as ~146 GB resident on a single replica, localized to `ColumnMergeBatcher::merge_by → Column::merge_from → reserve_for → RawVec<u64>::reserve`.
A before/after comparison across the live flag flip (same replica) put the arrange merge batcher at sub-2 GB under the columnation merger versus 146 GB under the columnar one — essentially all of the ~160 GB total heap growth.

### Description

In `page`'s `Skip` arm, serialize a `Typed` body into a `Vec<u64>` sized exactly to its content and store the fitting `Column::Align` instead of moving the over-reserved typed buffer into the chain.
The source column is cleared in place (matching the codec paths) so the high-capacity typed buffer stays with the caller for stash recycling.
`Align` / `Bytes` bodies are already fitting (or refcounted) and pass through unchanged.

Only the live accumulator and not-yet-shipped heads remain `Typed`; every chunk that reaches the chain is now serialized to a fitting buffer.
This only affects the resident path — when real paging is enabled, chunks already serialize on pageout.

### Verification

Adds `merge_chains_ships_fitting_align`, which asserts that under the disabled pager every shipped chain entry is `Resident(Column::Align)` (never `Typed`) and that the data round-trips.
Existing `column_pager` and `merge_batcher` unit tests continue to pass.

### DIFF
--- a/src/timely-util/src/column_pager.rs
+++ b/src/timely-util/src/column_pager.rs
@@ -369,7 +369,29 @@ impl ColumnPager {
                     bytes: len_bytes,
                     policy: Arc::clone(&self.policy),
                 };
-                return PagedColumn::Resident(std::mem::take(col), ticket);
+                // A resident chunk joins a merge chain and may live there
+                // across many merge rounds. A `Column::Typed` body arrives
+                // carrying `Column::merge_from`'s worst-case `reserve_for`
+                // capacity — sized for the unconsolidated union of both merge
+                // inputs — and parking that slack in the chain is the dominant
+                // source of merge-batcher resident memory. Serialize the body
+                // into a `Vec<u64>` sized exactly to its content and store the
+                // fitting `Column::Align` instead, clearing `col` in place (as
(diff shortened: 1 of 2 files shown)


DAlperin (approved):
Approved mod testing in staging. Though it does again make me want to keep the latest few chains resident...

PR #36956, opened 2026-06-10
adapter: make statement-logging end ownership linear, tolerate duplicate ends

aljoscha:
### Motivation

Alternative proposal for the statement-logging double-end panics fixed in #36943, put up to gather comments on the approach. It includes both commits of #36943 (the slow-path guard defuse and the tests are kept as-is) and replaces the fast-path mechanism with a design that fixes the *class* of bug rather than the two known instances.

The underlying problem: end-of-execution logging must happen exactly once per begun statement, but the frontend peek sequencing duplicated the capability to end a statement — one copy registered with the coordinator, one retained by the frontend — and decided "who logs the end" by inspecting response shapes after the fact. Every teardown (`DROP CLUSTER`, cancellation) racing a dispatch produced a double end, and the sink turned that into a process abort. Each new frontend-sequenced statement type (subscribes, read-then-write, …) would re-grow these races, window by window.

### Description

Three changes, one commit on top of #36943:

- **Linear, one-way end ownership.** The frontend's `StatementLoggingGuard` owns end-of-execution logging and is threaded through `try_frontend_peek_inner`. Dispatch sites that hand a statement to the coordinator for asynchronous completion (`RegisterFrontendPeek` Ok, `ExecuteSlowPathPeek` Ok, `ExecuteSubscribe` Ok) `defuse` the guard at the point of handoff; after that the frontend never logs the end, even on error. The response-shape matching in `try_frontend_peek` and the `end_already_logged` out-param from #36943 disappear; a still-armed guard at the end is simply retired with the result.
- **`UnregisterFrontendPeek` carries the end reason.** When `client.peek()` fails after registration, the coordinator (as owner) retires the peek with the frontend's error; if a concurrent teardown already retired it, the unregistration is a no-op — the same idempotency the pending-peeks layer already uses everywhere else. The `still_registered` bool reply is gone.
- **`end_statement_execution` tolerates duplicate ends** (first end wins, with a warning) instead of panicking. Exactly-once handoff is unattainable under async cancellation — a frontend future can be dropped between the coordinator registering a peek and the frontend defusing its guard, leaving both sides owning the end — and statement logging is observability: it must never abort environmentd. A missing `executions_begun` entry can only mean a duplicate (begins and the dispatch commands travel the same command channel in FIFO order), so nothing real is masked.

Drive-by fixes that fell out of the model:
- Dropping the frontend future mid-sequencing (client disconnect) used to leak a forever-running entry in `mz_statement_execution_history` (the guard was defused up front); the armed guard now emits an `Aborted` end.
- `verify_portal` failure with an inherited EXECUTE logging id used to discard the id without ending the statement; it now logs an `Errored` end.
- The `ExecuteSubscribe` handler defuses its coordinator-side guard on error, matching the slow-path contract (no error paths currently exist there, but the contract is now uniform).

`doc/developer/guide-adapter.md` gains a section documenting the ownership model and why the sink is tolerant.

Telemetry semantics: in the raced scenarios the recorded reason is the first end to arrive (e.g. `Canceled` from the teardown rather than the frontend's `Errored`); both descriptions are truthful. A pre-existing oddity worth a follow-up: frontend-emitted ends ship an `ended_at` that the coordinator discards and recomputes.

### Verification

The two `test/cluster` workflows from #36943 are kept and extended: since a duplicate end no longer panics, they now also assert that the logs contain no `duplicate end_statement_execution` warning, which keeps them sharp as protocol-regression tests (the deterministic fast-path failpoint test exercises exactly the raced window). `cargo check`/`clippy`/`fmt` pass; the mzcompose workflows and a nightly run have not been re-run on this branch yet.

### DIFF
--- a/doc/developer/generated/adapter/frontend_peek.md
+++ b/doc/developer/generated/adapter/frontend_peek.md
@@ -6,7 +6,7 @@ revision: e926ec3a86
 # adapter::frontend_peek
 
 Implements the "frontend peek" sequencing path, where SELECT query optimization and fast-path execution are performed in the pgwire connection task rather than the coordinator's main loop.
-`PeekClient::try_frontend_peek` is the main entry point; it verifies the portal, delegates statement logging setup to `PeekClient::begin_statement_logging` (which returns a `StatementLoggingGuard`), immediately defuses the guard, and then calls `try_frontend_peek_inner`.
+`PeekClient::try_frontend_peek` is the main entry point; it verifies the portal, delegates statement logging setup to `PeekClient::begin_statement_logging` (which returns a `StatementLoggingGuard` owning end-of-execution logging), threads the guard through `try_frontend_peek_inner` — dispatch sites that hand a statement to the coordinator for asynchronous completion defuse it — and retires any still-armed guard with the execution's outcome.
 `try_frontend_peek_inner` sequences the full pipeline — name resolution, planning, cluster selection, RBAC checks, timeline context determination, timestamp determination with read-hold acquisition, optimization (in a `spawn_blocking` task), and finally execution — handling `SELECT`, `EXPLAIN` (plan and pushdown), `COPY TO S3`, and `SUBSCRIBE` statement types.
 Fast-path peeks (constant, arrangement, or persist) are executed directly via `implement_fast_path_peek_plan`; slow-path dataflow plans are dispatched to the coordinator via `Command::ExecuteSlowPathPeek`.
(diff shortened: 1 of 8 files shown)


ggevay (approved):
Ok, let's merge this for now! (Sorry for taking so long to get back to this!)

(We can do the atomic handoff from the frontend task to the Coordinator later.)

I've rebased the PR to main and resolved the merge conflicts.

PR #36957, opened 2026-06-10
`jsonb_unpack` TableFunc, introduced by the optimizer

ggevay:
AI prototype for solving the json unpacking problem with the `TableFunc`/optimizer approach.

[Slack where multiple people are asking for fast json unpacking again.](https://materializeinc.slack.com/archives/CU7ELJ6E9/p1781002953424039?thread_ts=1780956289.880609&cid=CU7ELJ6E9)

Nightly (subset: random queries + feature benchmark, because of adding a specific feature benchmark scenario): https://buildkite.com/materialize/nightly/builds/16757

### DIFF
--- a/src/expr/src/relation.rs
+++ b/src/expr/src/relation.rs
@@ -680,10 +680,19 @@ impl MirRelationExpr {
                 }
                 result
             }
-            FlatMap { .. } => {
-                // FlatMap can add duplicate rows, so input keys are no longer
-                // valid
-                vec![]
+            FlatMap { func, .. } => {
+                if func.at_most_one_row_per_input() {
+                    // The table function emits at most one row per input row,
+                    // so it never duplicates input rows. It may drop rows, but
+                    // that cannot break uniqueness (same argument as Filter).
+                    // Input columns keep their positions, so input keys remain
+                    // valid as is.
+                    input_keys.next().unwrap().clone()
+                } else {
+                    // FlatMap can add duplicate rows, so input keys are no
+                    // longer valid.
+                    vec![]
+                }
             }
             Negate { .. } => {
                 // Although negate may have distinct records for each key,
--- a/src/expr/src/relation/func.rs
+++ b/src/expr/src/relation/func.rs
@@ -26,12 +26,13 @@ use mz_ore::{soft_assert_eq_no_log, soft_assert_or_log, soft_panic_or_log};
 use mz_repr::adt::array::ArrayDimension;
 use mz_repr::adt::date::Date;
 use mz_repr::adt::interval::Interval;
+use mz_repr::adt::jsonb::JsonbRef;
 use mz_repr::adt::numeric::{self, Numeric, NumericMaxScale};
 use mz_repr::adt::regex::{Regex as ReprRegex, RegexCompilationError};
 use mz_repr::adt::timestamp::{CheckedTimestamp, TimestampLike};
 use mz_repr::{
-    ColumnName, Datum, Diff, ReprColumnType, ReprRelationType, Row, RowArena, RowPacker, SharedRow,
-    SqlColumnType, SqlRelationType, SqlScalarType, datum_size,
+    ColumnName, Datum, Diff, InputDatumType, ReprColumnType, ReprRelationType, Row, RowArena,
+    RowPacker, SharedRow, SqlColumnType, SqlRelationType, SqlScalarType, datum_size, strconv,
 };
 use num::{CheckedAdd, Integer, Signed, ToPrimitive};
 use ordered_float::OrderedFloat;
@@ -48,7 +49,9 @@ use crate::explain::{HumanizedExpr, HumanizerMode};
 use crate::relation::{
     ColumnOrder, WindowFrame, WindowFrameBound, WindowFrameUnits, compare_columns,
 };
-use crate::scalar::func::{add_timestamp_months, jsonb_stringify};
+use crate::scalar::func::{
+    add_timestamp_months, jsonb_list_index_position, jsonb_path_step, jsonb_stringify,
+};
 
 // TODO(jamii) be careful about overflow in sum/avg
 // see https://timely.zulipchat.com/#narrow/stream/186635-engineering/topic/additional.20work/near/163507435
@@ -3392,6 +3395,25 @@ pub enum TableFunc {
     JsonbObjectKeys,
     JsonbArrayElements,
     JsonbArrayElementsStringify,
+    /// Extracts multiple fields from a single jsonb value in one pass, with
+    /// one output column per requested field, emitting **exactly one** output
+    /// row per input row (never zero, never more, never an error).
+    ///
+    /// Takes one argument, the jsonb value to unpack. Each output column
+    /// equals exactly what the corresponding scalar accessor (`->`, `->>`,
+    /// `#>`, `#>>`; see `jsonb_get_*` in `crate::scalar::func`) would have
+    /// returned when applied to that value — including on non-object inputs
+    /// and on SQL NULL input (where every output is NULL; note that
+    /// `empty_on_null_input` is false). The point over the scalar accessors
+    /// is performance: `k` scalar accessors scan the value `k` times, while
+    /// this function resolves all fields in a single pass (jsonb object keys
+    /// are stored sorted, enabling a sorted-merge lookup).
+    ///
+    /// Internal only: introduced by the `JsonbUnpack` transform in
+    /// `mz-transform`; not reachable from SQL and has no catalog entry.
+    JsonbUnpack {
+        fields: Vec<JsonbUnpackField>,
+    },
     RegexpExtract(AnalyzedRegex),
     CsvExtract(usize),
     GenerateSeriesInt32,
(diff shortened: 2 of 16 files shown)


src/expr/src/relation/func.rs
```diff
@@ -3392,6 +3395,25 @@ pub enum TableFunc {
     JsonbObjectKeys,
     JsonbArrayElements,
     JsonbArrayElementsStringify,
+    /// Extracts multiple fields from a single jsonb value in one pass, with
```
mgree, line 3398: Does this comment belong here, or on the function that implements it?

src/expr/src/relation/func.rs
```diff
@@ -3392,6 +3395,25 @@ pub enum TableFunc {
...
+    /// `#>`, `#>>`; see `jsonb_get_*` in `crate::scalar::func`) would have
+    /// returned when applied to that value — including on non-object inputs
+    /// and on SQL NULL input (where every output is NULL; note that
+    /// `empty_on_null_input` is false). The point over the scalar accessors
+    /// is performance: `k` scalar accessors scan the value `k` times, while
+    /// this function resolves all fields in a single pass (jsonb object keys
+    /// are stored sorted, enabling a sorted-merge lookup).
+    ///
+    /// Internal only: introduced by the `JsonbUnpack` transform in
+    /// `mz-transform`; not reachable from SQL and has no catalog entry.
```
mgree, line 3413: As discussed, I think we _should_ make it reachable by SQL---a nice escape hatch in case the transform is too fussy.

src/expr/src/relation/func.rs
```diff
@@ -3487,6 +3509,218 @@ struct WithOrdinality {
...
+    Eq,
+    PartialEq,
+    Ord,
+    PartialOrd,
+    Serialize,
+    Deserialize,
+    Hash,
+    MzReflect
+)]
+pub enum JsonbUnpackFieldKind {
```
mgree, line 3546: Is it not possible to have a path of indices for arrays nested in objects? Is possibly to have a mix of `Key` and `Index` in one unpack? I worry that there are some unrepresentable/doomed-to-fail states here, like `vec![Key("foo"), Index(0)]` (unless `Index` turns into fiedl access of a `"0"` field).

src/expr/src/relation/func.rs
```diff
@@ -3487,6 +3509,218 @@ struct WithOrdinality {
...
+        Datum::Null => {}
+        Datum::Map(map) => {
+            // Lookups to resolve against the object: `Key` fields and the
+            // heads of non-empty `Path` fields. `Index` fields are NULL on
+            // objects, as in `jsonb_get_int64`.
+            let mut wanted: Vec<(&str, usize)> = Vec::with_capacity(fields.len());
+            for (idx, field) in fields.iter().enumerate() {
+                match &field.kind {
+                    Key(k) => wanted.push((k.as_str(), idx)),
+                    Index(_) => {}
```
mgree, line 3585: Is silent drop the right behavior? What would an indexed get an object normally do? Push the `NULL` like we see below?

src/expr/src/relation/func.rs
```diff
@@ -3487,6 +3509,218 @@ struct WithOrdinality {
...
+
+    let mut row = Row::default();
+    let mut packer = row.packer();
+    for (field, value) in fields.iter().zip_eq(resolved) {
+        packer.push(match value {
+            Some(v) if field.stringify => jsonb_stringify(v, temp_storage)
+                .map(Datum::String)
+                .unwrap_or(Datum::Null),
+            Some(v) => v,
+            None => Datum::Null,
```
mgree, line 3718: Confirming the `Null` is right, as opposed to a `JsonNull`!

src/expr/src/relation/func.rs
```diff
@@ -3487,6 +3509,218 @@ struct WithOrdinality {
...
+            // as in `jsonb_get_string`. The array length is computed at most
+            // once, and only if some index is negative.
+            let mut cached_len: Option<usize> = None;
+            let mut wanted: Vec<(usize, usize)> = Vec::with_capacity(fields.len());
+            for (idx, field) in fields.iter().enumerate() {
+                let i = match &field.kind {
+                    Key(_) => None,
+                    Index(i) => Some(*i),
+                    Path(p) => match p.first() {
+                        Some(head) => strconv::parse_int64(head).ok(),
```
mgree, line 3645: Confirming: this is the right behavior for the usual path operator on arrays?

src/expr/src/relation/func.rs
```diff
@@ -3487,6 +3509,218 @@ struct WithOrdinality {
...
+                    // a non-collection value with 0 or -1 returns the value.
+                    Index(i) => (*i == 0 || *i == -1).then_some(other),
+                    // `#> '{}'` returns the input itself; any non-empty path
+                    // on a non-collection value is NULL (note the asymmetry
+                    // with `Index`, mirrored from `jsonb_get_path`).
+                    Path(p) => p.is_empty().then_some(other),
+                };
+            }
+        }
+    }
```
mgree, line 3708: 🥲

src/expr/src/relation/func.rs
```diff
@@ -3880,9 +4141,71 @@ impl TableFunc {
...
+    /// by `MirRelationExpr::keys_with_input_keys`.
+    ///
+    /// Compare the `FlatMapElimination` transform in `mz-transform`, which
+    /// discovers the same property dynamically, per instance, for FlatMaps
+    /// with all-literal arguments (of any table function), and removes the
+    /// FlatMap entirely.
+    ///
+    /// All variants are enumerated to ensure that added variants make a
+    /// conscious decision.
+    pub fn at_most_one_row_per_input(&self) -> bool {
```
mgree, line 4169: This is new! Do we _need_ this to prevent regressions, or just a nice-to-have? Might be good to consider it in a separate PR if the latter.

src/expr/src/relation/func.rs
```diff
@@ -3895,6 +4218,23 @@ impl fmt::Display for TableFunc {
...
+                        (JsonbUnpackFieldKind::Key(k), false) => write!(f, "->{:?}", k)?,
+                        (JsonbUnpackFieldKind::Key(k), true) => write!(f, "->>{:?}", k)?,
+                        (JsonbUnpackFieldKind::Index(i), false) => write!(f, "->{}", i)?,
+                        (JsonbUnpackFieldKind::Index(i), true) => write!(f, "->>{}", i)?,
+                        (JsonbUnpackFieldKind::Path(p), false) => write!(f, "#>{:?}", p)?,
+                        (JsonbUnpackFieldKind::Path(p), true) => write!(f, "#>>{:?}", p)?,
+                    }
+                }
+                f.write_str("]")
+            }
```
mgree, line 4237: Is this the surface syntax we would want? If we reveal it to users, we should make sure it matches up.

src/expr/src/relation/func.rs
```diff
@@ -3971,3 +4311,238 @@ impl WithOrdinality {
...
+        })
+    }
+
+    fn arb_field() -> impl Strategy<Value = JsonbUnpackField> {
+        // Keys/path elements from the same alphabet as `arb_json`, path
+        // elements also from small ints (to index into arrays); indices small
+        // (to hit), including 0/-1 (the scalar quirk) and the i64 extremes
+        // (`wrapping_sub` behavior).
+        let index = Union::new(vec![
+            (-6i64..6).boxed(),
```
mgree, line 4423: Is there a reason we only do 6.5% of the keyspace of `arb_json` here, but seemingly 100% of it elsewhere?

src/expr/src/scalar/func.rs
```diff
@@ -1645,17 +1645,29 @@ fn to_char_timestamp_tz_format(
...
 }
 
+/// Resolves a (possibly negative) json array index to the element position to
+/// fetch. Negative indices count backwards from the end, with the array length
+/// supplied on demand by `len` (and `wrapping_sub` semantics for extreme
+/// values).
+///
+/// Shared between the jsonb accessor functions (`jsonb_get_int64`,
+/// `jsonb_get_path`, and their stringifying variants) and
+/// `TableFunc::JsonbUnpack`, so that their semantics cannot drift apart.
```
mgree, line 1655: Maybe add a note about the `jsonb_unpack` caching? Kind of a key trick for asymptotics!

src/expr/src/scalar/func.rs
```diff
@@ -1697,6 +1709,26 @@ fn jsonb_get_string_stringify<'a>(
...
+    let v = match json.into_datum() {
+        Datum::Map(map) => map.iter().find(|(k, _)| key == *k).map(|(_k, v)| v),
+        Datum::List(list) => {
+            let i = strconv::parse_int64(key).ok()?;
+            let i = jsonb_list_index_position(i, || list.iter().count());
+            list.iter().nth(i)
+        }
+        _ => return None,
+    }?;
+    // `v` should be valid jsonb because it came from a jsonb container, but we
```
mgree, line 1726: Maybe a little prolix.

src/expr/src/relation.rs
```diff
@@ -680,10 +680,19 @@ impl MirRelationExpr {
...
+                    // so it never duplicates input rows. It may drop rows, but
+                    // that cannot break uniqueness (same argument as Filter).
+                    // Input columns keep their positions, so input keys remain
+                    // valid as is.
+                    input_keys.next().unwrap().clone()
+                } else {
+                    // FlatMap can add duplicate rows, so input keys are no
+                    // longer valid.
+                    vec![]
+                }
```
mgree, line 695: Nice! Can we attribute any unrelated plan improvements to this?

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+//! 1. every maximal Map/Filter/Project chain, treated as one
+//!    [`MapFilterProject`];
+//! 2. every `Reduce`'s group key and aggregate argument expressions, treated
+//!    as one expression bag. (By this phase of the pipeline, accessors in
+//!    GROUP BY and aggregate arguments live *inside* the Reduce:
+//!    `ReductionPushdown` absorbs Maps into Reduces in the logical pipeline.)
+//!
+//! Nothing else: not FlatMap arguments, TopK limits, Join equivalences, or
+//! ArrangeBy keys; MFPs already pushed into source imports are out of reach
+//! by pipeline position. Accessors in different sites never share an unpack.
```
mgree, line 31: What do we lose by not doing this on join equivalences? My intuition says all the batching should basically be done up front, but it's not clear to me. Maybe some kind of metric or mzexplore kind of work would help us know whether we want to work harder to batch JSON extractions in more obscure positions.

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+//!
+//! **What fires.** Within a site, after running the existing scalar ANF
+//! ([`MapFilterProject::memoize_expressions`]) on the bag:
+//!
+//! - A *candidate* is a memoized expression `<column> -> <literal>` (or
+//!   `->>`, `#>`, `#>>`) with a non-null literal key/index/path. Non-literal
+//!   keys are out of scope (no batched eval is possible without knowing the
+//!   fields up front; a future extension could handle keys that are at least
+//!   common across the site). Empty paths (`#> '{}'`, the identity) and
```
mgree, line 40: I think batched eval is just fine if we treat this as an eager expression, where we need to evaluate all of the fields to extract first.

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+//!   paths with null elements (constant NULL) are degenerate and skipped.
+//!   Accessors under `CASE` are out of scope as a consequence of reusing the
+//!   ANF: `eager_children` deliberately does not descend into `If` branches
+//!   (nor `COALESCE` tails), so such calls never become candidates.
+//! - At MFP sites, candidates transitively needed by a Filter predicate are
+//!   *pinned*: they stay below the unpacking, in their original scalar form,
+//!   so that predicates remain directly above the site's input where join
+//!   closures, arrangement-read fusion, and similar mechanisms consume them.
+//!   A pinned accessor that is also selected does not count toward the
+//!   threshold below. (Reduce sites have no predicates.)
```
mgree, line 50: I have trouble understanding this. Is the idea that we want to make sure we don't want to batch extract rows we would have filtered out anyway?

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+//! - Unpinned candidates are grouped by base column. A group **fires** iff
+//!   it has at least [`MIN_FIELDS_PER_UNPACK`] distinct fields and its base
+//!   is *below-computable*:
+//!   - any expression over input columns and pinned/below-computable columns
+//!     (materialized below the FlatMap stack if not already a column) — this
+//!     includes computed bases such as a string-to-jsonb cast; or
+//!   - a raw (`->`) field of another firing group, or an accessor over
+//!     below-computable columns — these bridge between stacked FlatMaps (the
+//!     latter inlined into the FlatMap argument expression).
```
mgree, line 59: This could be much clearer as, say, inference rules or careful definition (as opposed to prose).

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+//! **Output shape.** A fired MFP site is rebuilt as:
+//!
+//! ```text
+//! <input>
+//!   [Map + Filter]        — predicate support (in original scalar form) and
+//!                           computed bases; predicates evaluate before bases
+//!   FlatMap JsonbUnpack   — one per firing group, stacked in dependency order
+//!   [ FlatMap JsonbUnpack … ]
+//!   [Map] + [Project]     — leftover expressions rewritten over the unpacked
+//!                           columns + the site's original projection
```
mgree, line 80: Rewrite to be in `EXPLAIN` order (top down)?

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+//! ```
+//!
+//! At a fired Reduce site the FlatMap stack (plus a Map for computed bases)
+//! is spliced directly below the Reduce, and the rewritten expressions go
+//! back into the group key and aggregate arguments.
+//!
+//! A site where nothing fires is left **byte-for-byte unchanged**.
```
mgree, line 87: Gosh, thanks Claude! 😆

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+//! error-safe because there they evaluate after *all* of the site's
+//! predicates, i.e., on a subset of the rows they used to see.
+//!
+//! # Why didn't it fire?
+//!
+//! For debugging: the feature flag is off; the query is a fast-path peek
+//! (this transform is not part of the fast-path optimizer); the key is not a
+//! literal (or is an empty path); fewer than [`MIN_FIELDS_PER_UNPACK`]
+//! distinct fields on the same base within one site (chains count per link);
+//! the accessors are used by filters (pinned — even if also selected); the
```
mgree, line 104: This seems to conflict with the definition of *below-computable*... I'm confused!

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+/// of their own, which is safe. Processing this transform's own output never
+/// undoes or re-wraps what it built (unpacked accessors are gone from the
+/// expressions, predicate support gets re-pinned); it can only find
+/// *legitimately* remaining work — in particular, a group excluded because
+/// its base depended on an unpack output through a non-accessor expression
+/// becomes unpackable once that expression is an ordinary map expression of
+/// the rewritten upper layer. So running the transform to a fixpoint would
+/// be safe (each firing strictly decreases the number of accessor calls in
+/// sites, so it terminates) and would even lift the interleaved-base
+/// limitation; a single pass just doesn't chase that rare case.
```
mgree, line 189: What is the "interleaved-base limitation"?

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+fn skip_site(expr: &mut MirRelationExpr) -> &mut MirRelationExpr {
```
mgree, line 190: Inline this to the only use below?

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+            } => cur = input,
+            other => return other,
+        }
+    }
+}
+
+/// Descends through `JsonbUnpack` FlatMaps only (used below a rewritten
+/// Reduce, where the Map below the FlatMaps merges with the chain below into
+/// a site of its own).
+fn skip_flat_maps(expr: &mut MirRelationExpr) -> &mut MirRelationExpr {
```
mgree, line 210: Inline this to its only use?

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+    flat_maps: Vec<(Vec<JsonbUnpackField>, MirScalarExpr)>,
+    /// The site's outputs over the post-stack layout
+    /// `[input columns] ++ [demanded lower expressions] ++ [unpacked fields]`.
+    upper_mfp: MapFilterProject,
+}
+
+impl SitePlan {
+    /// Builds the lower MFP and the FlatMap stack on top of `input`. The
+    /// caller is responsible for the upper layer (`upper_mfp`), whose form
+    /// differs between MFP and Reduce sites.
```
mgree, line 321: I think it's generally better for a caller to pass information about its context and have someone do all the work than to expect the caller to take particular actions after---especially if those actions are easy to predict. But I see why it's a little tricky here: Reduce just doesn't really have a spot for it.

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+    /// An input column or a pinned expression: materialized below the stack.
+    Below,
+    /// A non-candidate expression over Below/BelowComputable columns: *can*
+    /// be materialized below the stack, but only is if some firing group's
+    /// base needs it.
+    BelowComputable,
+    /// An unfired candidate over a usable base: can be inlined into a
+    /// FlatMap argument to bridge to a deeper group. (If it is also used as
+    /// a value it is additionally evaluated in the upper layer; the rare
+    /// double evaluation of one accessor is accepted.)
```
mgree, line 350: I don't understand this.

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+    let mut state: Vec<State> = vec![State::Below; input_arity];
+    let mut fired: BTreeMap<usize, Vec<usize>> = BTreeMap::new();
+    for j in 0..n {
+        let next = if pinned[j] {
+            State::Below
+        } else if let Some((base, _)) = &candidates[j] {
+            let base_ok = matches!(
+                state[*base],
+                State::Below | State::BelowComputable | State::Link | State::Unpacked
+            );
```
mgree, line 427:
```suggestion
            let base_ok = state[*base] != State::Above;
```

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+        || mfp.predicates.iter().any(|(_, p)| contains_literal_err(p))
+    {
+        return None;
+    }
+
+    let mut mfp = mfp.clone();
+    mfp.memoize_expressions();
+    let input_arity = mfp.input_arity;
+    let exprs = &mfp.expressions;
+    let n = exprs.len();
```
mgree, line 377: Had to look back to see what `n` meant, would rename:
```suggestion
    let num_exprs = exprs.len();
```

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+    // demand: the direct support of the FlatMap arguments and of the upper
+    // layer. Everything else placed (e.g. predicate support) is consumed
+    // within the lower MFP and projected away below the stack.
+    //
+    // "The upper layer" is over-approximated as every expression not
+    // available from below, whether or not the site's projection transitively
+    // needs it; a slot demanded only by such dead expressions costs one
+    // unused pass-through column below the stack. (Dead slots are rare here:
+    // Demand and ProjectionPushdown ran earlier, and `memoize_expressions`
+    // only creates slots for expressions it found inside live ones.)
```
mgree, line 487: I only kind of understand this comment. Would be good to callout recursive nature of things here (due to `Link`).

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+        }
+    }
+
+    // Build the lower MFP: input columns stay put, then the pinned
+    // expressions (so that all predicates evaluate before any computed base
+    // can error), then the remaining placed expressions; the projection
+    // passes through the input columns and the demanded expressions.
+    // `optimize` inlines single-use expressions (e.g. predicate support back
+    // into the predicates) and drops anything unused.
+    let lower_slots: Vec<usize> = (0..n)
```
mgree, line 542: The order of events here is kind of confusing. We compute placements for the lower slots, _then_ the input, then add those placements in.

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+                e.visit_columns(|c| *c = remap_upper(*c));
+                e
+            }
+        })
+        .collect();
+    let mut upper_mfp = MapFilterProject::new(post_arity)
+        .map(upper_exprs)
+        .project(mfp.projection.iter().map(|p| remap_upper(*p)));
+    upper_mfp.optimize();
+
```
mgree, line 637: Add some assertion that the final arity is correct?

src/transform/src/jsonb_unpack.rs
```diff
@@ -0,0 +1,757 @@
...
+/// True iff the expression contains a literal error anywhere.
+fn contains_literal_err(expr: &MirScalarExpr) -> bool {
+    let mut found = false;
+    expr.visit_pre(|e| {
+        if e.is_literal_err() {
+            found = true;
+        }
+    });
+    found
+}
```
mgree, line 757: `MSE::contains_err` already does this, more efficiently (early return).

src/transform/src/lib.rs
```diff
@@ -891,6 +892,16 @@ impl Optimizer {
...
+            // Replace multiple jsonb accessors on a common value with a single
+            // multi-field unpacking table function. Placed late in the
+            // pipeline because it introduces FlatMaps, which many transforms
+            // handle poorly, and because it is a physical optimization: it
+            // changes how the data is accessed, not what is computed. Nothing
+            // after this point re-canonicalizes MFPs or rearranges what it
+            // builds; the final `fold_constants_fixpoint` below only needs a
+            // correct `TableFunc::eval`.
+            Box::new(jsonb_unpack::JsonbUnpack);
+                if ctx.features.enable_jsonb_unpack_transform,
```
mgree, line 904: Since the FlatMaps we produce here are somehow permanent, I wonder if there was some set of smaller manipulations we could do. Right now, we wait until we're done with all of our MFP/scalar manipulation, and then we batch together everything and call it done. Alternatively, we could batch together nearby JSON accesses earlier, maybe a few times, leaving behind messy maps and filters and projects (that other transforms can clean up after).

test/sqllogictest/transform/jsonb_unpack.slt
```diff
@@ -0,0 +1,412 @@
...
+# Use of this software is governed by the Business Source License
+# included in the LICENSE file.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0.
+
+# Tests for the JsonbUnpack optimizer transform, which replaces multiple
+# jsonb accessors on a common value with a single multi-field unpacking
+# table function.
```
mgree, line 12: Would be good to have bigger queries here---like, how does it behave around joins and other features it doesn't support?

mgree (commented):
This is looking pretty good! I like the scenario test, but maybe we can do some more targeted random testing, too?

We should talk a bit about some alternatives---I'm particularly curious about taking smaller bites and letting other passes clean up. For code/impl strategy: maybe get one more opinion (@antiguru and/or @frankmcsherry)?

ggevay:
Posting it also here that unfortunately there is an unexpected performance issue in the rendering (uncovered by benchmarking this prototype), where render_flat_map does something less performant than other code paths, making the optimized query not as fast as one would hope for, sometimes even a regressing. This looks fixable (maybe by making render_flat_map know about this particular TableFunc always producing 1 row), but needs some further thinking to estimate.

PR #36959, opened 2026-06-10
Scoped feature flags: resolution and gating

antiguru:
### Motivation

The resolution layer for scoped feature flags. The foundation (#37079), the
durable persistence and introspection (#37080), and the catalog-implications
foundation (#37151, which derives the replica-scoped controller push from the
committed catalog diff) have all merged. Create-time render-frozen correctness
is completed in the follow-up #37158. Design:
doc/developer/design/20260609_scoped_feature_flags.md (#36947).

Behind the `enable_scoped_system_parameters` dyncfg (off by default), this PR
consumes the durable scoped overrides and resolves them at the per-scope
boundaries.

### Description

* The sync loop evaluates the `cluster` (replica-free) and `replica` (size, size
  family, owning cluster) LaunchDarkly contexts and records an override only when
  the scoped value **differs from the environment-wide value**, compared in
  canonical encoding so bool spellings (`on`/`off` vs `true`/`false`) match. LD
  multi-context evaluation cannot reveal which context produced a value, so the
  differs-from-env test, not the `variation_detail` reason, is the recording
  rule. The evaluation is skipped while the gate is off, and the gate is read
  from the sync working copy so a disabled environment takes no catalog snapshot
  per tick.
* Resolution boundaries: cluster-coherent overrides feed plan-time
  `OptimizerFeatureOverrides` (with a lenient bool decode so `on`/`off` cannot
  panic the optimizer). Replica-local overrides reach the compute controller
  through the catalog implication from #37151.
* Create-time: a post-transact resolve folds a freshly created object's overrides
  into the working copy so it need not wait for the next sync tick. The follow-up
  #37158 moves this into the create transaction so render-frozen replica flags are
  set before `create_replica`.

### Verification

Canonical bool bridging, lenient optimizer decode, and frontend context unit
tests, plus the `test/launchdarkly` mzcompose workflow extended with
cluster/replica/durability/gating/create-time cases, validated end-to-end
against a live LaunchDarkly project.

https://claude.ai/code/session_01FUorxx5m6B8TT2VAek2STo

### DIFF
--- a/src/adapter/src/catalog/transact.rs
+++ b/src/adapter/src/catalog/transact.rs
@@ -73,7 +73,7 @@ use crate::catalog::{
     is_reserved_role_name, object_type_to_audit_object_type,
     system_object_type_to_audit_object_type,
 };
-use crate::config::ScopedParameters;
+use crate::config::{ScopedParameters, ScopedParametersScope};
 use crate::coord::ConnMeta;
 use crate::coord::catalog_implications::parsed_state_updates::ParsedStateUpdate;
 use crate::coord::cluster_scheduling::SchedulingDecision;
@@ -247,14 +247,19 @@ pub enum Op {
     },
     ResetAllSystemConfiguration,
     /// Persists the durable cache of scoped (per-cluster and per-replica)
-    /// system parameters to match `scoped`, the complete desired state computed
-    /// by the system-parameter sync loop (the sole writer). The handler diffs
-    /// against the current durable contents: it upserts changed/added entries
-    /// and removes entries the sync loop no longer serves, and lazily prunes
-    /// entries whose owning object id is absent from the catalog. See the
-    /// scoped feature flags design.
+    /// system parameters towards `scoped`. The handler diffs against the current
+    /// durable contents: it upserts changed/added entries and removes entries
+    /// the update no longer serves.
+    ///
+    /// `prune_scope` bounds removals to the objects the update was evaluated
+    /// for. `Some(scope)` removes a row only when its owning object is in
+    /// `scope`, so an object created after the update's evaluation snapshot, and
+    /// the override it folded into its own create transaction, survives a
+    /// concurrent full-state reconcile. `None` is an unscoped full replace, used
+    /// only by the disabled-feature clear path where no create-time writes race.
     UpdateScopedSystemParameters {
         scoped: ScopedParameters,
+        prune_scope: Option<ScopedParametersScope>,
     },
     /// Injects audit events into the catalog.
     ///
@@ -2712,18 +2717,38 @@ impl Catalog {
                     EventDetails::ResetAllV1,
                 )?;
             }
-            Op::UpdateScopedSystemParameters { scoped } => {
-                // The sync loop pushes the *complete* desired state each tick;
-                // diff it against the durable cache and persist only the delta.
-                // Entries whose owning object id is absent from the catalog are
-                // pruned (lazy GC); object ids are never reused, so such entries
-                // are inert and pruning is hygiene only.
+            Op::UpdateScopedSystemParameters {
+                scoped,
+                prune_scope,
+            } => {
+                // Diff `scoped` against the durable cache and persist only the
+                // delta: upsert changed/added entries, then remove entries the
+                // update no longer serves. A removal for a still-live object is
+                // bounded by `prune_scope` so this update does not delete a row
+                // for an object it was not evaluated for (e.g. one created after
+                // the update's snapshot, whose override rode its own create
+                // transaction). `None` prunes unconditionally (the
+                // disabled-feature clear path). Rows whose owning object is no
+                // longer live are always pruned regardless of `prune_scope`,
+                // because nothing else garbage collects them and ids are never
+                // reused, so this lazily reclaims orphans left by dropped
+                // objects.
                 let live_clusters: BTreeSet<ClusterId> =
                     tx.get_clusters().map(|cluster| cluster.id).collect();
                 let live_replicas: BTreeSet<ReplicaId> = tx
                     .get_cluster_replicas()
                     .map(|replica| replica.replica_id)
                     .collect();
+                let prune_cluster = |id: &ClusterId| {
+                    prune_scope
+                        .as_ref()
+                        .map_or(true, |s| s.clusters.contains(id))
+                };
+                let prune_replica = |id: &ReplicaId| {
+                    prune_scope
+                        .as_ref()
+                        .map_or(true, |s| s.replicas.contains(id))
+                };
 
                 // Cluster-coherent scope.
                 let existing_cluster: BTreeMap<(ClusterId, String), String> = tx
@@ -2742,10 +2767,10 @@ impl Catalog {
                         }
                     }
                 }
-                // Non-live clusters never enter `desired_cluster` (they `continue`
-                // above), so a missing entry already covers dropped clusters.
                 for (cluster_id, name) in existing_cluster.into_keys() {
-                    if !desired_cluster.contains(&(cluster_id, name.clone())) {
+                    if (!live_clusters.contains(&cluster_id) || prune_cluster(&cluster_id))
+                        && !desired_cluster.contains(&(cluster_id, name.clone()))
+                    {
                         tx.remove_cluster_system_config(cluster_id, &name);
                     }
                 }
@@ -2768,7 +2793,9 @@ impl Catalog {
                     }
                 }
                 for (replica_id, name) in existing_replica.into_keys() {
-                    if !desired_replica.contains(&(replica_id, name.clone())) {
+                    if (!live_replicas.contains(&replica_id) || prune_replica(&replica_id))
+                        && !desired_replica.contains(&(replica_id, name.clone()))
+                    {
                         tx.remove_replica_system_config(replica_id, &name);
                     }
                 }
@@ -3399,6 +3426,245 @@ mod tests {
             let all_t2 = state_all_at_once.try_get_entry(&id_t2).expect("all t2");
             assert_eq!(inc_t2.name(), all_t2.name());
             assert_eq!(inc_t2.owner_id, all_t2.owner_id);
+
+            catalog.expire().await;
+        })
+        .await
+    }
+
+    /// Exercises the diff-and-prune semantics of the
+    /// `Op::UpdateScopedSystemParameters` apply handler over the cluster scope.
+    /// Covers four behaviors: upsert of a desired row, the bounded-prune
+    /// protection that spares a live cluster outside `prune_scope`, removal of a
+    /// stale in-scope row, and the orphan-prune that reclaims a row whose owning
+    /// cluster was dropped even though that id is absent from `prune_scope`.
+    #[mz_ore::test(tokio::test)]
+    #[cfg_attr(miri, ignore)] // unsupported operation: can't call foreign function `TLS_client_method`
+    async fn test_transact_update_scoped_system_parameters_prune() {
+        use std::collections::{BTreeMap, BTreeSet};
+
+        use mz_catalog::memory::objects::{ClusterConfig, ClusterVariant};
+        use mz_controller_types::ClusterId;
+
+        use crate::catalog::DropObjectInfo;
+        use crate::config::{ScopedParameters, ScopedParametersScope};
+
+        Catalog::with_debug(|mut catalog| async move {
+            let param = "max_query_result_size".to_string();
+            let value = "1MB".to_string();
+
+            // Allocate and create two live clusters, A and B, the same way
+            // production code does.
+            let commit_ts = catalog.current_upper().await;
+            let cluster_a = catalog
+                .allocate_user_cluster_id(commit_ts)
+                .await
+                .expect("allocate cluster A");
+            let commit_ts = catalog.current_upper().await;
+            let cluster_b = catalog
+                .allocate_user_cluster_id(commit_ts)
+                .await
+                .expect("allocate cluster B");
+
+            let make_cluster_op = |id: ClusterId, name: &str| Op::CreateCluster {
+                id,
+                name: name.to_string(),
+                introspection_sources: Vec::new(),
+                owner_id: MZ_SYSTEM_ROLE_ID,
+                config: ClusterConfig {
+                    variant: ClusterVariant::Unmanaged,
+                    workload_class: None,
+                },
+            };
+
+            let oracle_write_ts = catalog.current_upper().await;
+            catalog
+                .transact(
+                    None,
+                    oracle_write_ts,
+                    None,
+                    vec![
+                        make_cluster_op(cluster_a, "test_cluster_a"),
+                        make_cluster_op(cluster_b, "test_cluster_b"),
+                    ],
+                )
+                .await
+                .expect("create clusters");
+
+            // Reads the durable cluster system configuration rows.
+            async fn rows(catalog: &Catalog) -> BTreeSet<(ClusterId, String, String)> {
+                let mut storage = catalog.storage().await;
+                let tx = storage.transaction().await.expect("open transaction");
+                tx.get_cluster_system_configurations()
+                    .map(|c| (c.cluster_id, c.name, c.value))
+                    .collect()
+            }
+
+            let scope_with = |ids: &[ClusterId]| ScopedParametersScope {
+                clusters: ids.iter().copied().collect(),
+                replicas: BTreeSet::new(),
+            };
+            let cluster_scoped = |id: ClusterId| {
+                let mut cluster = BTreeMap::new();
+                let mut overrides = BTreeMap::new();
+                overrides.insert(param.clone(), value.clone());
+                cluster.insert(id, overrides);
+                ScopedParameters {
+                    cluster,
+                    replica: BTreeMap::new(),
+                }
+            };
+            let empty_scoped = || ScopedParameters {
+                cluster: BTreeMap::new(),
+                replica: BTreeMap::new(),
+            };
+
+            // Behavior 1: upsert. A is live and in scope, so the row is created.
+            let oracle_write_ts = catalog.current_upper().await;
+            catalog
+                .transact(
+                    None,
+                    oracle_write_ts,
+                    None,
+                    vec![Op::UpdateScopedSystemParameters {
+                        scoped: cluster_scoped(cluster_a),
+                        prune_scope: Some(scope_with(&[cluster_a])),
+                    }],
+                )
+                .await
+                .expect("upsert A");
+            assert!(
+                rows(&catalog)
+                    .await
+                    .contains(&(cluster_a, param.clone(), value.clone())),
+                "upsert must create the row for A"
+            );
+
+            // Set up an additional row for B so the prune cases have something to
+            // act on.
+            let oracle_write_ts = catalog.current_upper().await;
+            catalog
+                .transact(
+                    None,
+                    oracle_write_ts,
+                    None,
+                    vec![Op::UpdateScopedSystemParameters {
+                        scoped: cluster_scoped(cluster_b),
+                        prune_scope: Some(scope_with(&[cluster_b])),
+                    }],
+                )
+                .await
+                .expect("upsert B");
+            assert!(
+                rows(&catalog)
+                    .await
+                    .contains(&(cluster_b, param.clone(), value.clone())),
+                "setup must create the row for B"
+            );
+
+            // Behavior 2: bounded prune. Running with empty `scoped` and a scope
+            // that excludes the live cluster B must leave B's row intact, because
+            // the update was not authoritative for B.
+            let oracle_write_ts = catalog.current_upper().await;
+            catalog
+                .transact(
+                    None,
+                    oracle_write_ts,
+                    None,
+                    vec![Op::UpdateScopedSystemParameters {
+                        scoped: empty_scoped(),
+                        prune_scope: Some(scope_with(&[cluster_a])),
+                    }],
+                )
+                .await
+                .expect("bounded prune");
+            assert!(
+                rows(&catalog)
+                    .await
+                    .contains(&(cluster_b, param.clone(), value.clone())),
+                "a live cluster outside prune_scope must keep its row"
+            );
+
+            // Behavior 3: stale in-scope prune. A is live and in scope but not
+            // desired, so its row is removed.
+            let oracle_write_ts = catalog.current_upper().await;
+            catalog
+                .transact(
+                    None,
+                    oracle_write_ts,
+                    None,
+                    vec![Op::UpdateScopedSystemParameters {
+                        scoped: empty_scoped(),
+                        prune_scope: Some(scope_with(&[cluster_a])),
+                    }],
+                )
+                .await
+                .expect("stale in-scope prune");
+            assert!(
+                !rows(&catalog)
+                    .await
+                    .iter()
+                    .any(|(id, _, _)| *id == cluster_a),
+                "a live in-scope undesired row must be removed"
+            );
+
+            // Behavior 4: orphan prune. Re-create A's row, drop cluster A, then run
+            // the update with a scope that does NOT contain A. The sync loop only
+            // ever places live ids in prune_scope, so the orphan must still be
+            // reclaimed because its owning cluster is no longer live.
+            let oracle_write_ts = catalog.current_upper().await;
+            catalog
+                .transact(
+                    None,
+                    oracle_write_ts,
+                    None,
+                    vec![Op::UpdateScopedSystemParameters {
+                        scoped: cluster_scoped(cluster_a),
+                        prune_scope: Some(scope_with(&[cluster_a])),
+                    }],
+                )
+                .await
+                .expect("re-create A row");
+            assert!(
+                rows(&catalog)
+                    .await
+                    .contains(&(cluster_a, param.clone(), value.clone())),
+                "re-created row for A must exist"
+            );
+
+            let oracle_write_ts = catalog.current_upper().await;
+            catalog
+                .transact(
+                    None,
+                    oracle_write_ts,
+                    None,
+                    vec![Op::DropObjects(vec![DropObjectInfo::Cluster(cluster_a)])],
+                )
+                .await
+                .expect("drop cluster A");
+
+            // B is the only live cluster the sync loop would have evaluated, so
+            // only B appears in prune_scope. A's id is deliberately absent.
+            let oracle_write_ts = catalog.current_upper().await;
+            catalog
+                .transact(
+                    None,
+                    oracle_write_ts,
+                    None,
+                    vec![Op::UpdateScopedSystemParameters {
+                        scoped: cluster_scoped(cluster_b),
+                        prune_scope: Some(scope_with(&[cluster_b])),
+                    }],
+                )
+                .await
+                .expect("orphan prune");
+            assert!(
+                !rows(&catalog)
+                    .await
+                    .iter()
+                    .any(|(id, _, _)| *id == cluster_a),
+                "orphan row for a dropped cluster must be removed even when absent from prune_scope"
+            );
 
             catalog.expire().await;
         })
--- a/src/adapter/src/config/frontend.rs
+++ b/src/adapter/src/config/frontend.rs
@@ -18,6 +18,8 @@ use hyper_tls::HttpsConnector;
 use launchdarkly_server_sdk as ld;
 use mz_build_info::BuildInfo;
 use mz_cloud_provider::CloudProvider;
+use mz_cluster_client::ReplicaId;
+use mz_controller_types::ClusterId;
 use mz_ore::now::NowFn;
 use mz_sql::catalog::EnvironmentId;
 use serde_json::Value as JsonValue;
@@ -38,6 +40,11 @@ pub struct SystemParameterFrontend {
     /// to use when populating the [SynchronizedParameters]
     /// instance in [SystemParameterFrontend::pull].
     key_map: BTreeMap<String, String>,
+    /// The environment ID, used to build scoped (`cluster` / `replica`)
+    /// evaluation contexts.
+    env_id: EnvironmentId,
+    /// Build info, used to build scoped evaluation contexts.
+    build_info: &'static BuildInfo,
     /// Frontend metrics.
     metrics: Metrics,
 }
@@ -71,13 +78,20 @@ impl SystemParameterFrontend {
             super::SystemParameterSyncClientConfig::File { path } => Ok(Self {
                 client: SystemParameterFrontendClient::File { path: path.clone() },
                 key_map: sync_config.key_map.clone(),
+                env_id: sync_config.env_id.clone(),
+                build_info: sync_config.build_info,
                 metrics: sync_config.metrics.clone(),
             }),
             SystemParameterSyncClientConfig::LaunchDarkly { sdk_key, now_fn } => Ok(Self {
                 client: SystemParameterFrontendClient::LaunchDarkly {
                     client: ld_client(sdk_key, &sync_config.metrics, now_fn).await?,
-                    ctx: ld_ctx(&sync_config.env_id, sync_config.build_info)?,
+                    // The environment-wide context carries no cluster/replica
+                    // scope. Scoped evaluation passes a `cluster` or `replica`
+                    // context per pass via [`ld_ctx`].
+                    ctx: ld_ctx(&sync_config.env_id, sync_config.build_info, None, None)?,
                 },
+                env_id: sync_config.env_id.clone(),
+                build_info: sync_config.build_info,
                 metrics: sync_config.metrics.clone(),
                 key_map: sync_config.key_map.clone(),
             }),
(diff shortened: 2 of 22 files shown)


src/adapter/src/catalog/apply.rs
```diff
@@ -524,6 +530,63 @@ impl CatalogState {
         }
     }
 
+    #[instrument(level = "debug")]
+    fn apply_cluster_system_configuration_update(
```
antiguru: Please add doc comments, here and below.

src/adapter/src/config/frontend.rs
```diff
@@ -271,5 +570,67 @@ fn ld_ctx(
...
+        let cluster = ClusterScopeContext {
+            id: "u1".into(),
+            name: "quickstart".into(),
+            is_builtin: false,
+        };
+        let replica = ReplicaScopeContext {
+            id: "u1-replica-1".into(),
+            name: "r1".into(),
+            is_builtin: false,
+            size: "D.1".into(),
```
antiguru: This would be incorrect, the size is something like `D.1-xsmall`. We have to avoid confusion here and elsewhere what family and size are.

src/adapter/src/catalog/apply.rs
antiguru: Done in 877ff6a2f8 — added doc comments to both `apply_{cluster,replica}_system_configuration_update`.

src/adapter/src/config/frontend.rs
antiguru: Fixed in 877ff6a2f8. You're right — the size is the full name (`D.1-xsmall`); `D` is the family. Changed the test to `size: "D.1-xsmall"`, `size_family: "D"`, and reworded the `ReplicaScopeContext`/`ReplicaAllocation::family` docs to state explicitly that the family is the coarse axis and not a prefix of the size name in general.

def- (changes requested):
Per-cluster LD feature flags in combination with a cluster-targeted optimizer flag crash:
Running `cargo test -p mz-repr --lib bool_override_decodes_lenient_spellings` with this diff:
```diff
diff --git a/src/repr/src/optimize.rs b/src/repr/src/optimize.rs
index 75d4a56c71..d0d3fcef05 100644
--- a/src/repr/src/optimize.rs
+++ b/src/repr/src/optimize.rs
@@ -189,3 +189,21 @@ macro_rules! impl_optimizer_feature_type {
 // Implement `OptimizerFeatureType` for all types used in the
 // `optimizer_feature_flags!(...)`  call above.
 impl_optimizer_feature_type![bool, usize];
+
+#[cfg(test)]
+mod tests {
+    use std::collections::BTreeMap;
+
+    use super::OptimizerFeatureOverrides;
+
+    #[mz_ore::test]
+    fn bool_override_decodes_lenient_spellings() {
+        for (stored, want) in [("true", true), ("false", false), ("on", true), ("off", false)] {
+            let map = BTreeMap::from([("enable_eager_delta_joins".to_string(), stored.to_string())]);
+            assert_eq!(
+                OptimizerFeatureOverrides::from(map).enable_eager_delta_joins,
+                Some(want),
+            );
+        }
+    }
+}
```
panics:
```
thread 'optimize::tests::bool_override_decodes_lenient_spellings' (4146015) panicked at src/repr/src/optimize.rs:191:1:
called `Result::unwrap()` on an `Err` value: ParseBoolError
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: <bool as mz_repr::optimize::OptimizerFeatureType>::decode
   4: <mz_repr::optimize::OptimizerFeatureOverrides as core::convert::From<alloc::collections::btree::map::BTreeMap<alloc::string::String, alloc::string::String>>>::from
   5: mz_repr::optimize::tests::bool_override_decodes_lenient_spellings::test_impl
   6: mz_repr::optimize::tests::bool_override_decodes_lenient_spellings
   7: mz_repr::optimize::tests::bool_override_decodes_lenient_spellings::{closure#0}
   8: <mz_repr::optimize::tests::bool_override_decodes_lenient_spellings::{closure#0} as core::ops::function::FnOnce<()>>::call_once
```

antiguru:
Good catch — fixed in `8ea690964b` (folded into the catalog/adapter commit). `<bool as OptimizerFeatureType>::decode` now accepts the canonical `on`/`off` spellings in addition to `true`/`false`, so a cluster-coherent override served from LaunchDarkly as `on`/`off` no longer panics the optimizer at plan time. Added your `bool_override_decodes_lenient_spellings` test (passes).

src/adapter/src/config/frontend.rs
```diff
@@ -144,6 +158,195 @@ impl SystemParameterFrontend {
...
+            // Compare in the parameter's canonical encoding. `base` is the
+            // var-formatted env-wide value (a `bool` is `"on"`/`"off"`), whereas
+            // the raw LaunchDarkly value spells a boolean `"true"`/`"false"`; a
+            // direct string compare would treat every boolean flag as differing,
+            // even on `Fallthrough`. We still *store* the raw `value` (downstream
+            // consumers parse `"true"`/`"false"`); only the decision is canonical.
+            let differs = match params.canonicalize(param_name, &value) {
+                Some(canonical) => canonical != base,
+                // Unparseable for this var: fall back to a raw comparison.
+                None => value != base,
```
DAlperin: There is a risk here. If LD gives us a badly typed response for a bool value (lets say `maybe`) canconicalize will return false `maybe` != `off` so we store `maybe` but then `bool::decode` will panic for every query plan I think. Probably better to just do None here.

DAlperin (approved):
LGTM, modulo one fix

aljoscha (commented):
Reviewed the top commit only (`d77eb1bc`, the resolution layer), per the stack note.

Overall this is a solid, cohesive change and the correctness reasoning holds up well: the differs-from-env recording rule, the canonical bool bridging so `on`/`off` vs `true`/`false` don't register as spurious overrides, the lenient optimizer `decode` guarded by `canonicalize` so a stored value can't panic the optimizer, the create-time resolution ordered before render-freeze, and the revert-via-full-`compute_config`-re-push are all sound. The end-to-end LD test covers the interesting paths (differs-from-env guard, recreate-by-name, bidirectional reconcile, gating, durability, create-time). No blocking correctness issues found.

A few things to clean up, mostly around our writing/comment conventions:

**Strong suggestions**

1. **Em-dashes and semicolons in comments.** Our style is to avoid em-dashes and sentence-structuring semicolons everywhere, including code comments. This commit adds ~22 lines with em-dashes and a similar number of structuring semicolons (heaviest in `coord.rs` and `config/frontend.rs`). Please do a pass replacing them with full stops and commas. Inline suggestions below mark representative spots.

2. **Doc comments that narrate / duplicate the body.** `evaluate_scoped_overrides` is the clearest case: the differs-from-env-vs-`variation_detail` rationale is spelled out almost verbatim in both the doc comment and the inline comment at the decision point. Our convention is for the doc comment to state the contract and for the reasoning to live inline at the decision point. Trim the doc comment (suggestion inline).

3. **`scoped_ld_ctx` is a dead public wrapper.** It only forwards to the private `ld_ctx`, and the real callers (`pull_cluster_overrides` / `pull_replica_overrides`) already call `ld_ctx` directly. Its only users are the unit tests, which live in the same module and can call `ld_ctx` too. Suggest dropping both the wrapper and its `pub use` re-export.

**Nits**

- `push_replica_dyncfg_overrides`: the "compute push reaches storage via the shared persist `ConfigSet`" subtlety is genuinely worth recording, so keep it, but the doc comment is long and could be tightened (and is part of the em-dash pass).
- Consider a unit-level test for the `evaluate_scoped_overrides` differs-from-env decision itself. The `canonicalize` and `decode` units are good, but the core recording decision is currently only exercised by the live-LD mzcompose workflow. Not blocking given the LD client dependency.

src/adapter/src/config/frontend.rs
```diff
@@ -144,6 +158,201 @@ impl SystemParameterFrontend {
...
+    /// The difference — not the `variation_detail` reason — is the signal that
+    /// the scope context produced a genuine opinion. The reason cannot say which
+    /// context kind's clause matched (an env-level rollout rule and a
+    /// cluster-specific rule both report `RULE_MATCH`), and `FALLTHROUGH` serves
+    /// the env-wide value to every object; either would record a row for every
+    /// object (dense) and let a coarse env-wide value override a finer manual
+    /// `CREATE CLUSTER ... FEATURES` pin. Comparing against the env-wide baseline
+    /// captures exactly "this scope context changed the answer". A silent LD
+    /// (flag absent / off / error / failed prerequisite) resolves to the `base`
+    /// default and is dropped. See the scoped feature flags design, §Resolution.
```
aljoscha: The differs-from-env rationale here is duplicated almost verbatim by the inline `// Record iff ...` comment at the recording decision below. Per our convention the doc comment states the contract and the reasoning lives inline at the decision point, so this paragraph can go. (It also carries several em-dashes/semicolons.) Suggested trim:

```suggestion
    /// Evaluates each of `param_names` against `ctx`, returning only the values
    /// that differ from the environment-wide value held in `params`. Shared by
    /// the cluster and replica passes, so the returned map is sparse.
    ///
    /// We record on the differs-from-env test, not the `variation_detail`
    /// reason. The inline comment at the recording decision explains why.
```

src/adapter/src/config/frontend.rs
```diff
@@ -209,9 +418,109 @@ async fn ld_client(
...
+
+/// Builds a multi-context for evaluating scoped feature flags.
+///
+/// Composes the base contexts (`environment` + `organization` + `build`) with:
+/// - a `cluster` context for cluster-coherent (replica-free) resolution, and/or
+/// - a `replica` context for replica-local resolution.
+///
+/// The environment-wide pass passes `None` for both. This is the single entry
+/// point the sync loop uses to evaluate each scoped pass.
+pub fn scoped_ld_ctx(
```
aljoscha: `scoped_ld_ctx` only forwards to the private `ld_ctx`, and the real callers (`pull_cluster_overrides` / `pull_replica_overrides`) already call `ld_ctx` directly. Its only users are the unit tests in this module, which can call `ld_ctx` directly too. Suggest dropping this wrapper and the `pub use frontend::{... scoped_ld_ctx}` re-export in `config.rs`.

src/adapter/src/coord.rs
```diff
@@ -2013,6 +2033,222 @@ pub struct Coordinator {
...
+    /// `handle_update_configuration` applies the pushed dyncfg updates both to
+    /// compute's own worker `ConfigSet` *and* to the shared persist client
+    /// `ConfigSet` (`persist_clients.cfg()`), which the co-located storage server
+    /// reads from the same `Arc`. So persist-backed and process-global
+    /// replica-local configs — the design's examples: the persist pager, LZ4,
+    /// persist client tuning, `lgalloc` — take effect on storage as well, even
+    /// though the push goes through the compute controller. The only thing this
+    /// would miss is a future `Replica`-scoped config realized *solely* in the
+    /// storage worker's own `ConfigSet` (applied only via the storage
+    /// controller's `UpdateConfiguration`); none exists today.
```
aljoscha: nit: the "compute push reaches storage via the shared persist `ConfigSet`" subtlety is worth recording, so keep it, but this doc comment is long and stacks several em-dashes/semicolons. Consider tightening it and folding the example list into prose as part of the style pass.

src/adapter/src/config/frontend.rs
antiguru: Applied your suggested trim — the doc now states the contract and points at the inline comment for the differs-from-env reasoning.

src/adapter/src/config/frontend.rs
antiguru: Dropped the `scoped_ld_ctx` wrapper and its `config.rs` re-export; the unit tests call the private `ld_ctx` directly (same module), and its doc moved onto `ld_ctx`. Real callers were already on `ld_ctx`.

src/adapter/src/coord.rs
antiguru: Tightened: folded the example list into prose, cut the em-dashes/semicolons, kept the shared-persist-`ConfigSet` storage-reach subtlety.

src/adapter/src/config/frontend.rs
aljoscha: Confirmed. The doc now states the contract and defers to the inline comment for the reasoning. Thanks.

src/adapter/src/config/frontend.rs
aljoscha: Confirmed. The wrapper and the re-export are gone, the tests call `ld_ctx` directly, and the stray `[`scoped_ld_ctx`]` doc reference was updated too. Thanks.

src/adapter/src/coord.rs
aljoscha: Confirmed. Reads well now and keeps the shared-persist-`ConfigSet` storage-reach subtlety. Thanks.

aljoscha:
Re-reviewed the responses. The three inline fixes all look good, including the follow-on cleanup of the stray `scoped_ld_ctx` doc reference.

One item from the review summary is still open: the broader em-dash and structuring-semicolon pass across the new comments. The `coord.rs` doc comment I flagged inline is clean now, but the rest of the new comments still carry them. By a quick count there are about 19 added lines with em-dashes and about 17 comment lines with structuring semicolons. Examples: the `// Record iff ...` block in `config/frontend.rs`, the `scoped_frontend` field doc and the `resolve_scoped_for_new_objects` / `push_new_replica_scoped_override` docs in `coord.rs`, and the `ReplicaScopeContext::size_family` and `canonicalize` docs. It is a mechanical pass, flagging only because it is a project-wide convention. Happy to list the exact lines if that helps.

antiguru:
Done the em-dash and structuring-semicolon pass over the new comments. Removed the em-dashes (reworded into separate sentences or commas) and the structuring semicolons (split into sentences) across `config/frontend.rs`, `config/params.rs`, `config/sync.rs`, `command.rs`, `coord.rs`, and `test/launchdarkly/mzcompose.py` — including the `// Record iff ...` block, the `scoped_frontend` field doc, the `resolve_scoped_for_new_objects` / `push_new_replica_scoped_override` docs, and the `ReplicaScopeContext::size_family` / `canonicalize` docs you called out. The remaining em-dashes in `client.rs` / `coord/introspection.rs` / `frontend_peek.rs` are pre-existing lines this PR does not touch, so I left them.

On the non-blocking suggestion: I agree a direct unit test of the differs-from-env decision would be nice, but `evaluate_scoped_overrides` takes a live `ld::Client` with no seam to inject canned variations, so it would need a fake-client shim that does not exist yet. The decision's two halves are unit-covered (`canonicalize` bridging and the lenient `decode`), and the end-to-end behavior is exercised by the `test/launchdarkly` workflow. Happy to add the shim as a follow-up if you think it's worth the surface.

aljoscha:
Did a focused performance pass on the hot paths and continual triggers. Summary: the read/plan path is fine, the costs are all on the write side and only when the feature is enabled.

**Plan / peek path: no concern.** `cluster_scoped_optimizer_overrides()` is added to every optimizer-config build (peek, MV, index, subscribe, introspection, fast-path peek). With no override for the target cluster (always the case while gated off) it is one miss on an empty `BTreeMap` plus a default 19-field `OptimizerFeatureOverrides` and 19 failed `override_from` branches. No allocation, no locking. Good.

**`push_new_replica_scoped_override` evaluates and discards all cluster overrides (every `CREATE ... REPLICA`, when on).** It calls `evaluate_scoped_parameters(.., None, Some(&replicas))` with `cluster_filter = None`, so it runs an LD pass over every cluster's cluster-coherent flags and then uses only `evaluated.replica`. The cluster-create path passes an empty set for the mirror case. Suggest passing `Some(&BTreeSet::new())` as the cluster filter here, so a replica creation does not run and throw away an O(#clusters x #cluster-params) LD evaluation.

**`push_new_replica_scoped_override` re-pushes the full env-wide compute config unconditionally (every `CREATE ... REPLICA`, when on).** `push_replica_dyncfg_overrides_from` always calls `update_replica_dyncfg_overrides` then `update_configuration(compute_config)`. `update_configuration` clones the full dyncfg `ComputeParameters` per instance and broadcasts to every replica (`controller.rs:603`), the same cost as an env-wide `ALTER SYSTEM` push. It fires even when the new replica has no override (the common case), and it runs synchronously on the coordinator thread. Creating a managed cluster with N replicas triggers N such re-pushes. Suggest an early-out when the new replica contributes nothing, e.g. `if evaluated.replica.is_empty() { return; }`.

That early-out also removes the common case of a subtle transient: replica create-time overrides are not written back to the catalog working copy (in-memory/controller only, by design), so `merged` is rebuilt from the stale working copy. If replica A was given a create-time override and replica B is then created before the next reconcile, the push for B rebuilds from the working copy that does not contain A, momentarily clearing A's override at the controller until the next tick. Bounded by the 1s reconcile window and harmless for render-frozen flags, but worth being aware of for replica-local configs that take effect on update.

**Continual trigger: per-tick O(fleet) evaluation when enabled.** With the feature on, `sync_scoped_params` runs every tick (1s in prod) and rebuilds eval contexts for every cluster and replica (several `String` allocations each) and runs an LD `variation` per object per scoped param, regardless of whether anything changed. This is on the dedicated sync-loop task so it does not block the coordinator, but it is continual O(#clusters + #replicas) CPU and allocation churn every second. Fine for small fleets, worth keeping in mind for large ones. The coordinator-side per-tick cost is just the `CatalogSnapshot` Arc-clone plus the `UpdateScopedSystemParameters` equality early-out, both negligible.

**Minor:** `SynchronizedParameters::new(catalog.system_config().clone())` on the create-cluster and create-replica paths deep-clones the whole `SystemVars` (~450 vars) and builds a ~450-element `BTreeSet`. Only on those DDL paths and only when enabled, so acceptable, just noting it is not a cheap call.

Net: nothing on the read path, and the write-path costs are gated off by default. The two worth acting on are the discarded cluster evaluation and the unconditional full compute re-push on replica creation.

antiguru:
Thanks for the perf pass. Acted on the two write-path items in `push_new_replica_scoped_override`:

* **Discarded cluster evaluation:** now passes `Some(&BTreeSet::new())` as the cluster filter, so a `CREATE ... REPLICA` runs only the replica-local pass and no longer evaluates+throws away the O(#clusters x #cluster-params) cluster-coherent LD pass.
* **Unconditional full compute re-push:** added an early-out — `if evaluated.replica.is_empty() { return; }` — before the merge and push. The new replica having no override is the common case, so we skip the per-instance `update_configuration` broadcast entirely. As you noted, this also removes the common-case transient where rebuilding the pushed map from the working copy (which does not carry other replicas not-yet-reconciled create-time overrides) momentarily clears them at the controller. The rare has-override case keeps the current behavior, with that transient bounded by the 1s reconcile window.

On the rest, agreed and no change:

* Plan/peek path: as you found, gated-off is an empty-map miss + a default-overrides no-op. Nothing to do.
* Per-tick O(fleet) evaluation when enabled: real, but on the dedicated sync-loop task and only when the feature is on. Noted for large fleets; if it bites we can diff-and-skip unchanged objects or widen the tick.
* `SynchronizedParameters::new(... .clone())` on the DDL paths: only on create-cluster/create-replica and only when enabled. Acceptable, noted.

aljoscha:
Follow-up on the per-tick note, on whether this is new work or just more of what the sync loop already did. It is net-new, and it runs unconditionally, so worth calling out explicitly.

Before this feature, the sync loop did two things per tick: `backend.pull` to fetch current values from the coordinator, and `frontend.pull` to evaluate each synced param against the single env-wide LD context. That env-wide pass is `O(#synced params)` against one context, and it only does the downstream `backend.push` when a value actually changed. No catalog snapshot, no per-object work.

This PR adds, on every tick:

* an unconditional `catalog_snapshot().await`, and
* when enabled, `evaluate_scoped_parameters` which builds eval contexts for every cluster and replica (`O(#clusters + #replicas)`, several `String` allocations each) and runs an LD `variation` per object per scoped param (`O(#objects x #scoped-params)`), followed by an `update_scoped_system_parameters().await`.

So this is not the env-wide pass getting a bit bigger. The object factor is entirely additional. And unlike the env-wide pass, which gates its expensive work on "did anything change", the scoped pass runs the full fleet evaluation every tick regardless. The coordinator early-outs on the equality check in `reconcile_scoped_system_parameters`, but the sync-task-side evaluation does not, so we redo the whole `O(#objects x #scoped-params)` LD pass and the per-object context allocations once per second whether or not anything changed.

It is on the dedicated sync-loop task so it does not block the coordinator, and it is gated off by default. But it is continual extra work proportional to fleet size, so worth a deliberate look before this goes on by default, especially for large environments. A cheap guard would be to skip the pass when neither the set of live objects nor the relevant LD state has changed since the last tick, though detecting the latter is the hard part.

antiguru:
Agreed, the object factor is net-new and the scoped pass runs every tick regardless of change. Two parts here:

The off-path should not regress, since a patch release to undo it would be painful. Fixed in 210c04ff30: the scoped reconcile now reads the gate from the sync working copy (`params`) instead of taking a `catalog_snapshot().await` every tick, so a disabled environment (the default) does no per-tick coordinator round-trip. A small `maybe_dirty` flag still does the one-time clear of any leftover overrides on startup and on an enable->disable transition, then goes quiet.

The on-path full-fleet evaluation is the part worth optimizing, and since this will be on by default we will want it. I would rather do that in a follow-up than widen this PR: a cheap object-set/env-change guard is unsafe on its own (a scoped LD rule can change with neither the live-object set nor the env-wide context changing, which would silently freeze scoped overrides), so the real fix is either a dedicated slower scoped tick or hooking LD streaming changes. Tracking that separately.

src/adapter/src/catalog/transact.rs
```diff
@@ -2732,10 +2752,10 @@ impl Catalog {
...
                     }
                 }
-                // Non-live clusters never enter `desired_cluster` (they `continue`
-                // above), so a missing entry already covers dropped clusters.
                 for (cluster_id, name) in existing_cluster.into_keys() {
-                    if !desired_cluster.contains(&(cluster_id, name.clone())) {
+                    if prune_cluster(&cluster_id)
+                        && !desired_cluster.contains(&(cluster_id, name.clone()))
+                    {
                         tx.remove_cluster_system_config(cluster_id, &name);
```
aljoscha, line 2774: Blocking: this `prune_scope` change drops the lazy GC of orphaned rows for dropped objects, with a few knock-on effects.

The sync loop builds `prune_scope` from `catalog.clusters()`, so it only ever contains live objects. A dropped cluster or replica is never in `prune_scope`, so `prune_cluster` / `prune_replica` returns false for it and its rows are never removed. Nothing else GCs these rows either: `remove_clusters` / `remove_cluster_replicas` cascade only to `cluster_replicas` and `introspection_sources`, and the only callers of `remove_*_system_config` are right here. The previous code pruned any row not in the desired state, which lazily covered dropped objects (the old comment said exactly that).

Consequences:
1. Durable and introspection leak. `mz_cluster_system_parameters` / `mz_replica_system_parameters` are unfiltered projections of these collections, so orphan rows with a dangling `cluster_id` / `replica_id` accumulate until the feature is toggled off (the only full wipe).
2. The per-tick early return is defeated. `reconcile_scoped_system_parameters` skips the durable write when `working_copy == scoped`, but the sync loop's `scoped` is live-only while the working copy retains orphans, so they can never be equal once an orphan exists. The coordinator then runs a `catalog_transact` every tick that can never converge.
3. The recreate-then-clear case in `test/launchdarkly/mzcompose.py` should fail: after DROP plus CREATE `ld_role`, clearing the rules and asserting `count(*) ... = 0` will see the orphan row from the dropped incarnation. I could not run it locally since it needs a live LD project.

Suggested fix, local to this handler: also prune rows whose owning object is not live. `live_clusters` / `live_replicas` are already computed just above, so something like `(!live_clusters.contains(&cluster_id) || prune_cluster(&cluster_id)) && !desired_cluster.contains(...)` (and the replica analogue). That restores dropped-object GC while keeping the concurrent-create protection that `prune_scope` was added for. Orphan pruning is always safe since ids are never reused.

test/launchdarkly/mzcompose.py
```diff
@@ -242,6 +278,306 @@ def sys(command: str) -> None:
...
+            # *remove* an override once LD no longer serves a differing value.
+            # Clear the rules so every cluster falls through to env-wide. All
+            # rows for the parameter disappear. Restoring the rules brings them
+            # back, exercising propagation in both directions.
+            ld_client.update_targeting(LD_OPTIMIZER_FLAG_KEY, rules=[])
+            sleep(5)
+            c.testdrive(
+                "\n".join(
+                    [
+                        f"> SELECT count(*) FROM mz_internal.mz_cluster_system_parameters WHERE name = '{OPTIMIZER_PARAM}'",
```
aljoscha, line 448: This `count(*) = 0` assertion is where the orphan-GC regression surfaces. After the DROP plus CREATE of `ld_role` above, the dropped incarnation's row is not pruned by the reconcile, since its id is no longer in `prune_scope`, so this count should be 1, not 0. See the comment on `src/adapter/src/catalog/transact.rs`. Worth re-running against the live LD project once the prune fix is in.

aljoscha (commented):
Reviewed the resolution and gating layer locally against main. Overall the design is solid and the comments are unusually good. One blocking correctness item from the new `prune_scope` rework, plus a couple of suggestions. Main item is inline on `catalog/transact.rs`, the rest is below.

## Blocking

See the inline comment on `src/adapter/src/catalog/transact.rs`: dropped clusters and replicas leak their scoped-config rows, and as a knock-on the periodic sync stops converging. Details and a suggested fix are there.

## Strong suggestions

- No Rust unit test covers the `UpdateScopedSystemParameters` diff and prune semantics. The only coverage is the `test/launchdarkly` mzcompose workflow, which needs live credentials and is not in standard PR CI. A catalog-level unit test exercising upsert, remove, `prune_scope` bounding, and the dropped-object case would have caught the blocking item and guards this boundary going forward. The other new logic (canonicalize, lenient decode, context building) is unit tested, which is great.
- The cluster-scoped optimizer override is wired in ad hoc at around a dozen plan-path call sites via `.override_from(&self.cluster_scoped_optimizer_overrides(cluster_id))` (peek, index, MV, subscribe, introspection, bootstrap, frontend_peek). It looks complete today, but a future plan path that builds a per-cluster `OptimizerConfig` will silently miss the scoped layer. Consider a single helper that assembles the per-cluster features (system config plus `config.features()` plus scoped overrides) so the precedence lives in one place. Partly pre-existing, since `config.features()` is already repeated, but this PR roughly doubles the override points.

## Nits

- `bool::decode` in `src/repr/src/optimize.rs` still panics on truly invalid input, even though the doc says decoding must not crash the optimizer. It is effectively unreachable, since `canonicalize` gates what gets stored and the `FEATURES` clause is bool typed, so it is a defensive assertion rather than a real hazard. The wording just oversells the guarantee.
- Replica create-time resolution is not done here. Only the cluster paths call `resolve_scoped_for_new_objects`, always with an empty replica set, so a freshly created replica's render-frozen flags use env-wide values until the next sync tick. The description documents this as deferred to #37158. Fine given the gate, just calling it out.

## What is good

- The differs-from-env recording rule with canonical bool bridging is the right call and is well explained and tested.
- The gating works as intended: the loop reads the gate from `params`, and `backend.pull` seeds `params` from the catalog system vars every tick, so `ALTER SYSTEM SET` and `additional_system_parameter_defaults` are observed even though the gate is not an LD flag.
- The `prune_scope` protection for concurrently created objects is sound, and the create-time `merge` correctly preserves existing overrides.
- Create-time resolution does a local in-memory LD evaluation only, with no network round-trip on the coordinator loop, so it is cheap.

aljoscha:
Thanks for the quick turnaround. The rebased revision (926f50110c) resolves the blocking item: the prune loop now reclaims orphaned rows for dropped objects (`!live_clusters.contains(..) || prune_cluster(..)`), and the new `test_transact_update_scoped_system_parameters_prune` covers upsert, bounded prune, stale in-scope prune, and the orphan case. That also clears my note on the recreate-then-clear assertion in the launchdarkly test. (My earlier inline comments are now outdated against this push.)

Remaining items, all non-blocking:

- The cluster-scoped optimizer override is still applied ad hoc at roughly a dozen plan-path call sites. Not a problem today, but a future plan path that builds a per-cluster `OptimizerConfig` can silently miss the scoped layer. A single per-cluster feature-assembly helper would keep the precedence in one place. Fine as a follow-up.
- `bool::decode` in `repr/optimize.rs` still panics on truly invalid input even though the doc says it must not crash the optimizer. It is effectively unreachable since `canonicalize` gates what gets stored, so this is just a wording nit.
- Replica create-time resolution remains deferred to #37158, as documented.

LGTM once you are happy with those. Nice work on the orphan-prune test in particular.

aljoscha:
## Review: resolution and gating

Overall this is a careful, well-documented PR. It's the resolution layer on top of an already-merged stack, gated behind `enable_scoped_system_parameters` (off by default). The hot-path and continual-path additions are minimal, the central race-protection mechanism (`prune_scope`) is sound and tested, and the code largely follows the adapter guide's invariants. I found no hard blockers for a gated, stacked increment, but there are a few correctness notes worth addressing.

### Context verified
- The in-memory working copy (`CatalogState::scoped_system_parameters`) is updated **per-row from the durable diff** (`apply_scoped_system_configuration_update`), not by wholesale replacement with the pushed `scoped`. This is the invariant that makes everything else hold: in-memory always mirrors the durable cache, so the `prune_scope` sparing of out-of-snapshot rows can't desync the two.
- The `bool` optimizer decode accepts a strict **superset** of what the system-var bool parser accepts (`t/true/on`, `f/false/off`), and is case-insensitive, so any value that passes `canonicalize` (and is therefore stored) is always decodable. No new panic surface versus the prior `str::parse().unwrap()`.

### Correctness — most important to flag

1. **Inline replicas of a newly-created managed cluster get no create-time replica resolution (render-frozen gap when enabled).** `sequence_create_cluster` (`cluster.rs:747`, `:954`) calls `resolve_scoped_for_new_objects(&{cluster_id}, &BTreeSet::new())` — empty replica set — even though the managed path creates replicas in the same transaction (`cluster.rs:719-739`). With the feature ON, those replicas hit `create_replica` (via the implication) with env-wide config, and their replica-local overrides only land on the next sync tick. For render-frozen flags (e.g. `enable_lgalloc`) a later push doesn't take effect until replica restart, so the value is wrong for the replica's lifetime. This matches exactly the bug described in `guide-adapter.md` ("The catalog is the source of truth..."). The PR description and the guide both say #37158 is the fix and the feature is off by default, so this is acceptable for a stacked increment — but it's the thing to make sure lands before the gate is flipped. Worth a `TODO` pointer at the call site rather than only in the PR description.

2. **Misleading comment at `cluster.rs:743-746`.** It claims "The durable write drives the replica-scoped controller push through the catalog implication," but this create-cluster path writes only the cluster-coherent override and never resolves replicas, so no replica push happens here. The comment overstates what this path does (see point 1). Please correct it.

### Strong suggestions

3. **`bool::decode` hard-panics on the plan path (`repr/optimize.rs:207`).** The comment motivating the lenient decode says decoding "must not crash the optimizer at plan time," yet the fallthrough still `panic!`s. The guard that prevents bad values (the `canonicalize` → `None` skip) lives in a *different crate* (adapter config), so the "stored values are always decodable" invariant is enforced at a distance. If any future writer ever persists a cluster override without going through `evaluate_scoped_overrides`, every query on that cluster panics at plan time — a severe availability hit. Consider `soft_panic_or_log!` plus a safe fallback for defense in depth. It matches the prior `.unwrap()` behavior, so this isn't a regression, just a chance to harden a critical path that this PR newly feeds from a remote source.

4. **`reconcile_scoped_system_parameters` doc says "the sync loop is the sole writer" (`coord.rs:2040`, `:2064`), but `resolve_scoped_for_new_objects` now also writes through it.** Both run serially on the coordinator loop and the cross-tick race is handled by `prune_scope`, so it's correct — but the doc is now inaccurate. Reword to something like "the sync loop and the create-time resolve are the only writers, both on the coordinator loop."

### Continual / hot path — looks good, with notes

5. **Plan path:** every optimize now adds `.override_from(&cluster_scoped_optimizer_overrides(id))`. When the feature is off (or the cluster has no override) this is a single `BTreeMap` lookup returning `None` + a default + a no-op `override_from`. Negligible, and only allocates (clone + decode of a small map) when a cluster actually has overrides. Fine for peek / `frontend_peek`.

6. **Sync loop when disabled (default):** after one clearing push, subsequent ticks do zero extra work and take no catalog snapshot. Good, and matches the gate intent.

7. **Sync loop when enabled:** each tick does two coordinator round-trips (`catalog_snapshot` + `update_scoped_system_parameters`) plus a full LD evaluation of every cluster and replica. The LD eval is off the coordinator loop (good), but a per-tick coordinator command happens even in steady state (the equality check then skips the durable write). With a 1s tick and large environments this is non-trivial but bounded and only when enabled. Worth keeping in mind for rollout.

### Nits

8. `build_replica_eval_contexts` iterates all clusters/replicas even when `filter` is `Some(empty)` (the create-cluster case), just to skip them all. A cheap early-return when the filter is empty would avoid an O(replicas) loop on the DDL path.
9. The `src/repr/src/optimize.rs` `bool` change crosses a CODEOWNERS boundary from the adapter work. Small and tightly coupled, so probably fine, but flagging per the one-team-scope guideline.

### Tests
Good coverage of the tricky parts: the prune handler's four behaviors (incl. orphan and bounded-prune), canonical bool bridging, lenient decode, and context construction. The end-to-end LaunchDarkly workflow exercises the differs-from-env recording rule, bidirectional reconcile, gating, durability, and create-time — though it needs a live LD project, so it won't run in normal CI. No Rust-level test for the `reconcile` / `resolve_scoped_for_new_objects` merge + short-circuit logic (needs a coordinator), which is an acceptable gap.

antiguru:
Thanks for the thorough re-review. Addressed the items in `44e8c4dd9a`:

1. **Render-frozen gap / create-time replica resolution** — added a `TODO` at the create-cluster call site spelling out that this path resolves only the cluster-scoped overrides, that inline replicas reach the controller with env-wide config and pick up replica-local overrides on the next tick (too late for render-frozen flags), and that the follow-up #37158 folds replica resolution into the create transaction.
2. **Misleading comment** — dropped the "the durable write drives the replica-scoped controller push" sentence from the create-cluster comment, since this path resolves no replicas. The comment now states only what it does (cluster-scoped overrides for the first plan).
3. **`bool::decode` hard-panic** — replaced the fallthrough `panic!` with `mz_ore::soft_panic_or_log!` and a `false` fallback. The `canonicalize` gate still makes the arm unreachable in practice, but a stored bad value now logs instead of crashing the optimizer on every plan for the affected cluster.
4. **Stale "sole writer" doc** — reworded `reconcile_scoped_system_parameters` (doc + inline) to say the sync loop and the create-time resolve are the only writers, both serialized on the coordinator loop.
8. **`build_replica_eval_contexts` nit** — early-return on a `Some(empty)` filter, so the create-cluster path skips the O(replicas) scan.

On 9 (CODEOWNERS): the `repr/optimize.rs` change stays minimal and tightly coupled to the override decode, so I kept it here rather than splitting it out.

aljoscha (approved):
Thanks, the revision in `44e8c4dd9a` addresses all the items. Verified the two code fixes too: the `soft_panic_or_log!` + `false` fallback is the right defense-in-depth (still panics under soft-asserts in CI, degrades gracefully in prod), and the empty-filter early-return is behavior-preserving. Nothing further from me.

Note for unblocking: `def-`'s changes-requested was the same optimizer-panic crash, which is fixed here — probably worth a re-review to clear it.

def- (approved):
Would be nice to add e2e tests for replica flags, not just cluster flags.

I think it's easy to misuse atm because we ignore misscoped variables instead of causing a failure that would make you notice that the flag you set won't apply.

antiguru:
Thanks! Both points are good and I'd like to handle them as follow-ups rather than widen this PR:

* **e2e tests for replica flags.** Agreed. Replica create-time resolution lands in #37158 (this PR only does the cluster paths with an empty replica set), so the natural home for replica-scoped e2e cases is there, alongside the existing cluster/durability/gating cases in the `test/launchdarkly` workflow. I'll extend it there. The one caveat is that the workflow needs a live LaunchDarkly project, so it doesn't run in standard PR CI.
* **Misscoped variables are silently ignored.** Strong agree this is a footgun: setting a cluster-only flag with a replica scope (or vice versa) currently just does nothing, with no signal. I'd rather make that loud (reject or at least warn at set time) than leave it silent. I'll file a follow-up for it since it's a separate, cross-cutting change from the resolution layer here.

Will link the two follow-up issues here once filed.

antiguru:
Thanks for the reviews! Tracking the e2e test as a follow-up in [DB-151](https://linear.app/materializeinc/issue/DB-151/consolidate-scoped-feature-flag-eval-helpers-scope-partition)

PR #36961, opened 2026-06-10
[ss-104] Mysql drop create corruption testdrive

peterdukelarsen:
### Motivation

Splits out tests from https://github.com/MaterializeInc/materialize/pull/36932 to simplify the diff.

### Description

 At a high level we're taking care to validate that we're able to handle all forms of DROP TABLE that are thrown at us. Tests are split out into:
1. Validating concurrency handling is correct now (drop-recreate-table.td)
2. Validating the new parsing/handling of DROP TABLE events works correctly for varied statements - i.e. multiple tables in the statement, various quoting/escaping schemes, etc.... This is being done in all other test files.


test/mysql-cdc/drop-table-detection.td
```diff
@@ -0,0 +1,590 @@
...
+# By default MySQL quotes identifiers with backticks and treats `"..."` as a
+# string literal. A session whose sql_mode includes ANSI_QUOTES flips this --
+# `"` becomes the identifier quote -- so the client below writes the DDL with
+# double quotes. The detector parses the binlog with a backtick-only dialect, so
+# this would break it IF the double quotes reached the binlog. They do not:
+# MySQL re-quotes the identifiers with backticks when it regenerates the
+# DROP TABLE into the binlog, regardless of the session's sql_mode (the same
+# binlog normalization that strips comments in section 9). This test pins that
+# behavior -- it would fail loudly if a future MySQL preserved the ANSI_QUOTES
+# `"` quoting in the binlog.
```
martykulma: this seems to contradict the unit test [ansi_quotes_parsed_properly](https://github.com/MaterializeInc/materialize/blob/b0754e44ef2ad644491781468cdbafb23b1cc343/src/storage/src/source/mysql/replication/events.rs#L643), which validates that if a double quoted identifier came through, we would parse it correctly.

test/mysql-cdc/drop-table-detection.td
```diff
@@ -0,0 +1,590 @@
...
+
+# The sibling is unaffected and continues to replicate new rows.
+> SELECT * FROM sel_keep;
+1
+2
+
+> DROP SOURCE mz_source CASCADE;
+
+# 9. Comments in awkward positions must not cause missed drop. This test is
+# more testing mysql's binlog replication scrubbing comments than excercising Materialize
```
martykulma: `s/excercising/exercising/`

test/mysql-cdc/mzcompose.py
```diff
@@ -176,6 +182,25 @@ def workflow_replica_connection(c: Composition, parser: WorkflowArgumentParser)
         )
 
 
+def workflow_cdc_ansi_quotes(c: Composition, parser: WorkflowArgumentParser) -> None:
```
martykulma: MySQL uses the global sql mode when a client connects - `SET GLOBAL` and `--sql-mode` result in the same observed behavior. If this behavior changed it would have an adverse effect on a large MySQL client population.

I would drop this workflow and adjust `ansi-quotes.td` to `SET GLOBAL` before MZ establishes the connection. The `SET SESSION` test seems to cover the case where MZ uses one sql mode, and the user/application uses a different one.

martykulma (approved):
Nice - thanks @peterdukelarsen!

Nothing block, minor cleanup (comments and some duplicate test coverage).

test/mysql-cdc/drop-table-detection.td
peterdukelarsen: Ah, good catch -- it does contradict and this is a stale comment from my assumption that the parser I pulled in wouldn't handle it.

PR #36962, opened 2026-06-10
expr: fold identity-operand and involution scalar calls in reduce

frankmcsherry:
## Summary

A batch of small scalar simplifications in `MirScalarExpr::reduce`, each factored as a self-contained arm in the post-order reducers, none touching existing dispatch logic:

- **Binary identity operands** (`reduce/binary.rs`): a data table of (function, identity literal, commutes) triples. Covers integer and unsigned `a+0`, `0+a`, `a-0`, `a*1`, `1*a`, `a/1`; bitwise `a|0`, `a#0`, `a&-1` (and unsigned all-ones), `a<<0`, `a>>0`; `btrim`/`ltrim`/`rtrim` with an empty character set; and `±INTERVAL '0'` on timestamp/timestamptz/time/interval. The call folds to its surviving operand.
- **Unary involutions** (`reduce/unary.rs`): a single metadata-driven arm, `f(g(x)) → x` when `g` advertises `f` as its inverse and `g` preserves uniqueness — per the `inverse()` contract, exactly the condition under which the advertised function is a *left* inverse — gated on both calls being infallible and null-propagating. Covers `NOT(NOT(x))`, `~(~x)`, `reverse(reverse(x))` (via a newly declared self-inverse on `Reverse`), numeric `-(-x)`, and infallible bijective cast roundtrips such as `bool::int4::bool`. An audit of all 146 `inverse` declarations (plus the manual impls) confirms every pair admissible under this gate is a genuine left inverse; the fallibility gates exclude every narrowing/parsing cast and integer `-(-x)` (which errors on the minimum value).
- **Variadic** (`reduce/variadic.rs`): `greatest`/`least` drop literal `NULL` operands (both ignore null inputs) and unwrap single survivors; two-argument `substr(s, 1)` folds to `s`.

## Soundness discipline

Every fold preserves error and null behavior exactly; functions whose evaluation can error even at their identity are deliberately excluded, with reasoning in comments: `s || ''` and `repeat(s, 1)` re-validate operand length (eliding them would suppress `LengthTooLarge`), float additive identities are inexact (`-0.0 + 0.0` is `+0.0`), numerics await an answer on result scale, and constant-result folds (`a%1`, `a-a`, annihilators) need null guards and are left for a follow-up.

One concrete motivation: the cardinality expressions synthesized by `ProjectionPushdown`'s unused-`generate_series` collapse (#36776) contain a division by a literal step, so `((y - x) / 1) + 1` plans now simplify to `((y - x) + 1)`.

## Testing

- Unit tests in each reducer file pin the folds and the deliberate non-folds (including the audit-derived boundary: numeric `-(-x)` and `bool::int4::bool` fold; integer `-(-x)` and `int4::int8::int4` do not).
- New `test/sqllogictest/scalar_identity_folds.slt` asserts optimized plans and results (including NULL rows) for both directions of the boundary.

### DIFF
--- a/src/expr/src/scalar/func/impls/string.rs
+++ b/src/expr/src/scalar/func/impls/string.rs
@@ -158,11 +158,7 @@ fn cast_string_to_uint64(a: &str) -> Result<u64, EvalError> {
     strconv::parse_uint64(a).err_into()
 }
 
-#[sqlfunc(
-    sqlname = "reverse",
-    preserves_uniqueness = true,
-    inverse = to_unary!(Reverse)
-)]
+#[sqlfunc(preserves_uniqueness = true, inverse = to_unary!(Reverse))]
 fn reverse<'a>(a: &'a str) -> String {
     a.chars().rev().collect()
 }
--- a/src/expr/src/scalar/func/impls/string.rs
+++ b/src/expr/src/scalar/func/impls/string.rs
@@ -158,7 +158,11 @@ fn cast_string_to_uint64(a: &str) -> Result<u64, EvalError> {
     strconv::parse_uint64(a).err_into()
 }
 
-#[sqlfunc(sqlname = "reverse")]
+#[sqlfunc(
+    sqlname = "reverse",
+    preserves_uniqueness = true,
+    inverse = to_unary!(Reverse)
+)]
 fn reverse<'a>(a: &'a str) -> String {
     a.chars().rev().collect()
 }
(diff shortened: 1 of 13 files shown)


src/expr/src/scalar/func/impls/string.rs
```diff
@@ -158,7 +158,11 @@ fn cast_string_to_uint64(a: &str) -> Result<u64, EvalError> {
     strconv::parse_uint64(a).err_into()
 }
 
-#[sqlfunc(sqlname = "reverse")]
+#[sqlfunc(
+    sqlname = "reverse",
```
antiguru:
```suggestion
```

Don't need to mention the name if it's equal to the function name.

antiguru (approved):
Looks good!

PR #36963, opened 2026-06-10
spec sheet: Move to separate buildkite pipeline

def-:
https://buildkite.com/materialize/spec-sheet


antiguru (approved):
Great, thank you!

PR #36969, opened 2026-06-10
[catalog] convert mz_clusters/mz_cluster_replicas/mz_cluster_schedules to BuiltinMaterializedView

mtabebe:
Problem:
mz_catalog.mz_clusters, mz_catalog.mz_cluster_replicas, and
mz_internal.mz_cluster_schedules are BuiltinTables.

This requires coordinated writes from a single environmentd process,
conflicting with the multi-envd goal.

Solution:
Convert all three to BuiltinMaterializedViews backed by queries over
mz_internal.mz_catalog_raw.

The disk column on mz_clusters and mz_cluster_replicas mirrors
CatalogState::cluster_replica_size_has_disk: it depends on the replica
size's swap_enabled and disk_limit allocation properties, which live in
process configuration and aren't durable in the catalog. To expose
swap_enabled to SQL we introduce a new internal BuiltinTable
mz_internal.mz_cluster_replica_size_internal(size, swap_enabled),
populated alongside mz_cluster_replica_sizes from the in-memory
ClusterReplicaSizeMap. The MVs LEFT JOIN against both tables, with
COALESCE handling missing sizes the same way the Rust function did.

For mz_cluster_schedules, the rust ClusterSchedule enum serializes
externally-tagged: Manual as the bare string Manual and Refresh as
{Refresh: {rehydration_time_estimate: {secs, nanos}}}. The MV
discriminates on these shapes and reconstructs the rehydration interval
via string cast (Materialize has no make_interval).

Testing:
New behavioural lockdown tests for mz_clusters, mz_cluster_replicas,
and mz_cluster_schedules


ggevay:
Some initial comments:
- (The PR description is from the previous PR, mentioning mz_roles.)
- Have to be rebased on `main` to pick up `catalog_server_explain.slt`, to avoid [another](https://materializeinc.slack.com/archives/C01LKF361MZ/p1781267287626409?thread_ts=1781258579.129039&cid=C01LKF361MZ) merge skew.

src/adapter/src/catalog/open/builtin_schema_migration.rs
```diff
@@ -197,6 +197,24 @@ static MIGRATIONS: LazyLock<Vec<MigrationStep>> = LazyLock::new(|| {
...
+            "mz_clusters",
+        ),
+        MigrationStep::replacement(
+            "26.29.0-dev.0",
+            CatalogItemType::MaterializedView,
+            MZ_CATALOG_SCHEMA,
+            "mz_cluster_replicas",
+        ),
+        MigrationStep::replacement(
+            "26.29.0-dev.0",
```
ggevay: (These three need to be adjusted when rebasing on `main`.)

src/catalog/src/builtin/mz_catalog.rs
```diff
@@ -2085,103 +2085,171 @@ pub static MZ_OPERATORS: LazyLock<BuiltinTable> = LazyLock::new(|| BuiltinTable
...
+            (
+                (data->'value'->'config'->'variant'->'Managed'->'logging'->'interval'->>'secs')
+                || ' seconds '
+                || ((data->'value'->'config'->'variant'->'Managed'->'logging'->'interval'->>'nanos')::bigint / 1000)::text
+                || ' microseconds'
+            )::interval
+    END AS introspection_interval
+FROM mz_internal.mz_catalog_raw
+LEFT JOIN mz_internal.mz_cluster_replica_size_internal internal
+    ON internal.size = data->'value'->'config'->'variant'->'Managed'->>'size'
```
ggevay: **Missing-size misconfiguration loses its signal for paused clusters — suggest a soft assert.** If a size is removed from the ClusterReplicaSizeMap while referenced, environments with a *running* replica of that size still panic at boot (`apply_cluster_replica_update` → `concretize_replica_location().expect(…)`, unchanged by this PR). But a managed cluster with **no** replicas (RF=0) previously panicked in `pack_cluster_update` and now boots silently with `disk = false`. Trading a whole-env crash-loop for a wrong display column is a good trade, but the signal shouldn't vanish entirely: suggest a `mz_ore::soft_panic_or_log` where the panic used to fire (e.g. when applying a managed `Cluster` update, if `config.size` is not in the size map) — loud in CI/staging, Sentry-visible but non-fatal in production.

ggevay: Btw. `workflow_test_sql_cluster_disk` could be extended with a restart that flips one in-use size to `disabled: true` (assert `disk` still resolves from the real allocation) and optionally removes a size entirely (assert `disk=false`, no crash). The disabled-row preservation is the most carefully-handled behavior in the PR and currently has zero executions; exposure is low (only `free` is disabled in prod, and nothing live should have actual size `free`), hence nice-to-have rather than blocking.

src/catalog/src/builtin/mz_catalog.rs
```diff
@@ -2257,74 +2325,116 @@ WHERE
...
+    mz_internal.parse_catalog_id(data->'value'->'owner_id') AS owner_id,
+    CASE
+        WHEN data->'value'->'config'->'location' ? 'Managed' THEN
+            NOT COALESCE(internal.swap_enabled, false)
+            AND COALESCE(internal.disk_bytes, 0) != 0
+    END AS disk
+FROM mz_internal.mz_catalog_raw
+LEFT JOIN mz_internal.mz_cluster_replica_size_internal internal
+    ON internal.size = data->'value'->'config'->'location'->'Managed'->>'size'
+WHERE data->>'kind' = 'ClusterReplica'",
```
ggevay: Both the LEFT JOIN here and in the other MV have the issue that the `WHERE` condition will get stuck above the join, because we unfortunately have this optimizer issue: https://github.com/MaterializeInc/database-issues/issues/8495
I suggest doing a manual predicate pushdown, i.e., adding this `WHERE` condition into the `FROM mz_internal.mz_catalog_raw`.

ggevay (approved):

PR #36888, opened 2026-06-03
adapter,sql: Add backtraces to `RecursionLimitError`

mgree:
### Motivation

Makes good on an old TODO. Improves diagnosis of recursion errors.

### Description

Adds a `std::backtrace::Backtrace` to `RecursionLimitError`. These are not `Clone`, which means some other small code changes as well.

Backtraces are _enormous_ (e.g., 10k frames). There is no stable way to limit the size of the backtrace.

I use `Backtrace::force_capture()`, which ignores environment variables. This feels right for blowing the stack (where paying the cost of the trace is not so bad). We could use `Backtrace::capture()` to instead rely on the environment variables.

### Verification

CI green. Any test of concrete backtraces will be _very_ brittle, so I'm reluctant to add things there. This is fundamentally debugging tooling for us.

PR #36971, opened 2026-06-10
adapter,sql: Abbreviate extremely long stack traces, storing full trace in TMPDIR

mgree:
### Motivation

https://github.com/MaterializeInc/materialize/pull/36888#issuecomment-4622933680

### Description

Pre-materializes backtraces, possibly excerpting the head and tail (25 frames/50 lines each). Full traces are saved to an appropriate TMPDIR.

### Verification

Manual verification using the given example. (It's possible to write a more complex test to, e.g., ensure that the trace exists and has the right number of lines, but it doesn't feel worth it.)

### DIFF
--- a/src/ore/src/stack.rs
+++ b/src/ore/src/stack.rs
@@ -17,12 +17,13 @@
 
 use std::backtrace::Backtrace;
 use std::cell::RefCell;
-use std::collections::VecDeque;
 use std::error::Error;
 use std::fmt;
 use std::io::{BufWriter, Write};
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicU64, Ordering};
+
+use tracing::error;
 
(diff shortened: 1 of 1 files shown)


def- (commented):
What if you keep storing /tmp files until environmentd runs out of disk? I'd just log the stack trace instead

mgree:
Sure! (Though I imagine that log is also being written to disk, in addition to being sent somewhere?)

Should I log full trace, or the abbreviated one?

def-:
I thought we want to have the full trace somewhere, not sure how DoSable it is.

mgree:
Okay, this emits the full trace attached to an event in the logs. Every trace gets a sequentially increasing id (per envd) that the error message can correlate back to.

def- (approved):

PR #36972, opened 2026-06-10
[SS-166] Widen `dimension : u8` to `i32` when writing Array values to Iceberg.

ublubu:
This bug and fix is essentially the same thing as:
https://linear.app/materializeinc/issue/SS-143/iceberg-failed-to-convert-row-to-recordbatch-failed-to-add-insert-row

https://github.com/MaterializeInc/materialize/pull/36500

Iceberg doesn't have unsigned ints or ints narrower than 32 bits. So if we need to write an Array value, the (normally u8) dimension must be stored as i32.


DAlperin (approved):
LGTM, thank you

PR #36975, opened 2026-06-11
ci: Fix nightly failures (2026-06-11)

def-:
@ublubu Is it correct that we really require the`rsa` dependency? CC @jasonhernandez 

The benchmark timeouts are required so that we show which scenarios are failing instead of just timing out the entire test when there is a big regression

Test run: https://buildkite.com/materialize/nightly/builds/16763 & https://buildkite.com/materialize/nightly/builds/16767

### DIFF
--- a/ci/nightly/pipeline.template.yml
+++ b/ci/nightly/pipeline.template.yml
@@ -165,7 +165,7 @@ steps:
       - id: feature-benchmark
         label: "Feature benchmark against merge base or 'latest'"
         depends_on: build-x86_64
-        timeout_in_minutes: 60
+        timeout_in_minutes: 120
         parallelism: 12
         agents:
           queue: hetzner-x86-64-dedi-16cpu-64gb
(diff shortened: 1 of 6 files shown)


jasonhernandez:
aws-lc-rs is our preferred cryptography library to align with eventual FIPS compliance

ublubu:
We need `reqsign` (which brings in `rsa`) for `iceberg-storage-opendal`'s GCS support.

from Claude:
> reqsign's google/signer.rs uses rsa::pkcs1v15::SigningKey to sign GCS requests with
  the service-account RSA private key. So removing it would require either dropping the opendal-gcs feature (losing GCS Iceberg support) or upstream reqsign
  switching to a different RSA implementation (e.g. ring/aws-lc-rs).

def-:
We could fork reqsign and change the dependency, but that's probably a larger change and I'm not sure we want to go down that route since it adds to our maintenance overhead.

bosconi (approved):
Looks good.

PR #36979, opened 2026-06-11
Remove sync MV sink implementation and consolidate to single version

antiguru:
### Motivation

The codebase had two implementations of the MV sink operator: an async version using `OperatorBuilder` and a sync version using `OperatorBuilderRc` with Tokio tasks. The sync version was gated behind the `ENABLE_SYNC_MV_SINK` dynamic config flag. This PR removes the experimental sync implementation and consolidates to a single, production-ready sync implementation.

### Description

This change:

1. **Removes the experimental sync MV sink module** (`src/compute/src/sink/materialized_view_v2.rs`) which contained an alternative implementation using sync Timely operators with Tokio task delegation.

2. **Consolidates the implementation** by moving the sync operator code from `materialized_view_v2.rs` into `materialized_view.rs`, making it the sole implementation. The sync approach (using `OperatorBuilderRc` with `mpsc` channels to Tokio tasks) is now the standard.

3. **Removes the feature flag** (`ENABLE_SYNC_MV_SINK`) from `src/compute-types/src/dyncfgs.rs` that previously gated between the two implementations.

4. **Removes system parameter** `enable_compute_sync_mv_sink` from the Python test infrastructure (`misc/python/materialize/mzcompose/__init__.py` and related test files).

5. **Updates module structure** by removing the `materialized_view_v2` module reference from `src/compute/src/sink.rs`.

6. **Preserves documentation** by moving the "Channel ordering requirements" section from the removed module into the main `materialized_view.rs` module docs, ensuring the critical ordering invariants for the sync implementation are documented.

The sync implementation uses a cleaner architecture where Timely operators communicate with Tokio tasks via `mpsc` channels, with careful ordering guarantees documented in the module-level comments. This replaces the previous async operator approach.

### Verification

The change is a consolidation of existing code paths. The sync implementation was already in production use when the flag was enabled. Removing the flag and the alternate implementation simplifies the codebase while keeping the proven sync approach. Existing tests that exercise the MV sink functionality will continue to validate this implementation.

https://claude.ai/code/session_012CGcAwA1hBcqefjw9ynizh


src/compute/src/sink/materialized_view.rs
```diff
@@ -18,6 +18,10 @@
 //!    up with undesired contents for any reason, this is corrected the next time the sink manages
 //!    to append to the shard.
 //!
+//! The sink's operators are sync Timely operators that delegate all persist I/O to Tokio tasks,
+//! communicating with them over `mpsc` channels. The "Channel ordering requirements" section below
+//! documents the invariants this split relies on.
```
bosconi: The implementation moved in here is the June snapshot of `materialized_view_v2.rs`, and main has changed that file four times since this branch was cut: #37233, #38118, #38358 and #38364. The last three are the ones that matter, because they move the correction maintenance and the batch read-back off the Tokio worker threads, which is the failure mode that led to this implementation being turned off. None of that is present in this copy: `spawn_blocking` and `block_in_place` appear in `origin/main:src/compute/src/sink/materialized_view_v2.rs` and nowhere in this file.

A rebase will surface this as a modify/delete conflict on `materialized_view_v2.rs`, and the resolution has to be to re-take the move from main's version rather than keep this copy, otherwise the merge silently reverts those fixes.

src/compute-types/src/dyncfgs.rs
```diff
@@ -123,13 +123,6 @@ pub const ENABLE_MZ_JOIN_CORE: Config<bool> = Config::new(
...
 );
 
-/// Use sync Timely operators with Tokio tasks for the MV sink.
-pub const ENABLE_SYNC_MV_SINK: Config<bool> = Config::new(
-    "enable_compute_sync_mv_sink",
-    true,
-    "Use sync Timely operators with Tokio tasks for the MV sink.",
-);
-
 /// Whether rendering should use the new MV sink correction buffer implementation.
```
bosconi: Since this branch was cut, #38074 flipped this flag's compiled-in default to `false`, as a follow-up to INC-1175 where the sync implementation was observed to schedule long-running blocking work on Tokio worker threads and starve the other tasks on them, losing reader leases. That commit also notes that cloud had already disabled the flag for every environment through LaunchDarkly, and says the default should only be flipped back after a gradual rollout.

So removing the flag together with the async implementation makes the currently-disabled path the only path and removes the switch incident response used. That is a decision the Cluster team should sign off on explicitly, and the answer depends on whether #38118, #38358 and #38364 are considered to have settled the blocking-work problem. Whatever the answer, it belongs in the description, because right now the stated justification is that the sync path is the one in production, which is the opposite of the current state.

No objection to the removal mechanics themselves: an `ALTER SYSTEM SET` of a name that no longer exists is tolerated on startup, `catalog/apply.rs` warns on the unknown parameter and moves on.

src/compute-types/src/dyncfgs.rs
```diff
@@ -487,7 +480,6 @@ pub fn all_dyncfgs(configs: ConfigSet) -> ConfigSet {
     configs
         .add(&ENABLE_HALF_JOIN2)
         .add(&ENABLE_MZ_JOIN_CORE)
-        .add(&ENABLE_SYNC_MV_SINK)
         .add(&ENABLE_CORRECTION_V2)
```
bosconi: `enable_compute_sync_mv_sink` exists as a LaunchDarkly flag, so dropping it from `all_dyncfgs` leaves a flag in LaunchDarkly with no parameter behind it. `test/launchdarkly-flag-consistency` tolerates that for exactly one release, because it treats a flag as stale only when it is absent from both the current build and the last published version. One release after this merges the check turns red.

Either delete the flag on the LaunchDarkly side as part of merging, which is the preferred direction since the allowlist is meant to shrink, or add `enable_compute_sync_mv_sink` to `KNOWN_STALE_LD_FLAGS` in `test/launchdarkly-flag-consistency/mzcompose.py` in this PR.

src/compute/src/sink/materialized_view.rs
```diff
@@ -807,134 +955,283 @@ mod write {
...
+        // `WriteCommand::Batch { persist_frontier, .. }`, which the Timely closure populates only
+        // when an input frontier actually moves. On startup, the input frontiers begin at
+        // `Timestamp::MIN`, so no `Batch` carries `persist_frontier` until the persist input
+        // catches up — yet a `WriteBatch(desc)` with `desc.lower = as_of` can already be sent.
+        // Snapshot-replay updates inserted in the meantime stay at their original timestamps
+        // (`Correction::insert` rounds to `max(t, since)` and `since == MIN`), and slip into the
+        // batch, tripping persist's `UpdateNotBeyondLower` invariant. Advancing
+        // `corrections.since` here keeps the Tokio side in lockstep with the Timely side, the
+        // same invariant `write::State::new` upholds via
+        // `apply_persist_frontier_advancement`.
```
bosconi: `apply_persist_frontier_advancement` was a method on the async implementation's `write::State`, which this PR deletes. After the move, `write::State` in this same file is the sync one, which upholds the invariant through the two `advance_persist_*_frontier` helpers called from `State::new`. As written the comment points at a name that no longer exists anywhere in the tree.

```suggestion
        // same invariant `write::State::new` upholds via
        // `advance_persist_ok_frontier` and `advance_persist_err_frontier`.
```

src/compute/src/sink/materialized_view.rs
```diff
@@ -232,73 +264,6 @@ pub(super) type BatchesStream<'s> = StreamVec<'s, Timestamp, (BatchDescription,
...
-    // Report sink frontier updates to the `ComputeState`.
-    let collection = compute_state.expect_collection_mut(sink_id);
-    collection.sink_write_frontier = Some(sink_frontier);
-
-    Rc::new((persist_token, mint_token, write_token, append_token))
-}
-
 /// Generic wrapper around ok/err pairs (e.g. streams, frontiers), to simplify code dealing with
 /// such pairs.
 pub(super) struct OkErr<O, E> {
```
bosconi: `pub(super)` on these items existed so that `materialized_view_v2` could reach them. With that module gone, nothing outside `materialized_view` refers to them: `sink.rs` declares `mod materialized_view;` privately and no `materialized_view::` path appears anywhere else in `src/`. The `mint`, `write` and `append` submodules reach these through `use super::*` and can see private parent items, so dropping the visibility compiles unchanged and stops the file advertising an interface it no longer has.

```suggestion
/// Type of the shared sink write frontier.
type SharedSinkFrontier = Rc<RefCell<Antichain<Timestamp>>>;

/// Generic wrapper around ok/err pairs (e.g. streams, frontiers), to simplify code dealing with
/// such pairs.
struct OkErr<O, E> {
```

The same applies to the rest of the moved items: `DesiredStreams`, `PersistStreams`, `DescsStream`, `BatchesStream`, the `OkErr` fields and methods, `advance`, `PersistApi` and its fields and methods, `persist_source`, `BatchDescription` and its fields and constructor, and `operator_name`.

misc/python/materialize/mzcompose/__init__.py
```diff
@@ -208,11 +208,6 @@ def get_variable_system_parameters(
             "true",
             ["true", "false"],
         ),
-        VariableSystemParameter(
-            "enable_compute_sync_mv_sink",
-            "true",
-            ["true", "false"],
-        ),
         VariableSystemParameter(
```
bosconi: Removing the entry is the right call once the flag is gone, but it is worth noting what it costs in coverage: this was the only knob that ran the MV sink both ways in CI, and it defaulted to `true`, so CI has been exercising the sync implementation while production has been on the async one since #38074. After this lands, every suite runs the consolidated sink and there is no longer a configuration in which the async path can be compared against it.

Given that, please run the rebased branch through a nightly, plus 0dt and platform-checks, and record which suites were run in the Verification section. The failure mode from INC-1175 was runtime scheduling behavior under load rather than anything a unit test would catch, so the argument for merging has to rest on those runs.

src/compute/src/sink/materialized_view.rs
```diff
@@ -113,22 +117,50 @@
...
 //! At the same time, the `write` operator still observes changes to the `desired` and `persist`
 //! collections, allowing it to keep its correction buffer up-to-date.
+//!
+//! ### Channel ordering requirements
+//!
+//! Each operator splits state across a Timely thread (which observes inputs and frontiers) and a
+//! Tokio task (which owns persist I/O state). They communicate via `mpsc` command channels, which
+//! preserve send order on a single sender. Each operator instance constructs its own
+//! `(tx, rx)` pair inside `render` and never clones the sender, so there is exactly one producer
+//! per channel — sends are totally ordered. Different worker instances of the same operator
```
bosconi: nit: em-dash in a comment, which the project style asks us to restructure with commas or full stops. Same on lines 570 and 1014 of this file. These are moved lines rather than new prose, so this is only worth doing while the text is already being touched.

```suggestion
//! per channel, so sends are totally ordered. Different worker instances of the same operator
```

PR #36980, opened 2026-06-11
SQL-344: Create info level metrics

SangJunBak:
We sync the catalog to a few info metrics such that users can join IDs in our metric labels with their names as well as other info (like the size of a clustesr). We do this by creating a background task on serve which, every 30 seconds (this is arbitrary), we check if the catalog has changed and if so, reconcile the info metrics. 

There does exist a case where a scrape can happen after DDL but before the reconciliation, leading to wrong names in grafana. However it's okay to be eventually consistent here given all joins in promql are effectively left joins (so if there doesn't exist an info metric for an object, it can handle that gracefully). Also given info metrics can be scraped at different intervals in a scrape config, we can assume the default behavior is eventual consistent.

Some other implementations I explored were:
- Reconciling using the catalog implications framework
     - I didn't like how we had to still snapshot on bootstrap and didn't feel like it belonged with the same code that creates cluster replicas?
- Having reconciliation trigger on group_commit (similar to our builtin tables)
     - It required passing around a top level channel which I wasn't a fan of

### Motivation

Fixes https://linear.app/materializeinc/issue/SQL-344/create-info-level-metrics 

Initially we were going to mutate the labels via a rules header in the stable endpoint https://github.com/MaterializeInc/materialize/pull/36633. However given scalability concerns, we plan on having self managed customers consume metrics via the internal endpoint. I've since [created a revert PR for that feature](https://github.com/MaterializeInc/materialize/compare/main...SangJunBak:materialize:revert-c8a90af6a2cc52b2816e60d0ecef86c89f0ca945?expand=1). However, they still need a way to join names to IDs. The common practice in Prometheus is to expose an "info" metric such that a user can join on it to get what they need. 

### Verification

Created tests and also did manual testing of a workload replay of a customer's environment.

### DIFF
--- a/src/adapter/src/coord/info_metrics.rs
+++ b/src/adapter/src/coord/info_metrics.rs
@@ -23,6 +23,7 @@
 
 use std::time::{Duration, Instant};
 
+use mz_adapter_types::dyncfgs::CATALOG_INFO_METRICS_RECONCILE_INTERVAL;
 use mz_catalog::memory::objects::{
     CatalogItem, Cluster, ClusterReplica, ClusterVariant, DataSourceDesc,
 };
@@ -36,15 +37,16 @@ use mz_repr::CatalogItemId;
 use mz_sql::names::{FullItemName, RawDatabaseSpecifier};
 use prometheus::core::AtomicU64;
 use tokio::sync::oneshot;
-use tokio::time::MissedTickBehavior;
 use tracing::{debug, warn};
 
 use crate::catalog::{Catalog, catalog_type_to_audit_object_type};
 use crate::command::{CatalogSnapshot, Command};
 use crate::coord::{Coordinator, Message};
 
-/// How often the info metrics are reconciled with the catalog.
-pub(crate) const RECONCILE_INTERVAL: Duration = Duration::from_secs(30);
+/// Fallback reconcile cadence: the re-poll cadence used while reconciliation is
+/// disabled (a zero [CATALOG_INFO_METRICS_RECONCILE_INTERVAL]), so it can be
+/// re-enabled at runtime.
+const RECONCILE_INTERVAL: Duration = Duration::from_secs(30);
 
 /// Reconciles that take longer than this are logged at warn level.
 const RECONCILE_WARN_THRESHOLD: Duration = Duration::from_secs(5);
@@ -259,10 +261,7 @@ impl Coordinator {
         let internal_cmd_tx = self.internal_cmd_tx.clone();
         let mut metrics = CatalogInfoMetrics::new(&self.catalog_info_metrics_registry);
         task::spawn(|| "catalog_info_metrics", async move {
-            let mut interval = tokio::time::interval(RECONCILE_INTERVAL);
-            interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
             loop {
-                interval.tick().await;
                 let (tx, rx) = oneshot::channel();
                 let send = internal_cmd_tx.send(Message::Command(
                     OpenTelemetryContext::obtain(),
@@ -275,8 +274,23 @@ impl Coordinator {
                 let Ok(CatalogSnapshot { catalog }) = rx.await else {
                     break;
                 };
-                // Reconcile the metrics with the catalog.
-                metrics.reconcile(&catalog);
+
+                // The reconcile cadence is a dyncfg; a zero interval disables
+                // reconciliation.
+                let interval =
+                    CATALOG_INFO_METRICS_RECONCILE_INTERVAL.get(catalog.system_config().dyncfgs());
+                if !interval.is_zero() {
+                    metrics.reconcile(&catalog);
+                }
+
+                // When disabled, keep polling at the fallback cadence so it can
+                // be re-enabled at runtime.
+                let sleep = if interval.is_zero() {
+                    RECONCILE_INTERVAL
+                } else {
+                    interval
+                };
+                tokio::time::sleep(sleep).await;
             }
         });
     }
--- a/src/adapter/src/coord/info_metrics.rs
+++ b/src/adapter/src/coord/info_metrics.rs
@@ -63,7 +63,8 @@ pub(crate) struct CatalogInfoMetrics {
     object_info: UIntGaugeVec,
     cluster_info: UIntGaugeVec,
     replica_info: UIntGaugeVec,
-    storage_object_info: UIntGaugeVec,
+    source_info: UIntGaugeVec,
+    sink_info: UIntGaugeVec,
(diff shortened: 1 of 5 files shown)


def- (commented):
The QA LLM review had feedback which seems pretty reasonable to me, especially since we've struggled with the coordinator thread before and tried to move things away from it, not into it:

### MEDIUM — Catalog-proportional synchronous series rebuild on the coordinator main loop, retriggered by *any* catalog transaction

**Where:** `src/adapter/src/coord/info_metrics.rs` (`reconcile`/`populate`),
select arm in `src/adapter/src/coord.rs` (`catalog_info_metrics_interval.tick()`).

**Mechanism:**

1. The reconcile guard keys off `Catalog::transient_revision`, which is bumped
   by **every** `Catalog::transact` with non-empty ops
   (`src/adapter/src/catalog/transact.rs:487`; `transact_inner` returns
   `Cow::Owned` for any applied update, including temporary-item-only ones).
   That includes operations that cannot change any reported series: temporary
   tables (explicitly excluded from the metrics), `COMMENT`, `GRANT`/`REVOKE`,
   role DDL, `ALTER SYSTEM`, etc. In any environment with steady DDL-ish
   activity (dbt runs, apps using temp tables), the guard is effectively
   always stale, so the full rebuild runs on every 30s tick.

2. `populate` is not incremental: it drops **all** existing series and
   recreates one series per catalog item × global ID (plus clusters, replicas,
   storage objects), iterating `catalog.entries()` — which includes the ~2k+
   system catalog items and per-cluster introspection source indexes (those
   are filtered, but only after a `resolve_full_name` call each). Each series
   costs ~6 `String` allocations, a label hash, and an `RwLock`-guarded map
   insert in the prometheus vec — and the same again on the drop side for the
   old generation.

3. This all runs synchronously on the coordinator main loop (the new select
   arm calls `self.catalog_info_metrics.reconcile(&self.catalog)` inline and
   `continue`s). Everything in the system — peeks, commits, group commit,
   controller messages — serializes behind it. For small catalogs this is
   single-digit milliseconds and fine; for large catalogs (tens of thousands
   of objects is real customer scale; limits tests go to 100k) it is a
   recurring stall plausibly in the hundreds of milliseconds, every 30
   seconds, manifesting as periodic tail-latency spikes on all sessions.

4. The stall is also **unattributable with existing tooling**: because the arm
   `continue`s before the message-processing instrumentation, the time is
   invisible to the `message_handling` histogram, the
   `coord_slow_message_warn_threshold` warning, and the `LastMessage`
   stuck-coordinator watchdog. When it does cause latency spikes in
   production, nothing will point at this code.

**Why MEDIUM:** no correctness impact and the frequency is bounded (1 rebuild
per 30s), but it is a new O(catalog) recurring block on the most
latency-sensitive single-threaded loop in the system, triggered by routine
activity, and invisible to the existing slow-coordinator diagnostics.

**Suggested directions (any of):**

- Move the rebuild off the main loop: `Catalog` is cheaply cloneable
  (CoW `CatalogState`, `impl Clone for Catalog` at
  `src/adapter/src/catalog.rs:146`) — clone the catalog on the tick and do the
  rebuild in a `spawn_blocking`/task, keeping only the revision check on the
  main loop.
- Make the rebuild incremental (diff old/new state, or consume the parsed
  catalog updates / catalog-implications stream) so unchanged objects keep
  their series.
- At minimum, time the reconcile and feed it into the `message_handling`
  metrics / slow-message warning so stalls are attributable.

aljoscha (changes requested):
I agree with "Dennis", we should change that to do work off the coord loop. Also, we might want to make the update interval configurable.

src/adapter/src/coord/info_metrics.rs
```diff
@@ -0,0 +1,604 @@
...
+                help: "Maps cluster IDs to the cluster's name and size. Constant 1.",
+                var_labels: ["cluster_id", "name", "size"],
+            )),
+            replica_info: registry.register(metric!(
+                name: "mz_replica_info",
+                help: "Maps cluster replica IDs to the replica's name and size. Constant 1.",
+                var_labels: ["replica_id", "cluster_id", "name", "size"],
+            )),
+            storage_object_info: registry.register(metric!(
+                name: "mz_storage_object_info",
```
Alphadelta14: wonder if we should split into separate source and sink metrics. I was also initially confused as how we just call them storage objects but those are internal-only names.
Started slack discussion ->

src/adapter/src/coord/info_metrics.rs
```diff
@@ -0,0 +1,604 @@
...
+    /// from. Used to avoid rebuilding the metrics if the catalog has not changed.
+    last_revision: Option<u64>,
+}
+
+impl CatalogInfoMetrics {
+    pub fn new(registry: &MetricsRegistry) -> Self {
+        Self {
+            object_info: registry.register(metric!(
+                name: "mz_object_info",
+                help: "Maps catalog object IDs to the object's name, schema, database, and \
```
Alphadelta14, line 84: do things like indexes and mzviews show up here?

SangJunBak: Yes!

src/adapter/src/coord/info_metrics.rs
```diff
@@ -0,0 +1,675 @@
...
+use tracing::{debug, warn};
+
+use crate::catalog::{Catalog, catalog_type_to_audit_object_type};
+use crate::command::{CatalogSnapshot, Command};
+use crate::coord::{Coordinator, Message};
+
+/// Fallback reconcile cadence: the re-poll cadence used while reconciliation is
+/// disabled (a zero [CATALOG_INFO_METRICS_RECONCILE_INTERVAL]), so it can be
+/// re-enabled at runtime.
+const RECONCILE_INTERVAL: Duration = Duration::from_secs(30);
```
aljoscha:
```suggestion
const FALLBACK_RECONCILE_INTERVAL: Duration = Duration::from_secs(30);
```

maybe?

aljoscha (approved):
Looks good from the technical Coordinator/task side. I'll defer to @Alphadelta14 and you on how you want to split up and name the metrics.

src/adapter/src/coord/info_metrics.rs
SangJunBak: Sounds good to me!

src/adapter/src/coord/info_metrics.rs
SangJunBak: Addressed here https://github.com/MaterializeInc/materialize/pull/36980/changes/71b309b6660fc650b7da460815ee2d35c73fccb4

PR #36982, opened 2026-06-11
tests: Add the cargo-fuzz fuzzing suite

def-:
Adds the cargo-fuzz suite: fuzz targets, seed corpora, dictionaries, and the runner/CI integration (release-qualification, 24h). Covers the SQL parser/pretty-printer, repr (strconv, jsonb, Row codec/proto, arithmetic oracles), the expr optimizer transforms, the Avro/Protobuf/CSV/pgwire/pgcopy decoders, pgrepr/pgtz, the upsert state machine, persist durable-state decode, and proto round-trips across storage-types/persist/catalog/external table descs.

Also includes the production-side enablement the targets require (the `fuzzing`/`fuzz` Cargo features and `#[doc(hidden)]`/`cfg`-gated re-exports) and the macOS build fixes those exposures necessitated.

This is the infrastructure PR, mostly mechanical (generated corpora/dicts). The individual bugs it surfaced are split into separate per-subsystem PRs.

Depends on these fixes for bugs it found:
- [x] https://github.com/MaterializeInc/materialize/pull/36983
- [x] https://github.com/MaterializeInc/materialize/pull/36984
- [x] https://github.com/MaterializeInc/materialize/pull/36985
- [x] https://github.com/MaterializeInc/materialize/pull/36986
- [x] https://github.com/MaterializeInc/materialize/pull/37047
- [x] https://github.com/MaterializeInc/materialize/pull/37048
- [x] https://github.com/MaterializeInc/materialize/pull/37049
- [x] https://github.com/MaterializeInc/materialize/pull/37096
- [x] https://github.com/MaterializeInc/materialize/pull/37128
- [x] https://github.com/MaterializeInc/materialize/pull/37130
- [x] https://github.com/MaterializeInc/materialize/pull/37145
- [x] https://github.com/MaterializeInc/materialize/pull/37177
- [x] https://github.com/MaterializeInc/materialize/pull/37299

Green test run: https://buildkite.com/materialize/release-qualification/builds/1294

PR #36983, opened 2026-06-11
sql: Fix pretty/display round-trip bugs found by fuzzing

def-:
Print→reparse round-trip bugs in the SQL parser and pretty-printer, surfaced by the grammar-aware fuzz target. Each fix has a regression test; the sqllogictest/testdrive plan goldens are refreshed to match.

Themes: quoting bare keyword identifiers (any/all/some/list, context-sensitive keywords), parenthesizing low-precedence operands (prefix ops, casts, COLLATE, quantified comparisons), special-form display correctness (EXTRACT/POSITION/SUBSCRIBE), and bounding parser recursion/backtracking to reject pathological inputs.

Found by the cargo-fuzz suite ([separate infra PR](https://github.com/MaterializeInc/materialize/pull/36982)).


src/sql-parser/tests/testdata/copy
```diff
@@ -37,7 +37,7 @@ Copy(CopyStatement { relation: Select(SelectStatement { query: Query { ctes: Sim
 parse-statement
 COPY (subscribe (SELECT 1)) TO STDOUT
 ----
-COPY (SUBSCRIBE (SELECT 1)) TO STDOUT
+COPY (SUBSCRIBE TO (SELECT 1)) TO STDOUT
```
aljoscha: These SUBSCRIBE TO changes feel a bit weird, why do we have them again?

aljoscha (commented):
Changes look good, but I haven't looked at the tests and largely only eyeballed the parse changes. On these I would pretty much trust our testing infra. Had one question about SUBSCRIBE TO which we added in a number of places that we should resolve

src/sql-parser/tests/testdata/copy
def-: `SUBSCRIBE TO TO` with the second `to` being an identifier was causing problems. I have now fixed it so that we only add the `TO` in the display when there is a `to` identifier following it.

def-:
Thanks for reviewing @aljoscha , and sorry for the huge change! I didn't want to open 50 PRs for these minor changes

src/sql-parser/src/ast/metadata.rs
```diff
@@ -245,7 +257,51 @@ impl AstDisplay for RawDataType {
...
+                                | FLOAT
+                                | INT
+                                | INTEGER
+                                | REAL
+                                | BOOLEAN
+                                | BYTES
+                                | JSON
+                                | CHAR
+                                | CHARACTER
+                        )
```
antiguru: Could we extract this to a function, or is it really specific to this code block?

test/sqllogictest/advent-of-code/2023/aoc_1207.slt
```diff
@@ -324,7 +324,7 @@ Explained Query:
                 Get l6 // { arity: 2 }
             Get l5 // { arity: 1 }
     cte l8 =
-      Project (#0, #1, #3{any}) // { arity: 3 }
+      Project (#0, #1, #3{"any"}) // { arity: 3 }
```
antiguru: Do we need quoting here? This is always a column reference, so it shouldn't be ambiguous, but I might miss something.

antiguru (approved):
Seems fine, left some inline comments, but I lack some context here. Happy to hit approve, but only merge if you're sure it moves us in the right direction!

src/sql-parser/src/ast/metadata.rs
def-: Done

test/sqllogictest/advent-of-code/2023/aoc_1207.slt
def-: A bit more complex, but trying now

test/sqllogictest/quote_ident.slt
antiguru: Is this comment still true?

test/sqllogictest/quote_ident.slt
```diff
@@ -342,7 +342,7 @@ materialize
 query T
 SELECT quote_ident('table');
 ----
-table
+"table"
```
antiguru, line 347: Seems `table` is now a reserved keyword?

antiguru (approved):

test/sqllogictest/quote_ident.slt
def-: Nope, updated!

test/sqllogictest/quote_ident.slt
def-: It's somewhere inbetween.

test/sqllogictest/pretty.slt
antiguru: Likely stale comment, too.

PR #36984, opened 2026-06-11
repr: Harden proto/Row decoding against malformed input

def-:
Hardens `mz-repr` proto/Row decoding against malformed/untrusted bytes — these paths are reachable from persisted state and the wire, so a panic is an availability bug. Replaces panics/asserts with proper decode errors and validates ranges (Date, CheckedTimestamp, ProtoNumeric, ProtoRange, ProtoRelationDesc, ProtoRow dict ordering, uuid parsing, leap-second truncation, Avro decimal).

### DIFF
--- a/src/repr/src/row/encode.rs
+++ b/src/repr/src/row/encode.rs
@@ -2126,14 +2126,12 @@ impl RowPacker<'_> {
                     // Map keys must be unique and strictly ascending; iterating a
                     // map that violates this trips a debug_assert. A crafted
                     // proto can, so reject it as a decode error here instead.
-                    if let Some(prev) = prev_key {
-                        if e.key.as_str() <= prev {
-                            return Err(format!(
-                                "dict keys must be unique and in ascending order, \
-                                 but {:?} came after {:?}",
-                                e.key, prev,
-                            ));
-                        }
+                    if let Some(prev) = prev_key && e.key.as_str() <= prev {
+                        return Err(format!(
+                            "dict keys must be unique and in ascending order, \
+                             but {:?} came after {:?}",
+                            e.key, prev,
+                        ));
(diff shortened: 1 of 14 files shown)


def-:
Ready for review

src/repr/src/row/encode.rs
```diff
@@ -2121,7 +2121,21 @@ impl RowPacker<'_> {
...
+                    // proto can, so reject it as a decode error here instead.
+                    if let Some(prev) = prev_key {
+                        if e.key.as_str() <= prev {
+                            return Err(format!(
+                                "dict keys must be unique and in ascending order, \
+                                 but {:?} came after {:?}",
+                                e.key, prev,
+                            ));
+                        }
+                    }
```
antiguru:
```suggestion
                    if let Some(prev) = prev_key && e.key.as_str() <= prev {
                        return Err(format!(
                            "dict keys must be unique and in ascending order, \
                             but {:?} came after {:?}",
                            e.key, prev,
                        ));
                    }
```

src/repr/src/row/encode.rs
```diff
@@ -2134,6 +2148,14 @@ impl RowPacker<'_> {
             Some(DatumType::Numeric(x)) => {
                 // Reminder that special values like NaN, PosInf, and NegInf are
                 // represented as variants of ProtoDatumOther.
+                //
+                // `decPackedToNumber` (called via `Decimal::from_packed_bcd`)
+                // doesn't bounds-check its input and segfaults on empty bcd —
```
antiguru: Here and elsewhere: Try to avoid em-dashes and write full sentences.

antiguru (approved):
Looks good, thank you!

PR #36985, opened 2026-06-11
persist: Harden durable-state and binary decoding against malformed input

def-:
Hardens decode paths in `persist`, `persist-client`, `persist-types`, `pgrepr`, and `postgres-util` against malformed/untrusted input: reject rollup-less/empty-frontier/hollow-only/invalid StateDiff state instead of panicking or debug-asserting, guard UUID parsing, fix i16 overflow in pgrepr numeric-scale binary decode, and propagate u16-conversion errors in the postgres table-desc protos.

### DIFF
--- a/src/persist-client/src/internal/encoding.rs
+++ b/src/persist-client/src/internal/encoding.rs
@@ -2412,7 +2412,37 @@ mod tests {
             });
         });
         let err = rollup_decode_err(proto);
-        assert!(err.contains("do not cover"), "{err}");
+        assert!(err.contains("do not tile"), "{err}");
+    }
+
+    /// A spine batch whose parts hit the right endpoints but overlap in the
+    /// middle (id `[0, 3)` tiled by parts `[0, 2)` and `[1, 3)`) passes the
+    /// endpoint check but is not a valid tiling. Such parts reach compaction's
+    /// `id_range`, whose contiguity `assert_eq!` would panic later, so decoding
+    /// must reject them here instead.
+    #[mz_ore::test]
+    fn rollup_proto_with_noncontiguous_spine_parts_is_rejected() {
+        let outer = SpineId(0, 3);
+        let part_ids = [SpineId(0, 2), SpineId(1, 3)];
+        let proto = rollup_proto_with_trace(|trace| {
+            trace.spine_batches.push(ProtoIdSpineBatch {
+                id: Some(outer.into_proto()),
+                batch: Some(ProtoSpineBatch {
+                    level: 0,
+                    desc: Some(u64_desc_proto(0, 3, 0)),
+                    parts: part_ids.iter().map(|id| id.into_proto()).collect(),
+                    descs: vec![],
+                }),
+            });
+            for id in part_ids {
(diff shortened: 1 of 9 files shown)


def-:
Ready for review
Edit: Last blocker for https://github.com/MaterializeInc/materialize/pull/36982

src/persist-client/src/internal/encoding.rs
```diff
@@ -295,6 +295,13 @@ pub(crate) fn parse_id(id_prefix: &str, id_type: &str, encoded: &str) -> Result<
         Some(x) => x,
         None => return Err(format!("invalid {} {}: incorrect prefix", id_type, encoded)),
     };
+    // `Uuid::parse_str` panics (rather than erroring) while building its parse
+    // error for some malformed inputs. For example, non-ASCII bytes make it
+    // slice the input on a non-char boundary. Reject anything that can't be a
+    // UUID first: the shortest valid form is 32 hex digits, all ASCII.
+    if uuid_encoded.len() < 32 || !uuid_encoded.is_ascii() {
+        return Err(format!("invalid {} {}: malformed UUID", id_type, encoded));
+    }
```
petrosagg: I can't find any evidence that this method panics. I tried to parse the strings in the regression test below and they also produce proper errors without panicking. Which input can cause a panic?

src/persist-client/src/internal/encoding.rs
```diff
@@ -512,7 +519,13 @@ impl<T: Timestamp + Codec64> RustType<ProtoStateDiff> for StateDiff<T> {
             proto.latest_rollup_key.into_rust()?,
         );
         if let Some(field_diffs) = proto.field_diffs {
-            debug_assert_eq!(field_diffs.validate(), Ok(()));
+            // `field_diffs` is decoded from an untrusted blob, so validate it and
+            // return a decode error on a malformed/crafted diff rather than a
+            // `debug_assert` (which panics under debug assertions / fuzzing and
+            // is compiled out in release, where the bad data flowed through).
```
petrosagg: Classic Claude verbosity comment

> return a decode error on a malformed/crafted diff

obvious from the code

> rather than a `debug_assert`

after merging this PR there won't be any `debug_assert` around so mentioning it is strange

> (which panics under debug assertions / fuzzing and is compiled out in release, where the bad data flowed through)

No need to document `debug_assert` here

"`field_diffs` is decoded from an untrusted blob, so validate it" should be more than enough

petrosagg (commented):
Looks good overall. The comments could be more laconic and still be to the point

src/persist-client/src/internal/encoding.rs
def-: Thanks, it was already fixed by another commit in this PR, removed!

src/persist-client/src/internal/encoding.rs
def-: Thanks, made them less verbose

DAlperin (approved):

DAlperin (approved):
Looks good to me! Thanks for the fixes!

PR #36986, opened 2026-06-11
avro: Bound and validate decoding against malformed input

def-:
Hardens the `mz-avro` decoder against adversarial input (Avro bytes/schemas arrive from Kafka and an external registry, so a panic/OOM is an availability bug): bound per-block array/map lengths and object counts by remaining input, cap object-container block byte length, bound schema-parse/value-decode recursion, and fix two schema-resolution panics on unmatched named types.

### DIFF
--- a/src/avro/src/decode.rs
+++ b/src/avro/src/decode.rs
@@ -1347,8 +1347,40 @@ pub struct GeneralDeserializer<'a> {
     pub schema: SchemaNode<'a>,
 }
 
+/// Cap on recursive `GeneralDeserializer::deserialize` calls. Avro records
+/// may reference themselves (`{"name":"X","type":"record","fields":[
+/// {"name":"x","type":"X"}]}`), so a malicious file plus matching wire
+/// bytes can recurse forever and overflow the stack.
+const MAX_DECODE_DEPTH: usize = 128;
+
+thread_local! {
+    static DECODE_DEPTH: std::cell::Cell<usize> = const { std::cell::Cell::new(0) };
+}
+
+struct DecodeDepthGuard;
+impl DecodeDepthGuard {
+    fn enter() -> Result<Self, AvroError> {
+        DECODE_DEPTH.with(|d| {
+            let new = d.get() + 1;
+            if new > MAX_DECODE_DEPTH {
+                return Err(AvroError::Decode(DecodeError::Custom(format!(
+                    "Avro decode depth exceeds limit {MAX_DECODE_DEPTH}"
+                ))));
+            }
+            d.set(new);
+            Ok(DecodeDepthGuard)
+        })
+    }
+}
+impl Drop for DecodeDepthGuard {
+    fn drop(&mut self) {
+        DECODE_DEPTH.with(|d| d.set(d.get().saturating_sub(1)));
+    }
+}
+
 impl<'a> AvroDeserializer for GeneralDeserializer<'a> {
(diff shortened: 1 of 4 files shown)


src/avro/src/decode.rs
```diff
@@ -576,6 +650,21 @@ impl<'a, R: AvroRead> AvroArrayAccess for SimpleArrayAccess<'a, R> {
...
+            // A block can't hold more items than there are bytes left to decode
+            // them from; reject a count that claims otherwise rather than letting
+            // it drive an unbounded allocation (see `Skip::remaining_input`).
+            if let Some(remaining) = self.r.remaining_input() {
+                if len > remaining {
+                    return Err(AvroError::Decode(DecodeError::Custom(format!(
+                        "Avro array block length {len} exceeds remaining input ({remaining} bytes)"
+                    ))));
+                }
+            }
```
martykulma, line 906: subtle issue here for zero-byte elements

```rust
    #[mz_ore::test]
    fn valid_null_array_falsely_rejected() {
        // `remaining_input` assumes >=1 byte per element, false for null
        // (and other zero-byte) elements, so after the count VARINT, only
        // the 1-byte terminator remains. count >= 2 trips the check.
        use std::str::FromStr;

        use super::{AvroDeserializer, GeneralDeserializer};
        use crate::encode::encode_to_vec;
        use crate::types::Value;
        use crate::{Schema, ValueDecoder};

        let schema = Schema::from_str(r#"{"type": "array", "items": "null"}"#).unwrap();
        let value = Value::Array(vec![Value::Null, Value::Null]);
        let body = encode_to_vec(&value, &schema);

        let dsr = GeneralDeserializer {
            schema: schema.top_node(),
        };
        let mut reader: &[u8] = &body;
        let res = dsr.deserialize(&mut reader, ValueDecoder);
        assert!(
            res.is_ok(),
            "an encoder-produced array of nulls should round-trip, but got: {res:?}"
        );
    }

```

src/avro/src/decode.rs
```diff
@@ -560,6 +628,12 @@ impl<'a> AvroArrayAccess for ValueArrayAccess<'a> {
     }
 }
 
+/// Sanity cap on the per-block element count an Avro array/map can claim
+/// from the wire. Without it, a malicious or corrupt file can claim up
+/// to `i64::MAX` items and the generic array/map decode loop will run
+/// until it eventually OOMs or hits `Vec` capacity-overflow.
+const MAX_BLOCK_LEN: usize = 1 << 24;
```
martykulma: nit: would you be opposed to calling this `MAX_BLOCK_ELEMENTS` - as I was reading this, I kept thinking this was bytes not element count.

martykulma (commented):
Nice! Since we're introducing bounds that don't exist in the AVRO spec, this PR should also update the docs so customers are aware. 

There's a subtle issue with remaining_input, see inline for the test case. Could we just drop the remaining input check and rely on the `maximum element count * size_of::<Value>`?  We might want to drop the max element count in that case.

src/avro/src/decode.rs
def-: done

src/avro/src/decode.rs
def-: Thanks, fixing

def-:
> Could we just drop the remaining input check and rely on the maximum element count * size_of::<Value>? We might want to drop the max element count in that case.

It only measures the enum header, not heap allocations required, also not packed row bytes.

martykulma (approved):
lgtm - thanks @def- !

PR #36987, opened 2026-06-11
expr, storage: Fix reduce error-propagation, LIKE, and CSV decoder bugs

def-:
Three correctness/robustness fixes found by fuzzing:

- **expr:** don't propagate an operand's error through non-strict AND/OR in `reduce` (`false AND <error>` evaluates to `false`, so folding it to the error introduced a spurious error into a query that succeeds).
- **expr:** avoid super-linear LIKE matching on multi-`%` patterns.
- **storage:** fix CSV source-decoder panics on malformed untrusted input (cursor reset + per-field UTF-8 validation).

Each fix has a regression test. Found by https://github.com/MaterializeInc/materialize/pull/36982


src/expr/src/scalar/like_pattern.rs
```diff
@@ -149,7 +149,18 @@ pub fn compile(pattern: &str, case_insensitive: bool) -> Result<Matcher, EvalErr
...
+    // single `%` is near-linear, but with two or more the backtracking nests and
+    // the cost becomes super-linear in the text length — an adversarial pattern
+    // like `%a%a%a` against a long run of `a`s takes time proportional to
+    // `len(text)^(number of %)`, which can stall a worker for many minutes. The
+    // regex engine matches the same patterns in linear time with no backtracking,
+    // so fall back to it whenever more than one `many` subpattern is present (and
+    // for the existing case-insensitive / too-many-subpatterns reasons).
+    let many_subpatterns = subpatterns.iter().filter(|s| s.many).count();
+    let use_regex = case_insensitive || subpatterns.len() > MAX_SUBPATTERNS || many_subpatterns > 1;
+    let matcher_impl = match use_regex {
```
antiguru, line 163: What's even the point of having our own string matcher? I feel we could switch to the regex-rewriting for most (all?) cases.

src/expr/src/scalar/reduce/variadic.rs
```diff
@@ -51,9 +51,17 @@ pub(super) fn reduce_call_variadic(
...
-    if let Some(err) = exprs.iter().find_map(|x| x.as_literal_err()) {
-        *e = MirScalarExpr::literal(Err(err.clone()), e.typ(column_types).scalar_type);
-        return;
+    // Only a strict (null-propagating) function propagates an operand's error
+    // unconditionally. A non-strict function such as AND/OR has a dominating
+    // operand (`false`/`true`) that absorbs another operand's error at runtime
+    // — e.g. `false AND <error>` evaluates to `false` — so folding the whole
+    // call to the error here would introduce an error the evaluated expression
+    // never raises. (Coalesce, also non-strict, bailed out above.)
+    if func.propagates_nulls() {
```
antiguru, line 60: Needs optimizer review.

antiguru (commented):
I only now realized this is still draft, so take my reviews as hints. I couldn't review the CSV decoder change as I can't see the actual diff.

src/expr/src/scalar/like_pattern.rs
def-: It'd be up to 700x slower in extreme cases based on trying it out.

src/expr/src/scalar/reduce/variadic.rs
def-: @ggevay Are you familiar with this code?

antiguru (commented):
The regex change LGTM. I find it difficult to review the CSV decoder change. You could split this into three PRs?

src/expr/src/scalar/reduce/variadic.rs
ggevay: Yes, I'll review it

PR #36992, opened 2026-06-11
row-spine: carry first-byte bitmap through safe codec install

antiguru:
### Motivation

Fixes a dictionary-codec soundness bug surfaced by [arrangement-dictionary-compression.td](<http://arrangement-dictionary-compression.td/>): with enable_arrangement_dictionary_compression_alpha *and* enable_column_paged_batcher on, SELECT v FROM t WHERE k = 17 returned a long dictionary value (xxx…135) instead of the stored "1". Disabling either flag returned correct results.

### Root cause

A from-scratch DatumContainer stores its pre-install rows raw while gathering statistics, then installs a safe codec via new_safe once it crosses STATS_THRESHOLD. new_safe discarded the first-byte bitmap (stats.1) gathered over those raw rows.

That bitmap is soundness-critical: new_from (the merge path) hands out any unused one-byte tag as a dictionary key and decides "unused" purely from the OR of its inputs' bitmaps. A safe codec is sound on its own — it only uses tags >= SAFE_TAG_BASE, which no datum first-byte can equal — so the dropped bitmap goes unnoticed until two such containers merge. With the pre-install bitmap gone, the installed codec's bitmap covered only post-install rows, so the merge could assign a low dictionary tag equal to a pre-install datum's first byte (StringTiny = 19 here). decode then resolves that literal datum to the dictionary entry and returns the wrong value.

### Fix

Carry the observed-tag bitmap into the installed codec so it covers both pre- and post-install rows. The MisraGries summary is still reset, since the installed codec re-accumulates it from rows it sees post-install. The symmetric reset in new_from is correct as-is: a new_from container re-encodes every row through encode, so observe sees them all — there are no bypassed raw rows there.

### Tests

Adds test_safe_codec_merge_bitmap_carryover, which drives the install-then-merge lifecycle directly. It trips encode's soundness debug_assert (raw datum first-byte 19 collides with a dictionary tag) before the fix and passes after.

Closes CLU-110

### DIFF
--- a/src/row-spine/src/lib.rs
+++ b/src/row-spine/src/lib.rs
@@ -279,6 +279,76 @@ mod tests {
         }
     }
 
+    /// Regression test for a dictionary-codec soundness bug in the safe-install
+    /// path (`new_safe`), reachable with the paged batcher enabled.
+    ///
+    /// A from-scratch container stores its pre-install rows *raw* while gathering
+    /// statistics, then installs a *safe* codec via `new_safe`. `new_safe` used to
+    /// discard the first-byte bitmap gathered over those raw rows. That bitmap is
+    /// soundness-critical: a later `new_from` merge consults it to decide which
+    /// one-byte tags are free to hand out as dictionary keys. With the bitmap
+    /// dropped, the merge could assign a dictionary tag equal to a raw datum's
+    /// first byte, after which `decode` resolves that literal datum to the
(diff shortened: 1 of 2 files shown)


def-:
Closes https://linear.app/materializeinc/issue/CLU-110/dictionary-compressed-arrangements-cause-wrong-results I assume?

DAlperin (approved):
LGTM, thanks

PR #36993, opened 2026-06-11
ci-builder/mzbuild: Retry docker push

def-:
Recently seen a lot of flakiness in pushes, for example in https://materializeinc.slack.com/archives/C01LKF361MZ/p1781186475438249


antiguru (approved):
Thank you!

PR #36995, opened 2026-06-11
sqllogictest: snapshot EXPLAIN plans for all mz_catalog_server objects

antiguru:
### Motivation

Closes database-issues#9730 (SQL-350).

Changing the SQL of one builtin catalog object can silently shift the query plans, and thus the maintained arrangements and memory profile, of other objects on `mz_catalog_server`.
Those downstream changes are otherwise invisible in a PR diff.

### Description

Adds an auto-generated sqllogictest snapshot of the `EXPLAIN` plan for every index, materialized view, and system-schema view on `mz_catalog_server` (276 objects).
`test/catalog-server-explain/mzcompose.py` owns the query list (enumerated from the catalog) and CI asserts it covers exactly the current objects; sqllogictest owns the recorded plans.

Two determinism fixes were needed:

* Per-replica introspection log indexes (`ClusterIntrospectionSourceIndex`, `si` id prefix) are excluded: they have no optimizer plan and their names embed a cluster id.
* A new `replace` sqllogictest directive masks non-deterministic output before compare/rewrite, used here for the folded `mz_now()` timestamp in `mz_wallclock_global_lag`.

### Verification

Re-keying `mz_cluster_replica_frontiers_ind` produces a focused diff limited to the index and its genuine dependents (`mz_hydration_statuses_ind`, `mz_compute_hydration_statuses`), confirming the snapshot surfaces exactly the affected plans.


ci/test/pipeline.template.yml
```diff
@@ -329,6 +329,20 @@ steps:
...
+    timeout_in_minutes: 30
+    inputs:
+      - test/catalog-server-explain
+      - test/sqllogictest/catalog_server_explain.slt
+      - src/catalog/src/builtin
+    plugins:
+      - ./ci/plugins/mzcompose:
+          composition: catalog-server-explain
+    agents:
+      queue: hetzner-aarch64-8cpu-16gb
```
def-: 4cpu-8gb is probably more than enough

ci/test/pipeline.template.yml
```diff
@@ -329,6 +329,20 @@ steps:
...
       queue: hetzner-aarch64-16cpu-32gb
 
+  - id: catalog-server-explain
+    label: Catalog server EXPLAIN coverage
+    depends_on: build-aarch64
+    timeout_in_minutes: 30
+    inputs:
+      - test/catalog-server-explain
+      - test/sqllogictest/catalog_server_explain.slt
+      - src/catalog/src/builtin
```
def-: I think changes in other files can also have an impact on the plans. We also run all tests for any change in src anyway from what I can remember, so this line seems useless.

doc/developer/sqllogictest.md
```diff
@@ -168,6 +168,28 @@ pipes the contents of file.csv to the following SQL statement:
...
+### `replace` extension
+
+The `replace` directive registers a regular-expression substitution that is
+applied to the actual output of every subsequent `query` before it is compared
+against (or rewritten into) the expected output.
+It exists to mask non-deterministic tokens that would otherwise make a snapshot
+unstable, for example a folded `mz_now()` timestamp in an `EXPLAIN` plan.
+
+```
+replace (> )\d{13}  ${1}<TIMESTAMP>
```
def-, line 180: How did we solve this previously? I thought we always had a bunch of timestamps in our output. I remember that we have regex-replacements in testdrive already.

ci/test/pipeline.template.yml
```diff
@@ -329,6 +329,20 @@ steps:
     agents:
       queue: hetzner-aarch64-16cpu-32gb
 
+  - id: catalog-server-explain
+    label: Catalog server EXPLAIN coverage
```
def-: The name is quite confusing. It just fails if the catalog_server_explain.slt isn't up to date? Then it feels more like a linter

test/catalog-server-explain/mzcompose.py
def-, line 1: I vote for putting this entire logic into test/sqllogictest/mzcompose.py, and running the validation as part of the default workflow we use in CI (for one of the shards). Then the regeneration just becomes a single command like `bin/mzcompose --find sqllogictest run catalog_server_explain_rewrite`.

def- (commented):
The whole idea is that when the plans change we have to take a look at them and consider if the change makes sense, right?
I *think* this is useful to have, but we have previously had similar things (in testdrive) to capture resource usage, and have removed them again because it was too annoying to have the CI roundtrip and it seemed not to cause any reflection.

def- (changes requested):
The replacements in SLT have to be reset at the end of each handled SLT file, otherwise they also replace in the next file!

antiguru:
Addressed in 5b26fe65ef:

* **Replacement reset** — `replace` directives are now cleared at the start of each file (`run_string`/`rewrite_file`), so they no longer leak into later files in a multi-file invocation. Verified with a two-file run where the second file keeps a value the first file's `replace` would have rewritten.
* **Folded into the SLT composition** — moved everything into `test/sqllogictest/mzcompose.py` and deleted `test/catalog-server-explain/`. Coverage validation now runs on shard 0 of the existing `sqllogictest-fast` step, so the separate composition and pipeline step are gone (which also resolves the step-name, `if_changed`, and agent-size comments). Regeneration is a single command: `bin/mzcompose --find sqllogictest run catalog-server-explain --rewrite`.
* I kept rewrite as a `--rewrite` modifier on the check workflow rather than a standalone `*_rewrite` workflow, so the destructive path is explicit and isn't auto-discovered/auto-run by the `default` workflow.
* Re prior art: sqllogictest had no output-substitution mechanism; this `replace` directive mirrors testdrive's regex replacements.

def- (approved):
Thanks

doc/developer/sqllogictest.md
antiguru: I'm not sure we did, or it was redacted and hence would compare equal?

PR #36997, opened 2026-06-11
[SS-188] Check the `token_uri` in Google Cloud Service Account keys

ublubu:
This PR adds a simple framework for checking the Secrets used by a Connection when either the Secret or the Connection changes.

So far, it's just a hook to check GCP secrets for a bad `token_uri` field.

This is a followup to #36694.

_Just for fun: [Opus vs Fable](https://github.com/ublubu/materialize/compare/f1cc6f85bd0e70bfafc6aec843e54b64087fc762..ublubu:materialize:3eb3a022489401038eab4f84cbdff0a184ff2063?diff=split&w) code comparison_

### DIFF
--- a/src/adapter/src/coord/sequencer/inner.rs
+++ b/src/adapter/src/coord/sequencer/inner.rs
@@ -647,8 +647,8 @@ impl Coordinator {
         secret_id: CatalogItemId,
         contents: &str,
     ) -> Result<(), AdapterError> {
-        for user_id in self.catalog().get_entry(&secret_id).used_by() {
-            if let CatalogItem::Connection(conn) = self.catalog().get_entry(user_id).item() {
+        for dependent_id in self.catalog().get_entry(&secret_id).used_by() {
+            if let CatalogItem::Connection(conn) = self.catalog().get_entry(dependent_id).item() {
                 for (guarded_id, guard) in conn.details.secret_content_guards() {
                     if guarded_id == secret_id {
                         guard(contents)?;
--- a/src/adapter/src/coord/sequencer/inner.rs
+++ b/src/adapter/src/coord/sequencer/inner.rs
@@ -64,7 +64,6 @@ use mz_sql::plan::{
     StatementContext,
 };
 use mz_sql::pure::{PurifiedSourceExport, generate_subsource_statements};
-use mz_storage_types::connections::gcp::GcpServiceAccountKeyTokenUri;
 use mz_storage_types::sinks::StorageSinkDesc;
 // Import `plan` module, but only import select elements to avoid merge conflicts on use statements.
 use mz_sql::plan::{
@@ -625,6 +624,41 @@ impl Coordinator {
         }
     }
 
+    /// Applies `details.secret_content_guards()` by reading each guarded
+    /// secret's current contents. Must be called whenever a connection is
+    /// created or its options altered, before the change takes effect.
+    async fn check_connection_secret_content_guards(
+        &self,
+        details: &ConnectionDetails,
+    ) -> Result<(), AdapterError> {
+        for (secret_id, guard) in details.secret_content_guards() {
+            let contents = self.caching_secrets_reader.read_string(secret_id).await?;
+            guard(&contents)?;
+        }
+        Ok(())
+    }
+
+    /// Applies the secret-content guards of every connection that uses
+    /// `secret_id` against proposed new `contents` for that secret. Must be
+    /// called whenever a secret's contents change, before the new value is
+    /// persisted.
+    pub(super) fn check_secret_content_guards_of_dependents(
+        &self,
+        secret_id: CatalogItemId,
+        contents: &str,
+    ) -> Result<(), AdapterError> {
+        for user_id in self.catalog().get_entry(&secret_id).used_by() {
+            if let CatalogItem::Connection(conn) = self.catalog().get_entry(user_id).item() {
+                for (guarded_id, guard) in conn.details.secret_content_guards() {
+                    if guarded_id == secret_id {
+                        guard(contents)?;
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
     #[instrument]
     pub(super) async fn sequence_create_connection(
         &mut self,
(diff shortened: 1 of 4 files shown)


src/sql/src/plan.rs
```diff
@@ -1715,6 +1715,24 @@ impl ConnectionDetails {
...
+    /// the returned secrets change (e.g. `ALTER SECRET`).
+    pub fn secret_content_guards(
+        &self,
+    ) -> Vec<(CatalogItemId, fn(&str) -> Result<(), anyhow::Error>)> {
+        match self {
+            // A service-account key defines its own OAuth2 token URI. We only
+            // want to send requests to the actual Google OAuth2 token API.
+            ConnectionDetails::Gcp(gcp) => vec![(
+                gcp.credentials_json,
+                GcpServiceAccountKeyTokenUri::validate_json,
```
peterdukelarsen, line 1738: So this is tuple of item ID (which I assume points to json credentials) and a validation function for that json.

Then more broadly this is structured so we can add validation functions from varied ConnectionDetails.

Commentary: Returning a mapping of catalog item and function from each connection detail to later call feels similar to me to having an abstract method of a parent class implemented by a handful of subclasses.

src/adapter/src/coord/sequencer/inner.rs
```diff
@@ -625,6 +624,41 @@ impl Coordinator {
...
+    /// Applies the secret-content guards of every connection that uses
+    /// `secret_id` against proposed new `contents` for that secret. Must be
+    /// called whenever a secret's contents change, before the new value is
+    /// persisted.
+    pub(super) fn check_secret_content_guards_of_dependents(
+        &self,
+        secret_id: CatalogItemId,
+        contents: &str,
+    ) -> Result<(), AdapterError> {
+        for user_id in self.catalog().get_entry(&secret_id).used_by() {
```
peterdukelarsen: user_id is a bit of a confusing name for the dependent catalog item ids.

Also, just to confirm:
1. used_by is only direct dependencies
2. connection always directly references the the secret -- as opposed to some possibility of transitive relationships/constraints
3. reads from catalog are fast in-memory reads so sequential reads or relatively high counts of used_by entries aren't a concern for latency of processing the query

src/adapter/src/coord/sequencer/inner.rs
```diff
@@ -664,22 +698,18 @@ impl Coordinator {
...
             }
-            ConnectionDetails::Gcp(gcp) => {
-                // A service account key defines its own OAuth2 token URI.
-                // We only want to send requests to the actual Google OAuth2 token API,
-                // so we inspect the key as early as we can.
-                if let Err(err) = self
-                    .caching_secrets_reader
-                    .read_string(gcp.credentials_json)
-                    .await
-                    .and_then(|json| GcpServiceAccountKeyTokenUri::validate_json(&json))
```
peterdukelarsen, line 675: Just to confirm, at a high level we're doing two things in this PR overall:
1. Porting this logic to more cases -- i.e. validating connection details on alter secret ad not just creating the connection
2. Updating the code to make it more abstract so we can set up constraints on the contents of a secret depended on by a connection

test/testdrive/connection-create-external.td
```diff
@@ -263,3 +263,37 @@ contains:token_uri must be https://oauth2.googleapis.com/token
...
+contains:token_uri must be https://oauth2.googleapis.com/token
+
+# Benign changes must still go through: a guard that rejected every ALTER would
+# also pass the tests above. ALTER SECRET first, while gcp_alter_conn still
+# depends on gcp_good -- afterwards the connection moves off it and the
+# dependent-based guard no longer applies.
+> ALTER SECRET gcp_good AS '{"type":"service_account","project_id":"y","client_email":"a@b.com","private_key":"${gcp-ssrf-key}","token_uri":"https://oauth2.googleapis.com/token"}'
+
+> CREATE SECRET gcp_good_2 AS '{"type":"service_account","project_id":"x","client_email":"a@b.com","private_key":"${gcp-ssrf-key}","token_uri":"https://oauth2.googleapis.com/token"}'
+> ALTER CONNECTION gcp_alter_conn SET (SERVICE ACCOUNT KEY = SECRET gcp_good_2) WITH (VALIDATE = false);
```
peterdukelarsen, line 299: This testing looks good to me.

src/adapter/src/coord/sequencer/inner/secret.rs
```diff
@@ -282,6 +282,8 @@ impl Coordinator {
         let caching_secrets_reader = self.caching_secrets_reader.clone();
         let secrets_controller = Arc::clone(&self.secrets_controller);
         let payload = self.extract_secret(session, &mut secret_as)?;
+        let contents = std::str::from_utf8(&payload).expect("validated as UTF-8 by extract_secret");
```
peterdukelarsen, line 285: validated that "validated as UTF-8 by extract_secret" is true

peterdukelarsen (approved):

src/adapter/src/coord/sequencer/inner.rs
ublubu: I had a poke at making it less abstract (I mean, there's only one branch, "Gcp". Come on.) but I couldn't do it.
So I added a comment to `secret_content_guards` to explain why it's like that.

```rust
    /// We rely on the caller to actually execute these checks because we don't know:
    /// - which secrets the caller cares about
    /// - which secrets require an async operation to fetch
    /// 
    /// For example, the ALTER SECRET caller should only perform checks on its own secret,
    /// while the ALTER CONNECTION caller fetches and checks every secret from its connection.
    pub fn secret_content_guards(
        &self,
    ) -> Vec<(CatalogItemId, fn(&str) -> Result<(), anyhow::Error>)>
```

src/adapter/src/coord/sequencer/inner.rs
ublubu: 1. Yes
2. For GCP Connections, yes. (I actually don't know if it holds for all the other connection types 😮)
3. Good question. Time to ask Marty. _EDIT: Yes, we have a cheap in-memory view of the catalog, including `used_by`._

src/sql/src/plan.rs
ublubu: re: Commentary

The short answer is yes, it's definitely similar. And enum variants are kinda like subclasses.

Slightly longer answer: https://github.com/MaterializeInc/materialize/pull/36997#discussion_r3398966286

PR #37006, opened 2026-06-12
feature-benchmark: Ignore ~10% regression in ParallelDataflows

def-:
Reported in https://linear.app/materializeinc/issue/CLU-116/dictionary-compressed-arrangements-causes-10percent-time-regression-in


antiguru (approved):
Thank you.

PR #37008, opened 2026-06-12
clusterd-test-driver: headless frontend to clusterd for scripted compute tests

antiguru:
### Motivation

Running a small compute experiment today means standing up a full `environmentd` — the SQL layer, the catalog, the coordinator — even when all you want is to hand `clusterd` a few commands and watch what it does. That is slow to set up and hard to control precisely.

This adds `mz-clusterd-test-driver`: a headless frontend that speaks the compute protocol to `clusterd` directly, with no `environmentd`. A test drives it from a text script, so it controls the exact persist state, the exact commands the replica sees, and the exact timestamps. Design doc: `doc/developer/design/20260612_headless_clusterd_test_driver.md`.

### What a script looks like

Each stanza is a command followed by a `----` block holding its expected output. That block *is* the assertion, and `REWRITE=1` regenerates it in place. Here is the whole lifecycle of a `count(*)` materialized view over a persist shard, read back at the end:

```
create-instance
----
ok

initialization-complete
----
ok

write-single-ts shard=events ts=0 count=3000
----
wrote 3000

define-schema name=count_out
  count bigint
----
ok

create-dataflow name=count-mv as-of=0
  import source=1000 shard=events upper=1
  build id=2000
    Reduce aggregates=[count(*)]
      Get u1000
  export kind=materialized-view sink=2001 on=2000 shard=mv schema=count_out
----
ok

schedule id=2001
----
ok

allow-writes id=2001
----
ok

peek id=2001 schema=count_out ts=0
----
3000
```

A command that fails renders as `error: <message>`, so an expected failure is just another golden block — there is no special assertion command. Because the waits are level-triggered on monotonic frontiers, the order a script waits in does not change the result, so a single sequential script stays deterministic.

### How it works

The crate is a generic mechanism, a dataflow builder, and the scripting layer on top.

* The mechanism hosts the persist PubSub server, connects over CTP (sending only `Hello`), and exposes a `Driver` that sends any `ComputeCommand`, submits dataflows without auto-scheduling, watches frontiers, and peeks.
* `DataflowBuilder` takes generic parts — persist imports, MIR objects to build, and index/materialized-view/subscribe exports — and runs the real MIR → LIR → `RenderPlan` lowering, because a `RenderPlan` can't be hand-built outside `mz-compute-types`. A `build`'s computation is written in the `mz-expr-parser` `.spec` syntax (`Reduce aggregates=[count(*)]` over `Get u1000`) rather than a bespoke vocabulary, since `MirRelationExpr`'s own serde isn't hand-authorable (`Row` literals are opaque bytes).
* `create-dataflow` is the one abstraction behind index, materialized-view, and subscribe exports (`copy-to` isn't implemented yet). A materialized view is read back by `peek`ing the sink id — that becomes a persist peek of its output shard, the same path `SELECT * FROM mv` takes — and a subscribe streams responses that the driver buffers and `await-subscribe` drains. Dataflows start read-only, so a sink needs `allow-writes` before its writes land.
* The handshake is explicit (`create-instance`, optional `update-configuration`, `initialization-complete`), and `reconnect` re-runs it without `initialization-complete` to exercise reconciliation.

### Verification

`mzcompose` runs each scenario script against a real `clusterd` and fails on any golden mismatch; the scenarios are index, deep-history, side-effects, multi-dataflow, reconciliation, error-behavior, reduce, materialized-view, and subscribe. Unit tests cover the direct-write round trip, the lowered dataflow structure, and the script parser, and `run-local.py` runs the same scripts on the host (with `REWRITE=1`) without docker images.


src/compute-test-driver/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+pub mod responses;
+pub mod target;
+
+#[cfg(test)]
+mod tests {
+    #[mz_ore::test]
+    fn it_links() {
+        assert_eq!(2 + 2, 4);
+    }
+}
```
antiguru: Please remove this test.

antiguru (commented):
What we're missing is a  specific language to script the headless driver. Encoding the interactions in Rust is a good first step, but ultimately we want something that is easier to iterate on. I could even imagine Python scripts that we load through pyo3 or so.

src/compute-test-driver/src/lib.rs
antiguru: Removed in d04a969676.

antiguru:
Agreed on a scripting language being the goal. I captured it under "Future work" in the design doc: the mechanism is already a thin scriptable surface (`send`, `submit_dataflow`, `schedule`, `expect_frontier`, `peek`, `subscribe_raw`), so a declarative script or Python-via-pyo3 layer would bind to it rather than replace it. Left it out of this PR to keep the first step focused.

def- (commented):
I have once built something very similar at a previous company, also using json inputs to test an execution engine directly instead of going through the entire database stack. Back then the test framework turned out to not be very useful since it was easier to write unit tests and full system tests instead of the middle-ground headless json tests.

Do we have some concrete examples of regression tests we've wanted to build and failed before, which would be possible with clusterd-test-driver?

I still think it's a worthy experiment, and I can especially imagine that building a [fuzzer](https://github.com/MaterializeInc/materialize/pull/36982) for clusterd-test-driver with some invariants could lead us to interesting bugs.

test/clusterd-test-driver/mzcompose.py
```diff
@@ -0,0 +1,89 @@
...
+                # /workdir/scripts (the convention testdrive uses for its files);
+                # the driver reads the file named by DRIVER_SCRIPT.
+                "volumes": [".:/workdir"],
+                "ports": [6879],
+            },
+        )
+
+
+SERVICES = [
+    Cockroach(setup_materialize=True),
```
antiguru: Can we use the metadata store that automatically dispatches to the DB configured through the environment?

antiguru: Done in d8bf9727b1. The composition now pulls the store in via `metadata_store_companions()` (dispatches to the env-configured DB, default `postgres-metadata`) and derives the consensus URI from `METADATA_STORE` instead of hardcoding Cockroach. Guarded to a SQL store, since persist consensus needs a Postgres-wire backend (FoundationDB wouldn't work).

src/clusterd-test-driver/src/script.rs
```diff
@@ -0,0 +1,1199 @@
...
+            } => {
+                let timeout = Duration::from_secs(timeout_secs.unwrap_or(DEFAULT_TIMEOUT_SECS));
+                let updates = self
+                    .driver
+                    .await_subscribe(GlobalId::User(id), Timestamp::from(up_to), timeout)
+                    .await?;
+                Ok(render_updates(&updates))
+            }
+            Command::CreateInstance => {
+                self.driver.create_instance()?;
```
antiguru: This should carry all `CreateInstance` parameters. The default if not specified can be the default that we're currently supplying.

src/clusterd-test-driver/src/script.rs
```diff
@@ -0,0 +1,1199 @@
...
+                Ok("ok".to_string())
+            }
+            Command::UpdateConfiguration {
+                peek_response_stash,
```
antiguru: This is overly specific to `peek_response_stash`. `peek_response_stash` must be `false` for the driver to work, so we can just patch it, but otherwise this should accept a table of config name, type, value:
```
update-configuration
  enable_my_flag bool true
  duration_falg duration 1s
----
ok
```

src/clusterd-test-driver/src/script.rs
antiguru: Done in 1eb7d5522d. `create-instance` now carries the settable `InstanceConfig` knobs — `expiration-offset` (a duration) and `arrangement-dictionary-compression` — defaulting to the values it supplied before. `logging` stays at its default (introspection logging off; enabling it safely needs `index_logs` wiring) and `peek_stash_persist_location` is necessarily the host's, since the driver hosts persist. The peek-response stash is force-disabled in the driver rather than exposed as a knob (see the next thread).

src/clusterd-test-driver/src/script.rs
antiguru: Done in 1eb7d5522d. `peek_response_stash` is now force-disabled at `create-instance` (patched in the driver, not a knob — see the previous thread). `update-configuration` takes a generic `name type value` table, e.g.:

```
update-configuration
  enable_my_flag bool true
  my_dur duration 1s
----
ok
```

Types `bool`/`u32`/`usize`/`f64`/`duration`/`string`, parsed by `mz_dyncfg`'s own `ConfigType::parse` and applied via `ConfigUpdates::add_dynamic` (so arbitrary names go through by name).

test/clusterd-test-driver/scripts/join.spec
```diff
@@ -0,0 +1,67 @@
...
+  1 x
+  2 y
+  2 z
+----
+wrote 3
+
+# Join the two sources on their key column — `#0` (left key) equals `#2` (right key)
+# in the 4-column output — and arrange the result. `optimize` is what lets the
+# `Join` lower.
+create-dataflow name=join as-of=0 optimize
```
DAlperin, line 45: Not blocking, but it occurs to me that when the caller opts into the optimize verb they might want to be able to assert that the optimized variant looks a certain way so that subtle optimizer behavior drift can't unknowingly meaningfully change the test.

DAlperin: Actually the more I think about this its probably pretty easy and worth doing since it otherwise breaks the spirit of the test framework a bit IMO

src/clusterd-test-driver/src/ctp.rs
```diff
@@ -0,0 +1,77 @@
...
+
+/// Connects to a clusterd compute controller address and sends `Hello`, leaving
+/// the controller handshake (`CreateInstance` onward) to the caller.
+///
+/// A reconnect re-runs exactly this: a fresh transport connection plus `Hello`.
+/// The reconciliation window then opens when the script sends `CreateInstance`
+/// and closes when it sends `InitializationComplete`; in between, the replica
+/// reconciles the replayed dataflows against its live ones rather than
+/// rehydrating.
+pub async fn connect_and_hello(compute_addr: &str) -> anyhow::Result<ComputeCtpClient> {
```
DAlperin, line 37: This is only called from the driver. Might not need to be its own file

src/clusterd-test-driver/src/driver.rs
```diff
@@ -0,0 +1,270 @@
...
+                    let mut iter = collection.into_row_iter();
+                    while let Some(row_ref) = iter.next() {
+                        rows.push(row_ref.to_owned());
+                    }
+                }
+                Ok(rows)
+            }
+            PeekResponse::Error(e) => anyhow::bail!("peek error: {e}"),
+            PeekResponse::Canceled => anyhow::bail!("peek canceled"),
+            PeekResponse::Stashed(_) => anyhow::bail!("unexpected stashed peek result"),
```
DAlperin, line 220: I thought for sure I had claude out here but it explicitly disabled the peek stash so this is fine:)

DAlperin (approved):
Seems sane to me and like a good test framework to have.

test/clusterd-test-driver/scripts/join.spec
antiguru: Good call — done in a stacked PR (antiguru#182). Added an `explain` verb: same body as `create-dataflow`, but it renders the lowered LIR (`EXPLAIN PHYSICAL PLAN` form) as the golden instead of submitting. `join.spec` now asserts the differential-join plan shape alongside the count, so optimizer/lowering drift surfaces. The render is e.g.:

```
u2000:
  →Differential Join %0:u1000[#0] » %1:u1001[#0]
    →Arrange (#0)
      →Stream u1000
    →Arrange (#0)
      →Stream u1001
```

src/clusterd-test-driver/src/ctp.rs
antiguru: It's shared by two modules, not just the driver: `driver.rs` uses `connect_and_hello`, and `responses.rs` uses the `ComputeCtpClient` type for the response pump. Folding it into `driver.rs` would make the response pump depend on the driver just for the transport type, so I'd keep the CTP primitives in their own small module. Happy to revisit if you feel strongly.

DAlperin: Missed it in my grep. My bad!

antiguru:
Thanks for the review!

PR #37007, opened 2026-06-12
sqllogictest: regenerate catalog_server_explain for new builtin MVs

antiguru:
### Motivation

`catalog_server_explain.slt` broke on main (build 125374).
PR #36995 snapshotted EXPLAIN plans for all `mz_catalog_server` objects, but two builtin materialized views (`mz_role_parameters`, `mz_roles`) merged concurrently.
The snapshot was stale: `mz_materialized_views` and `mz_builtin_materialized_views` reported 13 instead of 15 constant rows, and the two new MVs lacked their own EXPLAIN entries.

### Description

Regenerated the file with `bin/mzcompose --find sqllogictest run catalog-server-explain --rewrite` followed by `bin/sqllogictest -- --rewrite-results test/sqllogictest/catalog_server_explain.slt`.
This adds EXPLAIN blocks for the two new MVs and corrects the row counts.

### Verification

`bin/sqllogictest -- test/sqllogictest/catalog_server_explain.slt` → PASS 280/280.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #37009, opened 2026-06-12
sqllogictest: fill catalog_server_explain plans in one rewrite step

antiguru:
### Motivation

Follow-up to #37007. As def- noted there, regenerating `catalog_server_explain.slt` was a two-step dance: `catalog-server-explain --rewrite` only rebuilt the query list (new objects landed with empty plans), then a separate `bin/sqllogictest -- --rewrite-results` was needed to fill the plans.

### Description

Run the plan fill inside the `--rewrite` workflow, so a single `bin/mzcompose --find sqllogictest run catalog-server-explain --rewrite` produces a ready-to-commit file.
The fill reuses the existing `SltRunStepConfig.to_command` machinery in rewrite mode.
Replica configuration is shared between the fill and `run_sqllogictest`'s defaults via new `DEFAULT_REPLICA_SIZE`/`DEFAULT_REPLICAS` constants, so the rewrite reproduces the plans CI validates.

### Verification

Ran `bin/mzcompose --find sqllogictest run catalog-server-explain --rewrite` on a clean tree: the single invocation regenerated the file byte-for-byte with no diff.

### DIFF
--- a/test/sqllogictest/mzcompose.py
+++ b/test/sqllogictest/mzcompose.py
@@ -82,11 +82,6 @@ CATALOG_SERVER_SYSTEM_SCHEMAS = [
     "mz_unsafe",
 ]
 
-REGENERATE_HINT = (
-    "  bin/mzcompose --find sqllogictest run catalog-server-explain --rewrite"
-)
-
-
 def workflow_default(c: Composition) -> None:
     # `catalog-server-explain` validation runs inside `fast_tests` (shard 0), so
     # skip it here to avoid running it twice (and it is `--rewrite`-capable, so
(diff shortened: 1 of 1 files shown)


test/sqllogictest/mzcompose.py
```diff
@@ -77,8 +83,7 @@
 ]
 
 REGENERATE_HINT = (
-    "  bin/mzcompose --find sqllogictest run catalog-server-explain --rewrite\n"
-    "  bin/sqllogictest -- --rewrite-results test/sqllogictest/catalog_server_explain.slt"
+    "  bin/mzcompose --find sqllogictest run catalog-server-explain --rewrite"
 )
```
antiguru: Seems we can inline the hint as it is rather short now.

def- (approved):

PR #37012, opened 2026-06-12
ci: Annotate SLT failures more reliably

def-:
Seen in https://buildkite.com/materialize/test/builds/125383#019ebb7c-6de7-48c5-9d99-04c1987c908c
Follow-up to https://github.com/MaterializeInc/materialize/pull/36995 which extended SLT to have a new kind of error, not in junit.xml

### DIFF
--- a/misc/python/materialize/cli/ci_annotate_errors.py
+++ b/misc/python/materialize/cli/ci_annotate_errors.py
@@ -551,6 +551,32 @@ def annotate_logged_errors(
     errors = get_errors(log_files)
 
     if not errors:
+        # No error pattern was detected in the logs and no junit report
+        # contains an error, but the test may still have failed, so annotate
+        # the failure anyway so that it doesn't go unnoticed, and show the
+        # current failures on main branch.
+        # Only fetch the main branch status when we are running in CI, but not
+        # on main, so inside of a PR or a release branch instead.
+        if (
+            ui.env_is_truthy("BUILDKITE")
+            and os.getenv("BUILDKITE_BRANCH") != "main"
+            and test_result != 0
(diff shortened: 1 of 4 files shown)


antiguru (commented):
Claude says:

Nice — this fixes the actual mechanism, and handling the crash-with-no-report case (flag only sets when a non-empty junit was written) is the part a pure log-regex misses.

Three things:

1. (the one I'd want resolved) Moving the fallback into `if not errors:` narrows it versus the old guard. The old tail block fired on `len(unknown_errors)==0 and len(known_errors)==0`, but `errors` is the raw detected set, before IGNORE_RE / known-issue classification. So the case "errors were detected but all classified away to zero unknown+known, yet `test_result != 0`" used to get the safety-net annotation and now gets nothing — the new block is unreachable for it and the old tail is deleted. Intended? That "failed for only-ignored reasons but still red" path seems like exactly what the net was for.

2. Passing sqllogictest shards now emit an mzcompose-level junit where they emitted none before (the static `"sqllogictest"` exclusion is gone and the flag only sets on failure). Not a duplicate, since per-file junits are `os.remove`d on success — but it's a new Test Analytics entry per shard. Just confirming that's intended and doesn't skew analytics.

3. `c.has_sqllogictest_junit = True` is set from the `ThreadPoolExecutor` workers. Fine under the GIL (bool store), just noting the cross-thread mutation.

Only 1 blocks; 2 and 3 are confirm-and-move-on.

def-:
Thanks, fixed #1. The other two seems fine to me.

antiguru (approved):

antiguru (approved):
Thank you!

PR #37013, opened 2026-06-12
ci: Run slow single replica SLT files in Nightly

def-:
Test run: https://buildkite.com/materialize/nightly/builds/16777

They were entirely forgotten for a few months, so had to fix up some test results.


bosconi (approved):
Looks good.

test/sqllogictest/introspection/singlereplica_attribution_sources.slt
antiguru, line 1: We could use the new `replace` syntax to guard the tests against changing global IDs. They're not static, so changes to the catalog can cause churn here. Specifically, we could replace t[\d]+ with a placeholder.

def-: Oooh right, good idea.

PR #37014, opened 2026-06-12
compute: fix "cannot clone ReadHold" panic during shutdown

def-:
During process shutdown the tokio runtime drops tasks in arbitrary order, so the `StorageCollections` background task can disappear while an `Instance` task still processes a queued `maintain()`. Rehydrating a failed replica then panics in `ReadHold::clone` when registering the clone with the hung-up issuer (seen in sqllogictest, which tears down a server per file).

Add a fallible `ReadHold::try_clone` and use it in `add_replica_state`: on a hung-up issuer, stop populating per-replica collection state but still insert the replica to keep bookkeeping consistent with the controller. `Clone` keeps panicking for all other callers, since a clone that fails to register is not a real read hold.

Fixes [database-issues#9964](https://github.com/MaterializeInc/database-issues/issues/9964)

### DIFF
--- a/src/compute-client/src/controller/instance.rs
+++ b/src/compute-client/src/controller/instance.rs
@@ -378,7 +378,34 @@ impl Instance {
                 collection.read_frontier().to_owned()
             };
 
-            let input_read_holds = collection.storage_dependencies.values().cloned().collect();
+            // Cloning a `ReadHold` fails when its issuer has hung up. For these holds the issuer
+            // is the `StorageCollections`, which doesn't hang up as long as the `Instance` exists,
+            // except during process shutdown, when the tokio runtime drops tasks in arbitrary
+            // order. In that case there is no way of correctly initializing the per-replica
+            // collection state, so we stop trying. We still add the replica itself, to keep the
+            // bookkeeping consistent with the controller's, but it will never hydrate.
+            let mut input_read_holds = Vec::with_capacity(collection.storage_dependencies.len());
+            let mut hung_up = None;
+            for hold in collection.storage_dependencies.values() {
+                match hold.try_clone() {
+                    Ok(hold) => input_read_holds.push(hold),
+                    Err(_) => {
+                        hung_up = Some(hold.id());
+                        break;
+                    }
+                }
+            }
+            if let Some(input_id) = hung_up {
+                tracing::info!(
+                    replica_id = %id,
+                    %collection_id,
+                    %input_id,
+                    "giving up on adding replica collections: storage read hold issuer hung \
+                     up, the process is shutting down",
+                );
+                break;
+            }
+
             replica.add_collection(*collection_id, as_of, input_read_holds);
         }
 
--- a/src/compute-client/src/controller/instance.rs
+++ b/src/compute-client/src/controller/instance.rs
@@ -345,7 +345,7 @@ impl Instance {
         client: ReplicaClient,
         config: ReplicaConfig,
         epoch: u64,
-    ) {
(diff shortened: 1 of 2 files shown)


src/compute-client/src/controller/instance.rs
```diff
@@ -378,7 +378,34 @@ impl Instance {
...
 
-            let input_read_holds = collection.storage_dependencies.values().cloned().collect();
+            // Cloning a `ReadHold` fails when its issuer has hung up. For these holds the issuer
+            // is the `StorageCollections`, which doesn't hang up as long as the `Instance` exists,
+            // except during process shutdown, when the tokio runtime drops tasks in arbitrary
+            // order. In that case there is no way of correctly initializing the per-replica
+            // collection state, so we stop trying. We still add the replica itself, to keep the
+            // bookkeeping consistent with the controller's, but it will never hydrate.
+            let mut input_read_holds = Vec::with_capacity(collection.storage_dependencies.len());
+            let mut hung_up = None;
```
antiguru: Should this be a vector? It'd name all collections in the error message.

src/storage-types/src/read_holds.rs
```diff
@@ -165,10 +165,16 @@ impl ReadHold {
...
-    fn clone(&self) -> Self {
+    /// Clones this [ReadHold], returning an `Err` when the issuer of the read
+    /// hold has hung up, in which case the clone would not actually hold back
+    /// the since of the collection.
+    ///
+    /// The issuer hanging up is only expected during process shutdown, when
+    /// the tokio runtime drops tasks in arbitrary order. Callers that may run
+    /// concurrently with shutdown can use this method to handle that case
+    /// gracefully, instead of panicking via [Clone::clone].
+    pub fn try_clone(&self) -> Result<Self, SendError<(GlobalId, ChangeBatch<Timestamp>)>> {
```
antiguru: The `SendError` is not helpful for the caller, as it's just an indication that the other end has hung up. Either return a concrete error, or an option.

src/compute-client/src/controller/instance.rs
```diff
@@ -378,7 +378,34 @@ impl Instance {
                 collection.read_frontier().to_owned()
             };
 
-            let input_read_holds = collection.storage_dependencies.values().cloned().collect();
+            // Cloning a `ReadHold` fails when its issuer has hung up. For these holds the issuer
+            // is the `StorageCollections`, which doesn't hang up as long as the `Instance` exists,
+            // except during process shutdown, when the tokio runtime drops tasks in arbitrary
+            // order. In that case there is no way of correctly initializing the per-replica
+            // collection state, so we stop trying. We still add the replica itself, to keep the
+            // bookkeeping consistent with the controller's, but it will never hydrate.
```
antiguru: The issue with the below code is that it leaves the replica in an invalid state, even if it's just momentarily. Ideally we return an error to indicate an unrecoverable error that can only happen during shutdown, which in turn should shutdown the compute controller.

This is a risk since the log is an info message, which can easily be overlooked. Consider making it an error message.

antiguru (changes requested):
I left some comments, but the main hesitation I have with this change is that it leaves the replica and in turn the controller state in an invalid configuration. I'd prefer if we bubble up the error instead of swallowing it, to protect us from genuine storage dependency failures that now would turn into info logs.

def- (commented):
lgtm. Keeps failing in CI

antiguru (approved):

antiguru:
Thanks for the re-review! ;-)

PR #37015, opened 2026-06-12
Update catalog_server_explain.slt

def-:
Seen failing in https://buildkite.com/materialize/test/builds/125398
Following https://github.com/MaterializeInc/materialize/pull/36999

Is it expected that we have -1600 lines? Seems kind of surprising to me.


antiguru (approved):
Thank you!

PR #37016, opened 2026-06-12
docs: note subnet NACL ephemeral ports for PrivateLink Kafka

materializekatrina:
### Motivation
Follow up to [#incident-1060](https://materializeinc.slack.com/archives/C0B880EQKFE). A listener-port-only NACL on broker/NLB subnets silently drops the ephemeral-port return traffic of cross-AZ connections once cross-zone load balancing is enabled, causing intermittent, self-resolving, all-or-nothing per-broker timeouts. Document the requirement to allow the ephemeral port range (1024-65535) from the VPC CIDR in both the "Verify security groups and health checks" step and the source-creation troubleshooting list of the shared privatelink-kafka shortcode (covers the PrivateLink, Amazon MSK, and self-hosted Kafka pages).

Co-Authored-By: Claude Opus 4.8

### DIFF
--- a/doc/user/layouts/shortcodes/network-security/privatelink-kafka.md
+++ b/doc/user/layouts/shortcodes/network-security/privatelink-kafka.md
@@ -48,6 +48,8 @@ and retrieve the AWS principal needed to configure the AWS PrivateLink service.
 
     b. You can't use the security groups for the clients as a source in the security groups for the targets. Therefore, the security groups for your targets must use the IP addresses of the clients to allow traffic. For more details, check the [AWS documentation](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-register-targets.html).
 
+    c. If the broker or NLB subnets have [network ACLs](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-network-acls.html), they must allow inbound traffic on the ephemeral port range (1024–65535) from the VPC CIDR, not just the listener port. NACLs are stateless, so a listener-port-only rule drops the ephemeral return traffic on cross-AZ connections once [cross-zone load balancing](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html) is enabled, causing intermittent per-broker timeouts.
+
 1. Create a VPC [endpoint service](https://docs.aws.amazon.com/vpc/latest/privatelink/create-endpoint-service.html) and associate it with the **Network Load Balancer** that you’ve just created.
 
     Note the **service name** that is generated for the endpoint service.
@@ -165,3 +167,9 @@ connection you just configured:
 
    * For **in-region connections**, the correct availability zone is specified
       for each broker.
+
+   * The broker/NLB subnets' [network ACL](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-network-acls.html)
+      allows the ephemeral port range (1024–65535) from the VPC CIDR, not just
+      the listener port. A listener-port-only NACL causes intermittent,
+      self-resolving, all-or-nothing per-broker timeouts once cross-zone load
+      balancing is enabled.


doc/user/layouts/shortcodes/network-security/privatelink-kafka.md
```diff
@@ -48,6 +48,8 @@ and retrieve the AWS principal needed to configure the AWS PrivateLink service.
 
     b. You can't use the security groups for the clients as a source in the security groups for the targets. Therefore, the security groups for your targets must use the IP addresses of the clients to allow traffic. For more details, check the [AWS documentation](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/target-group-register-targets.html).
 
+    c. If the broker or NLB subnets have [network ACLs](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-network-acls.html), they must allow inbound traffic on the ephemeral port range (1024–65535) from the VPC CIDR, not just the listener port. NACLs are stateless, so a listener-port-only rule drops the ephemeral return traffic on cross-AZ connections once [cross-zone load balancing](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html) is enabled, causing intermittent per-broker timeouts.
```
kay-kim, line 51: Just want to flag per slack convo that this isn't limited to kafka. 

Some minor:

- would introduce the acronym here [network ACLs (NACLs)] since you reference it in the second sentence.
- the "they" -- ambiguous.

More importantly, from the second sentence ... it's only when you enable cross zone load balancing  and not just if using NACLs? If that is the case (?):
  - We should probably incorporate that into the If ... condition.
  - And, maybe have the second sentence explain the NLB behavior with cross-zone load balancing. and then go into the stateless ...
  - Additionally, Claude is saying that even in a single AZ, brokers and NLB can end up in different subnets.  So, is it slightly broader? 

Let me know which condition is the one we're trying to account for. Depending on the answer to that, have some suggestions.

kay-kim (changes requested):

doc/user/layouts/shortcodes/network-security/privatelink-kafka.md
kay-kim: Oh, also, just a friendly reminder about this other PR: https://github.com/MaterializeInc/materialize/pull/36623/changes

jubrad: My suggestion was 
> If you are using [network ACLs](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-network-acls.html), and network traffic between the NLB and the source application crosses subnet boundaries, NACLs must allow bidirectional flow for all possible paths. This will occur if cross zone load balancing is enabled on the NLB, or if the NLB and the source application are separate subnets even within the same Availability Zone. For example, If a source application is available on port is 9098, the subnet the application is hosted must allow 9098 ingress from *all* NLB ip ranges, and must allow egress to ephemeral port ranges (1024–65535) to all NLB IP ranges. Similarly, if a network load balancer is reaching out to a service on 9098, it must have an egress rule allowing that traffic and an ingress rule allowing return traffic on the ephemeral port range.

Which answers this question:
> Additionally, Claude is saying that even in a single AZ, brokers and NLB can end up in different subnets. So, is it slightly broader?

Yes it is broader... further, it's broader than just kafka, the same issue can arise on RDS (postgres/mysql/sqlserver). 

> it's only when you enable cross zone load balancing and not just if using NACLs? 

It's not just when you enable cross zone load balancing, see my comment above.

kay-kim: Thanks much.  I'm going to create a new branch with both PRs and try to single source.

kay-kim:
Closing since https://github.com/MaterializeInc/materialize/pull/37056 supersedes this.

PR #37019, opened 2026-06-12
Add Glue interchange and wire format header codec

martykulma:
Add Glue wire format decoding to `mz-interchange`.  This only supports uncompressed payloads at the moment, just to keep the changes to a minimum.  Compression support to be added at a later date.

### DIFF
--- a/src/interchange/src/glue.rs
+++ b/src/interchange/src/glue.rs
@@ -1,0 +1,160 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! Wire-format helpers for the AWS Glue Schema Registry framing.
+//!
+//! Glue prepends an 18-byte header to each Kafka record payload:
+//!
+//! | Offset | Bytes | Meaning                                              |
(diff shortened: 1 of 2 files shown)


src/interchange/src/glue.rs
```diff
@@ -0,0 +1,158 @@
...
+//! Wire-format helpers for the AWS Glue Schema Registry framing.
+//!
+//! Glue prepends an 18-byte header to each Kafka record payload:
+//!
+//! | Offset | Bytes | Meaning                                              |
+//! |--------|-------|------------------------------------------------------|
+//! | 0      | 1     | Header version. Glue currently emits `0x03`.         |
+//! | 1      | 1     | Compression byte. `0x00` = none, `0x05` = zlib.      |
+//! | 2..18  | 16    | Schema-version UUID, big-endian.                     |
+//! | 18..   | N     | The serialized record payload.                       |
```
peterdukelarsen, line 19: nit: what do you think about linking the docs/code where this is pulled from for future reference? I think this is where this is coming from, right? https://github.com/awslabs/aws-glue-schema-registry/blob/d1d3cddaa3bfcb8a8f74330c523365e8ba27cc19/common/src/main/java/com/amazonaws/services/schemaregistry/utils/AWSSchemaRegistryConstants.java#L54-L57

peterdukelarsen (approved):
LGTM

src/interchange/src/glue.rs
martykulma: good catch - done!

PR #37023, opened 2026-06-12
docs: add guide for ingesting from a dedicated PostgreSQL replica

josharenberg:
Adds a goal-first guide for the "sidecar" pattern: standing up a dedicated PostgreSQL replica via native logical replication and pointing a Materialize PostgreSQL source at it instead of the production primary.

Previously this pattern was only mentioned inside the PostgreSQL source FAQ, filed under "what if I can't use REPLICA IDENTITY FULL?" — so users whose actual motivation was keeping replication load/WAL-retention risk off their primary, or who couldn't reconfigure the primary, had no discoverable entry point. The new page is titled and framed around that goal and is reachable directly from the PostgreSQL ingest menu.

The guide makes explicit the two logical-replication hops involved (primary -> replica via native PG logical replication, then replica -> Materialize), which means the replica itself needs wal_level = logical, REPLICA IDENTITY FULL on its tables, and a separate publication for Materialize. It leads with a strong recommendation to connect directly to the primary and frames the replica as an exception path, then hands off to the self-hosted guide for source creation rather than duplicating it.

Cross-links added from the FAQ sidecar entry and the self-hosted guide so both motivations funnel into the new page.

No code changes; documentation only.

### Motivation

Closes <issue>
Fixes <bug>
etc.

### Description

diff *can't* tell a reviewer.

### Verification

### DIFF
--- a/doc/user/content/ingest-data/postgres/logical-replica.md
+++ b/doc/user/content/ingest-data/postgres/logical-replica.md
@@ -1,0 +1,246 @@
+---
+title: "Guide: Ingest from a dedicated PostgreSQL replica"
+description: "How to point a Materialize PostgreSQL source at a dedicated logical replica instead of your production primary."
+aliases:
+  - /ingest-data/postgres/sidecar/
+menu:
+  main:
+    parent: "postgresql"
+    name: "Ingest from a dedicated replica"
+    weight: 55
+---
+
+{{< note >}}
+We **nearly always recommend** connecting Materialize directly to your primary
+(see [self-hosted PostgreSQL](/ingest-data/postgres/self-hosted/) or the guide
+for your [hosted service](/ingest-data/postgres/)). The dedicated-replica setup
+on this page is an exception path: reach for it only when you have a **serious,
+specific concern** that direct replication can't address. It adds an extra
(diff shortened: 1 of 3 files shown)


kay-kim:
Thanks for this.  I'll put up a patch to address some tweaks and slight reordering to try to minimize switching from primary to replica.

doc/user/content/ingest-data/postgres/logical-replica.md
```diff
@@ -0,0 +1,294 @@
...
+If none of these apply, prefer connecting directly to the primary — the replica
+adds an extra system to operate and its own replication lag on top of the
+primary.
+
+## How it works
+
+There are **two** logical-replication hops to set up:
+
+```
+primary  ──(native PG logical replication)──▶  replica  ──(Materialize source)──▶  Materialize
```
kay-kim, line 60: We can have this be a mermaid diagram if you want. We just need to
- set in the frontmatter `mermaid: true`
- modify this to a flowchart LR ...

kay-kim (approved):
I left a comment about being able to use mermaid diagram.  Not sure if we want that.  Just let me know if you do.  But, this can also go as is.

PR #37025, opened 2026-06-12
adapter: migrate LaunchDarkly SDK off fork to upstream 3.1.1

jasonhernandez:
**Part 1 of 3** in the LaunchDarkly upstream-SDK stack:

1. **#37025 (this PR)** — migrate off the fork to upstream 3.1.1
2. #37026 — + the reconnect integration test
3. #37460 — + pin to pre-fix 3.0.1 to prove the test fails (**DO NOT MERGE**)

## What this does

Moves `launchdarkly-server-sdk` from the `MaterializeInc/rust-server-sdk` fork back to upstream crates.io **3.1.1**, restoring the `launchdarkly-sdk-transport` + `MetricsTransport` setup and dropping the `[patch.crates-io]` override.

The fork existed for [rust-server-sdk#116](https://github.com/launchdarkly/rust-server-sdk/issues/116): a non-`Eof` stream error left the `StreamingDataSource` stuck with no reconnect, silently breaking LD sync. A prior upgrade to 3.0.1 was reverted (incident-984) because that bug was still unfixed upstream. The fixes have since landed — rust-server-sdk#168 and eventsource-client#134/#135 — and 3.1.1 resolves `eventsource-client` to 0.17.5, which carries them.

- Uses the rustls + aws-lc-rs features (`hyper-rustls-native-roots`, `crypto-aws-lc-rs`), avoiding the OpenSSL path.
- Because LD now builds a rustls client and the workspace links both rustls provider features (`aws_lc_rs` + `ring`, the latter transitively), each binary that builds an LD client installs the aws-lc-rs provider explicitly (environmentd, clusterd, balancerd, sqllogictest, testdrive) or it panics on first client build. orchestratord already installs it.
- Adds `MetricsTransport` unit tests, including `test_metric_frozen_on_midstream_error` modeling the incident-984 failure mode.

Rebased onto current `main`. Draft until the nightly LaunchDarkly jobs are confirmed green across the stack.

### DIFF
--- a/src/balancerd/src/bin/balancerd.rs
+++ b/src/balancerd/src/bin/balancerd.rs
@@ -172,6 +172,14 @@ pub struct ServiceArgs {
 fn main() {
     let args: Args = cli::parse_args(CliConfig::default());
 
+    // Pin the rustls crypto provider to aws-lc-rs. The LaunchDarkly SDK uses
+    // hyper-rustls, so building its client resolves the process-default rustls
+    // provider. The workspace also links rustls' `ring` feature (pulled by
+    // other hyper-rustls chains), and with both provider features enabled
+    // rustls cannot choose a default on its own and panics. The call is
+    // idempotent, so ignore the result.
+    let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
+
     // Mirror the tokio Runtime configuration in our production binaries.
     let ncpus_useful = usize::max(1, std::cmp::min(num_cpus::get(), num_cpus::get_physical()));
     let runtime = tokio::runtime::Builder::new_multi_thread()
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -31,7 +31,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
@@ -1940,6 +1940,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
 
 [[package]]
+name = "chacha20"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
+dependencies = [
+ "cfg-if",
+ "cpufeatures 0.3.0",
+ "rand_core 0.10.1",
+]
+
+[[package]]
 name = "chrono"
 version = "0.4.41"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2005,6 +2016,12 @@ dependencies = [
  "ciborium-io",
  "half 1.6.0",
 ]
+
+[[package]]
+name = "cidr"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "579504560394e388085d0c080ea587dfa5c15f7e251b4d5247d1e1a61d1d6928"
 
 [[package]]
 name = "cipher"
@@ -2345,6 +2362,15 @@ name = "cpufeatures"
 version = "0.2.17"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "59ed5838eebb26a2bb2e58f6d5b5316989ae9d08bab10e0e6d103e656d1b0280"
+dependencies = [
+ "libc",
+]
+
+[[package]]
+name = "cpufeatures"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b2a41393f66f16b0823bb79094d54ac5fbd34ab292ddafb9a0456ac9f87d201"
 dependencies = [
  "libc",
 ]
@@ -3028,7 +3054,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users",
- "windows-sys 0.61.1",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -3442,17 +3468,18 @@ dependencies = [
 
 [[package]]
 name = "eventsource-client"
-version = "0.16.0"
+version = "0.17.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c26451361cde19fe2322835b6e684f4825902edf3139ce9d6a70e6542566a0b2"
+checksum = "96df8cfa11d3c8e4e1a48b81c9c600ecbe8c11d1326f41e1524cc4d42f7bf6d9"
 dependencies = [
  "base64 0.22.1",
+ "bytes",
  "futures",
- "hyper 0.14.32",
- "hyper-timeout 0.4.1",
+ "http 1.4.2",
+ "launchdarkly-sdk-transport",
  "log",
  "pin-project",
- "rand 0.8.5",
+ "rand 0.10.2",
  "tokio",
 ]
 
@@ -3997,9 +4024,21 @@ dependencies = [
  "cfg-if",
  "js-sys",
  "libc",
- "r-efi",
+ "r-efi 5.3.0",
  "wasi 0.14.2+wasi-0.2.4",
  "wasm-bindgen",
+]
+
+[[package]]
+name = "getrandom"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "300e883d756b2e4ec94e02791f39b04b522276138852cfc41d9fb7e904106099"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "r-efi 6.0.0",
+ "rand_core 0.10.1",
 ]
 
 [[package]]
@@ -4425,6 +4464,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "hyper-http-proxy"
+version = "1.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8021e0ae20c08eadc94d0bdafdeda66d4f0858541c146ae6e46b219bfe58497e"
+dependencies = [
+ "bytes",
+ "futures-util",
+ "headers",
+ "http 1.4.2",
+ "hyper 1.9.0",
+ "hyper-rustls",
+ "hyper-util",
+ "pin-project-lite",
+ "tokio",
+ "tokio-rustls",
+ "tower-service",
+]
+
+[[package]]
 name = "hyper-openssl"
 version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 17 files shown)


src/balancerd/src/bin/balancerd.rs
```diff
@@ -170,6 +170,10 @@ pub struct ServiceArgs {
 }
 
 fn main() {
+    // Both the aws-lc-rs and ring rustls backends are linked, so rustls can't
+    // auto-select a provider and panics on first TLS use unless one is installed.
+    let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
```
alex-hunt-materialize: Do we know what is pulling in ring? Can we remove it?

It seems to be pulled in by rustls, but that's probably only because we have some feature enabled by something else.

alex-hunt-materialize (approved):

def- (commented):
QA LLM review notes:

## MEDIUM — SSE staleness metric updates on connection establishment (200 head), not just on received data, which can suppress the `launchdarkly-stale-sse-tier-2` alert this migration exists to enable

**File:** `src/adapter/src/config/frontend.rs`, `MetricsTransport::request` (the
`gauge.set(now_fn() / 1000)` call *before* the body is wrapped).

### What the code does

```rust
Box::pin(async move {
    let resp = inner_fut.await?;
    if resp.status().is_success() {
        gauge.set(now_fn() / 1000);          // (A) fires on the 200 response HEAD
        let (parts, body) = resp.into_parts();
        let wrapped: ByteStream = Box::pin(body.inspect_ok(move |_| {
            gauge.set(now_fn() / 1000);      // (B) fires on each received body chunk
        }));
        Ok(http::Response::from_parts(parts, wrapped))
    } else {
        Ok(resp)
    }
})
```

The same `MetricsTransport` type is used for both the event processor (CSE metric,
`last_cse_time_seconds`) and the streaming data source (SSE metric,
`last_sse_time_seconds`). For the CSE (request/response) path, update (A) on the
200 head is correct. For the **streaming/SSE data source**, update (A) is the
problem.

### Why it is a defect

The metric's documented contract (`doc/user/data/metrics.yml:1339`,
`src/adapter/src/config.rs:161`) is:

> "The last known time when the LaunchDarkly client **received an event** from
> the LaunchDarkly server."

The pre-migration code honored that exactly: `last_sse_time_seconds` was set only
from the per-event callback (`frontend.rs:383-390`, `move |_ev| { ... set(ts) }`),
i.e. only when an SSE event was actually *received*. The new code additionally
sets the gauge whenever a streaming **connection is established** (200 head),
which is not "received an event."

This matters because the whole reason for this migration is that the upstream SDK
now **reconnects** on non-`Eof` stream errors (the incident-984 fix). Consider the
persistent version of the incident-984 failure mode — LD returns `200 OK`, then
the body times out before (or without) delivering usable data
(`hyper::Error(Body, Kind(TimedOut))`), repeatedly:

1. Reconnect → `request()` called again → 200 head → **update (A) sets the gauge to `now`**.
2. Body yields no chunk (times out after the configured `read_timeout` of 300s) → update (B) never fires.
3. SDK reconnects (backoff) → back to step 1 → gauge set to `now` again.

So a data source that is receiving **zero** actual events keeps its
`last_sse_time_seconds` advancing roughly every `read_timeout + reconnect_backoff`
(~300s+). The old code would have frozen the gauge at the last real event, letting
the staleness alert fire. The new head-update re-arms the "freshness clock" on
every reconnect, so a persistently wedged/reconnecting data source can look fresh
and the `launchdarkly-stale-sse-tier-2` alert — the detection this PR is built to
preserve — may never fire.

Note that update (A) is also *unnecessary* for the healthy SSE case: on a good
connection LD immediately sends the `put` (full flag state) as the first body
chunk, and sends periodic heartbeat bytes thereafter, so update (B) already keeps
the gauge fresh. Update (A) adds nothing on the happy path and only injects
spurious freshness on the failure path.

The PR's new test `test_metric_frozen_on_midstream_error` does not catch this: it
exercises a **single** connection (200 → one event → error) and asserts the gauge
freezes, but never models the reconnect that follows in production, which is where
the head-update re-freshens the gauge.

### Reachability / severity

- Reachable via a realistic operational scenario: a persistent LD-edge/proxy/network
  condition that returns 200 then stalls the body — exactly incident-984, made
  recurring rather than one-shot.
- Full alert suppression depends on the external `stale-sse-tier-2` threshold being
  longer than the ~300s `read_timeout` + reconnect backoff cycle. Tier-2 (the
  less-urgent tier) plausibly uses a multi-minute threshold, in which case the alert
  is suppressed; with a shorter threshold the alert firing is merely delayed by up
  to one reconnect cycle on each reconnect.
- Impact is confined to observability/alerting (a stuck config sync goes undetected),
  not data loss, crash, or security — hence MEDIUM, not HIGH.

### Suggested fix

Do not update the gauge on the response head for the **streaming (SSE) transport**;
update it only from received body chunks (update B). Keep update (A) for the
**event-processor (CSE) transport**, where the response is a normal
request/response and there is no meaningful streamed body. Concretely, give
`MetricsTransport` a flag such as `update_on_response_head: bool`, set `true` for
`cse_transport` and `false` for `data_source_transport`:

```rust
struct MetricsTransport<T> {
    inner: T,
    last_success_gauge: UIntGauge,
    now_fn: NowFn,
    update_on_response_head: bool,
}

// in request():
if resp.status().is_success() {
    if self.update_on_response_head {
        gauge.set(now_fn() / 1000);
    }
    let (parts, body) = resp.into_parts();
    let wrapped: ByteStream = Box::pin(body.inspect_ok(move |_| gauge.set(now_fn() / 1000)));
    Ok(http::Response::from_parts(parts, wrapped))
} else {
    Ok(resp)
}
```

This restores the documented "last received an event" semantics for the SSE metric
while preserving the CSE metric behavior, and keeps the staleness alert able to
detect a wedged data source.

(I haven't checked this yet, but thought I'd post it before we merge this too quickly.)

jasonhernandez:
> QA LLM review notes:
> ## MEDIUM — SSE staleness metric updates on connection establishment (200 head), not just on received data, which can suppress the `launchdarkly-stale-sse-tier-2` alert this migration exists to enable
> 
> **File:** `src/adapter/src/config/frontend.rs`, `MetricsTransport::request` (the `gauge.set(now_fn() / 1000)` call _before_ the body is wrapped).
> ### What the code does
> 
> ```rust
> Box::pin(async move {
>     let resp = inner_fut.await?;
>     if resp.status().is_success() {
>         gauge.set(now_fn() / 1000);          // (A) fires on the 200 response HEAD
>         let (parts, body) = resp.into_parts();
>         let wrapped: ByteStream = Box::pin(body.inspect_ok(move |_| {
>             gauge.set(now_fn() / 1000);      // (B) fires on each received body chunk
>         }));
>         Ok(http::Response::from_parts(parts, wrapped))
>     } else {
>         Ok(resp)
>     }
> })
> ```
> 
> The same `MetricsTransport` type is used for both the event processor (CSE metric, `last_cse_time_seconds`) and the streaming data source (SSE metric, `last_sse_time_seconds`). For the CSE (request/response) path, update (A) on the 200 head is correct. For the **streaming/SSE data source**, update (A) is the problem.
> ### Why it is a defect
> 
> The metric's documented contract (`doc/user/data/metrics.yml:1339`, `src/adapter/src/config.rs:161`) is:
> 
> > "The last known time when the LaunchDarkly client **received an event** from
> > the LaunchDarkly server."
> 
> The pre-migration code honored that exactly: `last_sse_time_seconds` was set only from the per-event callback (`frontend.rs:383-390`, `move |_ev| { ... set(ts) }`), i.e. only when an SSE event was actually _received_. The new code additionally sets the gauge whenever a streaming **connection is established** (200 head), which is not "received an event."
> 
> This matters because the whole reason for this migration is that the upstream SDK now **reconnects** on non-`Eof` stream errors (the incident-984 fix). Consider the persistent version of the incident-984 failure mode — LD returns `200 OK`, then the body times out before (or without) delivering usable data (`hyper::Error(Body, Kind(TimedOut))`), repeatedly:
> 
>     1. Reconnect → `request()` called again → 200 head → **update (A) sets the gauge to `now`**.
> 
>     2. Body yields no chunk (times out after the configured `read_timeout` of 300s) → update (B) never fires.
> 
>     3. SDK reconnects (backoff) → back to step 1 → gauge set to `now` again.
> 
> 
> So a data source that is receiving **zero** actual events keeps its `last_sse_time_seconds` advancing roughly every `read_timeout + reconnect_backoff` (~300s+). The old code would have frozen the gauge at the last real event, letting the staleness alert fire. The new head-update re-arms the "freshness clock" on every reconnect, so a persistently wedged/reconnecting data source can look fresh and the `launchdarkly-stale-sse-tier-2` alert — the detection this PR is built to preserve — may never fire.
> 
> Note that update (A) is also _unnecessary_ for the healthy SSE case: on a good connection LD immediately sends the `put` (full flag state) as the first body chunk, and sends periodic heartbeat bytes thereafter, so update (B) already keeps the gauge fresh. Update (A) adds nothing on the happy path and only injects spurious freshness on the failure path.
> 
> The PR's new test `test_metric_frozen_on_midstream_error` does not catch this: it exercises a **single** connection (200 → one event → error) and asserts the gauge freezes, but never models the reconnect that follows in production, which is where the head-update re-freshens the gauge.
> ### Reachability / severity
> 
>     * Reachable via a realistic operational scenario: a persistent LD-edge/proxy/network
>       condition that returns 200 then stalls the body — exactly incident-984, made
>       recurring rather than one-shot.
> 
>     * Full alert suppression depends on the external `stale-sse-tier-2` threshold being
>       longer than the ~300s `read_timeout` + reconnect backoff cycle. Tier-2 (the
>       less-urgent tier) plausibly uses a multi-minute threshold, in which case the alert
>       is suppressed; with a shorter threshold the alert firing is merely delayed by up
>       to one reconnect cycle on each reconnect.
> 
>     * Impact is confined to observability/alerting (a stuck config sync goes undetected),
>       not data loss, crash, or security — hence MEDIUM, not HIGH.
> 
> 
> ### Suggested fix
> 
> Do not update the gauge on the response head for the **streaming (SSE) transport**; update it only from received body chunks (update B). Keep update (A) for the **event-processor (CSE) transport**, where the response is a normal request/response and there is no meaningful streamed body. Concretely, give `MetricsTransport` a flag such as `update_on_response_head: bool`, set `true` for `cse_transport` and `false` for `data_source_transport`:
> 
> ```rust
> struct MetricsTransport<T> {
>     inner: T,
>     last_success_gauge: UIntGauge,
>     now_fn: NowFn,
>     update_on_response_head: bool,
> }
> 
> // in request():
> if resp.status().is_success() {
>     if self.update_on_response_head {
>         gauge.set(now_fn() / 1000);
>     }
>     let (parts, body) = resp.into_parts();
>     let wrapped: ByteStream = Box::pin(body.inspect_ok(move |_| gauge.set(now_fn() / 1000)));
>     Ok(http::Response::from_parts(parts, wrapped))
> } else {
>     Ok(resp)
> }
> ```
> 
> This restores the documented "last received an event" semantics for the SSE metric while preserving the CSE metric behavior, and keeps the staleness alert able to detect a wedged data source.
> 
> (I haven't checked this yet, but thought I'd post it before we merge this too quickly.)

thanks - I'll take it into account and likely patch. Please look at #37026 as well because that includes some more testing that I used to de-risk this upgrade

PR #37027, opened 2026-06-13
ci: Nightly fixes (2026-06-13)

def-:
Based on https://buildkite.com/materialize/nightly/builds/16780


test/launchdarkly-flag-consistency/mzcompose.py
```diff
@@ -248,6 +248,7 @@
     enable_load_generator_counter
     enable_load_generator_datums
     enable_managed_cluster_availability_zones
+    enable_mcp_developer_query_tool
```
def-, line 251: @bobbyiliev We should add this to LaunchDarkly I assume

antiguru (approved):

PR #37037, opened 2026-06-15
adapter: bound COPY FROM STDIN blocking-pool usage

def-:
setup_copy_from_stdin spawned available_parallelism() workers on the shared tokio blocking pool eagerly at COPY start -- before any data -- and each worker held its blocking thread for the whole COPY by driving an idle batch_rx.recv().await via block_on. With no cap on worker count or concurrent COPYs, a low-privilege client could open ~512/cores idle, zero-byte COPY ... FROM STDIN connections and pin the entire 512-thread pool, stalling every other blocking-pool user (e.g. the mandatory "optimize peek" stage of any SELECT) process-wide until the client disconnected.

Run each worker as a regular async task instead, so it holds no thread while parked waiting for the next chunk; offload only the CPU-bound per-chunk decode to the blocking pool for the duration of that decode. Also cap workers per COPY at COPY_FROM_STDIN_MAX_WORKERS (8) so one COPY cannot reserve an unbounded share of the pool even while actively decoding.

Closes SQL-372.

### DIFF
--- a/test/copy/mzcompose.py
+++ b/test/copy/mzcompose.py
@@ -17,6 +17,7 @@ import csv
 import json
 import random
 import string
+import threading
 import time
 from io import BytesIO, StringIO
 from textwrap import dedent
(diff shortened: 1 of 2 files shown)


test/copy/mzcompose.py
```diff
@@ -700,3 +701,79 @@ def workflow_copy_from_csv_crlf_large_end_marker(c: Composition) -> None:
...
+    held: list[tuple] = []
+    try:
+        for _ in range(_NUM_IDLE_SESSIONS):
+            held.append(_open_idle_copy(c))
+        assert _select_1_responsive(c, _SELECT_TIMEOUT_S), (
+            f"SELECT 1 did not return within {_SELECT_TIMEOUT_S}s while "
+            f"{len(held)} idle COPY FROM STDIN sessions were open"
+        )
+    finally:
+        _close_idle_copies(held)
```
ggevay, line 794: From Claude:

The new regression test never runs in CI. The copy pipeline step uses run: ci (ci/test/pipeline.template.yml:665), which invokes workflow_ci, and that lists only ["auth", "http", "copy-from-csv-header", "copy-from-ssrf-redirect"] (test/copy/mzcompose.py:207). workflow_copy_from_stdin_many_idle_sessions is reachable only through workflow_default, which no pipeline step runs for this composition (nightly uses run: nightly). So the test — the PR's proof and its guard against silent reintroduction — provides zero CI protection. One-line fix: add "copy-from-stdin-many-idle-sessions" to the workflow_ci list.

(The four copy_from_csv_* workflows added by #36732 are unwired the same way.)

test/copy/mzcompose.py
```diff
@@ -700,3 +701,79 @@ def workflow_copy_from_csv_crlf_large_end_marker(c: Composition) -> None:
             f"expected count={rows_each_side}, max_id={rows_each_side - 1} "
             "(rows after the bare \\. leaked through parallel workers)"
         )
+
+
+_NUM_IDLE_SESSIONS = 128
```
ggevay: Claude:

The reproduction is zero-margin and silently machine-coupled. Once wired in, the test runs on hetzner-aarch64-4cpu-8gb, so the buggy code spawns available_parallelism() = 4 workers per COPY, and _NUM_IDLE_SESSIONS = 128 gives exactly 128 × 4 = 512 pinned threads — precisely the pool size, making SELECT 1 the starved 513th request. That works, but 128 is implicitly "512 / 4 cores" with no headroom: on any agent reporting fewer effective cores (smaller queue, or cgroup CPU limits), the buggy code wouldn't exhaust the pool and the test would false-pass, silently stopping guarding the fix. Recommend (a) a comment documenting the 128 ↔ pool(512) ↔ cores coupling, and (b) bumping _NUM_IDLE_SESSIONS for margin — max_connections defaults to 5000 so there's ample connection headroom (keep it moderate, e.g. ~256

ggevay (approved):
Thank you!

PR #37039, opened 2026-06-15
sql: Gate all paths that set the bounded staleness isolation level

antiguru:
### Motivation

Fixes CLU-115.

`enable_bounded_staleness_isolation` was meant to gate the `bounded staleness <duration>` isolation level, but only `SET transaction_isolation TO ...` was blocked, and that check compared the variable name case-sensitively.
The flag could be bypassed three ways: a quoted/upper-cased name (`SET "TRANSACTION_ISOLATION" TO ...`), `ALTER ROLE ... SET transaction_isolation = ...`, and passing `transaction_isolation` as a connection option.

### Description

The gating logic now lives in a single case-insensitive helper, `check_transaction_isolation_feature_flag`, called from the chokepoint `SessionVars::set` (which all of `SET`, `SET TRANSACTION`, and connection options route through) and at plan time for both `SET` and `ALTER ROLE ... SET`.
A role default that names a gated level whose flag is later disabled is dropped when building session defaults, so a fresh session falls back to the built-in default rather than silently using the gated level.
The same helper also gates `strong session serializable` behind `enable_session_timelines`, closing the equivalent bypasses for that level.

### Verification

Added coverage in `test/sqllogictest/bounded_staleness.slt` for the quoted-name `SET`, `ALTER ROLE ... SET`, and the role-default fallback when the flag is disabled, plus a unit test in `src/sql/src/session/vars.rs` exercising the helper directly (flag on/off, name case-insensitivity, non-gated levels, unrelated variables).

### DIFF
--- a/src/adapter/src/coord/command_handler.rs
+++ b/src/adapter/src/coord/command_handler.rs
@@ -52,7 +52,8 @@ use mz_sql::rbac;
 use mz_sql::rbac::CREATE_ITEM_USAGE;
 use mz_sql::session::user::User;
 use mz_sql::session::vars::{
-    EndTransactionAction, NETWORK_POLICY, OwnedVarInput, STATEMENT_LOGGING_SAMPLE_RATE, Value, Var,
+    EndTransactionAction, NETWORK_POLICY, OwnedVarInput, STATEMENT_LOGGING_SAMPLE_RATE,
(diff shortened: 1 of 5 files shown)


def- (approved):
Thanks for adding tests, no complaints from QA side

mtabebe (approved):
Thanks for the fix, and the validating test.

antiguru:
Thanks for the reviews!

PR #37040, opened 2026-06-15
adapter: Drop comments when cascade-dropping replica-targeted MVs

antiguru:
### Motivation

Resizing a managed cluster that hosts a replica-targeted materialized view with a `COMMENT` panicked the coordinator:

```
coordinator inconsistency detected
  left: Err(... comments: [Dangling(MaterializedView(User(2)))] ...)
```

Fixes CLU-118.

### Description

Resizing a managed cluster tears down its replica via an internally-built `Op::DropObjects` that names only the replica, not the replica-targeted materialized views it implicitly drops.
The `ClusterReplica` branch of `ObjectsToDrop::add_item` expanded those MVs and their dependents into `self.items` but never recorded their comments in `self.comments`, so the MV was dropped while its comment survived.
Plan-driven drops (e.g. `DROP CLUSTER`) avoid this because the plan stage enumerates each dependent as its own `DropObjectInfo::Item`, whose comment is recorded by the top-of-function insert; only the direct-op resize path learns of the MV solely through this cascade.

The branch is rewritten to reuse `cluster_replica_dependents` — removing the duplicated `bound_objects`/`target_replica` loop — and to insert each dropped item's comment id alongside the item.
`seen` is still seeded from the items already collected, so the plan-driven path (dependents processed before the replica, in reverse-dependency order) does not re-add them.

### Verification

Extended `test/testdrive/materialized-view-replica-targeted.td` with a CLU-118 regression: a managed cluster with a commented replica-targeted MV is resized, then asserts the MV is gone, no dangling comment remains, and the cluster is still usable.

### DIFF
--- a/test/testdrive/materialized-view-replica-targeted.td
+++ b/test/testdrive/materialized-view-replica-targeted.td
@@ -193,9 +193,44 @@ contains:unknown catalog item 'mv_r1'
 r2 true
 
 # ---------------------------------------------------------------------------
+# Regression test for CLU-118: resizing a managed cluster that hosts a
+# replica-targeted MV with a COMMENT must not leave a dangling comment behind
+# (which previously tripped the coordinator consistency check and panicked).
+# ---------------------------------------------------------------------------
+
+> CREATE CLUSTER managed (SIZE 'scale=1,workers=1', REPLICATION FACTOR 1)
+
+> CREATE MATERIALIZED VIEW mv_managed IN CLUSTER managed REPLICA r1 AS SELECT sum(a) AS s FROM t
+
+> COMMENT ON MATERIALIZED VIEW mv_managed IS 'pinned to r1'
+
+> SELECT * FROM mv_managed
+10
+
+# Resizing tears down and recreates the replica, which implicitly drops the
+# targeted MV and must also drop its comment.
+> ALTER CLUSTER managed SET (SIZE 'scale=1,workers=2')
+
+# The targeted MV is gone.
+! SELECT * FROM mv_managed
+contains:unknown catalog item 'mv_managed'
+
+# No dangling comment is left behind.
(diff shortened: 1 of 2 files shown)


test/testdrive/materialized-view-replica-targeted.td
```diff
@@ -192,10 +192,45 @@ contains:unknown catalog item 'mv_r1'
...
+# targeted MV and must also drop its comment.
+> ALTER CLUSTER managed SET (SIZE 'scale=1,workers=2')
+
+# The targeted MV is gone.
+! SELECT * FROM mv_managed
+contains:unknown catalog item 'mv_managed'
+
+# No dangling comment is left behind.
+> SELECT count(*) FROM mz_internal.mz_comments c
+  JOIN mz_materialized_views mv ON (mv.id = c.id)
```
ggevay: The new “No dangling comment is left behind” assertion is effectively vacuous for the specific condition it claims to test. Because it joins mz_internal.mz_comments to mz_materialized_views, once mv_managed is dropped the query returns 0 even if an orphaned mz_comments row for the dropped MV still exists. The preceding ALTER CLUSTER likely still reproduces the original coordinator panic, so the regression test has value, but this assertion should directly check the unique comment row, e.g.:
```
> SELECT count(*) FROM mz_internal.mz_comments
  WHERE comment = 'pinned to r1'
0
```
That would make the test actually verify that the comment row was removed.

ggevay (approved):
Thank you!

test/testdrive/materialized-view-replica-targeted.td
antiguru: Good catch, you're right the join was vacuous. Switched to checking `mz_comments` directly by comment text in 550c044. Still passes.

PR #37042, opened 2026-06-15
fix clippy on main

sjwiesman:
## Motivation

`cargo clippy -p mz-deploy` surfaced 7 `clippy::disallowed_methods` warnings. The crate was calling `tokio_postgres` query/execute methods directly, which `clippy.toml` disallows repo-wide in favor of the `mz_postgres_util` `Sql` wrappers — these force SQL through the validated `Sql` type rather than raw strings.

## Changes

- **`src/mz-deploy/Cargo.toml`** — add `mz-postgres-util` as a path dependency (single-line `Cargo.lock` change, no version bumps).
- **`src/client/connection.rs`** — the five pass-through wrapper methods (`execute`, `query_one`, `query`, `simple_query`, `batch_execute`) now delegate to `mz_postgres_util::*`. Every call site passes string SQL (no prepared `Statement`), so the generic `ToStatement` bound is replaced with `&str`, wrapped via `Sql::raw_unchecked` — the sanctioned constructor for internally-assembled, already-trusted SQL.
- **`src/client/deployment_ops.rs`** — the two raw `Transaction` calls (the `SUBSCRIBE` cursor `execute` and the `FETCH ALL` `query`) go through `mz_postgres_util::execute`/`query`; the static `"FETCH ALL c"` uses `Sql::new`.
- **`src/client/errors.rs`** — add `From<mz_postgres_util::PostgresError> for ConnectionError`, mapping the `Postgres` variant back to `ConnectionError::Query` so the existing rich `as_db_error()` formatting is preserved.

## Testing

No behavioral change; this is a lint-compliance refactor that preserves the existing public API and error formatting of the `Client` wrapper. `cargo clippy -p mz-deploy --all-targets -- -D warnings` now passes clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6552,6 +6552,7 @@ dependencies = [
  "mz-controller-types",
  "mz-expr",
  "mz-ore",
+ "mz-postgres-util",
  "mz-repr",
  "mz-secrets",
  "mz-sql",
--- a/src/mz-deploy/Cargo.toml
+++ b/src/mz-deploy/Cargo.toml
@@ -20,6 +20,7 @@ rayon = { workspace = true }
 junit-report = { workspace = true }
 mz-build-info = { path = "../build-info" }
 mz-ore = { path = "../ore", default-features = false, features = ["async"] }
+mz-postgres-util = { path = "../postgres-util" }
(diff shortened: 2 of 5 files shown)


def- (approved):
Thanks Seth!

PR #37045, opened 2026-06-15
[SS-187] Unify max_lsn behind a single function

peterdukelarsen:
### Motivation

This PR is to try to unify the two spots where we call `SELECT pg_current_wal_lsn()`.

This is a small refactor ahead of adding support for replication from a physical standby.

### Description

Unifies all calls to select pg_current_wal_lsn to the method in src/postgres-util/src/replication.rs. This is not a pure refactor as the potential errors returned aren't the same on the off chance nothing is returned.

### DIFF
--- a/src/postgres-util/src/replication.rs
+++ b/src/postgres-util/src/replication.rs
@@ -237,7 +237,11 @@ pub async fn get_timeline_id(client: &Client) -> Result<u64, PostgresError> {
     }
 }
 
-pub async fn get_current_wal_lsn(client: &Client) -> Result<PgLsn, PostgresError> {
+pub async fn fetch_max_lsn(client: &Client) -> Result<PgLsn, PostgresError> {
+    // Based on the documentation, it appears that `pg_current_wal_lsn` has
+    // the same "upper" semantics of `confirmed_flush_lsn`:
+    // <https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-BACKUP>
+    // We may need to revisit this and use `pg_current_wal_flush_lsn`.
     let row = query_one(client, crate::sql!("SELECT pg_current_wal_lsn()"), &[]).await?;
     let lsn: PgLsn = row.get(0);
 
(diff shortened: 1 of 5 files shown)


src/postgres-util/src/replication.rs
```diff
@@ -237,7 +237,11 @@ pub async fn get_timeline_id(client: &Client) -> Result<u64, PostgresError> {
...
 }
 
-pub async fn get_current_wal_lsn(client: &Client) -> Result<PgLsn, PostgresError> {
+pub async fn fetch_max_lsn(client: &Client) -> Result<PgLsn, PostgresError> {
+    // Based on the documentation, it appears that `pg_current_wal_lsn` has
+    // the same "upper" semantics of `confirmed_flush_lsn`:
+    // <https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-BACKUP>
+    // We may need to revisit this and use `pg_current_wal_flush_lsn`.
     let row = query_one(client, crate::sql!("SELECT pg_current_wal_lsn()"), &[]).await?;
     let lsn: PgLsn = row.get(0);
```
peterdukelarsen: Biggest concern I had with this change is that before in the case where there wasn't a row or the row was `None` we explicitly returned a transient error:
```
...TransientError::Generic...
```

Now I think we still (somewhat implicitly) return a transient error through conversion from   `tokio_postgres::Error` to `PostgresError` and then  `TransientError::PostgresError` from `PostgresError`.

I think this code panics on a single Null row so that would have more of a blast radius, but I think this is guaranteed non-null from Postgres.

patrickwwbutler: I feel like in an ideal world, we don't panic on a bizarre null from pg, it is an invariant but it's an external invariant so that feels more like a "we don't know what happened here, the source is bricked" situation rather than "we cannot continue doing anything on this cluster because something is irreparably borked". 

I won't block on it though since this isn't actually changing any behavior, and also it might be a rather large PITA to update handling to specifically be a `DefiniteError` in that case, and `TransientError` for all others.

patrickwwbutler (approved):
LGTM!

with some non-blocking musing on the `.get()` behavior

src/postgres-util/src/replication.rs
peterdukelarsen: That's a good point -- specifically that we don't want unexpected behavior from the upstream to mess up the whole database and would prefer it just impacts the relevant source.

I'm also thinking this is ok to merge given we're already exposed to this issue on RTR queries. It is worth stating that this expands the surface error of the exposure to a potential panic.

peterdukelarsen: Ok, went ahead and fixed the panic risk: https://github.com/MaterializeInc/materialize/pull/37045/changes/d5531484131d08fc181d8c0b045c02b9ae476b4c. Sticking with something that will be a transient error for now.

PR #37048, opened 2026-06-15
storage: Fix CSV source decoder panics on malformed input

def-:
Fuzzing the CSV decoder turned up two ways untrusted source bytes could panic the decode operator (crashing source ingestion), and one silent data-corruption path:

  1. The error paths (wrong column count, invalid UTF-8) returned without resetting output_cursor/ends_cursor, unlike the success path. csv_core resets its own per-record output position, so after a malformed record the next record read back stale, non-monotonic field offsets — slicing output[ends[i]..ends[i+1]] then panicked (start > end), or silently produced rows stitched from leftover buffer bytes. One bad row corrupted the rest of the source. Fixed by clearing the buffers after every completed record, success or error.

  2. The whole record's concatenated field bytes were validated as UTF-8 in one shot, then sliced as a &str at each field boundary. A delimiter can sit between two bytes that, with the delimiter removed, form one multi-byte character: the concatenation validates, but a field boundary lands mid-character, panicking the &str slice (and even when it didn't panic, bytes were silently merged across fields). Fixed by validating each field's bytes as UTF-8 independently; a field that isn't valid UTF-8 on its own is now a decode error.

To make the decoder fuzzable without mz-storage's heavy dependency tree (rocksdb/rdkafka/aws), the pure CsvDecoderState moves to mz-storage-types alongside the CsvEncoding it consumes; mz-storage re-imports it. Adds regression tests for both panics.


def-:
Ready for review

src/storage-types/src/sources/encoding.rs
```diff
@@ -366,3 +368,266 @@ impl ColumnSpec {
...
+                                match std::str::from_utf8(
+                                    &self.output[self.ends[i]..self.ends[i + 1]],
+                                ) {
+                                    Ok(field) => fields.push(field),
+                                    Err(e) => {
+                                        utf8_error = Some(e);
+                                        break;
+                                    }
+                                }
+                            }
```
martykulma: Do we need to allocate the vec on every row here? Seems this could just work with row_packer directly.

martykulma (approved):
lgtm - thanks @def!  One note about fields vec allocation, but nothing blocking.

PR #37049, opened 2026-06-15
expr: Don't propagate an operand's error through non-strict AND/OR in reduce

def-:
The generic variadic fold in `reduce` replaced a call with any operand's literal error unconditionally. That is wrong for a non-strict function: AND/OR have a dominating operand (`false`/`true`) that absorbs an erroring operand at runtime, so `false AND <error>` evaluates to `false`, not an error. Folding `col AND <error>` to the error made `reduce` introduce an error into a query that evaluates successfully — a correctness bug. Guard error-propagation with `func.propagates_nulls()`, mirroring the adjacent null-propagation guard (the other non-strict variadic, Coalesce, already bails out earlier). Found by the mir_scalar_reduce fuzz target.

### DIFF
--- a/src/expr/src/scalar.rs
+++ b/src/expr/src/scalar.rs
@@ -922,6 +922,24 @@ impl MirScalarExpr {
         while old_self != *self {
             old_self = self.clone();
             self.reduce_and_canonicalize_and_or(); // We don't want to deal with 1-arg AND/OR at the top
+
+            // Undistribution (factoring out a common operand, and its
+            // absorption-law special cases like `a OR (a AND c) --> a`)
+            // preserves SQL's three-valued logic but NOT its error semantics:
+            // AND/OR let only the dominating value (`false`/`true`) absorb an
+            // operand's error, while a NULL operand does not — `NULL AND <err>`
+            // and `NULL OR <err>` both evaluate to `<err>`. So a rewrite such as
+            // `a OR (a AND c) --> a` can silently turn a row's error into NULL
+            // when `a` is NULL and a dropped/relocated operand errors. Skip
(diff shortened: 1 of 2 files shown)


def-:
Ready for review

ggevay (approved):
Well, it's not totally clear what should happen here.

The [Postgres docs say](https://www.postgresql.org/docs/9.0/sql-expressions.html#SYNTAX-EXPRESS-EVAL) stuff like `WHERE x > 0 AND y/x > 1.5` is "an untrustworthy way of trying to avoid division by zero" (and instead recommend `CASE`). Fwiw, Postgres' actual behavior _kind of_ agrees with the PR's proposed behavior:
```
postgres=# select false and 1/0 = 5;
 ?column? 
----------
 f
(1 row)

postgres=# select 1/0 = 5 and false;
ERROR:  division by zero
```
But there is weirdness in related situations:
```
create table t(x int);
insert into t values (3),(0);

postgres=# select 7/x from t;
ERROR:  division by zero
postgres=# select 7/x = 5 and false from t;
 ?column? 
----------
 f
 f
(2 rows)

postgres=# select false and 7/x = 5 from t;
 ?column? 
----------
 f
 f
(2 rows)
```

If our current behavior is frustrating some automated tests, I guess we can merge the PR. There are many other situations where we nope out of optimizations when we see a literal error being involved, so one more of these probably won't break stuff, but it's hard to know for sure. Sorry that I can't be more decisive here!

(In any case, https://github.com/MaterializeInc/database-issues/issues/4972 lists a gazillion other things where our error behavior is buggy or slightly off or unclear.)

PR #37050, opened 2026-06-15
[catalog] convert mz_default_privileges/mz_system_privileges to BuiltinMaterializedView

mtabebe:
Problem:
mz_catalog.mz_default_privileges and mz_catalog.mz_system_privileges are
BuiltinTables. This requires coordinated writes from a single
environmentd process, conflicting with the multi-envd goal.

Solution:
Convert both to BuiltinMaterializedViews backed by queries over
mz_internal.mz_catalog_raw.

To express the acl_mode -> char-code mapping in SQL, this commit adds a
new scalar function mz_internal.parse_catalog_acl_mode(jsonb) -> text,
which mirrors AclMode::from_bits + Display.

The populator functions pack_default_privileges_update and
pack_system_privileges_update are deleted from builtin_table_updates,
and the corresponding StateUpdateKind arms in apply.rs short-circuit to
Vec::new().

Testing:
New behavioural SLTs for mz_default_privileges and mz_system_privileges


mtabebe:
The test failures are failures on main too in nightly (2 caused by me .. I'm working on a fix)

src/catalog/src/builtin/mz_catalog.rs
```diff
@@ -2653,109 +2653,175 @@ pub static MZ_CLUSTER_REPLICA_FRONTIERS_IND: LazyLock<BuiltinIndex> =
...
+        WHEN '7'  THEN 'type'
+        WHEN '8'  THEN 'role'
+        WHEN '9'  THEN 'cluster'
+        WHEN '10' THEN 'cluster replica'
+        WHEN '11' THEN 'secret'
+        WHEN '12' THEN 'connection'
+        WHEN '13' THEN 'database'
+        WHEN '14' THEN 'schema'
+        WHEN '15' THEN 'function'
+        WHEN '17' THEN 'network policy'
```
ggevay, line 2730: We should definitely add at least code comments in the places where these cases and numbers come from. Plus it would be nice to have a test that this doesn't drift, although I'm not sure how painful would writing such a test be.

src/expr/src/scalar/func/impls/jsonb.rs
```diff
@@ -359,6 +359,30 @@ fn parse_catalog_privileges<'a>(a: JsonbRef<'a>) -> Result<ArrayRustType<MzAclIt
         .map_err(|e| EvalError::InvalidCatalogJson(e.into()))
 }
 
+/// Converts a catalog JSON-serialized `AclMode` bitflags object into a
+/// PostgreSQL ACL char-code string (e.g. `{"bitflags": 514}` → `"ar"`).
+#[sqlfunc]
+fn parse_catalog_acl_mode<'a>(a: JsonbRef<'a>) -> Result<String, EvalError> {
```
ggevay, line 374: nit: this duplicates the `bitflags` → `AclMode::from_bits` decode already inlined in `parse_catalog_privileges`'s `acl_mode` arm. Could extract a shared `fn jsonb_to_acl_mode(...) -> Result<AclMode, String>` so they can't drift — optional, since they return different types.

test/sqllogictest/mz_default_privileges.slt
```diff
@@ -0,0 +1,254 @@
...
+WHERE role_id = (SELECT id FROM mz_roles WHERE name = 'r_scope')
+----
+false  false  r
+false  true  r
+true  true  a
+
+# --- PUBLIC grantee -----------------------------------------------------------
+
+# grantee='p' indicates the PUBLIC pseudo-role. Most grants in this file are
+# TO PUBLIC, plus the four bootstrap rows.
```
ggevay: nit: only one of the four bootstrap rows has `grantee='p'` (the `type` row); the other three grant to `s2`. The `14` is correct (1 + 7 + 3 + 3), but this reads as if all four count. Maybe "…plus the single PUBLIC-granted bootstrap row (type)."

src/catalog/src/builtin/mz_catalog.rs
```diff
@@ -2653,109 +2653,175 @@ pub static MZ_CLUSTER_REPLICA_FRONTIERS_IND: LazyLock<BuiltinIndex> =
...
+                "grantee",
+                "Privileges described in this row will be granted to `grantee`. The role ID `p` stands for the `PUBLIC` pseudo-role and applies to all roles.",
+            ),
+            ("privileges", "The set of privileges that will be granted."),
+        ]),
+        // `object_type` in `mz_catalog_raw` is the numeric `Serialize_repr` form
+        // of `proto::ObjectType`. The CASE mirrors `mz_sql::catalog::ObjectType::
+        // Display` lowercased, matching the prior `pack_default_privileges_update`
+        // populator. Note that variant `16` is reserved/unused in the proto enum
+        // (see `src/catalog-protos/src/objects.rs:1995-2013`); any new variant
```
ggevay: nit: the `objects.rs:1995-2013` line reference will rot. Since you already name `proto::ObjectType`, the line range can be dropped.

ggevay (approved):
LGTM, just some minor comments.

PR #37051, opened 2026-06-15
Refactor AvroSchemaResolver: WriterSchemas enum + WriterSchemaKey

martykulma:
Generalize the handling of writer schemas and their keys.

When MZ reads a message from Kafka, it will parse the wire format header to determine what schema the message was written with. It may have been the reader schema, but it may be a different schema (the writer schema) that MZ will need to dynamically look up.

### DIFF
--- a/src/interchange/src/avro/schema.rs
+++ b/src/interchange/src/avro/schema.rs
@@ -53,6 +53,7 @@ use mz_repr::adt::numeric::{NUMERIC_DATUM_MAX_PRECISION, NumericMaxScale};
 use mz_repr::adt::timestamp::TimestampPrecision;
 use mz_repr::{ColumnName, RelationDesc, SqlColumnType, SqlScalarType, UNKNOWN_COLUMN_NAME};
 use tracing::warn;
+use uuid::Uuid;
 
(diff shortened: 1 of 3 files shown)


src/interchange/src/avro/schema.rs
```diff
@@ -360,106 +361,153 @@ mod tests {
...
+}
+
+/// Source of writer schemas for an [`AvroSchemaResolver`].
+///
+/// Mirrors the `WireFormat<C>` enum on the catalog side: a source can run
+/// without any wire-format framing or with Confluent framing (optionally
+/// without a registry to fetch from). Each variant owns its cache type by
+/// construction, so the resolver cannot mis-route a key to the wrong
+/// cache.
+pub enum WriterSchemas {
```
patrickwwbutler: Just at first glance the name `WriterSchemas` feels a bit ambiguous to me, because it kind of implies that the schemas themselves literally exist in this object (which I guess they _kind of_ do when there's a cache?), when it seems like it's more generally a way to get the schemas?

patrickwwbutler (approved):
LGTM with an opinion comment on the class name of `WriterSchemas`!

PR #37053, opened 2026-06-15
[test] normalize quickstart REPLICATION FACTOR in mz_clusters/mz_cluster_replicas SLTs

mtabebe:
New behavioural tests for mz_catalog.mz_clusters and mz_catalog.mz_cluster_replicas assert that the default quickstart cluster has replication_factor=1 and exactly one replica row.

Nightly's slow-tests workflow runs sqllogictest with --replicas=2, which bootstraps quickstart at replication_factor=2 and breaks these assertions.

Normalize quickstart to REPLICATION FACTOR 1 right after reset-server so the assertions are deterministic regardless of mzcompose's --replicas override.

### DIFF
--- a/test/sqllogictest/mz_cluster_replicas.slt
+++ b/test/sqllogictest/mz_cluster_replicas.slt
@@ -13,6 +13,14 @@ mode cockroach
 
 reset-server
 
+# Normalize quickstart's REPLICATION FACTOR so assertions about its replicas
+# pass regardless of mzcompose --replicas overrides (nightly runs slow-tests
+# with --replicas=2).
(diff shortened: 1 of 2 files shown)


mtabebe:
@def- this will fix a nightly blocker

def- (changes requested):
You can rename them to singlereplica_....slt instead, we have a convention to name SLTs requiring a single replica so.

def-:
Actually I just did that in https://github.com/MaterializeInc/materialize/pull/37059

mtabebe:
Today I learned. Thanks

PR #37054, opened 2026-06-15
Annotate 'public' self managed metrics

SangJunBak:
See commit messages for details. Currently hidden for now. The 'public' list is derived from https://materializeinc.slack.com/archives/C0A8HUDLZHQ/p1781035892355479?thread_ts=1781034098.598749&cid=C0A8HUDLZHQ

Preview link: https://preview.materialize.com/materialize/37054/manage/monitor/essential-metrics/

### DIFF
--- a/doc/user/content/manage/monitor/essential-metrics.md
+++ b/doc/user/content/manage/monitor/essential-metrics.md
@@ -1,0 +1,24 @@
+---
+title: "Essential metrics"
+description: "The Prometheus metrics Materialize recommends building dashboards and alerts on."
+disable_toc: true
+disable_list: true
+# TODO (SangJunBak): Unhide once ready to publish
+# menu:
+#   main:
+#     parent: "monitor"
+#     identifier: "monitor-essential-metrics"
+#     weight: 19
+---
+
+This page lists the essential Prometheus metrics exposed by Materialize: the
+ones we recommend building dashboards and alerts on. This list may evolve as
+we add observability for new features and refine what's most useful. A `*` in
+a metric name denotes a family of metrics whose name is completed at runtime
+(for example, `mz_persist_*_bytes`).
+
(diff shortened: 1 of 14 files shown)


src/metrics-catalog/src/main.rs
```diff
@@ -32,6 +33,23 @@ use walkdir::WalkDir;
...
+    };
+    match path
+        .path
+        .segments
+        .last()
+        .map(|s| s.ident.to_string())
+        .as_deref()
+    {
+        Some("Public") => MetricVisibility::Public,
+        _ => MetricVisibility::Internal,
```
mtabebe, line 49: 👍

doc/user/content/manage/monitor/essential-metrics.md
```diff
@@ -0,0 +1,23 @@
...
+# TODO (SangJunBak): Unhide once ready to publish
+# menu:
+#   main:
+#     parent: "monitor"
+#     identifier: "monitor-essential-metrics"
+#     weight: 19
+---
+
+This page lists the essential Prometheus metrics exposed by Materialize: the
+ones we recommend building dashboards and alerts on. A `*` in a metric name
```
mtabebe: In the metrics.rs we say 

```
/// A metric we want customers to build dashboards and
/// alerts on. We do not guarantee stability for this group
```

I don't know if we want to have the same phrasing in the public docs but probably worth a mention that metrics may change... maybe something like

"This list may evolve as we add observability for new features and refine what's most useful"

mtabebe (approved):

doc/user/content/manage/monitor/essential-metrics.md
SangJunBak: That sounds good to me!

PR #37057, opened 2026-06-16
sql: Fix panic on duplicate positional GROUP BY keys

aljoscha:
### Motivation

A query that repeats a positional `GROUP BY` reference to an input column, such as `SELECT *, count(*) FROM t GROUP BY 1, 1`, panicked the planner on the `group_hir_exprs.len() == group_exprs.len()` assertion in `plan_select_from_where`.

Fixes SQL-238

### Description

The GROUP BY loop in `plan_select_from_where` deduplicates keys that map to the same HIR expression, but the dedup check was gated on the AST expression being present (`Some`). `plan_group_by_expr` returns `None` for a positional reference to a bare input column, so a duplicate positional key (e.g. the second `1` in `GROUP BY 1, 1`, or the `1` in `GROUP BY a, 1`) skipped the dedup path: the HIR expr was pushed onto the `group_hir_exprs` Vec twice while the `group_exprs` map deduplicated it, tripping the length assertion.

The fix runs the dedup lookup unconditionally and only records the AST expression on the existing scope item when one is actually present (i.e. for named expressions, not bare positional references).

### Verification

Added regression tests to `test/sqllogictest/select_all_group_by.slt` covering `GROUP BY 1, 1`, `GROUP BY a, 1`, `GROUP BY 1, a`, and a multi-row `GROUP BY 1, 1`. The first three panicked before this change.

### Release notes

This release fixes a panic when a `GROUP BY` clause repeated a positional column reference, such as `GROUP BY 1, 1`.

### DIFF
--- a/src/sql/src/plan/query.rs
+++ b/src/sql/src/plan/query.rs
@@ -2382,14 +2382,24 @@ fn plan_select_from_where(
             let (group_expr, expr) = plan_group_by_expr(ecx, group_expr, &projection)?;
             let new_column = group_key.len();
 
-            if let Some(group_expr) = group_expr {
-                // Multiple AST expressions can map to the same HIR expression.
-                // If we already have a ScopeItem for this HIR, we can add this
-                // next AST expression to its set
-                if let Some(existing_scope_item) = group_exprs.get_mut(&expr) {
+            // Multiple AST expressions can map to the same HIR expression, e.g.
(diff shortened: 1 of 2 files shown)


antiguru (approved):
Looks good.

aljoscha:
@antiguru tyty! 🙇‍♂️

PR #37062, opened 2026-06-16
Mark idempotent FoundationDB transactions as retryable

antiguru:
### Motivation

FoundationDB transactions can fail with a `commit_unknown_result` error, indicating the transaction may or may not have committed. Currently, all transactions in the timestamp oracle use `TransactOption::default()`, which causes these errors to propagate as failures rather than being retried.

Several transactions in the timestamp oracle are idempotent and safe to retry on this error:
- Read-only transactions (they have no side effects)
- Transactions that use `GREATEST()` to monotonically advance values (retries only move the value further forward)

This change improves reliability by allowing safe retries instead of failures.

Closes PER-25.

### Description

Updated four transaction sites in `src/timestamp-oracle/src/foundationdb_oracle.rs` to use `TransactOption::idempotent()` instead of `TransactOption::default()`:

1. **`max_read_ts()`** — Pure read transaction, safe to retry
2. **`write_ts()` in `TimestampOracle::next()`** — Uses `max(write_ts + 1, proposed_next_ts)`, so retries only advance the timestamp further while preserving monotonicity
3. **`peek_write_ts()`** — Pure read transaction, safe to retry
4. **`read_ts()`** — Pure read transaction, safe to retry
5. **`apply_write()`** — Uses `GREATEST(current, write_ts)` to bump values, making it idempotent; retries cannot cause regression

Each change includes a comment explaining why the transaction is safe to retry on `commit_unknown_result` errors.

### Verification

No new tests added — this change improves error handling for existing transaction patterns. The idempotency properties are documented in the code comments and follow from the transaction logic (read-only operations and monotonic value updates via `GREATEST()`).

https://claude.ai/code/session_012DArBrNjtGVVBSdTomhsuF

### DIFF
--- a/src/timestamp-oracle/src/foundationdb_oracle.rs
+++ b/src/timestamp-oracle/src/foundationdb_oracle.rs
@@ -234,7 +234,10 @@ impl<N: Sync> FdbTimestampOracle<N> {
             .transact_boxed(
                 &(),
                 |trx, ()| self.max_rs_trx(trx).boxed(),
-                TransactOption::default(),
+                // This transaction is a pure read, so it is safe to retry on a
+                // `commit_unknown_result` ("transaction may or may not have
+                // committed") error.
+                TransactOption::idempotent(),
             )
             .await?;
         Ok(max_ts)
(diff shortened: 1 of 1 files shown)


def- (approved):
Closes https://linear.app/materializeinc/issue/PER-25/thread-tokiowork-2-panicked-at-srctimestamp-oraclesrcfoundationdb I assume

antiguru:
Yes; updated the PR body.

PR #37071, opened 2026-06-16
catalog: scope pg_description classoid lookups to pg_catalog schema

aljoscha:
### Motivation

Fixes SQL-278.

Creating any user object named `pg_class`, `pg_type`, or `pg_namespace`
in any schema broke `pg_catalog.pg_description` for everyone:

```
db error: ERROR: Evaluation error: more than one record produced in subquery
```

`pg_description` derives each comment's `classoid` from
`mz_internal.pg_description_all_databases`, which looked up the system
catalog oids with unscoped scalar subqueries (`WHERE relname =
'pg_class'`, etc.). A user object sharing one of those names made the
scalar subquery match multiple rows, so the whole view errored.

### Description

Scope the `pg_class`/`pg_type`/`pg_namespace` lookups to the
`pg_catalog` schema via a small `pg_catalog_class` CTE. The `pg_catalog`
schema is ambient and reserved (users cannot create a schema with the
`pg_` prefix), so each lookup resolves to exactly one row. Only the
classoid derivation needed scoping; no other builtin view has the same
unscoped-lookup pattern.

This matches PostgreSQL, where `pg_description` is a real catalog table
and is unaffected by user objects that happen to share those names.

### Verification

Verified against PostgreSQL 16: creating `public.pg_class` does not
affect `pg_description`; a commented user table's comment is still
returned.

Added a regression test to `test/sqllogictest/comment.slt` that creates
user tables named `pg_class`, `pg_type`, and `pg_namespace` and asserts
`pg_description` still returns a user object's comment instead of
erroring.

### DIFF
--- a/test/sqllogictest/catalog_server_explain.slt
+++ b/test/sqllogictest/catalog_server_explain.slt
@@ -3413,82 +3413,101 @@ mz_internal.pg_description_all_databases_ind:
 mz_internal.pg_description_all_databases:
   →With
     cte l0 =
-      →Arranged mz_internal.pg_class_all_databases
-    cte l1 =
-      →Differential Join %1[#0] » %0:l0[#1{relname}]
-        →Arranged l0
+      →Differential Join %1[#0] » %0:pg_namespace_all_databases[#1{nspname}]
+        →Arranged mz_internal.pg_namespace_all_databases
         →Arrange (#0)
           →Constant (1 row)
+    cte l1 =
+      →Differential Join %0:pg_class_all_databases[#2{relnamespace}] » %1[#0]
+        →Arrange (#2{relnamespace})
+          →Fused with Child Map/Filter/Project
+            Project: #1, #0, #2, #23
+              →Arranged mz_internal.pg_class_all_databases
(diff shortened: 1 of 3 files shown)


test/sqllogictest/catalog_server_explain.slt
```diff
@@ -3423,82 +3413,101 @@ mz_internal.pg_description_all_databases_ind:
 mz_internal.pg_description_all_databases:
   →With
     cte l0 =
-      →Arranged mz_internal.pg_class_all_databases
-    cte l1 =
-      →Differential Join %1[#0] » %0:l0[#1{relname}]
-        →Arranged l0
+      →Differential Join %1[#0] » %0:pg_namespace_all_databases[#1{nspname}]
```
aljoscha, line 3416: @ggevay I think you added these recently? I see that we're doing things differently now, which makes sense because we have an additional lookup. But couldn't say whether that's fine or not

ggevay (approved):
LGTM

The plan changes look fine to me!

aljoscha:
tyty! 🙇‍♂️

PR #37073, opened 2026-06-16
expr: Make CaseLiteral part of canonical reduce output

antiguru:
### Motivation

`CaseLiteral` was produced by a separate, flag-gated transform running late in physical optimization.
Index and arrangement keys are matched by structural equality on `Vec<MirScalarExpr>`, so the same expression existing as both an `If`-chain and a `CaseLiteral` depending on pipeline position made index matching sound only by accident of ordering.

### Description

Move `CaseLiteral` construction into `MirScalarExpr::reduce`, the canonical chokepoint every matcher relies on, and remove the `CaseLiteralTransform` pass together with the `enable_case_literal_transform` flag.
Index keys and query expressions now reduce to the same canonical `CaseLiteral`, so an index on a CASE-over-literals expression is matched as a fast-path lookup.
Arms are sorted by literal, so semantically-equal CASEs written with arms in different order canonicalize identically.
This also fixes a latent wrong-results bug the flag-off transform hid: the fold path now unions a folded arm's type into `return_type`, so a nullable arm no longer under-reports its nullability and gets dropped by `IsNull` folding.

### Verification

An eval-equivalence proptest (`reduce_preserves_eval`) checks the rewrite preserves evaluation and type nullability, and a sqllogictest confirms a CASE-keyed index is matched as a lookup.


antiguru:
Follow-up (separate stacked PR): extend `CaseLiteral` construction to fold an OR-of-equalities arm — `If(Or(Eq(x, l1), …, Eq(x, lN)), result, rest)` — into N lookup entries sharing one result slot.

This collapses patterns like the JSON-schema view's eleven numeric type names (`type = "int" OR … OR "double precision"` → `{"type":"number"}`) into the membership `case_lookup` instead of a leading linear chain of `=` comparisons. The main work is making `CaseLiteral::canonicalize` sharing-aware (dedupe equal result branches to one slot). The `ilike "timestamp%%"` arm stays a plain `If`, so the chain remains split there — a prefix match can't be a literal key.

src/transform/tests/test_transforms/case_literal.spec
ggevay, line 1: I'm wondering if we've lost valuable coverage with this test being deleted. But I'm not sure exactly where to move it, because I think the `.spec` framework doesn't play well with `reduce` at the moment. Could be just converted to slt for now.

antiguru: Good catch. Most of the `.spec`'s coverage already lives in `test/sqllogictest/transform/case_literal.slt` (kept; only the `ALTER SYSTEM SET enable_case_literal_transform` line was dropped): 2-arm, single-arm-no-fold, different-columns-no-merge, interrupting-arm → two lookups, and duplicate-first-wins — plus eval-correctness checks (matching / no-match / NULL input) and JSON serialization that the `.spec` didn't have.

The one genuinely-missing case — a nested CASE in a result arm folding independently — I've added to that slt in 55de812db1.

The remaining `.spec` case (a bool-result CASE in a Filter predicate folding to `case_lookup`) intentionally no longer holds: `reduce_if` rewrites bool-result `If` chains to AND/OR *before* CaseLiteral construction runs, so that one canonicalizes to boolean algebra now rather than a lookup. That's exactly the "`.spec` doesn't play well with `reduce`" mismatch — porting it verbatim would assert the old, pre-reduce shape.

def- (commented):
If we enable case literal transform we should probably fix https://linear.app/materializeinc/issue/DB-59/caseliteral-transform-returns-wrong-results-for-float-00

src/transform/tests/test_transforms/case_literal.spec
mgree: We could move to an insta/snapshot test instead?

src/expr/src/scalar/reduce/case_literal.rs
```diff
@@ -0,0 +1,524 @@
...
+            None
+        }
+    } else {
+        None
+    };
+
+    let (x, row) = result?;
+    if type_is_float_bearing(&x.typ(column_types).scalar_type) {
+        return None;
+    }
```
mgree, line 271: If `0.0` and `-0.0` are the only pain point... why not instead fudge the map, i.e., if we ever add a mapping for `0.0`, add an identical one for `-0.0` (and vice versa)?

src/expr/src/scalar/reduce/case_literal.rs
```diff
@@ -0,0 +1,524 @@
...
+            Ok(other) => panic!("unexpected datum {other:?}"),
+            Err(_) => Err(()),
+        }
+    }
+
+    proptest! {
+        #[mz_ore::test]
+        #[cfg_attr(miri, ignore)] // too slow under miri
+        fn reduce_preserves_eval(
+            arms in prop::collection::vec((-5i64..5, proptest::option::of(-100i64..100)), 2..6),
```
mgree, line 476: Can we do `arb_datum_for_scalar` instead? (Works best if we can accommodate floats, but failing that, we could just reject those samples or sample from a couple of different possible types.)

src/expr/src/scalar/func/impls/case_literal.rs
```diff
@@ -163,10 +163,107 @@ impl CaseLiteral {
...
+        assert_eq!(
+            exprs,
+            vec![MirScalarExpr::column(0), lit(200), lit(100), lit(0)]
+        );
+        assert_eq!(cl.lookup[0].expr_index, 1);
+        assert_eq!(cl.lookup[1].expr_index, 2);
+    }
+
+    #[mz_ore::test]
+    fn canonicalize_is_idempotent() {
```
mgree, line 241: Feels like a classic thing to proptest, if we care.

src/expr/src/scalar/reduce/case_literal.rs
```diff
@@ -0,0 +1,524 @@
...
+        return;
+    };
+
+    // Check that the CaseLiteral's input matches the If's common expression.
+    if exprs[0] != *common_candidate {
+        return;
+    }
+
+    if let Some(existing_idx) = cl.get(literal_row) {
+        // Duplicate literal: overwrite with the earlier arm's result (this If).
```
mgree, line 70: s/earlier/outer/, for clarity?

src/expr/src/scalar/reduce.rs
```diff
@@ -111,6 +112,14 @@ fn reduce_post(e: &mut MirScalarExpr, column_types: &[ReprColumnType], temp_stor
...
         }
-        MirScalarExpr::If { .. } => if_then::reduce_if(e, column_types),
+        MirScalarExpr::If { .. } => {
+            if_then::reduce_if(e, column_types);
+            // reduce_if may have rewritten the node away (e.g. bool branches
+            // become AND/OR); only attempt CaseLiteral construction if it is
+            // still an If.
+            if matches!(e, MirScalarExpr::If { .. }) {
+                case_literal::try_build(e, column_types);
+            }
```
mgree, line 136: Feels a little weird to sequentially re-testing here... why not update `reduce_if` itself to call `case_literal::try_build`? That's the pattern for the other MSE forms.

src/repr/src/optimize.rs
```diff
@@ -130,8 +130,6 @@ optimizer_feature_flags!({
     // See the feature flag of the same name.
     enable_cast_elimination: bool,
     // See the feature flag of the same name.
-    enable_case_literal_transform: bool,
-    // See the feature flag of the same name.
```
mgree, line 134: I don't _love_ disabling the feature flag at the same time we do the refactor. Of course, plumbing the feature flag to `MSE::reduce` will be pretty annoying, but not impossible (and maybe useful otherwise?).

mgree (commented):
I'm supportive of this move (and we should do a similar move for coalesce/case). I do think we probably want to keep the flag for now... which probably means plumbing `OptimizerFeatures` to `MSE::reduce`?

src/repr/src/optimize.rs
ggevay: I agree. For example, there is a phase ordering risk: `reduce` runs much earlier (and later, and ... everywhere) in the optimizer pipeline than the `CaseLiteral` transform did, so there is a possibility that some other transform breaks, expecting to see `If` but getting `CaseLiteral` now.

PR #37074, opened 2026-06-16
llm: Make `mz-test` skill more discoverable before running tests

ggevay:
It happens to me quite often that Claude (or other agents) try to run a test with a wrong command, due to not having read the `mz-test` skill, because it thinks it knows how to run tests even without a skill. This causes various glitches.

Add a pointer in AGENTS.md and lead the mz-test skill description with an explicit self-trigger plus the canonical-command anti-pattern, so the skill is consulted before running test commands rather than reaching for the obvious (wrong) command from memory.

### DIFF
--- a/.agents/skills/mz-test/SKILL.md
+++ b/.agents/skills/mz-test/SKILL.md
@@ -1,14 +1,17 @@
 ---
 name: mz-test
 description: >
-  Run tests + pick test framework in Materialize. Trigger: "run tests",
-  "run testdrive", "run sqllogictest", "run mzcompose", "run cargo test",
-  "run pgtest", "rewrite test results", "add a test", "reproduce a bug",
-  "write a regression test", or mentions testing, testdrive, sqllogictest,
-  mzcompose, pgtest, cargo test, nextest, flaky tests, test failures. Also
-  "test this" or "how do I verify this works". Specifics: mz-platform-checks
-  (upgrade/restart survival), mz-parallel-workload (concurrent stress),
-  mz-limits-test (many objects).
+  Run tests + pick test framework in Materialize. Invoke BEFORE running any test
+  command yourself, even mid-task — don't run one from memory, the canonical
+  command differs (e.g. `bin/sqllogictest --optimized`, not
+  `cargo build --bin sqllogictest`).
+  Trigger: "run tests", "run testdrive", "run sqllogictest", "run mzcompose",
+  "run cargo test", "run pgtest", "rewrite test results", "add a test",
+  "reproduce a bug", "write a regression test", or mentions testing, testdrive,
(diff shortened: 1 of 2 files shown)


.agents/skills/mz-test/SKILL.md
```diff
@@ -1,14 +1,17 @@
...
-  "run pgtest", "rewrite test results", "add a test", "reproduce a bug",
-  "write a regression test", or mentions testing, testdrive, sqllogictest,
-  mzcompose, pgtest, cargo test, nextest, flaky tests, test failures. Also
-  "test this" or "how do I verify this works". Specifics: mz-platform-checks
-  (upgrade/restart survival), mz-parallel-workload (concurrent stress),
-  mz-limits-test (many objects).
+  Run tests + pick test framework in Materialize. Invoke BEFORE running any test
+  command yourself, even mid-task — don't run one from memory, the canonical
+  command differs (e.g. `bin/sqllogictest --optimized`, not
+  `cargo build --bin sqllogictest`).
```
def-: I have seen this too. Did you try out if this works?

def- (approved):

.agents/skills/mz-test/SKILL.md
ggevay: Haven't tried it out yet in a new session. Will watch for this, and then we can adjust later if it's still using the wrong commands.

.agents/skills/mz-test/SKILL.md
```diff
@@ -1,14 +1,17 @@
...
-  "run testdrive", "run sqllogictest", "run mzcompose", "run cargo test",
-  "run pgtest", "rewrite test results", "add a test", "reproduce a bug",
-  "write a regression test", or mentions testing, testdrive, sqllogictest,
-  mzcompose, pgtest, cargo test, nextest, flaky tests, test failures. Also
-  "test this" or "how do I verify this works". Specifics: mz-platform-checks
-  (upgrade/restart survival), mz-parallel-workload (concurrent stress),
-  mz-limits-test (many objects).
+  Run tests + pick test framework in Materialize. Invoke BEFORE running any test
+  command yourself, even mid-task — don't run one from memory, the canonical
+  command differs (e.g. `bin/sqllogictest --optimized`, not
```
antiguru: I don't think this should be in the description, because it's instructional. I think the better place to have this is the top-level instructions.

ggevay: Moved it. The description is back to routing-only, and the "use mz-test before running tests" guidance (with the bin/sqllogictest vs cargo build --bin sqllogictest anti-pattern) now lives in the top-level instructions (AGENTS.md) instead.

PR #37075, opened 2026-06-16
llm: Note that bin/fmt takes no file arguments

ggevay:
It happens to me quite often that Claude tried to run `bin/fmt` with file arguments, but actually `bin/fmt` always formats the whole tree (only flag is --check); passing file paths errors. Make that explicit where it's documented in the `mz-run` and `mz-commit` skills, since the obvious formatter convention (scope to changed files) is wrong here.

(Separate PR from https://github.com/MaterializeInc/materialize/pull/37074, because that edits the top-level claude.md and skill descriptions, while this one just edits skill bodies, which is much less token-count sensitive.)

### DIFF
--- a/.agents/skills/mz-commit/SKILL.md
+++ b/.agents/skills/mz-commit/SKILL.md
@@ -14,7 +14,7 @@ Read `doc/developer/guide-changes.md` for the full conventions on submitting and
 
 Before committing, run these and fix any warnings:
 
-1. `bin/fmt` (formats `.rs`, `.py`, and `.proto` files)
+1. `bin/fmt` (formats all `.rs`, `.py`, and `.proto` files; takes no file arguments)
 2. `bin/lint` (can error if tools are missing; use `bin/ci-builder run stable bin/lint` as an alternative)
 3. `cargo clippy --all-targets -- -D warnings`
 
--- a/.agents/skills/mz-run/SKILL.md
+++ b/.agents/skills/mz-run/SKILL.md
@@ -41,7 +41,8 @@ Other useful flags:
 
 ## Formatting and linting
 
-Format Rust, Python, and Protobuf files with `bin/fmt`.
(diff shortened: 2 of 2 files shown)


antiguru:
Alternative take: Make bin/fmt accept file names?

ggevay:
`bin/fmt` takes less than 2 seconds on my machine, so I'd say it's fine to simply run it on the whole repo.

Edit: Claude's take is that it would be a lot of code for little benefit, with which I agree:

  Why it's genuinely substantial, not just argument plumbing. The five formatters each use a different discovery mechanism, and file-arg support means making all five conditional:
  - rustfmt — discovers targets via cargo metadata, batches across CPUs.
  - buf — operates on the src proto tree as a unit (buf format src), not arbitrary single files.
  - black / ruff / ruff-dbt — git_files "*.py" | xargs …, and there's a path-based split (files under misc/dbt-materialize/ get ruff with --target-version=py38, others don't).

  So you'd add positional parsing plus per-file classification by extension and path (to hit the right ruff config), plus rewrite each of the five task builders to bypass its own discovery, plus edge cases
  (unmatched extension, untracked file, mixed types) and tests. It touches every part of the file.

  The benefit doesn't justify it. The only reason to scope a formatter to files is speed — and at <2s for the whole repo, there's nothing to save. Worse, scoping introduces a real footgun: if the per-file routing
  ever diverges from the whole-repo invocation (e.g. the dbt/non-dbt ruff split, or rustfmt config), a local scoped run could format a file differently than CI's --check over the full tree — exactly the kind of
  confusing churn formatters are supposed to eliminate.

  The problem it'd solve is already better solved elsewhere. Agents passing file args → the docs fix we just made. Humans wanting per-file formatting → their editor already runs rustfmt/black on save; they don't
  shell out to bin/fmt per file.

def- (approved):

PR #37079, opened 2026-06-16
Scoped feature flags 1/3: scope declaration and per-replica override plumbing

antiguru:
### Motivation

First of a three-PR stack splitting #36959 (scoped feature flags) for review:
**1/3 (this) → 2/3 #37080 → 3/3 #36959**. Design:
doc/developer/design/20260609_scoped_feature_flags.md (#36947).

This PR is the additive foundation: it introduces the vocabulary the later PRs
build on, with no behavior change on its own.

### Description

* `ParameterScope` (`Environment` / `Cluster` / `Replica`), declared on system
  vars and dyncfg `Config`s and carried through to the synced system vars. The
  declaration is the single source of truth for which contexts get evaluated.
* The size-family taxonomy: `ReplicaAllocation::family` plus a size-map `family`
  field, with a `cc` / `legacy` fallback for sizes that don't set one.
* The compute controller's per-replica dyncfg override layer
  (`update_replica_dyncfg_overrides` + per-replica command specialization),
  inert until the adapter wires it in 3/3.

Nothing consumes the scope or the override layer yet, so this is a no-op.

### Verification

`test_replica_allocation_family` covers the size→family fallback; the rest is
exercised by the later PRs in the stack.


src/controller/src/clusters.rs
```diff
@@ -106,6 +116,24 @@ pub struct ReplicaAllocation {
...
+    /// Falls back to a value derived from [`Self::is_cc`] when [`Self::family`]
+    /// is unset: `"cc"` for modern sizes and `"legacy"` for the legacy t-shirt
+    /// sizes. This keeps the legacy family targetable even before every size
+    /// gains an explicit `family` in the size configuration.
+    pub fn family(&self) -> &str {
+        match &self.family {
+            Some(family) => family.as_str(),
+            None if self.is_cc => "cc",
+            None => "legacy",
+        }
```
DAlperin, line 133: Slightly lossy heuristic especially for M series clusters, but we can aim to configure the families in tandem with rollout

antiguru: I'm on it to patch the cloud side of the house. I think we can remove this very soon.

src/compute-client/src/controller/instance.rs
```diff
@@ -1091,21 +1100,57 @@ impl Instance {
...
+        {
+            params.dyncfg_updates.extend(over.clone());
+        }
+        cmd
+    }
+
+    /// Replaces the per-replica dyncfg overrides. Callers should follow this
+    /// with a configuration push (e.g. `update_configuration`) so that existing
+    /// replicas observe the new overrides.
+    pub(super) fn update_replica_dyncfg_overrides(
```
DAlperin, line 1147: Nit: this function seems to be only called once and could be inlined.

DAlperin (approved):
This LGTM, thanks

src/compute-client/src/controller/instance.rs
antiguru: The field is private, like all other fields on `Instance`; I'd like to keep it this way.

antiguru: Every field on `Instance` is private, and `controller.rs` only mutates instance state through methods passed to `instance.call(...)`. Inlining `update_replica_dyncfg_overrides` would mean making `replica_dyncfg_overrides` `pub(super)` and breaking that encapsulation (it doesn't compile otherwise — E0616), so I'd keep the one-line setter for consistency with the rest of the struct.

src/sql/src/session/vars/definitions.rs
```diff
@@ -20,6 +20,7 @@ use mz_adapter_types::timestamp_oracle::{
     DEFAULT_PG_TIMESTAMP_ORACLE_CONNPOOL_MAX_SIZE, DEFAULT_PG_TIMESTAMP_ORACLE_CONNPOOL_MAX_WAIT,
     DEFAULT_PG_TIMESTAMP_ORACLE_CONNPOOL_TTL, DEFAULT_PG_TIMESTAMP_ORACLE_CONNPOOL_TTL_STAGGER,
 };
+use mz_dyncfg::ParameterScope;
```
aljoscha, line 23: might be a sign that `ParameterScope` should not be in `mz_dyncfg`, now that we're using it here in the vars.

src/sql/src/session/vars/definitions.rs
```diff
@@ -69,6 +70,9 @@ pub struct VarDefinition {
     /// When set, prevents getting or setting the variable unless the specified
     /// feature flag is enabled.
     pub require_feature_flag: Option<&'static FeatureFlag>,
+    /// The scope at which this variable's value may be overridden by the
+    /// LaunchDarkly sync loop. Defaults to [`ParameterScope::Environment`].
```
aljoscha:
```suggestion
    /// The scope at which this variable's value may be overridden by the
    /// LaunchDarkly sync loop.
```

aljoscha: the default is there in the code and this could rot

src/sql/src/session/vars/definitions.rs
```diff
@@ -176,6 +183,14 @@ impl VarDefinition {
         self
     }
 
+    /// Declares the [`ParameterScope`] of this variable, overriding the default
+    /// of [`ParameterScope::Environment`]. See the scoped feature flags design
+    /// for the semantics of each scope class.
```
aljoscha: isn't the semantics defined on the type as well?

src/sql/src/session/vars/definitions.rs
```diff
@@ -1682,6 +1701,14 @@ pub mod cluster_scheduling {
 /// Ensuring that all syntax-related feature flags *enable* behavior means that setting all such
 /// feature flags to `on` during catalog boot has the desired effect.
 macro_rules! feature_flags {
+    // Resolve an optional `scope:` field to a `ParameterScope`, defaulting to
+    // `Environment` when the field is omitted.
+    (@scope_or_default) => {
+        ParameterScope::Environment
```
aljoscha: we hardcode `Environment` but `ParameterScope` implements it's own default, no?

src/sql/src/session/vars/definitions.rs
```diff
@@ -1717,11 +1747,15 @@ macro_rules! feature_flags {
         // Should the feature be turned on during catalog rehydration when
         // parsing a catalog item.
         enable_for_item_parsing: $enable_for_item_parsing:expr,
+        // The optional scope class. Defaults to `ParameterScope::Environment`.
```
aljoscha: Again the default is "hard-coded" here 😅

aljoscha (commented):
looks good! I did have some inline comments that I'm sure the ai can easily clean up

src/sql/src/session/vars/definitions.rs
antiguru: `ParameterScope` is also used by `mz_dyncfg`'s own `Config::scoped()`, and `mz_sql` depends on `mz_dyncfg` (not the reverse), so `mz_dyncfg` is the shared low-level home for both dyncfgs and system vars — moving it into the vars crate would invert that dependency. Happy to relocate if you have another target crate in mind, but it can't live in `mz_sql`.

src/sql/src/session/vars/definitions.rs
antiguru: Applied your wording and dropped the "Defaults to…" sentence to avoid the rot.

src/sql/src/session/vars/definitions.rs
antiguru: Reworded to point at [`ParameterScope`] itself for the per-scope semantics, rather than restating the default / linking the design doc.

src/sql/src/session/vars/definitions.rs
antiguru: Fixed across all sites: added a `const ParameterScope::DEFAULT`, and `Default::default` now delegates to it. The const constructors and this macro use `ParameterScope::DEFAULT` (const contexts can't call `Default::default()`, which is why it was hardcoded).

src/sql/src/session/vars/definitions.rs
antiguru: Same — now `ParameterScope::DEFAULT` (and the doc comment updated to match).

aljoscha (approved):

PR #37080, opened 2026-06-16
Scoped feature flags 2/3: durable persistence and introspection (inert)

antiguru:
Second of a three-PR stack splitting #36959: **1/3 #37079 (merged) → 2/3 (this)
→ 3/3 #36959**. Design: doc/developer/design/20260609_scoped_feature_flags.md
(#36947).

Adds the durable storage and introspection for scoped overrides, inert until the
resolution layer (3/3) wires it.

### Description

* Two durable catalog collections, `cluster_system_configurations` and
  `replica_system_configurations`, keyed by object id, surfaced as builtin
  **materialized views** `mz_internal.mz_cluster_system_parameters` /
  `mz_replica_system_parameters` over `mz_catalog_raw` (no new builtin tables,
  per the ongoing MV migration). CATALOG_VERSION 86 with a no-op migration.
* `CatalogState::scoped_system_parameters` is the in-memory working copy used for
  resolution, maintained from the durable state with a value-matched retraction
  in `apply` (order-independent updates); a durable
  `Op::UpdateScopedSystemParameters` reconciles the persisted cache to a complete
  desired state.
* Nothing writes the durable collections or the working copy here, so the
  introspection views stay empty and resolution falls back to the
  environment-wide value everywhere — no behavior change until 3/3.

### Verification

Catalog migration round-trip and golden-encoding tests (full `mz-catalog` suite),
regenerated sqllogictest catalog enumerations, and `ScopedParameters` unit tests.


src/adapter/src/config.rs
```diff
@@ -27,6 +29,60 @@ pub use frontend::SystemParameterFrontend;
...
+/// The full desired state of the scoped system parameters, reconciled by the
+/// system-parameter sync loop from continuous LaunchDarkly evaluation.
+///
+/// Values are the raw (unparsed) strings LaunchDarkly served, keyed by object
+/// id; the coordinator parses and resolves them at the appropriate boundary
+/// (the controller's per-replica dyncfg push for `replica`, plan-time
+/// `OptimizerFeatureOverrides` for `cluster`). Only values that differ from the
+/// environment-wide value are present, so the maps stay sparse. The sync loop
+/// is the sole writer, and each push is the *complete* desired state, so the
+/// coordinator can apply removals by diffing against its working copy.
```
DAlperin: Nit: these comments have some opus-isms leaking abstractions from other part of the code, not really just describing this struct

src/catalog/src/builtin/mz_internal.rs
```diff
@@ -2866,6 +2866,56 @@ pub static MZ_SESSIONS: LazyLock<BuiltinTable> = LazyLock::new(|| BuiltinTable {
...
+        ),
+        ("name", "The name of the cluster-coherent system parameter."),
+        (
+            "value",
+            "The cluster-scoped value of the system parameter, as resolved from LaunchDarkly.",
+        ),
+    ]),
+    is_retained_metrics_object: false,
+    access: vec![PUBLIC_SELECT],
+    ontology: None,
```
DAlperin: lets add the ontology cluster_id links

src/catalog/src/builtin/mz_internal.rs
```diff
@@ -2866,6 +2866,56 @@ pub static MZ_SESSIONS: LazyLock<BuiltinTable> = LazyLock::new(|| BuiltinTable {
...
+        ),
+        ("name", "The name of the replica-local system parameter."),
+        (
+            "value",
+            "The replica-scoped value of the system parameter, as resolved from LaunchDarkly.",
+        ),
+    ]),
+    is_retained_metrics_object: false,
+    access: vec![PUBLIC_SELECT],
+    ontology: None,
```
DAlperin: same

DAlperin (approved):
LGTM, thanks

src/catalog/src/builtin/mz_internal.rs
```diff
@@ -2866,6 +2866,83 @@ pub static MZ_SESSIONS: LazyLock<BuiltinTable> = LazyLock::new(|| BuiltinTable {
...
+    is_retained_metrics_object: false,
+    access: vec![PUBLIC_SELECT],
+    ontology: Some(Ontology {
+        entity_name: "cluster_system_parameter",
+        description: "Cluster-coherent system parameter overrides",
+        links: &const {
+            [OntologyLink {
+                name: "scoped_to_cluster",
+                target: "cluster",
+                properties: LinkProperties::fk("cluster_id", "id", Cardinality::ManyToOne),
```
antiguru: Why is this not `mz_repr::SemanticType::ClusterId`?

antiguru: Good catch — inconsistent with the replica link right below it, which used `fk_typed(..., ReplicaId)`. Switched the cluster link to `LinkProperties::fk_typed("cluster_id", "id", Cardinality::ManyToOne, mz_repr::SemanticType::ClusterId)` to match. No golden drift.

src/catalog/src/builtin/mz_internal.rs
```diff
@@ -2866,6 +2866,88 @@ pub static MZ_SESSIONS: LazyLock<BuiltinTable> = LazyLock::new(|| BuiltinTable {
...
+                    Cardinality::ManyToOne,
+                    mz_repr::SemanticType::ClusterId,
+                ),
+            }]
+        },
+        column_semantic_types: &[("cluster_id", SemanticType::ClusterId)],
+    }),
+});
+
+pub static MZ_REPLICA_SYSTEM_PARAMETERS: LazyLock<BuiltinTable> = LazyLock::new(|| BuiltinTable {
```
aljoscha: I'm afraid we can't add new builtin tables anymore. We're in the process of moving builtin tables that can be views over `mz_catalog_raw` (the durable catalog, as json) over to be builtin materialized views.

See for example: https://github.com/MaterializeInc/materialize/pull/36969

As an alternative, we could also drop the new builtins completely, since we also don't have a builtin for regular system parameters right now, I think.

aljoscha (commented):
see the one inline comment about builtins please

src/catalog/src/builtin/mz_internal.rs
antiguru: Converted both to builtin materialized views over `mz_catalog_raw`, following the `mz_clusters` / `mz_cluster_replicas` pattern. Each projects its durable collection by `kind` (`ClusterSystemConfiguration` / `ReplicaSystemConfiguration`): `parse_catalog_id(data->'key'->'cluster_id'|'replica_id')`, `data->'key'->>'name'`, `data->'value'->>'value'`, `IN CLUSTER mz_catalog_server`. Removed the now-dead `pack_*_system_parameter_update` fns and the builtin-table-update arms in `apply` (the durable rows flow through `mz_catalog_raw`); the in-memory working copy used for resolution is unchanged. Goldens regenerated — `information_schema.tables` now reports MATERIALIZED VIEW, `mz_builtin_materialized_views` count +2, `mz_tables` s%-count 50→48. No durable schema change.

aljoscha (commented):
Reviewed the **second commit only** (`catalog: persist scoped system parameters and surface in introspection`).

Overall this is clean and closely mirrors the existing durable-collection plumbing; the inert-foundation framing is clear and the commit message is excellent. A few things, grouped by severity (details inline).

**Blocking**
- CI is red (`slt-1`): `test/sqllogictest/catalog_server_explain.slt` is out of date — it's missing the `EXPLAIN MATERIALIZED VIEW` entries for the two new MVs (the commit bumped the row counts but didn't add the per-MV blocks). Regenerate with `bin/mzcompose --find sqllogictest run catalog-server-explain --rewrite`.

**Strong suggestions**
- Don't name the vendor ("LaunchDarkly") in user-facing surfaces — the `value` column comments and the new `mz_internal.md` prose. It leaks an implementation detail and goes stale if the sync source ever changes; describe it generically.
- The two introspection MVs wrap the `kind` filter in a subquery citing database-issues/8495, but that issue is join-specific and these MVs have no join. The join-less catalog MVs (`mz_databases`, `mz_schemas`) just use a plain `WHERE`. Suggest matching them.

**Nits** (inline)
- A couple of rustdoc comments narrate design/abstractions that belong in the design doc (consistent with the earlier note on `config.rs`).
- The two `apply_*_system_configuration_update` fns are byte-identical save the id type — a small generic helper would remove the duplication.
- Redundant liveness check in the reconcile loops; inconsistent visibility on the inert accessors.
- The reconcile diff/prune and the value-matched retraction are the only non-trivial logic here and are currently untested (inert). A focused unit test would land cleanly now rather than waiting for 3/3.

src/catalog/src/builtin/mz_internal.rs
```diff
@@ -2866,6 +2866,128 @@ pub static MZ_SESSIONS: LazyLock<BuiltinTable> = LazyLock::new(|| BuiltinTable {
...
+            .finish(),
+        column_comments: BTreeMap::from_iter([
+            (
+                "cluster_id",
+                "The ID of the cluster. Corresponds to `mz_clusters.id`.",
+            ),
+            ("name", "The name of the cluster-coherent system parameter."),
+            (
+                "value",
+                "The cluster-scoped value of the system parameter, as resolved from LaunchDarkly.",
```
aljoscha: Avoid naming the vendor in a user-facing column comment — it leaks an implementation detail and goes stale if the sync source changes. Something generic like "The cluster-scoped value of the system parameter." reads fine. Same for the replica MV (~line 2948) and the prose in `doc/user/content/reference/system-catalog/mz_internal.md`.

src/catalog/src/builtin/mz_internal.rs
```diff
@@ -2866,6 +2866,128 @@ pub static MZ_SESSIONS: LazyLock<BuiltinTable> = LazyLock::new(|| BuiltinTable {
...
+            ),
+        ]),
+        // Projects the durable `cluster_system_configurations` collection out of
+        // `mz_catalog_raw` (the durable catalog as JSON): the key is
+        // `{cluster_id, name}` and the value is `{value}`. The `kind` filter is
+        // pushed into a subquery by hand, as in `mz_clusters`, so the optimizer
```
aljoscha: These MVs have no join, so the database-issues/8495 rationale (a `WHERE` pushed below a LEFT JOIN) doesn't apply here — the comment over-claims. The join-less catalog MVs (`mz_databases`, `mz_schemas`) just do `FROM mz_internal.mz_catalog_raw WHERE data->>'kind' = '...'`. Suggest dropping the subquery and this comment to match. Same in the replica MV below.

src/adapter/src/catalog/apply.rs
```diff
@@ -526,6 +532,77 @@ impl CatalogState {
...
+    /// Applies a durable update to the in-memory cluster-coherent scoped
+    /// parameter working copy (`scoped_system_parameters.cluster`), keyed by
+    /// cluster id. Mirrors the durable `cluster_system_configurations`
+    /// collection; consumed at plan time via
+    /// [`CatalogState::cluster_scoped_optimizer_overrides`]. See the scoped
+    /// feature flags design.
+    ///
+    /// [`CatalogState::cluster_scoped_optimizer_overrides`]: crate::catalog::CatalogState::cluster_scoped_optimizer_overrides
+    #[instrument(level = "debug")]
+    fn apply_cluster_system_configuration_update(
```
aljoscha: This and `apply_replica_system_configuration_update` are identical apart from the id type and target map. Consider a single generic helper over `&mut BTreeMap<Id, BTreeMap<String, String>>` taking `(id, name, value, diff)` to remove the duplication.

src/catalog/src/durable/objects.rs
```diff
@@ -1022,6 +1022,98 @@ impl DurableType for SystemConfiguration {
     }
 }
 
+/// A single cluster-coherent scoped system-parameter override: parameter `name`
+/// has value `value` on the cluster `cluster_id`.
+///
+/// This is the in-memory shape of the durable `cluster_system_configurations`
+/// collection that backs cluster-coherent scoped feature flags (step 3 of the
```
aljoscha: Two doc nits: (1) "step 3" here vs "step 2" on `ReplicaSystemConfiguration` reference design-doc step numbers that will rot and read as inconsistent — I'd drop the numbers. (2) "loaded into the coordinator's working copy on startup" is slightly off: the in-memory mirror is maintained continuously by `apply.rs` on every catalog update, not only at startup.

src/adapter/src/catalog/transact.rs
```diff
@@ -2691,6 +2702,69 @@ impl Catalog {
...
+                    for (name, value) in values {
+                        desired_cluster.insert((*cluster_id, name.clone()));
+                        if existing_cluster.get(&(*cluster_id, name.clone())) != Some(value) {
+                            tx.upsert_cluster_system_config(*cluster_id, name, value.clone())?;
+                        }
+                    }
+                }
+                for (cluster_id, name) in existing_cluster.into_keys() {
+                    if !desired_cluster.contains(&(cluster_id, name.clone()))
+                        || !live_clusters.contains(&cluster_id)
```
aljoscha: `|| !live_clusters.contains(&cluster_id)` is redundant: entries for non-live clusters never enter `desired_cluster` (we `continue` above), so `!desired_cluster.contains(...)` already removes them. Same in the replica loop. Fine to keep as defensive, but it could go for clarity.

src/adapter/src/catalog/state.rs
```diff
@@ -2299,6 +2310,35 @@ impl CatalogState {
...
+            .cloned()
+            .map(OptimizerFeatureOverrides::from)
+            .unwrap_or_default()
+    }
+
+    /// Returns the entire scoped system-parameter working copy.
+    // Read by the coordinator's scoped-parameter reconcile path, which lands in
+    // a later change; the durable mirror and this accessor are inert until then.
+    #[allow(dead_code)]
+    pub(crate) fn scoped_system_parameters(&self) -> &ScopedParameters {
```
aljoscha: Visibility is inconsistent across the inert accessors: this is `pub(crate)` + `#[allow(dead_code)]`, while `cluster_scoped_optimizer_overrides` (and `ScopedParameters::{merge,is_empty}`) are `pub` with no allow. Since all are unused until 3/3, worth aligning — `pub(crate)` + allow for the lot, or confirm which are intended public API.

src/catalog/src/builtin/mz_internal.rs
antiguru: Dropped the vendor name from both column comments and the prose in `mz_internal.md` (also fixed "table" → "view" there since these are MVs now).

src/catalog/src/builtin/mz_internal.rs
antiguru: Right — no join here, so the 8495 rationale does not apply. Dropped the subquery wrapper and the comment; both MVs now `FROM mz_internal.mz_catalog_raw WHERE data->>'kind' = '...'`, matching `mz_databases`/`mz_schemas`.

src/adapter/src/catalog/apply.rs
antiguru: Collapsed into one generic `apply_scoped_system_configuration_update<Id: Ord>(map, id, name, value, diff)`; both dispatch arms call it with the cluster/replica map.

src/catalog/src/durable/objects.rs
antiguru: Dropped the design-doc step numbers on both structs, and reworded "loaded into the coordinator's working copy on startup" → "the coordinator's in-memory working copy is maintained from it on every catalog update".

src/adapter/src/catalog/transact.rs
antiguru: Removed the redundant `|| !live_*.contains(...)` in both removal loops (added a one-line note that non-live objects never enter the desired set).

src/adapter/src/catalog/state.rs
antiguru: Aligned: made `scoped_system_parameters` `pub` and dropped the `#[allow(dead_code)]`, matching `cluster_scoped_optimizer_overrides` and `ScopedParameters::{merge,is_empty}`. All four are now `pub` with no allow (consistent with CatalogState's style); pub items aren't dead-code-flagged, so no allow is needed even while inert.

src/catalog/src/builtin/mz_internal.rs
antiguru: Thanks, fixed.

aljoscha (commented):
Re-reviewed the updated second commit (`ef3cfe1632`). Thanks for the quick turnaround — all six points are addressed cleanly:

- ✅ Vendor name removed from the `value` column comments and the `mz_internal.md` prose (and "table" → "view" there too).
- ✅ Both MVs dropped the subquery + the database-issues/8495 comment and now use a plain `WHERE data->>'kind' = '...'`, matching `mz_databases`/`mz_schemas`.
- ✅ The two `apply_*` fns collapsed into one generic `apply_scoped_system_configuration_update<Id: Ord>` helper — nice, and the value-matched-retraction rationale moved into its doc.
- ✅ `objects.rs` docs: dropped the fragile "step 2/3" references and corrected "on startup" → "maintained from it on every catalog update".
- ✅ Redundant `|| !live_*.contains(...)` checks removed, with a clarifying comment on why a missing entry already covers dropped objects.
- ✅ Accessor visibility made consistent (`scoped_system_parameters()` is now `pub`, no stray `#[allow(dead_code)]`).

**Still blocking (unchanged):** CI `slt-1` is still red on the same error — `test/sqllogictest/catalog_server_explain.slt` is missing the `EXPLAIN MATERIALIZED VIEW` entries for the two new MVs:

```
Missing EXPLAIN queries (2):
+ EXPLAIN MATERIALIZED VIEW "mz_internal"."mz_cluster_system_parameters";
+ EXPLAIN MATERIALIZED VIEW "mz_internal"."mz_replica_system_parameters";
```

The commit regenerated `mz_internal.slt` / the index-accounting slt and bumped the row counts in `catalog_server_explain.slt`, but the per-MV EXPLAIN blocks still need to be added. Regenerate with:

```
bin/mzcompose --find sqllogictest run catalog-server-explain --rewrite
```

Once that lands and slt-1 is green, this LGTM from my side.

src/adapter/src/config.rs
```diff
@@ -27,6 +29,46 @@ pub use frontend::SystemParameterFrontend;
...
+impl ScopedParameters {
+    /// Returns `true` if there are no cluster or replica overrides.
+    pub fn is_empty(&self) -> bool {
+        self.cluster.is_empty() && self.replica.is_empty()
+    }
+
+    /// Returns a copy of `self` with the cluster and replica overrides from
+    /// `other` merged in, replacing any existing entry for the same object.
+    ///
+    /// Create-time resolution uses this to add freshly-evaluated new objects to
```
aljoscha: nit: this second paragraph narrates the 3/3 resolution layer ("create-time resolution", "periodic full-state reconcile") that doesn't exist yet in this commit — the same kind of leaking-abstraction the comment thread above flagged. `merge` itself is just a two-map `extend`. I'd trim to the local contract, e.g.: "Returns a copy of `self` with `other`'s entries merged in, replacing any existing entry for the same object. Expresses no removals." The caller-side rationale can live where the caller lands.

src/adapter/src/catalog/state.rs
```diff
@@ -2299,6 +2310,33 @@ impl CatalogState {
         &self.system_configuration
     }
 
+    /// Returns the cluster-coherent scoped optimizer-feature overrides for
+    /// `cluster_id` from the in-memory scoped-parameter working copy.
+    ///
+    /// These are layered on top of the cluster's manual `CREATE CLUSTER ...
```
aljoscha: nit: this precedence model (cluster-scoped LD rule beats a manual `FEATURES` pin beats env-wide) is implemented in the 3/3 resolution layer, not here — and it re-introduces the vendor name we just took out of the user-facing surfaces. This accessor is just a lookup + `From`. Suggest trimming the doc to the contract (returns the cluster's scoped optimizer-feature overrides, empty if none) and keeping the precedence/layering reasoning where the layering actually happens (or in the design doc).

aljoscha (commented):
Two doc-comment nits from a closer re-read (the introspection views themselves look correct — verified `parse_catalog_id` handles both id types, the `kind` discriminator matches, and the absence of `.with_key` is required since the optimizer can't prove the composite key). Both nits are docs that narrate the 3/3 resolution layer rather than the local contract.

aljoscha (approved):
Left a couple more comments before, and would be good to resolve these. But otherwise good to merge!

src/adapter/src/config.rs
antiguru: Trimmed to the local contract: "Returns a copy of `self` with `other`'s entries merged in, replacing any existing entry for the same object. Expresses no removals." Caller-side rationale will live with the caller in 3/3.

src/adapter/src/catalog/state.rs
antiguru: Trimmed to the contract (returns the cluster's scoped optimizer-feature overrides, empty if none) and dropped the precedence/vendor prose — that layering lives in 3/3 / the design doc.

antiguru:
Thanks for the reviews!

PR #36773, opened 2026-05-28
[SS-69] user-facing docs for Iceberg sink, now with GCP support

ublubu:
Adding docs for:
- GCP connection
- GCP Lakehouse/BigLake Iceberg Catalog Connection _(BigLake is still the name of the API and everything in the GCP console, but it lives under a Lakehouse umbrella now.)_

Small changes to docs for Iceberg sink.

_stacked on #36695_

PR #37083, opened 2026-06-16
[SS-69] user-facing docs for Iceberg sink, now with GCP support

ublubu:
reopening #36773 

----

Adding docs for:
- GCP connection
- GCP Lakehouse/BigLake Iceberg Catalog Connection _(BigLake is still the name of the API and everything in the GCP console, but it lives under a Lakehouse umbrella now.)_

Small changes to docs for Iceberg sink.


doc/user/data/examples/create_connection.yml
```diff
@@ -707,3 +759,27 @@
...
+
+- name: "example-iceberg-catalog-gcp-connection"
+  description: |
+    The following example creates a GCP connection and an Iceberg catalog connection for Google Cloud BigLake:
+  code: |
+    -- Base64-encode the entire JSON payload of the GCP service account key.
+    -- e.g. `base64 < sa_key.json`
+    -- Paste the resulting base64-encoded string into a CREATE SECRET statement.
+    CREATE SECRET gcp_service_account_key
+      AS decode('<base64-encoded service account key JSON>', 'base64');
```
ublubu: @kay-kim I did something a little different than you suggested here (and in my other combined CREATE SECRET, CREATE CONNECTION TO GCP example)

I wanted to make sure the json/base64 instructions were always present whenever you're looking at a `CREATE CONNECTION _ TO GCP` example because:
- these instructions only apply to secrets being used for GCP connections
- every GCP connection must have a secret created in this way

ublubu: @def- ran into some trouble with the previous version of these instructions, so I'm hoping this wording is better.

doc/user/data/examples/create_connection.yml
```diff
@@ -656,44 +656,96 @@
...
+      description: |
+        *Value:* base64-encoded JSON. Required.
+
+        Create a [GCP service account](https://docs.cloud.google.com/iam/docs/service-account-overview)
+        for Materialize to use.
+
+        Create a [service account key](https://docs.cloud.google.com/iam/docs/keys-create-delete) (JSON).
+
+        Base64-encode the entire JSON service account key and paste it into the `CREATE SECRET` statement.
+        (e.g. `base64 < sa_key.json`)
```
ublubu: json + base64 instructions here

I also renamed it to `<service_account_key_json_base64>`.

doc/user/content/serve-results/sink/iceberg-gcp.md
```diff
@@ -0,0 +1,94 @@
...
+account](https://docs.cloud.google.com/iam/docs/service-account-overview) you own.
+
+1. Create the [service account](https://console.cloud.google.com/iam-admin/serviceaccounts).
+2. Grant the service account these roles on your **project**:
+    - `biglake.editor` (BigLake Editor)
+    - `serviceusage.serviceUsageConsumer` (Service Usage Consumer)
+3. Grant the service account this role on your **Iceberg warehouse bucket**:
+    - `storage.objectUser` (Storage Object User)
+4. [Create a service account key in JSON format.](https://docs.cloud.google.com/iam/docs/keys-create-delete#iam-service-account-keys-create-gcloud)
+   You will need the entire JSON payload for the next step.
```
ublubu: I didn't include the base64 instructions here, but they're part of the example in the next step.

In response to Dennis's feedback, I did add a note about "entire JSON payload for the next step" here.

doc/user/data/examples/create_sink_iceberg.yml
```diff
@@ -154,6 +195,17 @@
...
+- name: "restrictions-limitations-gcp-maintenance"
+  content: |
+    [Iceberg tables managed by the Lakehouse runtime catalog (BigLake API)](https://docs.cloud.google.com/lakehouse/docs/key-concepts#metastore-iceberg)
+    do not receive automated maintenance like expiring old snapshots and compacting manifests and data files.
+
+    Without table maintenance, table metadata grows over time and will eventually exceed BigLake's allowed limit.
+    This will prevent Materialize Iceberg sinks from committing new data.
+
+    _[Iceberg tables managed by BigQuery](https://docs.cloud.google.com/lakehouse/docs/key-concepts#iceberg-managed-tables)_
+    _do receive automated maintenance, but only BigQuery can write to them._
```
ublubu: I tried to preempt questions about Google's April announcement about automated Iceberg table maintenance without being too sassy.

doc/user/data/examples/create_connection.yml
kay-kim: This is fine.  we should include the same comments in the syntax code-block instead of the syntax table.  see comment above.

doc/user/data/examples/create_connection.yml
kay-kim: This isn't a part of the CREATE CONNECTION  syntax so that we're now mixing create secret and create connection syntax into one table that also includes instructions to perform outside of materialize. 
I'd:

- Make it into a comment in the code block and not here.

- If you also want to somehow call it out ... maybe include  a row for `SECRET <secret_name>`  
and for the description, something like (spitballing): Secret containing the GCP service account key (JSON).  Base64-encode it first and decode it in the `CREATE SECRET` statement to avoid escaping quotes and newlines in the SQL string literal.

doc/user/content/sql/create-connection.md
```diff
@@ -178,6 +178,18 @@ CREATE CONNECTION gcs_connection TO AWS (
 );
 ```
 
+If you are exporting to Google Cloud Storage using [Iceberg sinks](/sql/create-sink/iceberg), use a [GCP connection](#gcp).
+
+### GCP
+{{< private-preview />}}
+
+A GCP connection stores a [GCP service account key](https://docs.cloud.google.com/iam/docs/keys-create-delete).
+You can use a GCP connection to export data to [Lakehouse/BigLake](https://docs.cloud.google.com/lakehouse/docs/lakehouse-iceberg-rest-catalog) via [Iceberg sinks](/sql/create-sink/iceberg).
```
kay-kim: So .. flip the order of the two.

Instead of  "A GCP connection stores ..." ... go with "The GCP connection uses a [GCP service account key (JSON)](link) to connect." 
With the new ordering, you could also call out instructions about creating a service account and encoding the key here. that way, it comes before the syntax.

doc/user/content/serve-results/sink/iceberg-gcp.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Prerequisites
+
+- Google Cloud project with the [BigLake API enabled](https://docs.cloud.google.com/lakehouse/docs/lakehouse-iceberg-rest-catalog#before_you_begin).
+- Google Cloud [Storage bucket](https://console.cloud.google.com/storage/browser) to serve as the Iceberg warehouse.
+- [Lakehouse runtime catalog](https://docs.cloud.google.com/lakehouse/docs/lakehouse-iceberg-rest-catalog#create_a_catalog) backed by your warehouse bucket.<br/>
+- [Namespace in the catalog](https://docs.cloud.google.com/lakehouse/docs/lakehouse-iceberg-rest-catalog#create_a_namespace_or_schema).
+
+{{< note >}}
+  You may configure your catalog with either _End-user credentials_ or _Credential vending mode_.
```
kay-kim: This note belongs to the Lakehouse running catalog?
If so, I'd add it under that bullet point because it's relevant there, yes?

Also, I'd specify what this configuration is for:
For **Authentication method**, you can select either ...

doc/user/content/serve-results/sink/iceberg-gcp.md
```diff
@@ -0,0 +1,94 @@
...
+  You will provide Materialize with a [GCP service account key](https://docs.cloud.google.com/iam/docs/keys-create-delete#iam-service-account-keys-create-gcloud) for authentication in the next step.
```
kay-kim: not clear the purpose of this sentence here.  is it relevant here? If so, we might make this more clear.  Otherwise, I would get rid of it.

doc/user/content/serve-results/sink/iceberg-gcp.md
```diff
@@ -0,0 +1,94 @@
...
+{{< /note >}}
+
+## Create the Iceberg catalog connection in Materialize
+
+### Step 1. Set up permissions in GCP
+
+Materialize authenticates to BigLake as a Google Cloud [service
```
kay-kim: maybe reword to "Materialize uses a Google Cloud service account to authenticate ..."  that way, the sentence is focused on the service account

doc/user/content/serve-results/sink/iceberg-gcp.md
kay-kim: Since the base64 is only in the comment, I would also include the base64 encoding as a step here.

That is, leave the comment in the example ... but also, have a step here .. because they need to do it outside of Materialize.

doc/user/data/examples/create_connection.yml
```diff
@@ -707,3 +759,27 @@
         AWS CONNECTION = aws_connection
     );
   testable: false
+
+- name: "example-iceberg-catalog-gcp-connection"
+  description: |
+    The following example creates a GCP connection and an Iceberg catalog connection for Google Cloud BigLake:
+  code: |
+    -- Base64-encode the entire JSON payload of the GCP service account key.
```
kay-kim: Since we're explicitly doing the base64-encoding step (see comment), maybe tweak the comment to:
-- Using the base64-encoded service account key (e.g., `base64 < sa_key.json`)

doc/user/data/examples/create_connection.yml
```diff
@@ -707,3 +759,27 @@
         AWS CONNECTION = aws_connection
     );
   testable: false
+
+- name: "example-iceberg-catalog-gcp-connection"
+  description: |
+    The following example creates a GCP connection and an Iceberg catalog connection for Google Cloud BigLake:
```
kay-kim: Would add reference link to the two connections

kay-kim (approved):

PR #37085, opened 2026-06-16
Add in-map index for O(log n) jsonb field access

antiguru:
### Motivation

Accessing a `jsonb` field linearly scans the underlying `DatumMap`, so a "JSON to columns" query pulling `k` fields from an `n`-key object does `O(n * k)` work per row. Map entries are already key-sorted at pack time, so a small in-map index lets a key be found by binary search.

### Description

Adds a deterministic index to the in-memory `Row` map encoding, making single-key lookup `O(log n)`. A non-empty map gains a suffix `[ offsets ][ count_word ]`: offsets are stored at the smallest width (u8/u16/u32) that fits the entries, and the width class is packed into the high bits of the count word, so the suffix costs `W*(n-1)+4` bytes rather than `4n`. The index is built at pack time with no re-walk on the entry-by-entry paths (JSON packing, columnar/proto decode, Avro map decoding, `text`-to-`map` casts, the pgwire value paths); `iter()` skips it, so equality, hashing, ordering, and columnar/proto encoding are unchanged.

Every operator that resolves a single key by name uses the index, so no call site pays its cost without collecting the benefit: `->`, `->>`, `#>`, `#>>`, and `?` on `jsonb`, `->`, `?`, `?&`, `?|` on `map`, and the map arms of both `@>` implementations. The two containment operators probed one map once per entry of the other, so they drop from `O(n * m)` to `O(m * log n)`.

This is safe because the `Tag` byte layout is never persisted (persist uses Arrow, so no migration), `Row` equality is byte equality and the index is a deterministic function of the sorted entries (equal maps stay byte-identical), and `Row` sort order is implementation-defined.

### Cost

The suffix is roughly `W` bytes per entry, so its share of a `Row` tracks `W / average_entry_bytes`: worst for objects with many small entries, negligible for objects with few large values. Measured over `Jsonb::from_slice`:

| Object | JSON | `Row` bytes | Index bytes | Overhead |
| --- | --- | --- | --- | --- |
| `{"a":1,"b":2,"c":3}` | 19 B | 42 B | 6 B | +16.7% |
| 12-key event object | 206 B | 209 B | 15 B | +7.7% |
| 50 keys, short values | 1001 B | 1011 B | 102 B | +11.2% |
| 50 keys, 100-byte values | 5601 B | 5611 B | 102 B | +1.9% |
| 3 nested 3-key objects | 73 B | 150 B | 24 B | +19.0% |
| 500 keys, integer values | 6891 B | 8511 B | 1002 B | +13.3% |

The cost is in-memory only. Durable encodings are built from `iter()`, so persisted bytes are unchanged and the growth lands in arrangements, `RowArena`s, and rows in flight. There is no opt-out, because a runtime toggle over the encoding would break byte-canonicality while both forms were live in arrangements. A workload that stores `jsonb` without doing field access therefore pays the memory without collecting the win.

### Behavior changes

Ascending keys were already a `DatumMap` contract, but a linear scan tolerated a violation: it still found the key. A binary search reports a present key as missing instead, so both index builders now check the order as they pack, under soft assertions, and the proto decode path rejects out-of-order keys outright since its input is untrusted.

`RowRef` hashes its raw bytes, so anything placing rows by `Row::hashed()` shifts with the new layout. `COPY TO` S3 assigns rows to output files that way, which is why `copy-s3-roundtrip-minio.td` now expects two files for `test/4_5`. The data is unchanged, only its distribution across files.

### Performance

On the `jsonb_to_columns` microbench (50-key object, 10k rows), indexed access runs at ~10x the linear scan, 31.6 ms against 313.4 ms, and the push-time builder takes ~13% off decode alone.

Both benchmarks measure the case the index is best at. Decode dominates the persist path: 64.1 ms of the 98.6 ms `decode_and_access` figure is re-parsing JSON text, so a query scanning `jsonb` straight out of persist sees far less than 10x. The `JsonbToColumns` feature benchmark arranges its table behind an index precisely so the parse is paid once and the measurement isolates field access, which is what a query reading from an arrangement or a materialized view actually pays.

Design doc: `doc/developer/design/20260616_jsonb_map_index.md`. Tests added: `test_datum_map_get` covers lookup against a linear scan across all three offset widths, and `test_dict_builders_agree` pins down that the closure and push-time builders emit byte-identical maps, which `Row`'s byte equality depends on.

### DIFF
--- a/doc/developer/design/20260616_jsonb_map_index.md
+++ b/doc/developer/design/20260616_jsonb_map_index.md
@@ -79,22 +79,27 @@ For a map with `n > 0` entries the payload (the bytes counted by the existing
 `u64` length) becomes:
 
 ```text
-[ count: u32 ][ offset_1: u32 ] .. [ offset_{n-1}: u32 ][ entries.. ]
+[ entries.. ][ offset_1: u32 ] .. [ offset_{n-1}: u32 ][ count: u32 ]
 ```
 
-* `count` is `n`.
-* `offset_i` is the start of entry `i` relative to the first entry. Entry 0 is
-  always at offset 0 and is omitted, so the header is exactly `n` little-endian
-  `u32`s and the entries begin `4 * n` bytes in.
 * Entries are unchanged: `(key, value)` datum pairs sorted ascending by key.
+* `offset_i` is the start of entry `i` relative to the first entry. Entry 0 is
+  always at offset 0 and is omitted.
+* `count` is `n`.
 
-Empty maps keep an **empty** payload (no header), so they stay byte-identical to
-`DatumMap::empty()` and the encoding of every value remains canonical.
+The index is a **suffix** of exactly `n` little-endian `u32`s, so the entries
+occupy the first `len - 4 * n` bytes. Empty maps keep an **empty** payload (no
+suffix), so they stay byte-identical to `DatumMap::empty()` and the encoding of
+every value remains canonical.
 
 The index is built once, at the end of `RowPacker::push_dict_with`
-(`finish_dict`), by walking the just-written entries to record their offsets and
-splicing the header in front of them. Pushing an existing `Datum::Map` copies
-its bytes verbatim, so the index is never rebuilt or duplicated.
+(`finish_dict`), by walking the just-written entries. Because this runs on the
+hot path that decodes every `Row` out of persist, it is written to be cheap: the
+suffix layout lets us **append** the index instead of splicing it in front (no
+memmove), and each offset is written straight into the buffer as it is computed
+(no temporary allocation). The only inherent cost is the single `O(n)` walk of
+the entries. Pushing an existing `Datum::Map` copies its bytes verbatim, so the
+index is never rebuilt or duplicated.
 
 `DatumMap::iter()` skips the header (so the columnar encoder, proto conversion,
 equality, hashing, and ordering — all iter-based — are unaffected), and the new
--- a/doc/developer/design/20260616_jsonb_map_index.md
+++ b/doc/developer/design/20260616_jsonb_map_index.md
@@ -1,0 +1,142 @@
+# Faster `jsonb` field access via an in-map index
+
+## Summary
+
+Accessing a field of a `jsonb` value scans the underlying `DatumMap` linearly.
+A "JSON to columns" query that pulls `k` fields out of an object with `n` keys
+therefore does `O(n * k)` work per row, re-scanning the map once per field.
+
+This change adds a small, deterministic **index** to the in-memory `Row`
+encoding of maps so a single key can be found with a binary search. Field
+access drops to `O(log n)` and the whole "JSON to columns" pattern from
+`O(n * k)` to `O(k * log n)` per row, transparently for every existing
+`->`, `->>`, `#>`, and `#>>` expression — no planner or SQL changes.
+
+## Background
+
+`Datum::Map` is a thin view over a byte slice in the `Row` (`src/repr/src/row.rs`).
+The bytes are a flat, **key-sorted** sequence of `(key, value)` datum pairs:
+
+```text
+[ Tag::Dict ][ u64 byte-length ][ key0 ][ val0 ][ key1 ][ val1 ] ..
+```
+
+Decoding a `Datum::Map` is `O(1)` — it just wraps the slice — but the only way
+to find a key is `DatumMap::iter().find(..)`, a linear scan. The keys are
+already sorted (`adt::jsonb` sorts them at pack time), yet nothing exploited
+that, because the variable-length entries gave no way to jump to the middle of
+the sequence.
+
+## Goals
+
+* Sub-linear single-key access on `jsonb` objects.
+* No on-disk migration, no SQL-visible behavior change.
+* Preserve the invariants the rest of the system relies on.
+
+## The two ideas considered
+
+**1. Multi-valued scalar functions** — a `jsonb_access` that takes a list of
+field paths and returns many columns in one decode (`O(n)` per row). This is the
+asymptotically best option for "all fields", but `MirScalarExpr` is
+single-`Datum`-out by construction; producing multiple columns needs new
+expression/relation infrastructure (a multi-output operator or an optimizer
+transform that fuses sibling accesses) and only helps queries rewritten to use
+it.
+
+**2. An index in the `Row`/`Datum`** — store enough information in the map
+encoding to binary search by key (`O(log n)` per access). Contained to
+`src/repr`, and it speeds up **every** existing field access with zero query
+changes.
+
+We chose **idea 2**. It is the smaller, lower-risk change and delivers the win
+transparently. Idea 1 remains a viable future step for the extreme
+"extract every field" case (where `k ≈ n` makes a single `O(n)` decode beat
+`O(k log n)`); the two are complementary.
+
+## Why idea 2 is safe
+
+The in-memory `Tag`-based encoding is constrained by exactly three properties,
+all of which this change respects:
+
+1. **It is not persisted.** Durable storage uses `ProtoRow` (protobuf) and a
+   separate Arrow/Parquet columnar encoding (`src/repr/src/row/encode.rs`); the
+   `Tag` byte layout never reaches S3. So there is **no migration** and no
+   format-version concern.
+2. **`Row` sort order is implementation-defined.** `RowRef::cmp` compares raw
+   bytes purely as an arbitrary-but-consistent total order; no correctness
+   depends on map bytes sorting in any particular logical order. Changing the
+   map layout is therefore free of ordering-semantics fallout.
+3. **`Row` equality is byte equality.** This is the one hard constraint: two
+   equal map values *must* encode to identical bytes. The index is a pure,
+   deterministic function of the (already sorted) entries, so equal maps still
+   produce equal bytes.
+
+## Design
+
+### Encoding
+
+For a map with `n > 0` entries the payload (the bytes counted by the existing
+`u64` length) becomes:
+
+```text
+[ count: u32 ][ offset_1: u32 ] .. [ offset_{n-1}: u32 ][ entries.. ]
+```
+
+* `count` is `n`.
+* `offset_i` is the start of entry `i` relative to the first entry. Entry 0 is
+  always at offset 0 and is omitted, so the header is exactly `n` little-endian
+  `u32`s and the entries begin `4 * n` bytes in.
+* Entries are unchanged: `(key, value)` datum pairs sorted ascending by key.
+
+Empty maps keep an **empty** payload (no header), so they stay byte-identical to
+`DatumMap::empty()` and the encoding of every value remains canonical.
+
+The index is built once, at the end of `RowPacker::push_dict_with`
+(`finish_dict`), by walking the just-written entries to record their offsets and
+splicing the header in front of them. Pushing an existing `Datum::Map` copies
+its bytes verbatim, so the index is never rebuilt or duplicated.
+
+`DatumMap::iter()` skips the header (so the columnar encoder, proto conversion,
+equality, hashing, and ordering — all iter-based — are unaffected), and the new
+`DatumMap::get()` binary searches the header.
+
+### Why not a new `Tag` or a runtime feature flag?
+
+A second tag (`DictIndexed` alongside `Dict`) would let indexed and legacy maps
+coexist — but coexistence is exactly what byte-equality forbids: if the same
+logical map could be emitted under either tag, two equal maps could differ in
+bytes and break dedup/join/arrangement keys. To stay canonical you would have to
+stop emitting plain `Dict` anyway, so the second tag buys nothing over changing
+`Dict` in place (and costs a discriminant plus dual read paths everywhere). The
+`Tag` `u8` space is not the constraint (~94 of 256 used); canonicalness is.
+
+For the same reason, a flag that *toggles the encoding* at runtime is unsafe:
+while both encodings are live in arrangements, equal maps have unequal bytes. If
+a kill-switch is ever wanted, the safe form gates only the **read** path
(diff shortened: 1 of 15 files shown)


def-:
## QA LLM Review

### 1. HIGH -- `push_dict_with` re-walks the entries it just wrote, so a fallible closure that fails between key and value panics

`src/repr/src/row.rs:2750`

`push_dict_with` calls `finish_dict` unconditionally after the closure returns, including when the closure returned an error partway through an entry. `finish_dict` walks the entries as `(key, value)` pairs, so a dict left with a trailing key and no value makes it `read_datum` off an empty slice and panic. `src/arrow-util/src/reader.rs:1092` has exactly that shape, so a Parquet file whose map column holds one unconvertible value aborts the `clusterd` process running `COPY INTO ... FROM ... (FORMAT PARQUET)` instead of returning the error it returned before.

<details>
<summary>Details</summary>

Minimal repro, as an `mz-repr` unit test on this branch:

```rust
let mut row = Row::default();
let res: Result<(), &str> = row.packer().push_dict_with(|packer| {
    packer.push(Datum::String("a"));
    packer.push(Datum::Int64(1));
    packer.push(Datum::String("b"));
    Err("value push failed")
});
```

```
thread 'row::tests::repro_partial_entry_panic' panicked at src/repr/src/row.rs:1599:16:
index out of bounds: the len is 0 but the index is 0
```

The reachable call site:

```rust
packer.push_dict_with(|packer| {
    while let Some((key, i)) = kv_sorted.next() {
        ...
        packer.push(Datum::String(key));
        values.read(i, packer)?;      // <- returns Err with the key already pushed
    }
    Ok::<_, anyhow::Error>(())
}).context("pack map")?;
```

Every scalar arm of `ColReader::read` computes its `Datum` and `?`s out *before* pushing, so an error leaves the key orphaned: `ColReader::TimestampSecond` on an out-of-range value (`src/arrow-util/src/reader.rs:832`), `ColReader::Jsonb` on a string that is not valid JSON (`:820`), `ColReader::Uuid` on a wrong-length blob (`:791`). A table column of `map[text=>timestamp]` or `map[text=>jsonb]` fed from a Parquet `map<utf8, ...>` reaches all three. `test/testdrive/copy-from-parquet-unsorted-map-keys.td` shows the map path is supported and exercised, but it uses `map[text=>text]`, whose value read cannot fail, so it will not catch this.

`decode_chunk` runs in the storage cluster (`src/storage/src/render.rs:508`) and previously surfaced these as `StorageErrorXKind::ParquetError`. Now they are a process abort that recurs on every retry of the statement, taking every dataflow on that replica with it. The input is entirely user-supplied.

`src/interchange/fuzz/fuzz_targets/json_encode.rs:444` has the same shape (`push_typed(packer, u, value, *value_nullable)?` after the key push), so the fuzz target will report this as a crash too.

The PR already converted the other fallible dict closures (`src/pgrepr/src/value.rs:295`, `:831`, `src/repr/src/row/encode.rs:2131`) to `push_indexed_dict_with`, which is immune because it never re-walks. These two were missed. Fix:

```diff
-                packer
-                    .push_dict_with(|packer| {
+                packer
+                    .push_indexed_dict_with(|builder| {
                         while let Some((key, i)) = kv_sorted.next() {
                             // Parquet docs state that if there are duplicate keys, the last value
                             // should be used, so skip duplicates here.
                             //
                             // sorted_by_key is a stable sort, so entries with duplicate keys will
                             // maintain their original order, and we can pick the last one here.
                             if let Some((next_key, _)) = kv_sorted.peek() {
                                 if key == *next_key {
                                     continue;
                                 }
                             }
-                            packer.push(Datum::String(key));
-                            values.read(i, packer)?;
+                            builder.push_entry(key, |packer| values.read(i, packer))?;
                         }
                         Ok::<_, anyhow::Error>(())
                     })
                     .context("pack map")?;
```

Converting the call sites fixes the reachable instances, but the hazard stays latent for the next fallible `push_dict_with` closure anyone writes: before this change a mid-entry error was harmless because nothing read the half-written map. Worth closing at the source too, either by having `finish_dict` drop a truncated trailing entry rather than reading past it, or by giving `try_push_dict_with` a rollback to `start` on error and steering fallible callers to it.

</details>

antiguru:
Confirmed and fixed in 5438bcb0a2. The repro reproduces exactly as written, panicking at `row.rs:1599` with `index out of bounds: the len is 0 but the index is 0`, and the analysis of the `ColReader::Map` call site is right: the key is pushed before a fallible value read that returns before pushing anything, so `COPY INTO ... FROM ... (FORMAT PARQUET)` over a `map[text=>timestamp]` or `map[text=>jsonb]` with one bad value aborted the process instead of returning `ParquetError`.

I took the suggestion to close it at the source rather than only at the call sites, since the hazard would otherwise stay latent. `finish_dict` now walks the entries read-only, collecting each one's offset into an inline `SmallVec`, stops at the last complete pair, and truncates the orphaned key away before appending the suffix. Doing the walk before any write is what makes this work: an offset for the truncated entry is never appended, so there is nothing to revise once the orphan turns up, and no memmove is needed to remove it. It then delegates to `finish_dict_from_offsets`, so the closure and push-time builders now emit identical suffixes by construction instead of by parallel implementation, which is a property `Row`'s byte equality depends on. Regression test is `test_dict_with_failed_entry_is_readable`, covering 0, 1, 2, and 40 complete entries before the failure and asserting the resulting map is readable and reports the right cardinality.

Call sites converted to `push_indexed_dict_with` as suggested: the Parquet map reader and the `json_encode` fuzz target. I also converted `jsonb_build_object` (`src/expr/src/scalar/func/variadic.rs`), which the review did not flag. It is safe today, because it rejects a null key before pushing it, but the ordering was load-bearing and invisible, and it is an entry-by-entry caller that should not be re-walking anyway.

Two notes on scope. The remaining `push_dict_with` callers are all infallible closures, so they cannot produce an orphan: `audit_log_details.rs:401`, `string.rs:607`, `jsonb.rs:214`, `dataflows.rs:695` (uses `expect`, not `?`), `storage-client/src/client.rs:272`, plus tests and benches. And `try_push_dict_with` was not the lever here, since the reachable site calls `push_dict_with` directly with a `Result`-returning closure and gets the error back as the closure's return value, so a rollback added to `try_push_dict_with` would not have covered it.

Verified: `bin/lint` clean, `cargo test` green across mz-repr (182), mz-expr, mz-arrow-util, mz-interchange, mz-pgrepr, mz-storage-types, the fuzz target compiles on nightly, and `jsonb.slt` plus `map.slt` pass 557/557.

Tracked as CPU-230.

def-:
## QA LLM Review

### 1. MEDIUM -- `DictBuilder::push_entry` indexes an entry before its value exists, so the push-time builder still emits an unreadable map when a closure fails mid-entry

`src/repr/src/row.rs:870`

`push_entry` appends the entry's offset before running the value closure, so a closure that fails between key and value leaves the orphaned key in the buffer *and* counted in the index. `finish_dict_from_offsets` then writes a count word claiming that entry, and `DatumMap::get`/`iter` panic on it with the same `index out of bounds: the len is 0 but the index is 0` the prior fix removed from `finish_dict`. Every fallible dict closure in the tree now sits on this path, so `test_dict_with_failed_entry_is_readable` guards the one builder whose remaining callers are all infallible.

<details>
<summary>Details</summary>

Reproduction, as an `mz-repr` unit test on this branch:

```rust
let mut row = Row::default();
let res: Result<(), &str> = row.packer().push_indexed_dict_with(|builder| {
    builder.push_entry("zzz", |_packer| Err("value push failed"))?;
    Ok(())
});
assert_eq!(res, Err("value push failed"));
let datum = row.unpack_first();
let map = datum.unwrap_map();
assert_eq!(map.len(), 0); // fails: left 1, right 0
map.get("zzz");           // panics at row.rs:1600
```

`map.len()` reports 1, and both `get("zzz")` and `iter()` panic at `read_byte`, because `entries()` ends immediately after the orphaned key. So the two builders are not in fact byte-identical on this path: on the same mid-entry failure `push_dict_with` yields a valid `n-1` entry map and `push_indexed_dict_with` yields a map with a count of `n` whose last entry has no value. `test_dict_builders_agree` only pins the success path down, and `push_indexed_dict_with`'s docs ("equivalent to `push_dict_with`", "the same contract applies") say nothing about the difference.

No live panic today: I checked each fallible call site and every one drops the row on the error path — `ArrowReader::read_all` (`src/arrow-util/src/reader.rs:113`) uses a local `Row`, `RowArena::try_make_datum` (`src/repr/src/row.rs:3839`) drops its scratch row, `decode_text_into_row`'s only caller packs into `SharedRow`, which `SharedRow::get` clears on the next borrow, and the `try_push_proto` callers either propagate or `expect`. That is what makes this latent rather than reachable, but it is not an invariant any of those call sites states, and the commit's own doc comment and regression test tell a reader the opposite. The next fallible closure that logs, keeps, or inspects the row after the error reintroduces the original panic.

Fix, keeping the offset out of the index until the value is actually written (verified: the whole `mz-repr` `row::` suite plus a builder-path version of `test_dict_with_failed_entry_is_readable` that also asserts byte-equality with the closure builder passes):

```diff
-        let offset = self.packer.row.data.len() - self.entries_start;
-        self.offsets
-            .push(u32::try_from(offset).expect("map larger than 4 GiB cannot be indexed"));
-        self.packer.push(Datum::String(key));
-        value(self.packer)
+        let entry_start = self.packer.row.data.len();
+        let offset = entry_start - self.entries_start;
+        self.packer.push(Datum::String(key));
+        let key_end = self.packer.row.data.len();
+        let res = value(self.packer);
+        if self.packer.row.data.len() > key_end {
+            self.offsets
+                .push(u32::try_from(offset).expect("map larger than 4 GiB cannot be indexed"));
+        } else {
+            // `value` wrote nothing, so it failed between the key and the value.
+            // Drop the orphaned key rather than indexing an entry without one.
+            self.packer.row.data.truncate(entry_start);
+        }
+        res
```

This needs no call-site changes and keeps the ascending-key soft assertion working, since `self.offsets.last()` on entry is still the previous complete entry. A `try_push_entry` taking `F: FnOnce(&mut RowPacker) -> Result<(), E>` and rolling back to `entry_start` on `Err` would be more explicit if you would rather not infer failure from "the closure wrote nothing"; either way, extending `test_dict_with_failed_entry_is_readable` to run both builders is what keeps them from drifting apart again.

</details>

doc/developer/design/20260616_jsonb_map_index.md
```diff
@@ -0,0 +1,249 @@
...
+  top two bits.
+
+Offsets are stored at width `W` ∈ {1, 2, 4} bytes, the smallest that holds any
+offset, selected from the entries' total byte length (`≤256 → u8`, `≤65536 →
+u16`, else `u32`). `W` is a deterministic function of the entries' bytes, so
+equal maps select the same width and stay byte-identical. The index is a
+**suffix** of `W * (n - 1) + 4` bytes, so the entries occupy the first
+`len - 4 - W * (n - 1)` bytes. Empty maps keep an **empty** payload (no
+suffix), so they stay byte-identical to `DatumMap::empty()` and the encoding of
+every value remains canonical.
```
mgree, line 107: The phrasing here took me a minute to realize that `W` varies per map, not per offset. Maybe something like:
```suggestion
Offsets are stored at width `W` ∈ {1, 2, 4} bytes, where `W` is fixed for the whole map.
We choose the smallest `W` that holds every
offset, selected from the entries' total byte length (`≤256 → u8`, `≤65536 →
u16`, else `u32`). `W` is a deterministic function of the entries' bytes, so
equal maps select the same width and stay byte-identical. The index is a
**suffix** of `W * (n - 1) + 4` bytes, so the entries occupy the first
`len - 4 - W * (n - 1)` bytes. Empty maps keep an **empty** payload (no
suffix), so they stay byte-identical to `DatumMap::empty()` and the encoding of
every value remains canonical.
```

src/repr/src/row.rs
```diff
@@ -812,6 +813,68 @@ pub struct RowPacker<'a> {
...
+            .field("entries_start", &self.entries_start)
+            .field("entries_pushed", &self.offsets.len())
+            .finish_non_exhaustive()
+    }
+}
+
+impl DictBuilder<'_, '_> {
+    /// Pushes one `(key, value)` entry, recording its start offset for the
+    /// index. `value` is run with the underlying packer to write the value,
+    /// which may itself be a nested list or dict.
```
mgree, line 842: We should highlight that `value` really should push only one element. (A badly behaved one could, say, push the next key!)

src/repr/src/row.rs
```diff
@@ -2504,6 +2765,48 @@ impl RowPacker<'_> {
...
+        // row buffer again.
+        drop(builder);
+
+        finish_dict_from_offsets(&mut self.row.data, entries_start, &offsets);
+
+        let len = u64::cast_from(self.row.data.len() - start - size_of::<u64>());
+        self.row.data[start..start + size_of::<u64>()].copy_from_slice(&len.to_le_bytes());
+
+        res
+    }
```
mgree, line 2808: I think this has the same problem @def-'s agent talked about: if there are errors from the closure, we don't ever notice and have unconditionally appended the bad offsets. My agent suggested the following fix, which I have not checked:

> An O(1) fix is possible: after the closure, read the key at the last recorded offset and check a value follows before the entries end; if not, pop the offset and truncate, matching finish_dict.

This is O(1) per key, afaict, so it may put us back into the re-walk scenario (but at least not recomputing the offsets post facto).

mgree (commented):
Overall looks good to me.

Beyond the comments, it seems like there are only six non-test uses of `push_dict_with` that aren't in tests or benchmarks, all of which could be moved to `push_indexed_dict_with`. If we don't migrate these calls now, we should do it soon---and drop `push_dict_with`. (A cursory inspection with an agent suggests that the usages in tests could _also_ be converted easily.)

mgree:
Oh also: the SLT failure points to some sneaky iteration order change, worth looking at.

PR #37087, opened 2026-06-17
avro: support numeric promotion for nullable (union) columns

tonydu-mz:
Fixes [SS-277](https://linear.app/materializeinc/issue/SS-277/avro-intdouble-promotion-fails-for-nullable-union-columns-during).

## Motivation

A customer's Kafka + Confluent Schema Registry source failed to decode records written with an older schema version after a nullable column's type was promoted from `int` to `double`:

```
Decode error: avro deserialization error: unable to decode row (Avro schema id = 38):
Schema resolution error: Failed to match writer union variant `Int` against any variant
in the reader for field `...rentable_area_size`
```

`int` → `double` is a valid, non-breaking Avro promotion. We already handle it for non-nullable columns, but a nullable column is encoded as a union (`["null","int"]` → `["null","double"]`), and union-variant matching used an exact `SchemaKind` lookup with no promotion. The writer's `Int` variant never matched the reader's `Double` variant, so the pairing was stored as an error in the resolved union and re-raised at decode time when a record expressed the int branch.

## Changes

- `src/avro/src/schema.rs`: add `can_promote()` (the numeric promotions, kept in sync with the pairs `reader.rs` can actually resolve) and promotion-aware `UnionSchema` matchers (`match_promote_writer`, `match_ref_promote_writer`, `match_ref_promote_reader`).
- `src/avro/src/reader.rs`: use the promotion-aware matchers in all three union resolution arms (union↔union, concrete→union, union→concrete). An exact match is still preferred; only on a miss do we look for a promotion target, choosing the first reader variant in declaration order. No decoder change is needed.

## Tests

- Three in-process unit tests in `src/avro/src/lib.rs`, one per resolution arm (`test_union_int_to_double_promotion`, `test_concrete_int_to_union_double_promotion`, `test_union_int_to_concrete_double_promotion`).
- `test/testdrive/avro-resolution-union-promote.td` as an end-to-end repro through a Kafka source + CSR.
- `cargo test -p mz-avro --lib` passes (66 tests). Confirmed the existing `avro-resolution-*` tests do not regress: every error-asserting case is a genuine narrowing / `Null` / `String` mismatch, for which `can_promote` returns false.

## Serialization

Checked the Avro sink/encode path per review: no analogous issue. The encoder is purely structural (it trusts the union index and writes against `variants()[index]`), the sink encoder builds unions by fixed index against a schema Materialize generates itself, and none of the changed matching functions are reachable from the encode path. Encoding never reconciles two schemas, so there is nothing to promote on the write side.

### Scope / follow-ups

- Limited to numeric promotion. Avro also permits `string` ↔ `bytes` promotion, which has the same union-matching gap and could be added to `can_promote` later.
- When a reader union contains multiple promotion targets (e.g. `["long","double"]` for a writer `int`), we take the first in declaration order. This is unambiguous for the common `["null", T]` shape.

## Release note

> Fix a schema-resolution error (`Failed to match writer union variant ...`) that caused Avro-formatted sources to fail decoding records written with an older schema after a nullable column's type was promoted to a wider numeric type (`int` → `long`/`float`/`double`, `long` → `float`/`double`, or `float` → `double`). These are valid, non-breaking Avro promotions; they already worked for non-nullable columns and now work for nullable columns (unions) too.

### DIFF
--- a/src/avro/src/schema.rs
+++ b/src/avro/src/schema.rs
@@ -868,7 +868,7 @@ impl UnionSchema {
     /// writer piece. An exact match (by kind for anonymous pieces, by name for
     /// named ones) is preferred; otherwise the first reader variant in
     /// declaration order that `writer` can be promoted into is returned. See
-    /// [`can_promote`].
+    /// `can_promote`.
     pub fn match_ref_promote_writer(
         &self,
         writer: SchemaPieceRefOrNamed,
@@ -893,7 +893,7 @@ impl UnionSchema {
     /// promotion. `self` is the *writer* union and `reader` is a concrete
     /// reader piece. An exact match is preferred; otherwise the first writer
     /// variant in declaration order that can be promoted into `reader` is
-    /// returned. See [`can_promote`].
+    /// returned. See `can_promote`.
     pub fn match_ref_promote_reader(
         &self,
         reader: SchemaPieceRefOrNamed,
--- a/src/avro/src/schema.rs
+++ b/src/avro/src/schema.rs
@@ -851,6 +851,89 @@ impl UnionSchema {
     ) -> Option<(usize, &SchemaPieceOrNamed)> {
         self.match_ref(other.as_ref(), names_map)
     }
+
+    /// Like [`UnionSchema::match_`], but additionally honors Avro numeric
+    /// promotion. Here `self` is the *reader* union and `writer` is a writer
+    /// variant being resolved against it.
+    pub fn match_promote_writer(
+        &self,
+        writer: &SchemaPieceOrNamed,
+        names_map: &BTreeMap<usize, usize>,
+    ) -> Option<(usize, &SchemaPieceOrNamed)> {
+        self.match_ref_promote_writer(writer.as_ref(), names_map)
+    }
+
+    /// Like [`UnionSchema::match_ref`], but additionally honors Avro numeric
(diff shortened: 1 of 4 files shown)


tonydu-mz:

src/avro/src/schema.rs
```diff
@@ -851,6 +851,89 @@ impl UnionSchema {
...
+
+/// Returns `true` if a value written with primitive schema `writer` can be read
+/// as primitive `reader` under Avro's numeric promotion rules: `int` →
+/// `long`/`float`/`double`, `long` → `float`/`double`, and `float` → `double`.
+///
+/// These are exactly the promotions performed during schema resolution in
+/// `reader.rs` (`ResolveIntLong`, `ResolveIntDouble`, ...). The two must be
+/// kept in sync: only return `true` here for a pair that resolution can
+/// actually decode, otherwise a union variant would match but fail to resolve.
+fn can_promote(writer: &SchemaPiece, reader: &SchemaPiece) -> bool {
```
martykulma, line 926: Bit of a time-bomb because nothing enforces this function stays in sync with the source of truth (reader.rs).  We should have a tracking issue to address this (either refactoring or a unit test) as a fast-follow.

martykulma (approved):
Nice @tonydu-mz ! The functionality lgtm.  Given the need to address the issue, marking as approved, but we need to address the duplicated promotion rules that can now drift.

PR #37089, opened 2026-06-17
sql: omit DETAILS from SHOW CREATE TABLE ... FROM SOURCE

ggevay:
Fixes SQL-231 and fixes https://github.com/MaterializeInc/database-issues/issues/10034.

SHOW CREATE TABLE on a table created with CREATE TABLE ... FROM SOURCE emitted an internal DETAILS option that is not user-typeable, so the output could not be replayed (frustrating workload replay, mz-deploy, copy-paste, etc.). Strip DETAILS in humanize_sql_for_show_create, matching how the internal AS OF of materialized views is already hidden.

The omitted DETAILS remains accessible for debugging in the raw create_sql (and redacted_create_sql) columns of mz_catalog.mz_tables.

Nightly (because of the fiddly version guards in td): https://buildkite.com/materialize/nightly/builds/16809 Edit: Green (the SLT fail is unrelated, also present on `main`)

### DIFF
--- a/src/sql/src/plan/statement/show.rs
+++ b/src/sql/src/plan/statement/show.rs
@@ -1055,13 +1055,24 @@ impl<'a> ShowColumnsSelect<'a> {
 
 /// Convert a SQL statement into a form that could be used as input, as well as
 /// is more amenable to human consumption.
+///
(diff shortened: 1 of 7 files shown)


martykulma (approved):
lgtm - thanks @ggevay !

PR #37091, opened 2026-06-17
doc: design for unifying the storage and compute cluster protocol

antiguru:
### Motivation

Separate storage and compute timely runtimes give no total order over the two command streams, which is the central obstacle to ever unifying the runtimes — a single runtime must render dataflows in identical order on every worker, and order can only come from one totally-ordered command stream ([#34713](https://github.com/MaterializeInc/materialize/pull/34713) failed on exactly this).

Motivating instance: [SS-199](https://linear.app/materializeinc/issue/SS-199/elevated-freshness-in-mz-catalog-server-in-staging) (storage's 10s tick holds back compute introspection freshness). Runtime unification would categorically eliminate that hold-back; this design is the prerequisite, not the [SS-199](https://linear.app/materializeinc/issue/SS-199) fix.

### Description

Design doc only. Step-1 proposal: a union `ClusterCommand`/`ClusterResponse`, a delegating `PartitionedState` reusing the existing storage/compute routing and merge impls verbatim, and a single unified per-replica `ReplicaTask` that owns one CTP connection per cluster process — the single serialization point that manufactures the total order. The cluster side demultiplexes back to the existing separate runtimes, so behavior is unchanged and the order is dormant until a follow-up unifies the runtimes.

Also covers the `SequentialHydration` refactor (from `GenericClient` wrapper to a synchronous interceptor state machine, dropping its forwarder task and channels) and the reconnect path, which is a supervisor merge over the already-identical die-and-respawn model on both sides.

Open questions tracked in the doc: shared epoch, unified CTP versioning, reconcile ordering on respawn.

### Verification

Design doc; no code. Implementation lands as a separate PR stack referencing this doc.

### DIFF
--- a/doc/developer/design/20260617_unify_storage_compute_protocol.md
+++ b/doc/developer/design/20260617_unify_storage_compute_protocol.md
@@ -57,8 +57,7 @@ SS-199 itself; the fix lands with the runtime-unification follow-up.
   storage and compute runtimes.
   Runtime unification is the follow-up that the total order enables.
 * Extending sequential hydration to storage collections.
-  Storage would benefit from the same hardening, but it is compute-only today and
-  applying it to storage is a separate follow-up (see Alternatives).
+  Storage has no `Schedule`/suspend protocol; adding one is a separate feature.
 * Independent failure domains for storage and compute.
   We accept shared fate, consistent with the existing recommendation to keep
   storage and compute objects on separate replicas.
@@ -228,12 +227,12 @@ nothing yet.
 
 ### Sequential hydration for all objects, including storage
 
-Deferred, not rejected.
-Storage does not gate hydration today, but it would benefit from the same
-hardening; the only reason it is compute-only so far is that compute has the more
-complex setups where the limit matters most.
-Applying the mechanism to storage is a reasonable follow-up, but it is out of
-scope for step 1, which must stay behavior-preserving.
+Rejected.
+Storage broadcasts every command and self-coordinates ordering internally; it
+has no `Schedule`/suspend protocol.
+Gating storage through sequential hydration would require inventing a
+storage-side suspend mechanism — a separate feature with changed semantics, not
+step-1 plumbing.
 
 ### Keep `SequentialHydration` as a wrapper, generalized to `ClusterCommand`, in front of the unified task
 
@@ -261,16 +260,16 @@ intrinsically, with no separate component to keep consistent.
 
 ## Open questions
 
-* **Epoch unification.**
-  Compute carries a per-incarnation `epoch` to discard stale responses
-  (`replica.rs:148`); storage keys responses by `ReplicaId` only.
-  The unified path needs the storage response path to honor the shared epoch so
-  a respawn can distinguish stale storage responses.
-  Confirm there is no other storage-side assumption that the absence of an epoch
-  encodes.
+None remaining; the items raised in review are resolved below.
 
 Resolved during review:
 
+* **Epoch unification** — storage adopts the same per-incarnation `epoch` compute
+  already carries to discard stale responses (`replica.rs:148`). Storage keys
+  responses by `ReplicaId` only today; the epoch is a beneficial hardening it
+  lacks (compute got it first because its setups are more complex), and the
+  unified respawn path needs it regardless. Add it to the storage response path
+  as part of unification.
 * **Protocol versioning** — a single shared CTP `version` is acceptable; storage
   and compute are already built and rolled out together.
 * **Reconcile ordering on respawn** — reconciliation is order-independent; there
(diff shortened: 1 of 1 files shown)


doc/developer/design/20260617_unify_storage_compute_protocol.md
```diff
@@ -0,0 +1,258 @@
...
+channel and `client.recv()`.
+The unified task replaces both:
+
+```text
+run():
+  state = hydration state machine (fresh per incarnation)
+  client = connect()                       // one union CTP connection per process
+  loop select:
+    cmd from storage_cmd_rx  -> specialize -> client.send(Storage(cmd))
+    cmd from compute_cmd_rx  -> specialize -> for c in state.absorb_command(cmd):
```
antiguru: Check that the rx's are cancellation safe.

doc/developer/design/20260617_unify_storage_compute_protocol.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Open questions
+
+* **Epoch unification.**
+  Compute carries a per-incarnation `epoch` to discard stale responses
+  (`replica.rs:148`); storage keys responses by `ReplicaId` only.
+  The unified path needs the storage response path to honor the shared epoch so
+  a respawn can distinguish stale storage responses.
+  Confirm there is no other storage-side assumption that the absence of an epoch
+  encodes.
```
antiguru: We can do the same for storage; the reason it doesn't do it is that we have more data with complex setups for compute, but storage would benefit if we applied the same hardening.

doc/developer/design/20260617_unify_storage_compute_protocol.md
```diff
@@ -0,0 +1,258 @@
...
+* **Protocol versioning.**
+  CTP `connect` takes a single `version`.
+  Unifying the connection means storage and compute are versioned together from
+  now on; confirm this does not break independent rollout assumptions.
```
antiguru: Same version is alright.

doc/developer/design/20260617_unify_storage_compute_protocol.md
```diff
@@ -0,0 +1,258 @@
...
+* **Reconcile ordering on respawn.**
+  Both controllers replay into the fresh task; confirm there is no implicit
+  dependency on storage reconciling before compute (or vice versa) that the
+  separate-connection model accidentally provided.
```
antiguru: Reconciliation should be order-independent, as long as the protocol invariants aren't violated. There is no dependence between storage and compute, so the order in which they reconcile is irrelevant. Slight preference to reconcile compute first, if you have to make a choice.

doc/developer/design/20260617_unify_storage_compute_protocol.md
antiguru: Added a cancel-safety note: both command `mpsc` receivers use `recv` (documented cancel safe) and `GenericClient::recv` is required cancel safe — the existing per-subsystem loops already rely on this. The `client.send` calls run inside the selected branch to completion. Flagged for re-verification against the union `Partitioned` and the hydration state machine at impl time.

doc/developer/design/20260617_unify_storage_compute_protocol.md
antiguru: Reframed from "rejected" to "deferred, not rejected": storage would benefit from the same hardening, compute-only today only because its setups are more complex. Updated both the Alternatives entry and the Out of Scope bullet.

doc/developer/design/20260617_unify_storage_compute_protocol.md
antiguru: Resolved: noted that a shared CTP `version` is acceptable since storage and compute are built and rolled out together. Moved out of Open questions into the design.

doc/developer/design/20260617_unify_storage_compute_protocol.md
antiguru: Resolved: documented reconciliation as order-independent (no storage/compute dependence), with a slight preference to reconcile compute first when a choice is forced.

doc/developer/design/20260617_unify_storage_compute_protocol.md
antiguru: Correction — I misread this earlier as being about sequential hydration. It's on epoch unification: storage adopts the same per-incarnation `epoch` compute already carries to discard stale responses — beneficial hardening storage lacks today (compute got it first due to more complex setups), and the unified respawn path needs it anyway. Resolved that in the doc and reverted my mistaken change to the sequential-hydration alternative (restored to rejected).

martykulma:
The plan seems reasonable, but Phase 3 changes to storage feel under specified.  It doesn't seem that this doc has intention to define all the way through phase 3, though.  I think some of the phase 3 work in storage could be handled before the runtime merge (e.g. error handling).

PR #37092, opened 2026-06-17
round-trip `SHOW CREATE TABLE` references for SQL Server and load generators

ggevay:
Fixes https://github.com/MaterializeInc/database-issues/issues/10010, a `SHOW CREATE` roundtripping issue.

SHOW CREATE TABLE on a `CREATE TABLE ... FROM SOURCE` prints a fully-qualified
3-part REFERENCE (database.schema.name) for SQL Server (the real upstream
database) and load generator (the synthetic `mz_load_generators`) sources.
Replaying that exact statement failed with "reference ... not found in source",
because the SourceReferenceResolver for those types was built with the
placeholder DATABASE_FAKE_NAME ("database") rather than the database actually
embedded in the stored reference. Only Postgres built its resolver with the real
database, so only Postgres round-tripped; MySQL and Kafka store no database
component in their reference, so they were unaffected.

Build the resolver with the database each source type stores in its external
reference: the real upstream database for SQL Server, `mz_load_generators` for
load generators. The resolver's database is only used for matching (resolve_idx)
and is never propagated, so stored/printed references are unchanged and no
catalog migration is needed.

Fix A vs Fix B: the alternative ("Fix B") was to drop the database from the
stored reference so it becomes 2-part like MySQL -- which is what the issue
reporter expected for load generators. We chose this approach ("Fix A") because
Fix B requires a catalog migration to rewrite existing tables' create_sql, which
is substantial hassle for little benefit, especially as load generator sources
may be deprecated soon. Fix A also keeps SQL Server consistent with Postgres,
where the database in the reference is genuinely meaningful.

Also rewrites the stale DATABASE_FAKE_NAME doc comment, which predated the SQL
Server source and wrongly claimed only Postgres stores a database in its
reference.

### DIFF
--- a/src/sql/src/pure/references.rs
+++ b/src/sql/src/pure/references.rs
@@ -16,7 +16,9 @@ use mz_repr::RelationDesc;
 use mz_sql_parser::ast::{ExternalReferences, Ident, IdentError, UnresolvedItemName};
 use mz_sql_server_util::SqlServerError;
 use mz_sql_server_util::desc::SqlServerTableRaw;
-use mz_storage_types::sources::load_generator::{LoadGenerator, LoadGeneratorOutput};
+use mz_storage_types::sources::load_generator::{
+    LOAD_GENERATOR_DATABASE_NAME, LoadGenerator, LoadGeneratorOutput,
+};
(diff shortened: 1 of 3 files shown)


martykulma (approved):
thanks @ggevay !!

🦈

PR #37094, opened 2026-06-17
compute: refactor SequentialHydration into a synchronous interceptor

antiguru:
### Motivation

Prepares for unifying the storage and compute replica tasks (design #37091).
Also a standalone simplification.

### Description

`SequentialHydration` was a `GenericClient` wrapper that owned the wrapped client and ran a forwarder task, with two mpsc channels existing solely to make its `recv` cancel safe while it internally sent commands to the wrapped client.

This turns it into a synchronous interceptor: `absorb_command` and `observe_response` now return the commands the replica task should send, and the replica task owns the send.
This drops the forwarder task, both mpsc channels, and the `GenericClient` impl.
Cancel safety is now provided by the replica task's `select!` loop — exactly as for the storage replica task, which already awaits `client.send` directly in its command arm.

Behavior is unchanged: the same `Schedule`-gating logic runs, only the plumbing around it changes.

### Verification

`cargo check`, `cargo clippy --all-targets`, and `cargo fmt` on `mz-compute-client`.
Behavior-preserving refactor; existing compute/cluster integration tests in CI exercise the hydration path.

### DIFF
--- a/src/compute-client/src/controller/sequential_hydration.rs
+++ b/src/compute-client/src/controller/sequential_hydration.rs
@@ -25,7 +25,7 @@
 //! implementation. If the assumption ever ceases to hold, we will need to adjust the code in this
 //! module.
 //!
-//! Sequential hydration is enforeced by a `SequentialHydration` client that sits between the
+//! Sequential hydration is enforced by a `SequentialHydration` interceptor that sits between the
 //! controller and the `PartitionedState` client that splits commands across replica processes.
 //! This location is important:
 //!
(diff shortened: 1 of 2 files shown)


antiguru:
Still valid after the design reframe (#37091). This refactor is orthogonal and independently landable; in the phased plan it slots into Phase 2 (the unified replica task pumps the interceptor), but it stands on its own as a simplification regardless.

src/compute-client/src/controller/sequential_hydration.rs
```diff
@@ -121,9 +108,9 @@ impl SequentialHydration {
         Arc::strong_count(&self.hydration_token) - 1
     }
 
-    /// Absorb a command and send resulting commands to the wrapped client.
-    fn absorb_command(&mut self, cmd: ComputeCommand) -> Result<(), anyhow::Error> {
-        // Whether to forward this command to the wrapped client.
+    /// Absorb a command the task intends to send, returning the commands it should actually send.
+    pub fn absorb_command(&mut self, cmd: ComputeCommand) -> Vec<ComputeCommand> {
```
martykulma: `SequentialHydration` is `pub(crate)` so this is effectively the same, I assume that's expected (and maybe pub here is just a shorthand?).

martykulma (approved):
lgtm!

src/compute-client/src/controller/sequential_hydration.rs
antiguru: Done; changed all to pub(super).

PR #37095, opened 2026-06-17
compute: add per-operator frontier and summary introspection relations

antiguru:
### Motivation

The only per-operator progress signal today is `mz_compute_operator_hydration_statuses`, which records a single `false`→`true` transition when an operator's output frontier first passes the dataflow's `as_of`. It never surfaces ongoing frontier movement, so it can't answer "which operator is holding this dataflow's frontier back?". The export-level `mz_compute_frontiers` is too coarse (one frontier per object), and the raw reachability logs only carry capabilities for the operators *currently pinning* a frontier — a sparse subset — without the operator connectivity needed to interpret them.

This is the first step toward replacing the bespoke `log_operator_hydration` dataflow operators with a query over introspection data, moving cost from dataflow run time to introspection query time.

### Description

Adds two introspection capabilities.

**Operator summaries.** Consume timely 0.30's `timely/summary/{type}` streams (`OperatesSummaryEvent`), which carry each operator's internal input→output path summaries (`Connectivity`). Each path-summary element is projected onto the outer timestamp and exposed via `mz_introspection.mz_dataflow_operator_summaries[_raw|_per_worker]` with columns `(id, worker_id, input_port, output_port, outer_impact, summary)`. `outer_impact` is the summary's effect on the outer timestamp — uniformly `0` in compute dataflows, since every path summary (including scope ingress/egress and a WMR feedback `+1`) acts only on the inner iteration coordinate. `summary` is a debug rendering of the element, so cyclic/feedback operators are identifiable.

Because the summaries arrive on a separate, timestamp-typed logger stream but describe per-operator state, the per-type loggers project them into shared state and the timely demux emits and retracts them on operator shutdown, exactly alongside `operates`/`addresses`. This keeps them on the same lifecycle and avoids leaking rows across dataflow churn.

**Reconstructed per-operator frontiers.** `mz_introspection.mz_dataflow_operator_frontiers[_per_worker]` is a recursive view that forward-propagates held capabilities (from reachability) along the operator graph (`mz_dataflow_channel_operators`). Since outer-time path summaries are identity, each edge transfers the frontier unchanged and a cyclic (WMR) region resolves to a single shared frontier. The result is an output frontier for *every* operator — not just the sparse set currently holding capabilities — which is exactly timely's outer frontier by construction.

### Verification

Validated manually against `mz_compute_frontiers` on a single-worker replica:

* For a non-iterative index and a `WITH MUTUALLY RECURSIVE` materialized view, the reconstructed sink-operator frontier matches the export frontier exactly, and every operator (including those inside the iterative scope) receives a frontier.
* With the upstream cluster scaled to zero replicas, the downstream object's frontier stalls; the reconstruction tracks the stalled frontier and attributes it to the stalled persist-import operators.
* Summary-log retraction: at steady state, the count of summary rows whose operator is no longer live is `0` across repeated create/drop cycles; transient nonzero values during churn reflect cross-arrangement skew, not leaked rows.

Draft: no automated tests yet, and `as_of` exposure plus retiring `log_operator_hydration` are follow-ups.

This release will add the `mz_dataflow_operator_summaries` and `mz_dataflow_operator_frontiers` introspection relations, exposing per-operator internal connectivity summaries and reconstructed per-operator output frontiers.


src/compute/src/logging/initialize.rs
```diff
@@ -317,6 +338,75 @@ impl LoggingContext<'_> {
...
+    /// time) carrying the operator's internal connectivity (`Connectivity<T::Summary>`,
+    /// indexed by input port). It projects each path-summary element onto the outer
+    /// timestamp dimension and pushes the rows into shared state, where the timely demux
+    /// picks them up — the demux owns operator lifecycle and retracts the rows on shutdown.
+    fn summary_logger<T>(&self) -> Logger<TimelySummaryEventBuilder<T::Summary>>
+    where
+        T: timely::progress::Timestamp,
+        T::Summary: ExtractSummary + std::fmt::Debug,
+    {
+        let shared_state = Rc::clone(&self.shared_state);
```
antiguru: Why do we use shared state and not a event queue?

antiguru: Why do we use shared state and not a event queue?

antiguru: Good call — shared state was the wrong instinct. The reason I avoided a standalone event-queue + `construct` (like reachability/differential) is retraction: `OperatesSummaryEvent` is logged once at operator build with no shutdown counterpart, so the summary stream is `+1`-only. A dedicated construct has no operator-shutdown signal, and semijoining against the retracting `addresses` collection would still leak the `+1`-only raw arrangement. The only place the shutdown signal lives is the timely demux.

Switching to: keep the `EventQueue` (drop shared state), and feed its replayed stream into the timely demux as a **second input**, so summaries ride the same operator lifecycle that already emits/retracts `operates`/`addresses`. That restores the conventional event-link structure and lets the demux retract summaries on `Shutdown`. Will push shortly along with the other review fixes.

antiguru: Done in 80a92c3bfb. Summaries now use the conventional event-queue + replay structure (no shared state); the replayed stream is a second input to the timely demux, which emits on build and retracts on shutdown alongside `operates`/`addresses`. Re-verified locally: frontier reconstruction still matches `mz_compute_frontiers` exactly, and orphaned summary rows settle to 0 after drops (cleaner than before — no transient skew).

Also addressed from the review pass:
- **Blocking:** BUILTINS ordering (frontier views now follow the channel-operator/reachability views they depend on); summary retraction moved before the missing-operator early-return in `handle_shutdown`.
- **Should-fix:** `outer_impact = 0` reliance now documented as an explicit assumption in the frontier view (folding it in needs port-level propagation — follow-up); `shutdown_before_summary` documented as bounded (Timely logs exactly one summary per operator, so entries always clear).
- **Deferred:** sqllogictest regeneration (`oid.slt`, `mz_introspection.slt`, etc.) before un-drafting.

PR #37096, opened 2026-06-17
avro: Bound array/map decode by Value-node count, not element count

def-:
The `reader_decode` cargo-fuzz target OOMed (peak ~4.3 GiB from a 175-byte input) on an `array<record{null}>`. Each element is zero-width on the wire, so the `min_encoded_len` byte-floor check can't bound the block — yet every element still allocates a `Value::Record` (a `Vec` slot, the record's own `Vec`, and a field-name `String`). The cumulative cap was on the element *count* (`MAX_BLOCK_ELEMENTS = 1 << 24`), so a single block claiming ~16M zero-width records amplified a handful of bytes into gigabytes.

Cap the number of decoded `Value` nodes instead: weight each element by `min_value_nodes`, a lower bound on the nodes it materializes (counted exactly for the amplifying case — `null` and records of only zero-width fields), and bound the weighted total by `MAX_VALUE_NODES = 1 << 22`. This also covers the analogous case of a huge block of wide, positive-floor records read from a large input, and applies symmetrically to maps (which previously had no cumulative cap at all). Worst case now peaks ~750 MiB, comfortably under the fuzzer's 2 GiB RSS limit, while still admitting any realistically-sized array/map.

Verified: the artifact peaks ~760 MiB and a 90s / 1.4M-run fuzz session finds no new OOM. Regression tests cover both the reject and the below-cap-still-decodes paths.

### DIFF
--- a/src/avro/src/decode.rs
+++ b/src/avro/src/decode.rs
@@ -279,11 +279,11 @@ mod tests {
     fn zero_width_array_total_len_bounded_across_blocks() {
         // A zero-width element type has no input-proportional per-block bound.
         // Keep a cumulative cap so repeated legal-size blocks cannot drive an
-        // unbounded decode. Seed the array access at the cap to test the edge
-        // without walking 16M null elements first.
+        // unbounded decode. Seed the shared node budget at the cap to test the
+        // edge without walking millions of null elements first.
         use std::str::FromStr;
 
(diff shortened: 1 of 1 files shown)


martykulma (approved):
nice catch, thanks @def- !

PR #37104, opened 2026-06-17
expr: keep `%substring%` LIKE on the fast string matcher

def-:
#37047 rerouted any LIKE pattern with more than one `%` (`many`) subpattern to the linear regex engine to kill a fuzzer-found super-linear backtracking blowup (`%a%a%a` against a long run of `a`s costs `len(text)^(#%)`). That fix is correct, but the routing predicate counts every `many` subpattern, including a trailing `%` whose suffix is empty.

A `%` with an empty suffix short-circuits in `is_match_subpatterns` with no `rfind`/backtracking, and per `build_subpatterns` such a `%` can only ever be the trailing subpattern, so it never contributes to the blowup. The canonical "contains" pattern `%foo%` decomposes into one non-empty-suffix `%` plus a trailing empty-suffix `%`, so it was counted as 2 > 1 and rerouted to regex — even though it runs as a single `text.rfind("foo")` in the string matcher. `%foo%` is one of the most common LIKE shapes and `IsLikeMatch` is evaluated per row, and the regex path is 2.4x-9x slower (and far worse on long values whose match sits near the end, where `rfind` scans from the end while the regex scans forward).

Count only `%` subpatterns with a non-empty suffix, since those are the ones that nest the backtracking loops. With at most one such `%` there is at most one `rfind` loop and matching stays linear in the text length — the property the original fix wants — while `%foo%`, `%foo`, `foo%`, etc. stay on the fast string matcher. The genuinely quadratic shapes (`%foo%bar%`, `%foo%bar`, `a%b%c`, `%a%a%a`) still route to the regex engine.

Adds `test_string_matcher_routing`, which asserts the `MatcherImpl` variant chosen for each shape so a future tweak to the predicate cannot silently de-optimize the common case. This is a separate test from the existing correctness tests because the regression produces no wrong results — it is invisible to any assertion over match outcomes.

### DIFF
--- a/src/expr/src/scalar/like_pattern.rs
+++ b/src/expr/src/scalar/like_pattern.rs
@@ -156,10 +156,22 @@ pub fn compile(pattern: &str, case_insensitive: bool) -> Result<Matcher, EvalErr
     // like `%a%a%a` against a long run of `a`s takes time proportional to
     // `len(text)^(number of %)`, which can stall a worker for many minutes. The
     // regex engine matches the same patterns in linear time with no backtracking,
-    // so fall back to it whenever more than one `many` subpattern is present (and
+    // so fall back to it whenever more than one backtracking `%` is present (and
     // for the existing case-insensitive / too-many-subpatterns reasons).
-    let many_subpatterns = subpatterns.iter().filter(|s| s.many).count();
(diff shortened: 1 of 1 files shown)


antiguru (approved):
Looks good!

PR #37106, opened 2026-06-17
sql: omit derived columns from SHOW CREATE TABLE ... FROM SOURCE

ggevay:
Fixes https://linear.app/materializeinc/issue/SQL-392

SHOW CREATE TABLE on a table created with CREATE TABLE ... FROM SOURCE emitted the column definitions and constraints that purification populates from the upstream schema (the `Defined` column variant). CREATE TABLE ... FROM SOURCE rejects those on input ("column definitions cannot be specified directly"), so the printed statement did not round-trip. This affected Postgres, MySQL, SQL Server, and multi-output load generator sources.

Strip the derived columns and constraints in humanize_sql_for_show_create, matching how the internal AS OF (materialized views) and DETAILS (this statement) are already hidden. The schema-affecting TEXT COLUMNS / EXCLUDE COLUMNS options are retained, so purification re-derives the same schema on replay. A user-typed `Named` column list is left intact, since it does round-trip.

**Strip vs accept**: the alternative was to make CREATE TABLE ... FROM SOURCE tolerate column definitions on input, which would keep them visible in SHOW CREATE. We strip instead because it is simpler (no validation of input columns against the derived schema, no schema-drift questions) and consistent with the AS OF / DETAILS precedent. Stripping loses no information: the column schema remains available via SHOW COLUMNS, mz_columns, and \d, and the full original (columns and constraints) is preserved verbatim in the raw create_sql / redacted_create_sql columns of mz_catalog.mz_tables.

Nightly: https://buildkite.com/materialize/nightly/builds/16815 (unrelated redness)

### DIFF
--- a/src/sql/src/plan/statement/show.rs
+++ b/src/sql/src/plan/statement/show.rs
@@ -1070,7 +1070,7 @@ fn humanize_sql_for_show_create(
     redacted: bool,
 ) -> Result<String, PlanError> {
     use mz_sql_parser::ast::{
-        CreateSourceConnection, MySqlConfigOptionName, PgConfigOptionName,
+        CreateSourceConnection, MySqlConfigOptionName, PgConfigOptionName, TableFromSourceColumns,
         TableFromSourceOptionName,
     };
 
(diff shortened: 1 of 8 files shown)


martykulma (approved):
thanks @ggevay!
🐟

PR #37107, opened 2026-06-17
catalog: add durable cluster-autoscaling state (migration v86->v87)

aljoscha:
**NOTE: This change looks bigger than it is because of the big durable state objects changes that are mostly copies. Please don't despair.**

**NOTE: This is PR1 in a stack of PRs, the full branch is at https://github.com/MaterializeInc/materialize/pull/36738**

Add the additive, behaviourally-inert durable state the cluster controller needs, with one catalog migration (v85->v86) defaulting it for existing clusters. No new field drives behaviour yet, so this is dark, save for one deliberate user-visible change: the public `mz_cluster_replicas.availability_zone` column now reports every zone a replica is provisioned under (details below).

On the managed cluster config we add three `Option` fields, all `None` by default: `auto_scaling_strategy` (user policy) and the in-flight runtime records `reconfiguration` and `burst`.

On the managed replica location we collapse the single `availability_zone` user pin into an `availability_zones: Vec<String>` recording the zones the replica was provisioned under. For a managed cluster that is the cluster's AVAILABILITY ZONES pool. For an unmanaged replica it is the single pin as a zero- or one-element list. The controller needs this durable to tell realized-shape from target-shape replicas by config shape, including an AVAILABILITY ZONES divergence. A single-AZ field plus a separate provisioned-list field would be mutually exclusive and both collapse to a list at the orchestrator, so one list is the honest shape. The migration backfills managed-cluster replicas from their cluster's `availability_zones` and unmanaged-cluster replicas from their pin.

With the durable record now storing the list unconditionally, the in-memory `ManagedReplicaAvailabilityZones` enum loses its reason to exist. It distinguished a managed cluster's pool (`FromCluster`) from an unmanaged cluster's single pin (`FromReplica`), but both already collapse to a list of acceptable zones at the orchestrator and the distinction is recoverable from whether the owning cluster is managed. We replace it on `ManagedReplicaLocation` with a bare `Vec<String>`, where an empty list means no AZ constraint, so the in-memory to durable `From` becomes a passthrough.

Concretization stays behaviour-preserving. On create and reshape it re-derives the managed pool from the cluster's current config, because the caller passes the cluster's AVAILABILITY ZONES as the override. On a rebuild from durable state it keeps the stored list, which is the pool the replica was provisioned under and equals the current pool unless a graceful reconfiguration is in flight. For an unmanaged replica it keeps the single pin. The orchestrator maps an empty list to "no constraint".

The public `mz_cluster_replicas.availability_zone` column changes behaviour. The durable collapse leaves the old view SQL reaching for an `availability_zone` JSON key that no longer exists, which would silently NULL the column for every replica, so the view has to change regardless. Rather than reconstruct the old single-pin behaviour, we now read the `availability_zones` list and report every zone the replica is provisioned under, comma-separated, for both managed and unmanaged clusters. It is `NULL` only when nothing constrains the replica's placement. This surfaces the full provisioned pool instead of hiding it behind a single zone, which is the more useful view and what an operator actually wants to see. The column name stays singular to avoid breaking existing queries.

This also adds the in-memory catalog mirror for the new durable state (`ReconfigurationState`, `ReconfigurationTarget`, `BurstState`, and the new fields on the in-memory `ClusterVariantManaged`, with conversions to and from the durable types) and the `AutoScalingStrategy` and `OnHydration` plan types the SQL surface will populate.

Implements the durable state model from
`doc/developer/design/20260522_cluster_autoscaling.md`


def- (changes requested):
In mz_catalog.rs:2400 we still use availability_zone:
```
data->'value'->'config'->'location'->'Managed'->>'availability_zone' AS availability_zone,
```
so the column will just be `NULL`. Small test to prove it:
```diff
diff --git a/test/cluster/mzcompose.py b/test/cluster/mzcompose.py
index f5076a57d9..e50b81c9e7 100644
--- a/test/cluster/mzcompose.py
+++ b/test/cluster/mzcompose.py
@@ -1455,6 +1455,43 @@ def workflow_test_bootstrap_vars(c: Composition) -> None:
         c.run_testdrive_files("resources/bootstrapped-system-vars.td")


+def workflow_test_replica_availability_zone_catalog(c: Composition) -> None:
+    """Regression test for the public mz_cluster_replicas.availability_zone
+    catalog column.
+
+    That column is a BuiltinMaterializedView that reaches into the durable
+    ClusterReplica JSON by the literal key `availability_zone`. Renaming that
+    durable Managed-location field (e.g. collapsing it into an
+    `availability_zones` list) without updating mz_catalog.rs silently NULLs the
+    column for every replica. test/sqllogictest/singlereplica_mz_cluster_replicas.slt
+    cannot catch this -- the sqllogictest environment has no availability zones
+    configured, so any user-specified AVAILABILITY ZONE is rejected and the
+    column never goes non-NULL there. Boot environmentd with a configured
+    availability-zone pool so the pin is accepted."""
+
+    materialized = Materialized(
+        options=[
+            "--availability-zone=us-east-1a",
+            "--availability-zone=us-east-1b",
+            "--availability-zone=us-east-1c",
+        ],
+        additional_system_parameter_defaults={
+            "enable_managed_cluster_availability_zones": "true",
+        },
+    )
+
+    with c.override(materialized, Testdrive()):
+        c.up("materialized")
+        c.run_testdrive_files("resources/replica-availability-zone.td")
+        c.kill("materialized")
+
+    # Restart and re-assert: the column is recomputed from durable state on
+    # boot, exercising the rebuild-from-durable concretization path too.
+    with c.override(materialized, Testdrive(no_reset=True)):
+        c.up("materialized")
+        c.run_testdrive_files("resources/replica-availability-zone-after-restart.td")
+
+
 def workflow_test_system_table_indexes(c: Composition) -> None:
     """Test system table indexes."""
```

aljoscha:
> In mz_catalog.rs:2400 we still use availability_zone:
> 
> ```
> data->'value'->'config'->'location'->'Managed'->>'availability_zone' AS availability_zone,
> ```
> 
> so the column will just be `NULL`. Small test to prove it:
> 
> ```diff
> diff --git a/test/cluster/mzcompose.py b/test/cluster/mzcompose.py
> index f5076a57d9..e50b81c9e7 100644
> --- a/test/cluster/mzcompose.py
> +++ b/test/cluster/mzcompose.py
> @@ -1455,6 +1455,43 @@ def workflow_test_bootstrap_vars(c: Composition) -> None:
>          c.run_testdrive_files("resources/bootstrapped-system-vars.td")
> 
> 
> +def workflow_test_replica_availability_zone_catalog(c: Composition) -> None:
> +    """Regression test for the public mz_cluster_replicas.availability_zone
> +    catalog column.
> +
> +    That column is a BuiltinMaterializedView that reaches into the durable
> +    ClusterReplica JSON by the literal key `availability_zone`. Renaming that
> +    durable Managed-location field (e.g. collapsing it into an
> +    `availability_zones` list) without updating mz_catalog.rs silently NULLs the
> +    column for every replica. test/sqllogictest/singlereplica_mz_cluster_replicas.slt
> +    cannot catch this -- the sqllogictest environment has no availability zones
> +    configured, so any user-specified AVAILABILITY ZONE is rejected and the
> +    column never goes non-NULL there. Boot environmentd with a configured
> +    availability-zone pool so the pin is accepted."""
> +
> +    materialized = Materialized(
> +        options=[
> +            "--availability-zone=us-east-1a",
> +            "--availability-zone=us-east-1b",
> +            "--availability-zone=us-east-1c",
> +        ],
> +        additional_system_parameter_defaults={
> +            "enable_managed_cluster_availability_zones": "true",
> +        },
> +    )
> +
> +    with c.override(materialized, Testdrive()):
> +        c.up("materialized")
> +        c.run_testdrive_files("resources/replica-availability-zone.td")
> +        c.kill("materialized")
> +
> +    # Restart and re-assert: the column is recomputed from durable state on
> +    # boot, exercising the rebuild-from-durable concretization path too.
> +    with c.override(materialized, Testdrive(no_reset=True)):
> +        c.up("materialized")
> +        c.run_testdrive_files("resources/replica-availability-zone-after-restart.td")
> +
> +
>  def workflow_test_system_table_indexes(c: Composition) -> None:
>      """Test system table indexes."""
> ```

Thank you for catching this! 🙇‍♂️

src/adapter/src/catalog/state.rs
```diff
@@ -2433,17 +2432,16 @@ impl CatalogState {
...
+                // paths pass the cluster's pool as `allowed_availability_zones`
+                // to stamp it, while rebuilds from durable state pass `None` to
+                // keep it. For an unmanaged cluster's replica it is the single
+                // user-pinned AZ.
+                availability_zones,
                 billed_as,
                 internal,
                 pending,
             } => {
-                if allowed_availability_zones.is_some() && availability_zone.is_some() {
```
mtabebe, line 2441: Why was this removed, should there be an assertion on the invariant around the availability_zone?

mtabebe (commented):
Also flagging the builtin issue: 
```
data->'value'->'config'->'location'->'Managed'->>'availability_zone' AS availability_zone
```

In general this reads very well and is nicely structured. I like how the catalog migration is a straight forward addition, and there are good unit tests as well. 

I think once you fix the built in, it should be good

src/adapter/src/catalog/state.rs
aljoscha: This is because before this change replicas of a managed cluster didn't have the AZs durable in their own state. With the change, we now store the AZ restriction on _all_ replicas, so the assertion can't hold anymore.

Is that an okay justification or should I expand?

aljoscha:
Updated and rebased on latest `main`, this has the potential to rot quickly ... 😬

def- (approved):

src/adapter/src/catalog/state.rs
mtabebe: Makes sense to me

mtabebe (approved):

PR #37109, opened 2026-06-17
SS-157 Fix jsonb serialization roundtrip panic with NaN numerics

patrickwwbutler:
Fixes https://linear.app/materializeinc/issue/SS-157/json-sink-encoder-thread-timelywork-0-panicked-at by adding handling to the `JsonbDatum` serialization specifically for non-finite numerics (NaN, and ±∞)

### DIFF
--- a/test/testdrive/jsonb-nan.td
+++ b/test/testdrive/jsonb-nan.td
@@ -32,7 +32,7 @@ $ set-arg-default default-storage-size=scale=1,workers=1
   SELECT '{"$serde_json::private::Number": "NaN"}'::jsonb AS j;
 
 > SELECT j::text FROM nan_jsonb_view;
-NaN
+"\"NaN\""
 
 > CREATE CLUSTER nan_sink_cluster SIZE '${arg.default-storage-size}';
 
--- a/test/testdrive/jsonb-nan.td
+++ b/test/testdrive/jsonb-nan.td
@@ -1,27 +1,11 @@
-# Reproducer: NaN-bearing JSONB crashes the JSON sink encoder.
+# Copyright Materialize, Inc. and contributors. All rights reserved.
 #
-# `strconv::parse_numeric` accepts positive quiet NaN. `Jsonb::from_str`
-# recognises the magic `$serde_json::private::Number` map key and calls
-# `parse_numeric` on the payload (see src/repr/src/adt/jsonb.rs), so a
-# user-supplied JSON text with that magic key smuggles `Datum::Numeric(NaN)`
-# past the `cast_jsonbable_to_jsonb` defence
-# (src/expr/src/scalar/func/impls/jsonb.rs:159) that the rest of the system
-# relies on.
+# Use of this software is governed by the Business Source License
+# included in the LICENSE file at the root of this repository.
 #
-# Once stored, sinks reach this datum via `JsonbRef::to_serde_json`
-# (src/repr/src/adt/jsonb.rs:213), which calls `serde_json::to_value(...)`
-# on a writer that produces the literal token `NaN`. `serde_json::to_value`
-# rejects that token as "invalid number"; the `.expect("conversion to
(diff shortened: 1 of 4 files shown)


test/testdrive/jsonb-nan.td
```diff
@@ -0,0 +1,31 @@
...
+$ set-arg-default default-storage-size=scale=1,workers=1
+
+> CREATE CONNECTION kafka_conn
+  TO KAFKA (BROKER '${testdrive.kafka-addr}', SECURITY PROTOCOL PLAINTEXT);
+
+> CREATE MATERIALIZED VIEW nan_jsonb_view AS
+  SELECT '{"$serde_json::private::Number": "NaN"}'::jsonb AS j;
+
+> SELECT j::text FROM nan_jsonb_view;
+"\"NaN\""
```
peterdukelarsen, line 19: Out of curiosity, is it expected that we would treat `'{"$serde_json::private::Number": "NaN"}'` as a number? Not sure if that's besides the point and there are other ways to trigger this issue.

test/testdrive/jsonb-nan.td
```diff
@@ -0,0 +1,31 @@
...
+
+> CREATE SINK nan_sink
+  IN CLUSTER nan_sink_cluster
+  FROM nan_jsonb_view
+  INTO KAFKA CONNECTION kafka_conn (TOPIC 'testdrive-jsonb-nan-${testdrive.seed}')
+  FORMAT JSON
+  ENVELOPE DEBEZIUM
+
+$ kafka-verify-data format=json sink=materialize.public.nan_sink key=false
+{"before":null,"after":{"j":"NaN"}}
```
peterdukelarsen, line 31: Is this the first place we've exposed "NaN" as part of the output to a sink? I don't think json really supports NaN in a native way, and I wonder if it would be possible/better to error rather than spit out the NaN?

peterdukelarsen (commented):
A couple of questions about desired behavior

test/testdrive/jsonb-nan.td
patrickwwbutler: Yes, as I understand it - that `$serde_json::private::Number` is supposed to indicate that the following string should be parsed as an arbitrary precision number:
https://github.com/MaterializeInc/materialize/blob/ae8f529217299aa81c014f15c75cf92761d1ce6f/src/repr/src/adt/jsonb.rs#L282

test/testdrive/jsonb-nan.td
patrickwwbutler: cc @maheshwarip on ideal behavior with NaNs in json. I'm not sure if we've expose NaN as output to sinks before? I would assume so?

I suppose there would be two error options - output this specific record with an error message, or set the whole sink into an errored state.

Personally I think this is definitely better than setting the sink into an errored state, that takes a fair amount of agency away from customers who might expect some NaNs and are okay with that scenario.

As for the other option, I think sinking the "NaN" value like is done here is basically topologically equivalent to outputting an error message for the specific row, IMO NaN is a pretty clear explanation that some "illegal" math operation was attempted here, but I suppose there could be some value in being more specific about why, but I think that becomes a different problem than just fixing the jsonb plumbing.

maheshwarip: If I'm understanding correctly, the question is: _should I error the sink or error an individual record_?

IMO the answer should be error the individual record. Otherwise as a user, you have a very brittle sink that is liable to easy failure!

As for what the error should look like, I'm not sure. Can we go with consistency with whatever we do elsewhere? From first principles, a clear error message + NaN / null would be my pick?

patrickwwbutler: It looks like we already do error out when someone does an illegal math operation in materialize, so this would really mostly apply to when a `NaN` comes in from a source. In which case I think just plumbing it along is probably the right move?

PR #37110, opened 2026-06-17
Replace ToDatumIter with a push-based ExtendDatums trait

frankmcsherry:
## What

Replaces the GAT-based `ToDatumIter` trait ("hand me a borrowing `Datum` iterator") with a push-based `ExtendDatums` trait, and removes the remaining places that iterated a trace row (`DatumSeq`) directly.

- `ToDatumIter` → `ExtendDatums`, a single method:
  `extend_datums(&self, arena: &RowArena, target: &mut Vec<Datum>, max: Option<usize>)`.
  Drops the `DatumIter<'a>` GAT, the `to_datum_iter` method, and the `Sized` supertrait.
- The `&RowArena` parameter lets an implementor decode into caller-provided storage and return `Datum`s that borrow from the arena rather than from `&self`. `Row` and row-spine's `DatumSeq` ignore it today; it is the hook a future compressed row representation needs — any `Datum<'a>` tied to `&'a self` is fundamentally incompatible with a representation whose decoded bytes don't live in `self`.
- Converts the last external users of `DatumSeq`'s inherent `Iterator` (reduce, delta/linear joins, peek, top_k, and render's `as_collection`/filtered-index import) to `extend_datums`, so the only remaining `DatumSeq::Iterator` users are row-spine internals.

## Why

A no-op refactor that finishes the push-based datum-decode interface (`main` had already started the `extend_datums` migration) and adds the arena hook. It is a prerequisite for compressing in-memory arrangements (per-column codecs, prototyped separately — early TPCH benchmarks show ~-65% arrangement memory); landing it on its own keeps that follow-up small and reviewable.

## Tests

Existing `repr`, `row-spine`, and `compute` tests cover the changed paths. The `encode.rs` ordering test switches from `to_datum_iter` to `Row::iter`. No behavior change; no release note.

### DIFF
--- a/doc/developer/generated/repr/_crate.md
+++ b/doc/developer/generated/repr/_crate.md
@@ -24,7 +24,7 @@ The lingua franca of Materialize: defines the core data types that all layers of
 * `namespaces` — well-known schema name constants
 * `optimize` — `OptimizerFeatures`, `OptimizerFeatureOverrides`, and `OverrideFrom` trait
 * `refresh_schedule` — REFRESH EVERY/AT schedule representation
-* `fixed_length` — `ToDatumIter` abstraction
+* `fixed_length` — `ExtendDatums` abstraction
 * `datum_vec` — reusable `Datum` scratch buffer
 * `bytes` — PostgreSQL-compatible `ByteSize`
 * `user` — external user auth metadata
--- a/src/compute/src/render/context.rs
+++ b/src/compute/src/render/context.rs
@@ -27,7 +27,7 @@ use mz_compute_types::plan::{ArrangementStrategy, AvailableCollections};
 use mz_dyncfg::ConfigSet;
 use mz_expr::{Eval, Id, MapFilterProject, MirScalarExpr};
 use mz_ore::soft_assert_or_log;
-use mz_repr::fixed_length::ToDatumIter;
+use mz_repr::fixed_length::ExtendDatums;
 use mz_repr::{DatumVec, DatumVecBorrow, Diff, GlobalId, Row, RowArena, SharedRow};
 use mz_storage_types::controller::CollectionMetadata;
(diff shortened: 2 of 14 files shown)


DAlperin (approved):
LGTM, thanks

doc/developer/generated/repr/_crate.md
antiguru, line 1: Please do not edit generated files.

src/compute/src/render/context.rs
```diff
@@ -323,10 +324,12 @@ impl<'scope, T: RenderTimestamp> ArrangementFlavor<'scope, T> {
                           d,
                           ok_session: &mut Session<T, DCB>,
                           err_session: &mut Session<T, ECB<T>>| {
+            // Declared before the borrow so it outlives any datums that decode into it.
+            let temp_storage = RowArena::new();
```
antiguru, line 328: You might be able to move this out of the closure and instead clear here.

src/compute/src/render/context.rs
```diff
@@ -377,10 +380,12 @@ impl<'scope, T: RenderTimestamp> ArrangementFlavor<'scope, T> {
     {
         let mut datums = DatumVec::new();
         let logic = move |k: DatumSeq, v: DatumSeq, t, d, ok_session: &mut Session<T, DCB>| {
+            // Declared before the borrow so it outlives any datums that decode into it.
+            let temp_storage = RowArena::new();
```
antiguru, line 384: Same here

src/compute/src/render/context.rs
frankmcsherry: Will investigate in a follow-up.

src/compute/src/render/context.rs
frankmcsherry: Same here

PR #37111, opened 2026-06-17
row-spine: per-column arrangement compression (alpha, default off)

frankmcsherry:
## Motivation

Row-spine arrangements currently compress at the whole-row level via the
just-shipped dictionary codec (`enable_arrangement_dictionary_compression_alpha`).
That works well when whole rows repeat, but leaves a lot on the table for wide
rows where individual columns have exploitable structure (low-cardinality enums,
clustered integers, repeated strings) but the rows as a whole are nearly unique.

This PR adds a **per-column compression framework**: each column in a row gets
its own codec, chosen independently from a menu, based on batch-wide statistics.

## What's here

A new `mz_row_spine::codec` module:

- **`ColumnCodec`** — a self-delimiting per-column codec enum: `Raw`,
  `Constant`, `Huffman` (boxed canonical Huffman, entropy), `For`
  (Frame-of-Reference = subtract-min then bitpack to a fixed byte width), and
  `Dictionary`. Because each codec is self-delimiting, they compose freely down
  a row.
- **`RowCodec`** — encodes/decodes a whole row by dispatching column-by-column.
- **`RowStats` / `ColumnStats`** — accumulate per-column statistics over a batch
  and pick the smallest-estimated codec per column, charging per-codec table
  storage (dictionary maps, Huffman tables) against the estimate so fixed
  overhead can't lose us bytes on small or high-cardinality columns.
- **Canonical Huffman** byte codec (`huffman.rs`) with forward-streaming decode.

The per-column dictionary is a **bounded heavy-hitter (MisraGries) summary** with
**single-byte references capped at 134 entries** — the same capacity and spare-tag
scheme as the existing `row_codec` dictionary. Popular values compress; the rare
tail falls through to raw, so compression stays available however large the tail
grows (no enumerate-all "give up" cliff).

Crucially, the codecs track the **full arrangement lifecycle**, mirroring the
dictionary codec:

- **Survive compaction** — `RowCodec` retains its (mergeable) stats and rebuilds
  a codec valid for the union via `new_from` on merge, re-harvesting counts from
  the rows it actually encodes — so a merge re-derives a codec instead of
  dropping the batch to raw.
- **Install mid-formation** — `build_safe` selects only codecs that stay sound as
  unseen values arrive (floored Huffman, heavy-hitter dictionary) and re-encodes
  the gathered rows at `STATS_THRESHOLD`, so reduce/upsert/CDC outputs get
  compressed before the first seal.

Decode is deferred to the operator that needs datums: `ReadItem` carries the
compressed bytes plus a reference to the codec, and decode happens into a
thread-local scratch for comparisons or into the caller's `RowArena` for
`extend_datums` / `to_row`.

## Relationship to dictionary compression

This is **purely additive and orthogonal** to the existing dictionary codec:

- It makes **zero changes** to the old dictionary path (`ColumnsCodec` /
  dictionary builder / `row_codec` module).
- It uses a **separate container field** (`column_codec`, alongside the existing
  `codec`) and a **separate flag**
  (`enable_arrangement_column_compression_alpha`, default off).
- The two are **mutually exclusive at runtime**: `build_column_codec` returns
  `None` whenever a dictionary codec is installed.

So the recently-shipped dictionary feature is untouched and independently
controllable; either flag can be flipped or walked back with no effect on the
other.

## Flag wiring

`enable_arrangement_column_compression_alpha` mirrors the dictionary flag end to
end: captured once at replica creation, threaded through
`InstanceConfig`/`ReplicaConfig`, and applied as a one-shot store into the
process-global `COLUMN_COMPRESSION` flag at instance creation (deliberately not
re-applied on config update, so flipping it never retroactively rewrites a live
replica's arrangements).

## Evaluation

Measured against a real TPCH SF1 load generator (6.0M `lineitem` rows), on a
dedicated single-worker index where the replica's process memory is essentially
the arrangement (no dilution), comparing the **current default (both alpha flags
off ⇒ raw)** against each codec:

| config | replica memory | hydration time |
| --- | --- | --- |
| raw (default) | 3137 MB | 8.0 s |
| whole-row dictionary | 1896 MB (−39.6%) | 28.1 s |
| per-column (this PR) | 1836 MB (−41.5%) | 29.3 s |

Takeaways:

- **vs the current default (raw):** per-column compression cuts arrangement
  memory ~42% on this workload.
- **vs whole-row dictionary:** the marginal gain on TPCH is small (~3%, ~60 MB).
  `lineitem` is dictionary-friendly — it has many low-cardinality columns that
  *both* codecs reduce to a single byte — and its one large text column
  (`l_comment`) is high-entropy random alphanumeric in the generator, so order-0
  Huffman only trims ~26% of it. The per-column codecs that the dictionary lacks
  (Huffman on text, FoR on fixed-width dates) therefore pay off most on workloads
  TPCH under-represents: genuinely repetitive / low-entropy text, and many
  bounded-range fixed-width numeric or date columns. (FoR in particular is
  limited by `mz_repr`'s little-endian integer/date encoding — it narrows
  fixed-width `Date` columns ~20% but cannot touch variable-length `Numeric`.)
- **Hydration cost (~3.5× over raw)** is dominated by the heavy-hitter statistics
  pass — one MisraGries insert per column per row — and is **shared with the
  dictionary codec**; it is not specific to per-column compression. The
  per-column path adds only ~4% over dictionary (a per-byte frequency count for
  Huffman). Reducing this shared stats cost would speed up both.

(Local `clusterd` uses the system allocator; production jemalloc may shift the
absolute numbers. The earlier "~-56%" figure was from a synthetic generator with
low-entropy text and over-stated the win — the table above supersedes it.)

## Testing

Unit tests in `codec.rs` / `huffman.rs` cover round-trip encode/decode for each
codec and the selector, plus merge-rebuild (`new_from`), mid-formation install,
and dictionary robustness to a long rare tail.

This release will add an alpha `enable_arrangement_column_compression_alpha`
feature flag (default off) for per-column arrangement compression.

### DIFF
--- a/misc/python/materialize/mzcompose/__init__.py
+++ b/misc/python/materialize/mzcompose/__init__.py
@@ -71,6 +71,7 @@ def get_minimal_system_parameters(
             "true" if version >= MzVersion.parse_mz("v0.132.0-dev") else "false"
         ),
         "enable_alter_swap": "true",
+        "enable_arrangement_column_compression_alpha": "false",
         "enable_arrangement_dictionary_compression_alpha": "false",
         "enable_case_literal_transform": "true",
         "enable_cast_elimination": "true",
--- a/src/compute-types/src/dyncfgs.rs
+++ b/src/compute-types/src/dyncfgs.rs
@@ -401,12 +401,15 @@ pub const ENABLE_ARRANGEMENT_DICTIONARY_COMPRESSION_ALPHA: Config<bool> = Config
 
 /// Whether to enable per-column arrangement compression (the `RowCodec`
 /// framework). Independent of, and not combined with, dictionary compression.
-/// Like dictionary compression, captured at replica creation and held fixed.
+///
+/// Replica-scoped: the per-replica-resolved value is applied to the process-global
+/// `mz_row_spine::COLUMN_COMPRESSION` flag in the compute replica's `apply_worker_config`.
 pub const ENABLE_ARRANGEMENT_COLUMN_COMPRESSION_ALPHA: Config<bool> = Config::new(
     "enable_arrangement_column_compression_alpha",
     false,
     "Enable per-column arrangement compression (alpha; not yet production-ready).",
-);
+)
+.scoped(ParameterScope::Replica);
 
 /// Whether to enable the peek response stash, for sending back large peek
 /// responses. The response stash will only be used for results that exceed
--- a/src/compute-types/src/dyncfgs.rs
+++ b/src/compute-types/src/dyncfgs.rs
@@ -399,6 +399,15 @@ pub const ENABLE_ARRANGEMENT_DICTIONARY_COMPRESSION_ALPHA: Config<bool> = Config
     "Enable arrangement dictionary compression (alpha; not yet production-ready).",
 );
 
+/// Whether to enable per-column arrangement compression (the `RowCodec`
+/// framework). Independent of, and not combined with, dictionary compression.
+/// Like dictionary compression, captured at replica creation and held fixed.
+pub const ENABLE_ARRANGEMENT_COLUMN_COMPRESSION_ALPHA: Config<bool> = Config::new(
+    "enable_arrangement_column_compression_alpha",
+    false,
+    "Enable per-column arrangement compression (alpha; not yet production-ready).",
+);
+
 /// Whether to enable the peek response stash, for sending back large peek
 /// responses. The response stash will only be used for results that exceed
 /// `compute_peek_response_stash_threshold_bytes`.
@@ -525,6 +534,7 @@ pub fn all_dyncfgs(configs: ConfigSet) -> ConfigSet {
         .add(&COMPUTE_LOGICAL_BACKPRESSURE_MAX_RETAINED_CAPABILITIES)
         .add(&COMPUTE_LOGICAL_BACKPRESSURE_INFLIGHT_SLACK)
         .add(&ENABLE_ARRANGEMENT_DICTIONARY_COMPRESSION_ALPHA)
+        .add(&ENABLE_ARRANGEMENT_COLUMN_COMPRESSION_ALPHA)
         .add(&ENABLE_PEEK_RESPONSE_STASH)
         .add(&PEEK_RESPONSE_STASH_THRESHOLD_BYTES)
         .add(&PEEK_RESPONSE_STASH_BATCH_MAX_RUNS)
--- a/src/compute/src/compute_state.rs
+++ b/src/compute/src/compute_state.rs
@@ -308,6 +308,16 @@ impl ComputeState {
             std::sync::atomic::Ordering::Relaxed,
         );
 
+        // Per-column arrangement compression is replica-scoped (see
+        // `ENABLE_ARRANGEMENT_COLUMN_COMPRESSION_ALPHA`), so its per-replica-resolved value
+        // arrives in `worker_config`; mirror it into the process-global flag row-spine reads at
+        // seal time. Safe to re-apply on every tick: each batch records its own codec, so a flip
+        // only affects batches sealed afterwards.
+        mz_row_spine::COLUMN_COMPRESSION.store(
+            ENABLE_ARRANGEMENT_COLUMN_COMPRESSION_ALPHA.get(config),
+            std::sync::atomic::Ordering::Relaxed,
+        );
+
         // NB: arrangement dictionary compression is deliberately NOT applied here. Unlike the
         // settings above, it is captured once at replica creation (see `handle_create_instance`
         // and `InstanceConfig::arrangement_dictionary_compression`) and held fixed, so that
@@ -489,11 +499,8 @@ impl<'a> ActiveComputeState<'a> {
             config.arrangement_dictionary_compression,
             std::sync::atomic::Ordering::Relaxed,
         );
-        // Same one-shot treatment for per-column compression; see the note above.
-        mz_row_spine::COLUMN_COMPRESSION.store(
-            config.arrangement_column_compression,
-            std::sync::atomic::Ordering::Relaxed,
-        );
+        // NB: per-column compression is replica-scoped and applied in `apply_worker_config`
+        // (called just above), not captured on `InstanceConfig`.
 
         if let Some(offset) = config.expiration_offset {
             self.compute_state.apply_expiration_offset(offset);
--- a/src/compute/src/compute_state.rs
+++ b/src/compute/src/compute_state.rs
@@ -489,6 +489,11 @@ impl<'a> ActiveComputeState<'a> {
             config.arrangement_dictionary_compression,
             std::sync::atomic::Ordering::Relaxed,
         );
+        // Same one-shot treatment for per-column compression; see the note above.
+        mz_row_spine::COLUMN_COMPRESSION.store(
+            config.arrangement_column_compression,
+            std::sync::atomic::Ordering::Relaxed,
+        );
 
         if let Some(offset) = config.expiration_offset {
             self.compute_state.apply_expiration_offset(offset);
--- a/src/row-spine/src/codec.rs
+++ b/src/row-spine/src/codec.rs
@@ -1,0 +1,363 @@
+// Copyright Materialize, Inc. and contributors. All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! Per-column compression for row-spine batches.
+//!
+//! Each column of a batch is compressed independently with a [`ColumnCodec`]
+//! chosen from batch-wide statistics, so a column of ids, a column of names, and
+//! a column of flags each get a scheme that fits. A [`RowCodec`] bundles one
+//! codec per column and encodes/decodes whole rows row-major: a row's columns
+//! are encoded back-to-back into the row's stored bytes.
+//!
+//! Codecs are **self-delimiting** in the encoded stream — each one consumes
+//! exactly its own bytes on decode — so they compose without a central
+//! length-prefix block: decoding a row just walks the column codecs in order.
+//!
+//! New codecs slot in by adding a [`ColumnCodec`] variant plus its size estimate
+//! in [`ColumnStats::choose`]; the selector then adopts it wherever it wins. The
+//! variants here are the row-major-compatible ones (each value encoded from a
+//! batch-wide stat); cross-row schemes (delta, RLE) would need a column-major
+//! layout and are out of scope.
+
+use mz_repr::read_datum;
+
+use crate::huffman::{BitReader, BitWriter, FrequencyCounter, HuffmanCode};
+
+/// A codec for a single column, selected per column at `seal()`.
+#[derive(Clone, Debug)]
+pub enum ColumnCodec {
+    /// Store the datum's bytes verbatim. Self-delimited by the row encoding
+    /// (`read_datum` knows where a datum ends).
+    Raw,
+    /// Every value in the column is identical batch-wide; nothing is stored per
+    /// row, and the value is reproduced on decode.
+    Constant(Box<[u8]>),
+    /// Huffman-code the datum's bytes: `uvarint(decoded_len) ++ byte-aligned bits`.
+    Huffman(HuffmanCode),
+}
+
+impl ColumnCodec {
+    /// Append the encoding of one column value to `out`.
+    fn encode(&self, value: &[u8], out: &mut Vec<u8>) {
+        match self {
+            ColumnCodec::Raw => out.extend_from_slice(value),
+            ColumnCodec::Constant(_) => {}
+            ColumnCodec::Huffman(code) => {
+                put_uvarint(value.len(), out);
+                let mut writer = BitWriter::new();
+                code.encode(value, &mut writer);
+                out.extend_from_slice(&writer.finish());
+            }
+        }
+    }
+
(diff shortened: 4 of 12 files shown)


src/compute-types/src/dyncfgs.rs
```diff
@@ -399,6 +399,15 @@ pub const ENABLE_ARRANGEMENT_DICTIONARY_COMPRESSION_ALPHA: Config<bool> = Config
...
 );
 
+/// Whether to enable per-column arrangement compression (the `RowCodec`
+/// framework). Independent of, and not combined with, dictionary compression.
+/// Like dictionary compression, captured at replica creation and held fixed.
+pub const ENABLE_ARRANGEMENT_COLUMN_COMPRESSION_ALPHA: Config<bool> = Config::new(
+    "enable_arrangement_column_compression_alpha",
+    false,
+    "Enable per-column arrangement compression (alpha; not yet production-ready).",
+);
```
antiguru: Please scope to Replica.

frankmcsherry: Happy to. Is there a mechanism to discover what this means and how to do it, or just spend tokens?

antiguru: Here's an example: https://github.com/MaterializeInc/materialize/blob/main/src/compute-types/src/dyncfgs.rs#L61

antiguru:
Edit: I @antiguru actually requested the review, but Claude didn't know, and got confused.

----

## Review — row-spine per-column arrangement compression

> _This review was written by Claude (Claude Code), at the request of @frankmcsherry / the PR author._

### TL;DR

The feature code (`codec.rs`, `huffman.rs`, the `lib.rs` wiring) is genuinely good: clean, well-documented, correctly self-delimiting codecs, sound mutual-exclusion with the dictionary path, and a default-off alpha flag wired exactly like its predecessor. But the PR as it stands is not mergeable, for one mechanical reason and a couple of correctness/robustness gaps worth closing before this goes in even behind a flag.

---

### ❌ Blocker 1 — 134,008 lines of unrelated junk

The final commit `6d74d99` ("compute: make per-column compression flag replica-scoped") swept in **80 unrelated working files** alongside the legitimate one-line flag-scoping change:

- `misc/build-time/*` (incl. a 55,905-line `cargo-timing.html`, llvm-lines dumps, logs)
- `misc/decoupling-scans/*` (~14k lines of scan output)
- `play/optimizer-rebuild/*`, `play/mz-bridge-recipe/package-lock.json`
- `misc/demo-data/*`, `.agents/skills/mz-demo-data`, `FEATURE_FLAG_AUDIT.md`

None are referenced by the feature. The real diff is just:

```
src/compute-types/src/dyncfgs.rs |  13 +
src/compute/src/compute_state.rs |  12 +
src/row-spine/src/codec.rs       | 592 +
src/row-spine/src/huffman.rs     | 410 +
src/row-spine/src/lib.rs         | 320 +
```

plus the two Python flag-registration files (`mzcompose/__init__.py`, `parallel_workload/action.py`), which **are** legitimate (`bin/lint-test-flags` requires them).

**Action:** rebuild the last commit so it carries only the flag-scoping change (`git reset` the junk paths out + `git commit --amend`, or interactive-rebase `6d74d99`). Until then the diff is unreviewable in a normal view and will trip `bin/lint` on a pile of these files.

---

### Correctness

**2. Variable / mixed arity is silent corruption in release (robustness gap).**
`RowStats::observe` grows its `columns` vec to the **max** arity ever seen, and `choose()` uses each column's own `count`. But `RowCodec::encode` indexes `self.columns[index]` per row column — a row with more columns than the codec models panics (index OOB); a row with fewer leaves `count != self.columns.len()`, caught only by a `debug_assert_eq!` (codec.rs:218) that's gone in release. And `decode_into` always walks **all** `self.columns`, so a short record makes `For`/`Dictionary::decode` slice `cursor[..width]` off the end → **panic even in release**, or `Raw` reads into the next record → corruption.

`build_column_codec` already hints the author saw this (it skips empty rows when gathering stats), but that's a half-measure: if a container ever holds both an empty (arity-0) row and non-empty rows, the empty row still goes through `push_into` → `encode` emits a 0-length record, and decoding it walks N columns off the end. In practice a single key/val container has uniform arity, so I believe this is **not reachable today** — but it's an unenforced invariant guarded only by a debug assert. Recommendation: have `RowStats` track the first row's arity and `build()` return `None` if any observed row's arity differs (and don't silently skip empties — they're an arity signal). Turns a latent release-mode panic into a clean "no codec."

**3. `HuffmanCode::from_frequencies` doc/behavior mismatch.** The doc says it returns `None` "when fewer than two distinct bytes appear," but `code_lengths` returns `Some` for the single-symbol case (length 1). So a single-distinct-byte column does build a code (1 bit/byte). Correct and round-trips, but the comment is wrong and no test exercises single-symbol selection through `ColumnCodec::Huffman`. Fix the comment, or actually return `None` for `present.len() == 1` if exclusion was intended.

**4. FoR / `be_uint` — verified correct ✓.** Checked against the row encoding (`Tag::NonNegativeInt64_24` = tag byte + N significant bytes). FoR is pure big-endian byte arithmetic over the fixed-length pattern including the tag, so it round-trips exactly regardless of tag/sign, correctly gated on `fixed_len && len <= 8 && max_int > min_int`, and `width` from `max-min` means it can never lose to `Raw`. `round_trip_for_numeric` is valid.

**5. Dictionary codes are insertion-ordered, not sorted.** `observe` assigns `code = distinct.len()` on first sight. Fine for correctness (decode is a lookup), but worth a one-line comment so a future reader doesn't assume order-preservation — and it forecloses the compare optimization below.

---

### Performance (flag is off, so non-blocking — but this is whether it can ship)

**6. The comparison path fully decodes both operands.** `PartialEq`/`PartialOrd` on encoded items decode both whole rows into thread-local scratch (Huffman bit-by-bit, dict lookups) then compare raw bytes. Differential merge/consolidate/seek is comparison-bound, so with the flag on this is O(full row) per comparison instead of O(distinguishing prefix). This is the single biggest risk to viability and deserves a dedicated benchmark (merge throughput, flag on vs off) before leaving alpha. The two-buffer thread-local design itself is sound. Possible win: decode column-by-column, stop at first differing column (`Constant` compares free); making `Dictionary`/`For` order-preserving would let those compare encoded, but Huffman isn't order-preserving so a general fast path needs the lazy column-wise decode.

**7. Per-row allocations on the read/scan path.** `to_row` does `Vec::new()` per call; `extend_datums` (the documented hot scan path) does `Vec::new()` **and then** `arena.push_bytes(buf)` — an allocation plus a copy per row, vs. the no-codec path's allocation-free tight loop. Both should reuse a thread-local scratch like `materialized` already does.

**8. `Huffman::decode_one` is bit-at-a-time** — O(max_len) branches + bounds-checked `read_bit` per symbol. A length-limited (`MAX_BITS=24`) table-driven decode would be much faster, worth it if Huffman is actually selected often (see test gap below).

**9. Merge re-encodes everything (inherent, fine).** `merge_capacity` correctly drops the codec and `push_into` decodes each incoming item to a `Row` then re-encodes under a fresh codec — unavoidable since codecs aren't reusable across merges; just another full decode + `to_row` alloc per row during merges.

---

### Test coverage gaps

1. **Huffman selection end-to-end.** No test asserts `ColumnCodec::Huffman` is ever *chosen* — `round_trip_text_heavy` and `round_trip_dictionary_lowcard` land on `Dictionary`, `round_trip_for_numeric` on `For`. So the `Huffman` arm through `RowCodec` and the model-cost accounting in `choose()` are never exercised. Add a case engineered to make Huffman win (high-cardinality, skewed bytes, >65,536 distinct so `Dictionary` is disqualified).
2. **The merge path** (`push_into` with `item.column_codec.is_some()`) — never tested; the container test only installs a codec and pushes fresh rows.
3. **Variable/empty-row arity** (finding #2) — once the guard is added.
4. **`MAX_BITS` decline path** in `code_lengths` — uncovered.
5. **`DICT_MAX_CARDINALITY` boundary** — exactly 65,536 / 65,537 distinct to confirm 2-byte width and fall-off to `None`.

---

### Minor nits

- `codec.rs:388` `let _ = best_bytes;` — the final `best_bytes` write in the `Dictionary` arm is dead; the discard silences the warning but reads oddly. A short comment or dropping the last assignment would help.
- `CMP_SCRATCH_A/B` thread-locals never shrink (retain the largest row seen for the thread's life) — negligible, worth a comment.
- `self_code` (codec.rs:192) is a one-line indirection over `index.get(...).expect(...)`.
- `compute_state.rs` "Safe to re-apply on every tick" is the right call and well-justified. 👍

---

### Suggested benchmarks

1. **Merge/consolidate throughput, flag on vs off** on a wide low-cardinality arrangement (e.g. TPCH `lineitem`) — exercises the full-row-decode comparison path (#6). This is the gating measurement.
2. **Arrangement scan latency** (`extend_datums`) on/off — captures the per-row alloc regression (#7).
3. **Memory footprint** on/off across the TPCH set — reproduce the "-56% lineitem / -43% mixed" claim in a repeatable harness (Feature Benchmark / `mz-benchmark` entry so CI tracks it).
4. **Seal cost** on high-cardinality columns — the per-batch stats pass builds a `BTreeMap` up to 65k entries per column and then discards it; confirm seal stays cheap.

---

Net: solid, well-isolated feature behind an off-by-default flag. Strip the junk commit (blocker), add the arity guard, and fill the Huffman-selection + merge-path test gaps; treat the comparison-path decode cost as the key thing to measure before promoting past alpha.

---

src/compute-types/src/dyncfgs.rs
frankmcsherry: Accomplished!

src/row-spine/src/lib.rs
```diff
@@ -23,10 +23,22 @@ pub use self::spines::{
 };
 use differential_dataflow::trace::implementations::OffsetList;
 
+// The per-column compression layer. Exercised by tests today and wired into the
+// containers in a follow-up; allow it to land unused in the meantime.
+#[allow(dead_code)]
+mod codec;
+#[allow(dead_code)]
+mod huffman;
```
DAlperin, line 31: Not actually dead anymore I think?

src/row-spine/src/lib.rs
```diff
@@ -1293,12 +1556,55 @@ mod dictionary {
 
         #[inline]
         pub fn to_row(&self) -> Row {
+            if let Some(codec) = self.column_codec {
+                // `iter.data` is RowCodec-encoded; decode to raw row bytes.
+                let mut buf = Vec::new();
+                codec.decode_into(self.iter.data, &mut buf);
+                return unsafe { Row::from_bytes_unchecked(&buf) };
+            }
```
DAlperin, line 1816: Probably better to decode into an arena here if one is available than allocate per row

frankmcsherry: Yup. We don't have that until #37115 lands, I think.

DAlperin: I should go read that then :)

src/row-spine/src/codec.rs
```diff
@@ -0,0 +1,710 @@
...
+        if self.is_constant {
+            if let Some(value) = &self.constant {
+                return ColumnCodec::Constant(value.clone());
+            }
+        }
+
+        let mut best = ColumnCodec::Raw;
+        let mut best_bytes = self.raw_bytes;
+
+        if let Some(code) = self.freq.build() {
```
DAlperin, line 388: Probably pretty minor in the grand scheme of things but we do the expensive check first. Calling build on the huffman code calls code_lengths which is slightly expensive (BinaryHeap over >= 256 symbols plus allocs). One could optimize this by running the cheap codecs first to lower best_bytes and then gate the huffman build behind a cheap lower bound like `let huffman_floor = self.count.saturating_mul(2).saturating_add(size_of::<HuffmanCode>() + 256);`

Again, pretty minor and might need some stats/profile to prove if its really a problem.

src/compute/src/compute_state.rs
```diff
@@ -489,6 +499,8 @@ impl<'a> ActiveComputeState<'a> {
             config.arrangement_dictionary_compression,
             std::sync::atomic::Ordering::Relaxed,
         );
+        // NB: per-column compression is replica-scoped and applied in `apply_worker_config`
+        // (called just above), not captured on `InstanceConfig`.
```
antiguru, line 503: Nit: Comment reads like a note-to-self.

src/row-spine/src/codec.rs
```diff
@@ -0,0 +1,710 @@
...
+            }
+            self.columns[index].observe(bytes);
+            count = index + 1;
+        }
+        // Track arity consistency. An empty (arity-0) row is a real arity signal here,
+        // not something to skip: a codec built from non-empty rows would walk its columns
+        // off the end of an empty record at decode time.
+        match self.arity {
+            None => self.arity = Some(count),
+            Some(arity) if arity != count => self.mixed_arity = true,
```
antiguru, line 602: In what situation would we expected mixed arity rows?

misc/python/materialize/mzcompose/__init__.py
```diff
@@ -71,6 +71,7 @@ def get_minimal_system_parameters(
             "true" if version >= MzVersion.parse_mz("v0.132.0-dev") else "false"
         ),
         "enable_alter_swap": "true",
+        "enable_arrangement_column_compression_alpha": "false",
```
antiguru, line 74: I'd like to see a test/nightly run with compression enabled.

antiguru (approved):
Seems good, but I'd like to see a test/nightly run with the flag enabled. At the moment it seems mostly untested delta the unit tests (thanks for adding!)

I'll approve, don't see a blocker here.

PR #37112, opened 2026-06-17
sql: Fix join in EXPLAIN ANALYZE and EXPLAIN ANALYZE CLUSTER

mgree:
### Motivation

Closes [SQL-356](https://linear.app/materializeinc/issue/SQL-356/explain-analyze-memory-cpu-with-skew-silently-drops-every).

### Description

Allow `worker_id` to be `NULL` in both per-worker queries.

### Verification

Updated SLT with tests from bug report.


def- (approved):
Thanks!

PR #37114, opened 2026-06-17
repr: back RowArena with a doubling region allocator

frankmcsherry:
## What

Reworks `RowArena` from "one `Vec<u8>` allocation per `push_bytes`, all freed on `clear`" into a bump allocator over a stack of byte regions:

- `push_bytes` **copies** bytes into the active (last) region. When that region lacks spare capacity it allocates a **new, larger** region (doubling) rather than growing the current one — so a region that already holds data is never reallocated and references returned earlier stay valid. The outer `Vec` may reallocate as regions are added, but that only moves the `Vec<u8>` headers, not the heap buffers they own.
- `push_bytes` now accepts any `B: Deref<Target = [u8]>` (e.g. `Vec<u8>`, `&[u8]`). Existing `Vec<u8>` callers are unchanged; borrowed sources no longer need a throwaway allocation just to hand bytes over.
- `clear` retains only the **single largest region** (emptied) and drops the rest, right-sizing the arena. An arena reused across `clear` cycles — e.g. decoding arrangement rows one at a time — becomes allocation-free in steady state.
- `reserve` now reserves *bytes* in the active region (was: slots in the outer vector); `with_capacity` sizes the initial region. Neither has load-bearing callers (the `reserve` users in `expr` pass size hints).

## Why

Follows up the `ExtendDatums` arena work (#37110) and the per-row arena hoist (#37113). Once the arena is the decode target for compressed arrangement rows (per-column codecs, #37111), a fresh allocation per value per row is the dominant cost; a reused doubling region removes it. The pattern mirrors columnation's region allocator.

The trade-off is the standard bump-allocator one: a single very large row leaves the arena holding a large region until the next `clear`. For the per-row-cleared decode/eval paths the high-water mark is just the largest single row, so it stays bounded.

## Tests

Adds `mz_ore::test`s in `repr::row` covering:
- references from `push_bytes` remaining valid after later pushes force new regions,
- `push_unary_row` reading a row back from a non-zero region offset (confirms decoding is position/alignment independent),
- reuse across `clear`.

No behavior change for callers; no release note.

### DIFF
--- a/src/repr/src/row.rs
+++ b/src/repr/src/row.rs
@@ -808,12 +808,21 @@ pub struct DatumDictTypedIter<'a, T> {
 /// `RowArena` is used to hold on to temporary `Row`s for functions like `eval` that need to create complex `Datum`s but don't have a `Row` to put them in yet.
 #[derive(Debug)]
 pub struct RowArena {
-    // Semantically, this field would be better represented by a `Vec<Box<[u8]>>`,
-    // as once the arena takes ownership of a byte vector the vector is never
-    // modified. But `RowArena::push_bytes` takes ownership of a `Vec<u8>`, so
-    // storing that `Vec<u8>` directly avoids an allocation. The cost is
-    // additional memory use, as the vector may have spare capacity, but row
-    // arenas are short lived so this is the better tradeoff.
+    // A stack of byte regions, used as a bump allocator. Bytes handed to
+    // `push_bytes` are *copied* into the active (last) region and a reference
+    // into that region is returned.
+    //
+    // The invariant that keeps returned references valid for the arena's
+    // lifetime is that a region is never reallocated once it holds data: when
+    // the active region lacks spare capacity for a push we allocate a *new*,
+    // larger region rather than growing the current one (which would move its
+    // bytes and dangle outstanding references). The outer `Vec` may itself
+    // reallocate as regions are added, but that only moves the `Vec<u8>`
+    // headers, not the heap buffers they own, so references remain valid.
+    //
+    // `clear` retains only the largest region (emptied) to right-size the arena
+    // for reuse; reusing one region across `clear` cycles makes a steady-state
+    // workload (e.g. decoding rows one at a time) allocation-free.
     inner: RefCell<Vec<Vec<u8>>>,
 }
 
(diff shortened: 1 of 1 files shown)


src/repr/src/row.rs
```diff
@@ -3036,45 +3045,88 @@ impl RowArena {
...
+            // The active region is empty, so nothing references it yet and it is safe to grow it
+            // in place (a reallocation cannot dangle a live reference).
+            Some(active) if active.is_empty() => {
+                if active.capacity() < additional {
+                    active.reserve_exact(additional);
+                }
+            }
+            // The active region holds live data; we cannot grow it without moving those bytes, so
+            // stage a fresh region for the upcoming pushes instead.
+            Some(_) => inner.push(Vec::with_capacity(additional)),
```
frankmcsherry: This may have the potential to pre-empt the doubling going on. Could be ok, but may also result in many small allocations.

src/repr/src/row.rs
```diff
@@ -3036,45 +3045,88 @@ impl RowArena {
...
+            //     `Vec<u8>` headers, not the heap buffers they own.
+            transmute::<&[u8], &'a [u8]>(copied)
         }
     }
 
-    /// Take ownership of `string` for the lifetime of the arena.
+    /// Copies `string` into the arena and returns a reference valid for its lifetime.
     pub fn push_string<'a>(&'a self, string: String) -> &'a str {
-        let owned_bytes = self.push_bytes(string.into_bytes());
+        let copied = self.push_bytes(string.as_bytes());
```
frankmcsherry, line 3131: Maybe worth deprecating? Seemingly any call site should be nudged to copy in the bytes instead, and avoid an allocation. Could even have a `push_format_args` if we end up building strings for that reason.

src/repr/src/row.rs
```diff
@@ -3084,39 +3136,23 @@ impl RowArena {
     /// If we had an owned datum type, this method would be much clearer, and
     /// would be called `push_owned_datum`.
     pub fn push_unary_row<'a>(&'a self, row: Row) -> Datum<'a> {
```
frankmcsherry, line 3143: Similar to `push_string`, we might want to audit for anyone creating rows for the purpose of pushing them here, and see if we have the bytes lying around in un-owned form.

src/repr/src/row.rs
```diff
@@ -3183,8 +3219,17 @@ impl RowArena {
     }
 
     /// Clear the contents of the arena.
+    ///
+    /// Retains the single largest region (emptied) so the arena can be reused without
+    /// reallocating; a workload that clears between uses of similar size becomes allocation-free.
     pub fn clear(&mut self) {
```
frankmcsherry, line 3230: If we double consistently this should probably be the last one, but without the double (due to reserving capacity) this checks out as the right logic. No strong opinion on which to use, but light bias towards keeping at most log-many allocations in the worst case.

src/repr/src/row.rs
antiguru: Yes, I'd at least change it to `push_str` to make it explicit that the function doesn't take ownership.

antiguru (approved):
Seems good!

src/repr/src/row.rs
frankmcsherry: The plan is to migrate these in future PRs, supporting a longer-lived buffer into which one can `format!` strings. Concretely https://github.com/MaterializeInc/materialize/pull/37115 does this, though tbd on the migration for users of this method (planned as a follow-up PR once we can rewrite those sites).

PR #37115, opened 2026-06-17
repr: add RowArenaBuf, a growable arena-backed byte buffer

frankmcsherry:
## What

Adds `RowArena::writer()` → `RowArenaBuf`: a `Vec`-shaped, writeable byte buffer backed by the arena. You build a value incrementally and commit it:

```rust
let mut w = arena.writer();
w.extend_from_slice(b"hello");
w.push(b'!');
let bytes: &[u8] = w.finish(); // valid for the arena's lifetime
```

It supports `push`, `extend_from_slice`, `std::io::Write`, reads back as `&[u8]` (via `Deref`/`as_slice`), and `finish()` commits the bytes into the arena and returns a reference valid for the arena's lifetime.

## Why

With the region-backed `RowArena` (#37114), `push_bytes` copies an already-assembled slice into a region. A producer that assembles bytes piecewise — decoding a compressed row, say — would otherwise build into a scratch buffer and then have `push_bytes` copy it in. `RowArenaBuf` lets it write **directly into arena storage**, removing the scratch and the commit-time copy.

Growth doubles capacity and **recycles the previous allocation** through a small bounded pool on the arena (`MAX_FREE_BUFFERS`) rather than freeing it; an abandoned (un-`finish`ed) buffer is recycled on drop, and `clear` feeds the same pool. So a steady-state producer reuses allocations across rows.

## Safety

The key invariant: no reference into the buffer escapes before `finish`, so the buffer is free to relocate (double-and-copy) while being built. Only `finish` hands out a `&'a [u8]`, and from then on those bytes live in a committed region that is never reallocated while it holds data — the same rule `push_bytes` relies on in #37114.

## Tests

`mz_ore::test`s (run under miri) cover incremental builds across many growth steps, `io::Write`, pooled reuse across writers, reference validity after a later writer commits, and empty/abandoned writers.

## Stacking

Builds on #37114 (region-backed `RowArena`). No call sites adopt the writer yet; wiring the per-column decode path onto it is a follow-up.

No behavior change for existing callers; no release note.

### DIFF
--- a/src/repr/src/lib.rs
+++ b/src/repr/src/lib.rs
@@ -67,8 +67,8 @@ pub use crate::row::encode::{RowColumnarDecoder, RowColumnarEncoder, preserves_o
 pub use crate::row::iter::{IntoRowIterator, RowIterator};
 pub use crate::row::{
     DatumDictTypedIter, DatumList, DatumListTypedIter, DatumMap, FromDatum, ProtoNumeric, ProtoRow,
-    Row, RowArena, RowPacker, RowRef, SharedRow, datum_list_size, datum_size, datums_size,
-    read_datum, row_size,
+    Row, RowArena, RowArenaBuf, RowPacker, RowRef, SharedRow, datum_list_size, datum_size,
+    datums_size, read_datum, row_size,
 };
 pub use crate::scalar::{
     ArrayRustType, AsColumnType, Datum, DatumKind, ExcludeNull, InputDatumType, Int2Vector,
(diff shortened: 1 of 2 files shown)


DAlperin:
I think this is good but I have two questions:
> RowArenaBuf lets it write directly into arena storage, removing the scratch and the commit-time copy.

I think this doesn't actually remove the commit-time copy right? `finish` calls `push_bytes` which does `region.extend_from_slice` which copies in my understanding. 

This still solves the alloc churn problem though which is good.

The other thing to think about is that writer() has a reentrancy risk which you could imagine doing a nested decode that tries to get a mutable reference to the same underlying stash

antiguru (approved):
Looks good!

PR #37116, opened 2026-06-17
[catalog] fix duplicate rows in SHOW COLUMNS after builtin type migration 

mtabebe:
After converting `mz_sources` from a BuiltinTable to a
BuiltinMaterializedView, `SHOW COLUMNS FROM mz_sources` returned each
column twice on environments that upgraded across the change. 

The cause is that  the builtin schema migration preserves the catalog id but does not 
drop the prior comment rows. 

`mz_internal.mz_comments` ends up with 15 table + 15 materialized-view
rows for the same id, and mz_show_columns joined against all of them.

The same class of bug will reappear any time a builtin's `CatalogItemType`
changes across a release.

Two complementary fixes (1 commit each)

1. Read side: mz_show_columns, predicate the comment join on object_type
against mz_objects.type.

2. Write side: on each boot, drop comments under every relation-style
CommentObjectId variant for the builtin's id, not just the current one.
This cleans up the stale rows on the next restart and self-heals any future migrations.

Tests:
- New `test/sqllogictest/show_columns.slt` locks in per-builtin column
  counts for `mz_sources`, `mz_tables`, `mz_secrets`, and exercises a
  user table with `COMMENT ON COLUMN`.
- Verified end-to-end: ran `environmentd` on the old version with
  `mz_sources` as a Table to populate the durable catalog, then
  restarted

### DIFF
--- a/src/adapter/src/catalog/open.rs
+++ b/src/adapter/src/catalog/open.rs
@@ -881,7 +881,16 @@ fn add_new_remove_old_builtin_items_migration(
             | Builtin::Index(_)
             | Builtin::Connection(_) => continue,
         };
-        txn.drop_comments(&BTreeSet::from_iter([comment_id]))?;
+        // Drop comments under every relation-style `CommentObjectId` variant
+        // for this id, not just the current one. When a builtin's type changes
+        // (e.g. Table -> MaterializedView) but its catalog id is preserved, the
+        // prior type's comment rows would otherwise linger forever.
+        txn.drop_comments(&BTreeSet::from_iter([
+            CommentObjectId::Table(id),
+            CommentObjectId::View(id),
+            CommentObjectId::MaterializedView(id),
+            CommentObjectId::Source(id),
+        ]))?;
 
         let mut comments = comments.clone();
         for (col_idx, name) in desc.iter_names().enumerate() {
--- a/src/catalog/src/builtin/mz_internal.rs
+++ b/src/catalog/src/builtin/mz_internal.rs
@@ -4964,7 +4964,8 @@ pub static MZ_SHOW_SECRETS: LazyLock<BuiltinView> = LazyLock::new(|| BuiltinView
     ontology: None,
 });
 
-pub static MZ_SHOW_COLUMNS: LazyLock<BuiltinView> = LazyLock::new(|| BuiltinView {
+pub static MZ_SHOW_COLUMNS: LazyLock<BuiltinView> = LazyLock::new(|| {
+    BuiltinView {
     name: "mz_show_columns",
     schema: MZ_INTERNAL_SCHEMA,
     oid: oid::VIEW_MZ_SHOW_COLUMNS_OID,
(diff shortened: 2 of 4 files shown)


def- (approved):
Test changes lgtm

src/adapter/src/catalog/open.rs
```diff
@@ -881,7 +881,16 @@ fn add_new_remove_old_builtin_items_migration(
             | Builtin::Index(_)
             | Builtin::Connection(_) => continue,
         };
-        txn.drop_comments(&BTreeSet::from_iter([comment_id]))?;
+        // Drop comments under every relation-style `CommentObjectId` variant
+        // for this id, not just the current one. When a builtin's type changes
+        // (e.g. Table -> MaterializedView) but its catalog id is preserved, the
+        // prior type's comment rows would otherwise linger forever.
+        txn.drop_comments(&BTreeSet::from_iter([
```
ggevay, line 888: Consider moving this drop above the `match`, so it runs for every builtin rather than only the ones whose *current* type is a relation. The `continue` for `Log | Type | Func | Index | Connection` happens before we get here, so a builtin that changes type from, say, Table to Log (logs are relations with columns too) would leave its stale `Table` comment rows behind forever, which is the same class of bug this PR fixes. Dropping the four relation variants keyed only on `id` for every builtin closes the whole class, and the comment-writing below can keep its current shape.

src/catalog/src/builtin/mz_internal.rs
```diff
@@ -4977,13 +4978,22 @@ pub static MZ_SHOW_COLUMNS: LazyLock<BuiltinView> = LazyLock::new(|| BuiltinView
...
     sql: "
-    SELECT columns.id, name, nullable, type, position, COALESCE(comment, '') as comment
+    SELECT columns.id, columns.name, columns.nullable, columns.type, columns.position, COALESCE(comment, '') as comment
     FROM mz_catalog.mz_columns columns
+    LEFT JOIN mz_catalog.mz_objects obj ON obj.id = columns.id
     LEFT JOIN mz_internal.mz_comments comments
-    ON columns.id = comments.id AND columns.position = comments.object_sub_id",
+    ON columns.id = comments.id
+       AND columns.position = comments.object_sub_id
+       AND comments.object_type = obj.type",
```
ggevay, line 4993: For completeness: a few other views join `mz_comments` without an `object_type` predicate and had the same exposure, e.g. the ontology column-description join in `src/catalog/src/builtin/ontology.rs` and the `cts_obj`/`cts_col` joins in the SHOW CREATE views further down this file. The boot-time cleanup in this PR protects them in steady state (stale rows are gone before the upgraded version serves queries), so leaving them seems fine, but it might be worth a follow-up issue so this doesn't come back one view at a time.

test/sqllogictest/show_columns.slt
```diff
@@ -0,0 +1,64 @@
...
+
+mode standard
+
+reset-server
+
+# Each catalog item should report exactly one row per column from
+# SHOW COLUMNS, regardless of whether it's a Table or MaterializedView.
+
+query I
+SELECT COUNT(*) FROM (SHOW COLUMNS FROM mz_sources)
```
ggevay, line 24: These hardcoded counts will need updating every time someone adds a column to one of these builtins, and a bare count change won't tell that person whether they broke deduplication or just added a column. Something like `SELECT COUNT(*) = COUNT(DISTINCT name) FROM (SHOW COLUMNS FROM mz_sources)` asserts the no-duplicates property directly and never needs maintenance. Keeping one absolute count as an anchor seems plenty.

test/sqllogictest/show_columns.slt
```diff
@@ -0,0 +1,64 @@
...
+----
+a
+a_comment
+b
+(empty)
+
+# Regression: even if mz_comments holds rows for the same id under
+# multiple object_types (which can happen after a builtin type change
+# where the catalog id is preserved), SHOW COLUMNS must still emit
+# exactly one row per column. We can't directly inject a stale row
```
ggevay, line 62: The write-side scenario can actually be covered by a unit test, since `Op::Comment` doesn't validate that the `CommentObjectId` variant matches the item's actual type (that validation lives in the sequencer's planning). A test in `src/adapter/src/catalog.rs` following the `test_catalog_revision` pattern can open a debug catalog, `transact` a stale `CommentObjectId::Table(id)` comment for `mz_sources`, reopen on the same persist client, and assert the stale row is gone from `state.comments` while the legitimate materialized-view column comments survived. Note `skip_migrations: true` in debug opens only skips the user-item AST migrations, `add_new_remove_old_builtin_items_migration` still runs. Optional.

ggevay (approved):
LGTM, just minor comments

src/catalog/src/builtin/mz_internal.rs
mtabebe: Good point!

src/adapter/src/catalog/open.rs
mtabebe: Great catch

PR #37084, opened 2026-06-16
ci/console: restore CSP header injection in Vercel deploys

jasonhernandez:
## Problem

The console's `Content-Security-Policy` header is **not** committed in `console/vercel.json` — it's injected at build time by `console/bin/apply-vercel-csp.js`, which must run *before* `vercel build`.

In the original `MaterializeInc/console` repo, both the production (`deploy-prod.yml`) and preview (`deploy-preview.yml`) deploys went through `bin/build-ci`, which ran `apply-vercel-csp.js` before building.

When the console was ported into this repo (#35032), the deploy logic was reimplemented as Buildkite scripts that call `vercel build` directly and **dropped the `apply-vercel-csp.js` step**. `console/bin/build-ci` was carried over intact but is no longer invoked by anything.

As a result, the real console deploys ship with **no CSP header at all**:

| Deploy | Script | CSP injected? |
|---|---|---|
| Production (`console.materialize.com`) | `ci/deploy/console.sh` | ❌ before |
| Impersonation (`internal.console.materialize.com`) | `ci/deploy/console-impersonation.sh` | ❌ before |
| PR preview | `ci/test/console/vercel-preview.sh` | ❌ before |
| e2e test (throwaway preview) | `ci/test/console/e2e-prod.sh` | ✅ (unchanged) |

Confirmed live — `curl -sI https://console.materialize.com` returns `x-frame-options` and `referrer-policy` but **no `content-security-policy`**.

## Fix

Run `bin/apply-vercel-csp.js --sentry-release="$SENTRY_RELEASE"` before `vercel build` in all three real deploy scripts, matching the original `bin/build-ci` behavior. All three already export `SENTRY_RELEASE`.

The policy itself (`console/contentSecurityPolicy.js`) is unchanged.

## Verification

After this lands and deploys, `curl -sI https://console.materialize.com | grep -i content-security` should return the CSP header.

Closes SEC-606

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #37119, opened 2026-06-17
ci/console: time-bound Playwright browser install so hangs retry

jasonhernandez:
## Problem

The Console E2E jobs (`Console E2E/staging`, `Console E2E/prod`) install Playwright browsers via:

```bash
retry yarn playwright install --with-deps
```

The `retry` helper loops `until "$@"` — it only re-runs on a **non-zero exit**. When the Chromium download from the Playwright CDN *stalls* (rather than failing), the command hangs indefinitely with no output, so `retry` never kicks in and the job runs all the way to the **60-minute pipeline timeout** before being killed (`signal: terminated`, exit `-1`).

This was observed on [test build 125767](https://buildkite.com/materialize/test/builds/125767) (PR #37084), where both E2E shards hung at the Chromium download:
- prod — download reached 100% then went silent for ~60 min
- staging — download stalled mid-fetch then went silent for ~60 min

Neither shard reached the actual test or deploy steps; the failure was purely the browser install hanging.

## Fix

Wrap the install in `timeout -k 30 600` so a stalled fetch is killed after 10 minutes (SIGTERM, then SIGKILL 30s later), producing a non-zero exit that `retry` catches and re-runs. Worst case is 3 × 10 min = 30 min, still comfortably under the 60-min job timeout, versus a single hang burning the full hour.

Applied to both `ci/test/console/e2e.sh` and `ci/test/console/e2e-prod.sh`.

### DIFF
--- a/ci/test/console/e2e-prod.sh
+++ b/ci/test/console/e2e-prod.sh
@@ -44,7 +44,9 @@ corepack enable
 
 ci_collapsed_heading "Installing dependencies"
 retry yarn install --immutable --network-timeout 30000
-retry yarn playwright install --with-deps
+# `retry` only re-runs on a non-zero exit, but a stalled browser download
+# hangs forever; wrap in `timeout` so a hung fetch is killed and retried.
+retry timeout -k 30 600 yarn playwright install --with-deps
 ci_collapsed_heading "Pulling Vercel production environment"
 npx vercel@latest pull --yes --environment=production --token="$VERCEL_TOKEN"
 mv .vercel/.env.production.local .vercel/.env.preview.local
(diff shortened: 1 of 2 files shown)


def- (changes requested):
I don't think the installation is actually the problem. The download succeeded from what I can tell but some next step is hanging (which doesn't print anything). Brought it up in https://materializeinc.slack.com/archives/CL68GT3AT/p1781768367958089
Edit: But maybe I'm wrong too!

jasonhernandez:
@def- closing this because it looks like you got to the bottom of the issue with #37129. Thanks!

def-:
Interestingly your Claude was closer than mine initially!

PR #37126, opened 2026-06-18
Increase Prometheus introspection scrape interval from 1s to 10s

antiguru:
This change increases the default collection interval for Prometheus metrics introspection from 1 second to 10 seconds. This reduces the frequency of scraping operations, which should decrease overhead and resource consumption while still maintaining reasonable observability into system metrics.

### Description

The `COMPUTE_PROMETHEUS_INTROSPECTION_SCRAPE_INTERVAL` configuration default has been updated from `Duration::from_secs(1)` to `Duration::from_secs(10)`. This is a straightforward configuration adjustment that affects how often the Prometheus metrics introspection source collects data.

### Verification

No testing needed — this is a configuration constant change. The existing configuration system will handle the new default value appropriately.


def- (approved):
Oops, now I merged the nightly fixes already, you could remove it from test/launchdarkly-flag-consistency/mzcompose.py again

antiguru:
> Oops, now I merged the nightly fixes already, you could remove it from test/launchdarkly-flag-consistency/mzcompose.py again

I totally anticipated this! All good, I removed the override here.

PR #37127, opened 2026-06-18
txn-wal: Port txns_progress_source off the async operator builder

antiguru:
### Motivation

`txns_progress_source` was the last operator in txn-wal still built on the async Timely operator builder.
Its sibling `txns_progress_frontiers` is already synchronous, so this removes the remaining async operator from the crate.

### Description

The operator's awaits are real persist I/O (`open_writer`, `data_subscribe`, `rx.recv`), so it ports to a synchronous `build_reschedule` operator paired with a Tokio task that owns the persist resources and the subscription receiver.
The task forwards `DataRemapEntry` values over an unbounded channel and wakes the operator through an `ArcActivator`.
The `physical_upper` emit decision and the capability downgrades stay on the Timely thread, so the operator still emits only when the physical upper advances and downgrades to `logical_upper`, on a single chosen worker.
A `Finished` sentinel lets the operator distinguish a clean subscription close from a task panic.
Shutdown reproduces `builder_async`'s two-phase handshake via `button()` and `local_pressed`/`all_pressed`, matching `txns_progress_frontiers`.

No behavior change is intended.

### Verification

Covered by the existing in-memory persist integration tests that drive `txns_progress` end-to-end (`operator::tests::data_subscribe`, `as_of_until`, `subscribe_shard_finalize`, `subscribe_shard_register_forget`).

### DIFF
--- a/src/txn-wal/src/operator.rs
+++ b/src/txn-wal/src/operator.rs
@@ -12,6 +12,7 @@
 use std::any::Any;
 use std::fmt::Debug;
 use std::future::Future;
+use std::rc::Rc;
 use std::sync::mpsc::TryRecvError;
 use std::sync::{Arc, mpsc};
 use std::time::Duration;
@@ -29,10 +30,8 @@ use mz_persist_client::{Diagnostics, PersistClient, ShardId};
 use mz_persist_types::codec_impls::{StringSchema, UnitSchema};
 use mz_persist_types::txn::TxnsCodec;
 use mz_persist_types::{Codec, Codec64, StepForward};
-use mz_timely_util::builder_async::{
-    OperatorBuilder as AsyncOperatorBuilder, PressOnDropButton, button,
-};
-use timely::container::CapacityContainerBuilder;
+use mz_timely_util::activator::ArcActivator;
+use mz_timely_util::builder_async::{PressOnDropButton, button};
 use timely::dataflow::channels::pact::Pipeline;
 #[cfg(test)]
 use timely::dataflow::operators::Input;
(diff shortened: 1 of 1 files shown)


src/txn-wal/src/operator.rs
```diff
@@ -178,55 +188,145 @@ where
...
+                };
+            }
+
+            let Some((event_rx, activation_ack, task)) = chosen_state.as_mut() else {
+                // Non-chosen worker: nothing to do. Stay alive (the button
+                // channel reschedules us) for the shutdown handshake above.
+                return false;
+            };
+            // Keep the task alive for as long as the operator runs.
+            let _ = &*task;
```
aljoscha: The `let _ = &*task;` below is a no-op: the task stays alive because `chosen_state` owns it, not because of this reference. The statement only silences the unused-binding warning, and the `chosen_state` doc comment above already documents the lifetime/abort contract, so this comment is misleading. Bind `_task` in the pattern instead and drop the two lines.

```suggestion
            let Some((event_rx, activation_ack, _task)) = chosen_state.as_mut() else {
                // Non-chosen worker: nothing to do. Stay alive (the button
                // channel reschedules us) for the shutdown handshake above.
                return false;
            };
```

src/txn-wal/src/operator.rs
```diff
@@ -178,55 +188,145 @@ where
...
+                        cap.downgrade(&logical_upper);
+                    }
+                    Ok(SourceEvent::Finished) => {
+                        // Subscription closed cleanly. Drop the capability.
+                        finished = true;
+                        capability = None;
+                    }
+                    Err(tokio::sync::mpsc::error::TryRecvError::Empty) => break,
+                    Err(tokio::sync::mpsc::error::TryRecvError::Disconnected) => {
+                        assert!(finished, "txns_progress_source task unexpectedly gone");
```
aljoscha, line 323: nit (probably fine to keep as is): this turns a task panic into a deferred channel-disconnect assertion rather than a panic on the worker thread, and on the panic path the task never calls `activate()`, so the assertion only fires if the operator is rescheduled for another reason. In production this is masked because the enhanced panic handler `process::abort()`s on any task panic, so the loud-crash behavior is preserved and this assertion is really just a safety net for non-abort (test) environments. No change needed, but a one-line note to that effect would save the next reader the trip.

src/txn-wal/src/operator.rs
antiguru: Done, bound `_task` and dropped the two lines.

src/txn-wal/src/operator.rs
antiguru: Added a note: the enhanced panic handler aborts the process on a task panic, so the assert is a safety net for non-abort environments, and on the panic path the task never calls `activate()`.

aljoscha (approved):
I _think_ this is correct 🤞 it's also much simpler than the other operators you de-asyncified.

antiguru:
Thanks for the review!

PR #37130, opened 2026-06-18
repr: fix panic decoding an empty Avro decimal value

def-:
`twos_complement_be_to_numeric_inner` read `input[0]` to determine the sign before checking that the input was non-empty, so a zero-length two's-complement byte run panicked with an out-of-bounds index.

This is reachable from Avro source ingestion: a `decimal` logical type backed by `bytes` (or `fixed`) whose value is encoded as a zero-length byte array is a valid wire form, and that data arrives from external, possibly hostile, schema registries and producers. A conformant producer encodes zero as `[0x00]`, but a non-conformant one can send empty bytes, crashing the decoder and so the source.

An empty big-endian two's-complement run canonically denotes 0. Guarding the sign check with `!input.is_empty()` makes the existing logic produce 0 for that case (`head == 0`, no chunks), with no other change needed. The only caller is the Avro decimal decode path.

Found by the avro_decode_fuzzed_schema cargo-fuzz target.

Tests: adds `test_twos_complement_empty_is_zero`, covering the wrapper at several scales and the generic inner at both numeric widths.

Release note: Fix a panic in Avro-formatted sources when decoding a `decimal` column whose value was encoded as a zero-length byte array.

### DIFF
--- a/src/repr/src/adt/numeric.rs
+++ b/src/repr/src/adt/numeric.rs
@@ -399,7 +399,12 @@ pub fn twos_complement_be_to_numeric(
 pub fn twos_complement_be_to_numeric_inner<D: Dec<N>, const N: usize>(
     input: &mut [u8],
 ) -> Result<Decimal<N>, anyhow::Error> {
-    let is_neg = if (input[0] & 0x80) != 0 {
+    // An empty big-endian two's-complement run encodes the value 0 (Avro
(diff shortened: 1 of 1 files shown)


def-:
This fix is already part of https://github.com/MaterializeInc/materialize/pull/36984. Really need to get this stuff merged or I'll keep rerunning into the same bugs.

PR #37134, opened 2026-06-18
compute: scope per-batch decode arenas to the activation

frankmcsherry:
## What

Moves the `RowArena` (and accompanying `DatumVec`) used for per-row decode in the `ArrangementFlatMap`, `ArrangementFlatMapOk`, and `FormArrangementKey` operators from operator-**build** scope into the per-**activation** closure, so it is dropped at the end of each scheduling invocation while still being reused across the records processed within that activation.

## Why

These three operators created their decode arena once at build time and captured it for the operator's lifetime, clearing it per record. That's the only place in the compute/storage layer (verified by an exhaustive audit of every `RowArena.clear()` site) where an arena is both **clear-reused** and **dataflow-lived**.

It matters because of the companion change that makes `RowArena::clear()` retain its largest region for reuse (#37114): under that semantics, a build-captured + per-row-cleared arena pins its high-water region for the operator's whole life — a small (one record's worth) but *permanent* per-operator memory tax. Scoping the arena to the activation eliminates the tax (the region is freed when the activation returns) while keeping the within-activation reuse that retention buys.

The discipline this establishes — *fresh arena per scheduling invocation, cleared per record* — is the one already used by `LinearJoinKeyPreparation`; this brings the remaining clear-reuse arenas in line with it.

## How

- `FormArrangementKey`: a one-line move of `RowArena::new()` from the build closure into the `move |_frontiers|` closure.
- `flat_map` / `flat_map_ok`: the per-row decode wrapper (which owned the arena + `DatumVec`) is removed; `flat_map_core_{fallible,ok}` now take `max_demand` and the **decoded-form** `logic` directly and build the decode closure inside their per-activation closure. `do_work`'s `logic` bound drops a vestigial `+ 'static` (never required by `walk_cursor`) so it can accept the activation-scoped, borrowing decode closure.

No public API change; no behavior change — the decode and emitted results are identical. This is intended to land before #37114 so the region-retention change can never introduce a dataflow-lived arena tax.

## Tests

Covered by existing compute unit tests and the SLT/testdrive suites that exercise indexed MFPs, table functions, and arrangement formation. No behavior change, so no new tests.

### DIFF
--- a/src/compute/src/render/context.rs
+++ b/src/compute/src/render/context.rs
@@ -300,7 +300,7 @@ impl<'scope, T: RenderTimestamp> ArrangementFlavor<'scope, T> {
         &self,
         key: Option<&Row>,
         max_demand: usize,
-        mut logic: L,
+        logic: L,
     ) -> (
         Stream<'scope, T, DCB::Container>,
         VecCollection<'scope, T, DataflowErrorSer, Diff>,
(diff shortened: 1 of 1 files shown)


src/compute/src/render/context.rs
```diff
@@ -1151,8 +1173,10 @@ impl<'scope, T: RenderTimestamp> CollectionBundle<'scope, T> {
             let mut key_buf = Row::default();
             let mut val_buf = Row::default();
             let mut datums = DatumVec::new();
```
antiguru, line 1175: In a follow-up, we could move these allocations into the subsequent closure, too.

antiguru (approved):
Looks good, thank you! I left a comment for a follow-up change.

PR #37135, opened 2026-06-18
storage: clear the oneshot-source decode arena per row

frankmcsherry:
## What

`render_decode_chunk` (the oneshot/COPY decode operator) created one `RowArena` in the operator body and reused it across every chunk and every row **without ever clearing it**. The arena owns whatever is pushed into it for its lifetime, so the MFP-evaluation temporaries for the *entire source* accumulated until the COPY finished — an unbounded arena proportional to the data volume.

This scopes the arena to each decoded chunk and clears it per row, so it holds at most one row's worth of temporaries and is released when the chunk is processed.

## Why

Found while auditing `RowArena` lifetimes (companion to the arena work in #37114 / #37134). This was the only arena in the audit that was both operator-lived **and** never cleared — i.e. a genuine unbounded accumulation rather than a bounded high-water. It brings the operator in line with the "arena per batch, cleared per row" idiom the other operators use.

## Tests

No behavior change (the MFP output is identical; only the arena's lifetime changes). Covered by the existing oneshot-source / COPY FROM tests.

### DIFF
--- a/src/storage-operators/src/oneshot_source.rs
+++ b/src/storage-operators/src/oneshot_source.rs
@@ -486,7 +486,6 @@ where
         let [_row_cap] = caps.try_into().unwrap();
 
         let mut datum_vec = DatumVec::default();
-        let row_arena = RowArena::default();
         let mut row_buf = Row::default();
 
         while let Some(event) = record_chunk_handle.next().await {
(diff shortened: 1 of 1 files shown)


martykulma (approved):
🦈

PR #37136, opened 2026-06-18
doc: collect row-encoding size limits

antiguru:
### Motivation

The size limits the in-memory `Row` encoding imposes are spread across unrelated constants (array dimensions, char/varchar length, numeric precision, regex size, the length-tiered string/bytes/list tags) with no single reference.

### Description

Adds `doc/developer/row-encoding.md` collecting them, each pointing at its enforcing constant and noting whether overflow is a graceful error, a panic, or unbounded. Covers per-datum limits and whole-`Row` bounds: arrangements store rows with `u64` offsets, while persist's Arrow encoding uses `i32` offsets, capping an encoded row at ~2 GiB. Links the doc from the developer docs index and adds a module doc on `row.rs`.

### DIFF
--- a/doc/developer/README.md
+++ b/doc/developer/README.md
@@ -70,6 +70,9 @@ land, and then browsed as reference material as you skill up on the codebase.
 
 * [reading.md](reading.md) contains a reading list.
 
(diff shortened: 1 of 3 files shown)


def- (approved):
No concerns from QA side

PR #37137, opened 2026-06-18
[catalog] bump mz_indexes for mz_cluster_replica_size_internal_ind

mtabebe:
A new builtin index was added at 26.30.0-dev.0 without bumping mz_indexes. make_mz_indexes inlines the builtin-index set as VALUES, so its SQL fingerprint changed without a matching migration step, causing 0dt upgrades from 26.29.0 to panic in update_fingerprints.

Re-enables the upgrade-smoke-test that was disabled in #37118.

Fixes SQL-398.


def- (approved):
I also disabled upgrade tests in nightly, reenabled them.

PR #37140, opened 2026-06-18
Re-enable flaky FoundationDB unit tests

antiguru:
### Motivation

`fdb_consensus` and `test_fdb_timestamp_oracle` were `#[ignore]`'d because they timed out sporadically (nextest terminated them at 180s).

Fixes database-issues#10076
Closes DB-103

### Description

The FoundationDB network can be booted once per process and never restarted.
Stopping it joins the network thread, which can block indefinitely when a `Database` or in-flight transaction is still alive; `fdb_consensus` called `shutdown_network()` while its `Database` was still in scope.
Conversely, skipping the shutdown lets the process tear down with the network thread still running, which segfaults the client.
Under nextest each test runs in its own process and each of these crates has a single network-using test, so booting once and shutting down once per process is correct.

Re-enable both tests and tear down cleanly:

* Drop all `Database` handles before calling `shutdown_network()`, so the network thread join does not block. `fdb_consensus` now drops its handle first; the oracle's handles are already dropped by the time it shuts down.
* Correct the `init_network` / `shutdown_network` documentation: boot once, drop handles, then shut down before exit.
* A 60s default transaction timeout on the database, applied only in tests, so a genuinely unresponsive server surfaces a prompt `transaction_timed_out` error instead of hanging. Production behavior is unchanged.
* `retries = 2` on both tests, so a transient stall reruns rather than failing the build.

### Verification

The flake is sporadic and load-dependent, so CI across repeated runs is the verification surface.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -7,6 +7,18 @@ slow-timeout = { period = "1200s", terminate-after = 2 }
 # For a given configuration parameter, the first override to match wins. Keep
 # these sorted in order from most specific to least specific.
 
+# The FoundationDB unit tests talk to an external FDB server; retry them so a
+# transient server stall reruns rather than failing the build (DB-103).
+[[profile.default.overrides]]
+filter = "package(mz-persist) and test(foundationdb::tests::fdb_consensus)"
+retries = 2
+priority = 100
+
+[[profile.default.overrides]]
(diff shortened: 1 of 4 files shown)


def- (approved):
Closes DB-103 I assume

PR #37142, opened 2026-06-18
mz-deploy: correctness and ergonomics fixes

sjwiesman:
A batch of independent correctness and ergonomics fixes for `mz-deploy`.

### How to review

**Each commit is a self-contained semantic change and can be reviewed on its
own.** Every commit is either *trivially correct* (a docs/string/lockfile change
with no behavioral risk) or *provably correct* against the test that ships in the
same commit. Reviewing commit-by-commit is the intended path.

### Commits

1. **run a source-table's grants/comments outside the create transaction**
   `CREATE TABLE ... FROM SOURCE` and its companion `GRANT`/`COMMENT` were bundled
   into one transaction, which the server rejects — the deploy failed. Splits the
   grants and comments out of the create transaction. *Provably correct:*
   regression test exercises a source table with a grant.

2. **store object identifiers as Idents so reserved-word names work**
   Object names that are SQL reserved words (or otherwise need quoting) were
   stored as raw strings and broke end-to-end. Switches the identity type to a
   typed `Ident`, quoting only at the edges. *Provably correct:* tests cover
   reserved-word object names.

3. **check in regenerated types.lock files**
   Ships the reserved-words test project's `types.lock` alongside the other test
   fixtures. *Trivially correct:* checked-in generated lockfile, no code change.

4. **correct stale `--name` flag in error hints**
   An error hint referenced a `--name` flag that no longer exists. Fixes the hint
   text. *Trivially correct:* user-facing string only.

5. **make init idempotent**
   `mz-deploy init` was not quite idempotent — re-running it in an initialized
   directory could error. Makes it a no-op on already-initialized projects.
   *Provably correct:* test runs `init` twice and in an existing git repo.

6. **drop nonexistent files from init/new help text**
   `init`/`new` help listed files the commands don't create. Trims the help to
   what is actually generated. *Trivially correct:* help text only.

7. **reject indexes on tables and sources**
   An index on a table/source could silently be dropped at promote when its
   cluster participated in a swap. Rejects indexes on tables and sources up front
   with a clear error hinting at creating a view and indexing that. *Provably
   correct:* test asserts the rejection and message.

8. **stop describing Docker as a type-checking tool**
   Documentation incorrectly described Docker as performing type checking. Docs
   wording fix. *Trivially correct:* docs only.

9. **add `stage --redeploy-schema` and `--redeploy-all`**
   Adds flags to force one or more schemas (or the whole project) to redeploy even
   when unchanged, by seeding the dirty-propagation fixed point with forced
   schemas. This commit also **fixes the replacement-MV (stable schema) dirty
   propagation**: the special "don't propagate" rule for stable MVs belongs only
   on downstream dependency propagation, not on schema atomicity. The corrected
   semantics: a dirty MV in a stable schema dirties all views in that schema; a
   dirty stable schema redeploys all its MVs atomically; a dirty stable MV's
   cluster still propagates; only *downstream* fan-out from a dirty stable MV is
   suppressed. *Provably correct:* unit tests for each rule (including inverted
   tests for the old behavior and a new downstream-skip test), plus the
   `redeploy-flags` and `replacement-mvs` mzcompose workflows. The rules were
   first validated as a runnable Soufflé datalog spec before porting to Rust.

10. **test promote crash recovery via fault injection**
    Adds a test-only crash hook and an mzcompose workflow that kills the process
    at chosen points during promote to exercise the resume paths. *Trivially
    correct:* test-only infrastructure; no production behavior change.

11. **resume promote after a crash between promotion and cleanup**
    A crash after a deploy was promoted but before its marker schemas were cleaned
    up left the deploy un-resumable — validation rejected the already-promoted
    state. Gates that validation on the not-yet-started state so promote resumes
    cleanly. *Provably correct:* the fault-injection test from commit 10 drives a
    crash at exactly this point and asserts a clean resume.


sjwiesman:
Splitting this into two focused PRs: one for the init/help/docs ergonomics fixes, and one for the substantive correctness fixes + forced-redeploy flags. Replaced by the two PRs linked below.

sjwiesman:
Replaced by #37231 (init/help/docs fixes) and #37232 (correctness fixes + forced-redeploy flags).

PR #37145, opened 2026-06-18
sql: parenthesize printing fixes

def-:
All found via https://github.com/MaterializeInc/materialize/pull/36982

### DIFF
--- a/src/sql-parser/src/ast/defs/expr.rs
+++ b/src/sql-parser/src/ast/defs/expr.rs
@@ -246,7 +246,17 @@ impl<T: AstInfo> AstDisplay for Expr<T> {
                 if *negated {
                     f.write_str("NOT ");
                 }
-                f.write_node(construct);
+                // `IS DISTINCT FROM <rhs>` parses the RHS at the `IS` precedence
+                // (see `Parser::parse_is`), so a RHS whose left spine binds at or
+                // below `IS` re-associates out of the `IS` unless parenthesized
+                // (`a IS DISTINCT FROM b OR c` is `(a IS DISTINCT FROM b) OR c`).
+                // The other constructs (`NULL`/`TRUE`/…) are bare keywords.
+                if let IsExprConstruct::DistinctFrom(rhs) = construct {
+                    f.write_str("DISTINCT FROM ");
+                    write_binary_operand(f, rhs, left_edge(rhs) <= prec::IS);
+                } else {
+                    f.write_node(construct);
+                }
             }
             Expr::InList {
                 expr,
--- a/src/sql-parser/src/ast/defs/expr.rs
+++ b/src/sql-parser/src/ast/defs/expr.rs
@@ -220,26 +220,28 @@ impl<T: AstInfo> AstDisplay for Expr<T> {
             Expr::Parameter(n) => f.write_str(&format!("${}", n)),
             Expr::Not { expr } => {
                 f.write_str("NOT ");
-                // `NOT` binds tighter than `AND`/`OR`, so a looser operand
-                // (`NOT (a OR b)`) must keep its parens once `Nested` is stripped.
-                write_binary_operand(f, expr, expr_precedence(expr) < prec::NOT);
+                // `NOT` binds tighter than `AND`/`OR`, so an operand exposing a
+                // looser operator on its left spine (`NOT (a OR b)`) must keep its
+                // parens once `Nested` is stripped — use `left_edge`, since the
+                // `NOT` sits to the operand's left.
+                write_binary_operand(f, expr, left_edge(expr) < prec::NOT);
             }
             Expr::And { left, right } => {
-                write_binary_operand(f, left, expr_precedence(left) < prec::AND);
+                write_binary_operand(f, left, right_edge(left) < prec::AND);
                 f.write_str(" AND ");
-                write_binary_operand(f, right, expr_precedence(right) <= prec::AND);
+                write_binary_operand(f, right, left_edge(right) <= prec::AND);
             }
             Expr::Or { left, right } => {
-                write_binary_operand(f, left, expr_precedence(left) < prec::OR);
+                write_binary_operand(f, left, right_edge(left) < prec::OR);
                 f.write_str(" OR ");
-                write_binary_operand(f, right, expr_precedence(right) <= prec::OR);
+                write_binary_operand(f, right, left_edge(right) <= prec::OR);
             }
             Expr::IsExpr {
                 expr,
                 negated,
                 construct,
             } => {
-                write_binary_operand(f, expr, expr_precedence(expr) < prec::IS);
+                write_binary_operand(f, expr, right_edge(expr) < prec::IS);
                 f.write_str(" IS ");
                 if *negated {
                     f.write_str("NOT ");
@@ -251,7 +253,7 @@ impl<T: AstInfo> AstDisplay for Expr<T> {
                 list,
                 negated,
             } => {
-                write_binary_operand(f, expr, expr_precedence(expr) < prec::LIKE);
+                write_binary_operand(f, expr, right_edge(expr) < prec::LIKE);
                 f.write_str(" ");
                 if *negated {
                     f.write_str("NOT ");
@@ -265,7 +267,7 @@ impl<T: AstInfo> AstDisplay for Expr<T> {
                 subquery,
                 negated,
             } => {
-                write_binary_operand(f, expr, expr_precedence(expr) < prec::LIKE);
+                write_binary_operand(f, expr, right_edge(expr) < prec::LIKE);
                 f.write_str(" ");
                 if *negated {
                     f.write_str("NOT ");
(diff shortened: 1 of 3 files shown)


src/sql-parser/src/ast/defs/expr.rs
```diff
@@ -549,29 +617,15 @@ fn write_dot_receiver<W: fmt::Write, T: AstInfo>(f: &mut AstFormatter<W>, expr:
...
-            | Expr::InSubquery { .. }
-            | Expr::And { .. }
-            | Expr::Or { .. }
-            | Expr::Not { .. }
-            | Expr::AnyExpr { .. }
-            | Expr::AllExpr { .. }
-            | Expr::AnySubquery { .. }
-            | Expr::AllSubquery { .. }
-    );
+    let needs_parens = right_edge(expr) <= prec::LIKE;
```
aljoscha: Correctness: this fixed threshold under-parenthesizes when the quantified operator binds tighter than `LIKE`. The `<op>` in `<left> <op> ANY/ALL (...)` can be any binary operator, including `+ - * / %`, which bind tighter than `LIKE`. For example `SELECT (a + b) * ANY (ARRAY[c])` with `Nested` stripped prints as `SELECT a + b * ANY (ARRAY[c])`, which reparses as `a + (b * ANY (ARRAY[c]))`, a structurally and semantically different expression. I verified this fails the round trip today.

The threshold should use the actual operator precedence, mirroring the binary-`Op` arm. Threading `op` in and using a strict `<` fixes the arithmetic cases and also stops over-parenthesizing comparison lefts (`a = b = ANY (...)`, `a LIKE b = ANY (...)`):

```rust
fn write_quantified_left<W: fmt::Write, T: AstInfo>(f: &mut AstFormatter<W>, expr: &Expr<T>, op: &Op) {
    let needs_parens = right_edge(expr) < binary_op_precedence(op);
    ...
}
```

(pass `op` from the four `Any*`/`All*` call sites). I patched this locally and confirmed it round-trips all the cases above. This is technically pre-existing, but since the PR rewrote this function and targets exactly this family of bugs, folding the fix plus a `(a + b) * ANY (...)` test in here seems right. Open to a follow-up instead if you would rather keep the PR tight.

aljoscha (commented):
Local review against our standards. Overall this is a careful, well-tested improvement to codebase health. I ran the `sqlparser_common` suite locally (36 tests pass, including the `datadriven` corpus, which confirms parser-produced ASTs still print byte-identically, so there is no golden-file churn and no user-visible reprint change). One correctness gap and a few style/cleanup items, posted inline.

Nits not anchored to a line:
- Title (and branch) typo: "paranthesize" should be "parenthesize". The commit subjects spell it correctly.
- The PR description is just the #36982 link. That is fine for a semantic-commits PR given the strong commit messages, but a one-liner such as "internal round-trip robustness, no user-visible reprint change, see commit messages" would help the release-notes classifier skip it cleanly.
- `prec::{COLLATE, POSTFIX}` are `#[allow(dead_code)]` placeholders. Well justified by the comment (keeps `PREFIX` at the parser's relative position), fine to leave as is.

Things I checked that are fine: parser ASTs unaffected, `binary_op_precedence`/`unary_prec` faithfully mirror `get_next_precedence`/`parse_prefix`, `prec` ordering matches the parser enum, single CODEOWNERS area, good per-commit test coverage.

src/sql-parser/src/ast/defs/expr.rs
```diff
@@ -220,36 +220,50 @@ impl<T: AstInfo> AstDisplay for Expr<T> {
             Expr::Parameter(n) => f.write_str(&format!("${}", n)),
             Expr::Not { expr } => {
                 f.write_str("NOT ");
-                f.write_node(expr);
+                // `NOT` binds tighter than `AND`/`OR`, so an operand exposing a
+                // looser operator on its left spine (`NOT (a OR b)`) must keep its
+                // parens once `Nested` is stripped — use `left_edge`, since the
```
aljoscha: Style (applies throughout the new comments, this line is just one example): our house style avoids em-dashes and semicolons for structuring sentences, in code comments too. There are roughly 10 em-dashes and 11 prose semicolons across the added comments. Could you split these into separate sentences with full stops and commas. For example here: "...must keep its parens once `Nested` is stripped. We use `left_edge`, since the `NOT` sits to the operand's left."

src/sql-parser/src/ast/defs/expr.rs
```diff
@@ -581,6 +635,182 @@ fn write_quantified_left<W: fmt::Write, T: AstInfo>(f: &mut AstFormatter<W>, exp
...
+        f.write_node(bound);
+    }
+}
+
+/// Output-precedence ranks, ordered like `Parser::get_next_precedence` /
+/// `Parser::Precedence` (higher binds tighter). They classify the *top* operator
+/// an expr prints with, so the binary-operator printer can parenthesize an
+/// operand that would otherwise re-associate on reparse. Keep in sync with the
+/// parser. `ATOM` covers the self-delimiting primaries — they never need parens.
+mod prec {
```
aljoscha, line 679: Design (not blocking): this `prec` table duplicates the parser's `Precedence` enum and has to be kept in sync by hand, which the comment itself flags as a hazard. Since both live in `mz-sql-parser`, consider exposing the parser's `Precedence` as `pub(crate)` and deriving these ranks from it, with the `ATOM` sentinel layered on top, so there is a single source of truth. The #36982 fuzzing is a safety net against drift, so this is a judgment call worth a sentence rather than a must-fix.

src/sql-parser/src/ast/defs/expr.rs
```diff
@@ -300,22 +325,40 @@ impl<T: AstInfo> AstDisplay for Expr<T> {
...
+                    // right) reaches into its right spine at a strictly looser
+                    // precedence (`right_edge`); the right operand needs parens
+                    // when the operator (printed to its left) reaches into its
+                    // left spine at an equal-or-looser precedence (`left_edge`,
+                    // `<=` because equal precedence re-associates left, so
+                    // `a - (b - c)` must keep its parens). The parser wraps such
+                    // operands in `Expr::Nested` (which ranks `ATOM` on both
+                    // edges, so it never re-parenthesizes), leaving parser-produced
+                    // ASTs unchanged; this re-adds parens for ASTs that lost them.
+                    let p = binary_op_precedence(op);
```
aljoscha, line 350: Nit: this inline comment (about 12 lines) restates the left/right-edge reasoning that already lives on the `left_edge` and `right_edge` doc comments. Consider trimming it to a one-liner that points at those helpers, so the same explanation is not maintained in three places.

src/sql-parser/src/ast/defs/expr.rs
```diff
@@ -300,22 +325,34 @@ impl<T: AstInfo> AstDisplay for Expr<T> {
...
+                write_between_bound(f, low);
                 f.write_str(" AND ");
-                f.write_node(&high);
+                write_between_bound(f, high);
             }
             Expr::Op { op, expr1, expr2 } => {
                 if let Some(expr2) = expr2 {
-                    f.write_node(&expr1);
+                    // Binary operators are left-associative: parenthesize an
+                    // operand that would re-associate once `Nested` is stripped —
```
aljoscha: Leftover em-dash in the freshly-trimmed comment. Could split into two sentences, e.g. "...re-associate once `Nested` is stripped. The left by its `right_edge`..."

src/sql-parser/src/parser.rs
```diff
@@ -343,8 +343,12 @@ const SPECULATIVE_FAILURES_PER_TOKEN: usize = 100;
 /// Defines a number of precedence classes operators follow. Since this enum derives Ord, the
 /// precedence classes are ordered from weakest binding at the top to tightest binding at the
 /// bottom.
+///
+/// The expression printer's `ast::defs::expr::prec` ranks are derived from this
+/// ladder via `as u8`, so this enum is the single source of truth for output
+/// parenthesization precedence too — keep the ordering authoritative.
```
aljoscha: Leftover em-dash in this new doc. e.g. "...parenthesization precedence too. Keep the ordering authoritative."

src/sql-parser/src/ast/defs/expr.rs
```diff
@@ -581,6 +641,191 @@ fn write_quantified_left<W: fmt::Write, T: AstInfo>(f: &mut AstFormatter<W>, exp
...
+}
+
+/// Output-precedence ranks, derived directly from the parser's [`Precedence`]
+/// ladder (higher binds tighter) so it stays the single source of truth:
+/// reordering or inserting a parser level reranks these automatically, and only
+/// the variant each rank maps to is maintained by hand. They classify the *top*
+/// operator an expr prints with, so the binary-operator printer can parenthesize
+/// an operand that would otherwise re-associate on reparse. `ATOM`, the one rank
+/// with no parser counterpart, is layered one above the tightest parser level to
+/// mark the self-delimiting primaries; they never need parens.
```
aljoscha: Leftover prose semicolon in this new doc comment. e.g. "...to mark the self-delimiting primaries. They never need parens."

aljoscha (approved):
LGTM, approving.

The correctness gap in `write_quantified_left` is genuinely fixed (real operator precedence instead of a fixed `Like` threshold), and the added tests cover both the under-parenthesization case it fixes and the over-parenthesization it stops. The `prec` table now derives from the parser's `Precedence` via `as u8`, so there is a single source of truth. I re-ran the `sqlparser_common` suite locally (37 tests pass, including the `datadriven` corpus) and `cargo check` is clean.

One trivial leftover: the em-dash/semicolon cleanup missed three spots, two of which are in text newly written for this commit. Flagged inline. Fine to fix on the way in, not blocking the approval.

PR #29673, opened 2024-09-19
adapter: synthesize and apply side effects (aka. controller commands) from catalog changes

aljoscha:
First step towards https://github.com/MaterializeInc/materialize/issues/29279

With this we only wire up the new side-effects logic for
creating/dropping tables.

### Checklist

- [ ] This PR has adequate test coverage / QA involvement has been duly considered. ([trigger-ci for additional test/nightly runs](https://trigger-ci.dev.materialize.com/))
- [ ] This PR has an associated up-to-date [design doc](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/README.md), is a design doc ([template](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/design/00000000_template.md)), or is sufficiently small to not require a design.
  
- [ ] If this PR evolves [an existing `$T ⇔ Proto$T` mapping](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/command-and-response-binary-encoding.md) (possibly in a backwards-incompatible way), then it is tagged with a `T-proto` label.
- [ ] If this PR will require changes to cloud orchestration or tests, there is a companion cloud PR to account for those changes that is tagged with the release-blocker label ([example](https://github.com/MaterializeInc/cloud/pull/5021)).
  
- [ ] If this PR includes major [user-facing behavior changes](https://github.com/MaterializeInc/materialize/blob/main/doc/developer/guide-changes.md#what-changes-require-a-release-note), I have pinged the relevant PM to schedule a changelog post.

PR #37151, opened 2026-06-19
adapter: derive replica-scoped override push from catalog implications

antiguru:
### Motivation

Scoped feature flags need the per-replica controller push to be *derived from**<br>**the committed catalog diff* rather than issued as a side effect in the<br>sequencer. Deriving it removes the working-copy/controller divergence that lets<br>a create-time replica override be wiped by a reconcile in the same DDL, and it<br>makes the push fire for a follower `environmentd` that only replays the catalog<br>diff. This is the pattern the catalog-implications framework ([#29673](<https://github.com/MaterializeInc/materialize/issues/29673>),<br>`TODO(aljoscha)`) exists to provide, which until now ignored config-type<br>changes.

### Description

`parse_state_update` no longer drops `ReplicaSystemConfiguration` updates; it<br>emits a `ParsedStateUpdateKind::ReplicaSystemConfiguration`. When such an update<br>is present in a transaction, `apply_catalog_implications` rebuilds the complete<br>per-replica compute dyncfg layer from the catalog working copy and pushes it<br>after clusters are created and before replicas are created, so a new replica's<br>first configuration replays with its override. The same push runs once on<br>bootstrap from the restored working copy.

Because the push reads the working copy that the same transaction's catalog<br>apply has already updated, the diff drives the push. The push mechanism<br>(`push_replica_dynogcfg_overrides`) lives here. Cluster-scoped overrides are read<br>at plan time and have no controller effect, so they remain unparsed.

This is the foundation under MaterializeInc/materialize#36959, which adds the writers that drive the<br>implication; the implication is dormant until that PR lands on top.

### Verification

Adds a unit test asserting a `ReplicaSystemConfiguration` update is parsed<br>rather than dropped. End-to-end controller behavior is covered by the<br>LaunchDarkly mzcompose test in MaterializeInc/materialize#36959.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

[https://claude.ai/code/session_01FUorxx5m6B8TT2VAek2STo](<https://claude.ai/code/session_01FUorxx5m6B8TT2VAek2STo>)

### DIFF
--- a/src/adapter/src/coord.rs
+++ b/src/adapter/src/coord.rs
@@ -117,6 +117,7 @@ use mz_controller::clusters::{
 };
 use mz_controller::{ControllerConfig, Readiness};
 use mz_controller_types::{ClusterId, ReplicaId, WatchSetId};
+use mz_dyncfg::ConfigUpdates;
 use mz_expr::{MapFilterProject, MirRelationExpr, OptimizedMirRelationExpr, RowSetFinishing};
 use mz_license_keys::{ExpirationBehavior, ValidatedLicenseKey};
 use mz_orchestrator::OfflineReason;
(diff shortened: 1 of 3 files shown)


src/adapter/src/coord.rs
```diff
@@ -2078,6 +2152,14 @@ impl Coordinator {
             .update_orchestrator_scheduling_config(scheduling_config);
         self.controller.update_configuration(dyncfg_updates);
 
+        // The scoped (per-cluster and per-replica) system-parameter working copy
+        // was restored from the durable cache into `CatalogState` while opening
+        // the catalog, so the last-known values are in effect before the first
+        // parameter sync and through a sync outage. Push the replica-local
+        // overrides into the compute controller so existing replicas observe them
+        // at startup. The cluster-coherent layer is read at plan time.
+        self.push_replica_dyncfg_overrides();
```
aljoscha: **Blocking:** this push is ineffective at bootstrap. It runs before the `create_cluster`/`create_replica` loop below (~2194-2215), so `update_replica_dyncfg_overrides` iterates an empty `self.instances` and drops every override. `create_instance` then builds each instance with empty overrides and applies only the env-wide `self.config`, so existing replicas never observe their overrides at startup (there is no controller-level stash of overrides for future instances, unlike `self.config`).

The DDL path works precisely because it pushes *after* `create_cluster`. Suggest moving this call to after the creation loop (after line 2215); `update_configuration`'s re-push then reaches the already-created replicas. No test currently covers this path.

src/adapter/src/coord/catalog_implications/parsed_state_updates.rs
```diff
@@ -64,6 +64,13 @@ pub enum ParsedStateUpdateKind {
         log: LogVariant,
         index_id: GlobalId,
     },
+    /// A replica-scoped system-parameter override changed. Carries no parsed
+    /// form: the implication re-pushes the complete per-replica dyncfg layer
+    /// from the catalog working copy, so it only needs to know that a
+    /// replica-scoped row changed.
```
aljoscha: `durable` is never read (matched as `_`/`..` in `apply_catalog_implications`, `absorb`, and the test). Keeping it is fine, but the doc says it "carries no parsed form," which reads as contradictory. Suggest wording that says the row is retained only for trace output:

```suggestion
    /// A replica-scoped system-parameter override changed. The implication
    /// re-pushes the complete per-replica dyncfg layer from the catalog working
    /// copy, so it does not consume `durable`. We keep the row only so it shows
    /// up in the `tracing::trace!` of the parsed update.
```

src/adapter/src/coord.rs
```diff
@@ -2013,6 +2014,79 @@ pub struct Coordinator {
...
+        self.controller
+            .compute
+            .update_replica_dyncfg_overrides(instance_overrides);
+        // Re-push the environment-wide compute configuration so existing replicas
+        // pick up their (possibly changed) overrides. This is also what *reverts*
+        // a removed override: the per-replica layer no longer carries the key, so
+        // the replica falls back to the value in this push. Correct revert
+        // therefore relies on `compute_config` being the complete env-wide set
+        // (it is: `compute_config` renders the full dyncfg set), so the
+        // formerly-overridden key is always present here with its env-wide value.
```
aljoscha: nit: dense clause-stacking. Reads cleaner as two sentences without the parenthetical:

```suggestion
        // Re-push the env-wide compute config so existing replicas pick up their
        // (possibly changed) overrides. This also reverts a removed override: the
        // per-replica layer no longer carries the key, so the replica falls back
        // to the env-wide value, which `compute_config` always includes because it
        // renders the full dyncfg set.
```

aljoscha (commented):
Reviewed against our change guidelines. One blocking correctness issue on the bootstrap path, plus a few comment cleanups. Details inline.

src/adapter/src/coord.rs
```diff
@@ -2013,6 +2014,79 @@ pub struct Coordinator {
 }
 
 impl Coordinator {
+    /// Resolves the replica-local scoped overrides from the catalog working copy
+    /// into the compute controller's per-replica dyncfg layer, then re-pushes
+    /// the environment-wide compute configuration so replicas observe the new
+    /// values. Driven by the catalog implication for replica-scoped
+    /// configuration changes, and called once on bootstrap.
+    ///
+    /// Only the compute controller's per-replica dyncfg layer is pushed, but on
```
aljoscha: nit: this paragraph is valuable, but it is reasoning about *why* pushing only the compute layer suffices. Per our comment conventions that fits better as an inline `//` note next to the `update_replica_dyncfg_overrides`/`update_configuration` calls than hoisted into the rustdoc contract.

src/adapter/src/coord.rs
antiguru: Good catch, fixed in dcba5b5a8e. Moved the push to after the create_cluster/create_replica loop, so it iterates populated instances and the subsequent update_configuration re-push reaches the now-created replicas. It also lands before dataflows are rendered later in bootstrap, so render-frozen replica flags still take effect.

src/adapter/src/coord/catalog_implications/parsed_state_updates.rs
antiguru: Applied your wording in dcba5b5a8e: the doc now says the row is kept only so it shows up in the tracing::trace! of the parsed update.

src/adapter/src/coord.rs
antiguru: Applied the two-sentence rewrite in dcba5b5a8e.

src/adapter/src/coord.rs
antiguru: Moved that paragraph out of the rustdoc into an inline note next to the update_replica_dyncfg_overrides / update_configuration calls in dcba5b5a8e; the rustdoc now states just the contract.

aljoscha (approved):

antiguru:
Thanks for the reviews!

PR #37153, opened 2026-06-19
persist: tuned Postgres queries with READ COMMITTED isolation

petrosagg:
This PR changes the mechanics of how Postgres tuned queries work in persist's consensus implementation. The new queries are able to:

* Increase throughput by at least 3x (see below)
* Avoid the throughput collapse of the SERIALIZABLE implementation

## Correctness

The way the queries interact with Postgres under READ COMMITTED and the correctness specification of the `Consensus` trait have been modeled in Lean4. In that system it was proven that any execution trace of the READ COMMITTED implementation is equivalent to an execution trace of an idealized linearizable system.

A version of the proof in prose is included as a comment in the source code.

## Performance characteristics 

Under my benchmarking the tuned queries are able to sustain **at least 3x** the throughput of the previous implementation. I say at least because I could not provision enough clusters to make the new configuration fail, so it's very likely even better.

The load test was made up of 20 clusters, each containing 600 trivial materialized views that select from an empty table. Each cluster starts with zero replicas and one by one is brought online, ramping up demand from the consensus implementation as more and more shards are written to.

These benchmarks were performed on a Materialize instance provisioned through our Terraform provider using a `db.r6g.2xlarge` RDS instance and 800cc replicas.

### READ COMMITTED results ([link to Grafana dashboard](https://snapshots.raintank.io/dashboard/snapshot/PaL74EvQIygv3s2EePGBEKbIO5aU4YQm?orgId=0&from=2026-06-25T19:29:45.000Z&to=2026-06-25T20:22:51.185Z&timezone=browser&refresh=5s))

image

In this screenshot we can see that the consensus implementation is able to easily follow the demanded load until the end of the experiment where it continuously sustains ~13k CaS ops/s. Anecdotally while working on these benchmarks there were situations that the sustained throughput was all the way up to 25k CaS ops/s.

The three big dips in throughput are due to the clusters not generating load versus the consensus implementation not being able to serve the demanded load. Supporting evidence for this is that during the dips we see the p99 latencies of CaS operations drop, connections being available in the connection pool, and RDS CPU dropping. The most likely explanation is that environmentd sometimes would struggle to advance the frontier of the singular table all those thousands of materialized views were reading from, stalling all the clusters.

### SERIALIZABLE results ([link to Grafana dashboard](https://snapshots.raintank.io/dashboard/snapshot/g6K99k3CBuDyJjAPk6JY9wsMzQ16Fd3a?orgId=0&from=2026-06-25T20:27:45.000Z&to=2026-06-25T20:53:47.775Z&timezone=browser&refresh=5s))

image

A very different outcome happened under the vanilla serializable queries. Here we see that the system entered a positive feedback loop at around 4.3k CaS ops/s after which throughput plummets and is unable to recover. CaS latencies skyrocket to 30 seconds and no connections are available in the connection pool. The culprit is the very large amount of SSIRead locks that are being taken, which peaked at ~100K.

I have uploaded snapshots of the grafana dashboards with the metrics here:

- [Read committed dashboard](https://snapshots.raintank.io/dashboard/snapshot/PaL74EvQIygv3s2EePGBEKbIO5aU4YQm?orgId=0&from=2026-06-25T19:29:45.000Z&to=2026-06-25T20:22:51.185Z&timezone=browser&refresh=5s)
- [Serializable dashboard](https://snapshots.raintank.io/dashboard/snapshot/g6K99k3CBuDyJjAPk6JY9wsMzQ16Fd3a?orgId=0&from=2026-06-25T20:27:45.000Z&to=2026-06-25T20:53:47.775Z&timezone=browser&refresh=5s)

### DIFF
--- a/src/postgres-client/src/lib.rs
+++ b/src/postgres-client/src/lib.rs
@@ -66,16 +66,54 @@ pub trait PostgresClientKnobs: std::fmt::Debug + Send + Sync {
     fn statement_timeout(&self) -> Duration;
 }
 
+/// The transaction isolation level applied to new connections.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum IsolationLevel {
+    /// `SERIALIZABLE` — the strongest level; the historical default for consensus.
+    Serializable,
+    /// `READ COMMITTED` — for callers (e.g. consensus) whose queries are correct without
+    /// serializable isolation (relying instead on the `PRIMARY KEY` / `FOR UPDATE` / `ON CONFLICT`).
+    ReadCommitted,
+}
+
+impl IsolationLevel {
+    /// The `SET SESSION CHARACTERISTICS` statement that selects this isolation level.
+    fn set_characteristics_sql(self) -> &'static str {
+        match self {
+            IsolationLevel::Serializable => {
+                "SET SESSION CHARACTERISTICS AS TRANSACTION ISOLATION LEVEL SERIALIZABLE"
+            }
+            IsolationLevel::ReadCommitted => {
+                "SET SESSION CHARACTERISTICS AS TRANSACTION ISOLATION LEVEL READ COMMITTED"
+            }
+        }
+    }
+}
+
+/// Resolves the isolation level to apply to a connection. It is invoked once per connection
+/// creation, so a dyncfg-backed resolver lets a change take effect as the pool cycles connections.
+pub type IsolationLevelFn = Arc<dyn Fn() -> IsolationLevel + Send + Sync>;
+
 /// Configuration for creating a [PostgresClient].
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct PostgresClientConfig {
     url: SensitiveUrl,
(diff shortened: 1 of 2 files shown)


src/postgres-client/src/lib.rs
```diff
@@ -66,16 +66,54 @@ pub trait PostgresClientKnobs: std::fmt::Debug + Send + Sync {
     fn statement_timeout(&self) -> Duration;
 }
 
+/// The transaction isolation level applied to new connections.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum IsolationLevel {
+    /// `SERIALIZABLE` — the strongest level; the historical default for consensus.
+    Serializable,
+    /// `READ COMMITTED` — for callers (e.g. consensus) whose queries are correct without
+    /// serializable isolation (relying instead on the `PRIMARY KEY` / `FOR UPDATE` / `ON CONFLICT`).
```
DAlperin, line 75: Stale comment

DAlperin (approved):
LGTM, correctness arguments hold up and match my understanding of the lean4 proof you shared. Thanks.

PR #37155, opened 2026-06-19
expr: Don't panic in the subquery-size guard on a count of zero

def-:
GuardSubquerySize soft-panicked ("subquery counting unexpectedly produced 0") whenever its count was 0, assuming an MIR count never yields 0. That holds for a grouped reduce, but constant folding can prove a scalar subquery's body empty and fold its count to a literal 0, at which point the guard aborted optimization of otherwise-valid SQL.

A count of 0 is benign, the empty subquery decorrelates to NULL via the outer lookup, so treat 0 like 1.

Seen in the RQG runs in https://buildkite.com/materialize/nightly/builds/16829, caused by https://github.com/MaterializeInc/materialize/pull/37049, but the bug was pre-existing and is just showing up now. Nightly run: https://buildkite.com/materialize/nightly/builds/16837

Nightly test run: https://buildkite.com/materialize/nightly/builds/16837 (failures are unrelated)

### DIFF
--- a/src/expr/src/relation/func.rs
+++ b/src/expr/src/relation/func.rs
@@ -22,7 +22,7 @@ use mz_lowertest::MzReflect;
 use mz_ore::cast::CastFrom;
 
 use mz_ore::str::separated;
-use mz_ore::{soft_assert_eq_no_log, soft_assert_or_log, soft_panic_or_log};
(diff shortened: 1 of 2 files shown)


antiguru (approved):
Thank you!

PR #37156, opened 2026-06-19
doc: document the catalog implications model in the adapter guide

aljoscha:
### Motivation

Fills in the empty "Architecture & Key Concepts" section of the adapter guide and adds one correctness invariant, distilled from a review discussion so the learnings are not rediscovered.

### Description

Two things recorded:

1. **The catalog-implications model.** DDL and system changes flow `sequencer -> catalog_transact -> apply_catalog_implications`. The sequencer decides what to write durably, and downstream side effects (in-memory state, controller commands, builtin-table updates) are derived from the committed catalog diff, not from sequencer closures or the input-op side-effect block in `catalog_transact_inner`. Deriving from the diff is what lets a side effect fire whether the change is applied locally or by following another writer, which the multi-`environmentd` direction depends on. Notes the infallibility and consolidation contracts, and which paths are legacy and being migrated away from.

2. **A correctness invariant:** if a reconcile rebuilds downstream state (e.g. a controller's per-replica config) from the catalog working copy, that state must live in the working copy. Pushing it only into a controller in-memory diverges, and the next rebuild silently clears it. Running the side effect as an implication is necessary but not sufficient if the value bypasses the catalog.

Docs only, no code or behavior change.

### DIFF
--- a/doc/developer/guide-adapter.md
+++ b/doc/developer/guide-adapter.md
@@ -7,7 +7,62 @@ decisions.
 
 ## Architecture & Key Concepts
 
-<!-- TODO: fill in -->
+### Catalog changes and their implications
+
+A DDL or system change flows through three conceptual phases:
+
+1. The sequencer (`src/adapter/src/coord/sequencer/`) decides *what to write
+   durably*. It builds a `Vec<catalog::Op>` and calls one of the
+   `catalog_transact*` entry points. It should not reach into controllers or
+   mutate downstream in-memory state directly.
+2. `catalog_transact` (via `catalog_transact_inner`) commits the ops durably and
+   applies them to the in-memory `CatalogState`, producing the committed catalog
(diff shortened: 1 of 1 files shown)


antiguru:
Forward-staleness heads-up: the legacy-paths bullet says `parse_state_update`'s catch-all "covers the system-config kinds, among others." That's true today, but #37151 moves `ReplicaSystemConfiguration` out of the catch-all into a real `ParsedStateUpdateKind` (so the replica-scoped controller push is derived from the diff). Once #37151 merges, this sentence should narrow to the cluster- and environment-scoped system-config kinds; replica-scoped will then be parsed, not `None`. Merge order decides whether it needs a one-line tweak here or there.

antiguru (approved):
Thank you!

mtabebe (approved):

PR #37157, opened 2026-06-19
environmentd: fix flaky test_auth_oidc_fetch_error

def-:
test_auth_oidc_fetch_error aborted its OIDC mock server and then pointed environmentd's configured issuer at the server's just-freed ephemeral port, expecting the JWKS discovery fetch to fail with "failed to fetch OIDC provider configuration".

Under high test parallelism (cargo nextest --test-threads=48) the OS can hand that freed port to another OIDC test's mock server before environmentd performs the discovery fetch. Because every OIDC test shares the same hardcoded kid ("test-key-1"), find_key() then succeeds against the wrong server's JWKS and JWT signature validation fails, so the assertion intermittently observed "failed to validate JWT" instead of the expected fetch error.

Point the configured issuer at a loopback port that is guaranteed to refuse connections (127.0.0.1:1, which is privileged and never bound by the ephemeral-port mock servers), so the discovery fetch fails deterministically regardless of port reuse. The mock server is now only used to mint a validly-signed JWT whose issuer matches the configured one.

Closes SQL-412

### DIFF
--- a/src/environmentd/tests/auth.rs
+++ b/src/environmentd/tests/auth.rs
@@ -2302,21 +2302,40 @@ async fn test_auth_oidc_fetch_error() {
     .unwrap();
 
     let oidc_user = "user@example.com";
-    let issuer = oidc_server.issuer.clone();
-    // Generate a JWT while the server is still up (correct issuer, valid signature).
+    // Point the issuer at a loopback port that is guaranteed to refuse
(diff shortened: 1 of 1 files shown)


SangJunBak (approved):
Ah that's a trippy flake! Thanks for the fix Dennis!

